Issue #1102, opened 2013-05-16
remove CFeatures::get_size()


sonney2k:

karlnapf:
any ideas what the motivation for this was?

lisitsyn:
@karlnapf is it needed somewhere?

sonney2k:
I used this like 10 years ago to alloc memory / know the size of the feature matrix etc. I guess we don't need this anymore.

On March 27, 2015 5:24:18 PM EET, Heiko Strathmann <notifications@github.com> wrote:
>any ideas what the motivation for this was?
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/shogun-toolbox/shogun/issues/1102#issuecomment-86973603

-- 
Sent from Kaiten Mail. Please excuse my brevity.

karlnapf:
ok thanks

PR #1112, opened 2013-05-18
Removed CFeatures::get_size() method (according to ...

vladislav-horbatiuk:
...https://github.com/shogun-toolbox/shogun/issues/1102).


karlnapf:
Any ideas what the motivation for this was?

Issue #1973, opened 2014-03-10
Implement heterogeneous (GPU+CPU) dot product computation routines (Deep learning project)


lisitsyn:
The dot product operation is one of the major building blocks for deep architecture neural networks. The routine implemented in this task should be able to handle batch computation of dot products. For some references see Theano, CUDA, OpenCL, ViennaCL. It is also worth to implement some tests to measure performance and memory specific things. 

Please join the discussion before starting working on any code. We expect to refine the task with further discussion.

nachitoys:
Hi all. Could you let me know if MKL can be performed in GPU? in case it is, there are examples or documentation about it?

Thank you..

karlnapf:
MKL as in multiple kernel learning?

Computing the kernel matrix for sure can be done on the GPU, though that might not be the bottleneck. I think the algorithms themselves (plane-cutting) cannot be parallelised to a large extend, but for sure all the linear algebra going on in there can be done using linalg, which will make things faster for sure.
Are you interested in looking into this with the help of someone?

@sonney2k How many dot-product-like and element-wise operations, and linear solves are there in the MKL code?

nachitoys:
In that case, yes, Im definitely interested in looking into this with the help of someone. Personally I have not reviewed the code yet. Neither I have not more knowledge about GPU parallelization than some indications from Theano, e.g. "in order to go to the GPU, variables must be declared as 'shared'...". I've an idea why it is necessary, but I have never programmed nothing like that.

Furthermore, I have performed some tests by using MKLMulticlass with lists of kernels. These tests were performed over the  fm_train_multiclass_digits corpus. Naturally it goes very slow, which becomes very difficult getting a good version of the code (it is thought even more difficult considering that Im also interested in to learn the widths of the MKLCombinedKernel, as well as higher level hyperkernels). The target dataset (text data) of my research is similar (in size) to the fm_train_multiclass_digits; that is why I think going to the GPU would be useful. 

Said the above, how do you suggest to proceed?

karlnapf:
I guess the first step would be to dive into the code, investigate performance systematically like you already started to do.

From this, it should be clear what the bottlenecks in the code are. Very often, some very simple multicore techniques can massively speed up things. As for GPU, you would have to identify which operations could be put into a GPU. Candidates are mostly operations with (fixed & big) matrices.

karlnapf:
Also have a look into the benchmark folder, there are a few example benchmarks we did in the past

[...]

PR #2125, opened 2014-04-10
dot product (GPU+CPU)

SunilMahendrakar:
basic dot product for #1973


nachitoys:
Hi all. Could you let me know if MKL can be performed in GPU? if yes, there are examples or documentation about it?

Thank you...

PR #2212, opened 2014-05-06
Refactor laplacian

yorkerlin:
@karlnapf  take a look at this.
I will send the link for the notebook tomorrow.

Note that the original implementation of LaplacianInferenceMethod in Shogun used log(lu.determinant()) to compute the log_determinant is not numerical stable. (In fact, this implementation do not follow the GPML code)

Maybe  MatrixOperations.h will be merged in Math.h.
However, I think in that case the Math.h file need to include the Eigen3 header.

Another issue is currently I use MatrixXd and VectorXd to pass variables in  MatrixOperations.h.
maybe SGVector and SGMatrix will be better. (should I use "SGVector &" or "SGVector")
I do now know whether passing SGVector to a function is to copy elements in the SGVector.

### DIFF
--- a/src/shogun/machine/gp/LaplacianInferenceMethod.cpp
+++ b/src/shogun/machine/gp/LaplacianInferenceMethod.cpp
@@ -21,6 +21,7 @@
 #include <shogun/mathematics/Math.h>
 #include <shogun/lib/external/brent.h>
 #include <shogun/mathematics/eigen3.h>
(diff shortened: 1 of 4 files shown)


yorkerlin:
Further clean-up work will be done once I complete the FITC stuff.
For now, I close this first.

karlnapf:
ok!

PR #2498, opened 2014-08-17
Added max-product LP relaxation

Jiaolong:


hushell:
@iglesias Sorry for late reply! I was busy for a final demo for another project, so this PR was suspended. Sorry that I missed this message as well. I am available again. I think I'll make the review done this weekend.

Jiaolong:
@hushell I am thinking about re-factory of GMPLP. Because it is essentially a message passing algorithm,   we may consider inheriting from belief propagation class. Anyway, we can discuss it before you start the review.

hushell:
@Jiaolong: I just went over the MPLP code carefully. I found it's not as clean as what I expected, and its factor implementation (aka Region in their case) is quite different from ours. So I don't think it's easy to reuse our factor implementation and keep their core code only. My suggestions to finish this PR soon are listed as below.
1. Let's not just treat MPLP (max-product LP) as LP relaxation MAP algorithm. In fact, there are many methods can be called LP relaxation, the categorization was a bit wrong. So we can add a new macro, let's say MAX_PRODUCT_LP_RELAXATION. We'll do other types later. 
2. We'll keep their inner data structure, i.e. Region. We don't change its name to EdgeFactor. It's helpful to try MPLP_ver1, which has simpler input/output. I can share my understandings for Region here: 
m_region_inds, lists all nodes participated, e.g., (3,4) is an edge and 3 is a node index, they are both regions (factors/potentials may be more well known). 
m_intersect_inds, lists all regions intersect with this region, stores region indices, e.g., region (3,4)'s index is 1, and region 3's index is 183, region 4's index is 184, then [1, 183, 184] will be stored. 
- UpdateMsgs implements basically the equation in Sontag's thesis (page 104). Note that messages are interpreted as dual variables in MPLP. But the update is different from their ealier conference papers. Here only messages passed from regions to variables, and Region can be viewed as a virtual concept comparing to the old update. 
- Let's replace their multidim_arr to SGNDArray. I think you have done this already. 
- You could keep all their STD stuffs, but should be hided well. Reference to BeliefPropagation.h
- Let's delete their commented code when pulling into Shogun. But make sure copying their first few lines in their README. 

I'll add other thoughts if they came up. If you have time, let's restart this PR, and try to finish it in 2 weeks. Our goal this week is just to import their code and make minimum needed changes.

Jiaolong:
@hushell I am currently using [darwin](http://drwn.anu.edu.au/). The organization of the inference classes in darwin looks elegant, e.g., [here](http://drwn.anu.edu.au/classdrwnMessagePassingMAPInference.html)
I will take your advise and restart this PR as soon as possible.

iglesias:
Hey, guys. Just giving my point here. Inference in PGMs is a big thing and there are already good open source libraries around there. Thus, imho, it might make no sense at all to re-implement all these stuff in Shogun. However, it can be very cool and powerful to connect it with some of the features of Shogun (e.g. sosvm solvers). Maybe an interface to such a library (e.g. Darwin?) like the one we have for OpenCV is a more suitable solution.

hushell:
My experience is that not all inference algorithms can be used for sosvm, e.g., loopy belief propagation. So at very beginning of this project, I chose 3 inference algorithms to implement: Tree BP for tree structure, GraphCuts for binary submodular general graph and MPLP for general graph. These are proved to be nice subroutines for sosvm. Another important reason that we want to re-implement these 3, because all other inference libs have their own graph/factorGraph data structure. We didn't want to tie with any of them. So we implemented our own factorGraph and factorGraphModel for sosvm. Yes, it will be good to implement a converter from other libs to our factorGraph.

hushell:
@Jiaolong I used darwin 2 years ago. I don't suggest to follow or import darwin's code. You'll find darwin's implementation is not as intuitive as it appears. If we want to support a more general inference lib. I'll suggest to go with OpenGM. 

Maybe we should think about how to deal with MPLP, as well as future support for other infernece algorithms. For MPLP, we can do either
1. to wrap MPLP as a 3rd-party, and convert factorGraph to MPLP's Region, which is an idea similar to OpenGM (they have a wrapper for MPLP). Such a converter can be general enough for other libs. 
2. to extract the core inference code of MPLP, and reimplement using our factor representation. It's not very hard if you understand MPLP, which is just one iterative update step. You'll find the core part of MPLP is in UpdateMsgs function (less than 50 lines). I prefer to do this if you have time. 
2.

iglesias:
I don't think it is a good idea to re-implement these inference algorithms. We could make some code that translates from Shogun factors to the ones used in other libraries (something similar to what we have with SGMatrix or CFeatures and OpenCV Mat).

hushell:
@iglesias Yes, this is what we should do to support many other MAP inference algorithms. However, recall that this transform has to be done for each sample in each iteration of sosvm, since factors change as parameters update. For computational efficiency concern, I suggest to reimplement MPLP at least, such that for a general loopy graph, we don't depend on 3rd-party lib. As I said, we have implemented inference algorithms for tree case, and binary loopy graph case.

PR #2668, opened 2015-01-03
Remove warnings in deb3 build.

sanuj:
The warnings are mainly due to documentation and swig. I will push more commits here.

### DIFF
--- a/src/shogun/mathematics/Math.h
+++ b/src/shogun/mathematics/Math.h
@@ -92,8 +92,6 @@ template <class T> struct radix_stack_t
 	uint16_t si;
 };
 
-///** pair */
-
 /** thread qsort */
 template <class T1, class T2> struct thread_qsort
 {
@@ -658,12 +656,13 @@ class CMath : public CSGObject
 		COMPLEX128_ERROR_ONEARG(atan)
 
 		/** Computes arc tangent with 2 parameters
-		 * @param x input
-		 * @return arctan(x/y)
+		 * @param y input(numerator)
+		 * @param x input(denominator)
+		 * @return arctan(y/x)
 		 */
-		static inline float64_t atan2(float64_t x, float64_t y)
+		static inline float64_t atan2(float64_t y, float64_t x)
 		{
-			return ::atan2((double) x, (double) y);
+			return ::atan2((double) y, (double) x);
 		}
 
 		/// atan2(x), x being a complex128_t not implemented
@@ -831,8 +830,25 @@ class CMath : public CSGObject
 			return area;
 		}
 
+		/** Converts string to float
+		 * @param str input string
+		 * @param float_result float pointer
+		 * @return returns true if successful, else returns false
+		 */
 		static bool strtof(const char* str, float32_t* float_result);
+
+		/** Converts string to double
+		 * @param str input string
+		 * @param double_result double pointer
+		 * @return returns true if successful, else returns false
+		 */
 		static bool strtod(const char* str, float64_t* double_result);
+
+		/** Converts string to long double
+		 * @param str input string
+		 * @param long_double_result long double pointer
+		 * @return returns true if successful, else returns false
+		 */
 		static bool strtold(const char* str, floatmax_t* long_double_result);
 
 		/** Computes factorial of input
@@ -1141,11 +1157,18 @@ class CMath : public CSGObject
 		/** Performs a bubblesort on a given matrix a.
 		 * it is sorted in ascending order from top to bottom
 		 * and left to right
-		 * @param a
-		 * @param cols
-		 * @param sort_col
+		 * @param a matrix to be sorted
+		 * @param cols number of columns
+		 * @param sort_col 
 		 */
 		static void sort(int32_t *a, int32_t cols, int32_t sort_col=0);
+
+		/** Performs a bubblesort on a given array a.
+		 * it is sorted in ascending order from left to right
+		 * @param a array to be sorted
+		 * @param idx index array
+		 * @param N length of array
+		 */
 		static void sort(float64_t *a, int32_t*idx, int32_t N);
 
 		/** Performs a quicksort on an array output of length size
@@ -1216,7 +1239,7 @@ class CMath : public CSGObject
 
 		/** Performs a in-place radix sort in ascending order
 		 * @param array array to be sorted
-		 * @size size size of array
+		 * @param size size of array
 		 */
 		template <class T>
 			inline static void radix_sort(T* array, int32_t size)
@@ -1224,11 +1247,10 @@ class CMath : public CSGObject
 				radix_sort_helper(array,size,0);
 			}
 
-		/*
-		 * Inline function to extract the byte at position p (from left)
+		/** Extract the byte at position p (from left)
 		 * of a 64 bit integer. The function is somewhat identical to
 		 * accessing an array of characters via [].
-		 * @param word
+		 * @param word input from which byte is extracted
 		 * @param p position
 		 * @return byte
 		 */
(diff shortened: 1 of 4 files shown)


src/shogun/mathematics/Math.h
```diff
@@ -658,7 +658,8 @@ class CMath : public CSGObject
 		COMPLEX128_ERROR_ONEARG(atan)
 
 		/** Computes arc tangent with 2 parameters
-		 * @param x input
+		 * @param x input(numerator)
+		 * @param y input(denominator)
 		 * @return arctan(x/y)
 		 */
 		static inline float64_t atan2(float64_t x, float64_t y)
```
iglesias: If we are doing changes here, I think it might be worth to change this to atan2(y,x) since it is much more common imo.

src/shogun/mathematics/Math.h
```diff
@@ -1141,11 +1159,18 @@ class CMath : public CSGObject
 		/** Performs a bubblesort on a given matrix a.
 		 * it is sorted in ascending order from top to bottom
 		 * and left to right
-		 * @param a
-		 * @param cols
-		 * @param sort_col
+		 * @param a matrix to be sorted
+		 * @param cols number of columns
+		 * @param sort_col
```
iglesias: Can we get a short description for this parameter?

src/shogun/mathematics/Math.h
```diff
@@ -1224,8 +1249,7 @@ class CMath : public CSGObject
 				radix_sort_helper(array,size,0);
 			}
 
-		/*
-		 * Inline function to extract the byte at position p (from left)
+		/** Inline function to extract the byte at position p (from left)
```
iglesias: Minor suggestion, I would start directly with "Extract the byte ...".

src/shogun/mathematics/Math.h
```diff
@@ -1224,8 +1249,7 @@ class CMath : public CSGObject
...
 		 * of a 64 bit integer. The function is somewhat identical to
 		 * accessing an array of characters via [].
 		 * @param word
```
iglesias: Short description for this parameter.

src/shogun/mathematics/Math.h
```diff
@@ -1245,7 +1269,11 @@ class CMath : public CSGObject
 			return uint8_t(0);
 		}
 
-		/// Helper function for radix sort.
+		/** Helper function for radix sort.
+		 * @param array
+		 * @param size
+		 * @param i
```
iglesias: Short descriptions for these ones.

src/shogun/mathematics/Math.h
```diff
@@ -1570,13 +1598,17 @@ class CMath : public CSGObject
...
 
-
+		/// helper function for parallel_qsort_index.
 		template <class T1,class T2>
 			static void* parallel_qsort_index(void* p);
 
 
-		/* finds the smallest element in output and puts that element as the
-		   first element  */
+		/** finds the smallest element in output and puts that element as the
```
iglesias: Start with capitals (just to be consistent with the doc of the other functions).

sanuj:
@iglesias I know these are swig warnings. Do you know how to sort these out?
```
src/shogun/lib/SGNDArray.h:116: Warning 367: operator*= ignored
src/shogun/mathematics/Math.h:1192: Warning 314: 'end' is a ruby keyword, renaming to 'C_end'
```

src/shogun/mathematics/Math.h
sanuj: I'm not sure what this parameter is doing exactly. Some help :) I don't seem to understand how the function is sorting. I know what is bubble sort. Not sure how the numbers are being swapped (vertically, horizontally or diagonally?)

iglesias:
It must the integer variable called `end` in the `binary_search` routine.

sanuj:
Shall I rename it to `c_end` and leave a comment that it is to avoid swig warning for ruby?

iglesias:
As you prefer, this warning is not extremely important. In case of renaming, I would prefer another name different from c_end that has more sense, c_end with the comment is fine too.

sanuj:
@iglesias I guess deleting these unused variables won't be a good idea to remove these [warnings](http://buildbot.shogun-toolbox.org/builders/deb3%20-%20modular_interfaces/builds/2576/steps/examples%20and%20unit%20tests/logs/warnings%20%2859%29). What do you think?

iglesias:
Which ones are the unused variables?

sanuj:
For eg- a warning says that:
```
/build/examples/undocumented/csharp_modular/kernel_combined_modular.cs(51,15): warning CS0219: The variable `km_train' is assigned but its value is never used
```
so km_train is assigned but is not used in the code anywhere in kernel_combined_modular.cs. I think these are examples so it is fine this way. Not sure though.

sanuj:
Moreover, the warnings in lua modular ([link](https://customspaces.com/office/DhXb6EKlE9/github-office-san-francisco/)) are SWIG warnings which can be removed by using `%ignore` or `ifndef SWIG` as swig does not support overloading functions ([link](http://www.swig.org/Doc1.3/SWIGPlus.html#SWIGPlus_nn25)). We can also rename these functions by using `%rename` but I think it's not necessary to remove these warnings as we are not getting them in python-modular etc. Your thoughts?

iglesias:
I also think that warnings for unused variables in the examples like the one in kernel_combined_modular.cs are fine. This showcases some of the things that can be done from modular interfaces.

iglesias:
Can you fix the first link in your last comment above?

src/shogun/mathematics/Math.h
```diff
@@ -658,10 +656,11 @@ class CMath : public CSGObject
 		COMPLEX128_ERROR_ONEARG(atan)
 
 		/** Computes arc tangent with 2 parameters
-		 * @param x input
+		 * @param y input(denominator)
```
iglesias: Change to y input (numerator)

src/shogun/mathematics/Math.h
```diff
@@ -658,10 +656,11 @@ class CMath : public CSGObject
...
+		 * @param x input(numerator)
```
iglesias: Change to x input (denominator)

src/shogun/mathematics/Math.h
```diff
@@ -658,10 +656,11 @@ class CMath : public CSGObject
...
 		 * @return arctan(x/y)
 		 */
-		static inline float64_t atan2(float64_t x, float64_t y)
+		static inline float64_t atan2(float64_t y, float64_t x)
 		{
 			return ::atan2((double) x, (double) y);
```
iglesias: change to atan2((double) y, (double) x);

iglesias: http://en.cppreference.com/w/cpp/numeric/math/atan2

sanuj:
@iglesias Oh Sorry! :-( That was a mistake. My apologies. I meant the warnings in lua modular ([link](http://buildbot.shogun-toolbox.org/builders/deb3%20-%20modular_interfaces/builds/2576/steps/lua%20modular/logs/warnings%20%2869%29)). Warning 509.

iglesias:
Regarding the warnings you mention above in lua modular, I think that sqrt and pow do not need to be exposed via swig, very likely all the languages we provide bindings for contain some math library with their own implementation of these functions so I'd say we do not need to provide them. Probably the same applies for random and normal_random, although these last ones might be arguable.

sanuj:
@iglesias I think it will be fine if we make only one implementation visible to SWIG (and hide the rest by using `SWIG` macro) as these warnings are due to overloading functions. For eg-
````
void func(float x);
void func(double y);
```
For scripting languages `float` and `double` will be the same thing so SWIG ignores `func(double y)` and throws up a warning. This can be corrected by doing -
```
void func(float x);
ifndef SWIG;
void func(double y);
endif;
```
But I'm not sure if this is the right approach.

Issue #2638, opened 2014-12-16
create unit test for MKLRegression


karlnapf:
With sensible data that makes sure things work as expected

sanuj:
http://asi.insa-rouen.fr/enseignants/~arakoto/code/mklindex.html
I was planning to use this but it doesn't seem to work.

sanuj:
SimpleMKL (the previous one) is a Matlab Toolbox. I will try SMO-MKL now.
http://research.microsoft.com/en-us/um/people/manik/code/smo-mkl/download.html

iglesias:
Hey, @sanuj. I just had a look at the unit test and you're definitely right; `labels_test` is not used. If I am not missing anything, the lines filling in the `lab_test` vector and creating the features from it can be removed. This test was probably created from another example where the test labels were used.

@karlnapf will have to tell you about the easy script, I don't know about that one. As I understand it, it has nothing to do with easysvm but the results used in the test are the output of LibSVM.

PR #2669, opened 2015-01-05
Remove redundant labels_test/lab_test in LibSVR_unittest

sanuj:
#2638 @iglesias Please review.


iglesias:
Merging as the first job failing in Travis (time out) is not caused by the changes here and the second job executes the unit tests without failures. Thanks!

PR #2670, opened 2015-01-13
Remove SGVector::fequal

sanuj:
@iglesias Please review.

### DIFF
--- a/src/shogun/lib/SGVector.cpp
+++ b/src/shogun/lib/SGVector.cpp
@@ -784,13 +784,6 @@ float32_t SGVector<float32_t>::sum_abs(float32_t* vec, int32_t len)
 }
 #endif
 
-/// return sum(abs(vec))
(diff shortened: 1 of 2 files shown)


iglesias:
Did you check for uses of it in modular examples and notebooks? Note that SGVector changes name in modular interfaces so grep for fequal, .fequal, or so, rather than for SGVector::fequal.

sanuj:
Yes I did. Only `fequal` from `CMath` is being used everywhere. If  that would have been the case then I would have got some error while building and testing on my machine. But it compiled properly and all the test cases also ran successfully.

iglesias:
There would have not necessarily been an error if it were used in the notebooks and (maybe) part of the examples (e.g. graphical ones?) as they are not covered by tests.

sanuj:
@iglesias Oh, alright! Thanks :)

PR #2671, opened 2015-01-15
Added missing HAVE_EIGEN3 guard in Statistics unittest

lambday:
Caught while testing linalg native backend without eigen3/viennacl being present.


lambday:
Travis error unrelated. Merging.

Issue #2667, opened 2014-12-29
Native implementations in Linalg.


sanuj:
@karlnapf @lambday I would like to work on this. Currently `HAVE_LINALG_LIB` is defined provided `eigen3` & `viennacl` are found, and supported c++11 compiler is used. We can change it to defining `HAVE_LINALG_LIB` only when the supported c++11 compiler is used. This can be achieved by altering `shogun/CMakeLists.txt` alone.

We can give runtime errors for the operations that can't be implemented natively in linalg. Please list down the operations that can be implemented natively (i guess Vector/Matrix dot and sum are one of them).

sanuj:
Nice to hear from you @lambday :D. I think not disturbing `HAVE_LINALG_LIB` is a good idea. Looking forward to your modifications.

lambday:
@sanuj check PR #2672. I have just added dot to check if the NATIVE backend thing is working. Once this change is merged, you can start working on native implementations for other operations in similar fashion. 

Please make sure that everything compiles even in absence of eigen3/viennacl. Unit-tests for NATIVE should work even when we don't have any external libs.

Another cleanup work would be to use this `linalg::dot(a, b)` instead of the existing dot methods (replace everywhere `dot` operation is used _without specifying any backend_ since it is safe now).

sanuj:
@lambday Great! The PR looks good :) I'll start the clean up work and add more native implementations in a similar fashion as soon as the PR is merged. Loads of functions can be cleaned up and more can be implemented natively (list is in the mailing list, Should I add it here just so we have a record?).

lambday:
@sanuj cool. So the key to add native implementation to linalg is that we'll keep those operations (in modules) outside of the `HAVE_LINALG_LIB` guard. No other special tricks required. Please keep in mind that we don't intend to provide native implementation for _everything_ . Just the basic trivial things (most of them were already there in SGVector and SGMatrix). Before you push, you _must_ check locally (compile and run the unit-tests)
- with all the backends configured 
- without the backends configured
(I do a silly thing - I just temporarily rename the include dirs of eigen3 and viennacl to something else so that cmake can't find them. Maybe there is a better way to turn this off just by cmake.)

Don't worry about adding the list here. We have the mailing list archived already :)
Wait for this PR to be merged first :) Once I hear green from @iglesias I will merge it.

lambday:
[...]

Issue #2582, opened 2014-10-27
SGVector cleanup.


iglesias:
Remove all the methods producing clutter in `SGVector`. This includes (although it is not limited to) math operations like `linspace`, `dot`, and other operations like `find`. The objective is to have an `SGVector` as lean as possible.

Systematically, the work to be done for each method reduced is:
1) Move the method from `SGVector` to another class, like for instance to `CMath` or to whichever it makes sense the most.
2) Refactor Shogun's internal code.
3) Refactor examples (libshogun and interface examples, if any).
4) In case the method already had unit tests in `SGVector`, then refactor them (and move them). Otherwise, write some good unit tests.

See this pull request to gain understanding of what this task consists of, #2579.

curiousguy13:
@iglesias  would it be fine , if I moved range_fill to CMath ?

iglesias:
@curiousguy13, it sounds good to me, both range_fill and range_fill_vector should go out of SGVector. Do something nice with them (perhaps, simplifying to only one method) and move to CMath. Keep in mind unit tests. Looking forward to see the pull request.

lambday:
@iglesias I think range_fill, zeros, these are perfect to be shifted to linalg instead.

iglesias:
It sounds good!

Hephaestus12:
I would like to work on this. Shall I start by moving add to CMath?

gf712:
> I would like to work on this. Shall I start by moving add to CMath?

You shouldn’t move to CMath but to linalg instead. However linalg already has add. But it would be good to remove add from SGVector. AFAIK CMath (now Math) should be dropped.

Hephaestus12:
This is for the method add() in SGVector.
I don't think simply removing it and replacing its calls with the add method in the Linalg framework will help. For the following reasons:

In file:` src/shogun/mathematics/linalg/backend/eigen/BasicOps.cpp` :
```
#define BACKEND_GENERIC_IN_PLACE_ADD(Type, Container)                          \
	void LinalgBackendEigen::add(                                              \
	    const Container<Type>& a, const Container<Type>& b, Type alpha,        \
	    Type beta, Container<Type>& result) const                              \
	{                                                                          \
		add_impl(a, b, alpha, beta, result);                                   \
	}
```
`add()` calls `add_impl()`


In file:` src/shogun/mathematics/linalg/backend/eigen/BasicOps.cpp` :
```
template <typename T>
[...]

PR #2509, opened 2014-08-20
added convolution to linalg

khalednasr:
@lisitsyn, @vigsterkr Pushed #2418 again, this time against the develop branch instead of feature/linalg

PR #2672, opened 2015-01-15
Added Shogun's native dot impl to linalg under NATIVE backend

lambday:
Addresses the basics for issue #2667 and #2582

### DIFF
--- a/src/shogun/mathematics/linalg/internal/modules/Core.h
+++ b/src/shogun/mathematics/linalg/internal/modules/Core.h
@@ -32,6 +32,8 @@
 #ifndef CORE_H_
 #define CORE_H_
 
+#ifdef HAVE_LINALG_LIB
+
 #include <shogun/mathematics/linalg/internal/implementation/ElementwiseSquare.h>
 #include <shogun/mathematics/linalg/internal/implementation/MatrixProduct.h>
 #include <shogun/mathematics/linalg/internal/implementation/Add.h>
@@ -45,18 +47,18 @@ namespace shogun
 namespace linalg
 {
 
-/** Performs matrix multiplication 
- * 
+/** Performs matrix multiplication
+ *
  * @param A First matrix
  * @param B Second matrix
  * @param C Result of the operation
  * @param transpose_A Whether to the transpose of A should be used instead of A
  * @param transpose_B Whether to the transpose of B should be used instead of B
- * @param overwrite If true, the values in C are overwritten with the result, 
+ * @param overwrite If true, the values in C are overwritten with the result,
  * otherwise, the result is added to the existing values
  */
 template <Backend backend=linalg_traits<Core>::backend,class Matrix>
-void matrix_product(Matrix A, Matrix B, Matrix C, 
+void matrix_product(Matrix A, Matrix B, Matrix C,
 	bool transpose_A=false, bool transpose_B=false, bool overwrite=true)
 {
 	implementation::matrix_product<backend, Matrix>::compute(A, B, C, transpose_A, transpose_B, overwrite);
@@ -64,7 +66,7 @@ void matrix_product(Matrix A, Matrix B, Matrix C,
 
 /** Performs the operation C = alpha*A + beta*B. Works for both matrices and vectors */
 template <Backend backend=linalg_traits<Core>::backend,class Matrix>
-void add(Matrix A, Matrix B, Matrix C, 
+void add(Matrix A, Matrix B, Matrix C,
 	typename Matrix::Scalar alpha=1.0, typename Matrix::Scalar beta=1.0)
 {
 	implementation::add<backend, Matrix>::compute(A, B, C, alpha, beta);
@@ -72,7 +74,7 @@ void add(Matrix A, Matrix B, Matrix C,
 
 /** Performs the operation C = alpha*A - beta*B. Works for both matrices and vectors */
 template <Backend backend=linalg_traits<Core>::backend,class Matrix>
-void subtract(Matrix A, Matrix B, Matrix C, 
+void subtract(Matrix A, Matrix B, Matrix C,
 	typename Matrix::Scalar alpha=1.0, typename Matrix::Scalar beta=1.0)
 {
 	implementation::add<backend, Matrix>::compute(A, B, C, alpha, -1*beta);
@@ -95,7 +97,7 @@ void elementwise_product(Matrix A, Matrix B, Matrix C)
 /**
  * Wrapper method for internal implementation of square of co-efficients that works
  * with generic dense matrices.
- * 
+ *
  * @param m the matrix whose squared co-efficients matrix has to be computed
  * @return another matrix whose co-efficients are \f$m'_{i,j}=m_(i,j}^2\f$
  * for all \f$i,j\f$
@@ -109,7 +111,7 @@ typename implementation::elementwise_square<backend,Matrix>::ReturnType elementw
 /**
  * Wrapper method for internal implementation of square of co-efficients that works
  * with generic dense matrices.
- * 
+ *
  * @param m the matrix whose squared co-efficients matrix has to be computed
  * @param result Pre-allocated matrix for the result of the computation
  */
@@ -145,4 +147,5 @@ void convolve(Matrix X, Matrix W, Matrix Y, bool flip = false,
 }
 
 }
+#endif // HAVE_LINALG_LIB
 #endif // CORE_H_
--- a/src/shogun/mathematics/linalg/internal/modules/SpecialPurpose.h
+++ b/src/shogun/mathematics/linalg/internal/modules/SpecialPurpose.h
@@ -29,11 +29,10 @@
  * either expressed or implied, of the Shogun Development Team.
  */
 
-/** Contains special purpose, algorithm specific functions. Uses the same 
- * backend as the Core module 
- */
 #ifndef SPECIAL_PURPOSE_H_
 #define SPECIAL_PURPOSE_H_
+
+#ifdef HAVE_LINALG_LIB
 
 #include <shogun/mathematics/linalg/internal/implementation/SpecialPurpose.h>
 
@@ -43,9 +42,12 @@ namespace shogun
 namespace linalg
 {
 
+/** Contains special purpose, algorithm specific functions. Uses the same
+ * backend as the Core module
+ */
 namespace special_purpose
 {
-	
+
 /** Applies the elementwise logistic function f(x) = 1/(1+exp(-x)) to a matrix */
 template <Backend backend=linalg_traits<Core>::backend,class Matrix>
 void logistic(Matrix A, Matrix result)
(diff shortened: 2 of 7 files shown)


iglesias:
It is looking good to me.

src/shogun/mathematics/linalg/linalg.h
```diff
@@ -35,40 +35,40 @@
...
+ * not set explicitly via cmake options. A Native backend is also defined to
+ * accommodate Shogun's native implementation of some operations.
  *
  * The enum defines these backends in order of priority as default backend, as
- * in, first defined one will be used as default
- *
- * Note - Currently EIGEN3 is the default (if its available)
+ * in, first defined one will be used as default.
  *
+ * Note - Currently EIGEN3 is the default (if it is available).
```
iglesias: Should this comment be updated?

lambday: @iglesias umm I think no.. it is still true :D

lambday:
@iglesias @lisitsyn I think we better merge this one _after_ we merge khaled's convolution patch #2509. I think he is not available for rebase. Since this patch also contains changes in Core.h I think we can get merge conflicts. I will then rebase and push again.

src/shogun/mathematics/linalg/linalg.h
iglesias: I'm afraid I am not completely understanding what initialisation of the last entry of the enum to 0 does. Aren't eigen and viennacl (if available) as well as native as default all getting the value zero?

lambday: @iglesias if both are available, then the enum would be defined as
```c++
enum class Backend
{
    EIGEN3, // will be 0
    VIENNACL, // will be 1
    NATIVE, // will be 2
    DEFAULT = 0 // therefore EIGEN3
};
```
If eigen3 and viennacl is not found, the enum is
```c++
enum class Backend
{
    NATIVE, // will be 0
    DEFAULT = 0 // therefore NATIVE
};
```
We're relying on the fact that enum by default starts from 0 and keeps on increasing by 1.

lambday: `DEFAULT` will be same as the first thing defined. So if both Eigen3 and ViennaCL are there, it is Eigen3. If only ViennaCL is there it is ViennaCL. If nothing is there it is NATIVE (which is limited to only a few operations that we'll provide native implementation).

sanuj: Of what I understand,
- if we have eigen and viennacl then 
`EIGEN3=0, VIENNACL=1, NATIVE=2 and DEFAULT=0`
in which case Backend::DEFAULT is EIGEN3
- if we have eigen and not viennacl then
`EIGEN3=0, NATIVE=1, DEFAULT=0`
in which case `Backend::DEFAULT` is `EIGEN3`
- if we have viennacl and not eigen then
`VIENNACL=0, NATIVE=1, DEFAULT=0`
in which case `Backend::DEFAULT` is `VIENNACL`
- if we neither have eigen nor viennacl then,
`NATIVE=0, DEFAULT=0`
in which case `Backend::DEFAULT` is `NATIVE`

@lambday ?

lambday: @sanuj yep. that's pretty much it :)

iglesias: I understood the enum above now, @lambday. Thanks for the explanation!

lambday: @iglesias thanks for merging the convolution patch. I assume it is okay then if I rebase and merge this patch then.

Please check the wiki on linalg as well. I could really use your suggestions on that.

iglesias: Hey, @lambday. A couple of comments regarding the linalg wiki:

* When the Redux and Core modules are introduced [here](https://github.com/shogun-toolbox/shogun/wiki/README_linalg#relying-on-globally-set-linalg-backend), maybe it would be nice to mention what is different between both? Just to give an idea where the names come from.
* I saw the usage of linalg in CustomKernel.cpp. I wonder why the `HAVE_LINALG_LIB` guard is required. I guess it is because there is a cmake option that allows to switch on/off compilation of linalg. In that case, maybe that can be mention in the wiki as well.

iglesias: It is very cool to have this documented in the wiki by the way. Awesome :-)

lambday: @iglesias thanks for your comments.
- yes you're right, I'll add some comments on the module names. It is not very clear from those names. IMO redux is for those operations which takes a matrix/vector and returns a scalar, e.g. dot, sum etc. rest generic util things should go to core.
- actually the guard is not required. `HAVE_LINALG_LIB` is defined whenever we have `HAVE_EIGEN3` _or_ `HAVE_VIENNACL` defined. currently there is no cmake option to turn linalg off. I think only `HAVE_EIGEN3` check will suffice there. I'll take care of that.

Thanks. I hope the wiki will be useful :) will finish it off by tonight.

iglesias: It sounds like we can get rid of `HAVE_LINALG_LIB` completely, right? With completely I mean everywhere and not only in CustomKernel.cpp. Operations in native will always be available so no guard required wherever they are used, and eigen3/viennacl stuff should always have their own have_eigen3/viennacl.

lambday: @iglesias actually we don't want to implement _everything_ natively, e.g. we most certainly don't want to write our own self adjoint eigensolver or SVD/QR/Cholesky. `HAVE_LINALG_LIB` guard is useful for those methods which are available _only_ when we _do_ have an external linalg lib. In linalg modules we're guarding those operations with `HAVE_LINALG_LIB` which do not have their native implementations (this modification is done in this patch only). I think it is still useful. A use-case is when we just want the result for some of these operations and don't wanna bother about which backend was used (i.e. relying on global backend setup).

In CustomKernel.cpp I'm explicitly using eigen3 backend because when I added this, ViennaCL specializations were not added till then for those methods. We can most certainly do native implementation for `sum` methods, and, in that case, then we can get rid of the guard in CustomKernel. Otherwise, as of now, we can get rid of `HAVE_EIGEN3` guard and remove the specific backend Eigen3 in those `sum` methods - it will then use Eigen3/ViennaCL - whichever is available.

lambday: I think I should mention about proper usage of `HAVE_LINALG_LIB` in the wiki as well as the limitations while using `NATIVE` backend more specifically. Thanks for your inputs :)

lambday: I thought of putting a matrix there in the wiki - mentioning the currently supported operations and which backends we can use for that. I hope that would be helpful.

iglesias: Oh, I got now the meaning of `HAVE_LINALG_LIB`; it refers to having an external library e.g. Eigen3 or ViennaCL. So far, I had read it as having the Shogun internal linalg library and that's why I got confused :-)

lambday: Yeah the meaning is not very clear. I should explain this macro properly in the wiki :)
BTW I just rebased this branch. Shall we merge it after travis is finished?

lambday: agh there are some errors due to c++11 code being exposed.. will fix it.

sanuj: @lambday How much work is left here?

lambday: @sanuj just addition of 4 lines. Sorry I just have been lazy P Will push by tonight.

sanuj: Cool :) i have some holidays now so i would get time to work on the cleanup
and native implementations. That's why i was asking. Hope you guys merge
this soon.

lambday: @sanuj yeah I guess @iglesias is fine with it (once travis stops whining). I'll merge this after I push the changes. Glad to see that you're psyched to work on this :D 

Go `linalg` !

sanuj:
@lambday what's up? :)

lambday:
Done. Travis failure not related. I'll ask @sonney2k if its okay to merge this patch right now.

src/shogun/mathematics/linalg/internal/modules/Core.h
```diff
@@ -45,18 +47,18 @@ namespace shogun
...
  * @param C Result of the operation
  * @param transpose_A Whether to the transpose of A should be used instead of A
  * @param transpose_B Whether to the transpose of B should be used instead of B
- * @param overwrite If true, the values in C are overwritten with the result, 
+ * @param overwrite If true, the values in C are overwritten with the result,
  * otherwise, the result is added to the existing values
  */
 template <Backend backend=linalg_traits<Core>::backend,class Matrix>
-void matrix_product(Matrix A, Matrix B, Matrix C, 
+void matrix_product(Matrix A, Matrix B, Matrix C,
```
karlnapf: @yorkerlin consider to start using such things, as they become available

src/shogun/mathematics/linalg/internal/modules/SpecialPurpose.h
```diff
@@ -67,15 +69,15 @@ void rectified_linear(Matrix A, Matrix result)
...
 	implementation::special_purpose::multiply_by_rectified_linear_derivative<backend, Matrix>::compute(A, C);
 }
 
-/** Applies the softmax function inplace to a matrix. The softmax function is 
- * defined as \f$ f(A[i,j]) = \frac{exp(A[i,j])}{\sum_i exp(A[i,j])} \f$ 
+/** Applies the softmax function inplace to a matrix. The softmax function is
+ * defined as \f$ f(A[i,j]) = \frac{exp(A[i,j])}{\sum_i exp(A[i,j])} \f$
  */
 template <Backend backend=linalg_traits<Core>::backend,class Matrix>
 void softmax(Matrix A)
```
karlnapf: @yorkerlin might be useful for the multiclass logistic GPs

karlnapf:
GREAT work guys :)
@lambday will read your tutorial for linalg these days

src/shogun/mathematics/linalg/internal/modules/SpecialPurpose.h
yorkerlin: Yes, I may use in near future once the four FITC methods are implemented.
There are two additional inference methods (all about Emti's KL dual inference method) to be added in the multiclass logistic GPs.  I will do it once I get time.

yorkerlin: BTW, @karlnapf  in theory, it seems FITC could work for the multiclass logistic GP. Not sure whether it is robust or not. Maybe, I take a look at this later.

lambday:
@karlnapf thanks. Lot more to do. Is this a good time for merging a new patch? I'll merge it then.

lambday:
Merging this one.

lambday:
@sanuj you can now start working on the things you have planned. Please add one method per PR (easy to review). Proper unit testing must.

sanuj:
Great! Yeah I'll keep that in mind @lambday Thanks!

src/shogun/mathematics/linalg/internal/modules/SpecialPurpose.h
karlnapf: I guess my point was more to use the logit and softmax implementation from linalg in the CLogitLikelihood and CSoftMaxLikelihood. So can change existing implementations to use linalg

FITC for multiclass might not work so well, we have to check, but lets discuss somewhere else as off topic here.

PR #2676, opened 2015-01-22
add the FITC framework 1

yorkerlin:
@karlnapf 
take a look at this
I will update the framework during these days


karlnapf:
Now starting to look at this....

src/shogun/kernel/GaussianARDKernel.cpp
```diff
@@ -23,16 +21,18 @@ CGaussianARDKernel::CGaussianARDKernel() : CLinearARDKernel()
 
 
 CGaussianARDKernel::CGaussianARDKernel(int32_t size, float64_t width)
-		: CLinearARDKernel(size), m_width(width)
```
karlnapf: I like these clean ups!

src/shogun/kernel/GaussianARDKernel.cpp
```diff
@@ -23,16 +21,18 @@ CGaussianARDKernel::CGaussianARDKernel() : CLinearARDKernel()
...
+		: CLinearARDKernel(size)
 {
 	init();
+	set_width(width);
 }
 
-CGaussianARDKernel::CGaussianARDKernel(CDenseFeatures<float64_t>* l,
-		CDenseFeatures<float64_t>* r, int32_t size, float64_t width)
-		: CLinearARDKernel(size), m_width(width)
+CGaussianARDKernel::CGaussianARDKernel(CFeatures* l,
```
karlnapf: I guess DotFeatures?

src/shogun/kernel/GaussianARDKernel.cpp
```diff
@@ -65,22 +64,9 @@ CGaussianARDKernel* CGaussianARDKernel::obtain_from_generic(CKernel* kernel)
 
 float64_t CGaussianARDKernel::compute(int32_t idx_a, int32_t idx_b)
 {
-	REQUIRE(lhs && rhs, "Features not set!\n")
```
karlnapf: the require is not necessray because this is a private method?
Or do you simply check at the places where the user defines the features?

src/shogun/kernel/GaussianARDKernel.cpp
```diff
@@ -65,22 +64,9 @@ CGaussianARDKernel* CGaussianARDKernel::obtain_from_generic(CKernel* kernel)
...
-		((CDenseFeatures<float64_t>*) rhs)->get_feature_vector(idx_b);
-
-	REQUIRE(avec.vlen==bvec.vlen, "Number of right and left hand "
-			"features must be the same\n");
-
-	float64_t result=0;
-
-	for (index_t i = 0; i < avec.vlen; i++)
-		result += CMath::pow((avec[i]-bvec[i])*m_weights[i], 2);
+	float64_t result=distance(idx_a,idx_b);
```
karlnapf: Can you please make sure we re-use the computational tricks from the Gaussian kernel?
Best would be to directly use the code.

src/shogun/kernel/GaussianARDKernel.cpp
```diff
@@ -65,22 +64,9 @@ CGaussianARDKernel* CGaussianARDKernel::obtain_from_generic(CKernel* kernel)
...
 
-	return CMath::exp(-result/m_width);
+	return CMath::exp(-result);
```
karlnapf: where is width?

src/shogun/kernel/GaussianARDKernel.cpp
```diff
@@ -96,10 +82,9 @@ SGMatrix<float64_t> CGaussianARDKernel::get_parameter_gradient(
 		{
 			for (index_t k=0; k<num_rhs; k++)
 			{
-				SGVector<float64_t> avec=
-					((CDenseFeatures<float64_t>*) lhs)->get_feature_vector(j);
-				SGVector<float64_t> bvec=
-					((CDenseFeatures<float64_t>*) rhs)->get_feature_vector(k);
+
+				SGVector<float64_t> avec=((CDotFeatures *)lhs)->get_computed_dot_feature_vector(j);
```
karlnapf: so it *is* DotFeatures, then you can change that above

src/shogun/kernel/GaussianARDKernel.cpp
```diff
@@ -122,21 +107,10 @@ SGMatrix<float64_t> CGaussianARDKernel::get_parameter_gradient(
...
-				SGVector<float64_t> avec=
-					((CDenseFeatures<float64_t>*) lhs)->get_feature_vector(j);
-				SGVector<float64_t> bvec=
-					((CDenseFeatures<float64_t>*) rhs)->get_feature_vector(k);
-
-				REQUIRE(avec.vlen==bvec.vlen, "Number of right and left hand "
-						"features must be the same\n");
-
-				float64_t result=0;
+				float64_t result=distance(j,k);
```
karlnapf: good that you re-use code here. Please also consider what I said above about re-using code from the GaussianKernel

src/shogun/kernel/GaussianARDKernel.cpp
```diff
@@ -148,3 +122,18 @@ SGMatrix<float64_t> CGaussianARDKernel::get_parameter_gradient(
 		return SGMatrix<float64_t>();
 	}
 }
+
+float64_t CGaussianARDKernel::distance(int32_t idx_a, int32_t idx_b)
+{
+	REQUIRE(lhs && rhs, "Features not set!\n")
```
karlnapf: Please check LHS and RHS sperately. Good for error message

src/shogun/kernel/GaussianARDKernel.cpp
```diff
@@ -148,3 +122,18 @@ SGMatrix<float64_t> CGaussianARDKernel::get_parameter_gradient(
...
+	
+	SGVector<float64_t> avec=((CDotFeatures *)lhs)->get_computed_dot_feature_vector(idx_a);
```
karlnapf: If you are checking for NULL you should maybe also check types here?

src/shogun/kernel/GaussianARDKernel.cpp
```diff
@@ -148,3 +122,18 @@ SGMatrix<float64_t> CGaussianARDKernel::get_parameter_gradient(
...
+	SGVector<float64_t> bvec=((CDotFeatures *)rhs)->get_computed_dot_feature_vector(idx_b);
+	
+	REQUIRE(avec.vlen==bvec.vlen, "Number of right and left hand "
```
karlnapf: Please add the numbers so that the user knows

src/shogun/kernel/GaussianARDKernel.cpp
```diff
@@ -148,3 +122,18 @@ SGMatrix<float64_t> CGaussianARDKernel::get_parameter_gradient(
...
+			"features must be the same\n");
+	float64_t result=0.0;
+	for (index_t i=0; i<avec.vlen; i++)
```
karlnapf: this can go to linalg, it is a very common call.
@lambday ? thoughts?

src/shogun/kernel/GaussianARDKernel.h
```diff
@@ -43,7 +43,7 @@ class CGaussianARDKernel: public CLinearARDKernel
 	 * @param size cache size
 	 * @param width kernel width
 	 */
-	CGaussianARDKernel(CDenseFeatures<float64_t>* l, CDenseFeatures<float64_t>* r,
+	CGaussianARDKernel(CFeatures* l, CFeatures* r,
```
karlnapf: DotFeatures?

src/shogun/kernel/GaussianARDKernel.h
```diff
@@ -85,6 +85,12 @@ class CGaussianARDKernel: public CLinearARDKernel
 	virtual SGMatrix<float64_t> get_parameter_gradient(const TParameter* param,
 			index_t index=-1);
 
+
+	/** set the kernel's width
```
karlnapf: Capital S, but thats a minor

src/shogun/kernel/GaussianARDKernel.h
```diff
@@ -96,6 +102,18 @@ class CGaussianARDKernel: public CLinearARDKernel
 	 */
 	virtual float64_t compute(int32_t idx_a, int32_t idx_b);
 
+	/** compute the distance between features a and b
```
karlnapf: Maybe re-use some existing code rather than adding this here?

src/shogun/kernel/GaussianKernel.cpp
```diff
@@ -31,7 +30,7 @@ CGaussianKernel::CGaussianKernel(int32_t size, float64_t w) : CDotKernel(size)
 	set_width(w);
 }
 
-CGaussianKernel::CGaussianKernel(CDotFeatures* l, CDotFeatures* r,
+CGaussianKernel::CGaussianKernel(CFeatures* l, CFeatures* r,
```
karlnapf: Dot?

src/shogun/kernel/GaussianKernel.cpp
```diff
@@ -108,24 +107,23 @@ float64_t CGaussianKernel::compute(int32_t idx_a, int32_t idx_b)
...
+	float64_t result=distance(idx_a,idx_b);
+	float64_t result_multiplier=1-(sqrt(result))/3;
 
 	if (result_multiplier<=0)
 		result_multiplier=0;
 	else
 		result_multiplier=pow(result_multiplier, power);
 
-	return result_multiplier*exp(-result/width);
+	return result_multiplier*exp(-result);
```
karlnapf: Please add a unit test for the Gaussina kernel against some matlab or python script with a length scale different to 1, just to make sure results dont change

src/shogun/kernel/GaussianKernel.cpp
```diff
@@ -181,3 +179,9 @@ void CGaussianKernel::init()
 	SG_ADD(&width, "width", "Kernel width", MS_AVAILABLE, GRADIENT_AVAILABLE);
 	SG_ADD(&m_compact, "compact", "Compact enabled option", MS_AVAILABLE);
 }
+
+float64_t CGaussianKernel::distance(int32_t idx_a, int32_t idx_b)
+{
+	REQUIRE(sq_lhs && sq_rhs, "Kernel not initialized!\n")
```
karlnapf: Proper error message please

src/shogun/kernel/Kernel.h
```diff
@@ -860,6 +862,62 @@ class CKernel : public CSGObject
...
+		 * @return value of a scalar auxiliary variable based on left feature and right feature
+		 * This auxiliary variable is used in FITC inference for Gaussian processes
+		 */
+		virtual float64_t get_auxiliary_value(int32_t idx_l, int32_t idx_r);
+
+		/** whether the kernel supports FITC inference
+		 *
+		 * @return false
+		 */
+		virtual bool supports_FITC() {return false;}
```
karlnapf: I dont like the fact that the kernel base class contains GP related stuff.
Can we change this?

karlnapf:
Very good patch. Also right size :)

There are some minor things which are easy to fix.

One major change would be to put the GP related parts in CKernel to another place, as we cannot convolute such a fundamental base class with algorithm specific details, thats bad design.
Any ideas how to do this alternatively?

yorkerlin:
@karlnapf 
I did write an alternate version for CFITCKernel by wrapping the existing kernel.
Only one modification n CKernel is changing 
`inline float64_t kernel(int32_t idx_a, int32_t idx_b)`
to
`virtual  float64_t kernel(int32_t idx_a, int32_t idx_b)`
since 
`template <class T> void* CKernel::get_kernel_matrix_helper(void* p)` is used the `float64_t kernel(int32_t idx_a, int32_t idx_b)` function. 

However, I think these wrapped FITC kernels may only work for GP framework since I only tested them in the GP modular.  I will update this version tonight.

src/shogun/kernel/GaussianKernel.cpp
yorkerlin: The `width` is used in `float64_t CGaussianKernel::distance(int32_t idx_a, int32_t idx_b)`.
Locally, I have unit tests for FITCInferenceMethod, which the width is not 1.
I will add some tests in `tests/unit/kernel/Kernel_unittest.cc`

src/shogun/kernel/GaussianKernel.cpp
yorkerlin: All issues about CDotFeatures are
CGaussianKernel is a subclass of CDotKernel() and 
in the following function
```
       virtual bool CDotKernel::init(CFeatures* l, CFeatures* r)
		{
			CKernel::init(l,r);

			ASSERT(l->has_property(FP_DOT))
			ASSERT(r->has_property(FP_DOT))
			ASSERT(l->get_feature_type() == r->get_feature_type())
			ASSERT(l->get_feature_class() == r->get_feature_class())

			if ( ((CDotFeatures*) l)->get_dim_feature_space() != ((CDotFeatures*) r)->get_dim_feature_space() )
			{
				SG_ERROR("train or test features #dimension mismatch (l:%d vs. r:%d)\n",
						((CDotFeatures*) l)->get_dim_feature_space(),((CDotFeatures*) r)->get_dim_feature_space());
			}
			return true;
		}
```

Maybe explicitly asking users to pass CDotFeatures are better.
What is more, maybe we should change the `CDotFeatures()` too
`CDotKernel::CDotKernel(CFeatures* l, CFeatures* r)`

src/shogun/kernel/GaussianARDKernel.cpp
lambday: @karlnapf yep. this would be useful. I think Redux module is the right place for this. Also, if using Eigen3's SIMD powered operations this would be lot faster.

karlnapf:
I see,
Can't you do this via one of the other methods? compute?
@lambday @lisitsyn you are the c++ gods and can probably tell me whether making this low level method virtual will bite us

@yorkerlin go ahead for now, we will see then. It might be ok if the FITC kernels only work for GP, but then put them into the GP folder, wrapping the existing kernel is then ok

karlnapf:
One more thing for your bugfix for the ARDKernel: unit tests.
Same goes for your new kernels in fact, very important if you wrap other objects as behaviour might implicitly change and that can slip through if no unit tests are there.

yorkerlin:
I can try to not modify CKernel by limiting the usage of FITC inference.
In this case, FITC inference only support FITC kernel since I will use
static_cast to convert CKernel to CFITCKernel. In fact, if we want to
compute the gradients wrt inducing points, kernels must support FITC
inference.  Maybe, automatic differentiation would be a solution to
non-FITC kernels to compute the gradients wrt inducing points. (I think
this option is better)

On the other hand,  If inducing points are provided by users and users do
not want to optimize these inducing points, normal kernels could be
directly used in FITC inference with the price of being not able to compute
the gradients wrt these inducing points.



On Thursday, January 29, 2015, Heiko Strathmann <notifications@github.com>
wrote:

> One more thing for your bugfix for the ARDKernel: unit tests.
> Same goes for your new kernels in fact, very important if you wrap other
> objects as behaviour might implicitly change and that can slip through if
> no unit tests are there.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/shogun-toolbox/shogun/pull/2676#issuecomment-71992711>
> .
>


-- 
best,

wu lin

yorkerlin:
It turns out that I need to copy those templated methods in CKernel which
call non-virtual methods and paste in CFITCKernel if CKernel remains
unchanged. Bad design :(

I am back to the original solution by changing the non-virtual method in
templated methods to the virtual version.

On Thursday, January 29, 2015, jaster yorker <yorkerjaster@gmail.com> wrote:

> I can try to not modify CKernel by limiting the usage of FITC inference.
> In this case, FITC inference only support FITC kernel since I will use
> static_cast to convert CKernel to CFITCKernel. In fact, if we want to
> compute the gradients wrt inducing points, kernels must support FITC
> inference.  Maybe, automatic differentiation would be a solution to
> non-FITC kernels to compute the gradients wrt inducing points. (I think
> this option is better)
>
> On the other hand,  If inducing points are provided by users and users do
> not want to optimize these inducing points, normal kernels could be
> directly used in FITC inference with the price of being not able to compute
> the gradients wrt these inducing points.
>
>
>
> On Thursday, January 29, 2015, Heiko Strathmann <notifications@github.com
> <javascript:_e(%7B%7D,'cvml','notifications@github.com');>> wrote:
>
>> One more thing for your bugfix for the ARDKernel: unit tests.
>> Same goes for your new kernels in fact, very important if you wrap other
>> objects as behaviour might implicitly change and that can slip through if
>> no unit tests are there.
>>
>> —
>> Reply to this email directly or view it on GitHub
>> <https://github.com/shogun-toolbox/shogun/pull/2676#issuecomment-71992711>
>> .
>>
>
>
> --
> best,
>
> wu lin
>
>

-- 
best,

wu lin

karlnapf:
Ok do this for now, but we need to confirm with someone who knows about this stuff.

karlnapf:
Asked Sergey, it is fine to make the call virtual, just make sure you check the indices

lisitsyn:
I wouldn't expect any non-marginal performance impact yeah

lisitsyn:
The good thing about having the kernel method is that it is a pattern: http://en.wikibooks.org/wiki/More_C++_Idioms/Non-Virtual_Interface

yorkerlin:
@karlnapf 
Take a look at this bug
https://github.com/shogun-toolbox/shogun/issues/2687
Currently, my class design requires to use the mapping trick. (`SGVector<float64_t> tmp(a.matrix, a.num_rows*a.num_cols, false);`)
However, the bug occurs when the trick is used.
If the bug is not easy to fix, I will rewrite an API.

src/shogun/kernel/GaussianARDKernel.cpp
yorkerlin: `ARDKernel` is a weighted GaussianKernel while the existing `GaussianKernel`  does not have the weights

karlnapf: Computing pairwise distances is the same though.

In fact, what about the following:
Rather than having different classes for ARD kernels, why not add an enum to the existing class.
KT_SCALAR, KT_DIAG, KT_FULL
for scalar (same as std), diagonal (your ARD kernel), and full pairwise (a symmetric matrix) weights?
Then all these seperate classes would just go. Let me know what you think?

yorkerlin: Yes. It could be done. 
The trick part is gradient wrt to parameters (eg, weights). ( in fact, in the future, automatic differentiation may be a good choice for gradient stuff when hand-crafted gradients are not available)    
Usually, the full pairwise matrix should be positive definitive.   (Not sure whether the linear algebra back end supports such checking since Eigen cannot be used in core API)
I am working on it. ( will soon be available next week). ( I am looking at codes about the `linalg`)

karlnapf: Why would that be a problem? The gradient method than just checks the enum and does compute different gradients?
Automatic would be good I agree, but thats for the future for now ;)

Anyway, I think this refactoring makes a lot of sense, check out GPy, they also do this, though they just distinguish two cases: normal and ARD. We should do the same via a flag. I think in fact we can use a boolean, too. The pairwise scaling thing is infeasible anyways, so no need to worry about it, sorry that I brought that up.

http://gpy.readthedocs.org/en/latest/tuto_GP_regression.html

yorkerlin: I am working on unit tests of the FITC Lapalce method.
Will soon update the kernel stuff.

On Sunday, February 1, 2015, Heiko Strathmann <notifications@github.com>
wrote:

> In src/shogun/kernel/GaussianARDKernel.cpp
> <https://github.com/shogun-toolbox/shogun/pull/2676#discussion_r23900579>:
>
> > @@ -122,21 +107,10 @@ SGMatrix<float64_t> CGaussianARDKernel::get_parameter_gradient(
> >  		{
> >  			for (index_t k=0; k<num_rhs; k++)
> >  			{
> > -				SGVector<float64_t> avec=
> > -					((CDenseFeatures<float64_t>*) lhs)->get_feature_vector(j);
> > -				SGVector<float64_t> bvec=
> > -					((CDenseFeatures<float64_t>*) rhs)->get_feature_vector(k);
> > -
> > -				REQUIRE(avec.vlen==bvec.vlen, "Number of right and left hand "
> > -						"features must be the same\n");
> > -
> > -				float64_t result=0;
> > +				float64_t result=distance(j,k);
>
> Why would that be a problem? The gradient method than just checks the enum
> and does compute different gradients?
> Automatic would be good I agree, but thats for the future for now ;)
>
> Anyway, I think this refactoring makes a lot of sense, check out GPy, they
> also do this, though they just distinguish two cases: normal and ARD. We
> should do the same via a flag. I think in fact we can use a boolean, too.
> The pairwise scaling thing is infeasible anyways, so no need to worry about
> it, sorry that I brought that up.
>
> http://gpy.readthedocs.org/en/latest/tuto_GP_regression.html
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/shogun-toolbox/shogun/pull/2676/files#r23900579>.
>


-- 
best,

wu lin

yorkerlin:
@karlnapf 
take a look at this.

yorkerlin:
@karlnapf 
`LinearARD` and `GaussianARD` support three version of ARD kernels, which you mentioned before.
I use the Shogun's `linear algebra lib` in these kernels.

yorkerlin:
@karlnapf 
Some Travis's tasks failed due to missing the `linalg` lib. Do you know how to fix this issue.
One error is
 ```/home/travis/build/shogun-toolbox/shogun/src/shogun/kernel/LinearARDKernel.cpp:86:4: error: ‘linalg’ has not been declared```

karlnapf:
Sorry for the delay, now checking

src/shogun/kernel/LinearARDKernel.cpp
```diff
@@ -47,99 +48,186 @@ CLinearARDKernel::~CLinearARDKernel()
...
+	else 
+	{
+		right=SGMatrix<float64_t> (m_weights.num_rows,1);
+
+		SGMatrix<float64_t> rtmp(right_vec.vector,right_vec.vlen,1,false);
 
-	SG_DEBUG("Initialized weights for LinearARDKernel (%p).\n", this)
+		if(m_ARD_type==KT_DIAG)
+		{
+			linalg::elementwise_product(m_weights, rtmp, right);
```
karlnapf: nice!

src/shogun/kernel/LinearARDKernel.h
```diff
@@ -77,30 +120,34 @@ class CLinearARDKernel: public CDotKernel
...
-	 * @param i index of feature
-	 */
-	virtual void set_weight(float64_t w, index_t i);
+	 * @return	weights in matrix form */
+	virtual SGMatrix<float64_t> get_weights();
 
-	/** get weight of particular feature/dimension
-	 *
-	 * @param i index of feature
+	/** setter for feature/dimension weights in vector form
```
karlnapf: Mmmmh, we should definitely have helper methods that you can give the intuitive representation that then wrap the call to this method. Could you add that in some next patch?

src/shogun/kernel/LinearARDKernel.h
```diff
@@ -4,24 +4,67 @@
...
+ * and \f$p\f$ is # of features
+ * To use this case,  please call set_weights(\f$\lambda\f$),
+ * where \f$\lambda\f$ is a \f$p\f$-by-1 matrix
+ *
+ * The last case is
+ * \f$({\bf x}-{\bf y})^T \Lambda^T \Lambda ({\bf x}-{\bf y})\f$
+ * where \f$\Lambda\f$ is a \f$d\f$-by-\f$p\f$ matrix,
+ * \f$p\f$ is # of features and \f$d\f$ can be \f$ d \ge p\f$ or \f$ d \le p\f$ 
+ * To use this case,  please call set_weights(\f$\Lambda\f$),
+ * where \f$\Lambda\f$ is a \f$d\f$-by-\f$p\f$ matrix
```
karlnapf: as I said below, please have helper methods for this to not confuse users by how to set the weights.
Should be: Scalar, vector, matrix -- as in c++ type

src/shogun/kernel/LinearARDKernel.cpp
```diff
@@ -150,3 +238,30 @@ SGMatrix<float64_t> CLinearARDKernel::get_parameter_gradient(
...
+
+SGMatrix<float64_t> CLinearARDKernel::get_weights()
+{
+	return SGMatrix<float64_t>(m_weights);
+}
+
+void CLinearARDKernel::set_weights(SGMatrix<float64_t> weights)
+{
+	REQUIRE(weights.num_cols>0 && weights.num_rows>0,
+		"Weight Matrix must not be empty\n");
```
karlnapf: Please print the number of columns and rows in the message

src/shogun/kernel/LinearARDKernel.cpp
```diff
@@ -47,99 +48,186 @@ CLinearARDKernel::~CLinearARDKernel()
...
+	}
+	else
+	{
+		SGMatrix<float64_t> left(avec.vector,1,avec.vlen,false);
+		SGMatrix<float64_t> right(bvec.vector,bvec.vlen,1,false);
+		SGMatrix<float64_t> res(1,1);
+
+		if (m_ARD_type==KT_SCALAR)
+		{
+			linalg::matrix_product(left, right, res);
```
karlnapf: @lambday say is matrix product parallelised with openmp?

src/shogun/kernel/LinearARDKernel.cpp
```diff
@@ -47,99 +48,186 @@ CLinearARDKernel::~CLinearARDKernel()
...
 {
-	if (i >= m_weights.vlen)
+	SGMatrix<float64_t> left;
+
+	float64_t scalar_weight=1.0;
+	if (m_ARD_type==KT_SCALAR)
 	{
-		SG_ERROR("Index %i out of range for LinearARDKernel."\
-				 "Number of features is %i.\n", i, m_weights.vlen);
+		left=SGMatrix<float64_t>(avec.vector,1,avec.vlen,false);
```
karlnapf: code formatting is not 100% correct here

src/shogun/kernel/Kernel.h
```diff
@@ -62,6 +62,7 @@ enum EKernelType
 	K_GAUSSIAN = 30,
 	K_GAUSSIANSHIFT = 32,
 	K_GAUSSIANMATCH = 33,
+	K_GAUSSIANFITC = 34,
```
karlnapf: Make sure to document this properly. Having all these different Gaussian kernels might get confusing otherwise.

I would still prefer a *single* GaussianKernel which can enable ARD via a flag and which also containts all the stuff needed for FITC inference. Though it might be good so seperate due to seperation of concerns rule. So let's leave it this way.

src/shogun/kernel/Kernel.cpp
```diff
@@ -1421,8 +1423,10 @@ SGMatrix<T> CKernel::get_kernel_matrix()
 	return SGMatrix<T>(result,m,n,true);
 }
 
+
 template SGMatrix<float64_t> CKernel::get_kernel_matrix<float64_t>();
```
karlnapf: I would really like to get rid of this and instead do a openmp version at some point.
@lambday another one for you :)
@lisitsyn or you? :)

src/shogun/kernel/GaussianShiftKernel.cpp
```diff
@@ -44,12 +41,11 @@ CGaussianShiftKernel::~CGaussianShiftKernel()
...
 	int32_t alen, blen;
-	bool afree, bfree;
 
-	float64_t* avec=
-		((CDenseFeatures<float64_t>*) lhs)->get_feature_vector(idx_a, alen, afree);
-	float64_t* bvec=
-		((CDenseFeatures<float64_t>*) rhs)->get_feature_vector(idx_b, blen, bfree);
+	SGVector<float64_t> avec=((CDotFeatures *)lhs)->get_computed_dot_feature_vector(idx_a);
+	SGVector<float64_t> bvec=((CDotFeatures *)rhs)->get_computed_dot_feature_vector(idx_b);
+	alen=avec.vlen;
```
karlnapf: these variabvles are not really needed.
The assert is also not really nice style.
I know both not your fault, but could you maybe quickly fix them in some later patch?

karlnapf: I added an entrance task on this, see
https://github.com/shogun-toolbox/shogun/issues/2702

src/shogun/kernel/GaussianKernel.h
```diff
@@ -155,6 +155,19 @@ class CGaussianKernel: public CDotKernel
...
+		/** compute the distance between features a and b
+		 * idx_{a,b} denote the index of the feature vectors
+		 * in the corresponding feature object
+		 *
+		 * @param idx_a index a
+		 * @param idx_b index b
+		 * @return computed the distance
+		 *
+		 * Note that in GaussianKernel,
+		 * kernel(idx_a, idx_b)=exp(-distance(idx_a, idx_b))
```
karlnapf: math mode latex?

src/shogun/kernel/GaussianKernel.cpp
karlnapf: good!

src/shogun/kernel/GaussianARDKernel.h
```diff
@@ -4,25 +4,60 @@
...
+ * k({\bf x},{\bf y})= exp(-\frac{||{\bf x}-{\bf y}||}{\tau})
+ * \f]
+ *
+ * where \f$\tau\f$ is the kernel width.
+ *
+ * There are three variants based on \f$||\cdot||\f$. 
+ * The default case is
+ * \f$\sum_{i=1}^{p}{{[\lambda \times ({\bf x_i}-{\bf y_i})] }^2}\f$
+ * where \f$\lambda\f$ is a scalar and \f$p\f$ is # of features
+ * To use this case,  please call set_weights(\f$\lambda\f$),
```
karlnapf: please also add convenience methods here and mention them.
Hide the actual method, make it protected

src/shogun/kernel/GaussianARDKernel.cpp
```diff
@@ -90,28 +75,19 @@ SGMatrix<float64_t> CGaussianARDKernel::get_parameter_gradient(
...
-						"features must be the same\n");
-
-				float64_t element=compute(j,k);
-				float64_t product=CMath::pow((avec[index]-bvec[index]), 2)
-						*(m_weights[index]/m_width);
-
-				derivative(j,k)=-2*element*product;
+				SGVector<float64_t> bvec=((CDotFeatures *)rhs)->get_computed_dot_feature_vector(k);
+				bvec.scale(-1.0);
+				bvec.add(avec);
```
karlnapf: we want to remove such methods from SGVector, please do not use them
@lisitsyn @lambday @iglesias this includes the scale and add, right?

karlnapf:
Ok I think the code is good to merge.

Great work, once again, Wu!
There are all these minor issues, which do not need to be addressed in this patch, but maybe in a next one? Or a small clean-up patch?

karlnapf:
@lambday linalg on travis?

karlnapf:
@yorkerlin once travis is sorted we can merge

karlnapf:
@yorkerlin I actually dont see a linalg include in the LinearARDKernel.cpp

src/shogun/kernel/GaussianARDKernel.cpp
yorkerlin: I need a vector operation to support subtraction. 
I only find `add` in the existing `linear algebra` examples.

src/shogun/kernel/Kernel.h
yorkerlin: @karlnapf 
Let me mention another reason why not use `CGaussianKernel` as the base class.
The reason is for the `scalar` version of Gaussian ARD kernel, we can use the `dot` operation in `CDotFeatures`, which is implemented in the exiting `CGaussianKernel`.
The `dot` operation can support `sparse` dot operation.

However, for the `vector` and `matrix` versions, I have to first get `Dense` vectors and then use them to compute kernel values.

yorkerlin:
@karlnapf 
I included the `linalg` in `LinearARDKernel.h`.
I will remove the header file in `LinearARDKernel.h` and add it in `LinearARDKernel.cpp` to see whether travis complains or not.

src/shogun/kernel/GaussianARDKernel.cpp
karlnapf: add a negative version? mmmmH, we should add this. Can you ping Rahul
@lambday ?

yorkerlin:
@karlnapf  and @lambday 
Can you help me?
I included `linalg` in `LinearARDKernel.cpp` but still get the following error in travis.
Local building is fine in my laptop.
```
/home/travis/build/shogun-toolbox/shogun/src/shogun/kernel/LinearARDKernel.cpp: In member function ‘virtual shogun::SGMatrix<double> shogun::CLinearARDKernel::compute_right_product(shogun::SGVector<double>, float64_t&)’:

/home/travis/build/shogun-toolbox/shogun/src/shogun/kernel/LinearARDKernel.cpp:87:4: error: ‘linalg’ has not been declared

/home/travis/build/shogun-toolbox/shogun/src/shogun/kernel/LinearARDKernel.cpp:91:4: error: ‘linalg’ has not been declared

/home/travis/build/shogun-toolbox/shogun/src/shogun/kernel/LinearARDKernel.cpp: In member function ‘virtual float64_t shogun::CLinearARDKernel::compute_helper(shogun::SGVector<double>, shogun::SGVector<double>)’:

/home/travis/build/shogun-toolbox/shogun/src/shogun/kernel/LinearARDKernel.cpp:120:4: error: ‘linalg’ has not been declared

/home/travis/build/shogun-toolbox/shogun/src/shogun/kernel/LinearARDKernel.cpp:124:4: error: ‘linalg’ has not been declared

/home/travis/build/shogun-toolbox/shogun/src/shogun/kernel/LinearARDKernel.cpp:134:2: error: ‘linalg’ has not been declared

/home/travis/build/shogun-toolbox/shogun/src/shogun/kernel/LinearARDKernel.cpp: In member function ‘virtual float64_t shogun::CLinearARDKernel::compute_gradient_helper(shogun::SGVector<double>, shogun::SGVector<double>, float64_t, index_t)’:

/home/travis/build/shogun-toolbox/shogun/src/shogun/kernel/LinearARDKernel.cpp:165:4: error: ‘linalg’ has not been declared

/home/travis/build/shogun-toolbox/shogun/src/shogun/kernel/LinearARDKernel.cpp:177:4: error: ‘linalg’ has not been declared

/home/travis/build/shogun-toolbox/shogun/src/shogun/kernel/LinearARDKernel.cpp:181:4: error: ‘linalg’ has not been declared

make[2]: *** [src/shogun/CMakeFiles/libshogun.dir/kernel/LinearARDKernel.cpp.o] Error 1

make[2]: *** Waiting for unfinished jobs....

make[1]: *** [src/shogun/CMakeFiles/libshogun.dir/all] Error 2

make: *** [all] Error 2
```

karlnapf:
this is weird
@lambday any ideas?
@vigsterkr  ?

karlnapf:
@yorkerlin Once we have figured this out, this is ready to merge. So feel free to open another PR that is based on this one?

lambday:
@karlnapf @yorkerlin `linalg` works only with c++11 (due to usage of `enum class` in linalg.h). You should guard your code with `HAVE_CXX11` and `HAVE_CXX0X` in order to avoid this problem. I haven't checked in detail but I think it should solve the issue.

@karlnapf @vigsterkr I think its time we use a c+11 supported compiler in travis (if that is in our hand, not sure how things work there).

src/shogun/kernel/GaussianARDKernel.cpp
lambday: @yorkerlin @karlnapf totally possible. As a matter of fact, it should just be one method with one [optional] functor param :D

src/shogun/kernel/GaussianARDKernel.h
```diff
@@ -4,25 +4,60 @@
...
+ * Written (W) 2015 Wu Lin
  * Written (W) 2012 Jacob Walker
  *
  * Adapted from WeightedDegreeRBFKernel.h
  */
 
-#ifndef GAUSSIANARDKERNEL_H_
-#define GAUSSIANARDKERNEL_H_
+#ifndef _GAUSSIANARDKERNEL_H_
+#define _GAUSSIANARDKERNEL_H_
```
lambday: @karlnapf umm I remember you suggesting something regarding the usage of underscores in the beginning of these. @yorkerlin is there any particular reason you changed this?

src/shogun/kernel/LinearARDKernel.cpp
```diff
@@ -47,99 +49,186 @@ CLinearARDKernel::~CLinearARDKernel()
...
 
-float64_t CLinearARDKernel::get_weight(index_t i)
-{
-	if (i >= m_weights.vlen)
-	{
-		SG_ERROR("Index %i out of range for LinearARDKernel."\
-				 "Number of features is %i.\n", i, m_weights.vlen);
+		SGMatrix<float64_t> left_transpose(left.matrix,left.num_cols,1,false);
+		if(m_ARD_type==KT_DIAG)
+		{
```
lambday: sorry but minor comment: if it is just one line, we don't use braces for `if` `for` etc things :)

src/shogun/kernel/GaussianARDKernel.cpp
karlnapf: @lambday I put up a trap to get you a GSoC applicant for this: #2717

karlnapf:
@yorkerlin ok with the guard it would work then. 

@lambday is it really a good idea to need to have these guards everywhere? I dont like guards.....

@vigsterkr Also it would be cool to be able to test these linalg things in travis, so what about having c++11?

yorkerlin:
@lambday 
I have one question about the `linalg` lib.
If the dimension is matched, does the lib support `in place` operations.
for example,
```
linalg::elementwise_product(a, b, a); // or linalg::matrix_product(b,a,a);
linalg::matrix_product(a, b,a ); // or linalg::matrix_product(b,a,a);
```
If the lib does not support `in place` operations
Having a element-wise `pow` and matrix `pow` may be good. 
```
linalg::elementwise_pow(a,2);  // linalg::elementwise_product(a,a,c);
linalg::matrix_pow(a,2);  // linalg::matrix_product(a,a,c);
```

lambday:
@karlnapf I don't like guards either. I opened an issue #2719.
@yorkerlin that's a really cool idea! I don't think presently it supports that. I may be wrong but I think [this](https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/mathematics/linalg/internal/implementation/ElementwiseProduct.h#L101) returns a new matrix. If you know how to do this with ViennaCL please add/change/modify. One question is - how to we determine that the matrices are **huge** and we need to rely on block-based processing?

lambday:
+1 for the pow operation.

yorkerlin:
@karlnapf  and @lambday 
It seems that `src/shogun/base/class_list.cpp` will automatically  create `CLinearARDKernel` and `CGaussianARDKernel` objects.
However, when `HAVE_CXX11` is not defined, these two ARD kernels are empty, which leads to a error in travis. 
```
/home/travis/build/shogun-toolbox/shogun/src/shogun/base/class_list.cpp: In function ‘shogun::CSGObject* __new_CLinearARDKernel(shogun::EPrimitiveType)’:

/home/travis/build/shogun-toolbox/shogun/src/shogun/base/class_list.cpp:808:113: error: expected type-specifier before ‘CLinearARDKernel’

/home/travis/build/shogun-toolbox/shogun/src/shogun/base/class_list.cpp:808:113: error: expected ‘:’ before ‘CLinearARDKernel’

/home/travis/build/shogun-toolbox/shogun/src/shogun/base/class_list.cpp:808:130: error: ‘CLinearARDKernel’ was not declared in this scope

/home/travis/build/shogun-toolbox/shogun/src/shogun/base/class_list.cpp:808:131: error: expected ‘;’ before ‘:’ token

/home/travis/build/shogun-toolbox/shogun/src/shogun/base/class_list.cpp:808:131: error: expected primary-expression before ‘:’ token

/home/travis/build/shogun-toolbox/shogun/src/shogun/base/class_list.cpp:808:131: error: expected ‘;’ before ‘:’ token
```
I know there is a trick to `create empty` `CLinearARDKernel’ and `CGaussianARDKernel` objects in `class_list.cpp`  to address this error.

However, do you know how to exclude  these two kernels in  `class_list.cpp`  when `HAVE_CXX11` is not defined? In order words,  `class_list.cpp` will `NOT create`  ‘CLinearARDKernel’ and `CGaussianARDKernel` objects

yorkerlin:
@lambday 
Since Shogun focuses on single desktop machine, In my opinion,  whether a matrix is huge or not should depend on CPU memory and GPU memory. 
We may view CPU memory as `a hard disk drive` and GPU memory as `a traditional memory`. We may be able to use some `IO-efficient blocking tricks` to solve some numerical linear algebra. 
For example, `I/O-Efficient Techniques for Computing Pagerank` may be a good example.

lambday:
@yorkerlin can you provide some links for the blocking tricks you're talking about - some talk/paper/anything? I've got hold of that pagerank paper. will it be useful for our purpose? we can include these ideas in linalg project ideas this year. cool stuff :+1:

lambday:
Regarding the class_list.cpp thing, I think its better not to touch class_list.cpp.py. We can of course blacklist these classes but its better if we can create empty instances with default constructor as well.

yorkerlin:
@lambday 
for the pagerank problem:

`Amy N. Langville and Carl D. Meyer.  A Reordering for the PageRank problem. SIAM Journal on Scientific Computing. ` exploit the sparsity of matrices  

`Amy N. Langville and Carl D. Meyer.  Deeper Inside PageRank. Internet Mathematics, Vol. 1(3):335-380, 2005.` theory behind pagerank

`Chen, Yen-Yu, Qingqing Gan, and Torsten Suel. "I/O-efficient techniques for computing PageRank." CIKM 2002.` one blocking trick

yorkerlin:
@lambday  or @karlnapf 
please restart the failed traivs task.
It seems the error is unrelated to my PR.

yorkerlin:
@karlnapf 
BTW, the original `LinearARDKernel` has a bug about computing the `gradient`.
Since I completely rewrite these ARD kernels, the bug is fixed. 

The FITC inference has a unifying framework to compute `gradients of hyper-parameters`, which may be useful to integrate `automatic differentiation`.

To-Do list for me in the GP module  (improving the existing framework)

* a unifying framework to compute `gradients of hyper-parameters` for non-FITC inference 
(Laplace, Exact, EP and all KL methods)
* model-based selection of GP `hyper-parameters`
* optimization framework in GP (non hyper-parameters) (about Newton method, Quasi-Newton, and other optimizers)

src/shogun/kernel/GaussianARDKernel.cpp
iglesias: @yorkerlin, I think you should be able to do it using [scale](https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/mathematics/linalg/internal/modules/Core.h#L85) and [add](https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/mathematics/linalg/internal/modules/Core.h#L69). Please try.

yorkerlin:
@iglesias 
I used the `linalg::add` for the `scale_and_add`. Once travis is green, this PR should be ready to be merged.

yorkerlin:
@iglesias
Please restart the failed Travis' tasks.

src/shogun/kernel/GaussianKernel.h
yorkerlin: Added the math

src/shogun/kernel/GaussianARDKernel.h
yorkerlin: No. Let me know if this is an issue.

src/shogun/kernel/LinearARDKernel.cpp
yorkerlin: It is the if... else if... else statement. Let me know if this is an issue

karlnapf:
@yorkerlin thanks for fixing the bug. Please put this in the NEWS file (and also mention that you also fixed in in GPML ;) )
You mentioned that the framework requires understanding of GPs. This means you should document it very carefully. Really important.

Ill restart travis

karlnapf:
The travis error is due to memory limit. I'll merge the patch!

karlnapf:
@yorkerlin please update the NEWS with all you additions, its good to expose it to people.

A few examples would also be nice once everything else is merged. Looking forward

src/shogun/kernel/GaussianARDKernel.h
karlnapf: There is some standard that say #defines should not start with underscores. 

http://stackoverflow.com/questions/6756059/c-using-c-code-using-double-underscores-in-defines-and-identifiers

not a big thing but maybe good to clean up in the next PR?

src/shogun/kernel/LinearARDKernel.cpp
karlnapf: we dont use { } if the statement is only one line. But thats minor here

i.e.
```
else if(m_ARD_type==KT_FULL)
    linalg::matrix_product(m_weights, ltmp, left_transpose);
```
rather than
```
else if(m_ARD_type==KT_FULL)
		{
			linalg::matrix_product(m_weights, ltmp, left_transpose);
		}
```

src/shogun/kernel/GaussianKernel.h
karlnapf: thanks!

src/shogun/kernel/LinearARDKernel.cpp
yorkerlin: ok
I will change it in the next PR.
BTW, one exception of this rule should be `macros` (e.g. REQUIRE) right?

lambday:
@yorkerlin thanks for the notes. maybe we can discuss regarding this sometime. I should open an issue on this so that we can move our in-place related discussions there.

src/shogun/kernel/LinearARDKernel.cpp
karlnapf: yeah true, otherwise the compiler might complain

Issue #2564, opened 2014-10-25
Reduce complexity of headers with d-pointers and "only declaration" rule


lisitsyn:
I propose to change:

``` cpp

class CClass
{
public:
  void method();
  int get_field() const
  {
    return field;
  }
private:
  int m_field;
};

// .cpp

void CClass::method()
{
   .. m_field ..
}
```

to a better looking (and faster compiling) variant:

``` cpp

class CClass 
{
public:
    void method();
    int get_field() const;
private:
    class Self;
    Unique<Self> self;
};

// .cpp

class CClass::Self
{
  int m_field;
};

CClass::get_field()
{
  return self->m_field;
}

CClass::method()
{
   .. get_field(); ..
}

```

It would involve massive refactoring so should be done once and fast. One thing that comes for free - in our code we will not use member variables w/o getters anymore (which is considered good). ~~Another advantage is that very likely SWIG will compile much faster.~~

I think possible approach involves writing some script instead of doing it manually.

iglesias:
@lisitsyn, why is it good not to use member variables w/o getters (and setters too, I guess) within the class where they are members?

lisitsyn:
@iglesias I'd say it is more flexible

iglesias:
I am not sure what flexible is here. Can you put an example, @lisitsyn? :-)

karlnapf:
why does this compile faster?

lisitsyn:
Okay guys a few cases that come into my mind:

1. If we go with generic get/set without specific getters for field adding a new field is free - you just recompile the .cpp file (@iglesias).
2. In the same scenario headers are binary compatible with various versions of shogun (@iglesias).
3. We can remove *all* mentions of std vector/list/.. from headers because they are only in .cpp files so it will be faster (@karlnapf).
4. We can probably use simpler saving/loading that just works with that subobject (declared in .cpp) but not the object (declared in .h) - that's a thing I am thinking about yet.

karlnapf:
+1 from my side to all this

iglesias:
It sounds pretty awesome!

iglesias:
Thanks a lot for the nice description, @lisitsyn :-)

vigsterkr:
obsolete

PR #2677, opened 2015-01-22
Added Unique handler to store implementations

lisitsyn:
This thing was described in #2564

### DIFF
--- a/src/shogun/base/unique.h
+++ b/src/shogun/base/unique.h
@@ -1,0 +1,44 @@
+#ifndef __SG_UNIQUE_H__
+#define __SG_UNIQUE_H__
+
+#ifdef HAVE_CXX11
+
+#include <memory>
+
+namespace shogun
+{
+
+	/** Holds unique pointer that is deleted once this holder is deleted. 
+	 * Its main usage is to hold a pointer to implementation (pimpl idiom):
+	 *
+	 * class Self;
+	 * Unique<Self> self;
+	 *
+	 */
+	template <typename T>
+		class Unique : protected std::unique_ptr<T>
+	{
+		public:
+			Unique() : std::unique_ptr<T>(new T)
+		{
+		}
+
+			Unique(const Unique& other) = delete;
+			Unique(Unique&& other) = delete;
+			Unique& operator=(const Unique& other) = delete;
+
+			~Unique()
+			{
+			}
+
+			inline T* operator->() const
+			{
+				return std::unique_ptr<T>::get();
+			}
+
+	};
+
+}
+
+#endif /* HAVE_CXX11 */
+#endif /* __SG_UNIQUE_H__ */
--- a/src/shogun/base/unique.h
+++ b/src/shogun/base/unique.h
@@ -25,7 +25,6 @@ namespace shogun
 			}
 
 			Unique(const Unique& other);
-			Unique(Unique&& other);
 			Unique& operator=(const Unique& other);
 
 			inline T* operator->() const
--- a/src/shogun/base/unique.h
+++ b/src/shogun/base/unique.h
@@ -1,10 +1,6 @@
 #ifndef __SG_UNIQUE_H__
 #define __SG_UNIQUE_H__
 
-#ifdef HAVE_CXX11
-
-#include <memory>
-
 namespace shogun
 {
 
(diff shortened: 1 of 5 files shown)


lisitsyn:
Comments anyone?

src/shogun/base/unique.h
```diff
@@ -0,0 +1,44 @@
...
+	 * Unique<Self> self;
+	 *
+	 */
+	template <typename T>
+		class Unique : protected std::unique_ptr<T>
+	{
+		public:
+			Unique() : std::unique_ptr<T>(new T)
+		{
+		}
```
iglesias: I think there's something weird with the indentation here.

iglesias:
It is looking cool to me!

src/shogun/base/unique.h
lisitsyn: Yeah will fix

src/shogun/base/unique.h
```diff
@@ -0,0 +1,44 @@
+#ifndef __SG_UNIQUE_H__
+#define __SG_UNIQUE_H__
+
+#ifdef HAVE_CXX11
+
+#include <memory>
+
+namespace shogun
```
iglesias: We should also avoid SWIG from going through this file, right? I am not sure if that is already happening since there is no mention to this in any .i file or we should also use the guard.

tests/unit/base/Unique_unittest.cc
```diff
@@ -0,0 +1,33 @@
+#include <shogun/base/some.h>
+#include <shogun/kernel/GaussianKernel.h>
```
iglesias: Why this one?

iglesias:
This looks like a pretty neat solution to me. From my side, feel free to merge ;-)

tests/unit/base/Unique_unittest.cc
lisitsyn: Will remove

src/shogun/base/unique.h
lisitsyn: Why?

iglesias: Because of the `std::unique_ptr` use in the header. Also, this is to be used internally, not something to expose, right?

src/shogun/base/unique.h
```diff
@@ -0,0 +1,44 @@
...
+
+	/** Holds unique pointer that is deleted once this holder is deleted. 
+	 * Its main usage is to hold a pointer to implementation (pimpl idiom):
+	 *
+	 * class Self;
+	 * Unique<Self> self;
+	 *
+	 */
+	template <typename T>
+		class Unique : protected std::unique_ptr<T>
```
abinashmeher999: Finally, got to see a `protected` mode of derivation :relieved: . Curious to know the reason! :)

lisitsyn: To restrict all operations but calling a member function of underlying object with ->. Unique pointer has some things like reset(), release(), etc and they can be easily mixed up with reset() of underlying object. E.g. these are different things but easy to confuse: 

```
std::unique_ptr<T> obj;
obj.reset();
obj->reset();
```

src/shogun/base/unique.h
lisitsyn: Hmm probably I should do some magic here to push all the std stuff to .cpp.

We can't ignore it in SWIG because it would be needed through concrete classes anyway

lisitsyn:
@iglesias take a look

src/shogun/transfer/multitask/MultitaskL12LogisticRegression.cpp
```diff
@@ -13,12 +13,27 @@
 #include <shogun/lib/SGVector.h>
 #include <shogun/features/DotFeatures.h>
 
+#ifdef HAVE_CXX11
```
iglesias: We should be able to take out this one now?

src/shogun/transfer/multitask/MultitaskL12LogisticRegression.h
```diff
@@ -4,13 +4,16 @@
...
  *
- * Copyright (C) 2012 Sergey Lisitsyn
+ * Copyright (C) 2014 Sergey Lisitsyn
  */
 
 #ifndef  MULTITASKL12LOGISTICREGRESSION_H_
 #define  MULTITASKL12LOGISTICREGRESSION_H_
 
 #include <shogun/lib/config.h>
+#ifdef HAVE_CXX11
```
iglesias: Here the same?

tests/unit/base/Unique_unittest.cc
iglesias: Still needed? :-)

lisitsyn: Yeahh will remove ;)

src/shogun/transfer/multitask/MultitaskL12LogisticRegression.cpp
lisitsyn: Yes

iglesias:
@lisitsyn, do you happen to know if there is any performance difference wrt to doing it with unique_ptr? I was just wondering whether this should wait until we decide if we take C++11 as compulsory dependency.

lisitsyn:
@iglesias no difference or faster. No need to use unique_ptr here anyway.

iglesias:
Perfect. Then, from my part, feel free to merge once you think it's ready
;-)

PR #2679, opened 2015-01-23
 Update revision of doc/md submodule

sanuj:
@iglesias I have edited the wiki but I don't think it's possible to open a PR for a wiki. Here is the [diff](https://github.com/sanuj/shogun/wiki/_compare/0a8ddcbf20d6268403dce1bc9af45d7eee15df2f%5E...0a8ddcbf20d6268403dce1bc9af45d7eee15df2f). I have pushed to changes to my fork. Here is the [link](https://github.com/sanuj/shogun/wiki/README_developer). We can close this PR if this is not the appropriate way.


iglesias:
I have updated now the README_devleoper in the wiki with your suggested changes, Sanuj. Thanks!

PR #2254, opened 2014-05-23
added support for configuring neural networks using JSON files

khalednasr:

PR #2680, opened 2015-01-24
added support for configuring neural networks using JSON files

iglesias:
Rebase of #2254.


iglesias:
We let Travis finish and then we merge it (everything should be fine anyway since #2254 was green).

Issue #2380, opened 2014-07-11
Problem with cygwin installation for python modular ? 


aveshd:
Hi,

Please accept my apologies if this issue has already been solved. But I need help, I am not a big time user of cygwin, but I try to find my way around. 

Before coming to this forum I have tried various options to the best of my ability. 
I am trying to follow the instructions for installation of shogun for cygwin. 

I am am getting this error and everything stops compiling further. 
# 
# 

[ 27%] Building CXX object src/shogun/CMakeFiles/libshogun.dir/io/File.cpp.o
/home/shogun-develop/src/shogun/io/File.cpp: In constructor ‘shogun::CFile::CFile(int, const char_, const char_)’:
/home/shogun-develop/src/shogun/io/File.cpp:44:22: error: ‘fdopen’ was not declared in this scope
  file=fdopen(fd, mode);
                      ^
/home/shogun-develop/src/shogun/io/File.cpp: In member function ‘void shogun::CFile::set_variable_name(const char_)’:
/home/shogun-develop/src/shogun/io/File.cpp:185:27: error: ‘strdup’ was not declared in this scope
  variable_name=strdup(name);
                           ^
/home/shogun-develop/src/shogun/io/File.cpp: In member function ‘char_ shogun::CFile::get_variable_name()’:
/home/shogun-develop/src/shogun/io/File.cpp:190:29: error: ‘strdup’ was not declared in this scope
  return strdup(variable_name);
                             ^
/home/shogun-develop/src/shogun/io/File.cpp:191:1: warning: control reaches end of non-void function [-Wreturn-type]
 }
 ^
src/shogun/CMakeFiles/libshogun.dir/build.make:5368: recipe for target 'src/shogun/CMakeFiles/libshogun.dir/io/File.cpp.o' failed
make[2]: **\* [src/shogun/CMakeFiles/libshogun.dir/io/File.cpp.o] Error 1
CMakeFiles/Makefile2:232: recipe for target 'src/shogun/CMakeFiles/libshogun.dir/all' failed
make[1]: **\* [src/shogun/CMakeFiles/libshogun.dir/all] Error 2
Makefile:136: recipe for target 'all' failed
# make: **\* [all] Error 2
# 

it seems the problem is with fdopen function. 

Here is my cmake output. Just in case if you need to look at that as well
# 
# 

cmake -DPythonModular=ON ..
-- The C compiler identification is GNU 4.8.3
-- The CXX compiler identification is GNU 4.8.3
[...]

PR #2682, opened 2015-01-24
Use get_strdup instead of strdup to avoid cygwin crashes

lisitsyn:
Fixes one of the reasons for the issue #2380


iglesias:
Travis failed jobs are not due to changes here.

Issue #2673, opened 2015-01-18
Document CMultidimensionalScaling


karlnapf:
and other dimensionality reduction methods from Tapkee....

This means: 
 * Link to Wikipedia page
 * Basic math
 * Objective function
 * Computational costs

All this should go to the doxygen class documentation

iglesias:
Part of the content at (tapkee.lisitsyn.me)[http://tapkee.lisitsyn.me/], under Dimension reduction techniques (upper left corner) can be probably recycled for this.

curiousguy13:
Hi, I'm new here and I would like to work on this. 
Also the pull requests have to be for the changes in one file at a time , right?

iglesias:
Nope. You can submit a pr with changes in several files. It is good though to keep prs small and focussing on one task.

curiousguy13:
Oh, okay. So, for this I need to change the documentation in the source files right? for example, for Isomap, I would need to add the changes in src/shogun/converter/Isomap.h ?

iglesias:
Exactly.

PR #2683, opened 2015-01-29
added documentation for Isomap doxygen class

curiousguy13:
I added some documentation to the Isomap doxygen class as required by issue #2673 .
I mostly recycled it from here: http://tapkee.lisitsyn.me/ .
I also added the wikipedia link as required in the issue.
Is there anything else I can add in this documentation that would be helpful?

### DIFF
--- a/src/shogun/converter/Isomap.h
+++ b/src/shogun/converter/Isomap.h
@@ -36,11 +36,13 @@ class CDistance;
  * classic Multidimensional Scaling algorithm. The algorithm itself
  * consists of the following steps:
  * - For each feature vector \f$x\in X\f$ find its \f$k\f$ nearest
- * neighbours and construct the sparse neighbourhood graph.
+ *   neighbours and construct the sparse neighbourhood graph.
  * - Compute squared distances matrix \f$D\f$ such as \f$D\{i,j\} = d^2(x_i,x_j)\f$.
  * - Relax distances with shortest(so-called geodesic) distances on the sparse neighbourhood graph (e.g. with sparse Dijkstra algorithm).
  * - Center the matrix \f$D\f$ with subtracting row mean, column mean and adding to the grand mean. Multiply \f$D\f$ element-wise with \f$-0.5\f$.
- * - Compute embedding with the \f$t\f$ eigenvectors that correspond to the largest eigenvalues of the matrix \f$D\f$; normalize these vectors dividing each eigenvector by the square root of its corresponding eigenvalue. Form the final embedding with eigenvectors as rows and projected feature vectors as columns.
+ * - Compute embedding with the \f$t\f$ eigenvectors that correspond to the largest eigenvalues of the matrix \f$D\f$; normalize these vectors
+ *   dividing each eigenvector by the square root of its corresponding eigenvalue. Form the final embedding with eigenvectors as rows and projected
+ *   feature vectors as columns.
 
  * It is possible to apply preprocessor to specified distance using
  * apply_to_distance.
--- a/src/shogun/converter/Isomap.h
+++ b/src/shogun/converter/Isomap.h
@@ -37,7 +37,7 @@ class CDistance;
  * consists of the following steps:
  * - For each feature vector \f$x\in X\f$ find its \f$k\f$ nearest
  *   neighbours and construct the sparse neighbourhood graph.
- * - Compute squared distances matrix \f$D\f$ such as \f$D\{i,j\} = d^2(x_i,x_j)\f$.
+ * - Compute the squared distances matrix \f$D\f$ such as \f$D\{i,j\} = d^2(x_i,x_j)\f$.
  * - Relax distances with shortest(so-called geodesic) distances on the sparse neighbourhood graph (e.g. with sparse Dijkstra algorithm).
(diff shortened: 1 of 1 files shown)


src/shogun/converter/Isomap.h
```diff
@@ -24,14 +24,24 @@ namespace shogun
...
+ * as described in:
  *
  * Silva, V. D., & Tenenbaum, J. B. (2003).
  * Global versus local methods in nonlinear dimensionality reduction.
  * Advances in Neural Information Processing Systems 15, 15(Figure 2), 721-728. MIT Press.
  * Retrieved from http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.9.3407&rep=rep1&type=pdf
  *
+ * The Isomap algorithm can be considered as a modification of the
+ * classic Multidimensional Scaling algorithm. The algorithm itself is
+ * consists following steps:
```
iglesias: The algorithm itself consists of

src/shogun/converter/Isomap.h
```diff
@@ -24,14 +24,24 @@ namespace shogun
...
+ * - For each feature vector \f$x\in X\f$ find its \f$k\f$ nearest
+ * neighbours and construct the sparse neighbourhood graph.
+ * - Compute squared distances matrix \f$D\f$ such as \f$D\{i,j\} = d^2(x_i,x_j)\f$.
```
iglesias: Compute the squared distances matrix

src/shogun/converter/Isomap.h
```diff
@@ -24,14 +24,24 @@ namespace shogun
...
+ * - Relax distances with shortest(so-called geodesic) distances on the sparse neighbourhood graph (e.g. with sparse Dijkstra algorithm).
+ * - Center the matrix \f$D\f$ with subtracting row mean, column mean and adding to the grand mean. Multiply \f$D\f$ element-wise with \f$-0.5\f$.
+ * - Compute embedding with the \f$t\f$ eigenvectors that correspond to the largest eigenvalues of the matrix \f$D\f$; normalize these vectors with dividing each eigenvectors with square root of its corresponding eigenvalue. Form the final embedding with eigenvectors as rows and projected feature vectors as columns.
```
iglesias: Break the three lines above to have a similar line width as the one used in the previous paragraphs (we don't follow any specific line width, but it is nice to be consistent at least within files).

iglesias: Normalise these vector dividing (remove with)

iglesias: Each eigenvector (singular)

iglesias: with square root should be by the square root of its corresponding eigenvalue

src/shogun/converter/Isomap.h
```diff
@@ -39,6 +49,9 @@ class CDistance;
  *
  * To use this converter with static interfaces please refer it by
  * sg('create_converter','isomap',k);
+
+ * cf. http://tapkee.lisitsyn.me/
+ * cf. https://en.wikipedia.org/wiki/Isomap
```
iglesias: A reference to the paper where Isomap was first introduced is a must as well :-)

http://isomap.stanford.edu/

iglesias: Mm I just saw there is already a reference for another paper at the beginning of the class description. I think the one in the link above is more suitable since it is an older paper but I am fine leaving it as it is as well.

src/shogun/converter/Isomap.h
```diff
@@ -24,14 +24,24 @@ namespace shogun
...
  * Advances in Neural Information Processing Systems 15, 15(Figure 2), 721-728. MIT Press.
  * Retrieved from http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.9.3407&rep=rep1&type=pdf
  *
+ * The Isomap algorithm can be considered as a modification of the
+ * classic Multidimensional Scaling algorithm. The algorithm itself is
+ * consists following steps:
+ * - For each feature vector \f$x\in X\f$ find its \f$k\f$ nearest
+ * neighbours and construct the sparse neighbourhood graph.
+ * - Compute squared distances matrix \f$D\f$ such as \f$D\{i,j\} = d^2(x_i,x_j)\f$.
+ * - Relax distances with shortest(so-called geodesic) distances on the sparse neighbourhood graph (e.g. with sparse Dijkstra algorithm).
```
iglesias: Is it "sparse" Dijkstra's algorithm a special version of Dijkstra's algorithm or just the same? If the latter, I would just say Dijkstra's algorithm,

iglesias: shortest (so-called) missing white space after shortest.

lisitsyn:
Haha that's very neat that you review my language here @iglesias haha

lisitsyn:
Actually I'd love to get these fixes back to the website as well

curiousguy13:
@lisitsyn don't worry, once the issues here are fixed , i'll send a PR for your website :-)

curiousguy13:
@iglesias thanks for your input. 
I made the changes you asked here, could you take a look if I missed something?
I looked at the implementation of isomap in tapkee and it seems like Dijkstra's algorithm only and not sparse sparse Dijkstra's , so changed it to Dijkstra's ( @lisitsyn correct me if i'm wrong here).

iglesias:
Thanks, @curiousguy13!

PR #2684, opened 2015-01-29
Optional fdopen calls

lisitsyn:
POSIX C and ANSI C differs in their support of the fdopen function.

Added cmake check and ifdef for these fdopen calls

Relates #2380

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -447,9 +447,7 @@ CHECK_CXX_SYMBOL_EXISTS(signgam "cmath" HAVE_DECL_SIGNGAM)
 AppendToDefines(HAVE_DECL_SIGNGAM)
 
 CHECK_CXX_SYMBOL_EXISTS(fdopen "stdio.h" HAVE_FDOPEN)
-IF(HAVE_FDOPEN)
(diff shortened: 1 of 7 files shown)


CMakeLists.txt
```diff
@@ -446,6 +446,11 @@ AppendToDefines(HAVE_DECL_ISNAN)
 CHECK_CXX_SYMBOL_EXISTS(signgam "cmath" HAVE_DECL_SIGNGAM)
 AppendToDefines(HAVE_DECL_SIGNGAM)
 
+CHECK_CXX_SYMBOL_EXISTS(fdopen "stdio.h" HAVE_FDOPEN)
+IF(HAVE_FDOPEN)
+    LIST(APPEND DEFINES HAVE_FDOPEN)
+ENDIF()
```
iglesias: What about `AppendToDefines(HAVE_FDOPEN)` instead of this if? It is more succint :)

lisitsyn: True

lisitsyn:
Ok to merge?

iglesias:
It's cool with me :)

PR #2688, opened 2015-01-30
fixes for CMake

besser82:
* failsafe fix for new CMake-policy CMP0053 introduced by CMake >= 3.1
* fix "Argument not separated from preceding token by whitespace." for R-modular
* fix doxy2swig to use "encoding='utf-8'" when opening files by default  
on e.g. Fedora there are encoding-errors when using Python 3.X

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -286,7 +286,7 @@ else()
 endif()
 
 ############### VERSION #####################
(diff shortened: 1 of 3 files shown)


besser82:
do not merge, yet…  there is more stuff to come…

PR #2689, opened 2015-02-01
Add native implementation for addition of vectors and matrices in linalg

sanuj:
@lambday @iglesias Please review.

### DIFF
--- a/src/shogun/mathematics/linalg/internal/implementation/Add.h
+++ b/src/shogun/mathematics/linalg/internal/implementation/Add.h
@@ -55,20 +55,89 @@ namespace linalg
 namespace implementation
 {
 
-/** Generic class which is specialized for different backends to perform addition */
+/**
+ * @brief Generic class which is specialized for different backends to perform addition
+ */
 template <enum Backend, class Matrix>
 struct add
 {
 	/** Scalar type */
 	typedef typename Matrix::Scalar T;
 	
-	/** Performs the operation C = alpha*A + beta*B. Works for both matrices and vectors */
+	/**
+	 * Performs the operation C = alpha*A + beta*B. Works for both matrices and vectors
+	 * @param A first matrix
+	 * @param B second matrix
+	 * @param C matrix to store the result
+	 * @param alpha constant to be multiplied by the first matrix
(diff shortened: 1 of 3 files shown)


lambday:
looks good to me. Please add your name in the header as well :)

lambday:
also, I think we should provide default values for alpha and beta (=1). can you please add that?

src/shogun/mathematics/linalg/internal/implementation/Add.h
```diff
@@ -55,20 +55,89 @@ namespace linalg
...
+	 * @param B second vector
+	 * @param C vector to store the result
+	 * @param alpha constant to be multiplied by the first vector
+	 * @param beta constant to be multiplied by the second vector
+	 * @param len length of the vectors/matrices
+	 */
+	static void compute(T* A, T* B, T* C,
+		T alpha, T beta, index_t len)
+	{
+		for (int32_t i=0; i<len; i++)
```
lambday: this can be done in parallel, openmp-ized :) but its fine for now.

sanuj: Do you mean by adding something like `#pragma omp parallel for ...`. I can do that in this PR itself.

sanuj: Also what all operations would we generally like to do in parallel? Do we always use openmp or something else as well? Just so that I take care of it in future :)

iglesias:
lgtm too.

src/shogun/mathematics/linalg/internal/implementation/Add.h
lambday: @sanuj yes that's what I meant. There's no such rule of thumb for what operations we like to be executed in parallel - but it's worth noting down the scopes where we can. Also, I'm not sure whether we should use pthread or openmp. So maybe leave that for now - we can anyway add it later :)

lambday:
Travis failure unrelated. Merging this one.

lambday:
@sanuj great job! looking forward to your next PR :)

PR #2691, opened 2015-02-03
Add openmp and default values in function add(linalg)

sanuj:
@lambday I added openmp. Not sure about it, can be incorrect.

### DIFF
--- a/src/shogun/mathematics/linalg/internal/implementation/Add.h
+++ b/src/shogun/mathematics/linalg/internal/implementation/Add.h
@@ -1,6 +1,7 @@
 /*
  * Copyright (c) The Shogun Machine Learning Toolbox
  * Written (w) 2014 Khaled Nasr
+ * Written (w) 2015 Sanuj Sharma
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
@@ -92,7 +93,7 @@ struct add<Backend::NATIVE, Matrix>
 	 * @param beta constant to be multiplied by the second matrix
 	 */
 	static void compute(SGMatrix<T> A, SGMatrix<T> B, SGMatrix<T> C,
-		T alpha, T beta)
+		T alpha=1, T beta=1)
 	{
 		REQUIRE(A.num_rows == B.num_rows && B.num_rows == C.num_rows,
 			"Matrices should have same number of rows!\n");
@@ -110,7 +111,7 @@ struct add<Backend::NATIVE, Matrix>
 	 * @param beta constant to be multiplied by the second vector
 	 */
 	static void compute(SGVector<T> A, SGVector<T> B, SGVector<T> C,
-		T alpha, T beta)
+		T alpha=1, T beta=1)
 	{
 		REQUIRE(A.vlen == B.vlen, "Vectors should have same length!\n");
 		compute(A.vector, B.vector, C.vector, alpha, beta, A.vlen);
--- a/tests/unit/mathematics/linalg/Add_unittest.cc
+++ b/tests/unit/mathematics/linalg/Add_unittest.cc
@@ -55,6 +55,7 @@ TEST(AddMatrix, native_backend)
 	SGMatrix<float64_t> A(3,3);
 	SGMatrix<float64_t> B(3,3);
 	SGMatrix<float64_t> C(3,3);
+	SGMatrix<float64_t> D(3,3);
 	
 	for (int32_t i=0; i<9; i++)
 	{
@@ -63,9 +64,13 @@ TEST(AddMatrix, native_backend)
 	}
 	
 	linalg::add<linalg::Backend::NATIVE>(A, B, C, alpha, beta);
+	linalg::add<linalg::Backend::NATIVE>(A, B, D);
 	
 	for (int32_t i=0; i<9; i++)
+	{
 		EXPECT_NEAR(alpha*A[i]+beta*B[i], C[i], 1e-15);
+		EXPECT_NEAR(A[i]+B[i], D[i], 1e-15);
+	}
 }
 
 TEST(AddVector, native_backend)
@@ -76,6 +81,7 @@ TEST(AddVector, native_backend)
 	SGVector<float64_t> A(9);
 	SGVector<float64_t> B(9);
 	SGVector<float64_t> C(9);
+	SGVector<float64_t> D(9);
 	
 	for (int32_t i=0; i<9; i++)
 	{
@@ -84,9 +90,13 @@ TEST(AddVector, native_backend)
 	}
 	
 	linalg::add<linalg::Backend::NATIVE>(A, B, C, alpha, beta);
+	linalg::add<linalg::Backend::NATIVE>(A, B, D);
 	
 	for (int32_t i=0; i<9; i++)
+	{
 		EXPECT_NEAR(alpha*A[i]+beta*B[i], C[i], 1e-15);
+		EXPECT_NEAR(A[i]+B[i], D[i], 1e-15);
+	}
 }
 
 #ifdef HAVE_LINALG_LIB


tests/unit/mathematics/linalg/Add_unittest.cc
```diff
@@ -63,9 +64,12 @@ TEST(AddMatrix, native_backend)
 	}
 	
 	linalg::add<linalg::Backend::NATIVE>(A, B, C, alpha, beta);
+	linalg::add<linalg::Backend::NATIVE>(A, B, D);
 	
-	for (int32_t i=0; i<9; i++)
+	for (int32_t i=0; i<9; i++) {
```
iglesias: Just for the sake of following our usual style, open and close braces in new lines. Here and below.

sanuj: @iglesias Done :)

iglesias:
@lambday or @lisitsyn, can you check if the openmp part (is just one line :) makes sense?

lambday:
@iglesias yeah I don't think it needs anything else. @lisitsyn ?
@sanuj can you please run valgrind on the test and paste the report here? you can use option `--gtest_filter=AddVector.*` for the executable `shogun-unit-tests` inside `build/tests/unit` directory while running valgrind, something like
`valgrind --leak-check=full --track-origins=yes /path/to/shogun-unit-tests --gtest_filter=AddVector.*`

sanuj:
@lambday Sure. I'll have a look at it and tell you by tonight. Have classes right now ;)

sanuj:
@lambday This is what i got after running valgrind
```
sanuj@sanuj-VAIO:~/git-reps/shogun$ valgrind build/tests/unit/shogun-unit-test --gtest_filter=AddVector.* --leak-check=full --track-origins=yes
==8410== Memcheck, a memory error detector
==8410== Copyright (C) 2002-2013, and GNU GPL'd, by Julian Seward et al.
==8410== Using Valgrind-3.10.0.SVN and LibVEX; rerun with -h for copyright info
==8410== Command: build/tests/unit/shogun-unit-test --gtest_filter=AddVector.* --leak-check=full --track-origins=yes
==8410== 
Note: Google Test filter = AddVector.*
[==========] Running 3 tests from 1 test case.
[----------] 3 tests from AddVector
[ RUN      ] AddVector.native_backend
[       OK ] AddVector.native_backend (23 ms)
[ RUN      ] AddVector.eigen3_backend
[       OK ] AddVector.eigen3_backend (17 ms)
[ RUN      ] AddVector.viennacl_backend
unknown file: Failure
C++ exception with description "ViennaCL: FATAL ERROR: ViennaCL encountered an unknown OpenCL error. In some cases, this might be due to an invalid global work size, but it can also be due to several compilation errors.
If you think that this is a bug in ViennaCL, please report it at viennacl-support@lists.sourceforge.net and supply at least the following information:
 * Operating System
 * Which OpenCL implementation (AMD, NVIDIA, etc.)
 * ViennaCL version
Many thanks in advance!" thrown in the test body.
[  FAILED  ] AddVector.viennacl_backend (856 ms)
[----------] 3 tests from AddVector (909 ms total)

[==========] 3 tests from 1 test case ran. (980 ms total)
[  PASSED  ] 2 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] AddVector.viennacl_backend

 1 FAILED TEST
==8410== 
==8410== HEAP SUMMARY:
==8410==     in use at exit: 14,171 bytes in 15 blocks
==8410==   total heap usage: 58,292 allocs, 58,277 frees, 3,867,887 bytes allocated
==8410== 
==8410== LEAK SUMMARY:
==8410==    definitely lost: 12,432 bytes in 5 blocks
==8410==    indirectly lost: 94 bytes in 4 blocks
==8410==      possibly lost: 0 bytes in 0 blocks
==8410==    still reachable: 1,645 bytes in 6 blocks
==8410==         suppressed: 0 bytes in 0 blocks
==8410== Rerun with --leak-check=full to see details of leaked memory
==8410== 
==8410== For counts of detected and suppressed errors, rerun with: -v
==8410== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 0 from 0)
```

lisitsyn:
Please benchmark that before merging. Simple add would benefit from vectorization more than from parallelization.

lisitsyn:
To answer the question whether openmp is needed here we need to consider what sizes we expect here. Unless we add really big vectors it would hurt to make it parallel.

lambday:
@lisitsyn it would be so cool if we write vectorized dot ourselves :D 
@sanuj http://bruun.co/2012/02/07/easy-cpp-benchmarking

sanuj:
@lisitsyn I read the benchmarking blog. Was very helpful. I greped for "BENCHMARK" in shogun and found nothing so i guess it hasn't been done anywhere.
@lisitsyn will openmp simd work for vectorization? Something like `#pragma openmp simd` before the for loop.

lisitsyn:
@lambday compilers are smart enough to vectorize this loop

@sanuj not sure but probably not, I'd rather let one core pass it with big strides than let it call all these omp routines

sanuj:
@lisitsyn so there is no need of any `#pragma omp` directives. The compiler
will vectorize on it's own?

lisitsyn:
@sanuj yes whether it is possible

src/shogun/mathematics/linalg/internal/implementation/Add.h
```diff
@@ -128,6 +129,7 @@ struct add<Backend::NATIVE, Matrix>
 	static void compute(T* A, T* B, T* C,
 		T alpha, T beta, index_t len)
 	{
+#pragma omp parallel for shared(A, B, C, alpha, beta)
```
sanuj: @lisitsyn @lambday shall i remove this then? And is there still a need of benchmarking if we remove this?

lambday: yeah I believe that makes sense. I tried locally with/without the omp directive for dot and the assembly it generated was same - compiler takes care of the vectorization.

lambday:
@lisitsyn the valgrind error for viennacl is most probably due to installation issue. what do you think?

lambday:
@sanuj you can just write a small gist comparing Eigen3's dot and our linalg dot using native backend - probably on randomly generated vectors of size ~10000, ~1000 times. It would be interesting to know whether Eigen3 is doing something different there with SIMD. you don't need to send a PR with that. Just give a link of the gist here, with the output (runtime comparison).

src/shogun/mathematics/linalg/internal/implementation/Add.h
sanuj: @lambday removed it.

sanuj:
@lambday I tried comparing the dot functions by calling them in a `int main()` function but I'm getting errors on compiling the code with `g++ -lshogun`. I think these are due to my local installing issues (I installed the libeigen3-dev package for ubuntu)
```
In file included from /usr/local/include/shogun/mathematics/linalg/internal/implementation/ElementwiseSquare.h:40:0,
                 from /usr/local/include/shogun/mathematics/linalg/internal/modules/Core.h:35,
                 from /usr/local/include/shogun/mathematics/linalg/linalg.h:181,
                 from rough1.cpp:2:
/usr/local/include/shogun/mathematics/eigen3.h:17:24: fatal error: Eigen/Eigen: No such file or directory
  #include <Eigen/Eigen>
```
Am I supposed to do it like this or is there any other way?

lambday:
@sanuj try compiling with `g++ -O3 -std=c++11 -I/path/to/eigen3/in/your/system  -lshogun` (first one is capital i, second one is small L).
in my system it is `/usr/include/eigen3`

sanuj:
@lambday [Here](https://gist.github.com/sanuj/1099abec85e90b11eb0d) is the link to my gist (used hayai). BTW, why don't we use [google benchmarking](https://github.com/google/benchmark)?
`g++ -std=c++11 -I/usr/include/eigen3 dot_comparison.cpp -lshogun -lOpenCL` worked for me. Thanks for your suggestion :) Native dot seems to be running faster :P (Might have done something wrong here.)

lisitsyn:
Add ` -O3 `

sanuj:
@lisitsyn Yeah that worked. Thanks :) Updated the [gist](https://gist.github.com/sanuj/1099abec85e90b11eb0d).

sanuj:
@lambday @lisitsyn Do we need to add something before merge?

lambday:
@sanuj nah it is good to go IMO :)

lambday:
Travis error unrelated. Merging.

src/shogun/mathematics/linalg/internal/implementation/Add.h
```diff
@@ -110,7 +111,7 @@ struct add<Backend::NATIVE, Matrix>
 	 * @param beta constant to be multiplied by the second vector
 	 */
 	static void compute(SGVector<T> A, SGVector<T> B, SGVector<T> C,
-		T alpha, T beta)
+		T alpha=1, T beta=1)
 	{
 		REQUIRE(A.vlen == B.vlen, "Vectors should have same length!\n");
```
karlnapf: This require statement is bad, see

src/shogun/mathematics/linalg/internal/implementation/Add.h
```diff
@@ -92,7 +93,7 @@ struct add<Backend::NATIVE, Matrix>
 	 * @param beta constant to be multiplied by the second matrix
 	 */
 	static void compute(SGMatrix<T> A, SGMatrix<T> B, SGMatrix<T> C,
-		T alpha, T beta)
+		T alpha=1, T beta=1)
 	{
 		REQUIRE(A.num_rows == B.num_rows && B.num_rows == C.num_rows,
```
karlnapf: same here

src/shogun/mathematics/linalg/internal/implementation/Add.h
karlnapf: I know its off topic for the patch, but still. We should keep those clean, to avoid user frustration

PR #2692, opened 2015-02-04
Add linalg max native implementation

sanuj:
@lambday @iglesias Max function for vectors is also available in CMath. We can remove it from CMath if we are adding it here after the C++11 issue is resolved.

### DIFF
--- a/src/shogun/mathematics/linalg/internal/implementation/Max.h
+++ b/src/shogun/mathematics/linalg/internal/implementation/Max.h
@@ -34,6 +34,7 @@
 #include <shogun/lib/config.h>
 #include <shogun/lib/SGMatrix.h>
 #include <shogun/lib/SGVector.h>
+#include <shogun/mathematics/Math.h>
 
 #ifdef HAVE_EIGEN3
 #include <shogun/mathematics/eigen3.h>
@@ -56,7 +57,8 @@ namespace linalg
 namespace implementation
 {
 
-/** Generic class which is specialized for different backends to perform 
+/** 
+ * @brief Generic class which is specialized for different backends to perform 
  * the max operation
  */
 template <enum Backend,class Matrix>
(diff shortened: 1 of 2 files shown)


src/shogun/mathematics/linalg/internal/implementation/Max.h
```diff
@@ -34,6 +34,7 @@
 #include <shogun/lib/config.h>
 #include <shogun/lib/SGMatrix.h>
 #include <shogun/lib/SGVector.h>
+#include <shogun/mathematics/Math.h>
```
sanuj: Should i instead `include <math.h>`?

src/shogun/mathematics/linalg/internal/implementation/Max.h
```diff
@@ -65,12 +67,63 @@ struct max
...
+	 * Returns the largest element in a vector or matrix passed as a pointer.
+	 * @param m input vector or matrix
+	 * @param len length of the vector or matrix
+	 * @return largest value in the input vector or matrix
+	 */
+	static T compute(T* vec, index_t len)
+	{
+		ASSERT(len>0)
+		T maxv=vec[0];
+		for (int32_t i=1; i<len; i++)
```
sanuj: @lambday Can we do this in parallel?

src/shogun/mathematics/linalg/internal/implementation/Max.h
lambday: nah shogun's math is fine.

src/shogun/mathematics/linalg/internal/implementation/Max.h
lambday: Nope I think that's unnecessary overhead for this one.

src/shogun/mathematics/linalg/internal/implementation/Max.h
```diff
@@ -65,12 +67,63 @@ struct max
...
+				maxv=std::max(vec[i], maxv);
```
iglesias: As we relying on the stl for doing the max, I suggest we just go and use `std::max_element` directly.

iglesias: That is, that `compute` basically calls std::max_element.

src/shogun/mathematics/linalg/internal/implementation/Max.h
```diff
@@ -65,12 +67,63 @@ struct max
...
+
+	/**
+	 * Returns the largest element in a vector or matrix passed as a pointer.
+	 * @param m input vector or matrix
+	 * @param len length of the vector or matrix
+	 * @return largest value in the input vector or matrix
+	 */
+	static T compute(T* vec, index_t len)
+	{
+		ASSERT(len>0)
```
iglesias: If this can crash from the usage of a user, I think we should use `REQUIRE` with a proper error message.

src/shogun/mathematics/linalg/internal/implementation/Max.h
```diff
@@ -65,12 +67,61 @@ struct max
...
+
+	/**
+	 * Returns the largest element in a vector or matrix passed as a pointer.
+	 * @param m input vector or matrix
+	 * @param len length of the vector or matrix
+	 * @return largest value in the input vector or matrix
+	 */
+	static T compute(T* vec, index_t len)
+	{
+		return *std::max_element(vec, vec+len);
```
sanuj: @iglesias switched to `max_element`

src/shogun/mathematics/linalg/internal/implementation/Max.h
```diff
@@ -65,12 +67,61 @@ struct max
...
+	typedef typename Matrix::Scalar T;
+
+	/**
+	 * Returns the largest element in a matrix.
+	 * @param mat input matrix
+	 * @return largest value in the input matrix
+	 */
+	static T compute(SGMatrix<T> mat)
+	{
+		REQUIRE(mat.num_cols*mat.num_rows > 0, "Matrix can not be empty!\n");
```
sanuj: @iglesias Added the error check here and below.

iglesias:
Perfect, thanks. Travis and merge.

PR #2693, opened 2015-02-04
added documentation for the FactorAnalysis class

curiousguy13:
in reference to: #2673 
added documentation for the FactorAnalysis doxygen class mainly from here: http://tapkee.lisitsyn.me/
is there any other relevent info about Factor Analysis that i can add here?

### DIFF
--- a/src/shogun/converter/FactorAnalysis.h
+++ b/src/shogun/converter/FactorAnalysis.h
@@ -40,6 +40,12 @@ namespace shogun
  *
  * Uses implementation from the Tapkee library.
  *
+ * cf. http://tapkee.lisitsyn.me/ <br>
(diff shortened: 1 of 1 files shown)


iglesias:
Thanks for polishing the documentation. I will merge this but next time, please use only one commit for a change as small as this one. You can squash more than one commit into one after having done them.

curiousguy13:
Alright , thanks for the tip. I'll remember it for next time. :)

karlnapf:
If you want infos on FA, check david barber's book "Bayesian reasoning and machine learning", which is available online.

I would love to see the math for the generative model of FA in the docs. Also a good thing to learn about

curiousguy13:
I'll definitely check it out :)

PR #2694, opened 2015-02-05
fix build of modular interfaces with SWIG 3.0.5 on MacOSX

besser82:
This is needed with SWIG >= 3.0.5 on MacOSX because of conflicting macros in AssertMacros.h of Carbon-framework.  Won't cause harm on other distros.


iglesias:
Hey, @besser82. I see you can fix stuff related to osx, that's great :-) Do you think you could have a look to the broken osx job in Travis? https://travis-ci.org/shogun-toolbox/shogun/jobs/49594806

PR #2698, opened 2015-02-11
fix segfaults when calling delete[] twice on SGMatrix-instances and fix minor issues with swig 3.X

besser82:
fix minor issues when building python_modular with "patched" swig 3.X
 * fix failing test when building with swig 3.X
 * fix segfault when calling delete[] twice on SGMatrix-instances


besser82:
Failures on Travis are related to Octave.  Not related to changes in this PR.

Issue #2702, opened 2015-02-12
Get rid of ASSERT macros


karlnapf:
We have lot's of ```ASSERT``` calls in Shogun. In many cases, these can fail due to user input, which is bad, as they don't provide feedback on which part of the user input is wrong and how.

This task is to scan for those ASSERT messages, and replace them by REQUIRE messages with *proper* error messages. See here:
https://github.com/shogun-toolbox/shogun/wiki/Assertions

somya-anand:
@karlnapf If assertion is for internal variables then is there a reason to convert them into `REQUIRE` messages? I mean for user, these things are a black box and she didn't require a feedback for information which she didn't provided.

Example:
```c
/** ./src/shogun/lib/SGVector.cpp:268 */
template<class T>
SGVector<T> SGVector<T>::operator+ (SGVector<T> x)
{
	ASSERT(x.vector && vector)
	ASSERT(x.vlen == vlen)

	SGVector<T> result=clone();
	result.add(x);
	return result;
}
```

iglesias:
Is `SGVector::operator+` public? I guess it is. Then, if  a user happily tries to add two vectors of different lengths using the method, the assert will say hi.

somya-anand:
@iglesias I have quoted the irrelevant example here, but what I was asking is - if something which is not provided by the user then should we change that assertion as well?

iglesias:
Then, the answer is no. To find out whether an assert can be evaluated to false due to wrong, unexpected, or user input not contemplated in the code might not be straightforward in every case, though.

karlnapf:
Sorry for late reply, but yeah. Everything that can be broken by a user call should be require, otherwise assert (with a comment) is fine. Thanks for taking care

erip:
Recommend closing.

PR #2703, opened 2015-02-13
shogun:lib Replace ASSERT with REQUIRE

somya-anand:
Fix issue #2702 in file src/shogun/lib/SGVector.cpp

### DIFF
--- a/src/shogun/lib/SGVector.cpp
+++ b/src/shogun/lib/SGVector.cpp
@@ -244,14 +244,14 @@ void SGVector<complex128_t>::range_fill_vector(complex128_t* vec,
 template<class T>
 const T& SGVector<T>::get_element(index_t index)
 {
-	ASSERT(vector && (index>=0) && (index<vlen))
+	REQUIRE(vector && (index>=0) && (index<vlen), "Provided index (%d) must be between 0 and %d.\n", index, vlen);
 	return vector[index];
 }
 
 template<class T>
 void SGVector<T>::set_element(const T& p_element, index_t index)
 {
-	ASSERT(vector && (index>=0) && (index<vlen))
+	REQUIRE(vector && (index>=0) && (index<vlen), "Provided index (%d) must be between 0 and %d.\n", index, vlen);
 	vector[index]=p_element;
 }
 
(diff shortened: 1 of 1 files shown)


src/shogun/lib/SGVector.cpp
```diff
@@ -304,7 +303,7 @@ void SGVector<T>::add(const SGSparseVector<T>& x)
 		for (int32_t i=0; i < x.num_feat_entries; i++)
 		{
 			index_t idx = x.features[i].feat_index;
-			ASSERT(idx < vlen)
+			REQUIRE(idx < vlen, "Feature index should be less than %d.\n", vlen);
```
iglesias: Can you display here idx as well? Maybe with `Features index (%d) should`...

src/shogun/lib/SGVector.cpp
```diff
@@ -371,7 +370,7 @@ template <>
 void SGVector<bool>::display_vector(const bool* vector, int32_t n, const char* name,
 		const char* prefix)
 {
-	ASSERT(n>=0)
+	REQUIRE(n>=0, "Vector size can not be negative.\n");
```
iglesias: Here and below maybe it is good to display n as well.

iglesias: Nah, never mind about that one. It should be understandable enough to say that a negative value was passed.

iglesias:
All right, looking good. We wait for Travis and then we merge.

src/shogun/lib/SGVector.cpp
karlnapf: No, please put the idx. Provides help what is going wrong.

src/shogun/lib/SGVector.cpp
```diff
@@ -868,7 +867,7 @@ void SGVector<complex128_t>::load(CFile* loader)
 
 template<class T> void SGVector<T>::save(CFile* saver)
 {
-	ASSERT(saver)
+	REQUIRE(saver, "Requires a valid 'c FILE pointer'\n");
```
karlnapf: "Provided CFile pointer is NULL\n"

src/shogun/lib/SGVector.cpp
```diff
@@ -518,7 +517,7 @@ template <>
 void SGVector<complex128_t>::display_vector(const complex128_t* vector, int32_t n,
 		const char* name, const char* prefix)
 {
-	ASSERT(n>=0)
+	REQUIRE(n>=0, "Vector size can not be negative.\n");
```
karlnapf: see above, please display index, and it is spelled "cannot", but it should really be "must not"

PR #2706, opened 2015-02-14
added GEMPLP

Jiaolong:
@hushell Could you please help me to check this patch?

### DIFF
--- a/src/shogun/structure/FactorGraphDataGenerator.h
+++ b/src/shogun/structure/FactorGraphDataGenerator.h
@@ -1,0 +1,138 @@
+/*
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Written (W) 2015 Jiaolong Xu
+ * Copyright (C) 2015 Jiaolong Xu
+ */
+
+#ifndef __FACTOR_GRAPH_DATA_GENERATOR_H__
+#define __FACTOR_GRAPH_DATA_GENERATOR_H__
+
+#include <shogun/mathematics/Math.h>
+#include <shogun/lib/SGVector.h>
+#include <shogun/lib/SGNDArray.h>
+#include <shogun/lib/SGMatrix.h>
+#include <shogun/base/DynArray.h>
+#include <shogun/base/init.h>
+#include <shogun/io/SGIO.h>
+
+#include <shogun/structure/FactorGraph.h>
+#include <shogun/structure/FactorGraphModel.h>
+#include <shogun/structure/FactorType.h>
+#include <shogun/structure/Factor.h>
+
+#include <shogun/structure/MAPInference.h>
+#include <shogun/structure/StochasticSOSVM.h>
+
+#include <shogun/features/FactorGraphFeatures.h>
+#include <shogun/labels/FactorGraphLabels.h>
+
+namespace shogun
+{
+/** @brief Class CFactorGraphDataGenerator Create factor graph data for  multiple unit tests
+ */
+class CFactorGraphDataGenerator: public CSGObject
+{
+public:
+	/** Constructor */
+	CFactorGraphDataGenerator();
+
+	/** @return class name */
+	virtual const char* get_name() const
+	{
+		return "FactorGraphDataGenerator";
+	}
+
+	/** Define a simple 2 node chain graph */
+	CFactorGraph* simple_chain_graph();
+
+	/** Convert grid coordinate into 1-d index
+	 *
+	 * @param x grid coordinate x
+	 * @param y grid coordinate y
+	 * @param w grid width
+	 * @return index in 1-d vector
+	 */
+	int32_t grid_to_index(int32_t x, int32_t y, int32_t w = 10);
+
+	/** Truncate energy table to ensure submodurality
+	 *
+	 * @param A value in (0,0)
+	 * @param B value in (0,1)
+	 * @param C value in (1,0)
+	 * @param D value in (1,1)
+	 */
+	void truncate_energy(float64_t &A, float64_t &B, float64_t &C, float64_t &D);
+
+	/** Define a four nodes chain graph
+	 * potentials are randomly generated
+	 *
+	 * @param assignment_expect expected assignment
+	 * @param min_energy_expect expected minimum energies
+	 * @param N size of the energy table (e.g., 2x2)
+	 */
+	CFactorGraph* random_chain_graph(SGVector<int> &assignment_expect, float64_t &min_energy_expect, int32_t N = 2);
+
+	/** Define a multiple state tree graph */
+	CFactorGraph* multi_state_tree_graph();
+
+	/** Generate random data following [1]:
+	 * Each example has exactly one label on.
+	 * Each label has 40 related binary features.
+	 * For an example, if label i is on, 4i randomly chosen features are set to 1
+	 *
+	 * [1] Finley, Thomas, and Thorsten Joachims.
+	 * "Training structural SVMs when exact inference is intractable."
+	 * Proceedings of the 25th international conference on Machine learning. ACM, 2008.
+	 *
+	 * @param len_label label length (10)
+	 * @param len_feat feature length (40)
+	 * @param size_data training data size (50)
+	 * @param feats generated feature matrix
+	 * @param labels generated label matrix
+	 */
+	void generate_data(int32_t len_label, int32_t len_feat, int32_t size_data,
+	                   SGMatrix<float64_t> &feats, SGMatrix<int32_t> &labels);
+
+	/** Get fully connected edges
+	 *
+	 * @param num_classes number of classes
+	 * @return matrix of edge
+	 */
+	SGMatrix< int32_t > get_edges_full(const int32_t num_classes);
+
+	/** Build factor graph
+	 *
+	 * @param feats features
+	 * @param labels labels
+	 * @param edge_list edge list
+	 * @param v_factor_type factor types
+	 * @param fg_feats features for factor graph
+	 * @param fg_labels labels for factor graph
+	 */
+	void build_factor_graph(SGMatrix<float64_t> feats, SGMatrix<int32_t> labels,
+	                        SGMatrix< int32_t > edge_list, const DynArray<CTableFactorType*> &v_factor_type,
+	                        CFactorGraphFeatures* fg_feats, CFactorGraphLabels* fg_labels);
+
+	/** Define factor type
+	 *
+	 * @param num_classes number of classes
+	 * @param dim dimension of the feature
+	 * @param num_edges number of edegs
+	 * @param v_factor_type factor types
+	 */
+	void define_factor_types(int32_t num_classes, int32_t dim, int32_t num_edges,
+	                         DynArray<CTableFactorType*> &v_factor_type);
+
+	/** Test sosvm inference algorithm with random data
+	 *
+	 * @param infer_type type of inference algorithm
+	 * @return average training loss (expected to be 0)
+	 */
+	float64_t test_sosvm(EMAPInferType infer_type);
+};
+}
+#endif
--- a/src/shogun/structure/GEMPLP.cpp
+++ b/src/shogun/structure/GEMPLP.cpp
@@ -1,0 +1,360 @@
+/*
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Written (W) 2014 Jiaolong Xu
+ * Copyright (C) 2014 Jiaolong Xu
+ */
+
+#include <shogun/structure/GEMPLP.h>
+#include <shogun/io/SGIO.h>
+#include <algorithm>
+
+using namespace shogun;
+
+CGEMPLP::CGEMPLP()
+	: CMAPInferImpl()
+{
+	m_fg = NULL;
+	m_factors = NULL;
+}
+
+CGEMPLP::CGEMPLP(CFactorGraph* fg, Parameter param)
+	: CMAPInferImpl(fg),
+	  m_param(param)
+{
+	ASSERT(m_fg != NULL);
+
+	init();
+}
+
+CGEMPLP::~CGEMPLP()
+{
+	if(m_factors != NULL)
+		SG_UNREF(m_factors);
+}
+
+void CGEMPLP::init()
+{	
+	SGVector<int32_t> fg_var_sizes = m_fg->get_cardinalities();	
+	m_factors = m_fg->get_factors();
+
+	int32_t num_factors = m_factors->get_num_elements();
+	m_region_intersections.resize(num_factors);
+
+	// get all the intersections
+	for (int32_t i = 0; i < num_factors; i++)
+	{
+		CFactor* factor_i = dynamic_cast<CFactor*>(m_factors->get_element(i));
+		SGVector<int32_t> region_i = factor_i->get_variables();
+		SG_UNREF(factor_i);
+
+		for (int32_t j = i; j < num_factors; j++)
+		{
+			CFactor* factor_j = dynamic_cast<CFactor*>(m_factors->get_element(j));
+			SGVector<int32_t> region_j = factor_j->get_variables();
+			SG_UNREF(factor_j);
+
+			const int32_t k = find_intersection_index(region_i, region_j);
+			if (k < 0) continue;
+
+			m_region_intersections[i].insert(k);
+
+			if (j != i)
+				m_region_intersections[j].insert(k);
+		}
+	}	
+
+	m_region_inds_intersections.resize(num_factors);
+	m_msgs_from_region.resize(num_factors);
+	m_theta_region.resize(num_factors);
+
+	for (int32_t c = 0; c < num_factors; c++)
+	{
+		CFactor* factor_c = dynamic_cast<CFactor*>(m_factors->get_element(c));
+		SGVector<int32_t> vars_c = factor_c->get_variables();
+		SG_UNREF(factor_c);
+		
+		m_region_inds_intersections[c].resize(m_region_intersections[c].size());
+		m_msgs_from_region[c].resize(m_region_intersections[c].size());
+
+		int32_t s = 0;
+
+		for (set<int>::iterator t = m_region_intersections[c].begin(); 
+				t != m_region_intersections[c].end(); t++)
+		{
+			SGVector<int32_t> curr_intersection = m_all_intersections[*t];
+			SGVector<int32_t> inds_s(curr_intersection.size());
+			SGVector<int32_t> dims_array(curr_intersection.size());
+			
+			for (int32_t i = 0; i < inds_s.size(); i++)
+			{
+				inds_s[i] = vars_c.find(curr_intersection[i])[0];
+				REQUIRE(inds_s[i] >= 0, 
+						"Intersection contains variable %d which is not in the region %d", curr_intersection[i], c);
+
+				dims_array[i] = fg_var_sizes[curr_intersection[i]];
+			}
+			
+			// initialize indices of intersections inside the region
+			m_region_inds_intersections[c][s] = inds_s;
+			
+			// initialize messages from region and set it 0
+			SGNDArray<float64_t> message(dims_array);
+			message.set_const(0);
+			m_msgs_from_region[c][s] = message.clone();
+			s++;
+		}
+
+		// initialize potential on region
+		m_theta_region[c] = convert_energy_to_potential(factor_c);
+	}
+
+	// initialize messages in intersections and set it 0
+	m_msgs_into_intersections.resize(m_all_intersections.size());
+	
+	for (uint32_t i = 0; i < m_all_intersections.size(); i++)
+	{
+		SGVector<int32_t> vars_intersection = m_all_intersections[i];
+		SGVector<int32_t> dims_array(vars_intersection.size());
+		
+		for (int32_t j = 0; j < dims_array.size(); j++)
+			dims_array[j] = fg_var_sizes[vars_intersection[j]];
+	
+		SGNDArray<float64_t> curr_array(dims_array);
+		curr_array.set_const(0);
+		m_msgs_into_intersections[i] = curr_array.clone();
+	}	
+}
+
+SGNDArray<float64_t> CGEMPLP::convert_energy_to_potential(CFactor* factor)
+{
+	SGVector<float64_t> energies = factor->get_energies();
+	SGVector<int32_t> cards = factor->get_cardinalities();
+
+	SGNDArray<float64_t> message(cards);
+
+	if (cards.size() == 1)
+	{
+		for (int32_t i = 0; i < energies.size(); i++)
+			message.array[i] = - energies[i];
+	}
+	else if (cards.size() == 2)
+	{
+		for (int32_t y = 0; y < cards[1]; y++)
+			for (int32_t x = 0; x < cards[0]; x++)
+				message.array[x*cards[1]+y] = - energies[y*cards[0]+x];
+	}
+	else
+		SG_ERROR("Index issue has not been solved for higher order (>=3) factors.");
+
+	return message.clone();
+}
+
+int32_t CGEMPLP::find_intersection_index(SGVector<int32_t> region_A, SGVector<int32_t> region_B)
+{
+	vector<int32_t> tmp;
+
+	for (int32_t i = 0; i < region_A.size(); i++)
+	{
+		for (int32_t j = 0; j < region_B.size(); j++)
+		{
+			if (region_A[i] == region_B[j])
+				tmp.push_back(region_A[i]);
+		}
+	}
+	
+	// return -1 if intersetion is empty
+	if (tmp.size() == 0)	return -1;
+
+	
+	SGVector<int32_t> sAB(tmp.size());
+	for (uint32_t i = 0; i < tmp.size(); i++)
+		sAB[i] = tmp[i];
+
+	// find (or add) intersection set
+	int32_t k;
+	for (k = 0; k < (int32_t)m_all_intersections.size(); k++)
+		if (m_all_intersections[k].equals(sAB))
+			break;
+
+	if (k == (int32_t)m_all_intersections.size())
+		m_all_intersections.push_back(sAB);
+	
+	return k;
+}
+
+float64_t CGEMPLP::inference(SGVector<int32_t> assignment)
+{
+	REQUIRE(assignment.size() == m_fg->get_cardinalities().size(),
+	        "%s::inference(): the output assignment should be prepared as"
+	        "the same size as variables!\n", get_name());
+
+	// iterate over message loop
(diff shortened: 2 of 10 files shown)


hushell:
@Jiaolong Good good! This one is much better than what I was thinking about. 
It's nice to get rid of Region class in the original implementation (OI). I think you have fully understand the MPLP algorithm. I can tell from update_messages(), which is almost different from the OI. You are amazing, man! 
The only thing that we should be more careful is the naming rule of variables, functions etc. I'll expand that in the code.

src/shogun/structure/GEMPLP.h
```diff
@@ -0,0 +1,160 @@
...
+		          const float64_t int_gap_thr = 0.0002)
+			: m_max_iter(max_iter),
+			  m_obj_del_thr(obj_del_thr),
+			  m_int_gap_thr(int_gap_thr)
+		{}
+
+		/** maximum number of outer iterations*/
+		int32_t m_max_iter;
+		/** threshold of the delta objective value */
+		float64_t m_obj_del_thr;
```
hushell: May need to explain more about these vars. 
obj_del = last_dual_obj - dual_obj
int_gap = duality gap
Correct me if I am wrong.

src/shogun/structure/GEMPLP.h
```diff
@@ -0,0 +1,160 @@
...
+	 */
+	void max_in_subdimension(SGNDArray<float64_t> tar_arr, SGVector<int32_t> &subset_inds, SGNDArray<float64_t> &max_res) const;
+
+	/** Find separators between clqiues
+	 *
+	 * @param clique_A clique A
+	 * @param clique_B clique B
+	 * @return index in all separators
+	 */
+	int32_t find_separator_index(SGVector<int32_t> clique_A, SGVector<int32_t> clique_B);
```
hushell: I guess this is analogy to sepset for clique trees. However, actually messages are passed between clusters/regions which are not formed in a hypertree but in general a hypergraph. I don't know if seperator set is a conventional in the context or not. Besides, calling it intersection sets is not a bad idea. The more we follow the OI names, the easier for people to understand.

hushell: I suggest to rename to find_intersection_set_index

src/shogun/structure/GEMPLP.h
```diff
@@ -0,0 +1,160 @@
...
+
+	/** Convert original energies to potentials of the cluster
+	 * GEMPLP objective function is a maximazation function, we use - energy
+	 * as potential. The indices of the table factor energy also need to be
+	 * converted to the order of the nd array.
+	 *
+	 * @param factor factor which contains energy
+	 * @return potential of the cluster in MPLP
+	 */
+	SGNDArray<float64_t> convert_message(CFactor* factor);
```
hushell: Suggest to rename to convert_energy_to_potential

src/shogun/structure/GEMPLP.h
```diff
@@ -0,0 +1,160 @@
...
+
+	/** Insert the new value into a vector
+	 *
+	 * @param v the target vector
+	 * @param k current value
+	 */
+	void exist_or_pushback(vector<int32_t>& v, int32_t k);
```
hushell: Is it better to use std:set? since you don't want to have copies. I mean: vector< set<int32_t> > m_clique_separators. Then use set:insert, rather than this function.

src/shogun/structure/GEMPLP.h
```diff
@@ -0,0 +1,160 @@
...
+
+public:
+	/** GEMPLP parameter */
+	Parameter m_param;
+	/** all factors in the graph*/
+	CDynamicObjectArray* m_factors;
+	/** all separators */
+	vector<SGVector<int32_t> > m_all_separators;
```
hushell: = m_all_intersects?

src/shogun/structure/GEMPLP.h
```diff
@@ -0,0 +1,160 @@
...
+	/** the separator indices (node indices) on each clique */
+	vector<vector<int32_t> > m_clique_separators;
```
hushell: It's nice to have members here since we don't have class Region. However, I suggest to rename separaters -> intersection_sets and clique -> region. Also applies to other members below.

src/shogun/structure/GEMPLP.h
```diff
@@ -0,0 +1,160 @@
...
+	/** the indices (orders in the clique) of the separators on each clique */
+	vector<vector<SGVector<int32_t> > > m_clique_inds_separators;
+	/** store the sum of messages into separators */
+	vector<SGNDArray<float64_t> > m_msgs_into_separators;
+	/** store the messages from cluster to separator */
+	vector<vector<SGNDArray<float64_t> > > m_msgs_from_cluster;
```
hushell: cluster -> region

src/shogun/structure/GEMPLP.h
```diff
@@ -0,0 +1,160 @@
...
+	/** store the original (-) energy of the factors */
+	vector<SGNDArray<float64_t> > m_theta_cluster;
```
hushell: -> m_region_theta
I know they call m_region_lambda, but lambda was used in the paper differently .

src/shogun/structure/GEMPLP.h
Jiaolong: Yes,  int_gap = primal_obj_val - dual_obj_val

src/shogun/structure/MAPInference.cpp
```diff
@@ -38,6 +39,9 @@ CMAPInference::CMAPInference(CFactorGraph* fg, EMAPInferType inference_method)
 		case GRAPH_CUT:
```
hushell: Now I think it's safe to remove other inference types, only keep TreeMaxProd (for tree), GraphCut (for binary loopy graph, better with submodular energy gurantee) and GEMPLP (general loopy graph), since we don't want to implement more methods. The default one choose GEMPLP. In a short time period, we don't want to support 3rd party inference lib.  
@Jiaolong @iglesias Thoughts?

src/shogun/structure/GEMPLP.h
Jiaolong: OK, I agree with you. I just followed DRWN. I will change the names.

src/shogun/structure/GEMPLP.h
Jiaolong: Agree!

src/shogun/structure/MAPInference.cpp
hushell: BTW, since the factor graph is an input, we could check the graph type, and warn if the inference type is not a best fit.

src/shogun/structure/GEMPLP.h
Jiaolong: Sounds very good!

src/shogun/structure/GEMPLP.h
Jiaolong: OK

src/shogun/structure/GEMPLP.h
Jiaolong: OK

src/shogun/structure/GEMPLP.h
Jiaolong: Yes, so I will keep theta.

Jiaolong:
Here I show you the results on yest dataset. The computation is slightly high.

Example 1: tree structure, max-product inference
----------------------------------------------------
SGD trained in   47.0507
sgd solver: average training loss = 0.030553
sgd solver: average testing error = 0.113155

Example 2.1: tree structure, graph-cuts inference
----------------------------------------------------
SGD trained in   42.9520
sgd solver: average training loss = 0.049546
sgd solver: average testing error = 0.125000

Example 2.2: full-connected graph, graph-cuts inference
----------------------------------------------------
SGD trained in   67.9839
sgd solver: average training loss = 0.082026
sgd solver: average testing error = 0.134894

Example 3.1: tree structure, GEMPLP inference
----------------------------------------------------
SGD trained in   76.5377
sgd solver: average training loss = 0.031104
sgd solver: average testing error = 0.113852

Example 3.2: full-connected graph, GEMPLP inference
----------------------------------------------------
SGD trained in  460.3569
sgd solver: average training loss = 0.030966
sgd solver: average testing error = 0.102425

src/shogun/structure/GEMPLP.cpp
```diff
@@ -0,0 +1,368 @@
...
+	SGNDArray<float64_t> lam_sum(cards);
+
+	if (m_theta_cluster[id_clique].len_array == 0)
+		lam_sum.set_const(0);
+	else
+		lam_sum = m_theta_cluster[id_clique].clone();
+
+	int32_t num_separators = m_clique_separators[id_clique].size();
+	vector<SGNDArray<float64_t> > lam_minus; // substract message: \lambda_s^{-c}(x_s)
+	// \sum_{\hat{s}} \lambda_{\hat{s}}^{-c}(x_{\hat{s}}) + \theta_c(x_c)
```
hushell: Is this one the same as the eq in section 6 of [1]?

src/shogun/structure/GEMPLP.cpp
```diff
@@ -0,0 +1,368 @@
...
+	for (int32_t s = 0; s < num_separators; s++)
+	{
+		// maximazation: \max_{x_c} \sum_{\hat{s}} \lambda_{\hat{s}}^{-c}(x_{\hat{s}}) + \theta_c(x_c)
+		SGNDArray<float64_t> lam_max(lam_minus[s].get_dimensions());
+		max_in_subdimension(lam_sum, m_clique_inds_separators[id_clique][s], lam_max);
+		int32_t id_separator = m_clique_separators[id_clique][s];
+		// weighted sum
+		lam_max *= 1.0/num_separators;
+		m_msgs_from_cluster[id_clique][s] = lam_max.clone();
+		m_msgs_from_cluster[id_clique][s] -= lam_minus[s];
```
hushell: lam_minus need to multiply (1-1/num_sep) ?

hushell:
@Jiaolong Good job! I'll check your code out  and test it on my machine, then I probably give more comments later.

src/shogun/structure/MAPInference.cpp
Jiaolong: Yes, I totally agree with you. 

I am also considering to re-factor energy table and use `SGNDArray` to store energies and messages.  

Table factor operation class could also be implemented in the future because the current operation can directly done by `SGNDArray` which is not straightforward and hard to understand. Such ideas are just inspired from `DRWIN`.

What do you think?

Jiaolong:
Here are the results from the MPLP V1.0 library. 

Example 3.1: tree structure, GEMPLP inference
----------------------------------------------------
SGD trained in   80.2897
sgd solver: average training loss = 0.040462
sgd solver: average testing error = 0.120262

Example 3.2: full-connected graph, GEMPLP inference
----------------------------------------------------
SGD trained in  213.0953
sgd solver: average training loss = 0.040462
sgd solver: average testing error = 0.120262

src/shogun/structure/GEMPLP.cpp
Jiaolong: Yes, you are right. In fact, I am confused by this line of code as well. The original implementation does not  use `lam_minus *= 1.0/num_separators;`. I will check it later.

src/shogun/structure/MAPInference.cpp
hushell: Definitely a good idea! Some of functions in CFactorType CFactor for operating energy tables are hard to follow. You may open an issue for this. So either you or I can do it when we have time.

hushell:
@Jiaolong Good to see these comparisons! As we expected, MPLP does a better job than GraphCuts in general graphs and similar to TreeBP in trees. Didn't expect MPLP is so much time consuming, but 460s = <8mins is acceptable? Since GraphCuts was not too bad, maybe better to set it as default.

hushell:
"the results from the MPLP V1.0 library" is keeping their code as before? 
You may want to timing each part of the code to see why the current CGEMPLP is slower. Also check if current CGEMPLP uses more iterations?

Jiaolong:
MPLP V1.0 library is the one that I kept all original implementation.

iglesias:
I am sorry for the long delay, guys, but this pull request is HUGE :-)

tests/unit/structure/GEMPLP_unittest.cc
```diff
@@ -0,0 +1,193 @@
...
+
+	EXPECT_NEAR(min_energy_expected, infer_met.get_energy(), 1E-10);
+
+	SG_UNREF(fg_random);
+	SG_UNREF(fg_test_data);
+}
+
+// Test with SOSVM
+/* This test is time consuming, uncomment it if needed
+TEST(GEMPLP, sosvm)
```
iglesias: Can we make the data for the unit test more minimal so that the test is not time consuming? Unit tests should be quite minimal, using very toy and simple data sets.

tests/unit/structure/GraphCut_unittest.cc
```diff
@@ -168,411 +78,39 @@ TEST(GraphCut, graph_cut_chain)
...
+	CFGTestData* fg_test_data = new CFGTestData();
+	SG_REF(fg_test_data);
+	
+	EXPECT_NEAR(fg_test_data->test_sosvm(GRAPH_CUT), 0, 0.1);
 
-	SG_UNREF(labels_sgd);
-	SG_UNREF(sgd);
-	SG_UNREF(model);
-	SG_UNREF(fg_feats_train);
-	SG_UNREF(fg_labels_train);
```
iglesias: I get a bit scared with all these changes with `UNREF` and `REF`. Did you run valgrind and ensure nothing is leaking?

src/shogun/structure/FGTestData.h
```diff
@@ -0,0 +1,136 @@
...
+
+#include <shogun/structure/MAPInference.h>
+#include <shogun/structure/StochasticSOSVM.h>
+
+#include <shogun/features/FactorGraphFeatures.h>
+#include <shogun/labels/FactorGraphLabels.h>
+
+namespace shogun
+{
+class CFGTestData: public CSGObject
```
iglesias: Can you describe a little bit what is the function of this class?

src/shogun/structure/GEMPLP.h
```diff
@@ -0,0 +1,152 @@
...
+
+#include <vector>
+#include <set>
+
+using namespace std;
+
+namespace shogun
+{
+#define IGNORE_IN_CLASSLIST
+/** GEMPLP (Generalized Max-product LP Relaxation) inference for fatcor graph
```
iglesias: Typo: factor graph.

src/shogun/structure/GEMPLP.h
```diff
@@ -0,0 +1,152 @@
...
+#include <shogun/lib/SGNDArray.h>
+
+#include <vector>
+#include <set>
+
+using namespace std;
+
+namespace shogun
+{
+#define IGNORE_IN_CLASSLIST
```
iglesias: Why are you using this macro here?

iglesias: I am not saying is wrong though, just really asking :-)

src/shogun/structure/GEMPLP.h
```diff
@@ -0,0 +1,152 @@
...
+
+#include <shogun/lib/config.h>
+#include <shogun/lib/SGVector.h>
+#include <shogun/structure/FactorGraph.h>
+#include <shogun/structure/Factor.h>
+#include <shogun/structure/MAPInference.h>
+#include <shogun/lib/SGNDArray.h>
+
+#include <vector>
+#include <set>
```
iglesias: Danger to have these includes if we want these headers to be public to modular interfaces!

src/shogun/structure/GEMPLP.h
```diff
@@ -0,0 +1,152 @@
...
+ * Amir Globerson, Tommi Jaakkola
+ * Advances in Neural Information Processing Systems (NIPS). Vancouver, Canada. 2007.
+ *
+ * [2] Approximate Inference in Graphical Models using LP Relaxations.
+ * David Sontag
+ * Ph.D. thesis, Massachusetts Institute of Technology, 2010.
+ *
+ * The original implementation of GEMPLP can be found:
+ * http://cs.nyu.edu/~dsontag/code/mplp_ver2.tgz
+ * http://cs.nyu.edu/~dsontag/code/mplp_ver1.tgz
```
iglesias: All good with licenses here, I mean, is a new implementation what we have here inspired from the original?

iglesias:
This looks like pretty amazing work, @Jiaolong. Great! Do you think you would be able to showcase in your structure output learning notebook what can we do with it?

tests/unit/structure/GEMPLP_unittest.cc
Jiaolong: Yes, thanks, I will try to make the data smaller.

tests/unit/structure/GraphCut_unittest.cc
Jiaolong: I have run valgrind before, it seemed everything is fine. I can double check it.

src/shogun/structure/FGTestData.h
Jiaolong: Yes, I will add the descriptions.

src/shogun/structure/GEMPLP.h
Jiaolong: Is this macro for avoiding generating python interface? Maybe I am wrong. :)

src/shogun/structure/GEMPLP.h
Jiaolong: Yes, currently, it can not be used for modular interfaces. :(

src/shogun/structure/GEMPLP.h
Jiaolong: Our implementation is not exactly the same as the original one, I mean it is not a copy & paste. However, we followed at least 80% of the original implementation for the core functions, e.g., `update_message()`. The main difference is that our version  is more compatible with the factor graph framework in shogun.

Jiaolong:
@iglesias Thanks for your review! In fact, I have already prepared the notebook. An early version can be found here:
http://nbviewer.ipython.org/gist/Jiaolong/95633b7efd64bbd096f2

I will update it with more detailed explanations and probably with more experiments on other datasets.

tests/unit/structure/GraphCut_unittest.cc
iglesias: If you already did it, then it should be fine ;-) I just wanted to be sure you did check it.

src/shogun/structure/GEMPLP.h
iglesias: You don't need any macro to avoid swig from generating wrappers for a class. This is what you mean by generating Python interface. In fact, a class has to be included in swig files (.i files under interfaces directory) to be accessible to modular interfaces. We use macros when we make a class accessible to interfaces but we do not want some of its methods to be accessible via modular interfaces (but it is not this macro, see e.g. SGVector).

iglesias: Thus, I suggest to remove this line.

src/shogun/structure/GEMPLP.h
iglesias: What does it mean to follow at least 80%? Is it that 80% of the lines are copy & paste but not 100% of them? :-)

src/shogun/structure/GEMPLP.h
iglesias: I understand, however, that we can use gemplp from Python since you have used it in your notebook, right? I guess because you an enumerate or similar mechanism to select the type of approximate inference to use in your MAP class.

That is perfect, then! The interface, in this case, MAPInference class, is accessible to swig and through it one can define whether to use gemplp, etc.

src/shogun/structure/GEMPLP.h
Jiaolong: OK, then I will remove it.

iglesias:
I did not see mplp mentioned anywhere in the notebook. Maybe I overlooked?

src/shogun/structure/GEMPLP.h
```diff
@@ -0,0 +1,152 @@
...
+ */
+IGNORE_IN_CLASSLIST class CGEMPLP: public CMAPInferImpl
```
iglesias: Remember to remove the ignore here as well.

src/shogun/structure/GEMPLP.h
Jiaolong: Yes, you are right! This trick was done by Shell. :)

src/shogun/structure/GEMPLP.h
Jiaolong: No, I mean the algorithm structure, the logic. :)

Jiaolong:
Sorry, it was an old version. The only clue about MPLP is `infer_alg = LP_RELAXATION`. :p And now it should be `infer_alg = GEMPLP`.

src/shogun/structure/GEMPLP.h
Jiaolong: I will do it.

src/shogun/structure/GEMPLP.h
iglesias: We are all right, then.

iglesias:
All right, got it.

Then, what we need for this pull request is:
1) Make unit tests simpler (probably just reducing input data) so we don't need to comment any of them, neither to wait long until they finish. If you want to have a larger test, then an integration one (that is, an example in Python modular) is the right place. Anyway, that would go to a future pull request.
2) Check the function of this FGTestData. I really wonder whether it makes sense to have it in C++. But I need to know more about it first.

Thanks a lot for getting back so quickly, @Jiaolong. Really cool.

Jiaolong:
The reason I created `FGTestData` is that the same testing data are used in multiple unit tests, e.g., GrapCuts, GEMPLP, ect, and probably in the future new algroithms. Thus there are too many redudant code. Another reason is that, we haven't implemented functions to read factor graph from files, so we have to run these code to construct factor graph data.

iglesias:
Then, it sounds like unit tests would be a better place for it, right?

Jiaolong:
Yes, maybe we can move it into unit test folder. However, it can also be used in integral test.

src/shogun/structure/FGTestData.cpp
```diff
@@ -0,0 +1,582 @@
...
+		v_feat.zero();
+
+		for (int32_t j = 0; j < 4 * (i + 1); j++)
+		{
+			int32_t r = random_indices[j];
+			v_feat[r] = 1;
+		}
+
+		for (int32_t f = 0; f < len_feat; f++)
+		{
```
iglesias: One-liner loop, no need for braces. Here and below.

iglesias:
I see your point. To use it from integration tests one could also just generate some data files and push them into the data repository. I admit this is not the most flexible solution though. I leave it up to you, then. If you want to keep the FG data class under structure, it is fine with me.

I will check for other data generators we have in Shogun and see whether it would make more sense there. I will let you know.

iglesias:
There is this DataGenerator class inside the features directory. In my opinion, it is fine to keep this factor graph test data class here under structure, since it is quite specific. @karlnapf?

@Jiaolong, I do suggest to rename it to `FactorGraphDataGenerator`, to follow the convention with other data classes.

Jiaolong:
All right, I will change the name.

Jiaolong:
I just updated the code with your comments. Regarding to the unit tests, valgrind tests are fine:
```
shogun/build/tests/unit$ valgrind ./shogun-unit-test --gtest_filter="GEMP*"
==4564== Memcheck, a memory error detector
==4564== Copyright (C) 2002-2013, and GNU GPL'd, by Julian Seward et al.
==4564== Using Valgrind-3.10.0.SVN and LibVEX; rerun with -h for copyright info
==4564== Command: ./shogun-unit-test --gtest_filter=GEMP*
==4564== 
Note: Google Test filter = GEMP*
[==========] Running 7 tests from 1 test case.
[----------] 7 tests from GEMPLP
[ RUN      ] GEMPLP.find_intersections_index
[       OK ] GEMPLP.find_intersections_index (27 ms)
[ RUN      ] GEMPLP.max_in_subdimension
[       OK ] GEMPLP.max_in_subdimension (8 ms)
[ RUN      ] GEMPLP.initialization
[       OK ] GEMPLP.initialization (64 ms)
[ RUN      ] GEMPLP.convert_energy_to_potential
[       OK ] GEMPLP.convert_energy_to_potential (6 ms)
[ RUN      ] GEMPLP.simple_chain
[       OK ] GEMPLP.simple_chain (32 ms)
[ RUN      ] GEMPLP.random_chain
[       OK ] GEMPLP.random_chain (34 ms)
[ RUN      ] GEMPLP.sosvm
[       OK ] GEMPLP.sosvm (30277 ms)
[----------] 7 tests from GEMPLP (30458 ms total)

[==========] 7 tests from 1 test case ran. (30498 ms total)
[  PASSED  ] 7 tests.
==4564== 
==4564== HEAP SUMMARY:
==4564==     in use at exit: 0 bytes in 0 blocks
==4564==   total heap usage: 7,657,410 allocs, 7,657,410 frees, 164,492,480 bytes allocated
==4564== 
==4564== All heap blocks were freed -- no leaks are possible
==4564== 
==4564== For counts of detected and suppressed errors, rerun with: -v
==4564== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 0 from 0)
```

src/shogun/structure/FactorGraphDataGenerator.h
```diff
@@ -0,0 +1,138 @@
...
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Written (W) 2015 Jiaolong Xu
+ * Copyright (C) 2015 Jiaolong Xu
+ */
+
+#ifndef __FGDATA_H__
```
iglesias: Update the name here as well, please, and in the line below this one.

iglesias:
Great, thanks a lot for the valgrind log. I see however the GEMPLP.sosvm is still taking too long. Can you reduce that time, please?

Jiaolong:
The time shown above is measured under valgrind, so it is longer than running at realse mode. It is `452 ms` on my PC. But anyway, I will try to reduce more. Currently, the SOSVM data has 4 classes and 16 samples.

src/shogun/structure/FactorGraphDataGenerator.h
Jiaolong: Oops!

Jiaolong:
Now the unit test time is reduced to 160 ms.
 `[       OK ] GEMPLP.sosvm (160 ms)`

iglesias:
Oh, sure. Sorry, I did not take into account valgrind when I looked at the running time. 0.4 seconds should be fine, no need to worry more about it.

iglesias:
Perfect, we are ready to go, then. Thanks, @Jiaolong.

Jiaolong:
Many thanks! :D

iglesias:
Feel free to send an update for the notebook when you have the time. It would be great if it includes as well some short discussion about gemplp.

Jiaolong:
Sure! I will try to polish the notebook and send it shortly. Thanks!

hushell:
@Jiaolong @iglesias It's a great news this patch is merged! Sorry I couldn't help for the merge. I spent last fews day on traveling. Thanks for your hardworking! Let's plan our next steps :)

PR #2711, opened 2015-02-16
Some without std::shared_ptr

lisitsyn:
Ok guys lets discuss if this is better

### DIFF
--- a/src/shogun/base/some.h
+++ b/src/shogun/base/some.h
@@ -1,10 +1,9 @@
 #ifndef __SG_SOME_H__
 #define __SG_SOME_H__
 
-#ifdef HAVE_CXX11
 #include <memory>
-
 #include <shogun/base/SGObject.h>
+#include <shogun/lib/RefCount.h>
 
 namespace shogun
 {
@@ -21,13 +20,12 @@ namespace shogun
 	 *
 	 */
 	template <typename T>
-	class Some : protected std::shared_ptr<T>
+	class Some
 	{
+		void* shared;
+
 		public:
-			Some() = delete;
-			Some(const std::shared_ptr<T>& shared);
 			Some(const Some<T>& other);
-			Some(Some<T>&& other);
 			Some& operator=(const Some<T>& other);
 			~Some();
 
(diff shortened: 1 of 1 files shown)


lambday:
@lisitsyn why not using c++11 version?

src/shogun/base/some.h
```diff
@@ -41,62 +39,164 @@ namespace shogun
...
+		__CREATE_SOME(__PARAMETERS_2(um, dois));
+	}
+	template <typename T, typename Um, typename Dois, typename Tres>
+	inline Some<T> some(const Um& um, const Dois& dois, const Tres& tres)
+	{
+		__CREATE_SOME(__PARAMETERS_3(um, dois, tres));
+	}
+	template <typename T, typename Um, typename Dois, typename Tres, typename Quatro>
+	inline Some<T> some(const Um& um, const Dois& dois, const Tres& tres, const Quatro& quatro)
+	{
```
karlnapf: If this is the cost of not having c++, I definitely vote for c++11. This is too messy to read (at least for me)

lisitsyn: Well it is just a cascade of all possible sizes

The main mess comes from no shared_ptr being used

lisitsyn:
@lambday to not force all the shogun be C++11 :) Just an idea

src/shogun/base/some.h
lambday: @lisitsyn only up to 5 constructor params - 5 should suffice and rest should be initialized by setters but still variadic is good - why restrict ? I will also vote for c++11 one :P

iglesias:
I think the methods in Some (constructor, assign, ref, unref, etc) might cause us terrible headaches in the future :-)

How bad would it be to make C++11 a requirement from a certain version of Shogun? I really have no idea whether it is just nuts doing that. What do other open source projects usually do? Are there important compilers for which we would directly lose support?

src/shogun/base/some.h
karlnapf: As we kind of agreed on c++11, lets DO it :)

PR #2721, opened 2015-02-18
move range_fill from SGVector to CMath

curiousguy13:
addresses issue #2582 .
i'll remove this method from SGVector.h and SGVector.cpp and  change other files that are using this method once it's finalised here.

### DIFF
--- a/src/shogun/mathematics/Math.h
+++ b/src/shogun/mathematics/Math.h
@@ -511,6 +511,33 @@ class CMath : public CSGObject
 			x = x*(1.5f - xhalf*x*x); // One round of Newton's method
 			return x;
 		}
+		
+		/** Range fill a vector with start...start+len-1
+		 *
+		 * @param v the vector to be filled
+		 * @param len length of the vector v
+		 * @param start - value to be assigned to first element of vector
+		 */
+		template <class T>
+			static void range_fill(SGVector<T> v, int32_t len, T start=0)
+			{
+				REQUIRE(len>0 && len<=v.size(),"Vector length is out of bounds.\n");
+				range_fill(v.vector,len,start);
+			}
+
+		/** Range fill a vector with start...start+len-1
+		 *
+		 * @param v the vector to be filled
+		 * @param len length of the vector v
+		 * @param start - value to be assigned to first element of vector
+		 */
+		template <class T>
+			static void range_fill(T* v, int32_t len, T start=0)
+			{
+
+				for(index_t i=0; i<len; ++i)
+					v[i]=i+start;
+			}
 
 		/**
 		 * @name Exponential methods (x^n)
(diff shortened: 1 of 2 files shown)


curiousguy13:
@iglesias could you check this out?

lambday:
@curiousguy13 I think this should go in `linalg` instead of CMath. `range_fill` can work for `SGMatrix` as well. Please have a look at [`linalg` README](https://github.com/shogun-toolbox/shogun/wiki/README_linalg)

The idea is to add it to `Core` module in `linalg`. 
1. You add a `NATIVE` implementation (see [`Add.h`](https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/mathematics/linalg/internal/implementation/Add.h#L63)) and add a convenient global function in modules (see ['Core.h'](https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/mathematics/linalg/internal/modules/Core.h#L69)). You can simply use `std::iota` in our code.
2. Write a unit-test for that
3. grep for `range_fill` in shogun and replace those with your newly added `linalg::range_fill()`
4. Remove `range_fill` from SGVector.

tests/unit/mathematics/Math_unittest.cc
```diff
@@ -510,4 +510,31 @@ TEST(CMath, dot)
...
+{
+	SGVector<index_t> v(4);
+
+	CMath::range_fill(v,v.vlen);
+
+	EXPECT_EQ(v.vlen, 4);
+	EXPECT_EQ(v[0], 0);
+	EXPECT_EQ(v[1], 1);
+	EXPECT_EQ(v[2], 2);
+	EXPECT_EQ(v[3], 3);
```
lambday: why not check them in loop?

src/shogun/mathematics/Math.h
```diff
@@ -511,6 +511,33 @@ class CMath : public CSGObject
...
+		/** Range fill a vector with start...start+len-1
+		 *
+		 * @param v the vector to be filled
+		 * @param len length of the vector v
+		 * @param start - value to be assigned to first element of vector
+		 */
+		template <class T>
+			static void range_fill(SGVector<T> v, int32_t len, T start=0)
+			{
+				REQUIRE(len>0 && len<=v.size(),"Vector length is out of bounds.\n");
```
iglesias: Print also the length and the size, please.

iglesias:
Remove the method in SGVector (therefore, change the calls to CMath) in the same PR.

I also agree with Rahul that this should better go to linalg. In any case, I am also fine I you finish this PR putting it in CMath and then in a later one you move it from CMath to linalg. It would not be the most efficient move doing it in two steps, but your choice :-)

curiousguy13:
@iglesias i think moving range_fill in two steps would be quite inefficient , so i'll move range_fill directly to linalg in my next PR according to the steps mentioned by @lambday .
I'll close this PR then.

Issue #2508, opened 2014-08-19
A Meta-Language for Shogun examples


karlnapf:
GSoC project description
 * http://shogun-toolbox.org/page/Events/gsoc2014_ideas#meta_language

Some discussion on the mailing list
 * http://permalink.gmane.org/gmane.comp.ai.machine-learning.shogun/4319

Related github issues
 * #1883
 * #1882

Other:
 * Question on other mailing list: https://groups.google.com/forum/#!topic/antlr-discussion/FcP9_seAaLk

PR #2722, opened 2015-02-18
Meta-Language Example Generation

sorig:
This pull request adds an example generation tool that simplifies that maintenance of the shogun API examples. It does so by implementing a meta-language in which examples can be written and automatically be translated to each target language. 

The following target languages are currently supported: java, R, python, and octave. More languages can easily be added (in most cases) by adding a JSON file with translation rules under the `example-generation/targets` folder.

I have added an option to the main `CMakeLists.txt` called `GENERATE_EXAMPLES` that translates all examples in the `examples` folder under `example-generation`. If `ENABLE_TESTING` is set, these translations are added as tests (I have not yet gotten this to work for the java examples, but I could not get it to work for the existing examples in `undocumented/java_modular/` either).

Please see the [README.md](https://github.com/sorig/shogun/blob/develop/examples/example-generation/README.md) file for more info and documentation.

See also: #2508


examples/example-generation/examples/gp.sg
```diff
@@ -0,0 +1,56 @@
...
+CSVFile Y("../data/gaussian_process_Y.csv")
+CSVFile Y_test("../data/gaussian_process_Y_test.csv")
+
+RegressionLabels labels(Y)
+RegressionLabels labels_test(Y_test)
+RealFeatures feats_train(X)
+RealFeatures feats_test(X_test)
+
+# GP specification
+GaussianKernel kernel(10, 2)
```
lisitsyn: Idea: can we add keywords here to annotate parameter names? `width=2`

examples/example-generation/examples/gp.sg
```diff
@@ -0,0 +1,56 @@
...
+#from pylab import plot, show, legend, fill_between
+#plot(feats_train[0],labels,'x') # training observations
+#plot(feats_test[0],labels_test,'-') # ground truth of test
+#plot(feats_test[0],mean, '-') # mean predictions of test
+#fill_between(feats_test[0],mean-1.96*sqrt(variance),mean+1.96*sqrt(variance),color='grey')  # 95% confidence interval
+#legend(["training", "ground truth", "mean predictions"])
+
+#show()
+
+print "Alpha:"
```
lisitsyn: Can we support `print x,y,...,z` syntax that concatenates all the stuff?

examples/example-generation/types/typelist
```diff
@@ -0,0 +1,873 @@
+ObjectDetector
```
lisitsyn: Should this file be in or it can be dynamic?

examples/example-generation/examples/gp.sg
sorig: We could support statements like that but if you are suggesting that the meta-language should actually look up the parameter names in the Shogun API in order to get the positions of the named arguments, that would complicate things quite a bit (a lot actually).
Furthermore, when the examples are translated these kinds of named arguments would be lost in all target languages that do not support this feature.

Say you have a meta-language example with the line (supposing the meta-language now supports this kind of statement):

```
GaussianKernel kernel(size=10, width=2)
```
Then this could potentially be translated to a python program as:

```
kernel = GaussianKernel(size=10, width=2)
```

But in Java this would just be translated to:

```
GaussianKernel kernel = new GaussianKernel(2, 10);
```

So for many target languages this wouldn't be very useful (keep in mind that most users will only look at the translations, not the meta-language versions). Another approach is to simply write the meta-language examples in a more explanatory way. E.g.

```
int size = 10
int width = 2
GaussianKernel kernel(size, width)
```
Which would translate well to all target languages.

examples/example-generation/examples/gp.sg
sorig: I guess this would be possible. I suspect there might be some complications casting the objects in some languages and target languages may handle string concatenation in different ways, but I think these issues should be manageable. 

My only concern with this is that the translations may turn out to be more complicated than the meta-language version. One could imagine that for some target language a statement like `print x,y,..,z` would have to be turned into several print statements with string casts all over the place, which might not be desirable (I can't think of a concrete example of this off the top of my head).

But if you think this feature would be really useful, it should definitely be doable.

examples/example-generation/types/typelist
sorig: I'm not sure to be honest. I'm not familiar with this class. The typelist is simply scraped from the `.i` files in `src/interfaces/modular` as you can see in `example-generation/types/gettypelist`. 

Please let me know if I need to do something more smarter than that.

examples/example-generation/examples/gp.sg
lisitsyn: No, I'd suggest to not bother if it takes more than an hour

examples/example-generation/examples/gp.sg
lisitsyn: I think it is ok to drop them or make the keyword a comment. Don't want to push this but if it is easy - would be cool

examples/example-generation/types/gettypelist
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
```
lisitsyn: Could you please name it `get_type_list.sh`?

examples/example-generation/types/typelist
lisitsyn: That's perfect but it would be better if that file is generated on each build so that we don't have to change it manually

examples/example-generation/examples/gp.sg
karlnapf: Good idea. Nice to have.

But maybe we can do this after the patch is merged? Also example translation is next step I would say, as well as adding more targets.

examples/example-generation/types/typelist
karlnapf: That is actually true. Could you add this to the CMake stuff and update the patch?

karlnapf:
Thanks for reviewing, @lisitsyn 

@sorig I think running the script that scrapes the .i files should happen at build time. So thats the only thing to be updated before this can be merged (at least from my side, lets see whether the others find time to review - we are currently overloaded with GSoC applicants)

Very nice work!

examples/example-generation/examples/liblinear.sg
```diff
@@ -0,0 +1,16 @@
...
+BinaryLabels labels(label_data)
+
+LibLinear svm(0.9, feats_train, labels)
+svm.set_liblinear_solver_type(enum LIBLINEAR_SOLVER_TYPE.L2R_L2LOSS_SVC_DUAL)
+svm.set_epsilon(0.001)
+svm.set_bias_enabled(True)
+svm.train()
+BinaryLabels predictions = svm.apply_binary(feats_test)
+
+print predictions.get_labels()
```
kevinhughes27: nitpick - bunch of unix new lines are missing

kevinhughes27:
super cool!

kevinhughes27:
also with this are we going to simplify the examples folder and consolidate the undocument and documented examples? I've always thought this was confusing

lisitsyn:
@kevinhughes27 yeah I remember back in 2011 I was thinking: why all the examples are stored in a folder that sounds like it is a trash or something

sorig:
Thanks for the feedback everyone! I will get around to the suggested changes ASAP. 

I have been doing some testing with a larger number of example files and it turns out that the pyparsing library used in `parse.py` is a major bottleneck. I have tried out two alternative approaches: handwriting a LL(1) parser, and using the `PLY` (python lex-yacc) library instead (which is a LALR parser generator).
Both of these alternative solutions are fast (about equally fast), but the handwritten parser is not very maintainable as it relies on hardcoding a LL(1) parse table which requires a moderate amount of knowledge of parsing methods. On the other hand, the `PLY` library is not part of the standard python distribution like `pyparsing` is, so using this library introduces a new dependency.

You can see the `parse.py` reimplementation using `PLY` [here](https://github.com/sorig/shogun/blob/fastparsing/examples/example-generation/parse.py)

Below I've put timing comparisons of parsing and translating 100 example files on my machine. The first is using the current implementation that relies on `pyparsing` and the second is the reimplementation using `PLY`
```
time python generate.py 

real    0m38.380s
user    0m37.393s
sys 0m0.676s
```

```
time python generate.py 

real    0m0.665s
user    0m0.576s
sys 0m0.081s
```

I would like to hear your opinion on what approach to go with - introducing a dependency on the `PLY` library or handwriting the parser without any dependencies.

karlnapf:
@kevinhughes27 @lisitsyn 
yes, we aim to completely replace the existing examples and migrate them into the new system.
We just need patience until this works, so I suggest to set things up next to each other and have the GSoC applicants translate the examples.

karlnapf:
@sorig Just checked, in Ubuntu, PLY is supported (not just part of it, but supported by Canonical)

Since python has a super easy package managment system, I dont see too many problems. However, it would still be better to not have this dependency in. Are there any other options? Re-implementing things is also not an option. What about the C++ generators?

@sonney2k @vigsterkr @besser82 what do you think?

kevinhughes27:
:sparkles:

karlnapf:
Thought about it again, what about offering two versions? Fast parsing would then be optional, but still everyone could generate examples?

lisitsyn:
Python dependencies are so easy to install so I'd not care

karlnapf:
I had severe troubles with this on a cluster machine on which I use Shogun. So I think having both versions, where the fast one is optional when PLY is available would be a good way to go.

sorig:
@karlnapf Using a C++ parser generator would introduce a similar dependency although it would probably be more cumbersome to install compared to a python library. Another option is using ANTLR, but the generated parsers it produces depend on the ANTLR runtime, so not much difference here either. 

Offering two versions is an option, but I'm thinking it's quite annoying to maintain two versions of essentially the same program. 

If we are going to introduce a dependency on a parser generator then I'd say something like PLY is easiest to handle because it's just a matter of installing a python library. The handwritten parser is still a pretty good option although slightly more difficult to maintain. However, assuming that the grammar of the meta-language won't change much, this won't be a very big problem (it would still be easy to add new target languages and maintain the translation program).

Personally I'd say PLY is the best option while we're transitioning to using the meta-language as the main way of writing Shogun examples. Once we have a good number of examples in the meta-language and we're confident that the grammar won't change, we can always switch to a handwritten parser and remove the dependency.

karlnapf:
The handwritten parser has the problem that nobody will ever touch it again, which is bad. We had this scenario before, so I would like to avoid that -- even though you are right that we probably don't need to touch it again.

PLY is the best option in my eyes too. However, it would be extremely annoying for people to not be able to produce any examples at all. Imagine you install Shogun, are not too familiar with python libraries, and then you cannot look at examples. That is why we *need* something that runs with no dependencies at all. I agree it's annoying to maintain two version of the same program, but its less annyoing that maintaining a manual parser, so I think that would be the way to go.

What are your thoughts?

sorig:
Alright, let's have two versions then. If the grammar isn't changing much, I guess it's not a big deal anyways.

How should we handle this in terms of the build process then? In case PLY isn't installed, running cmake with `GENERATE_EXAMPLES` enabled will be be very slow. Should we just output a cmake warning telling users that installing PLY will speed up example generation?

sorig:
I have now merged the two parser versions into a [parse.py](https://github.com/sorig/shogun/blob/fastparsing/examples/example-generation/parse.py) that detects wether PLY is installed and chooses the parser implementation based on that.

karlnapf:
Great, I like this.
Maybe add a unit test on one or two examples to make sure both parsers output the same?

Yes, a warning would be the best I think.

With this, we are soon ready to merge this.

karlnapf:
OK, I merge this now.
Let's see what happens.

@vigsterkr could you add a new build with this enabled?

karlnapf:
requests:
Could you  move the README.md into the doc/md folder?
This will make it editable from the wiki. Then we can put a symlink to this folder?

examples/example-generation/translate.py
```diff
@@ -0,0 +1,303 @@
...
+    except IOError, err:
+        if err.errno == 2:
+            print "Target \"" + targetJsonPath + "\" does not exist"
+        else:
+            print err
+        raise Exception("Could not load target dictionary")
+
+if __name__ == "__main__":
+    # Extract target language from arguments (default target is python)
+    targets = [target for target in map(lambda arg:arg[len("--target="):] if "--target=" in arg else "", sys.argv) if target != ""]
```
lisitsyn: Sorry I get back about that a bit late

There is a better standard api to do that: https://docs.python.org/2/library/argparse.html. Could you please change the code to use that so it gets much more flexible and easier to read?

sorig: You're right. I had already switched to argparse for `parse.py` and `generate.py`, but have for some reason forgotten about `translate.py`. Fix: #2821

PR #2724, opened 2015-02-18
Added maybe - a holder for potentially absent values

lisitsyn:
Work in progress but please comment

### DIFF
--- a/src/shogun/base/maybe.h
+++ b/src/shogun/base/maybe.h
@@ -112,7 +112,7 @@ namespace shogun
          *
          * Doesn't throw any exception.
          */
-        inline T value_or(const T& v)
+        inline T& value_or(T& v)
         {
             if (is_present())
                 return *get();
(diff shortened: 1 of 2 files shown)


karlnapf:
Out of my competence scope. Can you explain a bit or point to references?

lisitsyn:
@karlnapf it is useful if a function sometimes can not return anything. If you use pointer semantics absent value is nullptr but in case of value semantics one better use optional/maybe

karlnapf:
I see. We want this to make code look nicer?
I don't see the different between ```NULL->method()``` and an error caused by absent value apart from cosmetics? Or do I miss something?

lisitsyn:
@karlnapf yes that's safer and it has value semantics instead of pointer semantics

karlnapf:
safer in the sense that we dont cause segfaults or read uninitialised memory? Nicer because no pointer semantics?

Cool I get it then.

lisitsyn:
Just pushed reworked Maybe

vigsterkr:
:two_hearts:

lisitsyn:
Travis errors unrelated (gosh we have to fix travis), merging

PR #2725, opened 2015-02-19
add a kernel to support FITC inference

yorkerlin:
@karlnapf 
take a look at this

### DIFF
--- a/src/shogun/machine/gp/GaussianARDFITCKernel.cpp
+++ b/src/shogun/machine/gp/GaussianARDFITCKernel.cpp
@@ -1,0 +1,146 @@
+/*
+ * Copyright (c) The Shogun Machine Learning Toolbox
+ * Written (w) 2015 Wu Lin
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * The views and conclusions contained in the software and documentation are those
+ * of the authors and should not be interpreted as representing official policies,
+ * either expressed or implied, of the Shogun Development Team.
+ *
+ */
+
+#include <shogun/machine/gp/GaussianARDFITCKernel.h>
+
+using namespace shogun;
+
+CGaussianARDFITCKernel::CGaussianARDFITCKernel() : CGaussianARDKernel()
+{
+	init();
+}
+void CGaussianARDFITCKernel::init()
+{
+}
+
+CGaussianARDFITCKernel::~CGaussianARDFITCKernel()
+{
+}
+
+#ifdef HAVE_CXX11
+#include <shogun/mathematics/linalg/linalg.h>
+CGaussianARDFITCKernel::CGaussianARDFITCKernel(int32_t size, float64_t width)
+		: CGaussianARDKernel(size,width)
+{
+	init();
+}
+
+CGaussianARDFITCKernel::CGaussianARDFITCKernel(CDotFeatures* l,
+		CDotFeatures* r, int32_t size, float64_t width)
+		: CGaussianARDKernel(l, r, size, width)
+{
+	init();
+}
+
+CGaussianARDFITCKernel* CGaussianARDFITCKernel::obtain_from_generic(CKernel* kernel)
+{
+	if (kernel->get_kernel_type()!=K_GAUSSIANARDFITC)
+	{
+		SG_SERROR("Provided kernel is not of type CGaussianARDFITCKernel!\n");
+	}
+
+	/* since an additional reference is returned */
+	SG_REF(kernel);
+	return (CGaussianARDFITCKernel*)kernel;
+}
+
+#include <typeinfo>
+CSGObject *CGaussianARDFITCKernel::shallow_copy() const
+{
+	// TODO: remove this after all the classes get shallow_copy properly implemented
+	// this assert is to avoid any subclass of CGaussianARDFITCKernel accidentally called
+	// with the implement here
+	ASSERT(typeid(*this) == typeid(CGaussianARDFITCKernel))
+	CGaussianARDFITCKernel *ker = new CGaussianARDFITCKernel(cache_size, m_width);
+	if (lhs)
+	{
+		ker->CGaussianARDKernel::init(lhs, rhs);
+	}
+	return ker;
+}
+
+SGMatrix<float64_t> CGaussianARDFITCKernel::get_parameter_gradient(
+		const TParameter* param, index_t index)
+{
+	if (!strcmp(param->m_name, "latent_features"))
+	{
+		REQUIRE(lhs && rhs, "Features not set!\n")
+		REQUIRE(index>=0 && index<num_lhs,"")
+		int32_t idx_l=index;
+		SGVector<float64_t> left_vec=((CDotFeatures *)lhs)->get_computed_dot_feature_vector(idx_l);
+		SGMatrix<float64_t> res(left_vec.vlen, num_rhs);
+
+		for (int32_t idx_r=0; idx_r<num_rhs; idx_r++)
+		{
+			SGVector<float64_t> right_vec=((CDotFeatures *)rhs)->get_computed_dot_feature_vector(idx_r);
+			SGMatrix<float64_t> res_transpose(res.get_column_vector(idx_r),1,left_vec.vlen,false);
+			linalg::add(left_vec, right_vec, right_vec, 1.0, -1.0);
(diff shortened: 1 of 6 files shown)


yorkerlin:
@karlnapf 
There may be another the issue in `if` in different situations. 
Let me know if it is an issue and I will fix it

yorkerlin:
@karlnapf  or @iglesias 
please restart the failed travis' tasks

src/shogun/machine/gp/GaussianARDFITCKernel.h
```diff
@@ -0,0 +1,119 @@
...
+	 */
+	virtual const char* get_name() const { return "GaussianARDFITCKernel"; }
+
+	/** destructor */
+	virtual ~CGaussianARDFITCKernel();
+
+private:
+	void init();
+
+#ifdef HAVE_CXX11
```
karlnapf: I guess hiding the whole class is better than just hiding parts?

@lambday what about the c++ being turned on. I really would like to avoid these guards

src/shogun/machine/gp/GaussianARDFITCKernel.h
```diff
@@ -0,0 +1,119 @@
...
+
+	/** @param kernel is casted to CGaussianARDFITCKernel, error if not possible
+	 * is SG_REF'ed
+	 * @return casted CGaussianARDFITCKernel object
+	 */
+	static CGaussianARDFITCKernel* obtain_from_generic(CKernel* kernel);
+
+
+	/** Make a shallow copy of the kernel */
+	virtual CSGObject* shallow_copy() const;
```
karlnapf: please get rid of this method

src/shogun/machine/gp/GaussianARDFITCKernel.h
```diff
@@ -0,0 +1,119 @@
...
+	 *
+	 * @return name GaussianARDFITCKernel
+	 */
+	virtual const char* get_name() const { return "GaussianARDFITCKernel"; }
+
+	/** destructor */
+	virtual ~CGaussianARDFITCKernel();
+
+private:
+	void init();
```
karlnapf: Should be documented at least with one sentence

src/shogun/machine/gp/GaussianARDFITCKernel.h
```diff
@@ -0,0 +1,119 @@
...
+#include <shogun/kernel/GaussianARDKernel.h>
+
+namespace shogun
+{
+
+/** @brief Gaussian Kernel with Automatic Relevance Detection with supporting
+ * FITC inference
+ *
+ * This kernel supports to compute the gradient wrt latent features (inducing points),
+ * which are not hyper-parameters of the kernel.
```
karlnapf: Some references would be good here:
 * other classes that can use this
 * paper that explains inducing points? Maybe the JMLR ?

src/shogun/machine/gp/GaussianARDFITCKernel.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+SGMatrix<float64_t> CGaussianARDFITCKernel::get_parameter_gradient(
+		const TParameter* param, index_t index)
+{
+	if (!strcmp(param->m_name, "latent_features"))
+	{
+		REQUIRE(lhs && rhs, "Features not set!\n")
+		REQUIRE(index>=0 && index<num_lhs,"")
+		int32_t idx_l=index;
+		SGVector<float64_t> left_vec=((CDotFeatures *)lhs)->get_computed_dot_feature_vector(idx_l);
```
karlnapf: static casts without type checks....
Can this fail from wrong user input? if so, please put a type check in a require

src/shogun/machine/gp/GaussianARDFITCKernel.cpp
```diff
@@ -0,0 +1,146 @@
...
+	}
+	return ker;
+}
+
+SGMatrix<float64_t> CGaussianARDFITCKernel::get_parameter_gradient(
+		const TParameter* param, index_t index)
+{
+	if (!strcmp(param->m_name, "latent_features"))
+	{
+		REQUIRE(lhs && rhs, "Features not set!\n")
```
karlnapf: Please require them seperately for user to fix things easier

src/shogun/machine/gp/GaussianARDFITCKernel.cpp
```diff
@@ -0,0 +1,146 @@
...
+		REQUIRE(index>=0 && index<num_lhs,"")
```
karlnapf: please dont do empty error messages

src/shogun/machine/gp/GaussianARDFITCKernel.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+	/* since an additional reference is returned */
+	SG_REF(kernel);
+	return (CGaussianARDFITCKernel*)kernel;
+}
+
+#include <typeinfo>
+CSGObject *CGaussianARDFITCKernel::shallow_copy() const
+{
+	// TODO: remove this after all the classes get shallow_copy properly implemented
```
karlnapf: I see

src/shogun/machine/gp/GaussianARDFITCKernel.cpp
```diff
@@ -0,0 +1,146 @@
...
+	{
+		SG_SERROR("Provided kernel is not of type CGaussianARDFITCKernel!\n");
+	}
+
+	/* since an additional reference is returned */
+	SG_REF(kernel);
+	return (CGaussianARDFITCKernel*)kernel;
+}
+
+#include <typeinfo>
```
karlnapf: should go on top

src/shogun/machine/gp/GaussianARDFITCKernel.cpp
```diff
@@ -0,0 +1,146 @@
...
+void CGaussianARDFITCKernel::init()
+{
+}
+
+CGaussianARDFITCKernel::~CGaussianARDFITCKernel()
+{
+}
+
+#ifdef HAVE_CXX11
+#include <shogun/mathematics/linalg/linalg.h>
```
karlnapf: all includes on top please

src/shogun/machine/gp/GaussianARDFITCKernel.cpp
```diff
@@ -0,0 +1,146 @@
...
+			}
+			else
+			{
+				if(m_ARD_type==KT_DIAG)
+				{
+					SGMatrix<float64_t> weights(m_weights.matrix,1,m_weights.num_rows,false);
+					linalg::elementwise_product(right_transpose, weights, res_transpose);
+				}
+				else if(m_ARD_type==KT_FULL)
+				{
```
karlnapf: no {}

src/shogun/machine/gp/GaussianARDFITCKernel.cpp
```diff
@@ -0,0 +1,146 @@
...
+					linalg::matrix_product(right_transpose, m_weights, res_transpose);
+				}
+				else
+				{
```
karlnapf: No {}

src/shogun/machine/gp/GaussianARDFITCKernel.cpp
```diff
@@ -0,0 +1,146 @@
...
+					SG_ERROR("Unsupported ARD kernel\n");
```
karlnapf: The error message should be more clear. "Unsupported ARD type, see <name_of_enum>\"

karlnapf:
Cool!
What about a unit test?

src/shogun/machine/gp/GaussianARDFITCKernel.h
yorkerlin: This may be a issue.
see https://github.com/shogun-toolbox/shogun/pull/2676
@lambday suggested that it is better not to touch `class_list.cpp.py`

The following is my question related to this issue
It seems that `src/shogun/base/class_list.cpp` will automatically  create `CLinearARDKernel` and `CGaussianARDKernel` objects.
However, when `HAVE_CXX11` is not defined, these two ARD kernels are empty, which leads to a error in travis. 
I know there is a trick to `create empty` `CLinearARDKernel` and `CGaussianARDKernel` objects in   `class_list.cpp`  to address this error.

However, do you know how to exclude  these two kernels in  `class_list.cpp`  when `HAVE_CXX11` is not defined? In order words,  `class_list.cpp` will `NOT create`  `CLinearARDKernel` and `CGaussianARDKernel` objects

yorkerlin: Let me know which approach is better (empty instances or excluding these classes).

src/shogun/machine/gp/GaussianARDFITCKernel.h
yorkerlin: add the reference first since there are several fitc methods to be implemented

src/shogun/machine/gp/GaussianARDFITCKernel.cpp
yorkerlin: This kernel is a subclass of `CDotKernel`.  
`CDotKernel` requires `CDotFeatures` as input.
Let me know if you think extra checking is better

yorkerlin:
@karlnapf 
This kernel is tested in the upcoming `FITC` PRs since the existing `FITCInferenceMethod` does not support to compute the gradient wrt `latent features`

src/shogun/machine/gp/GaussianARDFITCKernel.h
karlnapf: I see. Leave the guard for now

I think @lambday is pushing to require c++11 to build Shogun. From then, we can just delete all guards.

src/shogun/machine/gp/GaussianARDFITCKernel.h
karlnapf: Maybe reference the class where this is used from?

src/shogun/machine/gp/GaussianARDFITCKernel.cpp
karlnapf: nops thats fine, but please put a comment for future developers to see

karlnapf:
Ok, all good to go. I will merge, but could you make a note of the minor style issues I noticed and address them in the next patch?

lambday:
@karlnapf @yorkerlin hey guys one thing I forgot to mention (sorry I didn't notice in details) - `HAVE_CXX11` guards are sufficient for those linalg methods for which we **have** native implementation. For others (those rely on external libs) we should guard with `HAVE_LINALG_LIB`. 

It is probably a good idea to make a list of native methods and put it in a list somewhere in the wiki.

src/shogun/machine/gp/GaussianARDFITCKernel.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+#include <shogun/machine/gp/GaussianARDFITCKernel.h>
+
+using namespace shogun;
+
+CGaussianARDFITCKernel::CGaussianARDFITCKernel() : CGaussianARDKernel()
+{
+	init();
+}
+void CGaussianARDFITCKernel::init()
```
lisitsyn: Please rename it as it 'clashes' with init(CFeatures*, CFeatures*) and causes warning on clang

src/shogun/machine/gp/GaussianARDFITCKernel.h
yorkerlin: see this
https://github.com/shogun-toolbox/shogun/issues/2735
I will remove it

src/shogun/machine/gp/GaussianARDFITCKernel.cpp
yorkerlin: done 
see
https://github.com/shogun-toolbox/shogun/issues/2735
For now, I will remove this function.

yorkerlin:
@lambday 
any suggestion about when we should guard `HAVE_LINALG_LIB` and/or `HAVE_CXX11`

lambday:
@yorkerlin since not all the operations have native implementation, I'd say that its safer to use have_linalg_lib guard instead of c++11 guard *as of now*. we'll anyway review all of these later.

Issue #2729, opened 2015-02-20
Take `linalg::add` out of `HAVE_LINALG_LIB` guard


lambday:
[This method](https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/mathematics/linalg/internal/modules/Core.h#L69) has to be taken out of the `HAVE_LINALG_LIB` guard, now that we have the native implementation. In absence of external linalg lib, this makes the compilation fail in [Add_unittest](https://github.com/shogun-toolbox/shogun/blob/develop/tests/unit/mathematics/linalg/Add_unittest.cc#L66)

lambday:
Done. Closing.

PR #2731, opened 2015-02-20
Fix for issue #2729

lambday:
- moved `add`, `max`, and `vector_sum` outside the `HAVE_LINALG_LIB` guard
- guarded `CLinearARDKernel` and `CGaussianARDKernel` with `HAVE_LINALG_LIB` instead of `HAVE_CXX11` since the linalg methods used there are still dependent on third party libs and are not available natively (yet).


lambday:
Travis errors unrelated. Merging.

PR #2733, opened 2015-02-22
Update SGSparseMatrix.cpp

dkostka:
Fixes segmentation fault when num_features > num_vectors.

### DIFF
--- a/src/shogun/lib/SGSparseMatrix.cpp
+++ b/src/shogun/lib/SGSparseMatrix.cpp
@@ -199,7 +199,7 @@ template<class T> SGSparseMatrix<T> SGSparseMatrix<T>::get_transposed()
 
 	SG_FREE(hist);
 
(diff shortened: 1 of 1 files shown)


lambday:
@dkostka thanks for the PR. Could you also please add a unit test for this? Say, three tests 
- one with square matrix 
- non-square matrix with num_feats > num_vec
- non-square matrix with num_feats < num_vec

You call get_transposed on those and verify sanity. See `SGSparseMatrix_unittest.cc`

iglesias:
Merging the bug fix now. @dkostka, it would be great if you could add the unit tests that Rahul has mentioned.

sudk1896:
@iglesias: Among the tests that @lamday mentioned. 2 and 3 are covered (`get_transposed_more_features` and `get_transposed_more_vectors`. I am adding the test for Transposing a square matrix.

Issue #2730, opened 2015-02-20
Update list of methods


karlnapf:
http://shogun-toolbox.org/doc/en/current/methods.html

We have *tons* of stuff in Shogun that is not listed here. This (easy) task is to populate the methods list.
Just go through examples and check whether the method is listed, if not, add.

The page is generated from:
https://github.com/shogun-toolbox/shogun/blob/develop/doc/pages/Methods.mainpage

This will be later used for our user-guide.

H

karlnapf:
Super high priority and easy - so go for it!

jimsotelo:
Has this been taken? I'm trying to get familiar with the shogun toolbox and this looks like a good place to start.

iglesias:
Yes. See #2738.

karlnapf:
Great!

But there are way more classes, so I re-open

adit-39:
Will fix that in a couple of hours by taking every file in the src folder
and its subfolders and only after that generate the list with classes
missing unit tests and examples! :)
Every *.h (except for template classes) in src is eligible to make this
list?

karlnapf:
Thats a good way to start, but keep in mind the list of methods should not contain every class we have, but every algorithm we implement, with a reference to the class(es) it is implemented in.

You don't need to look for missing unit tests, we have a test coverage script that does this, see buildbot

rohit12:
Is this closed? I would like to work on this.

iglesias:
@karlnapf can probably elaborate what else needs to be done here. Anyway, what I understand from his comment is that this list should not only be a list of class names but rather of ML methods/algorithms that are implemented by the classes.

karlnapf:
I think we can leave this for now as the last update was nice and we soon will switch to the new manual.
If you are keen, you can polish this a bit, many things are unresolved, general feedback is appreciated:

https://github.com/karlnapf/shogun_manual

karlnapf:
Closing for now

PR #2738, opened 2015-02-24
Issue #2730: Update Methods list [FIX]

adit-39:
Have updated the *completed*  **features**, **preprocesssors**, **kernels**, **distance measures** and **methods** based on the implementations found under respective directories in *src/shogun* instead of only the examples folder. Hope that is alright as it seemed to contain a more complete list than that in examples.

### DIFF
--- a/doc/pages/Methods.mainpage
+++ b/doc/pages/Methods.mainpage
@@ -47,7 +47,15 @@
   \li CDummyFeatures - Features without content; Only number of vectors is known.
   \li CExplicitSpecFeatures - Implement spectrum kernel feature space explicitly.
   \li CImplicitWeightedSpecFeatures - DotFeatures that implicitly implement weighted spectrum kernel features.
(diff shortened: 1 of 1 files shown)


iglesias:
Nice job, @adit-39. Good decision to check the directories instead of the examples. Since you are on it, it would be also nice if you detect the classes that are missing unit tests and/or api examples. Those should be implemented so we can, for instance, start creating entrance issues for them.

adit-39:
@iglesias sure, will get down to it! :)

adit-39:
@iglesias do you want separate lists of those without unit tests, those without examples and those without both?

iglesias:
A GitHub issue sounds reasonable to me. In the issue, you can have two separate lists: one with the missing unit tests, another one with the missing examples.

As they start gradually getting done, we can update those lists. Also, the PRs addressing missing unit tests and/or examples should reference the issue.

PR #2742, opened 2015-02-27
add the base FITC class

yorkerlin:
@karlnapf 
take a look at this

### DIFF
--- a/src/shogun/machine/gp/FITCInferenceBase.h
+++ b/src/shogun/machine/gp/FITCInferenceBase.h
@@ -1,0 +1,325 @@
+/*
+ * Copyright (c) The Shogun Machine Learning Toolbox
+ * Written (W) 2015 Wu Lin
+ * Written (W) 2013 Roman Votyakov
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * The views and conclusions contained in the software and documentation are those
+ * of the authors and should not be interpreted as representing official policies,
+ * either expressed or implied, of the Shogun Development Team.
+ *
+ */
+
+#ifndef CFITCINFERENCEBASE_H
+#define CFITCINFERENCEBASE_H
+
+#include <shogun/lib/config.h>
+
+#ifdef HAVE_EIGEN3
+
+#include <shogun/machine/gp/InferenceMethod.h>
+
+namespace shogun
+{
+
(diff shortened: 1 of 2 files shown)


yorkerlin:
@karlnapf 
there are three layers of class hierarchy for fitc inference.
This class is in the first layer, which is used for general fitc inference. (regression, binary classification and multi-classification). 

For multi-classification (including Laplace method) , currently the design only focuses on the `single kernel` scheme, which means each class uses the same kernel. (this base class for now is not implemented since fitc may not be roust for multi-classification )

Once this class is merged, I will send a base class for single fitc inference (regression and binary classification)

src/shogun/machine/gp/FITCInferenceBase.h
```diff
@@ -0,0 +1,291 @@
...
+
+#include <shogun/machine/gp/InferenceMethod.h>
+
+namespace shogun
+{
+
+/** @brief The Fully Independent Conditional Training inference base class.
+ *
+ * This inference method computes the Cholesky and Alpha vectors approximately
+ * with the help of latent variables. For more details, see "Sparse Gaussian
```
karlnapf: should we maybe consider renaming the latent features to "inducing variables" and also reference the JMLR paper "a unifying view on sparse GPs" ?

karlnapf: Oh, and the math of the model would be good here, including the inducing features

src/shogun/machine/gp/FITCInferenceBase.h
```diff
@@ -0,0 +1,291 @@
...
+	virtual SGMatrix<float64_t> get_cholesky();
+
+	/** update all matrices */
+	virtual void update()=0;
+
+	/** set the noise for inducing points
+	 *
+	 * @param noise noise for inducing points
+	 *
+	 * The noise is used to enfore the kernel matrix about the inducing points are positive definite
```
karlnapf: It would be good to have the math in the model for such things, and then explicitly reference what in the model this represents?

src/shogun/machine/gp/FITCInferenceBase.h
```diff
@@ -0,0 +1,291 @@
...
+
+	/** get the noise for inducing points
+	 *
+	 * @return noise noise for inducing points
+	 *
+	 */
+	virtual float64_t get_inducing_noise();
+
+	/** returns derivative of negative log marginal likelihood wrt latent features (input)
+	 * Note that in order to call this method, kernel must support FITC inference
```
karlnapf: please reference a method to check whether the kernel supports FITC inference, the enum, a method, or whatever

src/shogun/machine/gp/FITCInferenceBase.h
```diff
@@ -0,0 +1,291 @@
...
+	 * \f$p(f|y)\f$ exactly, and it returns covariance matrix \f$\Sigma\f$ of
+	 * the posterior Gaussian distribution \f$\mathcal{N}(\mu,\Sigma)\f$
+	 * otherwise.
+	 *
+	 * @return covariance matrix
+	 */
+	virtual SGMatrix<float64_t> get_posterior_covariance();
+
+protected:
+	/** convert latent features and features to the same represention
```
karlnapf: maybe consider renaming "latent" to "inducing" globally?

src/shogun/machine/gp/FITCInferenceBase.h
```diff
@@ -0,0 +1,291 @@
...
+	 *
+	 * Note that these two kinds of features can different types.
```
karlnapf: A verb is missing "can different types"

src/shogun/machine/gp/FITCInferenceBase.h
```diff
@@ -0,0 +1,291 @@
...
+	 *
+	 * @return derivative of negative log marginal likelihood
+	 */
+	virtual SGVector<float64_t> get_derivative_wrt_inference_method(
+			const TParameter* param)=0;
+
+	/** returns derivative of negative log marginal likelihood wrt parameter of
+	 * likelihood model
+	 *
+	 * @param param parameter of given likelihood model
```
karlnapf: state math of log marginal likelihood?

Also maybe say that it is the *approximate* log marginal likelihood

karlnapf:
I understand what you have in mind with these three layers. Sounds good
There are a number of documentation glitches. Could you fix them?

src/shogun/machine/gp/FITCInferenceBase.h
yorkerlin: next PR?
Focus on the math part first

src/shogun/machine/gp/FITCInferenceBase.h
yorkerlin: it is done in the base class for single fitc inference (regression and binary classification) since in the future this class may support multi-kernels.

src/shogun/machine/gp/FITCInferenceBase.h
yorkerlin: in the introduction of the class.

src/shogun/machine/gp/FITCInferenceBase.h
yorkerlin: updated

yorkerlin:
any suggestion about this PR?

yorkerlin:
one question about posterior `p(f_n | y)` of GP, where  `y` are labels, `f_n` are (latent) function points and `f_m` are inducing points.

There are two version of posterior 
* true posterior (for regression, is exact posterior) (for Laplace, is approximated posterior)
* approximated posterior by inducing points  (which is similar to the predicted distribution(

In some cases, the true posterior may be difficult to compute (which means O(n^3)) while the approximated posterior should be easy to compute (which means O(m*n^2))

For `get_posterior_mean()` and `get_posterior_covariance()`, which version I should use?

karlnapf:
Hi Wu,

I think whenevery somebody uses FITC based inference, we should return the approximation coming from the inducing points

karlnapf:
We only want the O(n^3) approximation if we *don't* use the inducing framework

karlnapf:
I am fine with doing the rest in the next PR

karlnapf:
travis is fine except for unrelated things. merging!
NICE! :)

Issue #2717, opened 2015-02-16
Add scale and add to linalg


karlnapf:
Search for "bvec.scale(-1.0);" in here
https://github.com/shogun-toolbox/shogun/pull/2676/files#r24749044

Talk to @lambday to get guidance.

iglesias:
I see both [scale](https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/mathematics/linalg/internal/modules/Core.h#L85) and [add](https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/mathematics/linalg/internal/modules/Core.h#L69) in linalg that should work for vectors. 

@lambday, is that right?

lambday:
@iglesias yes. The task is then to replace existing SGVector scale calls with linalg one.

lambday:
as well as remove scale from SGVector once everything is replaced. Links to SGVector cleanup issue #2582 as well :)

nginn:
Hi. I'm interested in contributing to Shogun and would like to start with this issue. @lambday is there anything else about this issue that I should know before starting?

lambday:
@nginn Hey! Thanks for your interest. This task is relatively simple - replace the existing member method calls for add and scale with linalg alternatives. Once it is done, we'd want the member methods for these two removed. Make sure you compile and run all the unit tests locally after your changes.

nginn:
Thanks, got it. Although, after doing some work on this issue, I have several questions:
1) how should I rely on unit-testing results if there are 4 failed tests currently? (build from d8eb73dd89f47e0da28f07163c4f635b96d0ec00) Should I consider it as a good result if after my changes there will be still the same number of failed tests?)
2) does this issue address the method SGVector::scale_vector, as long as SGVector::operator=+, too?

lambday:
Hi @nginn!

1) Yes! Your patch shouldn't introduce any new failing unit-tests than that are there  ;)
2) Good point. I think it does, because the overloaded `+` and `+=` operators use `add`. `scale_vector` should be fine for now since it doesn't depent on `scale`.

Suggestions: 
- handle `scale` and `add` in two different patches. Send one first, once that's merged, send the other patch.
- For both, you'd have to use the in-place version of the linalg methods. 
[...]

PR #2744, opened 2015-02-27
move scale to linalg

SunilMahendrakar:
towards #2717. Everything is not yet replaced.

### DIFF
--- a/src/shogun/mathematics/linalg/internal/implementation/Scale.h
+++ b/src/shogun/mathematics/linalg/internal/implementation/Scale.h
@@ -76,7 +76,7 @@ struct scale<Backend::NATIVE, Matrix>
 	static void compute(SGMatrix<T> A, SGMatrix<T> B, T alpha)
 	{
 		REQUIRE((A.num_rows==B.num_rows)&&(A.num_cols==B.num_cols),
-			"Dimensions of A(%dx%d) and B(%dx%d) don't match.\n",
+			"Dimensions of A(%dx%d) and B(%dx%d) don't match.",
 			A.num_rows, A.num_cols, B.num_rows, B.num_cols);
 		
 		compute(A.matrix, B.matrix, A.num_rows*A.num_cols, alpha);
@@ -86,7 +86,7 @@ struct scale<Backend::NATIVE, Matrix>
 	static void compute(SGVector<T> A, SGVector<T> B, T alpha)
 	{
 		REQUIRE(A.vlen==B.vlen,"Number of elements in A(%d) should be "
-			"equal to number of elements in (%d)!\n", A.vlen, B.vlen);
+			"equal to number of elements in (%d)!", A.vlen, B.vlen);
 		
 		compute(A.vector, B.vector, A.vlen, alpha);
 	}
@@ -94,7 +94,7 @@ struct scale<Backend::NATIVE, Matrix>
 	/** Performs the operation B = alpha*A for len elements*/	
 	static void compute(T* A, T* B, index_t len, T alpha)
 	{
-		REQUIRE(A!=NULL&&B!=NULL, "invalid pointers to matrices");
+		REQUIRE(A!=NULL&&B!=NULL, "Invalid pointers to matrices.");
 		
 		for (index_t i=0; i<len; i++)
 			B[i]=A[i]*alpha;
--- a/src/shogun/mathematics/linalg/internal/implementation/Scale.h
+++ b/src/shogun/mathematics/linalg/internal/implementation/Scale.h
@@ -66,6 +66,41 @@ struct scale
 	static void compute(Matrix A, Matrix B, Matrix C, T alpha, T beta);
 };
 
+/** Specialization of scale for the Native backend */
+template<class Matrix>
+struct scale<Backend::NATIVE, Matrix>
+{
+	typedef typename Matrix::Scalar T;
+	
+	/** Performs the operation B = alpha*A */
+	static void compute(SGMatrix<T> A, SGMatrix<T> B, T alpha)
+	{
+		REQUIRE((A.num_rows==B.num_rows)&&(A.num_cols==B.num_cols),
+			"Dimensions of A(%dx%d) and B(%dx%d) don't match.\n",
+			A.num_rows, A.num_cols, B.num_rows, B.num_cols);
+		
+		compute(A.matrix, B.matrix, A.num_rows*A.num_cols, alpha);
+	}
+	
+	/** Performs the operation B = alpha*A */
+	static void compute(SGVector<T> A, SGVector<T> B, T alpha)
+	{
+		REQUIRE(A.vlen==B.vlen,"Number of elements in A(%d) should be "
+			"equal to number of elements in (%d)!\n", A.vlen, B.vlen);
+		
+		compute(A.vector, B.vector, A.vlen, alpha);
+	}
+
+	/** Performs the operation B = alpha*A for len elements*/	
+	static void compute(T* A, T* B, index_t len, T alpha)
+	{
+		REQUIRE(A!=NULL&&B!=NULL, "invalid pointers to matrices");
+		
+		for (index_t i=0; i<len; i++)
+			B[i]=A[i]*alpha;
+	}
+};
+
 #ifdef HAVE_EIGEN3
 
 /** Specialization of scale for the Eigen3 backend */
--- a/src/shogun/mathematics/linalg/internal/modules/Core.h
+++ b/src/shogun/mathematics/linalg/internal/modules/Core.h
@@ -61,6 +61,20 @@ void add(Matrix A, Matrix B, Matrix C, typename Matrix::Scalar alpha=1.0,
 	implementation::add<backend, Matrix>::compute(A, B, C, alpha, beta);
 }
 
+/** Performs the operation B = alpha*A. Works for both matrices and vectors */
+template <Backend backend=linalg_traits<Core>::backend,class Matrix>
+void scale(Matrix A, Matrix B, typename Matrix::Scalar alpha)
+{
+	implementation::scale<backend, Matrix>::compute(A, B, alpha);
+}
+
+/** Performs the operation A = alpha*A. Works for both matrices and vectors */
+template <Backend backend=linalg_traits<Core>::backend,class Matrix>
+void scale(Matrix A, typename Matrix::Scalar alpha)
+{
+	implementation::scale<backend, Matrix>::compute(A, A, alpha);
+}
+
 #ifdef HAVE_LINALG_LIB
 /** Performs matrix multiplication
  *
@@ -85,13 +99,6 @@ void subtract(Matrix A, Matrix B, Matrix C,
 	typename Matrix::Scalar alpha=1.0, typename Matrix::Scalar beta=1.0)
 {
 	implementation::add<backend, Matrix>::compute(A, B, C, alpha, -1*beta);
-}
-
-/** Performs the operation B = alpha*A. Works for both matrices and vectors */
-template <Backend backend=linalg_traits<Core>::backend,class Matrix>
-void scale(Matrix A, Matrix B, typename Matrix::Scalar alpha)
-{
-	implementation::scale<backend, Matrix>::compute(A, B, alpha);
 }
 
 /** Performs the operation C = A .* B where ".*" denotes elementwise multiplication */
--- a/tests/unit/mathematics/linalg/Scale_unittest.cc
+++ b/tests/unit/mathematics/linalg/Scale_unittest.cc
@@ -47,6 +47,38 @@
 
 using namespace shogun;
 
+TEST(ScaleMatrix, native_backend)
+{
+	const float64_t alpha = 0.3;
+	
+	SGMatrix<float64_t> A(9,9);
+	SGMatrix<float64_t> B(9,9);
+	
+	for (int32_t i=0; i<9*9; i++)
+		A[i] = i;
+	
+	linalg::scale<linalg::Backend::NATIVE>(A, B, alpha);
+	
+	for (int32_t i=0; i<9*9; i++)
+		EXPECT_NEAR(alpha*A[i], B[i], 1e-15);
+}
+
+TEST(ScaleVector, native_backend)
+{
+	const float64_t alpha = 0.3;
+	
+	SGVector<float64_t> A(9);
+	SGVector<float64_t> B(9);
+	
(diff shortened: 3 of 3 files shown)


src/shogun/statistics/QuadraticTimeMMD.cpp
```diff
@@ -1023,7 +1024,7 @@ SGVector<float64_t> CQuadraticTimeMMD::sample_null_spectrum_DEPRECATED(
 
 	/* when m=n, return m*MMD^2 instead */
 	if (m==n)
-		null_samples.scale(0.5);
+		linalg::scale(null_samples, null_samples, 0.5);
```
karlnapf: hey

just curious, can't there be a convenience method for the case where both arugments are the same?

src/shogun/structure/CCSOSVM.cpp
```diff
@@ -538,7 +539,7 @@ SGSparseVector<float64_t> CCCSOSVM::find_cutting_plane(float64_t* margin)
 		if (result->psi_computed)
 		{
 			new_constraint.add(result->psi_truth);
-			result->psi_pred.scale(-1.0);
+			linalg::scale(result->psi_pred, result->psi_pred, -1.0);
```
karlnapf: for example here, this is bad due to the pointer de-referencing --- imagine this is an expensive operation

src/shogun/statistics/QuadraticTimeMMD.cpp
SunilMahendrakar: hi,
We'll have to add a simple method. I'll add it.

src/shogun/structure/CCSOSVM.cpp
SunilMahendrakar: Agreed.

SunilMahendrakar:
@karlnapf please take a look.

src/shogun/mathematics/linalg/internal/implementation/Scale.h
```diff
@@ -66,6 +66,20 @@ struct scale
...
+struct scale<Backend::NATIVE, Matrix>
+{
+		typedef typename Matrix::Scalar T;
+		
+		/** Performs the operation B = alpha*A */
+		static void compute(SGVector<T> A, SGVector<T> B, T alpha)
+		{
+			for (int32_t i=0; i<A.vlen; i++)
+				B[i]=A[i]*alpha;
+		}
```
lambday: Some indentation issue is there.

src/shogun/mathematics/linalg/internal/implementation/Scale.h
```diff
@@ -66,6 +66,20 @@ struct scale
...
 };
 
+/** Specialization of scale for the Native backend */
+template<> template<class Matrix>
+struct scale<Backend::NATIVE, Matrix>
+{
+		typedef typename Matrix::Scalar T;
+		
+		/** Performs the operation B = alpha*A */
+		static void compute(SGVector<T> A, SGVector<T> B, T alpha)
```
lambday: This should be able to handle both SGMatrix and SGVector. We can have two methods
`static void compute(SGMatrix<T> A, SGMatrix<T> B, T alpha)`
and
`static void compute(SGVector<T> A, SGVector<T> B, T alpha)`
both of these would call internally a
`static void compute(T* A, T* B, T alpha)`
where you pass the number of elements you want to scale. That would be consistent with the Eigen3 and ViennaCL specializations we have for scale.

src/shogun/mathematics/linalg/internal/implementation/Scale.h
```diff
@@ -66,6 +66,20 @@ struct scale
 	static void compute(Matrix A, Matrix B, Matrix C, T alpha, T beta);
 };
 
+/** Specialization of scale for the Native backend */
+template<> template<class Matrix>
```
lambday: extraneous `template <>`. I know we have it in linalg at many places but these are not really necessary. This is simply a partial specialization. We should aim at removing the existing ones. gcc doesn't complain but with clang we get a warning for this.

src/shogun/mathematics/linalg/internal/modules/Core.h
```diff
@@ -94,6 +94,13 @@ void scale(Matrix A, Matrix B, typename Matrix::Scalar alpha)
 	implementation::scale<backend, Matrix>::compute(A, B, alpha);
 }
 
+/** Performs the operation A = alpha*A. Works for both matrices and vectors */
+template <Backend backend=linalg_traits<Core>::backend,class Matrix>
+void scale(Matrix A, typename Matrix::Scalar alpha)
```
lambday: really useful :)

src/shogun/mathematics/linalg/internal/implementation/Scale.h
```diff
@@ -66,6 +66,20 @@ struct scale
...
+struct scale<Backend::NATIVE, Matrix>
+{
+		typedef typename Matrix::Scalar T;
+		
+		/** Performs the operation B = alpha*A */
+		static void compute(SGVector<T> A, SGVector<T> B, T alpha)
+		{
+			for (int32_t i=0; i<A.vlen; i++)
```
lambday: There should be a sanity check for size, in case the sizes of A and B differ.

lambday:
Hi @SunilMahendrakar. Sorry for such a late reply. Have been caught up with something. 

The PR looks good. I have some comments and it would be great if you could please address those - shouldn't take more than 5 mins. I know GSoC is not happening for us this summer which is depressing, but it would be great if we can finish at least those things that we started. Looking forward to it.

SunilMahendrakar:
@lambday Sure!. GSoC would have been great but I'll be contributing anyways.

src/shogun/mathematics/linalg/internal/implementation/Scale.h
SunilMahendrakar: @lambday both Eigen3 and ViennaCL don't have ```static void compute(T* A, T* B, index_t num_elements, T alpha)``` in the spcializations we have. I was about to add it since it would be easy for moving other ```SGVector::scale``` calls to ```linalg::scale```

lambday: That's fine. The main thing is that these specializations handle both matrices and vector. For our native implementation we can add any helper method we need to achieve that purpose. This is practically a different class altogether.

lambday:
That's really nice to hear :) Looking forward.

SunilMahendrakar:
@lambday done :) please take a look.

src/shogun/mathematics/linalg/internal/implementation/Scale.h
```diff
@@ -66,6 +66,34 @@ struct scale
...
+	static void compute(SGVector<T> A, SGVector<T> B, T alpha)
+	{
+		REQUIRE(A.vlen==B.vlen,"Number of elements in A(%d) should be "
+			"equal to number of elements in (%d)!\n", A.vlen, B.vlen);
+		
+		for (int32_t i=0; i<A.vlen; i++)
+			B[i]=A[i]*alpha;
+	}
+};
+
```
lambday: @SunilMahendrakar well what I meant here was to have an additional method that just takes raw pointer types and the length of the consecutive memory array. 
```c++
static void compute(T* A, T* B, index_t len, T alpha)
{
    // REQUIRE on null check for both A and B
    // the loop goes here
}
```
So you write the loop there just once and make the SGMatrix and SGVector ones to call that method instead. That way, we'll be able to handle handle raw arrays as well (we can later simply add another wrapper which calls that with NATIVE backend). See what I mean?

lambday:
Hi @SunilMahendrakar . This looks good. Just see my other comment. 

A couple of other things I forgot to mention 

- as of now C++11 is not a hard requirement for Shogun, but it is so for linalg. So whenever *using* linalg methods, we should (at present) guard our code with `HAVE_CXX0X` or `HAVE_CXX11`. Otherwise compilation will fail on platforms that ironically still use prehistoric compilers. The rule of thumb is -
    * If the linalg method *has* a native implementation, we should guard our code with `HAVE_CXX0X` or `HAVE_CXX11`. (we would get rid of this requirement soon!)
    * If the linalg method *doesn't* have a native implementation and only available with external linalg libs, we should guard it with `HAVE_LINALG_LIB`
    * If the linalg method only has implementation against a particular backend, we should guard it for that particular lib (`HAVE_EIGEN3` or `HAVE_VIENNACL`).
- Whenever we add a native implementation for some method, the purpose of that is to make it available even in absence of any external linalg libs. So in the wrapper, we should pull that method out of `HAVE_LINALG_LIB` guard. See that [`add`](https://github.com/SunilMahendrakar/shogun/blob/scale/src/shogun/mathematics/linalg/internal/modules/Core.h#L58) method resides outside that guard.

src/shogun/mathematics/linalg/internal/implementation/Scale.h
SunilMahendrakar: @lambday yeah. so that we just have a REQUIRE there which calls this method to perform the actual computation.

SunilMahendrakar:
@lambday thanks for the explanation. :)

src/shogun/mathematics/linalg/internal/implementation/Scale.h
lambday: Yeah. Keep the other requires there - they are useful. The null check REQUIRE is useful when people (hopefully mistakenly) passes scale(NULL, 1000, 0.5) in the raw ptr method.

SunilMahendrakar:
@lambday it would be better if we first work towards removing HAVE_CXX11 first then move the calls to linalg, since if we want to prevent compilation failures we would have to add the guard all over the place start from where we use it, i.e for things dependent on the method where we use it. Its easy for outermost methods like unit tests. Since we are going to lose it anyway lets start working towards losing it.

lambday:
@SunilMahendrakar that totally makes sense. We want the build to fail in cmake itself in absence of c++11. 

However, until that happens, I'd suggest that replacing everything with linalg calls can be done later, in separate patches. As of now, we'd be fine just adding it to linalg and having unit-tests. So don't replace everything yet - let's first get native scale merged.

lambday:
Please rebase.

SunilMahendrakar:
sure.

src/shogun/mathematics/linalg/internal/implementation/Scale.h
```diff
@@ -66,6 +66,41 @@ struct scale
...
+template<class Matrix>
+struct scale<Backend::NATIVE, Matrix>
+{
+	typedef typename Matrix::Scalar T;
+	
+	/** Performs the operation B = alpha*A */
+	static void compute(SGMatrix<T> A, SGMatrix<T> B, T alpha)
+	{
+		REQUIRE((A.num_rows==B.num_rows)&&(A.num_cols==B.num_cols),
+			"Dimensions of A(%dx%d) and B(%dx%d) don't match.\n",
```
iglesias: I think we can skip the \n.

src/shogun/mathematics/linalg/internal/implementation/Scale.h
```diff
@@ -66,6 +66,41 @@ struct scale
...
+			A.num_rows, A.num_cols, B.num_rows, B.num_cols);
+		
+		compute(A.matrix, B.matrix, A.num_rows*A.num_cols, alpha);
+	}
+	
+	/** Performs the operation B = alpha*A */
+	static void compute(SGVector<T> A, SGVector<T> B, T alpha)
+	{
+		REQUIRE(A.vlen==B.vlen,"Number of elements in A(%d) should be "
+			"equal to number of elements in (%d)!\n", A.vlen, B.vlen);
```
iglesias: Here as well.

src/shogun/mathematics/linalg/internal/implementation/Scale.h
```diff
@@ -66,6 +66,41 @@ struct scale
...
+		
+		compute(A.vector, B.vector, A.vlen, alpha);
+	}
+
+	/** Performs the operation B = alpha*A for len elements*/	
+	static void compute(T* A, T* B, index_t len, T alpha)
+	{
+		REQUIRE(A!=NULL&&B!=NULL, "invalid pointers to matrices");
```
iglesias: Following the style used above, I would start the message with capital i and finish with a stop.

iglesias:
It looks good to me, just very minor comments above. @lambday, ready to merge?

src/shogun/mathematics/linalg/internal/implementation/Scale.h
lambday: @iglesias I personally like the `\n` in the end - as per our wiki on [Assertions] -(https://github.com/shogun-toolbox/shogun/wiki/Assertions).

- Use **proper** English (Start with capital letters, write sentences)
- Don't forget `\n` at the end.
- Make sure that you **don't cause segfaults** in your assertions, check bounds, pointer valid, etc

Although it is just a matter of choice.

tests/unit/mathematics/linalg/Scale_unittest.cc
```diff
@@ -47,6 +47,38 @@
...
+	
+	for (int32_t i=0; i<9; i++)
+		A[i] = i;
+	
+	linalg::scale<linalg::Backend::NATIVE>(A, B, alpha);
+	
+	for (int32_t i=0; i<9; i++)
+		EXPECT_NEAR(alpha*A[i], B[i], 1e-15);
+}
+
```
lambday: @SunilMahendrakar could you please add a couple of these tests (one on SGMatrix and another on SGVector) that does the scale operation in place? This way, `linalg::scale(A, alpha)` is also tested.

lambday:
@SunilMahendrakar great job! Thanks for the patch :) Merging. If you could please add two more unit-tests that I mentioned, we can consider the matter close :)

tests/unit/mathematics/linalg/Scale_unittest.cc
lambday: Actually, it makes sense to have the *in place* scale operation for **all** the backends - since we are exposing it for all of 'em.

SunilMahendrakar: sure!

SunilMahendrakar:
@lambday buildbots are going crazy. test failing unrelated?

lambday:
@SunilMahendrakar yeah `SerializationJSON` is a long term serial killer. We should rename buildbot to blamebot.

Issue #2719, opened 2015-02-17
Make C++11 support mandatory for Shogun 4.x.x


lambday:
Let's vote here. Immediate benefits -
- We can use all the cool stuffs in `linalg` without having to worry about `HAVE_CXX11` guards
- @lisitsyn can use `std::shared_ptr` for `Some`
- GSoC students can use cool C++11 things this year only - for_each, lambdas, type traits.

@iglesias and @karlnapf are already on board. @vigsterkr @sonney2k @lisitsyn @besser82 what do you think? We just say that in order to build Shogun v4.x.x, you **got** to have a C++11 supported compiler - I don't see any harm in that.

iglesias:
+1. See also #2711.

sonney2k:
We can do that. I am sure that it will make shogun not compile on some cc's and archs. But we have to name the release that requires it 5.0!

So Ok from my side.

karlnapf:
+1 And also on the renaming, which is true.

iglesias:
Shall we do a 5.0 release soon then so from pre-gsoc time people can start
using c++11 with no problem?

What does Viktor think?

lambday:
@vigsterkr @lisitsyn @besser82 ?? 
I'll vote for pre-gsoc. Its redundant to write *useful things* from scratch when we *have it there already* in c++11. Of course we can write those on our own but 
(1) they need to be properly tested for bugs which is tricky sometimes.
(2) students will feel more comfy with existing c++ stuffs.

karlnapf:
Lets move soon. But not sure whether we should do another release before GSoC.
5.0 after GSoC is fine.

If you want to release before this switch, also good! But we did not fix too many things since then right?
Releasing as often as possible is a good practice though

lambday:
5.0 is a bit too much for just this change I think! :D 
major version change should include some *major* modification I think! how about we do it in our next cleaned up version within 4.x.x.?

iglesias:
In 4.x.x. sounds good to me. Soeren, any reason why it should be 5.0?

karlnapf:
Requiring a new standard is a major change in my eyes too. But well, it is a matter of taste, just change it in the Dev branch and we can discuss numbers later

sonney2k:
@lambday Well just require C++11 from now on but don't do a release until it's ready. So no need to do things twice/cumbersome.

If we have bugfixes - lets make a 4.0.X release.

lambday:
Already resolved. Closing.

PR #2745, opened 2015-02-28
added range_fill in linalg

curiousguy13:
[INCOMPLETE]
@lambday I had a couple of questions regarding this:
1)Should I place a guard like #ifdef HAVE_LINALG_LIB  before including linalg.h in files?
2)What would be the best way to deal with normal arrays like :
float64_t* w= SG_MALLOC(float64_t, d);
SGVector<float64_t>::range_fill_vector(w, d, 17.0);
should I make a separate method for this?


iglesias:
1) Check in other places where linalg is used. You can find them fast by
using grep.
2) Can they be substituted using SGVector? It is more convenient to use
them, less error- and memory leak-prone.

src/shogun/mathematics/linalg/internal/implementation/RangeFill.h
```diff
@@ -0,0 +1,116 @@
...
+{
+	/**Scalar type */
+	typedef typename Matrix::Scalar T;
+
+	/** Range fill a vector or a matrix with start...start+len-1
+     * @param A - the matrix to be filled
+     * @param len - length of the matrix to be filled
+	 * @param start - value to be assigned to first element of vector or matrix
+	 */	
+	static void compute(Matrix A, index_t len, T start);
```
lambday: This len parameter is not really required. In the specializations, we can extract them from the Matrix type itself.

src/shogun/mathematics/linalg/internal/implementation/RangeFill.h
```diff
@@ -0,0 +1,116 @@
...
+ */
+template <enum Backend, class Matrix>
+struct range_fill
+{
+	/**Scalar type */
+	typedef typename Matrix::Scalar T;
+
+	/** Range fill a vector or a matrix with start...start+len-1
+     * @param A - the matrix to be filled
+     * @param len - length of the matrix to be filled
```
lambday: Something wrong with indentation

src/shogun/mathematics/linalg/internal/implementation/RangeFill.h
```diff
@@ -0,0 +1,116 @@
...
+	/** Range fill a matrix with start...start+len-1
+     * @param A - the matrix to be filled
+     * @param len - length of the matrix to be filled
+	 * @param start - value to be assigned to first element of vector or matrix
+	 */	
+	static void compute(SGMatrix<T> A, index_t len, T start=0)
+	{
+		REQUIRE(len>0,"Entered matrix length (currently %d) should be greater than 0.\n", len);
+		REQUIRE(len<=A.num_rows*A.num_cols,"Entered matrix length (currently %d) should be less than matrix size(%d).\n", len, A.num_rows*A.num_cols);
+		compute(A.matrix, len, start);
```
lambday: We can just pass `A.num_rows * A.num_cols` here for len

lambday: The requires are not needed then

src/shogun/mathematics/linalg/internal/implementation/RangeFill.h
```diff
@@ -0,0 +1,116 @@
...
+
+	/** Range fill a vector with start...start+len-1
+     * @param A - the matrix to be filled
+     * @param len - length of the matrix to be filled
+	 * @param start - value to be assigned to first element of vector or matrix
+	 */	
+	static void compute(SGVector<T> A, index_t len, T start=0)
+	{
+		REQUIRE(len>0,"Entered vector length (currently %d) should be greater than 0.\n", len);
+		REQUIRE(len<=A.size(),"Entered vector length (currently %d) should be less than or equal to vector size(%d).\n", len, A.size());
```
lambday: Same here

src/shogun/mathematics/linalg/internal/implementation/RangeFill.h
```diff
@@ -0,0 +1,116 @@
...
+		compute(A.vector, len, start);
+	}
+
+	/** Range fill a vector or a matrix with start...start+len-1
+     * @param A - the matrix to be filled
+     * @param len - length of the matrix to be filled
+	 * @param start - value to be assigned to first element of vector or matrix
+	 */	
+	static void compute(T* A, index_t len, T start)
+	{
```
lambday: Just a `NULL` check here.

src/shogun/mathematics/linalg/internal/implementation/RangeFill.h
```diff
@@ -0,0 +1,116 @@
...
+{
+	/** Scalar type */
+	typedef typename Matrix::Scalar T;
+
+	/** Range fill a matrix with start...start+len-1
+     * @param A - the matrix to be filled
+     * @param len - length of the matrix to be filled
+	 * @param start - value to be assigned to first element of vector or matrix
+	 */	
+	static void compute(SGMatrix<T> A, index_t len, T start=0)
```
lambday: Default value of start is taken care of in the wrapper method itself. We don't have to put it here.

src/shogun/mathematics/linalg/internal/modules/Core.h
```diff
@@ -61,7 +62,20 @@ void add(Matrix A, Matrix B, Matrix C, typename Matrix::Scalar alpha=1.0,
...
+
 #ifdef HAVE_LINALG_LIB
+
+/** Range fill a vector or a matrix with start...start+len-1
+ * @param A - the matrix to be filled
+ * @param len - length of the matrix to be filled
+ * @param start - value to be assigned to first element of vector or matrix
+ */	
+template <Backend backend=linalg_traits<Core>::backend,class Matrix>
+void range_fill(Matrix A, index_t len, typename Matrix::Scalar start=0.0)
```
lambday: Here ^ (default value for start)

lambday: `len` not required here as well.

src/shogun/mathematics/linalg/internal/implementation/RangeFill.h
```diff
@@ -0,0 +1,116 @@
...
+ */
+template <class Matrix>
+struct range_fill<Backend::NATIVE, Matrix>
+{
+	/** Scalar type */
+	typedef typename Matrix::Scalar T;
+
+	/** Range fill a matrix with start...start+len-1
+     * @param A - the matrix to be filled
+     * @param len - length of the matrix to be filled
```
lambday: Something wrong with indentation

src/shogun/mathematics/linalg/internal/implementation/RangeFill.h
```diff
@@ -0,0 +1,116 @@
...
+	static void compute(SGMatrix<T> A, index_t len, T start=0)
+	{
+		REQUIRE(len>0,"Entered matrix length (currently %d) should be greater than 0.\n", len);
+		REQUIRE(len<=A.num_rows*A.num_cols,"Entered matrix length (currently %d) should be less than matrix size(%d).\n", len, A.num_rows*A.num_cols);
+		compute(A.matrix, len, start);
+	}
+
+	/** Range fill a vector with start...start+len-1
+     * @param A - the matrix to be filled
+     * @param len - length of the matrix to be filled
```
lambday: Something wrong with indentation

src/shogun/mathematics/linalg/internal/implementation/RangeFill.h
```diff
@@ -0,0 +1,116 @@
...
+	static void compute(SGVector<T> A, index_t len, T start=0)
+	{
+		REQUIRE(len>0,"Entered vector length (currently %d) should be greater than 0.\n", len);
+		REQUIRE(len<=A.size(),"Entered vector length (currently %d) should be less than or equal to vector size(%d).\n", len, A.size());
+		compute(A.vector, len, start);
+	}
+
+	/** Range fill a vector or a matrix with start...start+len-1
+     * @param A - the matrix to be filled
+     * @param len - length of the matrix to be filled
```
lambday: Something wrong with indentation

src/shogun/regression/KernelRidgeRegression.cpp
```diff
@@ -59,7 +61,7 @@ bool CKernelRidgeRegression::train_machine_pinv()
 
 	/* tell kernel machine that all alphas are needed as'support vectors' */
 	m_svs=SGVector<index_t>(m_alpha.vlen);
-	m_svs.range_fill();
+	linalg::range_fill<linalg::Backend::NATIVE>(m_svs,m_svs.vlen);
```
lambday: Would be so much neater if we don't have to pass vlen.

lambday:
hi @curiousguy13 
*1)Should I place a guard like #ifdef HAVE_LINALG_LIB before including linalg.h in files?*
-- HAVE_LINALG_LIB guard is required when we're doing something that absolutely requires external linalg libs. For example, those operations in linalg that doesn't have a native implementation. If the operation *does* have a native implementation, a `HAVE_CXX0X` or `HAVE_CXX111` guard would be fine.
*2)What would be the best way to deal with normal arrays like :
float64_t* w= SG_MALLOC(float64_t, d);
SGVector::range_fill_vector(w, d, 17.0);
should I make a separate method for this?*
-- Yes a separate wrapper method would be nice. We have to pass `len` for that method but for the others we don't really need that, as we can extract it from the matrix or vector type itself.

Thanks for the PR. It looks good already. Please check my other comments. Looking forward to your changes. Let's just finish the things we started :)

curiousguy13:
@lambday  this PR might get a bit delayed because I have my mid-term exams from next week (and also range_fill is used in many files) but i'll make the necessary changes and try to finish this PR as soon as possible.(most probably after my mid-terms are over next week).

lambday:
@curiousguy13 absolutely. All the best for the exams. And it is not necessary to make all the replacements in one single PR. You can just finish this PR and later send another one with the replacements. That would be easier to review as well :)

karlnapf:
In fact, grepping for ```SG_MALLOC(float64_t``` would be a good idea :)

src/shogun/mathematics/linalg/internal/implementation/RangeFill.h
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+};
+
+/**
+ * @brief Generic class which is specialized for different backends to perform addition
+ */
+template <enum Backend, class Vector>
+struct range_fill_vec
+{
```
curiousguy13: @lambday I made a new struct for range_fill_vec to handle float64* type of arrays.
Although this seems to work , is there a better way to do this?

lambday: @curiousguy13 no, this has to be handled within the other struct only. Sorry :)
Check the `linalg::add` method implementation to see how it is done there. That can be helpful :)

lambday: Also, don't worry about the template argument being named `Matrix` for both matrix and vector. It doesn't matter. When you pass a `SGVector` there then `Matrix` can very well mean a `SGVector`. Does it make sense? You'll see the same for `Add` there.

lambday: You'll need a separate wrapper though for handling `range_fill(T* A, index_t len)`. By wrapper I mean the methods that goes inside `/modules`, which in turn calls the compute methods in those implementation structs.

curiousguy13: @lambday  first of all sorry for the late reply.
"Also, don't worry about the template argument being named Matrix for both matrix and vector. It doesn't matter. When you pass a SGVector there then Matrix can very well mean a SGVector. Does it make sense?"
yeah i know that Matrix can mean SGVector as well as SGMatrix but afaik the problem here was that Matrix cannot mean double* or int* type of arrays because we are using Matrix::Scalar in the range_fill struct  which throws an error even after calling it from a separate wrapper function in Core.h(as is expected) because double/int don't have scalar type. So to deal with these kind of arrays I made a separate struct to deal with these arrays.
In my code the vector in the second struct is not for SGVector but for these double*/int* type of arrays.
"Check the linalg::add method implementation to see how it is done there. That can be helpful :)"
Yeah, it was helpful but linalg::add doesn't deal with dynamically allocated arrays and neither does any other method in linalg. 
Maybe I'm still missing something. So can you tell me an elegant way to incorporate these arrays and vectors and matrix in a single struct.:)

lambday: Okay here is the plan. 

- I think using separate methods for vectors and matrix is a not a very good idea, leads to a lot of code duplication. Their underlying data is just a plain memory array. Also, in shogun, matrices are always column-major. So treating them in the same manner shouldn't be an issue. I admit that it is done in linalg in quite a few places but that's bad. We should change that.  I just added methods `T* data()` and `index_t size()` in both SGMatrix and SGVector (check PR #2830). So using these we can handle the underlying memory exactly the same way as one would do with raw array types (once that PR is merged).

- Write one struct 
```c++
template <enum Backend, class Container>
struct range_fill
{
    // you don't need to put anything here. Your code should go only with
    // the specializations.
};
```
and partially specialize NATIVE. Access the data using `container.data()` and size using `container.size()`.

- Write two wrappers
```c++
template <Backend backend, class Container> 
range_fill(Container container, typename Container::Scalar begin=0)
{
}
```
and
```c++
template <typename T> // backend not required. raw pointers should work only with NATIVE
range_fill(T* memarray, index_t size, T begin=0)
{
    // construct a SGVector object with refcounting turned off
    SGVector<T> v(memarray, size, false);
    // instantiate the impl template with SGVector<T> type and NATIVE backend and call compute
}
```
- Since this supports NATIVE implementation, keep these above methods outside of the `HAVE_LINALG_LIB` guard (since they can be invoked even in absence of external linear algebra lib).

curiousguy13:
@lambday i did as you suggested and merged methods for SGVector and SGMatrix using data() and size() methods.Raw pointer array is working fine now too.
Any other changes required here(before I cleanup and replace range_fill with linalg::range_fill in codebase) ?

curiousguy13:
@lambday @karlnapf  bump.

karlnapf:
Hi and sorry for the delay. @lambday could you comment?

lambday:
@curiousguy13 Hey! Apologies for late reply. Thanks @karlnapf for tagging me.

The PR itself looks good to me. But there are some issues with travis builds
```
/home/travis/build/shogun-toolbox/shogun/src/shogun/structure/DisjointSet.cpp: In member function ‘void shogun::CDisjointSet::make_sets()’:
/home/travis/build/shogun-toolbox/shogun/src/shogun/structure/DisjointSet.cpp:50:2: error: ‘linalg’ has not been declared
```
Could you please make sure once that these issues are not being caused due to some forgotten header inclusion? Once you do that, could you please rebase once with develop and push again?

curiousguy13:
@lambday I rebased this branch , compiled and ran unit-tests on my machine but I cannot reproduce the errors in Travis-CI . I have included the linalg.h header file wherever I have used linalg.
Can you take a look at the code if I am missing something?

lambday:
Hi @curiousguy13 . These issues are caused by ancient gcc compiler that we're using in travis which doesn't have c++11 support. In absence of c++11 we're missing the entire linalg library which causes these errors in gcc. Although not a good solution, but I think putting a ifdef guard around those calls would solve the issue temporarily.

karlnapf:
what is the state here?

curiousguy13:
@karlnapf @lambday sorry for the delay. Could you check it out once more and what else needs to be done here?

vigsterkr:
couple of problems:
 * why all those indentation changes... please dont do that, makes bisecting incredibly hard.
 * dont change file permission, i.e. 644 -> 755

please fix those before further reviewing.

curiousguy13:
@vigsterkr Thanks for pointing it out. My text editor must have converted all the tabs to spaces since I have been working in Python3 recently. Anyways, check it out now .

src/shogun/features/DotFeatures.cpp
```diff
@@ -358,19 +365,19 @@ void CDotFeatures::benchmark_dense_dot_range(int32_t repeats)
 			dense_dot_range(out, 0, num, alphas, w, d, 23);
 
 #ifdef DEBUG_DOTFEATURES
-    CMath::display_vector(out, 40, "dense_dot_range");
```
vigsterkr: what are all these indentation changes?

src/shogun/mathematics/linalg/internal/implementation/RangeFill.h
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	/**Range fill a vector or a matrix with start...start+len-1
+	 * @param A - the matrix to be filled
+	 * @param len - length of the matrix to be filled
+	 * @param start - value to be assigned to first element of vector or matrix
+	 */
+	static void compute(T* A, index_t len, T start)
+	{
+		std::iota(A, A+len, start);
```
vigsterkr: isn't this c++11 only?

src/shogun/mathematics/linalg/internal/modules/Core.h
```diff
@@ -271,4 +294,4 @@ void convolve(Matrix X, Matrix W, Matrix Y, bool flip = false,
 }
 
 }
-#endif // CORE_H_
+#endif // CORE_H_
```
vigsterkr: add extra empty line as github says

src/shogun/features/DotFeatures.cpp
curiousguy13: I just converted all the spaces to tabs since that is what is mentioned in the coding guide.
It must have been spaces there before.

src/shogun/mathematics/linalg/internal/modules/Core.h
curiousguy13: okay.

src/shogun/mathematics/linalg/internal/implementation/RangeFill.h
curiousguy13: yes std::iota is available from c++11 onwards. would that be a problem?

vigsterkr: yes because i dont see an `ifdef` for it

tests/unit/mathematics/linalg/RangeFill_unittest.cc
```diff
@@ -0,0 +1,80 @@
...
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * The views and conclusions contained in the software and documentation are those
+ * of the authors and should not be interpreted as representing official policies,
+ * either expressed or implied, of the Shogun Development Team.
+ */
+
+#include <shogun/lib/config.h>
+
+#if defined(HAVE_CXX0X) || defined(HAVE_CXX11)
```
vigsterkr: why do you need these?

src/shogun/mathematics/linalg/internal/implementation/RangeFill.h
curiousguy13: As far as i know the linalg library is available only for c++11 onwards. @lambday  please correct me if i'm wrong here.
So, keeping that in mind i used #ifdef HAVE_LINALG_LIB and #else whenever range_fill is used in the code which makes the code compatible with older versions of C++ and C that do not support linalg.

lambday: yeah linalg in c++11 only. the macro `HAVE_LINALG_LIB` itself is not defined when we don't have c++11 supported compiler. the `ifdef` is taken care of in `linalg.h` itself. so only a check of `ifdef HAVE_LINALG_LIB` around the places where it is used would suffice.

lambday: as per the discussion in other threads (#2719, #2919), soon Shogun itself would be made c++11 only. It won't be an issue I guess in that sense.

tests/unit/mathematics/linalg/RangeFill_unittest.cc
curiousguy13: I think these are to ensure that the test would run only in the presence of c++11 since that is what linalg is built for and skip the tests otherwise.

karlnapf: Sounds reasonable to me
But please to them in seperate lines

src/shogun/features/DotFeatures.cpp
```diff
@@ -345,10 +349,13 @@ void CDotFeatures::benchmark_dense_dot_range(int32_t repeats)
 	float64_t* w= SG_MALLOC(float64_t, d);
 	float64_t* out= SG_MALLOC(float64_t, num);
 	float64_t* alphas= SG_MALLOC(float64_t, num);
+#ifdef HAVE_LINALG_LIB
```
karlnapf: ah noooo
This is something I really want to avoid ... double code.
Rather assume that linalg lib is available always.
@lambday @vigsterkr what about just requiring this? We had the same thing for eigen3 and it was a pain -- also lots of bugs in the manual code ...

vigsterkr: @karlnapf linalg requires c++11... imo it's a bit too strict requirement

src/shogun/features/DotFeatures.cpp
vigsterkr: yeah, still dont do this on old code... just on the code you add.

src/shogun/features/DotFeatures.cpp
karlnapf: @curiousguy13 we have to discuss this a bit. Will get back to you

karlnapf: I suggest you keep the guards for now. But please get rid of this NATIVE thing, we will drop that anyways and eigen3 should be the default.

tests/unit/mathematics/linalg/RangeFill_unittest.cc
karlnapf: HAVE_LINALG is the way to do it

src/shogun/features/DotFeatures.cpp
curiousguy13: @vigsterkr any particular reason why we should not try to improve existing code, even if it is just indentation? Am I missing something here?

tests/unit/mathematics/linalg/RangeFill_unittest.cc
curiousguy13: Alright. I'll update it

src/shogun/features/DotFeatures.cpp
vigsterkr: This is not improving this is breaking history of the code... Which makes traceability of a given change quite cumbersome.

curiousguy13: Hmm... fair enough. I'll revert it. Thank you :)

karlnapf:
I think this needs a re-work. It is a mess now. Maybe close and re-open something fresh?

curiousguy13:
continuing in #3071

PR #2755, opened 2015-03-02
unit test for SGSparseMatrix

yingryic:
@iglesias sorry, I rewind my commit from the fork which closed the original pull request. I have corrected the indentation, and changed the variable names "rows and cols" to "features and vectors"


tests/unit/lib/SGSparseMatrix_unittest.cc
```diff
@@ -157,4 +159,110 @@ TEST(SGSparseMatrix, io_libsvm)
 			EXPECT_NEAR(m[i].features[j].entry, m_from_file[i].features[j].entry, 1E-14);
 		}
 	}
```
iglesias: Leave one empty line after you close this block.

tests/unit/lib/SGSparseMatrix_unittest.cc
```diff
@@ -157,4 +159,110 @@ TEST(SGSparseMatrix, io_libsvm)
...
+    //remove file generated by test
```
iglesias: Fix indentation in the comment.

tests/unit/lib/SGSparseMatrix_unittest.cc
```diff
@@ -157,4 +159,110 @@ TEST(SGSparseMatrix, io_libsvm)
...
+{
+    CRandom randGenerator(0);
+    float64_t sparseLevel=0.1;
+    const index_t numberOfFeatures=50;
+    const index_t numberOfVectors=100;
+
+    SGSparseMatrix<float64_t> matrix(numberOfFeatures, numberOfVectors);
+
+    for (index_t featIndex=0; featIndex<numberOfFeatures; ++featIndex)
+        for (index_t vecIndex=0; vecIndex<numberOfFeatures; ++vecIndex){
```
iglesias: Open the brace in a new line.

iglesias: Should this be `vecIndex<numberOfVectors`?

tests/unit/lib/SGSparseMatrix_unittest.cc
```diff
@@ -157,4 +159,110 @@ TEST(SGSparseMatrix, io_libsvm)
...
+    const index_t numberOfFeatures=100;
+    const index_t numberOfVectors=50;
+    SGSparseMatrix<int32_t> sparseMatrix(numberOfFeatures, numberOfVectors);
+
+    for (index_t featIndex=0; featIndex<numberOfFeatures; ++featIndex)
+        for (index_t vecIndex=0; vecIndex<numberOfVectors; ++vecIndex){
+            float64_t randomNumber=randGenerator.random(0.0,1.0);
+            if (randomNumber<=sparseLevel)
+                sparseMatrix(vecIndex,featIndex)=randomNumber;
+        }
```
iglesias: Since you are using the same code to create the sparse matrix in several unit tests, why don't you use pull out that functionality into one function that then your tests call?

tests/unit/lib/SGSparseMatrix_unittest.cc
```diff
@@ -157,4 +159,110 @@ TEST(SGSparseMatrix, io_libsvm)
...
+
+    sparseMatrix.sort_features();
```
iglesias: I am not sure I understand why the use of `sort_features` here. Would you mind to explain, please?

tests/unit/lib/SGSparseMatrix_unittest.cc
```diff
@@ -157,4 +159,110 @@ TEST(SGSparseMatrix, io_libsvm)
...
+            if (randomNumber<=sparseLevel)
+                EXPECT_EQ(matrix(vecIndex,featIndex), randomNumber);
+            else
+                EXPECT_EQ(matrix(vecIndex,featIndex), 0.0);
+        }
+}
+
+TEST(SGSparseMatrix, get_transposed_more_features)
+{
+    CRandom randGenerator(0);//generate random number to control the level of sparsity
```
iglesias: Write comments rather before, in their own line, instead of at the end of a line of code.

iglesias:
Thanks for the new pull request, @yingryic. See the comments above and also have a look at the few comments regarding style left.

tests/unit/lib/SGSparseMatrix_unittest.cc
yingryic: changed

tests/unit/lib/SGSparseMatrix_unittest.cc
```diff
@@ -157,4 +177,101 @@ TEST(SGSparseMatrix, io_libsvm)
 			EXPECT_NEAR(m[i].features[j].entry, m_from_file[i].features[j].entry, 1E-14);
 		}
 	}
+
+    //remove file generated by test
+	unlink("SGSparseMatrix_io_libsvm_output.txt");
```
yingryic: not sure why the indentation is like this, it appears fine before the pull reques

tests/unit/lib/SGSparseMatrix_unittest.cc
yingryic: made a function to generate sparse matrix

tests/unit/lib/SGSparseMatrix_unittest.cc
yingryic: I just follow the test already in the file. Seems we dont need, thus I delete it in the new commit.

yingryic:
@iglesias I have made some change, please review

tests/unit/lib/SGSparseMatrix_unittest.cc
```diff
@@ -21,6 +23,24 @@ using namespace shogun;
 #include <shogun/mathematics/eigen3.h>
 using namespace Eigen;
 
+//function to create sparse matrix
+template<typename T>
+SGSparseMatrix<T> GenerateSparseMatrix(const int32_t numberOfFeatures,
+                                       const int32_t numberOfVectors,
+                                       const float64_t sparseLevel)
```
iglesias: Why making all these const? They are passed by value anyway. It's fine, just wondering.

tests/unit/lib/SGSparseMatrix_unittest.cc
iglesias: Double-check and fix it if so.

tests/unit/lib/SGSparseMatrix_unittest.cc
```diff
@@ -157,4 +177,101 @@ TEST(SGSparseMatrix, io_libsvm)
...
+    const index_t numberOfVectors=100;
+
+    SGSparseMatrix<float64_t> matrix(numberOfFeatures, numberOfVectors);
+
+    for (index_t featIndex=0; featIndex<numberOfFeatures; ++featIndex)
+        for (index_t vecIndex=0; vecIndex<numberOfVectors; ++vecIndex)
+        {
+            float64_t randomNumber=randGenerator.random(0.0,1.0);
+            if (randomNumber<=sparseLevel)
+                matrix(vecIndex,featIndex)=randomNumber;
```
iglesias: Can you call here `GenerateMatrix` as well so that this test only contains the `EXPECT_EQ` in the loops?

iglesias: That way, we reduce lines of code.

tests/unit/lib/SGSparseMatrix_unittest.cc
```diff
@@ -157,4 +177,101 @@ TEST(SGSparseMatrix, io_libsvm)
...
+    //check contents
+    for (index_t featIndex=0; featIndex<numberOfFeatures; ++featIndex)
+        for (index_t vecIndex=0; vecIndex<numberOfVectors; ++vecIndex)
+            EXPECT_EQ(sparseMatrix(vecIndex,featIndex), sparseMatrixT(featIndex,vecIndex));
+}
+
+TEST(SGSparseMatrix, from_dense)
+{
+    //GenerateSparseMatrix cannot templated to generate dense matrix,
+    //since () operator are defined differently for sparse and dense matrix
```
iglesias: Can you motivate a bit more? Fine that dense matrix uses (row,col) (or the other way around, do not remember now) and sparse the contrary (as you mentioned in the mailing list). But dense matrix is not obliged to have features in rows or columns, is just a generic matrix.

iglesias:
Merging for the moment since everything is fine. But it would be nice if you answer the comments above and address them if you like in new pull request.

tests/unit/lib/SGSparseMatrix_unittest.cc
yingryic: just make it clear it is unchanged in the function. Pass in reference is better, but leave in this way wont cost much more, it is just ints and floats.

tests/unit/lib/SGSparseMatrix_unittest.cc
yingryic: I have thought about for a while, I think as long as the sparse matrix and dense matrix use different interfaces, it will be very hard to make the code more elegant.
1. we can not use the same (templated) function to generate both sparse and dense matrix
2. we can write a function always generate two matrices (one sparse one dense), but it is not the best solution since some test only need one type of matrix.
I guess I'll leave it here, if you want to make the two interfaces compatible, I could refine the test in the future.

iglesias: Why cannot you make 1.?

yingryic: got an idea

PR #2756, opened 2015-03-03
minor fix

yorkerlin:
@karlnapf 
take a look at this

### DIFF
--- a/src/shogun/machine/gp/KLApproxDiagonalInferenceMethod.cpp
+++ b/src/shogun/machine/gp/KLApproxDiagonalInferenceMethod.cpp
@@ -74,7 +74,8 @@ void CKLApproxDiagonalInferenceMethod::init()
 CKLApproxDiagonalInferenceMethod* CKLApproxDiagonalInferenceMethod::obtain_from_generic(
 		CInferenceMethod* inference)
 {
-	REQUIRE(inference, "Inference pointer not set.\n");
+	if (inference==NULL)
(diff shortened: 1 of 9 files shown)


yorkerlin:
@karlnapf 
There is an issue about the initial value of `float` variables in shogun.
 `SerializationJSON` dose not support `CMath::INFITY` and many of values which are below `10^{-6}`. Setting unsupported initial values will lead to the failure of  the unit test of  `SerializationJSON`.
It implies that  initial values needed in an algorithm have to set `dummy` or `nonsense` value in `init()`.  Without setting initial values, `Coverity Scan` will complain.

The trick to address this issue is that to replace `proper` initial values with `dummy` before computation is performed.

src/shogun/machine/gp/KLApproxDiagonalInferenceMethod.cpp
```diff
@@ -74,7 +74,8 @@ void CKLApproxDiagonalInferenceMethod::init()
 CKLApproxDiagonalInferenceMethod* CKLApproxDiagonalInferenceMethod::obtain_from_generic(
 		CInferenceMethod* inference)
 {
-	REQUIRE(inference, "Inference pointer not set.\n");
+	if (inference==NULL)
```
karlnapf: thanks for that.

BTW there were some plans to provide obtina_from_generic for all Shogun classes automagically. Interested in doing this at some point? :)

PR #2759, opened 2015-03-04
modified sparse matrix unit test, made it cleaner

yingryic:
create a template to generate both sparse and dense matrix and made the rest of the test more concise. @iglesias

### DIFF
--- a/tests/unit/lib/SGSparseMatrix_unittest.cc
+++ b/tests/unit/lib/SGSparseMatrix_unittest.cc
@@ -26,7 +26,7 @@ using namespace Eigen;
 
 //function can generate both sparse and dense matrix accroding to sparse convention
 template<class MatrixType>
-void GenerateMatrixAccrodingToSparseConvention(const float64_t &sparseLevel,
+void GenerateMatrixAccordingToSparseConvention(const float64_t &sparseLevel,
                                                const int32_t   &numberOfFeatures,
                                                const int32_t   &numberOfVectors,
                                                const int32_t   &randSeed,
@@ -193,14 +193,15 @@ TEST(SGSparseMatrix, access_by_index_non_square)
 
     //generate a sparse and dense matrix and compare this entries
     SGSparseMatrix<float64_t> sparseMatrix(numberOfFeatures, numberOfVectors);
-    GenerateMatrixAccrodingToSparseConvention<SGSparseMatrix<float64_t> >(sparseLevel,
+    GenerateMatrixAccordingToSparseConvention<SGSparseMatrix<float64_t> >(sparseLevel,
                                                                           numberOfFeatures,
                                                                           numberOfVectors,
                                                                           randSeed,
                                                                           &sparseMatrix);
 
-    SGMatrix<float64_t> denseMatrix(numberOfVectors, numberOfFeatures); denseMatrix.zero();
-    GenerateMatrixAccrodingToSparseConvention<SGMatrix<float64_t> >(sparseLevel,
+    SGMatrix<float64_t> denseMatrix(numberOfVectors, numberOfFeatures);
+    denseMatrix.zero();
+    GenerateMatrixAccordingToSparseConvention<SGMatrix<float64_t> >(sparseLevel,
                                                                     numberOfFeatures,
                                                                     numberOfVectors,
                                                                     randSeed,
@@ -219,7 +220,7 @@ TEST(SGSparseMatrix, get_transposed_more_features)
     const index_t randSeed=0;
 
     SGSparseMatrix<float64_t> sparseMatrix(numberOfFeatures, numberOfVectors);
-    GenerateMatrixAccrodingToSparseConvention<SGSparseMatrix<float64_t> >(sparseLevel,
+    GenerateMatrixAccordingToSparseConvention<SGSparseMatrix<float64_t> >(sparseLevel,
                                                                           numberOfFeatures,
                                                                           numberOfVectors,
                                                                           randSeed,
@@ -245,7 +246,7 @@ TEST(SGSparseMatrix, get_transposed_more_vectors)
     const index_t randSeed=0;
 
     SGSparseMatrix<float64_t> sparseMatrix(numberOfFeatures, numberOfVectors);
-    GenerateMatrixAccrodingToSparseConvention<SGSparseMatrix<float64_t> >(sparseLevel,
+    GenerateMatrixAccordingToSparseConvention<SGSparseMatrix<float64_t> >(sparseLevel,
                                                                           numberOfFeatures,
                                                                           numberOfVectors,
                                                                           randSeed,
@@ -271,8 +272,9 @@ TEST(SGSparseMatrix, from_dense)
     const index_t numberOfVectors=100;
     const index_t randSeed=0;
 
-    SGMatrix<float64_t> denseMatrix(numberOfVectors, numberOfFeatures); denseMatrix.zero();
-    GenerateMatrixAccrodingToSparseConvention<SGMatrix<float64_t> >(sparseLevel,
+    SGMatrix<float64_t> denseMatrix(numberOfVectors, numberOfFeatures);
+    denseMatrix.zero();
+    GenerateMatrixAccordingToSparseConvention<SGMatrix<float64_t> >(sparseLevel,
                                                                     numberOfFeatures,
                                                                     numberOfVectors,
                                                                     randSeed,
--- a/tests/unit/lib/SGSparseMatrix_unittest.cc
+++ b/tests/unit/lib/SGSparseMatrix_unittest.cc
@@ -24,22 +24,22 @@ using namespace shogun;
 using namespace Eigen;
 #endif // HAVE_EIGEN3
 
-//function to create sparse matrix
-template<typename T>
-SGSparseMatrix<T> GenerateSparseMatrix(const int32_t numberOfFeatures,
-                                       const int32_t numberOfVectors,
-                                       const float64_t sparseLevel)
+//function can generate both sparse and dense matrix accroding to sparse convention
+template<class MatrixType>
+void GenerateMatrixAccrodingToSparseConvention(const float64_t &sparseLevel,
+                                               const int32_t   &numberOfFeatures,
+                                               const int32_t   &numberOfVectors,
+                                               const int32_t   &randSeed,
+                                               MatrixType* matrix)
 {
-    CRandom randGenerator(0);
-    SGSparseMatrix<T> matrix(numberOfFeatures, numberOfVectors);
+    CRandom randGenerator(randSeed);
     for (index_t featIndex=0; featIndex<numberOfFeatures; ++featIndex)
         for (index_t vecIndex=0; vecIndex<numberOfVectors; ++vecIndex)
         {
             float64_t randomNumber=randGenerator.random(0.0,1.0);
             if (randomNumber<=sparseLevel)
-                matrix(vecIndex,featIndex)=randomNumber;
+                (*matrix)(vecIndex,featIndex)=randomNumber*100;
         }
-    return matrix;
 }
 
 #ifdef HAVE_EIGEN3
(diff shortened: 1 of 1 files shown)


tests/unit/lib/SGSparseMatrix_unittest.cc
```diff
@@ -23,22 +23,22 @@ using namespace shogun;
...
 using namespace Eigen;
 
-//function to create sparse matrix
-template<typename T>
-SGSparseMatrix<T> GenerateSparseMatrix(const int32_t numberOfFeatures,
-                                       const int32_t numberOfVectors,
-                                       const float64_t sparseLevel)
+//function can generate both sparse and dense matrix accroding to sparse convention
+template<class MatrixType>
+void GenerateMatrixAccrodingToSparseConvention(const float64_t &sparseLevel,
```
iglesias: I think you mean according, typo.

iglesias:
I am at the phone with crappy connection now. Will continue reviewing later.

tests/unit/lib/SGSparseMatrix_unittest.cc
```diff
@@ -23,22 +23,22 @@ using namespace shogun;
...
+                                               const int32_t   &numberOfFeatures,
+                                               const int32_t   &numberOfVectors,
+                                               const int32_t   &randSeed,
```
iglesias: Why are all these const and passed by reference? I mean, the pass by reference is correct, but there is no gain here since the arguments are just primitive and very simple data types.

tests/unit/lib/SGSparseMatrix_unittest.cc
```diff
@@ -23,22 +23,22 @@ using namespace shogun;
...
+                                               MatrixType* matrix)
```
iglesias: Why not creating a `MatrixType` variable in the method that is then returned? In my opinion, it is more intuitive and readable that way.

tests/unit/lib/SGSparseMatrix_unittest.cc
```diff
@@ -184,37 +184,44 @@ TEST(SGSparseMatrix, io_libsvm)
...
+
+    //generate a sparse and dense matrix and compare this entries
+    SGSparseMatrix<float64_t> sparseMatrix(numberOfFeatures, numberOfVectors);
+    GenerateMatrixAccrodingToSparseConvention<SGSparseMatrix<float64_t> >(sparseLevel,
+                                                                          numberOfFeatures,
+                                                                          numberOfVectors,
+                                                                          randSeed,
+                                                                          &sparseMatrix);
+
+    SGMatrix<float64_t> denseMatrix(numberOfVectors, numberOfFeatures); denseMatrix.zero();
```
iglesias: Keep one instruction per line, please.

tests/unit/lib/SGSparseMatrix_unittest.cc
```diff
@@ -184,37 +184,44 @@ TEST(SGSparseMatrix, io_libsvm)
...
-
+    const index_t randSeed=0;
+
+    //generate a sparse and dense matrix and compare this entries
+    SGSparseMatrix<float64_t> sparseMatrix(numberOfFeatures, numberOfVectors);
+    GenerateMatrixAccrodingToSparseConvention<SGSparseMatrix<float64_t> >(sparseLevel,
+                                                                          numberOfFeatures,
+                                                                          numberOfVectors,
+                                                                          randSeed,
+                                                                          &sparseMatrix);
```
iglesias: I am not really a fan of this sort of indentation, but if you like it I guess it's fine.

tests/unit/lib/SGSparseMatrix_unittest.cc
```diff
@@ -252,26 +263,28 @@ TEST(SGSparseMatrix, get_transposed_more_vectors)
...
-                denseMatrix(featIndex,vecIndex)=randomNumber;
-        }
+    const index_t randSeed=0;
+
+    SGMatrix<float64_t> denseMatrix(numberOfVectors, numberOfFeatures); denseMatrix.zero();
+    GenerateMatrixAccrodingToSparseConvention<SGMatrix<float64_t> >(sparseLevel,
+                                                                    numberOfFeatures,
+                                                                    numberOfVectors,
+                                                                    randSeed,
+                                                                    &denseMatrix);
```
iglesias: Why is GitHub marking this in another colour? Something different with the indentation (using spaces instead of tabs) maybe?

lambday:
@yingryic hey could you please rebase and push again? There was a compilation error in absence of Eigen3 (your newly added method to generate test sparse matrix was under that ifdef guard which was being used outside of that guard in the other tests). small issue - #2760 fixed it.

thanks for the tests btw :) really useful :)

tests/unit/lib/SGSparseMatrix_unittest.cc
yingryic: I use const & to denote the variable is not changed inside the function, * to denoting something is changed. If you prefer not to have the reference here, I'm fine with deleting that.

tests/unit/lib/SGSparseMatrix_unittest.cc
yingryic: I don't think it is possible to create a sparse and dense matrix and fill in their value inside the function. since you need to use SparseMatrix(numberOfFeats, numberOfVects) but DenseMatrix(numberOfVecs, numberOfFeats) to initialize. Thus it is only possible to initialize the matrix outside the function and fill in their value inside the function.

tests/unit/lib/SGSparseMatrix_unittest.cc
yingryic: I use spaces, not sure why this happen.

lambday:
Hi @yingryic. If you're using a RNG in unit-tests, I think it would be better that you use a fixed seed. I think the following failure is due to that
```
[ RUN      ] SGSparseMatrix.from_dense
/usr/home/wiking/shogun-buildbot/bsd1_-_libshogun/build/tests/unit/lib/SGSparseMatrix_unittest.cc:278: Failure
Value of: denseMatrix(featIndex,vecIndex)
  Actual: nan
Expected: sparseMatrix(vecIndex,featIndex)
Which is: nan
/usr/home/wiking/shogun-buildbot/bsd1_-_libshogun/build/tests/unit/lib/SGSparseMatrix_unittest.cc:278: Failure
Value of: denseMatrix(featIndex,vecIndex)
  Actual: nan
Expected: sparseMatrix(vecIndex,featIndex)
Which is: nan
[  FAILED  ] SGSparseMatrix.from_dense (1 ms)
```
See [buildbot log](http://buildbot.shogun-toolbox.org/builders/bsd1%20-%20libshogun/builds/2666/steps/test/logs/stdio). It works fine on linux but fails on bsd.

lambday:
Ah sorry I didn't notice you're already using a fixed seed. Maybe something else then causing a nan value then?

tests/unit/lib/SGSparseMatrix_unittest.cc
iglesias: This is what I mean:

```
template<class MatrixType>
MatrixType GenerateMatrixAccordingToSparseConvention(float64_t sparseLevel, int32_t m, int32_t n, int32_t randSeed)
{
CRandom randGenerator(randSeed);
MatrixType matrix(m,n);
for (index_t i=0; i<m; ++i)
for (index_t j=0; j<n; ++j)
{
float64_t randomNumber=randGenerator.random(0.0,1.0);
if (randomNumber<=sparseLevel)
matrix(i,j)=randomNumber*100;
}

return matrix;
}
```

iglesias: The idea is that when you call it for a dense matrix then m has the meaning of #features and n of #vectors (or vice versa, maybe) and when you call it for a sparse one the other way around. It does not matter, they are just generic types for matrices, not actual feature objects!

I haven't compile the code above, just write it here directly, but it makes sense to me and hopefully illustrates the idea. Use at your own risk ;-)

yingryic: @iglesias  yes, I know what you mean, and I also think that is what we should do. 

But the problem is if we initialize matrix as "MatrixType matrix(m,n)" and later use matrix(i,j) to access its contents. For sparse matrix, i goes from 0 to n-1, and j is from 0 to m-1, while dense matrix is i: 0 to m-1, j: 0 to n-1. 

This is how sparse and dense matrix is currently implemented and what I mean the sparse and dense matrix use different interfaces.

yingryic:
@lambday I use unix, then how can I test it against bsd.

tests/unit/lib/SGSparseMatrix_unittest.cc
```diff
@@ -24,22 +24,22 @@ using namespace shogun;
...
 #endif // HAVE_EIGEN3
 
-//function to create sparse matrix
-template<typename T>
-SGSparseMatrix<T> GenerateSparseMatrix(const int32_t numberOfFeatures,
-                                       const int32_t numberOfVectors,
-                                       const float64_t sparseLevel)
+//function can generate both sparse and dense matrix accroding to sparse convention
+template<class MatrixType>
+void GenerateMatrixAccordingToSparseConvention(const float64_t &sparseLevel,
```
yingryic: fixed typo

tests/unit/lib/SGSparseMatrix_unittest.cc
```diff
@@ -24,22 +24,22 @@ using namespace shogun;
...
     for (index_t featIndex=0; featIndex<numberOfFeatures; ++featIndex)
         for (index_t vecIndex=0; vecIndex<numberOfVectors; ++vecIndex)
         {
             float64_t randomNumber=randGenerator.random(0.0,1.0);
             if (randomNumber<=sparseLevel)
-                matrix(vecIndex,featIndex)=randomNumber;
+                (*matrix)(vecIndex,featIndex)=randomNumber*100;
         }
-    return matrix;
 }
```
yingryic: In case you didnt notice my previous conversation @iglesias 

The problem is if we initialize matrix as "MatrixType matrix(m,n)" and later use matrix(i,j) to access its contents. For sparse matrix, i goes from 0 to n-1, and j is from 0 to m-1, while dense matrix is i: 0 to m-1, j: 0 to n-1.

This is how sparse and dense matrix is currently implemented and what I mean the sparse and dense matrix use different interfaces.

tests/unit/lib/SGSparseMatrix_unittest.cc
```diff
@@ -254,26 +266,29 @@ TEST(SGSparseMatrix, get_transposed_more_vectors)
...
-        for (index_t vecIndex=0; vecIndex<numberOfVectors; ++vecIndex)
-        {
-            float64_t randomNumber=randGenerator.random(0.0,1.0);
-            if (randomNumber<=sparseLevel)
-                denseMatrix(featIndex,vecIndex)=randomNumber;
-        }
+    const index_t randSeed=0;
+
+    SGMatrix<float64_t> denseMatrix(numberOfVectors, numberOfFeatures);
+    denseMatrix.zero();
```
yingryic: changed to one line per instruction

tests/unit/lib/SGSparseMatrix_unittest.cc
```diff
@@ -186,37 +186,45 @@ TEST(SGSparseMatrix, io_libsvm)
...
+    //generate a sparse and dense matrix and compare this entries
+    SGSparseMatrix<float64_t> sparseMatrix(numberOfFeatures, numberOfVectors);
+    GenerateMatrixAccordingToSparseConvention<SGSparseMatrix<float64_t> >(sparseLevel,
+                                                                          numberOfFeatures,
+                                                                          numberOfVectors,
+                                                                          randSeed,
+                                                                          &sparseMatrix);
+
+    SGMatrix<float64_t> denseMatrix(numberOfVectors, numberOfFeatures);
+    denseMatrix.zero();
```
yingryic: changed to one line per instruction

yingryic:
committed again, please tell me what you think @iglesias

tests/unit/lib/SGSparseMatrix_unittest.cc
iglesias: No. The for loops in my comment  made ¡ go from 0 to n-1 and j from 0 to m-1 (or vice versa) in any case.

PR #2761, opened 2015-03-06
get rid of extraneous template args in linalg impl

lambday:


lambday:
Travis fails unrelated. Merging.

PR #2763, opened 2015-03-07
Update NEWS

iglesias:


NEWS
```diff
@@ -48,6 +48,12 @@
 		- Removing redundant and fixing implicit imports [Thoralf Klein]
 		- Hide many methods from SWIG, reducing compile memory by 500MiB [Heiko Strathmann, Fernando Iglesias, Thoralf Klein]
 
+2014-12-16 Fernando Iglesias <fernando.iglesias@shogun-toolbox.org>
```
karlnapf: Could you also add the "new" news section? This is the one of the old release...

karlnapf: Wait I am confused. Why did you add this here? Looking at the file doesnt seem it has the correct structure now

iglesias: That seems to be true. I had this old commit in a local branch and pushed it overlooking the new structure.

iglesias: What is the "new" news section?

iglesias: I will move this into Cleanups in the section above. I am wondering though, how do keep track of news constantly if we put everything in a section after a release?

PR #2764, opened 2015-03-07
Refactoring in SGSparseMatrix unit test.

iglesias:


iglesias:
@yingryic, have a look. I also replaced the indentation with tabs instead of white spaces. In new files, white spaces are fine to indent but in an already existing file, please use the same characters that were already used. Configure your editor to make white space characters visible.

PR #2767, opened 2015-03-11
add the FITC framework2

yorkerlin:
@karlnapf 
take a look at this

### DIFF
--- a/src/shogun/machine/gp/SingleFITCLaplacianBase.cpp
+++ b/src/shogun/machine/gp/SingleFITCLaplacianBase.cpp
@@ -1,0 +1,401 @@
+/*
+ * Copyright (c) The Shogun Machine Learning Toolbox
+ * Written (W) 2015 Wu Lin
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * The views and conclusions contained in the software and documentation are those
+ * of the authors and should not be interpreted as representing official policies,
+ * either expressed or implied, of the Shogun Development Team.
+ *
+ */
+
+#include <shogun/machine/gp/SingleFITCLaplacianBase.h>
+
+#ifdef HAVE_EIGEN3
+
+#include <shogun/mathematics/Math.h>
+#include <shogun/mathematics/eigen3.h>
+#include <shogun/features/DotFeatures.h>
+
+using namespace shogun;
+using namespace Eigen;
+
+CSingleFITCLaplacianBase::CSingleFITCLaplacianBase() : CFITCInferenceBase()
+{
+	init();
+}
+
+CSingleFITCLaplacianBase::CSingleFITCLaplacianBase(CKernel* kern, CFeatures* feat,
+		CMeanFunction* m, CLabels* lab, CLikelihoodModel* mod, CFeatures* lat)
+		: CFITCInferenceBase(kern, feat, m, lab, mod, lat)
+{
+	init();
+	check_fully_FITC();
+}
+
+void CSingleFITCLaplacianBase::init()
+{
+	m_fully_FITC=false;
+	m_lock=new CLock();
+	SG_ADD(&m_al, "al", "alpha", MS_NOT_AVAILABLE);
+	SG_ADD(&m_t, "t", "noise", MS_NOT_AVAILABLE);
+	SG_ADD(&m_B, "B", "B", MS_NOT_AVAILABLE);
+	SG_ADD(&m_w, "w", "B*al", MS_NOT_AVAILABLE);
+	SG_ADD(&m_Rvdd, "Rvdd", "Rvdd", MS_NOT_AVAILABLE);
+	SG_ADD(&m_V, "V", "V", MS_NOT_AVAILABLE);
+	SG_ADD(&m_fully_FITC, "fully_FITC", "whether the kernel support fitc inference", MS_NOT_AVAILABLE);
+}
+
+void CSingleFITCLaplacianBase::set_kernel(CKernel* kern)
+{
+	CInferenceMethod::set_kernel(kern);
+	check_fully_FITC();
+}
+
+CSingleFITCLaplacianBase::~CSingleFITCLaplacianBase()
+{
+	delete m_lock;
+}
+
+void CSingleFITCLaplacianBase::check_fully_FITC()
+{
+	REQUIRE(m_kernel, "Kernel must be set first\n")
+	if (strstr(m_kernel->get_name(), "FITCKernel")!=NULL)
+		m_fully_FITC=true;
+	else
+	{
+		SG_WARNING( "The provided kernel does not support fully FITC inference\n");
+		m_fully_FITC=false;
+	}
+}
+
+SGVector<float64_t> CSingleFITCLaplacianBase::get_derivative_related_cov_diagonal()
+{
+	Map<MatrixXd> eigen_W(m_Rvdd.matrix, m_Rvdd.num_rows, m_Rvdd.num_cols);
+	Map<VectorXd> eigen_al(m_al.vector, m_al.vlen);
+
+	SGVector<float64_t> res(m_al.vlen);
+	Map<VectorXd> eigen_res(res.vector, res.vlen);
+	//-sum(W.*W,1)' - al.*al;
+	eigen_res=-eigen_W.cwiseProduct(eigen_W).colwise().sum().transpose()-eigen_al.array().pow(2).matrix();
+	return res;
+}
+
+float64_t CSingleFITCLaplacianBase::get_derivative_related_cov_helper(
+	SGMatrix<float64_t> dKuui, SGVector<float64_t> v, SGMatrix<float64_t> R)
+{
+	Map<VectorXd> eigen_w(m_w.vector, m_w.vlen);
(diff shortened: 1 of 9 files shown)


src/shogun/machine/gp/SingleFITCLaplacianBase.h
```diff
@@ -0,0 +1,269 @@
...
+	virtual void check_fully_FITC();
+
+	/** compute matrices which are required to compute negative log marginal
+	 * likelihood full derivatives wrt  cov-like hyperparameter
+	 *
+	 * Note that
+	 * scale, which is a hyperparameter in inference_method, is a cov-like hyperparameter
+	 * hyperparameters in cov function are cov-like hyperparameters
+	 *
+	 * @param the gradient wrt hyperparameter related to cov-like hyperparameter
```
karlnapf: Could you document all parameters here?
doxygen will complain otherwise?
Maybe reference the math somewhere?

src/shogun/machine/gp/SingleFITCLaplacianBase.h
```diff
@@ -0,0 +1,269 @@
...
+		SGMatrix<float64_t> dKuui, SGMatrix<float64_t> dKui);
+
+	/** compute matrices which are required to compute negative log marginal
+	 * likelihood derivatives wrt cov-like hyperparameter
+	 *
+	 * Note that
+	 * scale, which is a hyperparameter in inference_method, is a cov-like hyperparameter
+	 * hyperparameters in cov function are cov-like hyperparameters
+	 *
+	 * @param the gradient wrt hyperparameter related to cov-like hyperparameter
```
karlnapf: same here

src/shogun/machine/gp/SingleFITCLaplacianBase.h
```diff
@@ -0,0 +1,269 @@
...
+
+	/** helper function to compute matrices which are required to compute negative log marginal
+	 * likelihood derivatives wrt cov-like hyperparameter
+	 *
+	 * Note that
+	 * scale, which is a hyperparameter in inference_method, is a cov-like hyperparameter
+	 * hyperparameters in cov function are cov-like hyperparameters
+	 * what is more, derivative wrt inducing_noise will also use this function
+	 *
+	 * @param the explicit gradient wrt hyperparameter related to cov
```
karlnapf: and here

src/shogun/machine/gp/SingleFITCLaplacianBase.h
```diff
@@ -0,0 +1,269 @@
...
+	 * @return derivative of negative log marginal likelihood
+	 */
+	virtual SGVector<float64_t> get_derivative_related_cov_diagonal();
+
+	/** helper function to compute matrices which are required to compute negative log marginal
+	 * likelihood derivatives wrt mean
+	 *
+	 * @param the explicit gradient wrt hyperparameter related to mean
```
karlnapf: ```@param dmu the explicit gradient .... ```

src/shogun/machine/gp/SingleFITCLaplacianBase.h
```diff
@@ -0,0 +1,269 @@
...
+
+	/** returns derivative of negative log marginal likelihood wrt inducing features (input)
+	 * Note that in order to call this method, kernel must support FITC inference,
+	 * which means derivatives wrt inducing features can be computed
+	 *
+	 * @return derivative of negative log marginal likelihood
+	 */
+	virtual SGVector<float64_t> get_derivative_wrt_inducing_features(const TParameter* param);
+
+	/* Note that alpha is NOT post.alpha */
```
karlnapf: With all those vectors around, it might get a bit confusing what is what.

I suggest you write down the math of the thing somewhere, with well defined expressions, and then you can reference this. I.e. "this method computes the $L_R$ matrix of expression XYZ"

What do you think?

src/shogun/machine/gp/SingleFITCLaplacianBase.h
```diff
@@ -0,0 +1,269 @@
...
+	/* B*al */
+	SGVector<float64_t> m_w;
+
+	/** Rvdd=W is defined in infFITC.m 
+	* Note that W is NOT the diagonal matrix
+	*/
+	SGMatrix<float64_t> m_Rvdd;
+
+	/* a lock used to parallelly compute derivatives wrt hyperparameters */
+	CLock m_lock;
```
karlnapf: No pointer here?

karlnapf: C* classes should never be on stack

src/shogun/machine/gp/SingleFITCLaplacianBase.cpp
```diff
@@ -0,0 +1,409 @@
...
+
+CSingleFITCLaplacianBase::~CSingleFITCLaplacianBase()
+{
+}
+
+void CSingleFITCLaplacianBase::check_fully_FITC()
+{
+	REQUIRE(m_kernel, "Kernel must be set first\n")
+	if (strstr(m_kernel->get_name(), "FITCKernel")!=NULL)
+	{
```
karlnapf: Minor style: Shogun doesnt do { ... } if there is only one expression inside

src/shogun/machine/gp/SingleFITCLaplacianBase.cpp
```diff
@@ -0,0 +1,409 @@
...
+	result+=(eigen_ddiagKi.dot(eigen_t))/2.0-
+			eigen_w.dot((eigen_dKui*eigen_al));
+	return result;
+
+}
+
+SGVector<float64_t> CSingleFITCLaplacianBase::get_derivative_wrt_inference_method(
+		const TParameter* param)
+{
+	REQUIRE(!(strcmp(param->m_name, "scale")
```
karlnapf: Put a REQUIRE(param, "param not set\n") here to avoid segfaults in the next check

src/shogun/machine/gp/SingleFITCLaplacianBase.cpp
```diff
@@ -0,0 +1,409 @@
...
+	result[0]=get_derivative_related_cov(deriv_trtr, deriv_uu, deriv_tru);
+	return result;
+}
+
+SGVector<float64_t> CSingleFITCLaplacianBase::get_derivative_wrt_kernel(
+		const TParameter* param)
+{
+	SGVector<float64_t> result;
+
+	if (param->m_datatype.m_ctype==CT_VECTOR ||
```
karlnapf: same here, a NULL check is good

src/shogun/machine/gp/SingleFITCLaplacianBase.cpp
```diff
@@ -0,0 +1,409 @@
...
+	Map<VectorXd> eigen_al(m_al.vector, m_al.vlen);
+	Map<VectorXd> eigen_dmu(dmu.vector, dmu.vlen);
+	return -eigen_dmu.dot(eigen_al);
+}
+
+SGVector<float64_t> CSingleFITCLaplacianBase::get_derivative_wrt_mean(
+		const TParameter* param)
+{
+	// create eigen representation of al vector
+
```
karlnapf: param NULL check

src/shogun/machine/gp/SingleFITCLaplacianBase.cpp
```diff
@@ -0,0 +1,409 @@
...
+	Map<VectorXd> eigen_v(v.vector, v.vlen);
+	Map<MatrixXd> eigen_B(m_B.matrix, m_B.num_rows, m_B.num_cols);
+	Map<MatrixXd> eigen_BdK(BdK.matrix, BdK.num_rows, BdK.num_cols);
+
+	int32_t dim=((CDotFeatures *)m_inducing_features)->get_dim_feature_space();
+	int32_t num_samples=m_inducing_features->get_num_vectors();
+	SGVector<float64_t>deriv_lat(dim*num_samples);
+	deriv_lat.zero();
+
+	m_lock.lock();
```
karlnapf: Could you explain me this lock business here?

karlnapf: Need to be careful about deadlocks when using this.
Where is the parallelisation happening?

karlnapf:
You are writing beautiful code here, @yorkerlin 

There are a few minor style things and some missing REQUIRE checks, but good to go otherwise.

The lock thing I don't understand, could you explain it to me?

Good to merge otherwise

karlnapf:
Most of the other travis errors seem unrelated

src/shogun/machine/gp/SingleFITCLaplacianBase.cpp
yorkerlin: The `CInferenceMethod` will compute the gradients wrt the hyper-paramters in parallel.

The existing `FITCInferenceMethod` only use evaluate  the kernel function when `CFITCInferenceMethod::get_derivative_wrt_kernel` is called.

However, if we want to compute derivative_wrt_inducing_features` and `derivative_wrt_inducing_noise` in parallel, we have to `protect` the `kernel function`.

yorkerlin: To avoid this `lock` and gain speed-up, we need three kernel objects via `copy`.  In this case, these three kernel objects will be initialized only once (\Sigma_n, \Sigma_m and \Sigma_{nm}).

yorkerlin:
For mathematical equations related to sparse Gaussian inference, I may include the final equations in a new IPython notebook.
A tutorial work is done by Yarin Gal etc at http://arxiv.org/abs/1402.1412.

src/shogun/machine/gp/SingleFITCLaplacianBase.cpp
karlnapf: Lets keep the lock to  be stable in terms of memory. I dont think the speed gain is too massive?
Thanks for explaining nicely

karlnapf:
Mmmmmh, maybe at least some of the math? Doesnt need all the derivations.
But just something so that if one has the paper lying next to him can read the code and see easily what is what?

karlnapf:
I did not know the paper yet. Its good to reference it!

karlnapf:
Ok, once the style and REQUIRE things are in, I will merge

yorkerlin:
@karlnapf 
I am writing the theory part for the new notebook. Only the `final` results of the derivations are include in the notebook.

Will update this PR tonight.

src/shogun/machine/gp/SingleFITCLaplacianBase.cpp
yorkerlin: For computing gradients wrt cov, induing points, and scale, we have to initialized the kernels  (\Sigma_n, \Sigma_m and \Sigma_{nm}), which means we have to compute the kernel matrix 3*n^2 + 3*m*n + 3*m^2, where n is the number of  input points and m is the number of inducing points.

yorkerlin: @karlnapf 
If you think the `copy` approach is more reasonable, I can  send a new PR for this.
I am not sure that whether the existing base class `CKernel` has a `copy()` to do the following operation. (Or all sub-classes have a `copy()` to do so)
```
CKernel * ptr=new GaussianARDFITCKernel();
Ckernel * new_ptr=obj.copy() //where new_ptr is the pointer to GaussianARDFITCKernel
```

src/shogun/machine/gp/SingleFITCLaplacianBase.h
yorkerlin: see updated PR

src/shogun/machine/gp/SingleFITCLaplacianBase.h
yorkerlin: see updated PR

src/shogun/machine/gp/SingleFITCLaplacianBase.cpp
yorkerlin: currently we check the name of the provided kernel to determine whether the kernel can compute the derivatives wrt inducing_features.
The reason is the base class `CKernel` does not have the `is_support_FITC()`.
The `wrapped kernel` approach is hard to implement since the static template methods in `CKernel`.

yorkerlin: Indeed, if a kernel(left_features, right_features) can compute the gradients wrt left_features, the kernel supports FITC  operation.
However, the existing `CKernel::get_parameter_gradient(const TParameter* param, index_t index=-1)` cannot handle this operation, since `left_features` are not hyper-parameters of the kernel

karlnapf:
I saw you sent a patch, is this the final one? Travis was ok except for some other errors

yorkerlin:
For the  "lock" approach, this PR should be final

On Monday, March 16, 2015, Heiko Strathmann <notifications@github.com>
wrote:

> I saw you sent a patch, is this the final one? Travis was ok except for
> some other errors
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/shogun-toolbox/shogun/pull/2767#issuecomment-81966384>
> .
>


-- 
best,

wu lin

karlnapf:
I think the lock is better than creating all these multiple matrices 
So merging!
Well done!

Issue #2769, opened 2015-03-12
make sparse matrix and dense matrix have the same interface


yingryic:
the current constructors for sparse and dense matrix are matrix(int num_feat, int num_vec), but () operator are implemented differently, denseMatrix(feat_index, vec_index) and sparseMatrix(vec_index, featIndex). Thus we need to change the interface of SGSparseMatrix to make it consistent, and test other related member functions. @karlnapf

iglesias:
I even think we should remove the notion of features and vectors from these matrix classes. They should be treated as math objects and use more abstract terms (row and column) than feature and vector. The latter appear when we use matrices to represent feature matrices.

iglesias:
And the same for `SGSparseVector`. From my point view, there should be no feature mention there.

yingryic:
the variables "num_features" and "num_vectors" are defined as public in the sparsematrix class, as well as "num_feat_entries" and "*features" in the sparsevector class. Every file using those sparse structures are taking advantage of it, i.e, the statement "numFeatures=sparseMatrix.num_features" are everywhere, changing the convention would be problematic.
Just check all the files including the SGSparseMatrix.h file, seems all of them are taking advantage of of the sparse matrix data structure rather than calling the (,) operator. All the test using the (,) are just initialize a square matrix and set the diagonal entries. That's why nothing is broken.
Tell me if I need any change on the pull request. @iglesias

iglesias:
Oh, I see. It is understood now. Thanks a lot for the explanation. This makes me wonder whether we should provide after the operator() :D

yingryic:
yes, but some of the unit test are still using () operator, getting rid of it would also pose problems.

karlnapf:
I agree with @iglesias 
No features and vectors in a matrix.

As for semantics. Sparse matrices should really only be treated as linear operators that can be applied to vectors or chained. As a data structure, its more a dictionary of (i,j) pairs to a real number.

The () operator is useful for debuggin/construction, but it should never be used in practice. Imagine you iterate over num_rows, num_columns! Harmful!

It probably makes sense to mimic other libraries implenentation: Maybe check eigen3, scipy, and matlab?

iglesias:
They do provide that operator in other implementations (at least, I am somewhat confident Matlab does).

PR #2770, opened 2015-03-12
modified () operator for sparse matrix to be consistent with dense matrix

yingryic:
not sure anywhere else still using the sparse matrix () operator, but get all the unit test passed. tell me what you think. @iglesias @karlnapf #2769

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -47,12 +47,8 @@
 		- PCA now depends on Eigen3 instead of LAPACK [Parijat Mazumdar]
 		- Removing redundant and fixing implicit imports [Thoralf Klein]
 		- Hide many methods from SWIG, reducing compile memory by 500MiB [Heiko Strathmann, Fernando Iglesias, Thoralf Klein]
-
-2014-12-16 Fernando Iglesias <fernando.iglesias@shogun-toolbox.org>
-
-	* Cleanups:
 		- Drop empty generalised HMM [Fernando Iglesias]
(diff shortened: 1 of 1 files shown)


iglesias:
I am really surprised to see that nothing breaks after this change! Does this mean that `SGSparseMatrix` is not really used anywhere? I would expect that all around where `SGSparseMatrix::operator()` is used, the features and vector indices should be swapped. Am I missing something?

karlnapf:
Careful!

This is the final patch, see list here: http://permalink.gmane.org/gmane.comp.ai.machine-learning.shogun/5463
We cannot merge this without being absolutely certain that it doesnt break things.

You  basically have to search for all places where SGVSparseMatrix is imported and then check for the () operatior being invoked. A bit messy.

karlnapf:
However, after thats done, we can merge. Thanks a lot for your help

iglesias:
I don't think there is any risk as the operator() is used nowhere but the unit test.

iglesias:
In fact, I really wonder whether we need it if it is not used.

karlnapf:
It kind of is a dangerous thing for sparse structures as one should always iterate through the list of non-zero elements

yingryic:
@karlnapf I checked before, actually the () operator is not used anywhere except for one unit test (ProbingSample test), where it constructed a symmetric matrix, thus the order actually doesn't matter.

karlnapf:
Sorry for the late reply.
I agree if it is not used then the order doesnt matter.
We should probably unify the interface with other langauges as @iglesias suggested.

PR #2774, opened 2015-03-19
minor fix and comments about time complexity of fitc regression

yorkerlin:
@karlnapf 
take a look at this.
There are some issues about fitc inference.
In order to efficiently implement fitc inference, I need to a method to compute diagonal elements of a kernel matrix and related gradients.
I can implement the method for GaussianARD kernel.

Since we have to support general kernels, I may need to modify/add a `init` method in `CKernel`.

### DIFF
--- a/src/shogun/machine/gp/EPInferenceMethod.cpp
+++ b/src/shogun/machine/gp/EPInferenceMethod.cpp
@@ -465,21 +465,10 @@ SGVector<float64_t> CEPInferenceMethod::get_derivative_wrt_kernel(
 	// create eigen representation of the matrix Q
 	Map<MatrixXd> eigen_F(m_F.matrix, m_F.num_rows, m_F.num_cols);
 
+	REQUIRE(param, "Param not set\n");
 	SGVector<float64_t> result;
-
-	if (param->m_datatype.m_ctype==CT_VECTOR ||
-			param->m_datatype.m_ctype==CT_SGVECTOR ||
-			param->m_datatype.m_ctype==CT_MATRIX ||
-			param->m_datatype.m_ctype==CT_SGMATRIX)
-	{
-		REQUIRE(param->m_datatype.m_length_y,
-				"Length of the parameter %s should not be NULL\n", param->m_name)
-		result=SGVector<float64_t>(*(param->m_datatype.m_length_y));
-	}
-	else
-	{
-		result=SGVector<float64_t>(1);
-	}
+	int64_t len=const_cast<TParameter *>(param)->m_datatype.get_num_elements();
+	result=SGVector<float64_t>(len);
 
 	for (index_t i=0; i<result.vlen; i++)
 	{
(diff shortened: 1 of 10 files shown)


yorkerlin:
@karlnapf 
Please fix the bug in version.cpp and restart travis since there is a unrelated error in version.cpp

iglesias:
Hey @yorkerlin. Hopefully, my last commit will fix the version problem. You may need to rebase.

yorkerlin:
@karlnapf  @iglesias 
why  TParameter *  param ->m_datatype.`get_num_elements()` is `not` a `const` method ?

iglesias:
It makes sense to me to refactor it so that it is const.

src/shogun/machine/gp/EPInferenceMethod.cpp
```diff
@@ -465,21 +465,10 @@ SGVector<float64_t> CEPInferenceMethod::get_derivative_wrt_kernel(
 	// create eigen representation of the matrix Q
 	Map<MatrixXd> eigen_F(m_F.matrix, m_F.num_rows, m_F.num_cols);
 
+	REQUIRE(param, "Param not set\n");
 	SGVector<float64_t> result;
-
-	if (param->m_datatype.m_ctype==CT_VECTOR ||
```
karlnapf: so you remove those checks everywhere.
Is the idea that if param is set, then all those are redundant? Or are these things checked elsewhere anyways?

karlnapf:
I like these complexity comments a lot!
Could you put a few of those in the class description (not the details, just the current implementation's costs and that one could speed this up via tuning implementation)

karlnapf:
About diagonal of kernel matrix:

 * check whether we have this. I remember we considered adding this thing a while ago
 * if not, add a method to the base class, that does it naively
 * sub-classes may override, but as high as possible in the hierarchy

@lambday might have a comment here?

karlnapf:
Travis seems fine. Ill merge

src/shogun/machine/gp/EPInferenceMethod.cpp
yorkerlin: see here
```
int64_t TSGDataType::get_num_elements()
{
	switch (m_ctype)
	{
		case CT_SCALAR:
			return 1;
		case CT_VECTOR: case CT_SGVECTOR:
			/* length_y contains the length for vectors */
			return *m_length_y;
		case CT_MATRIX: case CT_SGMATRIX:
			return (*m_length_y)*(*m_length_x);
		case CT_NDARRAY:
			SG_SNOTIMPLEMENTED
		case CT_UNDEFINED: default:
			SG_SERROR("Implementation error: undefined container type\n");
			break;
	}
	return 0;
}
```
Since the method is done the job, I remove the redundant code.

karlnapf: ah great, very good style!

PR #2778, opened 2015-03-23
refactor FITC regression

yorkerlin:
@karlnapf 
take a look at this.
I will update the diagonal kernel stuff once PRs for FITC regression and FITC binary Laplace are merged.

### DIFF
--- a/src/interfaces/modular/GaussianProcess.i
+++ b/src/interfaces/modular/GaussianProcess.i
@@ -20,6 +20,8 @@
 %rename(MultiLaplacianInferenceMethod) CMultiLaplacianInferenceMethod;
 %rename(SingleLaplacianInferenceMethodWithLBFGS) CSingleLaplacianInferenceMethodWithLBFGS;
 %rename(FITCInferenceMethod) CFITCInferenceMethod;
+%rename(SingleFITCLaplacianBase) CSingleFITCLaplacianBase;
+%rename(FITCInferenceBase) CFITCInferenceBase;
 %rename(EPInferenceMethod) CEPInferenceMethod;
 
 %rename(LikelihoodModel) CLikelihoodModel;
@@ -85,6 +87,8 @@
 %include <shogun/machine/gp/ExactInferenceMethod.h>
 %include <shogun/machine/gp/SingleLaplacianInferenceMethodWithLBFGS.h>
 %include <shogun/machine/gp/FITCInferenceMethod.h>
+%include <shogun/machine/gp/SingleFITCLaplacianBase.h>
+%include <shogun/machine/gp/FITCInferenceBase.h>
 %include <shogun/machine/gp/EPInferenceMethod.h>
 
 %include <shogun/machine/gp/KLInferenceMethod.h>
--- a/src/shogun/machine/GaussianProcessMachine.cpp
+++ b/src/shogun/machine/GaussianProcessMachine.cpp
@@ -42,7 +42,8 @@
 #include <shogun/machine/GaussianProcessMachine.h>
 #include <shogun/mathematics/Math.h>
 #include <shogun/kernel/Kernel.h>
-#include <shogun/machine/gp/FITCInferenceMethod.h>
+#include <shogun/machine/gp/SingleFITCLaplacianBase.h>
+
 #include <shogun/mathematics/eigen3.h>
 
 using namespace shogun;
@@ -79,12 +80,14 @@ SGVector<float64_t> CGaussianProcessMachine::get_posterior_means(CFeatures* data
 	CFeatures* feat;
 
 	// use inducing features for FITC inference method
-	if (m_method->get_inference_type()==INF_FITC)
+	if (m_method->get_inference_type()==INF_FITC_REGRESSION ||
+		m_method->get_inference_type()==INF_FITC_LAPLACIAN_SINGLE)
 	{
-		CFITCInferenceMethod* fitc_method=
-			CFITCInferenceMethod::obtain_from_generic(m_method);
+		CSingleFITCLaplacianBase* fitc_method=
+			dynamic_cast<CSingleFITCLaplacianBase *>(m_method);
+		REQUIRE(fitc_method, "Inference method %s must support FITC inference\n",
+			m_method->get_name());
 		feat=fitc_method->get_inducing_features();
-		SG_UNREF(fitc_method);
 	}
 	else
 		feat=m_method->get_features();
@@ -139,12 +142,14 @@ SGVector<float64_t> CGaussianProcessMachine::get_posterior_variances(
 	CFeatures* feat;
 
 	// use inducing features for FITC inference method
-	if (m_method->get_inference_type()==INF_FITC)
+	if (m_method->get_inference_type()==INF_FITC_REGRESSION ||
+		m_method->get_inference_type()==INF_FITC_LAPLACIAN_SINGLE)
 	{
-		CFITCInferenceMethod* fitc_method=
-			CFITCInferenceMethod::obtain_from_generic(m_method);
+		CSingleFITCLaplacianBase* fitc_method=
+			dynamic_cast<CSingleFITCLaplacianBase *>(m_method);
+		REQUIRE(fitc_method, "Inference method %s must support FITC inference\n",
+			m_method->get_name());
 		feat=fitc_method->get_inducing_features();
-		SG_UNREF(fitc_method);
 	}
 	else
 		feat=m_method->get_features();
@@ -160,6 +165,8 @@ SGVector<float64_t> CGaussianProcessMachine::get_posterior_variances(
 
 	// get kernel matrix and create eigen representation of it
 	SGMatrix<float64_t> k_tsts=kernel->get_kernel_matrix();
+	//To DO in order to speed up the prediction process,
+	//we only need to obtain the diagonal elements of the kernel matrix
 	Map<MatrixXd> eigen_Kss(k_tsts.matrix, k_tsts.num_rows, k_tsts.num_cols);
 
 	// compute Kss=Kss*scale^2
--- a/src/shogun/machine/gp/FITCInferenceMethod.cpp
+++ b/src/shogun/machine/gp/FITCInferenceMethod.cpp
@@ -1,17 +1,34 @@
 /*
- * This program is free software; you can redistribute it and/or modify
- * it under the terms of the GNU General Public License as published by
- * the Free Software Foundation; either version 3 of the License, or
- * (at your option) any later version.
- *
+ * Copyright (c) The Shogun Machine Learning Toolbox
+ * Written (W) 2015 Wu Lin
  * Written (W) 2013 Roman Votyakov
- * Copyright (C) 2012 Jacob Walker
- * Copyright (C) 2013 Roman Votyakov
+ * All rights reserved.
  *
- * Code adapted from Gaussian Process Machine Learning Toolbox
- * http://www.gaussianprocess.org/gpml/code/matlab/doc/
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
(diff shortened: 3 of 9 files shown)


yorkerlin:
Completing the implementation of the first method at  https://github.com/shogun-toolbox/shogun/issues/2779

yorkerlin:
Travis is done but the flag is still yellow.
@karlnapf

src/interfaces/modular/GaussianProcess.i
```diff
@@ -20,6 +20,8 @@
 %rename(MultiLaplacianInferenceMethod) CMultiLaplacianInferenceMethod;
 %rename(SingleLaplacianInferenceMethodWithLBFGS) CSingleLaplacianInferenceMethodWithLBFGS;
 %rename(FITCInferenceMethod) CFITCInferenceMethod;
+%rename(SingleFITCLaplacianBase) CSingleFITCLaplacianBase;
+%rename(FITCInferenceBase) CFITCInferenceBase;
```
karlnapf: Do these base classes need to be exposed via swig?

src/shogun/machine/GaussianProcessMachine.cpp
```diff
@@ -79,12 +80,14 @@ SGVector<float64_t> CGaussianProcessMachine::get_posterior_means(CFeatures* data
...
 	// use inducing features for FITC inference method
-	if (m_method->get_inference_type()==INF_FITC)
+	if (m_method->get_inference_type()==INF_FITC_REGRESSION ||
+		m_method->get_inference_type()==INF_FITC_LAPLACIAN_SINGLE)
 	{
-		CFITCInferenceMethod* fitc_method=
-			CFITCInferenceMethod::obtain_from_generic(m_method);
+		CSingleFITCLaplacianBase* fitc_method=
+			dynamic_cast<CSingleFITCLaplacianBase *>(m_method);
+		REQUIRE(fitc_method, "Inference method %s must support FITC inference\n",
```
karlnapf: Slight change:
"Inference method %s does not support FITC inference\n"

More easily understandable for user

src/shogun/machine/GaussianProcessMachine.cpp
```diff
@@ -160,6 +165,8 @@ SGVector<float64_t> CGaussianProcessMachine::get_posterior_variances(
 
 	// get kernel matrix and create eigen representation of it
 	SGMatrix<float64_t> k_tsts=kernel->get_kernel_matrix();
+	//To DO in order to speed up the prediction process,
+	//we only need to obtain the diagonal elements of the kernel matrix
```
karlnapf: @lambday could you do something here?

src/shogun/machine/gp/FITCInferenceMethod.cpp
```diff
@@ -1,17 +1,34 @@
 /*
- * This program is free software; you can redistribute it and/or modify
- * it under the terms of the GNU General Public License as published by
- * the Free Software Foundation; either version 3 of the License, or
- * (at your option) any later version.
- *
+ * Copyright (c) The Shogun Machine Learning Toolbox
+ * Written (W) 2015 Wu Lin
  * Written (W) 2013 Roman Votyakov
- * Copyright (C) 2012 Jacob Walker
```
karlnapf: their code is deleted?

src/shogun/machine/gp/FITCInferenceMethod.cpp
```diff
@@ -76,17 +74,26 @@ void CFITCInferenceMethod::update()
 	SG_DEBUG("leaving\n");
 }
 
+CFITCInferenceMethod* CFITCInferenceMethod::obtain_from_generic(
+		CInferenceMethod* inference)
+{
+	ASSERT(inference!=NULL);
```
karlnapf: please return NULL in case you get NULL

src/shogun/machine/gp/FITCInferenceMethod.cpp
```diff
@@ -76,17 +74,26 @@ void CFITCInferenceMethod::update()
...
 void CFITCInferenceMethod::check_members() const
 {
-	CInferenceMethod::check_members();
+	CSingleFITCLaplacianBase::check_members();
 
 	REQUIRE(m_model->get_model_type()==LT_GAUSSIAN,
 			"FITC inference method can only use Gaussian likelihood function\n")
 	REQUIRE(m_labels->get_label_type()==LT_REGRESSION, "Labels must be type "
 			"of CRegressionLabels\n")
-	REQUIRE(m_inducing_features, "inducing features should not be NULL\n")
```
karlnapf: this might break python integration tests, did you check whether it does?
If not, then it is not covered by python examples, which is also bad

src/shogun/machine/gp/FITCInferenceMethod.cpp
```diff
@@ -246,6 +228,7 @@ void CFITCInferenceMethod::update_chol()
 	m_L=SGMatrix<float64_t>(m_kuu.num_rows, m_kuu.num_cols);
 	Map<MatrixXd> eigen_chol(m_L.matrix, m_L.num_rows, m_L.num_cols);
 
+	//post.L  = solve_chol(Lu*Luu,eye(nu)) - iKuu;
```
karlnapf: @lambday we need something as
http://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.cho_solve.html

or at least a triangular solve in linalg

src/shogun/machine/gp/FITCInferenceMethod.cpp
```diff
@@ -317,74 +306,102 @@ void CFITCInferenceMethod::update_deriv()
...
-	MatrixXd R=2*dKui-dKuui*eigen_B;
+	m_mu=SGVector<float64_t>(m_al.vlen);
+	Map<VectorXd> eigen_mu(m_mu.vector, m_mu.vlen);
 
-	// compute v=ddiagKi-sum(R.*B, 1)'
-	VectorXd v=ddiagKi-R.cwiseProduct(eigen_B).colwise().sum().adjoint();
+	/*
+	//true posterior mean with equivalent FITC prior
+	//time complexity of the following operations is O(n)
+	Map<VectorXd> eigen_al(m_al.vector, m_al.vlen);
```
karlnapf: ?

src/shogun/machine/gp/FITCInferenceMethod.cpp
```diff
@@ -317,74 +306,102 @@ void CFITCInferenceMethod::update_deriv()
...
+			m_chol_utr.num_cols);
 
-	return result;
+	/*
+	//true posterior mean with equivalent FITC prior
+	CGaussianLikelihood* lik=CGaussianLikelihood::obtain_from_generic(m_model);
+	float64_t sigma=lik->get_sigma();
+	SG_UNREF(lik);
+	Map<VectorXd> eigen_t(m_t.vector, m_t.vlen);
+	VectorXd diag_part=CMath::sq(sigma)*eigen_t;
```
karlnapf: ?

tests/unit/machine/gp/FITCInferenceMethod_unittest.cc
```diff
@@ -590,5 +655,428 @@ TEST(FITCInferenceMethod,get_marginal_likelihood_derivatives_sparse)
...
+	//while off-diagonal elements are in natrual domain
+
+	EXPECT_NEAR(dnlz_weights(0,0), -0.326900947549438, 1E-10);
+	EXPECT_NEAR(dnlz_weights(0,1), -6.119973491296843, 1E-10);
+	//dnlz_weights(1,0) should be different (non-zero)
+	//since the kernel matrix in the Shogun's implementation does not
+	//have the "upper triangular matrix" constraint
+	EXPECT_NEAR(dnlz_weights(1,1), 0.002268754179272, 1E-10);
+
+	EXPECT_NEAR(dnlZ_lik, 0.260126513715885,1E-10);
```
karlnapf: technically, there should be seperate tests for all of those variables.
Unit tests should be as small and atomic as possible

tests/unit/machine/gp/FITCInferenceMethod_unittest.cc
```diff
@@ -590,5 +655,428 @@ TEST(FITCInferenceMethod,get_marginal_likelihood_derivatives_sparse)
...
+	//Note that in the latest GPML3.5, derivatives wrt xu (covSEfact) does not support yet
+	//In Shohun's implementation, the derivatives wrt xu are supported. 
+	//TParameter* lat_param=inf->m_gradient_parameters->get_parameter("inducing_features");
+	//SGVector<float64_t> tmp=gradient->get_element(lat_param);
+	//SGMatrix<float64_t> deriv_lat(tmp.vector, dim, m, false);
+
+	// clean up
+	SG_UNREF(gradient);
+	SG_UNREF(parameter_dictionary);
+	SG_UNREF(inf);
```
karlnapf: same here, the test is a bit too long.
I suggest that you code up some set_up method that initialises everything and then the unit tests are very short and seperate for the gradients?

karlnapf:
This is a beautiful piece of work!

I would love to see a bit shorter unit tests.

Can I suggest that we re-work over the notebook a bit? 
Also, API examples for FITC would be good. Minimal ones.
Finally, could you benchmark your code against the other toolboxes to make sure there are no bottlenecks?

karlnapf:
As my comments for only style, I merge this so that you can go on. As usual, it would be great if you could send a mini patch for them later

karlnapf:
Really excited about your work, Wu. You are pushing Shogun towards being one of the most complete and efficient approx GP implementations out there! :)

src/interfaces/modular/GaussianProcess.i
yorkerlin: if seems that without adding them, some compiler(s) may throw some warning.

src/shogun/machine/gp/FITCInferenceMethod.cpp
yorkerlin: It seems that Roman used Jacob's framework.
In fact, I changed most of the code in `FITCInferenceMethod`

src/shogun/machine/gp/FITCInferenceMethod.cpp
yorkerlin: The reason is the base class has to the same checking job since `m_inducing_features` is a protected member of the base class.

src/shogun/machine/gp/FITCInferenceMethod.cpp
yorkerlin: As mentioned before, we have two posteriors.
* the true posterior p(f_n|y) \propto p(y|f_n)*p(f_n), where p(f_n) is the FITC equivalent FITC prior
* FITC approximated posterior  
p(f_n|y) = \int_{f_m} { p(f_n|f_m)*p(f_m|y) df_{m} }

src/shogun/machine/gp/FITCInferenceMethod.cpp
yorkerlin: see above

yorkerlin:
will do benchmark and notebook once the FITC for binary classification and diagonal `init` for ARD kernels are merged.
will send a mini patch to address your comments soon

src/shogun/machine/GaussianProcessMachine.cpp
lambday: @karlnapf sorry man I didn't follow. If only diagonal is required, couldn't we have used here `CKernel::get_kernel_diagonal()` instead?

yorkerlin: @lambday 
Yes. we can use `CKernel::get_kernel_diagonal()`
However, I may do some similar tricks when the diagonal elements of derivatives wrt hyper-parameter are required. say `get_parameter_gradient_diagonal(param, i)` if needed.

see `deriv_trtr=m_kernel->get_parameter_gradient(param, i).get_diagonal_vector();`

lambday: @yorkerlin well, sorry I don't know how that algorithm works - if the working memory is the same as the matrix size, then I don't think adding a separate method for diagonal will gain something. But if you think you can manage with just an array of the size of the diagonal as the working memory (for computing the parameter gradient) - then it is useful.

Does this get_parameter_gradient work for *all* the kernel functions?

yorkerlin: @lambday  
`get_parameter_gradient` is a method in `CKernel`

I may add a method called `get_parameter_gradient_diagonal` in `CKernel`, which will naively call get_parameter_gradient(param, i).get_diagonal_vector()

In ARD kernels, It seems  `get_parameter_gradient_diagonal` can be efficiently implemented.

lambday: That's great! Although I'd make sure that adding something generic to the base is actually something that is useful. You can add extra methods in the sub classes (wherever appropriate) and at the time of usage you can just dynamic cast the kernel to call that method.. for things that are not supposed to be general, keeping them limited to the scope where they are actually used helps at times. I don't know much about the context but I hope that it makes sense :)

Great work by the way :) Hope to go through your additions soon!

yorkerlin: @lambday 
As long as we can speed up the performance of computing diagonal elements of kernel matrix, we can gain the same speed for computing the diagonal elements of gradients wrt hyper-parameter(s).

@karlnapf  your opinion?

However, for now I will implement efficient `get_parameter_gradient_diagonal()` `ONLY for the ARD kernels`.

src/interfaces/modular/GaussianProcess.i
karlnapf: mmmh
should not be. @lisitsyn any ideas about this?
@sonney2k

src/shogun/machine/gp/FITCInferenceMethod.cpp
karlnapf: Yeah I see but why the commented out code?

src/shogun/machine/GaussianProcessMachine.cpp
karlnapf: I think putting the naive method in the base class (with a warning on costs in documentation) and then in the ARD kernels overload with the efficient version

src/shogun/machine/gp/FITCInferenceMethod.cpp
karlnapf: okok

karlnapf:
Hey!
Any news on this patch?

PR #2781, opened 2015-03-24
accidentally delete the original pull request as #2770

yingryic:


src/shogun/lib/SGMatrix.cpp
```diff
@@ -48,6 +48,7 @@ SGMatrix<T>::SGMatrix(index_t nrows, index_t ncols, bool ref_counting)
 	: SGReferencedData(ref_counting), num_rows(nrows), num_cols(ncols)
 {
 	matrix=SG_MALLOC(T, ((int64_t) nrows)*ncols);
+	zero();
```
karlnapf: why that?
We dont want that I think

src/shogun/lib/SGSparseMatrix.h
```diff
@@ -107,15 +107,15 @@ template <class T> class SGSparseMatrix : public SGReferencedData
 		{
 			REQUIRE(i_row>=0, "index %d negative!\n", i_row);
 			REQUIRE(i_col>=0, "index %d negative!\n", i_col);
-			REQUIRE(i_row<num_vectors, "index should be less than %d, %d provided!\n",
-				num_vectors, i_row);
-			REQUIRE(i_col<num_features, "index should be less than %d, %d provided!\n",
-				num_features, i_col);
+			REQUIRE(i_row<num_features, "index should be less than %d, %d provided!\n",
```
karlnapf: I know it is very minor. But could we do
"Provided index (%d) is larger than number of rows (%d)\n"

Capital letters at beginning of sentences, uniform style across Shogun. This is what our users see and it should be as clean as possible.

src/shogun/lib/SGMatrix.cpp
sonney2k: Yes indeed that would be slower (when we init the matrix later and
besides that one could use SG_CMALLOC if we really wanted this)

tests/unit/lib/SGSparseMatrix_unittest.cc
```diff
@@ -189,16 +191,15 @@ TEST(SGSparseMatrix, access_by_index_non_square)
 	const index_t randSeed=0;
 
 	//generate a sparse and dense matrix and compare this entries
-	SGSparseMatrix<float64_t> sparseMatrix(numberOfFeatures, numberOfVectors);
-	FillMatrixAccordingToSparseConvention(sparseMatrix, sparseLevel, numberOfFeatures, numberOfVectors, randSeed);
+	SGSparseMatrix<float64_t> sparseMatrix=
```
karlnapf: like this overloading

karlnapf:
Looks fine!
Thanks.

karlnapf:
Ah sorry I forgot about the zero().
Please send another patch to remove that.

Issue #2709, opened 2015-02-15
Support DotFeatures in neural networks


lisitsyn:
Currently neural nets require DenseFeatures input which restricts its usage a bit.

There is a good idea by @sonney2k (see [COFFIN](http://cmp.felk.cvut.cz/ftp/articles/franc/Sonnenburg-COFFIN-ICML10.pdf) paper) that most of the things depend on dot product of feature vector with some weight vector. This enables magic training on out-of-core data like local or remote files and even data generated on the fly. This task proposes to use input of CDotFeatures instead of CDenseFeatures as a general idea, but concrete suggestions are welcome! 

Good for [Deep learning](https://github.com/shogun-toolbox/shogun/wiki/GSoC_2015_project_deep_learning) applicants

sanuj:
@lisitsyn From what I understand, `CDenseFeatures` have to be replaced by `CDotFeatures` as neural nets training involves lots of dot products. Tell me if I'm wrong.
I know what Sparse (most zero entries) and Dense (most non-zero entries) matrices are. Are `CSparseFeatures` and `CDenseFeatures` also derived from the same logic? Then there are `SGSparseVectors` in shogun as well.
I didn't get the "magic training" part. Please elaborate, how training will be easy for out-of-core data (external memory, data not stored in RAM) etc. if we use `CDotFeatures`?
I saw the COFFIN paper is on Linear SVMs. Is there any particular section that I need to focus on? I'll read it tomorrow.

lisitsyn:
@sanuj 
1) Yes, ```CDenseFeatures -> CDotFeatures```
2) Yes, both dense and sparse features are examples of dot features
3) Dense features have to be in-memory but DotFeatures can be anything. As access is usually sequential it could be file or some resource accessed through network

sanuj:
@lisitsyn I see. Basically `CDenseFeatures` are matrices of all the features (as columns) so if we make a `CDenseFeature` then all the features are loaded at once in the matrix. I don't know how `CDotFeatures` are stored. Seems quite an abstract class to me. Also there are no unit tests for `CDotFeatures` from where I can get the idea of how to use them.
If I'm not wrong, the idea is if we use `CDotFeatures`, then all the features which are child classes of `CDotFeatures` can be supplied to the neural nets. (Right?)
I think following classes would be changed (`CDenseFeatures -> CDotFeatures`):
[...]

PR #2788, opened 2015-03-28
Some refactoring of neural nets

lisitsyn:
- Moved layers to separate dir
- Use DotFeatures api
- Dropped multiple inputs (that's controversial)

@khalednasr please take a look


src/shogun/neuralnets/NeuralLayer.h
```diff
@@ -147,10 +148,9 @@ class CNeuralLayer : public CSGObject
 	/** Computes the activations of the neurons in this layer, results should 
 	 * be stored in m_activations. To be used only with input layers
 	 * 
-	 * @param inputs activations of the neurons in the 
-	 * previous layer, matrix of size previous_layer_num_neurons * batch_size
+	 * @param inputs dot features to store activations from
 	 */
-	virtual void compute_activations(SGMatrix<float64_t> inputs) { }
+	virtual void compute_activations(CFeatures* inputs) { }
```
khalednasr: I think it would be better to keep ```inputs``` a matrix. If we gpu-ized things, and ```inputs``` was of type ```CFeatures```,  since ```CFeatures``` has its data on CPU memory, we'd have to copy it into GPU memory in every iteration of training, which would be pretty slow.

src/shogun/neuralnets/NeuralNetwork.cpp
```diff
@@ -270,6 +269,9 @@ bool CNeuralNetwork::train_gradient_descent(SGMatrix<float64_t> inputs,
 
 	bool continue_training = true;
 	float64_t alpha = gd_learning_rate;
+	SGVector<index_t> inputs_batch_subset(gd_mini_batch_size);
+	SGMatrix<float64_t> targets_batch;
+		//(targets.matrix+j*get_num_outputs(),
```
khalednasr: Got a stray comment here :)

src/shogun/neuralnets/NeuralNetwork.h
```diff
@@ -264,7 +265,7 @@ friend class CDeepBeliefNetwork;
 	 * 
 	 * @return activations of the last layer
 	 */
-	virtual SGMatrix<float64_t> forward_propagate(SGMatrix<float64_t> inputs, int32_t j=-1);
+	//virtual SGMatrix<float64_t> forward_propagate(SGMatrix<float64_t> inputs, int32_t j=-1);
```
khalednasr: and here

khalednasr:
@lisitsyn Thanks! looks good apart from them minor comments above.

I have no problems with dropping multiple inputs, they're not really useful.

src/shogun/neuralnets/NeuralNetwork.cpp
lisitsyn: Oops missed that

src/shogun/neuralnets/NeuralLayer.h
lisitsyn: I think we rather need to have GPU dot features that can do dot product right in the GPU memory.

If we go with dotfeatures API we get at least two advantages:

1) We can train with datasets that doesn't fit into RAM and GPU memory (but still use it as a kind of cache)
2) We can use complex features (random fourier, homogeneous kernel mapping, random mapping, etc) that can be computed on the fly with less memory footprint. I think (with giant batches and deep nets) GPUing things that are done after computing input would significantly speed up the learning process anyway.

That's all motivated by this COFFIN thing: http://www.icml2010.org/papers/280.pdf. That's a really cool idea by @sonney2k that could be one of the main features of our neural nets

khalednasr: I see. Well, as long as ```CDotFeatures``` will support GPU memory there won't be a problem.

lisitsyn:
@khalednasr probably multiple inputs can be emulated through features as well

I was thinking about siamese networks - this is a case of multiple inputs being used..

lisitsyn:
@khalednasr I am thinking of dropping input layer. We can directly ask CDotFeatures for dot product in linear unit. Convolution layer could retrieve features via ```add_to_dense_vec```

khalednasr:
@lisitsyn That's possible. But if we're gonna add multiple inputs through features, why remove them in the first place?

I like that ```CNeuralNetworkOptimizer``` thing. makes things much cleaner :)

src/shogun/neuralnets/DeepAutoencoder.cpp
```diff
@@ -240,7 +238,7 @@ void CDeepAutoencoder::init()
 	pt_contraction_coefficient.set_const(0.0);
 	
 	pt_optimization_method = SGVector<int32_t>((m_num_layers-1)/2);
-	pt_optimization_method.set_const(NNOM_LBFGS);
+	// pt_optimization_method.set_const(NNOM_LBFGS);
```
khalednasr: comments. here and throughout this file

sanuj:
@lisitsyn @khalednasr How much time is this PR going to take? I can start working on #2709

lisitsyn:
@sanuj yeah sorry I stopped at some point here. Can get back to that this weekend.

karlnapf:
any news here?

PR #2795, opened 2015-04-01
update the comments

yorkerlin:


src/shogun/machine/gp/MultiLaplacianInferenceMethod.cpp
```diff
@@ -362,6 +362,8 @@ void CMultiLaplacianInferenceMethod::update_alpha()
 		m_nlz+=Psi_New;
 	}
 
+	if (Psi_Old-Psi_New>m_tolerance && iter>=m_iter)
+		SG_WARNING("Max iterations attained but the optimizer is not converged\n");
```
karlnapf: maybe print the difference here?

karlnapf: and also the m_tolerance level

"Max iterations (%d) reached, but convergence level (%f) is not yet below tolerance (%f)"

karlnapf:
This is good to merge but I suggest to change these warning messages to be more informative

yorkerlin:
@karlnapf 
let me know whether this PR is merge-able

src/shogun/machine/gp/MultiLaplacianInferenceMethod.cpp
```diff
@@ -362,6 +362,10 @@ void CMultiLaplacianInferenceMethod::update_alpha()
 		m_nlz+=Psi_New;
 	}
 
+	if (Psi_Old-Psi_New>m_tolerance && iter>=m_iter)
```
karlnapf: Shouldn't this b absolute value?

karlnapf:
All good
Double check the abs value of convergence check

karlnapf:
I'll merge when back on my computer, cannot do from a phone

karlnapf:
@iglesias Yeah I know, but I guess there is a point in not merging things on a phone, since it might be "too quickly" ;) I am very inclined to such actions!

karlnapf:
Here we go

PR #2809, opened 2015-04-07
Added matrix-type independent elementwise-product in linalg

lambday:
- Also Put `VCLMatrixBase CGPUMatrix::vcl_matrix()` under SWIG guard to avoid possible SWIG issue.


lambday:
Travis issue seems unrelated. Merging.

karlnapf:
+1

PR #2811, opened 2015-04-07
Added matrix-type independent add in linalg

lambday:
@karlnapf @yorkerlin

### DIFF
--- a/src/shogun/mathematics/linalg/internal/implementation/Add.h
+++ b/src/shogun/mathematics/linalg/internal/implementation/Add.h
@@ -89,6 +89,29 @@ struct add<Backend::NATIVE, Matrix>
 	 * Performs the operation C = alpha*A + beta*B.
 	 * @param A first matrix
 	 * @param B second matrix
+	 * @param alpha constant to be multiplied by the first matrix
+	 * @param beta constant to be multiplied by the second matrix
+	 * @return The return matrix
+	 */
+	static SGMatrix<T> compute(SGMatrix<T> A, SGMatrix<T> B, T alpha=1, T beta=1)
+	{
+		REQUIRE(A.matrix, "Matrix A is not initialized!\n");
+		REQUIRE(B.matrix, "Matrix B is not initialized!\n");
+
+		REQUIRE(A.num_rows == B.num_rows && A.num_cols == B.num_cols,
+				"Dimension mismatch! A(%d x %d) vs B(%d x %d)\n",
+				A.num_rows, A.num_cols, B.num_rows, B.num_cols);
+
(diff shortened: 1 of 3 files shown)


src/shogun/mathematics/linalg/internal/implementation/Add.h
```diff
@@ -96,10 +119,12 @@ struct add<Backend::NATIVE, Matrix>
 	static void compute(SGMatrix<T> A, SGMatrix<T> B, SGMatrix<T> C,
 		T alpha=1, T beta=1)
 	{
-		REQUIRE(A.num_rows == B.num_rows && B.num_rows == C.num_rows,
-			"Matrices should have same number of rows!\n");
-		REQUIRE(A.num_cols == B.num_cols && B.num_cols == C.num_cols,
-			"Matrices should have same number of columns!\n");
+		REQUIRE(A.matrix, "Matrix A is not initialized!\n");
+		REQUIRE(B.matrix, "Matrix B is not initialized!\n");
```
lambday: Missed a few REQUIRES. Adding.

lambday:
Travis failure unrelated. Merging.

karlnapf:
+1

iglesias:
Hello, @lambday. Do you think it would make sense / be worth to use inheritance (or something else, maybe) in order to avoid some code duplication here?

For instance, we have now the specialised structs `add<Backend::Native, Matrix>`, `add<Backend::EIGEN3, Matrix>` , etc. with the same method signature and implementation for
`static SGMatrix<T> compute(SGMatrix<T> A, SGMatrix<T> B, T alpha=1, T beta=1) `,
among other methods.

lambday:
@iglesias nice point. As of the now, the implementation (apart from the REQUIRES) is (just a tiny bit) different. For this method,
```c++
static SGMatrix<T> compute(SGMatrix<T> A, SGMatrix<T> B, T alpha=1, T beta=1)
```
with NATIVE, calls a `compute()` that handles raw pointers, 
```c++
static void compute(T* A, T* B, T*C, T alpha, T beta, index_t len)
```
whether the EIGEN3 one calls a different `compute()`.
```c++
static void compute(SGMatrix<T> A, SGMatrix<T> B, SGMatrix<T> C, T alpha=1, T beta=1)
```
The inheritance thing can be useful if we make the NATIVE one call the same `compute()` that we're using for Eigen3. I used the raw-pointer one for NATIVE to avoid additional method calls (because the other one in turn calls that very same method in NATIVE - so one additional method call if we do that).

iglesias:
You are right, I didn't notice the difference in the types of the arguments. Thanks for explaining ;-)

PR #2813, opened 2015-04-08
fix an error when viennaCL is used

yorkerlin:
@karlnapf 
let's see whether this PR can fix the error when viennaCL is used.

### DIFF
--- a/src/shogun/kernel/GaussianARDKernel.cpp
+++ b/src/shogun/kernel/GaussianARDKernel.cpp
@@ -124,7 +124,7 @@ float64_t CGaussianARDKernel::get_parameter_gradient_helper(
 
 	if (!strcmp(param->m_name, "weights"))
 	{
-		linalg::add(avec, bvec, bvec, 1.0, -1.0);
+		bvec=linalg::add(avec, bvec, 1.0, -1.0);
 		float64_t scale=-kernel(idx_a,idx_b)/m_width;
 		return	compute_gradient_helper(bvec, bvec, scale, index);
 	}
@@ -192,7 +192,7 @@ float64_t CGaussianARDKernel::distance(int32_t idx_a, int32_t idx_b)
 
 	SGVector<float64_t> avec=((CDotFeatures *)lhs)->get_computed_dot_feature_vector(idx_a);
 	SGVector<float64_t> bvec=((CDotFeatures *)rhs)->get_computed_dot_feature_vector(idx_b);
-	linalg::add(avec, bvec, avec, 1.0, -1.0);
+	avec=linalg::add(avec, bvec, 1.0, -1.0);
 	float64_t result=compute_helper(avec, avec);
 	return result/m_width;
 }
--- a/src/shogun/kernel/LinearARDKernel.cpp
+++ b/src/shogun/kernel/LinearARDKernel.cpp
@@ -82,14 +82,12 @@ SGMatrix<float64_t> CLinearARDKernel::compute_right_product(SGVector<float64_t>r
 	}
 	else
 	{
-		right=SGMatrix<float64_t> (m_weights.num_rows,1);
-
 		SGMatrix<float64_t> rtmp(right_vec.vector,right_vec.vlen,1,false);
 
 		if(m_ARD_type==KT_DIAG)
-			linalg::elementwise_product(m_weights, rtmp, right);
+			right=linalg::elementwise_product(m_weights, rtmp);
 		else if(m_ARD_type==KT_FULL)
-			linalg::matrix_product(m_weights, rtmp, right);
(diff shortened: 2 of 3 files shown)


src/shogun/kernel/GaussianARDKernel.cpp
```diff
@@ -124,7 +124,7 @@ float64_t CGaussianARDKernel::get_parameter_gradient_helper(
 
 	if (!strcmp(param->m_name, "weights"))
 	{
-		linalg::add(avec, bvec, bvec, 1.0, -1.0);
+		bvec=linalg::add(avec, bvec, 1.0, -1.0);
```
lambday: @yorkerlin ah due the usage of this, we're missing the _in-place_ benefits here :(
Shall I add a param to specify whether you want this _in-place_ or not?

src/shogun/kernel/LinearARDKernel.cpp
```diff
@@ -108,29 +106,24 @@ float64_t CLinearARDKernel::compute_helper(SGVector<float64_t> avec, SGVector<fl
...
-		SGMatrix<float64_t> left_transpose(left.matrix,left.num_cols,1,false);
 		if(m_ARD_type==KT_DIAG)
-			linalg::elementwise_product(m_weights, ltmp, left_transpose);
+			left_transpose=linalg::elementwise_product(m_weights, ltmp);
 		else if(m_ARD_type==KT_FULL)
-			linalg::matrix_product(m_weights, ltmp, left_transpose);
+			left_transpose=linalg::matrix_product(m_weights, ltmp);
 		else
 			SG_ERROR("Unsupported ARD type\n");
+		left=SGMatrix<float64_t>(left_transpose.matrix,1,left_transpose.num_rows,false);
```
lambday: @yorkerlin you need non-refcounted SGMatrix here because you're constructing it from a SGVector? I remember talking a couple of constructors in SGMatrix/SGVector to use it back and forth _with_ refcounting on. Maybe adding this now would be a good idea.

src/shogun/kernel/GaussianARDKernel.cpp
iglesias: +1 to supporting *in-place* too.

yorkerlin: yes, if possible

src/shogun/kernel/LinearARDKernel.cpp
yorkerlin: yes. Let me know once they are available.

karlnapf:
Let me know when to merge this

karlnapf:
Good to go from travis side

src/shogun/kernel/GaussianARDKernel.cpp
karlnapf: In place would be good in general, but keep in mind this is just a vector operation (and the vector is not large)
In-place is more crucial when we create copies of the kernel matrices (which are the main memory bottleneck of GPs)

karlnapf:
Seems good to me, Ill merge

src/shogun/kernel/LinearARDKernel.cpp
lambday: @yorkerlin I just merged the code. Two new constructors in SGMatrix are now available to share the same memory and refcount with an SGVector. Check #2814 (the usage is there in the unit tests added in that PR). Should be helpful.

src/shogun/kernel/GaussianARDKernel.cpp
lambday: @karlnapf yeah but presently these operations work the same way for matrices as well (in-place not allowed if we're not using an explicit backend). Maybe I'll think of something about how to handle this backend-independently.

yorkerlin:
@karlnapf  @lambday 
It seems there is still some error when viennaCL is used.
Could you tell me how to configure `cmake` so that `eigen3` is used in GP module while `viennaCL` is used in `linalg`

lambday:
@yorkerlin if both eigen3 and viennacl is there, presently it is not possible to set the latter as default with cmake. If you want to reproduce the error (and if you need both eigen3 and viennacl for that, with viennacl being default), you can locally switch their preferences in linalg.h and test. You need to change their order in the enum Backend.
```c++
enum class Backend
{
#ifdef HAVE_VIENNACL
    VIENNACL,
#endif
#ifdef HAVE_EIGEN3
    EIGEN3,
#endif
    NATIVE,
    DEFAULT = 0 
};
```
BTW the cmake that the trusty-libshogun-viennacl builder uses is
`cmake -DENABLE_TESTING=ON -DLinAlgBackend=VIENNACL ..`

PR #2814, opened 2015-04-10
Added two constructors in SGMatrix with SGVectors

lambday:
- Both copy the refcount of the SGVector to make sure that memory
  is not accidentally deleted. Therefore we don't have to turn off
  refcount for the SGMatrix as well

### DIFF
--- a/src/shogun/lib/SGMatrix.cpp
+++ b/src/shogun/lib/SGMatrix.cpp
@@ -51,6 +51,30 @@ SGMatrix<T>::SGMatrix(index_t nrows, index_t ncols, bool ref_counting)
 }
 
 template <class T>
+SGMatrix<T>::SGMatrix(SGVector<T> vec) : SGReferencedData(vec)
+{
+	REQUIRE(vec.vector, "Vector not initialized!\n");
+	matrix=vec.vector;
+	num_rows=vec.vlen;
+	num_cols=1;
+}
+
+template <class T>
+SGMatrix<T>::SGMatrix(SGVector<T> vec, index_t nrows, index_t ncols)
+: SGReferencedData(vec)
+{
+	REQUIRE(vec.vector, "Vector not initialized!\n");
+	REQUIRE(nrows>0, "Number of rows (%d) has to be a positive integer!\n", nrows);
+	REQUIRE(ncols>0, "Number of cols (%d) has to be a positive integer!\n", ncols);
+	REQUIRE(vec.vlen==nrows*ncols, "Number of elements in the matrix (%d) must "
+			"be the same as the number of elements in the vector (%d)!\n",
+			nrows*ncols, vec.vlen);
+	matrix=vec.vector;
+	num_rows=nrows;
+	num_cols=ncols;
+}
+
+template <class T>
 SGMatrix<T>::SGMatrix(const SGMatrix &orig) : SGReferencedData(orig)
 {
 	copy_data(orig);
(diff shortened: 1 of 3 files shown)


lambday:
@karlnapf @sonney2k please have a look.
@yorkerlin this can be useful for you.

src/shogun/lib/SGMatrix.h
```diff
@@ -57,6 +57,25 @@ template<class T> class SGMatrix : public SGReferencedData
...
+		 * Constructor for creating a SGMatrix from a SGVector with refcounting.
+		 * This constructor assumes that the vector is the column 0 in the matrix.
+		 *
+		 * @param vec The SGVector
+		 */
+		SGMatrix(SGVector<T> vec);
+
+		/**
+		 * Constructor for creating a SGMatrix from a SGVector with refcounting.
+		 * The number of elements in the matrix *MUST* be same with the number
```
iglesias: Same as instead of same with, sorry for being annoying :-)

src/shogun/lib/SGMatrix.h
```diff
@@ -57,6 +57,25 @@ template<class T> class SGMatrix : public SGReferencedData
 		/** Constructor to create new matrix in memory */
 		SGMatrix(index_t nrows, index_t ncols, bool ref_counting=true);
 
+		/**
+		 * Constructor for creating a SGMatrix from a SGVector with refcounting.
```
iglesias: Even if it should be enough saying "with refcounting" as it is, I would perhaps remark here that no data is copied. Also in the method below.

iglesias:
We probably don't need these constructors from interfaces, right? Then, it may be a good idea to have them within `#ifndef SWIG`.

Nice one, @lambday!

src/shogun/lib/SGMatrix.h
lambday: Hehe no issues :D That indeed would be the right one. Will update :)

lambday:
@iglesias well, I'm not sure if we'd ever be using this from interfaces. But since we face a lot of problem with swig already due to gigantic size of that cxx, I'd better put these under the guard. If anyone requires, we can anytime remove that guard.

Thanks for your comments :)

iglesias:
As I like to think about it, SGVector and SGMatrix should never really be used directly from interfaces. In the interfaces, the most preferred/default data structure for vector and matrices should be used (e.g. numpy.array in Python) and our work is to provide nice type maps to these data structures from the interfaced languages.

lambday:
@iglesias yeah that makes sense. For other languages also I suppose we have things like that, e.g. java?

lambday:
@iglesias updated the PR. Could you please check once if the things look good now? :)

iglesias:
:+1:

By the way, sorry that I didn't answer what you asked about Java and the type maps, I completely overlooked the question! I am not 100% sure about it, but I think in particular in Java we use type maps to convert between SGVector, SGMatrix, etc, and JBlas data structures. Here there seems to be something related to it: https://github.com/iglesias/shogun/blob/master/src/interfaces/java_modular/swig_typemaps.i

lambday:
Thanks @iglesias. Now that you mentioned, it does ring a bell. I remember writing a blog post on using Shogun with java long back. I used something related to jblas there, then totally forgot myself (never used java modular after that :D) 

BTW shall I merge this code?

iglesias:
Sure, looking great to me.

lambday:
Merging then.

src/shogun/lib/SGMatrix.cpp
```diff
@@ -51,6 +51,30 @@ SGMatrix<T>::SGMatrix(index_t nrows, index_t ncols, bool ref_counting)
 }
 
 template <class T>
+SGMatrix<T>::SGMatrix(SGVector<T> vec) : SGReferencedData(vec)
```
karlnapf: +1

src/shogun/lib/SGMatrix.h
```diff
@@ -57,18 +57,45 @@ template<class T> class SGMatrix : public SGReferencedData
...
+		 * ref-count object).
+		 *
+		 * This constructor assumes that the vector is the column 0 in the matrix.
+		 *
+		 * @param vec The SGVector
+		 */
+		SGMatrix(SGVector<T> vec);
+
+		/**
+		 * Constructor for creating a SGMatrix from a SGVector with refcounting.
```
karlnapf: Please mention that Shogun does COL-major matrix storage

karlnapf:
Good stuff

Issue #2819, opened 2015-04-14
CParameterCombination prints SGVectors incorrectly


sorig:
When training the hyperparameters of the spectral mixture kernel (see #2818) I noticed that SGVectors didn't print correctly when using the `print_tree` method on the `CParameterCombination` class. The SGVector parameter on the kernel object has a container datatype on the `TSGDataType` object set to `CT_VECTOR` instead of `CT_SGVECTOR`. Consequently the SGVector parameter is not being printed correctly.

I have created a fix, but I don't know if the fact that the container type is wrong constitutes and underlying issue.

sorig:
Can anyone tell me if there is an underlying problem with the container datatype or if my fix is adequate?

karlnapf:
As we discussed, this is a bug related to legacy code.
The method itself should probably be removed as the terminal might not be available in modular interfaces.
We should rather return an adequate data structure with the resulting parameters. But well

karlnapf:
Fixed in #2820

PR #2820, opened 2015-04-14
Fix parameter combination print for SGVectors

sorig:
See issue #2819

### DIFF
--- a/src/shogun/modelselection/ParameterCombination.cpp
+++ b/src/shogun/modelselection/ParameterCombination.cpp
@@ -308,6 +308,8 @@ void CParameterCombination::print_tree(int prefix_num) const
 		SG_SPRINT("%s", prefix)
 		for (index_t i=0; i<m_param->get_num_parameters(); ++i)
 		{
+			EContainerType ctype = m_param->get_parameter(i)->m_datatype.m_ctype;
+
 			/* distinction between sgobject and values */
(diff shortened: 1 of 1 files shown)


src/shogun/modelselection/ParameterCombination.cpp
```diff
@@ -316,22 +318,21 @@ void CParameterCombination::print_tree(int prefix_num) const
 				SG_SPRINT("\"%s\":%s at %p ", param->m_name,
 						current_sgobject->get_name(), current_sgobject);
 			}
-
-		    else if (m_param->get_parameter(i)->m_datatype.m_ctype == CT_SGVECTOR)
+		    else if (ctype == CT_SGVECTOR || ctype == CT_VECTOR)
```
lisitsyn: Does it correspond to the if at line 314? Something strange with indentation

sorig:
@lisitsyn The indentation issue is fixed now (the indentation in the original version of the file is also weird). The fix is simply to add an extra check for the container type being `CT_VECTOR` in the the if statement at [line 320 in the original version of the file](https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/modelselection/ParameterCombination.cpp#L320)

karlnapf:
Seems fine to me!

karlnapf:
And travis errors are unrelated too, merging
Thanks!

PR #2829, opened 2015-04-24
replace obsolete `json_object_object_get` with `json_object_object_get_ex`

besser82:
This funct is deprecated since json-c >= 0.9, we need >=0.11.  So no problems are expected nor ifdefs.


besser82:
Build fails are not related…  Merging…

PR #2818, opened 2015-04-14
Spectral mixture kernel

sorig:
Implementation of the [spectral mixture kernel](http://jmlr.org/proceedings/papers/v28/wilson13.pdf) which can be used for extrapolation with GPs and approximation of any stationary kernel.

PR #2830, opened 2015-04-25
Added element-wise unary operators in linalg

lambday:
- Works with predefined operations (sin, cos, log, exp etc.)
- Works with custom unary operators as well (functors, std::function, lambda expressions, function pointers) that developers can provide.
- Implicit type conversion is **turned off** for CPU to/from GPU matrix/vector data-types. NATIVE/EIGEN3 works only with SG-types and VIENNACL works only with CGPU-types.
- Type safety ensured.
- A benchmark of element-wise operation with a custom unary operator.

### DIFF
--- a/benchmarks/elementwise_benchmark.cpp
+++ b/benchmarks/elementwise_benchmark.cpp
@@ -1,0 +1,130 @@
+/*
+ * Copyright (c) The Shogun Machine Learning Toolbox
+ * Written (w) 2015 Soumyajit De
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * The views and conclusions contained in the software and documentation are those
+ * of the authors and should not be interpreted as representing official policies,
+ * either expressed or implied, of the Shogun Development Team.
+ */
+
+#include <shogun/lib/SGMatrix.h>
+#include <shogun/lib/GPUMatrix.h>
+#include <shogun/mathematics/linalg/linalg.h>
+#include <vector>
+#include <algorithm>
+#include <hayai/hayai.hpp>
+
+using namespace shogun;
+
+/**
+ * Instructions :
+ * 1. Install benchmarking toolkit "hayai" (https://github.com/nickbruun/hayai)
+ * 2. Compile against libhayai_main, e.g.
+ * g++ -O3 -std=c++11 elementwise_benchmark.cpp -I/usr/include/eigen3 -lshogun -lhayai_main -lOpenCL -o benchmark
+ * 3. ./benchmark
+ */
+
+/** Generate data only once */
+struct Data
+{
+	Data()
+	{
+		init();
+	}
+
+	void init()
+	{
+		m_cpu=SGMatrix<float32_t>(num_rows, num_cols);
+		std::iota(m_cpu.data(), m_cpu.data()+m_cpu.size(), 1);
+		m_gpu=CGPUMatrix<float32_t>(m_cpu);
+	}
+
+	SGMatrix<float32_t> m_cpu;
+	CGPUMatrix<float32_t> m_gpu;
+
+	static constexpr index_t num_rows=1000;
+	static constexpr index_t num_cols=1000;
+};
+
+Data data;
+
+BENCHMARK(SGMatrix, elementwise, 10, 1000)
+{
+	float32_t weights=0.6;
+	float32_t std_dev=0.2;
+	float32_t mean=0.01;
+
+	SGMatrix<float32_t> result=linalg::elementwise_compute(data.m_cpu,
+	[&weights, &std_dev, &mean](float32_t& sqr_dist)
+	{
+		float32_t outer_factor=-2*CMath::PI*CMath::sqrt(sqr_dist)*CMath::sq(weights);
+		float32_t exp_factor=CMath::exp(-2*CMath::sq(CMath::PI)*sqr_dist*CMath::sq(std_dev));
+		float32_t sin_factor=CMath::sin(2*CMath::PI*CMath::sqrt(sqr_dist)*mean);
+		return outer_factor*exp_factor*sin_factor;
+	});
+
+}
+
+BENCHMARK(SGMatrix, loop, 10, 1000)
+{
+	float32_t weights=0.6;
+	float32_t std_dev=0.2;
+	float32_t mean=0.01;
+
+	SGMatrix<float32_t> result(data.m_cpu.num_rows, data.m_cpu.num_cols);
+
+	for (index_t j=0; j<data.m_cpu.num_cols; ++j)
+	{
+		for (index_t i=0; i<data.m_cpu.num_rows; ++i)
+		{
+			float32_t sqr_dist=data.m_cpu(i, j);
+			float32_t outer_factor=-2*CMath::PI*CMath::sqrt(sqr_dist)*CMath::sq(weights);
+			float32_t exp_factor=CMath::exp(-2*CMath::sq(CMath::PI)*sqr_dist*CMath::sq(std_dev));
+			float32_t sin_factor=CMath::sin(2*CMath::PI*CMath::sqrt(sqr_dist)*mean);
+			result(i, j)=outer_factor*exp_factor*sin_factor;
+		}
+	}
+}
+
+BENCHMARK(CGPUMatrix, elementwise, 10, 1000)
+{
+	float32_t weights=0.6;
+	float32_t std_dev=0.2;
+	float32_t mean=0.01;
+
+	std::string data_type=linalg::implementation::ocl::get_type_string<float32_t>();
+
+	std::string s_weights=std::to_string(weights);
+	std::string s_std_dev=std::to_string(std_dev);
+	std::string s_mean=std::to_string(mean);
+	std::string s_pi=std::to_string(CMath::PI);
+
+	std::string operation;
+	operation.append(data_type+" outer_factor=-2*"+s_pi+"*sqrt(element)*pow("+s_weights+", 2);\n");
+	operation.append(data_type+" exp_factor=exp(-2*pow("+s_pi+",2)*element*pow("+s_std_dev+", 2));\n");
+	operation.append(data_type+" sin_factor=sin(2*"+s_pi+"*sqrt(element)*"+s_mean+");\n");
+	operation.append("return outer_factor*exp_factor*sin_factor;");
+
+	linalg::elementwise_compute_inplace(data.m_gpu, operation);
+}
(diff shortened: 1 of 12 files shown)


lambday:
@karlnapf @lisitsyn please have a look!

lambday:
The result for the benchmark which uses the same operation that Esben used in #2818. 
```
[==========] Running 3 benchmarks.
[ RUN      ] SGMatrix.elementwise (10 runs, 1000 iterations per run)
[     DONE ] SGMatrix.elementwise (583458.460000 ms)
[   RUNS   ]        Average time: 58345846.000 us
                         Fastest: 57482036.000 us (-863810.000 us / -1.480 %)
                         Slowest: 59081968.000 us (+736122.000 us / +1.262 %)
                                  
             Average performance: 0.01714 runs/s
                Best performance: 0.01740 runs/s (+0.00026 runs/s / +1.50275 %)
               Worst performance: 0.01693 runs/s (-0.00021 runs/s / -1.24593 %)
[ITERATIONS]        Average time: 58345.846 us
                         Fastest: 57482.036 us (-863.810 us / -1.480 %)
                         Slowest: 59081.968 us (+736.122 us / +1.262 %)
                                  
             Average performance: 17.13918 iterations/s
                Best performance: 17.39674 iterations/s (+0.25756 iterations/s / +1.50275 %)
               Worst performance: 16.92564 iterations/s (-0.21354 iterations/s / -1.24593 %)
[ RUN      ] SGMatrix.loop (10 runs, 1000 iterations per run)
[     DONE ] SGMatrix.loop (630158.857000 ms)
[   RUNS   ]        Average time: 63015885.700 us
                         Fastest: 62119686.000 us (-896199.700 us / -1.422 %)
                         Slowest: 64610928.000 us (+1595042.300 us / +2.531 %)
                                  
             Average performance: 0.01587 runs/s
                Best performance: 0.01610 runs/s (+0.00023 runs/s / +1.44270 %)
               Worst performance: 0.01548 runs/s (-0.00039 runs/s / -2.46869 %)
[ITERATIONS]        Average time: 63015.886 us
                         Fastest: 62119.686 us (-896.200 us / -1.422 %)
                         Slowest: 64610.928 us (+1595.042 us / +2.531 %)
                                  
             Average performance: 15.86901 iterations/s
                Best performance: 16.09796 iterations/s (+0.22894 iterations/s / +1.44270 %)
               Worst performance: 15.47726 iterations/s (-0.39176 iterations/s / -2.46869 %)
[ RUN      ] CGPUMatrix.elementwise (10 runs, 1000 iterations per run)
[     DONE ] CGPUMatrix.elementwise (188.860000 ms)
[   RUNS   ]        Average time: 18886.000 us
                         Fastest: 9224.000 us (-9662.000 us / -51.160 %)
                         Slowest: 85618.000 us (+66732.000 us / +353.341 %)
                                  
             Average performance: 52.94927 runs/s
                Best performance: 108.41284 runs/s (+55.46356 runs/s / +104.74848 %)
               Worst performance: 11.67979 runs/s (-41.26949 runs/s / -77.94155 %)
[ITERATIONS]        Average time: 18.886 us
                         Fastest: 9.224 us (-9.662 us / -51.160 %)
                         Slowest: 85.618 us (+66.732 us / +353.341 %)
                                  
             Average performance: 52949.27459 iterations/s
                Best performance: 108412.83608 iterations/s (+55463.56148 iterations/s / +104.74848 %)
               Worst performance: 11679.78696 iterations/s (-41269.48763 iterations/s / -77.94155 %)
[==========] Ran 3 benchmarks.
```

src/shogun/lib/SGMatrix.h
```diff
@@ -170,6 +177,18 @@ template<class T> class SGMatrix : public SGReferencedData
...
 		}
 
+		/** The data */
+		inline T* data() const
+		{
+			return matrix;
+		}
+
+		/** The size */
+		inline index_t size() const
```
karlnapf: overflow risk with 32  bit and product?

src/shogun/mathematics/linalg/internal/implementation/ElementwiseUnaryOperation.h
```diff
@@ -0,0 +1,198 @@
...
+	using T = typename Operand::Scalar;
+
+	/** The scalar type of the result */
+	using ST = typename ReturnType::Scalar;
+
+#ifdef HAVE_VIENNACL
+	/** Ensure that this struct is not being instantiated with any GPU operand types */
+	static_assert(std::is_same<SGMatrix<T>, Operand>::value
+			|| std::is_same<SGVector<T>, Operand>::value,
+			"NATIVE backend not allowed for GPU operands! Use SGMatrix/SGVector "
```
karlnapf: This is good!
BTW do we have both cases covered in our builds @vigsterkr @besser82 @lambday ?

benchmarks/elementwise_benchmark.cpp
```diff
@@ -0,0 +1,130 @@
...
+Data data;
+
+BENCHMARK(SGMatrix, elementwise, 10, 1000)
+{
+	float32_t weights=0.6;
+	float32_t std_dev=0.2;
+	float32_t mean=0.01;
+
+	SGMatrix<float32_t> result=linalg::elementwise_compute(data.m_cpu,
+	[&weights, &std_dev, &mean](float32_t& sqr_dist)
```
karlnapf: I have to admid, I just learned about this construct ;)

benchmarks/elementwise_benchmark.cpp
```diff
@@ -0,0 +1,130 @@
...
+
+	std::string s_weights=std::to_string(weights);
+	std::string s_std_dev=std::to_string(std_dev);
+	std::string s_mean=std::to_string(mean);
+	std::string s_pi=std::to_string(CMath::PI);
+
+	std::string operation;
+	operation.append(data_type+" outer_factor=-2*"+s_pi+"*sqrt(element)*pow("+s_weights+", 2);\n");
+	operation.append(data_type+" exp_factor=exp(-2*pow("+s_pi+",2)*element*pow("+s_std_dev+", 2));\n");
+	operation.append(data_type+" sin_factor=sin(2*"+s_pi+"*sqrt(element)*"+s_mean+");\n");
```
karlnapf: Not sure I understand this way of constructing strings to tell the GPU what to do

karlnapf: I see, so there is a number of predefined operation strings that we provide, but users also can put their own ones.

How does a wrongly typed operation's error look like?

karlnapf:
Is this benchmark with compiler optimisation enabled?
Do I get this correct
 * CPU element wise: 44 it/s
 * CPU explicit loop 18 it/s
 * GPU: 2413 it/s

Factor 500?

karlnapf:
If so, awesome :) Good to merge from my side! Check travis though, just restarted a few jobs

Awesome work
@sorig check it out, and make use of it in your kernel once this is merged
Also @sorig , once your project is done, are you interested in putting in the base class for stationary kernels we discussed? That pulls out computing the distances from the sub-classes and tries to speed them up with GPU?

src/shogun/lib/SGMatrix.h
lambday: Yes I should change it to uint64_t or size_t.

benchmarks/elementwise_benchmark.cpp
lambday: Hehe. Just doing justification to my IRC handle :D There are lambda functions.

src/shogun/mathematics/linalg/internal/implementation/ElementwiseUnaryOperation.h
lambday: We have a dedicated ViennaCL build in buildbot. But that just sets the only the global backends to ViennaCL. I think we should get rid of global ViennaCL backend setups, because when that's done, one would take a massive performance hit due to all the data transfer in all the operations. Backends should always be specified explicitly. This static assert is an attempt to prevent from doing the such unintentional things.

src/shogun/mathematics/linalg/internal/modules/ElementwiseOperations.h
```diff
@@ -0,0 +1,179 @@
...
+ *
+ * This method returns the result in a newly allocated matrix/vector.
+ *
+ * @param operand The operand on which the element-wise operation has to be performed
+ * @param unary_op The custom unary operator
+ * @return The result of the unary operator applied element-wise on the operand
+ */
+template <class Operand, class UnaryOp>
+auto elementwise_compute(Operand operand, UnaryOp unary_op)
+-> typename Operand::template container_type<decltype(unary_op(operand.data()[0]))>
```
lambday: Took me a while to come up with this one here :)

benchmarks/elementwise_benchmark.cpp
lambday: @karlnapf it gives a runtime error from OpenCL backend while compiling the kernel. If one follows the error description, it is not hard to figure out what went wrong.

There is not much magic regarding how the string should look like. It is plain C code, just written inside a string. This whole string that we're passing forms the body of a function, named `operation`, where the current element in the element-wise operation is called `element`. For example, here, the `operation` function looks like
```c
inline float operation(float element)
{
    // the string that we provide gets replaced here
    float outer_factor=-2*3.1415+sqrt(element)*pow(0.6, 2);
    float exp_factor=exp(-2*pow(3.1415,2)*element*pow(0.2, 2));
    float sin_factor=sin(2*3.1415*sqrt(element)*0.01);
    return outer_factor*exp_factor*sin_factor;
}
```
And then this `operation` is used inside an OpenCL `kernel`
```c
__kernel void kernel_some_unique_hash_float(__global float* vec, int size, int vec_offset, 
            __global float* result, int result_offset)
{
    int i = get_global_id(0);
    if (i<size)
        result[i+result_offset] = operation(vec[i+vec_offset]);
}
```
The user would just have to provide the body of the operation. Everything else is taken care of inside the framework (thanks to Khaled's [awesome code](https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/mathematics/linalg/internal/opencl_util.h#L136)

The only limitation is, the functions we can use inside those string must be supported by OpenCL. Like sin/cos/exp/pow/log etc are fine, but rand is not (so far I know).

lambday:
@karlnapf these results are on my `Intel HD Graphics IvyBridge M GT2` (which is pretty poor). These results could crank up a notch if someone could benchmark these with `NVIDIA GeForce GTX` I am hoping. One interesting thing is that, when I tried this on CPU+OpenCL, then I got even better performance. Not sure I understand why.

lambday:
@karlnapf  BTW you were right. Using `OpenMP` for `NATIVE` I noticed an improved performance than `SIMD` for these complex operations. Plus these function objects are most likely to be inlined by the compiler, therefore they are expected to be faster than a normal function pointer.

lambday:
Travis error is unrelated. Merging this one.

benchmarks/elementwise_benchmark.cpp
```diff
@@ -0,0 +1,130 @@
...
+
+BENCHMARK(CGPUMatrix, elementwise, 10, 1000)
+{
+	float32_t weights=0.6;
+	float32_t std_dev=0.2;
+	float32_t mean=0.01;
+
+	std::string data_type=linalg::implementation::ocl::get_type_string<float32_t>();
+
+	std::string s_weights=std::to_string(weights);
```
lisitsyn: What about wrapping it like ```parameter(weights)``` so it does the same thing converting to string?

benchmarks/elementwise_benchmark.cpp
```diff
@@ -0,0 +1,130 @@
...
+	std::string s_std_dev=std::to_string(std_dev);
+	std::string s_mean=std::to_string(mean);
+	std::string s_pi=std::to_string(CMath::PI);
+
+	std::string operation;
+	operation.append(data_type+" outer_factor=-2*"+s_pi+"*sqrt(element)*pow("+s_weights+", 2);\n");
```
lisitsyn: Two possible approaches to avoid strings concat here:

```Typed(data_type, "outer_factor") = -2*Parameter(CMath::PI)*Sqrt(Element)```

Implementation: all functions create wrappers for strings, all ops (like multiplication) just concat it to the left or to the right, assignment concats with ```=```. Result is convertible to std::string

or something formatted

```Format("{type} {name} = -2*{pi}*sqrt(element)*pow({weights},2)}", Parameter("type") = data_type, Parameter("name") = "outer_factor", ...)```

Implementation is probably easy

benchmarks/elementwise_benchmark.cpp
karlnapf: haha, awesome!!! :)

benchmarks/elementwise_benchmark.cpp
karlnapf: ok cool

and overhead?

src/shogun/mathematics/linalg/internal/implementation/ElementwiseUnaryOperation.h
karlnapf: Yeah I agree, we should be as explicit as possible

src/shogun/mathematics/linalg/internal/modules/ElementwiseOperations.h
karlnapf: btw do we also have these unary operations inline?

karlnapf: ehm inplace I mean

karlnapf:
Cool!


@sorig feel free to use this in your patch!

karlnapf:
@lambday awesome work!

benchmarks/elementwise_benchmark.cpp
lambday: @lisitsyn hey man could you please have a look at [this](https://gist.github.com/lambday/21618e6bac4898dd2149)? Didn't get much time to work on it earlier. This one is for the formatted solution.

I spent some time earlier trying to work out things for the first solution but I don't think it can be done. Operators for basic types cannot be overloaded in C++. Maybe you have some ideas?

lisitsyn: That looks cool for me.

Actually you don't need rewriting operators for basic types - just operators of ```Parameter``` and ```float/int``` so that any multiplication/addition/etc goes inside your object that represents expression.

src/shogun/mathematics/linalg/internal/modules/ElementwiseOperations.h
lambday: @karlnapf yes in-place versions are also there.

PR #2835, opened 2015-05-01
users can disable the gradient computation

yorkerlin:


yorkerlin:
@karlnapf 
take a look at this

src/shogun/machine/gp/InferenceMethod.cpp
```diff
@@ -75,6 +75,7 @@ void CInferenceMethod::init()
 	SG_ADD((CSGObject**)&m_mean, "mean_function", "Mean function", MS_AVAILABLE);
 	SG_ADD((CSGObject**)&m_labels, "labels", "Labels", MS_NOT_AVAILABLE);
 	SG_ADD((CSGObject**)&m_features, "features", "Features", MS_NOT_AVAILABLE);
+	SG_ADD(&m_is_compute_gradients, "is_compute_gradients", "whether gradients are computed", MS_NOT_AVAILABLE);
```
karlnapf: why is the field called "is" ?

what about "m_compute_gradients?

Minor "Whether gradients are computed" (capital W)

src/shogun/machine/gp/InferenceMethod.cpp
```diff
@@ -294,4 +297,21 @@ void CInferenceMethod::update_train_kernel()
 	m_ktrtr=m_kernel->get_kernel_matrix();
 }
 
+void CInferenceMethod::set_compute_gradients(bool is_enabled)
```
karlnapf: super minor: we usually call the variable in setters the same as the member, but without the "m_" prefix

src/shogun/machine/gp/InferenceMethod.cpp
```diff
@@ -294,4 +297,21 @@ void CInferenceMethod::update_train_kernel()
...
+}
+
+bool CInferenceMethod::is_compute_gradients()
+{
+	return m_is_compute_gradients;
+}
+
+void CInferenceMethod::check_compute_gradients()
+{
+	REQUIRE(m_is_compute_gradients,
```
karlnapf: This seems weird to have an error in this method?

src/shogun/machine/gp/InferenceMethod.h
```diff
@@ -428,6 +439,8 @@ class CInferenceMethod : public CDifferentiableFunction
 	 */
 	static void* get_derivative_helper(void* p);
 
+	/** ensure that gradients computation is enabled*/
```
karlnapf: I see, so this is more "assert_compute_gradients"

tests/unit/classifier/GaussianProcessClassification_unittest.cc
```diff
@@ -3231,6 +3243,7 @@ TEST(GaussianProcessClassificationUsingMultiLaplacian,get_variance_vector)
 
 	const float64_t scale=CMath::sqrt(497.3965463400368);
 	inf->set_scale(scale);
+	inf->set_compute_gradients(false);
```
karlnapf: if you have to change this everywhere, shouldnt this be "false" by default?

karlnapf:
It is good to have this option.
BUT:

 * I don't like that users have to think about this
 * Gradients should be computed if model-selection is performed and otherwise not
 * without the user interacting
 * If users explicitly asks for gradients, should be computed lazily
 * So the setter for the flag is not the most elegant way to do all this

Thoughts?

yorkerlin:
for lazy update, it is related to this
https://github.com/shogun-toolbox/shogun/issues/2838

the fact is in SGObject
The hash value is `changed` if `parameter_hash_changed()` is called even when values of variables in the object remain `unchanged`.

for lazy update, I need 
No matter how many times  `parameter_hash_changed()` is called, the hash value is `unchanged` if values of variables in the object remain `unchanged`.

yorkerlin:
@karlnapf 
the idea using lazy update is at https://github.com/shogun-toolbox/shogun/pull/2839

I close this PR.

karlnapf:
Ok, please do the lazy one after the bugfix. We dont want users to need to be explicit about this

Issue #2838, opened 2015-05-05
hash in SGObject


yorkerlin:
@vigsterkr , @lisitsyn , @karlnapf 
there may be a bug.
If CSGObject::update_parameter_hash() is called more than once, CSGObject::parameter_hash_changed() always returns true even when variables remain the same.

for example
```
void CExactInferenceMethod::update()
{
	
	CInferenceMethod::update();
	update_chol();
	update_alpha();
	update_deriv();
	update_mean();
	update_cov();

	update_parameter_hash();
	SG_SPRINT("hash value1 %d \n", m_hash);  
	update_parameter_hash();
	SG_SPRINT("hash value2 %d \n", m_hash);  

}
```
`hash value1` and `hash value2` are different since
```
void CSGObject::update_parameter_hash()
{
	uint32_t carry=0;
	uint32_t length=0;

	get_parameter_incremental_hash(m_hash, carry, length); 
	m_hash=CHash::FinalizeIncrementalMurmurHash3(m_hash, carry, length);
}
```
where `m_hash` becomes `non-zero` after `update_parameter_hash()` is called

what is worse,
```
bool CSGObject::parameter_hash_changed()
{
	uint32_t hash=0;
	uint32_t carry=0;
	uint32_t length=0;

	get_parameter_incremental_hash(hash, carry, length); 
	hash=CHash::FinalizeIncrementalMurmurHash3(hash, carry, length);

	return (m_hash!=hash);
}
```
Note that `hash` is `zero`, which means if `update_parameter_hash()` is called more than once, `CSGObject::parameter_hash_changed()` always returns `true` even when variables remain the same.

yorkerlin:
I am not sure whether the hash value should be changed if `parameter_hash_changed()` is called even when `values` of variables in the object remain the same

yorkerlin:
if you think this is a bug, the fix is quite simple.

```
void CSGObject::update_parameter_hash()
{
    uint32_t carry=0;
    uint32_t length=0;

    //CHANGE m_hash to 0
    get_parameter_incremental_hash(0, carry, length);  
    m_hash=CHash::FinalizeIncrementalMurmurHash3(m_hash, carry, length);
}
```

karlnapf:
I think this is a bug! This should not be. Do unit tests break if you fix it?
Make sure to add unit tests for the hashing if you do fix it
Thanks!

yorkerlin:
this issue is solved.
I close this

Issue #2817, opened 2015-04-14
make model selection on GP more flexible


yorkerlin:
At default, inference methods in GP module will compute gradients wrt model parameters, which is not always necessary.
* Provide users an option to ask inference methods do not compute these gradients.
* (for now) Make the model selection framework work with `FITC inference` 
* (suspicious bug, some GP parameters are in natural domain while some gradients wrt parameters are in log domain)

yorkerlin:
@karlnapf 
will work on this issue once `notebooks` are done

karlnapf:
* I would actually go to move all GP parameters to log-domain. We can still interface and store them in standard domain, but all dervatives in log-domain would be better for many reasons
 * inducing point model selection would be good to have I agree
 * not computing all gradients definitely is a good thing to have.

 * I would like to add more efficient (GPU, parallel) ways to compute gradients. A unified way to speed up is needed rather than all this ad-hoc code. What are your thoughts here?

yorkerlin:
@karlnapf 
Currently, `pthread` is used to parallelly compute the gradients.

yorkerlin:
about moving all GP parameters to log-domain:
* not all parameters are positive (for GaussianARDKernel, where weights are FULL_Matrix)
* many existing parameters in Shogun are in standard domain.

It will break many codes if we change existing  positive parameters in Shogun from standard domain to log domain.
My idea is (for example in GaussianKernel) to add a `log` flag 
```
CGaussianKernel::CGaussianKernel(int32_t size, float64_t w) 
```
becomes
```
CGaussianKernel::CGaussianKernel(int32_t size, float64_t w, bool is_log_domain=false) 
```
if flag `is_log_doamin` is true, the gradient will be first computed in standard domain (the existing code) and then transformed in log domain.

yorkerlin:
@karlnapf 
Let me know if you are agree with my idea.

yorkerlin:
@karlnapf 
the existing model selection (`CGradientModelSelection`) does not offer gradient transformation, which means there is a bug if a parameter and its gradient are in different domain.

yorkerlin:
@karlnapf 
Any feedback on the model selection

karlnapf:
I usually prefer general solutions that do things automatically to those that require extra arugments and thinking by the user. Users should not be able to decide whether a parameter is in log-domain or not, but we should do the appropriate thing automatically (in a transparent way)

[...]

PR #2839, opened 2015-05-05
users can disable the gradient computation [new]

yorkerlin:
@karlnapf 
take a look at this

### DIFF
--- a/src/shogun/base/SGObject.cpp
+++ b/src/shogun/base/SGObject.cpp
@@ -254,6 +254,7 @@ void CSGObject::update_parameter_hash()
 	uint32_t carry=0;
 	uint32_t length=0;
 
+	m_hash=0;
 	get_parameter_incremental_hash(m_hash, carry, length);
 	m_hash=CHash::FinalizeIncrementalMurmurHash3(m_hash, carry, length);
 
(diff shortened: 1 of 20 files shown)


src/shogun/base/SGObject.cpp
```diff
@@ -254,6 +254,7 @@ void CSGObject::update_parameter_hash()
 	uint32_t carry=0;
 	uint32_t length=0;
 
+	m_hash=0;
```
yorkerlin: @karlnapf 
the bug is mentioned at
https://github.com/shogun-toolbox/shogun/issues/2838

karlnapf: please send a seperate PR for this fix, and add unit tests that make sure the rationale is explicit

karlnapf:
How does this work from the user perspective now?

yorkerlin:
ref
https://github.com/shogun-toolbox/shogun/issues/2817

yorkerlin:
@karlnapf 
the gradient computation will only be activated when any of the following things:
* gradients 
* posterior mean/cov  
are required

karlnapf:
Cool thats perfpect.

As said, could you send a seperate PR with the bugfix and a unit test for the hash behavious? thanks!

yorkerlin:
@karlnapf 
the unit tests are added.

karlnapf:
Great thanks!

PR #2842, opened 2015-05-07
bug fix

yorkerlin:
@vigsterkr 
this patch should fix the bug in `LinearARDKernel`


karlnapf:
yep

karlnapf:
thanks for that
@vigsterkr does that do it?

Issue #2752, opened 2015-03-01
Use nested namespaces that resembles directory structure


lisitsyn:
Currently it is very hard to navigate through documentation and code completion due to flat namespace structure. I am pretty sure we should have nested namespaces to help with that.

```
shogun::features::DenseFeatures
```

@karlnapf @vigsterkr @sonney2k @iglesias what do you think?

vigsterkr:
i would go for it! :)

iglesias:
I like the idea too.

lisitsyn:
@karlnapf I guess if like it too we should go for it ;)

karlnapf:
Oh yes, please please +1 +1

AND do a little clean up on the fly, some of the structures do not make too much sense

ajaybhat:
Hi, I want to contribute to Shogun to sharpen my C++ and ML skills. What is the status of this issue? Can I take it up? 
> AND do a little clean up on the fly, some of the structures do not make too much sense
Can you give some examples of these, so I can include them in a PR?

lisitsyn:
Hey @ajaybhat, sure, feel free to take this over.

I am not sure about structures to clean up, but I think you would notice some during namespaces moving.

karlnapf:
Good to see you are one this!

PR #2847, opened 2015-05-27
Use Nested Namespaces resembling directory structure

ajaybhat:
This is a fix for #2752 
I have started with the <code>shogun.distance</code> package.
There is a lot of refactoring to be done, however I'd like to know if this is the way to go.

@lisitsyn @karlnapf

### DIFF
--- a/src/shogun/distance/AttenuatedEuclideanDistance.h
+++ b/src/shogun/distance/AttenuatedEuclideanDistance.h
@@ -17,7 +17,9 @@
 #include <shogun/distance/RealDistance.h>
 #include <shogun/features/DenseFeatures.h>
 
-namespace shogun
+using namespace shogun;
+
+namespace distance
 {
 /** @brief class AttenuatedEuclideanDistance
  *
--- a/src/shogun/clustering/Hierarchical.cpp
+++ b/src/shogun/clustering/Hierarchical.cpp
@@ -37,7 +37,7 @@ CHierarchical::CHierarchical()
 {
 }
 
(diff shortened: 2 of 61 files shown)


karlnapf:
@lisitsyn has to comment here. I am not sure what is the most elegant way to go ahead here

lisitsyn:
Hmm I am so sorry I missed this PR. Gonna glance over it now

src/shogun/distance/AttenuatedEuclideanDistance.h
```diff
@@ -17,7 +17,9 @@
 #include <shogun/distance/RealDistance.h>
 #include <shogun/features/DenseFeatures.h>
 
-namespace shogun
+using namespace shogun;
```
lisitsyn: Here and everywhere when we do ```using namespace``` and then declare namespace, we don't do it nested. I think it should be like 
```namespace shogun
{
  namespace distance
  {```

ajaybhat: @lisitsyn Is use of the scope resolution operator <code>::</code> for specifying type of the fields of the different classes okay or is it something that you discourage?

lisitsyn: I mean we better have namespace `shogun::distance` but currently it is just `distance`

karlnapf:
Any updates here?

karlnapf:
closing as inactive

Issue #2857, opened 2015-06-17
add a new features for subsamples


yorkerlin:
@karlnapf 
Shogun has a implementation of `CDenseSubsetFeatures`, which uses a subset of `features`.

I need another `features` class, which uses a subset of `samples`. The `features` class is used to create a mini batch, which is a subset of samples.
Please let me know if there exist such features class for subset of samples. If not, I will add my implementation.

yorkerlin:
`CDenseFeatures` has a `copy_subset(SGVector<index_t> indices)` method. However, I do not want to copy the subset.

yorkerlin:
BTW,  in `features` classes,
the following methods 
* get_num_vectors()
* get_num_features()
* get_dim_feature_space()
are confusing.

For `denseFeatures`  
* get_num_vectors() is samples size
* get_num_features() == get_dim_feature_space() is feature size

Maybe it may be good to document the difference between `get_num_features()` and `get_dim_feature_space()`.

yorkerlin:
@karlnapf 
the  `CDenseSubSamplesFeatures` class is implemented.
I will send a PR soon.

karlnapf:
I dont understand why we need such subsample features?
We can just set the subset manually?

Is there anything that is not good about this? Convenience?

karlnapf:
PS I agree with you that the num_features and num_vectors notation is confusing.....

yorkerlin:
this issue is solved.
I closed this.

PR #2860, opened 2015-06-18
a new features class

yorkerlin:
@karlnapf


yorkerlin:
ref:
https://github.com/shogun-toolbox/shogun/issues/2857

karlnapf:
Again the question why we cannot just use the subset feature of DenseFeatures and just create a convenience class here?

yorkerlin:
@karlnapf 
The `DenseSubSamplesFeatures` class wraps `DenseFeatures` and the indices of a subset.
I think this class is the `convenience` class you mentioned.

There are reasons why the `DenseSubSamplesFeatures` class.
* memory efficiency  
** we do `not` need to copy the subset to create a new `DenseFeatures` class
* API design
```
class StochasticGPInference(CFeatures *) 
{
protected:
Minimizer* m_opt
};

class StreamingGPInference: public StochasticGPInference
{
}

class BatchGPInference(CFeatures *) : public StochasticGPInference
{
protected:
Scheduler* m_sch // generate indices of  a mini-batch to create DenseSubSamplesFeatures
}
```
 
Let me know your suggestions?

karlnapf:
No I agree then! Good!

PR #2868, opened 2015-07-11
Remove inline-functions option

lisitsyn:
Currently it requires `-finline-functions` which is missed on yosemite's clang.


lisitsyn:
@besser82 @wiking ping

lisitsyn:
Too late! ;)

Issue #2859, opened 2015-06-18
bug in CDenseSubsetFeatures


yorkerlin:
Please note that
in `CDenseSubsetFeatures` class

`template<class ST> class CDenseSubsetFeatures: public CDotFeatures`

```
    virtual EFeatureClass get_feature_class() const
    {
        return m_fea->get_feature_class(); // BUG!!!
    }
    CDenseFeatures<ST> *m_fea;
```

The bug comes from the get_feature_class()  method. 
The class is indeed `not` `CDenseFeatures` class. (eg, it does not have `get_feature_vector` method)
However, it claims that it is like `CDenseFeatures` class since the `get_feature_class()` method.

If we mix `CDenseFeatures` and  `CDenseSubsetFeatures`, there is a bug. 
the bug can be reproduced by

```
aa->dot(0,bb,0);
```

where aa is CDenseFeatures\*  and bb is CDenseSubsetFeatures \*  

Reason:
the `CDenseSubsetFeatures` does not the `get_feature_vector` method from `CDenseFeatures`.

See

```
template<class ST> float64_t CDenseFeatures<ST>::dot(int32_t vec_idx1, CDotFeatures* df,
        int32_t vec_idx2)
{
    ASSERT(df)
    ASSERT(df->get_feature_type() == get_feature_type())
    ASSERT(df->get_feature_class() == get_feature_class()) //BUG!!! CDenseSubsetFeatures passed the assert
    CDenseFeatures<ST>* sf = (CDenseFeatures<ST>*) df; //we cast CDenseSubsetFeatures  to CDenseFeatures 

    int32_t len1, len2;
    bool free1, free2;

    ST* vec1 = get_feature_vector(vec_idx1, len1, free1);
    ST* vec2 = sf->get_feature_vector(vec_idx2, len2, free2);

    float64_t result = CMath::dot(vec1, vec2, len1);

    free_feature_vector(vec1, vec_idx1, free1);
    sf->free_feature_vector(vec2, vec_idx2, free2);

    return result;
}
```

yorkerlin:
@karlnapf 
@lisitsyn 
please take a look

karlnapf:
Yes, bug.

Why do we even have this class? See my other comment?

karlnapf:
I think it should be removed

We should rather define masks for CFeatures in general, just like numpy does it

PR #2871, opened 2015-07-18
Compatible kernel computation

yorkerlin:
@karlnapf

### DIFF
--- a/src/shogun/features/DenseSubSamplesFeatures.h
+++ b/src/shogun/features/DenseSubSamplesFeatures.h
@@ -174,6 +174,25 @@ public:
 	 * @param iterator as returned by get_feature_iterator
 	 */
 	virtual void free_feature_iterator(void* iterator);
+
+
+	/** does this class support compatible computation bewteen difference classes?
+	 * for example, this->dot(rhs_prt),
+	 * can rhs_prt be an instance of a difference class?
+	 *
+	 * @return whether this class supports compatible computation
+	 */
+	virtual bool support_compatible_class() const {return true;}
+
+	/** Given a class in right hand side, does this class support compatible computation? 
+	 *
+	 * for example, is this->dot(rhs_prt) valid,
+	 * where rhs_prt is the class in right hand side 
+	 *
+	 * @param rhs, the class in right hand side 
+	 * @return whether this class supports compatible computation
+	 */
+	virtual bool get_feature_class_compatibility (EFeatureClass rhs) const;
 private:
 	/* init */
 	void init();
--- a/src/shogun/kernel/DotKernel.h
+++ b/src/shogun/kernel/DotKernel.h
@@ -68,8 +68,19 @@ class CDotKernel : public CKernel
 
 			ASSERT(l->has_property(FP_DOT))
 			ASSERT(r->has_property(FP_DOT))
-			ASSERT(l->get_feature_type() == r->get_feature_type())
-			ASSERT(l->get_feature_class() == r->get_feature_class())
+			ASSERT(l->get_feature_type() == r->get_feature_type());
+			if (l->support_compatible_class())
+			{
+				REQUIRE(l->get_feature_class_compatibility(r->get_feature_class()),
+					"Right hand side of features (%s) must be compatible with left hand side features (%s)\n",
+					l->get_name(), r->get_name());
+			}
+			else
+			{
+				REQUIRE(l->get_feature_class()==r->get_feature_class(),
+					"Right hand side of features (%s) must be compatible with left hand side features (%s)\n",
+					l->get_name(), r->get_name())
+			}
 
 			if ( ((CDotFeatures*) l)->get_dim_feature_space() != ((CDotFeatures*) r)->get_dim_feature_space() )
 			{
--- a/src/shogun/kernel/Kernel.cpp
+++ b/src/shogun/kernel/Kernel.cpp
@@ -106,7 +106,18 @@ bool CKernel::init(CFeatures* l, CFeatures* r)
 	REQUIRE(r, "CKernel::init(%p, %p): Right hand side features required!\n", l, r)
 
 	//make sure features are compatible
-	ASSERT(l->get_feature_class()==r->get_feature_class())
+	if (l->support_compatible_class())
+	{
+		REQUIRE(l->get_feature_class_compatibility(r->get_feature_class()),
+			"Right hand side of features (%s) must be compatible with left hand side features (%s)\n",
+			l->get_name(), r->get_name());
+	}
+	else
+	{
+		REQUIRE(l->get_feature_class()==r->get_feature_class(),
+			"Right hand side of features (%s) must be compatible with left hand side features (%s)\n",
+			l->get_name(), r->get_name())
+	}
 	ASSERT(l->get_feature_type()==r->get_feature_type())
 
 	//remove references to previous features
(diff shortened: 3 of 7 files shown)


src/shogun/kernel/DotKernel.h
```diff
@@ -68,8 +68,19 @@ class CDotKernel : public CKernel
 
 			ASSERT(l->has_property(FP_DOT))
 			ASSERT(r->has_property(FP_DOT))
-			ASSERT(l->get_feature_type() == r->get_feature_type())
-			ASSERT(l->get_feature_class() == r->get_feature_class())
```
yorkerlin: the old Kernel class only supports computation between instances in the same features class

src/shogun/kernel/DotKernel.h
```diff
@@ -68,8 +68,19 @@ class CDotKernel : public CKernel
...
+			ASSERT(l->get_feature_type() == r->get_feature_type());
+			if (l->support_compatible_class())
```
yorkerlin: the new Kernel class now can supports computation between difference features classes if the features in left hand side support compatible computation.

src/shogun/kernel/Kernel.cpp
```diff
@@ -106,7 +106,18 @@ bool CKernel::init(CFeatures* l, CFeatures* r)
 	REQUIRE(r, "CKernel::init(%p, %p): Right hand side features required!\n", l, r)
 
 	//make sure features are compatible
-	ASSERT(l->get_feature_class()==r->get_feature_class())
```
yorkerlin: see comments in DotKernel

yorkerlin:
We need compatible computation when stochastic sparse inference is added.
`CDenseFeatures<float_64> data_points;`
`CDenseFeatures<float_64> inducing_points;`
The reason is below:
* `Scheduler` class will generate a subset of `data_points` as `CDenseSubSamplesFeatures<float64_t> subsample_points`.  Note that `subsample_points` only stores indices of `data_points'
* we need to compute `kernel(subsample_points, inducing_points)`. Note that these two instance are in different  features classes.

src/shogun/features/DenseSubSamplesFeatures.h
```diff
@@ -174,6 +174,25 @@ template<class ST> class CDenseSubSamplesFeatures: public CDotFeatures
...
+
+	/** Given a class in right hand side, does this class support compatible computation? 
+	 *
+	 * for example, is this->dot(rhs_prt) valid,
+	 * where rhs_prt is the class in right hand side 
+	 *
+	 * @param rhs, the class in right hand side 
+	 * @return whether this class supports compatible computation
+	 */
+	virtual bool get_feature_class_compatibility (EFeatureClass rhs) const;
```
yorkerlin: this implementation also offer one way to deal with the issue https://github.com/shogun-toolbox/shogun/issues/2859

src/shogun/kernel/DotKernel.h
karlnapf: I like this!

src/shogun/features/DenseSubSamplesFeatures.h
```diff
@@ -174,6 +174,25 @@ template<class ST> class CDenseSubSamplesFeatures: public CDotFeatures
 	 * @param iterator as returned by get_feature_iterator
 	 */
 	virtual void free_feature_iterator(void* iterator);
+
+
+	/** does this class support compatible computation bewteen difference classes?
```
karlnapf: Start sentences with capital letters.... it's English after all ;)

src/shogun/features/DenseSubSamplesFeatures.h
```diff
@@ -174,6 +174,25 @@ template<class ST> class CDenseSubSamplesFeatures: public CDotFeatures
...
+	 * for example, this->dot(rhs_prt),
+	 * can rhs_prt be an instance of a difference class?
+	 *
+	 * @return whether this class supports compatible computation
+	 */
+	virtual bool support_compatible_class() const {return true;}
```
karlnapf: Why virtual?

karlnapf: Ah sorry, forget about this, needs to be virtual in fact

src/shogun/features/DenseSubSamplesFeatures.h
```diff
@@ -174,6 +174,25 @@ template<class ST> class CDenseSubSamplesFeatures: public CDotFeatures
...
+
+	/** Given a class in right hand side, does this class support compatible computation?
```
karlnapf: The documentation should be a bit more clear here.

Returns whether this instance supports dot products /whatever with the feature class specified in the argument

src/shogun/features/DenseSubSamplesFeatures.h
karlnapf: Can you elaborate?

src/shogun/kernel/DotKernel.h
karlnapf: I feel all those should be a REQUIRE rather than an assert since the user can make them fail

src/shogun/kernel/DotKernel.h
```diff
@@ -68,8 +68,19 @@ class CDotKernel : public CKernel
...
+			{
+				REQUIRE(l->get_feature_class_compatibility(r->get_feature_class()),
+					"Right hand side of features (%s) must be compatible with left hand side features (%s)\n",
```
karlnapf: Maybe put in a reference to the "get_feature_class_compatibility" method called? So that a user can check .... or if there is a bug, it can be identified more easily

tests/unit/features/DenseSubSamplesFeatures_unittest.cc
```diff
@@ -440,3 +440,91 @@ TEST(DenseSubSamplesFeatures, test6)
 	SG_UNREF(features_train2);
 	SG_UNREF(latent_features_train2);
 }
+
+TEST(DenseSubSamplesFeatures, test7)
+{
+	index_t n=6;
```
karlnapf: seems a bit large for a unit test....but well

tests/unit/features/DenseSubSamplesFeatures_unittest.cc
```diff
@@ -440,3 +440,91 @@ TEST(DenseSubSamplesFeatures, test6)
...
+
+	feat_train2(1,0)=0.5;
+	feat_train2(1,1)=5.17637;
+	feat_train2(1,2)=4.57765;
+
+	CDenseFeatures<float64_t>* features_train0=new CDenseFeatures<float64_t>(feat_train2);
+	CDenseFeatures<float64_t>* latent_features_train0=new CDenseFeatures<float64_t>(lat_feat_train);
+
+	SGVector<int32_t> idx1(n/2);
+	for(int i=0; i<n; i++)
```
karlnapf: we now have auto

@lisitsyn

tests/unit/features/DenseSubSamplesFeatures_unittest.cc
```diff
@@ -440,3 +440,91 @@ TEST(DenseSubSamplesFeatures, test6)
...
+	SG_REF(latent_features_train0);
+	kernel->init(features_train0, latent_features_train0);
+	SGMatrix<float64_t> res0=kernel->get_kernel_matrix();
+
+	SG_REF(features_train2);
+	SG_REF(latent_features_train0);
+	kernel->init(features_train2, latent_features_train0);
+	SGMatrix<float64_t> res1=kernel->get_kernel_matrix();
+
+	for(index_t i=0; i<res1.num_rows; i++)
```
karlnapf: What about having a Macro to compare matrices with EXPECT/ASSERT_EQUAL?

Where would we add this?
@lisitsyn do you have an opinion on this?

karlnapf:
Good stuff @yorkerlin I'll merge if travis is good

PR #2875, opened 2015-07-22
add base classes for minimizer

yorkerlin:
the base classes for sgd


yorkerlin:
@karlnapf 
I close this due to the conflict of one branch

PR #2878, opened 2015-07-28
Feature/gaussian distribution

Ialong:
New overloaded constructors and getter/setter methods (compiles fine)

### DIFF
--- a/src/shogun/distributions/classical/GaussianDistributionNew.cpp
+++ b/src/shogun/distributions/classical/GaussianDistributionNew.cpp
@@ -1,11 +1,38 @@
 /*
- * This program is free software; you can redistribute it and/or modify
- * it under the terms of the GNU General Public License as published by
- * the Free Software Foundation; either version 3 of the License, or
- * (at your option) any later version.
- *
+ * Copyright (c) The Shogun Machine Learning Toolbox
+ * Written (w) 2015 Alessandro Ialongo
  * Written (W) 2013 Heiko Strathmann
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * The views and conclusions contained in the software and documentation are those
+ * of the authors and should not be interpreted as representing official policies,
+ * either expressed or implied, of the Shogun Development Team.
+ *
  */
+
+
+// Do we need to initialise m_dimension or is it somehow done automatically?
+// How should we store (i.e. under which type) m_cov_type?
 
 #include <shogun/distributions/classical/GaussianDistributionNew.h>
 
@@ -31,22 +58,26 @@ CGaussianDistributionNew::CGaussianDistributionNew(SGVector<float64_t> mean,
 	m_mean=mean;
 
 	m_cov=SGMatrix<float64_t>(1,1);
-	m_cov(1,1)=cov;
+	m_cov(0,0)=cov;
+
+	m_cov_type="Float";
 }
 
 CGaussianDistributionNew::CGaussianDistributionNew(SGVector<float64_t> mean,
 		SGVector<float64_t> cov, bool cov_is_cholesky) :
 				CProbabilityDistribution(mean.vlen)
 {
-	REQUIRE(mean.vlen==cov.vlen, "Mean must have same dimension as "
-			"covariance, which is %dx%d, but is %d\n",
-			cov.vlen, cov.vlen, mean.vlen);
+	REQUIRE(mean.vlen==cov.vlen,
+			"Mean dimension (%d) and covariance dimension (%d) do not match\n",
+			mean.vlen, cov.vlen);
 
 	init();
 
 	m_mean=mean;
 
 	m_cov=cov.convert_to_matrix(cov,1,cov.vlen,false);
+
+	m_cov_type="Vector";
 }
 
 CGaussianDistributionNew::CGaussianDistributionNew(SGVector<float64_t> mean,
@@ -55,9 +86,9 @@ CGaussianDistributionNew::CGaussianDistributionNew(SGVector<float64_t> mean,
 {
 	REQUIRE(cov.num_rows==cov.num_cols, "Covariance must be square but is "
 			"%dx%d\n", cov.num_rows, cov.num_cols);
-	REQUIRE(mean.vlen==cov.num_cols, "Mean must have same dimension as "
-			"covariance, which is %dx%d, but is %d\n",
-			cov.num_rows, cov.num_cols, mean.vlen);
+	REQUIRE(mean.vlen==cov.num_cols,
+			"Mean dimension (%d) and covariance dimension (%d) do not match\n",
+			mean.vlen, cov.num_cols);
 
 	init();
 
@@ -65,28 +96,12 @@ CGaussianDistributionNew::CGaussianDistributionNew(SGVector<float64_t> mean,
 
 	if (!cov_is_cholesky)
 	{
-		Map<MatrixXd> eigen_cov(cov.matrix, cov.num_rows, cov.num_cols);
-		m_cov=SGMatrix<float64_t>(cov.num_rows, cov.num_cols);
-		Map<MatrixXd> eigen_L(m_cov.matrix, m_cov.num_rows, m_cov.num_cols);
-
-		/* compute cholesky */
-		LLT<MatrixXd> llt(eigen_cov);
-		if (llt.info()==NumericalIssue)
-		{
-			/* try to compute smalles eigenvalue for information */
-			SelfAdjointEigenSolver<MatrixXd> solver(eigen_cov);
-			if (solver.info() == Success)
-			{
-				VectorXd ev=solver.eigenvalues();
-				SG_ERROR("Error computing Cholesky of Gaussian's covariance. "
-						"Smallest Eigenvalue is %f.\n", ev[0]);
-			}
-		}
-
-		eigen_L=llt.matrixL();
+		compute_cholesky(cov);
 	}
 	else
 		m_cov=cov;
+
+	m_cov_type="Matrix";
 }
 
 CGaussianDistributionNew::~CGaussianDistributionNew()
@@ -102,17 +117,23 @@ void CGaussianDistributionNew::set_mean(SGVector<float64_t> mean)
 void CGaussianDistributionNew::set_cov(float64_t cov)
 {
 	m_cov=SGMatrix<float64_t>(1,1);
-	m_cov(1,1)=cov;
+	m_cov(0,0)=cov;
+
+	m_cov_type="Float";
 }
 
 void CGaussianDistributionNew::set_cov(SGVector<float64_t> cov)
 {
 	m_cov=cov.convert_to_matrix(cov,1,cov.vlen,false);
+
+	m_cov_type="Vector";
 }
 
 void CGaussianDistributionNew::set_cov(SGMatrix<float64_t> cov)
 {
 	m_cov=cov;
+
+	m_cov_type="Matrix";
 }
 
 SGVector<float64_t> CGaussianDistributionNew::get_mean() const
@@ -133,6 +154,11 @@ SGVector<float64_t> CGaussianDistributionNew::get_cov_diag() const
 SGMatrix<float64_t> CGaussianDistributionNew::get_cov_full() const
 {
 	return m_cov;
+}
+
+char* CGaussianDistributionNew::get_cov_type() const
+{
+	return m_cov_type;
 }
 
 SGMatrix<float64_t> CGaussianDistributionNew::sample(int32_t num_samples,
@@ -174,6 +200,31 @@ SGMatrix<float64_t> CGaussianDistributionNew::sample(int32_t num_samples,
 	eigen_samples.colwise()+=eigen_mean;
 
 	return samples;
+}
+
+void CGaussianDistributionNew::compute_cholesky(SGMatrix<float64_t> cov)
+{
+	Map<MatrixXd> eigen_cov(cov.matrix, cov.num_rows, cov.num_cols);
+	m_cov=SGMatrix<float64_t>(cov.num_rows, cov.num_cols);
+	Map<MatrixXd> eigen_L(m_cov.matrix, m_cov.num_rows, m_cov.num_cols);
+
+	/* compute cholesky */
+	LLT<MatrixXd> llt(eigen_cov);
+	if (llt.info()==NumericalIssue)
+	{
+		/* try to compute smallest eigenvalue for information */
+		SelfAdjointEigenSolver<MatrixXd> solver(eigen_cov);
+		if (solver.info() == Success)
+		{
+			VectorXd ev=solver.eigenvalues();
+			SG_ERROR("Error computing Cholesky of Gaussian's covariance. "
+					"Smallest Eigenvalue is %f.\n", ev[0]);
+		}
+		else
+			SG_ERROR("Error computing Cholesky of Gaussian's covariance. "
+					"Numerical Issue (could not compute eigenvalues)");
+	}
+	eigen_L = llt.matrixL();
 }
 
 SGVector<float64_t> CGaussianDistributionNew::log_pdf_multiple(SGMatrix<float64_t> samples) const
@@ -237,8 +288,10 @@ SGVector<float64_t> CGaussianDistributionNew::log_pdf_multiple(SGMatrix<float64_
 void CGaussianDistributionNew::init()
 {
 	SG_ADD(&m_mean, "mean", "Mean of the Gaussian.", MS_NOT_AVAILABLE);
-	SG_ADD(&m_cov, "L", "Lower factor of covariance matrix, "
-			"depending on the factorization type.", MS_NOT_AVAILABLE);
+	SG_ADD(&m_cov, "cov", "Covariance matrix, "
+			"can be float, SGVector, or SGMatrix (cholesky).", MS_NOT_AVAILABLE);
+//	SG_ADD(&m_cov_type, "cov_type", "Type of covariance matrix, "
+//				"can be 'Float', 'Vector', or 'Matrix'.", MS_NOT_AVAILABLE);
 }
 
 #endif // HAVE_EIGEN3
--- a/src/shogun/distributions/classical/GaussianDistributionNew.cpp
+++ b/src/shogun/distributions/classical/GaussianDistributionNew.cpp
@@ -153,7 +153,9 @@ float64_t CGaussianDistributionNew::get_cov_spherical() const
 SGVector<float64_t> CGaussianDistributionNew::get_cov_diag() const
 {
 	REQUIRE(m_cov_type==COV_DIAGONAL,"Covariance is not diagonal");
-	return * m_cov.get_column_sg_vector(0);
+	SGVector<float_64_t> cov_copy(m_dimension);
+	memcpy(m_cov.matrix,cov_copy.vector,sizeof(float64_t)*m_dimension);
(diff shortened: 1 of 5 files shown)


karlnapf:
Could you send the PR against the feature branch in shogun?

src/shogun/distributions/classical/GaussianDistributionNew.cpp
```diff
@@ -0,0 +1,244 @@
+/*
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Written (W) 2013 Heiko Strathmann
```
karlnapf: Could you put a BSD license here and add your name?
See any recent Shogun file for a template

src/shogun/distributions/classical/GaussianDistributionNew.h
```diff
@@ -0,0 +1,175 @@
+/*
+ * Copyright (c) The Shogun Machine Learning Toolbox
+ * Written (w) 2014 Wu Lin
+ * Written (W) 2013 Heiko Strathmann
+ * All rights reserved.
```
karlnapf: License like here

src/shogun/distributions/classical/GaussianDistributionNew.h
```diff
@@ -0,0 +1,175 @@
...
+	/** Constructor for which takes Gaussian mean and its covariance matrix.
+	 * It is also possible to pass a precomputed matrix factor of the specified
+	 * form. In this case, the factorization is not explicitly computed.
+	 *
+	 * @param mean mean of the Gaussian
+	 * @param cov covariance of the Gaussian, or covariance factor
+	 * @param cov_is_cholesky whether cov is a Cholesky factorization of the covariance or not
+	 * (default is false). If false, the Cholesky factorization is explicitly computed
+	 */
+	CGaussianDistributionNew(SGVector<float64_t> mean, float64_t cov,
```
karlnapf: three times the same?

src/shogun/distributions/classical/GaussianDistributionNew.h
```diff
@@ -0,0 +1,175 @@
...
+	CGaussianDistributionNew(SGVector<float64_t> mean, SGVector<float64_t> cov,
+			bool cov_is_cholesky=false);
+
+	CGaussianDistributionNew(SGVector<float64_t> mean, SGMatrix<float64_t> cov,
+			bool cov_is_cholesky=false);
+
+	/** Destructor */
+	virtual ~CGaussianDistributionNew();
+
+	void set_mean(SGVector<float64_t> cov);
```
karlnapf: parameter should be called "mean" not "cov"

src/shogun/distributions/classical/GaussianDistributionNew.h
```diff
@@ -0,0 +1,175 @@
...
+	void set_cov(float64_t cov);
+	void set_cov(SGVector<float64_t> cov);
+	void set_cov(SGMatrix<float64_t> cov);
+
+	SGVector<float64_t> get_mean() const;
+	float64_t get_cov_spherical() const;
+	SGVector<float64_t> get_cov_diag() const;
+	SGMatrix<float64_t> get_cov_full() const;
```
karlnapf: Good!
These now need documentation, but we can add that later

src/shogun/distributions/classical/GaussianDistributionNew.h
```diff
@@ -0,0 +1,175 @@
...
+private:
+
+	/** Initialses and registers parameters */
+	void init();
+
+protected:
+	/** Mean */
+	SGVector<float64_t> m_mean;
+
+	/** Lower factor of covariance matrix (depends on factorization type).
```
karlnapf: This should be updated and explain the different parametrisations possible here, and what the matrix contains in the case and which dimensions it has

src/shogun/distributions/classical/GaussianDistributionNew.h
karlnapf: Sorry just saw the different type

karlnapf: These need documentation eventually that explains what parametrisation it corresponds to

src/shogun/distributions/classical/GaussianDistributionNew.cpp
```diff
@@ -0,0 +1,244 @@
...
+CGaussianDistributionNew::CGaussianDistributionNew(SGVector<float64_t> mean,
+		float64_t cov, bool cov_is_cholesky) :
+				CProbabilityDistribution(mean.vlen)
+{
+	init();
+
+	m_mean=mean;
+
+	m_cov=SGMatrix<float64_t>(1,1);
+	m_cov(1,1)=cov;
```
karlnapf: Shogun starts indexing at 0

src/shogun/distributions/classical/GaussianDistributionNew.cpp
```diff
@@ -0,0 +1,244 @@
...
+}
+
+CGaussianDistributionNew::CGaussianDistributionNew(SGVector<float64_t> mean,
+		SGVector<float64_t> cov, bool cov_is_cholesky) :
+				CProbabilityDistribution(mean.vlen)
+{
+	REQUIRE(mean.vlen==cov.vlen, "Mean must have same dimension as "
```
karlnapf: Maybe we can improve the error message a bit to
"Mean dimension (%d) and covariance dimension (%d) do not match"

src/shogun/distributions/classical/GaussianDistributionNew.cpp
```diff
@@ -0,0 +1,244 @@
...
+	m_cov=cov.convert_to_matrix(cov,1,cov.vlen,false);
+}
+
+CGaussianDistributionNew::CGaussianDistributionNew(SGVector<float64_t> mean,
+		SGMatrix<float64_t> cov, bool cov_is_cholesky) :
+				CProbabilityDistribution(mean.vlen)
+{
+	REQUIRE(cov.num_rows==cov.num_cols, "Covariance must be square but is "
+			"%dx%d\n", cov.num_rows, cov.num_cols);
+	REQUIRE(mean.vlen==cov.num_cols, "Mean must have same dimension as "
```
karlnapf: see above for better error messages

src/shogun/distributions/classical/GaussianDistributionNew.cpp
```diff
@@ -0,0 +1,244 @@
...
+			"covariance, which is %dx%d, but is %d\n",
+			cov.num_rows, cov.num_cols, mean.vlen);
+
+	init();
+
+	m_mean=mean;
+
+	if (!cov_is_cholesky)
```
karlnapf: This should go to a private/protected method

src/shogun/distributions/classical/GaussianDistributionNew.cpp
```diff
@@ -0,0 +1,244 @@
...
+		/* compute cholesky */
+		LLT<MatrixXd> llt(eigen_cov);
+		if (llt.info()==NumericalIssue)
+		{
+			/* try to compute smalles eigenvalue for information */
+			SelfAdjointEigenSolver<MatrixXd> solver(eigen_cov);
+			if (solver.info() == Success)
+			{
+				VectorXd ev=solver.eigenvalues();
+				SG_ERROR("Error computing Cholesky of Gaussian's covariance. "
```
karlnapf: The error is only printed if Eigenvalue computation was a success. 
I suggest to print the smallest eigenvalue in a warning, and then throw the error afterwards, independent of the eigenvalue solver

src/shogun/distributions/classical/GaussianDistributionNew.h
```diff
@@ -0,0 +1,175 @@
...
+	 * Covariance (approximation) is given by \f$\Sigma=LL^T\f$ */
+	SGMatrix<float64_t> m_cov;
+};
```
karlnapf: need a new field here with the m_cov_type

src/shogun/distributions/classical/GaussianDistributionNew.cpp
```diff
@@ -0,0 +1,297 @@
...
+	eigen_result=eigen_result.array()+(log_det_part+const_part);
+
+	/* contains everything */
+	return result;
+
+}
+
+void CGaussianDistributionNew::init()
+{
+	SG_ADD(&m_mean, "mean", "Mean of the Gaussian.", MS_NOT_AVAILABLE);
```
karlnapf: m_dimension should be initialised to 0 here

src/shogun/distributions/classical/GaussianDistributionNew.cpp
```diff
@@ -0,0 +1,297 @@
...
+using namespace Eigen;
+
+CGaussianDistributionNew::CGaussianDistributionNew() : CProbabilityDistribution()
+{
+	init();
+}
+
+CGaussianDistributionNew::CGaussianDistributionNew(SGVector<float64_t> mean,
+		float64_t cov, bool cov_is_cholesky) :
+				CProbabilityDistribution(mean.vlen)
```
karlnapf: calling the base constructor here doesnt work for initialising the dimension.
This is since the init() method is called afterwards. So you need a m_dimension=mean.vlen statement after init();

src/shogun/distributions/classical/GaussianDistributionNew.cpp
```diff
@@ -0,0 +1,297 @@
...
+{
+	init();
+
+	m_mean=mean;
+
+	m_cov=SGMatrix<float64_t>(1,1);
+	m_cov(0,0)=cov;
+
+	m_cov_type="Float";
```
karlnapf: pls use enum, not string

src/shogun/distributions/classical/GaussianDistributionNew.h
```diff
@@ -0,0 +1,179 @@
...
+#include <shogun/lib/config.h>
+
+#ifdef HAVE_EIGEN3
+
+#include <shogun/distributions/classical/ProbabilityDistribution.h>
+#include <shogun/lib/SGVector.h>
+#include <shogun/mathematics/Math.h>
+
+namespace shogun
+{
```
karlnapf: put the enum ECovType here, see Gaussian.h

src/shogun/distributions/classical/GaussianDistributionNew.cpp
```diff
@@ -0,0 +1,297 @@
+/*
+ * Copyright (c) The Shogun Machine Learning Toolbox
+ * Written (w) 2015 Alessandro Ialongo
+ * Written (W) 2013 Heiko Strathmann
```
karlnapf: I think Wu Lin was also in here

src/shogun/distributions/classical/GaussianDistributionNew.cpp
```diff
@@ -0,0 +1,297 @@
...
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * The views and conclusions contained in the software and documentation are those
+ * of the authors and should not be interpreted as representing official policies,
+ * either expressed or implied, of the Shogun Development Team.
+ *
+ */
+
+
+// Do we need to initialise m_dimension or is it somehow done automatically?
```
karlnapf: see below

src/shogun/distributions/classical/GaussianDistributionNew.cpp
```diff
@@ -0,0 +1,297 @@
...
+
+	m_cov_type="Float";
+}
+
+CGaussianDistributionNew::CGaussianDistributionNew(SGVector<float64_t> mean,
+		SGVector<float64_t> cov, bool cov_is_cholesky) :
+				CProbabilityDistribution(mean.vlen)
+{
+	REQUIRE(mean.vlen==cov.vlen,
+			"Mean dimension (%d) and covariance dimension (%d) do not match\n",
```
karlnapf: minor: maybe "covariance (diagonal) dimension (%d)"?

src/shogun/distributions/classical/GaussianDistributionNew.cpp
```diff
@@ -0,0 +1,297 @@
...
+			mean.vlen, cov.vlen);
+
+	init();
+
+	m_mean=mean;
+
+	m_cov=cov.convert_to_matrix(cov,1,cov.vlen,false);
+
+	m_cov_type="Vector";
```
karlnapf: enum, and I think it should be called COV_DIAG

src/shogun/distributions/classical/GaussianDistributionNew.cpp
```diff
@@ -0,0 +1,297 @@
...
+}
+
+CGaussianDistributionNew::CGaussianDistributionNew(SGVector<float64_t> mean,
+		SGMatrix<float64_t> cov, bool cov_is_cholesky) :
+				CProbabilityDistribution(mean.vlen)
```
karlnapf: base constructor issues as above

src/shogun/distributions/classical/GaussianDistributionNew.cpp
karlnapf: exactly like that!

src/shogun/distributions/classical/GaussianDistributionNew.cpp
```diff
@@ -0,0 +1,297 @@
...
+}
+
+SGVector<float64_t> CGaussianDistributionNew::get_mean() const
+{
+	return m_mean;
+}
+
+float64_t CGaussianDistributionNew::get_cov_spherical() const
+{
+	return *m_cov.data();
```
karlnapf: minor: could you make this m_cov(0,0), its cleaner

src/shogun/distributions/classical/GaussianDistributionNew.cpp
```diff
@@ -0,0 +1,297 @@
...
+}
+
+SGVector<float64_t> CGaussianDistributionNew::get_cov_diag() const
+{
+	return m_cov.get_row_vector(0);
```
karlnapf: I think this copies the vector, 
http://shogun-toolbox.org/doc/en/latest/SGMatrix_8cpp_source.html#l01088
but it would be better to return the same memory (this should be documented in the getter doxygen)

Shogun stores vectors in column-major (fortran) format, which means you can extract subsequent columns of matrices without copying. So get_column_vector is your friend here:
http://shogun-toolbox.org/doc/en/latest/SGMatrix_8h_source.html#l00115
Just realised there is no version of it which returns SGVector, so you could add that for convenience. Ask me if unclear

Summary:
We dont want to copy here, but return a reference to the memory. And this should be stated in the documentation

src/shogun/distributions/classical/GaussianDistributionNew.cpp
karlnapf: Should throw an error if m_cov is not COV_SPHERICAL

src/shogun/distributions/classical/GaussianDistributionNew.cpp
karlnapf: Should throw an error if m_cov is not COV_DIAG

src/shogun/distributions/classical/GaussianDistributionNew.cpp
```diff
@@ -0,0 +1,297 @@
...
+}
+
+SGMatrix<float64_t> CGaussianDistributionNew::get_cov_full() const
+{
+	return m_cov;
```
karlnapf: Should throw an error if m_cov is not COV_FULL

This is wrong as the m_cov stores a cholesky factor. So you need to allocate a new matrix here, compute LL^T, and return that.  Also the documentation should state that this gives a copy.

I also suggest to add another methods get_cov_full_cholesky() which returns m_cov if m_cov_type == COV_FULL

src/shogun/distributions/classical/GaussianDistributionNew.cpp
```diff
@@ -0,0 +1,297 @@
...
+				" (%d) does not desired number of samples (%d)\n",
+				pre_samples.num_cols, num_samples);
+
+		samples=pre_samples;
+	}
+	else
+	{
+		/* allocate memory and sample from std normal */
+		samples=SGMatrix<float64_t>(m_dimension, num_samples);
+		for (index_t i=0; i<m_dimension*num_samples; ++i)
```
karlnapf: I think linalg can now do this, could you check?

src/shogun/distributions/classical/GaussianDistributionNew.cpp
```diff
@@ -0,0 +1,297 @@
...
+			samples.matrix[i]=sg_rand->std_normal_distrib();
+	}
+
+	/* map into desired Gaussian covariance */
+	Map<MatrixXd> eigen_samples(samples.matrix, samples.num_rows,
```
karlnapf: This is the way to sample if m_cov_type == COV_FULL.
The other  covariance types work differently, so you need to do a switch(m_cov_type) here and impleent the corresponding behaviour (which is computationally cheaper)

src/shogun/distributions/classical/GaussianDistributionNew.cpp
```diff
@@ -0,0 +1,297 @@
...
+	/* compute cholesky */
+	LLT<MatrixXd> llt(eigen_cov);
+	if (llt.info()==NumericalIssue)
+	{
+		/* try to compute smallest eigenvalue for information */
+		SelfAdjointEigenSolver<MatrixXd> solver(eigen_cov);
+		if (solver.info() == Success)
+		{
+			VectorXd ev=solver.eigenvalues();
+			SG_ERROR("Error computing Cholesky of Gaussian's covariance. "
```
karlnapf: This should be a warning, and only say 
"Smallest Eigenvalue is %f.\n", ev[0]"

src/shogun/distributions/classical/GaussianDistributionNew.cpp
```diff
@@ -0,0 +1,297 @@
...
+					"Smallest Eigenvalue is %f.\n", ev[0]);
+		}
+		else
```
karlnapf: no else here, the error should be thrown in any case

src/shogun/distributions/classical/GaussianDistributionNew.cpp
```diff
@@ -0,0 +1,297 @@
...
+	eigen_samples=eigen_L*eigen_samples;
+
+	/* add mean */
+	Map<VectorXd> eigen_mean(m_mean.vector, m_mean.vlen);
+	eigen_samples.colwise()+=eigen_mean;
+
+	return samples;
+}
+
+void CGaussianDistributionNew::compute_cholesky(SGMatrix<float64_t> cov)
```
karlnapf: Since this method does not return anything, I would call it:
"update_cholesky"

src/shogun/distributions/classical/GaussianDistributionNew.cpp
```diff
@@ -0,0 +1,297 @@
...
+			samples.num_cols);
+	REQUIRE(samples.num_rows==m_dimension, "Sample dimension (%d) does not match"
+			"Gaussian dimension (%d)\n", samples.num_rows, m_dimension);
+
+	/* for easier to read code */
+	index_t num_samples=samples.num_cols;
+
+	float64_t const_part=-0.5 * m_dimension * CMath::log(2 * CMath::PI);
+
+	/* determinant is product of diagonal elements of triangular matrix */
```
karlnapf: the whole method is only correct for m_cov_type == COV_FULL.

I suggest (as above) to add static methods to the class that compute things based on what the covariance is, and then add a switch over m_cov_type here and call the helper methods

src/shogun/distributions/classical/GaussianDistributionNew.cpp
```diff
@@ -0,0 +1,297 @@
...
+			samples.num_cols);
+	Map<VectorXd> eigen_mean(m_mean.vector, m_mean.vlen);
+
+	/* substract mean from samples (create copy) */
+	SGMatrix<float64_t> centred(m_dimension, num_samples);
+	Map<MatrixXd> eigen_centred(centred.matrix, centred.num_rows,
+			centred.num_cols);
+	for (index_t dim=0; dim<m_dimension; ++dim)
+	{
+		for (index_t sample_idx=0; sample_idx<num_samples; ++sample_idx)
```
karlnapf: Can be done with eigen

src/shogun/distributions/classical/GaussianDistributionNew.cpp
```diff
@@ -0,0 +1,297 @@
...
+	}
+
+	/* solve the linear system based on factorization */
+	MatrixXd solved=eigen_L.triangularView<Lower>().solve(eigen_centred);
+	solved=eigen_L.transpose().triangularView<Upper>().solve(solved);
+
+	/* one quadratic part x^T C^-1 x for each sample x */
+	SGVector<float64_t> result(num_samples);
+	Map<VectorXd> eigen_result(result.vector, result.vlen);
+	for (index_t i=0; i<num_samples; ++i)
```
karlnapf: I think this can be done with Eigen without a loop, using matrix expressions

karlnapf: or even more clever, with dot products and linalg

src/shogun/distributions/classical/GaussianDistributionNew.h
```diff
@@ -0,0 +1,179 @@
+/*
+ * Copyright (c) The Shogun Machine Learning Toolbox
+ * Written (w) 2015 Alessandro Ialongo
```
karlnapf: add WU

src/shogun/distributions/classical/GaussianDistributionNew.h
```diff
@@ -0,0 +1,179 @@
...
+	void set_cov(float64_t cov);
+	void set_cov(SGVector<float64_t> cov);
+	void set_cov(SGMatrix<float64_t> cov);
+
+	SGVector<float64_t> get_mean() const;
+	float64_t get_cov_spherical() const;
+	SGVector<float64_t> get_cov_diag() const;
+	SGMatrix<float64_t> get_cov_full() const;
+
+	/** Samples from the distribution multiple times
```
karlnapf: Computational costs should be stated here for differenc cov types

src/shogun/distributions/classical/GaussianDistributionNew.h
```diff
@@ -0,0 +1,179 @@
...
+	char* get_cov_type() const;
+
+	/** Computes the univariate pdf for one given sample.
+	 *
+	 * @param sample is a given sample
+	 * @param mu is the mean of univariate Normal distribution (default value is 0.0)
+	 * @param sigma2 is the variance of univariate Normal distribution (default value is 1.0)
+	 * @return the pdf of the distribution given the sample
+	 */
+	static float64_t univariate_log_pdf(float64_t sample, float64_t mu = 0.0, float64_t sigma2 = 1.0)
```
karlnapf: I think it would be good to add static methods for:

sampling:
 * full covariance

log-pdf:
 * full_covariance
 * diagonal covariance
 * spherical covariance

and then call these in the class methods

karlnapf:
I will close this one for now. Feel free to re-open and continue to work on any time. It still would be good to have this

PR #2881, opened 2015-08-07
enable CMake-policy CMP0056, if available

besser82:
Needed for hardening-flags in Fedora.  This policy will pass all compiler-flags to stuff like 'FIND_PACKAGE()'.  Introduced with CMake 3.2.0.

Shouldn't break any builds.


besser82:
Travis-failures are not related.  Merging…

PR #2884, opened 2015-08-15
added log1p in CMath

yorkerlin:


src/shogun/mathematics/Math.h
```diff
@@ -924,6 +924,19 @@ class CMath : public CSGObject
...
+		 *
+		 * The method is computed in a way that is accurate
+		 * even if the value of v is near zero
+		 *
+		 * @param v the argument
+		 * @return log base e of (1.0+v) or ln(1.0+v)
+		 */
+		static inline float64_t log1p(float64_t v)
+		{
+			return ::log1p(v);
```
lisitsyn: Am I right it is C++11 only?

yorkerlin: It seems the function in supported in C99
ref
http://www.cplusplus.com/reference/cmath/log1p/

yorkerlin: @lisitsyn 
Other useful functions can be added in CMath

http://www.johndcook.com/blog/2010/06/07/math-library-functions-that-seem-unnecessary/

lisitsyn: @yorkerlin yeah I mean will it compile w/o C99 or C++11?

karlnapf:
updates here?

karlnapf:
closing as inactive

PR #2874, opened 2015-07-20
Nlopt minimizer

yorkerlin:
@karlnapf 
This is a wrapped NLOpt minimizer based on the minimizer framework for GP

ref:
https://github.com/shogun-toolbox/shogun/issues/2853

PR #2886, opened 2015-08-16
bug fixed

yorkerlin:


besser82:
That should fix the bug mentioned in #2874.  Merging…

PR #2887, opened 2015-08-17
fixed bug when NLOPT is missing

yorkerlin:


yorkerlin:
Let's see whether this could fix the bug.
Merge

PR #2896, opened 2015-08-21
update comments

yorkerlin:


yorkerlin:
Since I only changed comments, I merge this.

Issue #2766, opened 2015-03-09
Rename all ```initialize``` methods


karlnapf:
They clash with ruby:

See #2765

This is an easy grep exercise

PR #2899, opened 2015-08-21
Fix issue 2766. Changed all 'initialize' methods names

erip:
Refactor methods with name `initialize` to a more meaningful name as to not interfere with Ruby naming.

### DIFF
--- a/doc/OpenCV_docs/OpenCV_NN_vs_Shogun_NN.md
+++ b/doc/OpenCV_docs/OpenCV_NN_vs_Shogun_NN.md
@@ -286,7 +286,7 @@ The hidden layer has ```10``` neurons and similarly the output layer has ```4```
 * Initialize the network. The input is nothing but the standard deviation of the gaussian which is used to randomly initialize the parameters. We chose ```0.1``` here.
 
 ```CPP
-    network->initialize(0.1);
+    network->initialize_neural_network(0.1);
 ```
 
 * Specify the training parameters if needed.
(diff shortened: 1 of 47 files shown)


erip:
This is so annoying... Travis can't handle the build, so it fails.

lisitsyn:
@erip hey, what's that? :)

erip:
@lisitsyn When Travis builds, some of the tests will pass. When I rebuild (no changes in code), other tests will pass.

yorkerlin:
@erip 
Please try not to do the `open and close` operations.
Travis will fail due to the memory issue.

You can use the following trick:

For example,
* In the first commit, all tests are passed except test 1 and test 2  are failed.
* then do the `git force update` trick without changing the code
* In the second commit, if test 1 and test 2 are passed, you may consider `all` tests  are passed even other tests are failed.

the `git force update` trick is:
```
git reset --soft "HEAD^"
git commit -m "you comments" (the `hash id` will be changed)
git push origin +your_branch (force to update your branch even without no change)
```

yorkerlin:
BTW, @lisitsyn 
It seems currently `test 3` always fail due to the memory issue.

erip:
Ah, sorry. I'm new to OSS, so I didn't know this. My apologies. `test 12` seems to fail always, and I've created an issue in rudix to address it. In this latest build, all tests passed apart from `test 12`.

yorkerlin:
ref
https://github.com/shogun-toolbox/shogun/issues/2766

erip:
Merge?

besser82:
No objections…  Merging…

PR #2902, opened 2015-08-23
added more updaters and learning rate for the minimizer framework

yorkerlin:
The idea of Adaptive momentum is
* to increase the `next` gradient if the sign of velocity and the sign of `current` gradient are agreed
* to discount the `next` gradient if the sign of velocity and the sign of `current` gradient are not agreed

### DIFF
--- a/src/shogun/optimization/DescendCorrection.h
+++ b/src/shogun/optimization/DescendCorrection.h
@@ -68,11 +68,12 @@ public:
 	 *
 	 * @param negative_descend_direction the negative descend direction
 	 * @param idx the index of the direction
+	 * @param delta update the change to correct descend direction
 	 * 
 	 * @return corrected descend direction
 	 */
 	virtual float64_t get_corrected_descend_direction(float64_t negative_descend_direction,
-		index_t idx)=0;
+		index_t idx, float64_t& delta)=0;
 
 	/** Update a context object to store mutable variables
 	 * used in descend update
--- a/src/shogun/optimization/DescendCorrection.h
+++ b/src/shogun/optimization/DescendCorrection.h
@@ -36,6 +36,20 @@
 #include <shogun/optimization/MinimizerContext.h>
 namespace shogun
 {
+#define IGNORE_IN_CLASSLIST
+#ifndef DOXYGEN_SHOULD_SKIP_THIS
+IGNORE_IN_CLASSLIST struct DescendPair
+{
+	DescendPair()
+	{
+		descend_direction=0.0;
+		delta=0.0;
+	}
+
+	float64_t descend_direction;
+	float64_t delta;
+};
+#endif
 /** @brief This is a base class for descend based correction method.
  *
  * The interfact will be used in DescendUpdaterWithCorrection::update_variable()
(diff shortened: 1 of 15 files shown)


src/shogun/optimization/DescendCorrection.h
```diff
@@ -68,11 +68,12 @@ class DescendCorrection
 	 *
 	 * @param negative_descend_direction the negative descend direction
 	 * @param idx the index of the direction
+	 * @param delta update the change to correct descend direction
 	 * 
 	 * @return corrected descend direction
 	 */
 	virtual float64_t get_corrected_descend_direction(float64_t negative_descend_direction,
-		index_t idx)=0;
+		index_t idx, float64_t& delta)=0;
```
lisitsyn: Is it necessary for delta to be changed through reference?

src/shogun/optimization/NesterovMomentumCorrection.h
```diff
@@ -92,15 +92,18 @@ class NesterovMomentumCorrection: public MomentumCorrection
...
 	 */
-	virtual float64_t get_corrected_descend_direction(float64_t gradient, index_t idx)
+	virtual float64_t get_corrected_descend_direction(float64_t gradient, index_t idx,
+		float64_t& delta)
 	{
 		REQUIRE(idx>=0 && idx<m_previous_descend_direction.vlen,"The index (%d) is invalid\n", idx);
 		float64_t tmp=m_weight*m_previous_descend_direction[idx];
 		m_previous_descend_direction[idx]=tmp-gradient;
-		return (1.0+m_weight)*m_previous_descend_direction[idx]-tmp;
+		delta=m_previous_descend_direction[idx];
```
yorkerlin: @lisitsyn 
Yes.
For `NesterovMomentumCorrection`, the `return` is different from `delta`.

src/shogun/optimization/DescendCorrection.h
yorkerlin: Yes.
If we allow to return a `vector`, we do not need to use reference.
However, the `delta` is not a `corrected_descend_direction`.

yorkerlin: The idea of Adaptive momentum is 
* increase the gradient if the sign of `velocity` and the sign of `gradient` are agreed
* discount the gradient  if the sign of `velocity` and the sign of `gradient` are not agreed

yorkerlin:
@lisitsyn 
Locally, momentum methods (including the adaptive momentum method) can be introduced to `RmsProp`, `Adam`, `Adadelta` and `adagrad`.

I have tested the result for `RmsProp` and `Adadelta`.

src/shogun/optimization/DescendCorrection.h
lisitsyn: Can we return pair of values?

yorkerlin:
@lisitsyn 
@vigsterkr 
It seems `CMap` (including `CStringMap`) is `not` serializable.

yorkerlin:
@lisitsyn 
Since `CMap` and `CStringMap` do not support serialization, I disable  serialization for `MinimizerContext`.

yorkerlin:
@lambday @lisitsyn 
Currently, most of classes in the minimizer framework are not subclasses of SGObject.
If in future we allow all these classes in the minimizer framework become  subclasses of SGObject,
the `serialization` issue will be solved.
If not, I have to use `CStringMap` (a subclass of `CMap`) for serialization. 
In that case, do you guys plan to support serialization for `CMap`?

yorkerlin:
Errors are not related to this PR.
I merge  this.

Issue #2904, opened 2015-08-24
issue in CMap and CHash


yorkerlin:
CMap use thes `CHash::MurmurHash3` algorithm
```
	int32_t CMap::hash(const K& key)
	{
		return CHash::MurmurHash3((uint8_t*)(&key), sizeof(key), 0xDEADBEEF) % hash_size;
	}
```

```
	const char* cc="may";
	std::cout<<CHash::MurmurHash3((uint8_t *)&cc, sizeof("may"), 1)<<std::endl;

	char *c1=SG_MALLOC(char, strlen("may")+1);
	strcpy(c1, "may");
	std::cout<<CHash::MurmurHash3((uint8_t *)&c1, sizeof(c1), 1)<<std::endl;
	
	char *c2=SG_MALLOC(char, strlen("may")+1);
	strcpy(c2, "may");
	std::cout<<CHash::MurmurHash3((uint8_t *)&c2, sizeof(c2), 1)<<std::endl;
```
I guess the function uses the memory address to generate a hash value.
The three outputs are different.
What is worse, the outputs are different every time I run the program.

This is not good for  `serialization` and `deserialization`. 
For example, `CMap<String, SGVector<float64_t> >` 
Any idea to use the `content` not the `memory address` to generate a hash value.

yorkerlin:
@lisitsyn 
@iglesias

yorkerlin:
However, the MD5 algorithm works fine
```
	unsigned char buff[17];
	const char* cc="may";
 
	CHash::MD5((unsigned char *)cc, strlen(cc), buff);
	buff[16]='\0';
	std::cout<<"["<<buff<<"]"<<std::endl;

	char *c1=SG_MALLOC(char, strlen("may")+1);
	strcpy(c1, "may");
	
	CHash::MD5((unsigned char *)c1, strlen("may"), buff);
	buff[16]='\0';
	std::cout<<"["<<buff<<"]"<<std::endl;

	char *c2=SG_MALLOC(char, strlen("may")+1);
	strcpy(c2, "may");
	 
	CHash::MD5((unsigned char *)c2, strlen("may"), buff);
	buff[16]='\0';
	std::cout<<"["<<buff<<"]"<<std::endl;
```
The outputs are the same.

yorkerlin:
A bug is CHash
```
	int32_t index_of(const K& key)
	{
		int32_t index=hash(key);
		CMapNode<K ,T>* result=chain_search(index, key);

[...]

Issue #2903, opened 2015-08-23
about the MinimizerContext (CMap/CStringMap serialization)


yorkerlin:
@lisitsyn 
Because most of classes in the minimizer framework are not subclasses of `SGObject` for now, we use MinimizerContext to store the mutable variables.

There will be many mutable variables when Adam, RmsProp and other updaters are added.
For example
```
SGVector<float64_t> mut_var1_for_adam
SGVector<float64_t> mut_var1_for_rmsprop
```

Option 1
Should I add these two vectors in data members of `MinimizerContext`?

Or

Option 2
We use a `chain` to manage these variables.

For example
* MinimizerContext is a head node
* Each component (eg,  Adam, RmsProp and other updaters ) creates a node and create an implicit  `SGObject` link list.

Or

Option 3
* create a `CMap<String, SGVector<float64_t> >` to store `SGVector<float64_t>` mutable variables. 
* create a `CMap<String, float64_t >` to store `float64_t` mutable variables. 
* Key (string) = ClassName::DataMemberName

yorkerlin:
I vote for the CMap option.
@lisitsyn 
Any suggestion or anther option ?

lisitsyn:
@yorkerlin yeah I like map as well

yorkerlin:
@lisitsyn 
CMap (including CStringMap) does not support serialization.

yorkerlin:
done

PR #2905, opened 2015-08-24
String map

yorkerlin:
https://github.com/shogun-toolbox/shogun/issues/2904

### DIFF
--- a/src/shogun/lib/Map.h
+++ b/src/shogun/lib/Map.h
@@ -292,7 +292,7 @@ private:
 	 */
 	int32_t hash(const K& key)
 	{
-		return CHash::MurmurHash3((uint8_t*)(&key), sizeof(key), 0xDEADBEEF) % hash_size;
+		return get_hash_value(key) % hash_size;
 	}
 
 	/** is free? */
@@ -314,7 +314,6 @@ private:
 		else
 		{
 			CMapNode<K, T>* current=hash_array[index];
-
 			do // iterating all items in the list
 			{
 				if (current->key==key)
@@ -441,6 +440,15 @@ private:
 	}
 
 protected:
+	/** Get the hash of a given key
+	 * @param key a given key
+	 * @return hash of the key
+	 */
+	virtual uint32_t get_hash_value(const K& key)
+	{
+		return CHash::MurmurHash3((uint8_t*)(&key), sizeof(key), 0xDEADBEEF);
+	}
+
 	/** whether SG_MALLOC or just malloc etc shall be used */
 	bool use_sg_mallocs;
 
--- a/src/shogun/lib/StringMap.h
+++ b/src/shogun/lib/StringMap.h
@@ -1,0 +1,72 @@
+/*
+ * Copyright (c) The Shogun Machine Learning Toolbox
+ * Written (w) 2015 Wu Lin
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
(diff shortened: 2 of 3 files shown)


yorkerlin:
@vigsterkr 
take a look at this.
this PR address two issues:
* the bug `(uint32_t) % (int32_t), it seems the compiler will automatically convert unit32_t to int32_t.`
* use the content instead of the memory address to generate a hash value

src/shogun/lib/Map.h
```diff
@@ -292,7 +292,8 @@ IGNORE_IN_CLASSLIST template<class K, class T> class CMap: public CSGObject
 	 */
 	int32_t hash(const K& key)
 	{
-		return CHash::MurmurHash3((uint8_t*)(&key), sizeof(key), 0xDEADBEEF) % hash_size;
+		int32_t hv=get_hash_value(key)% (uint32_t) hash_size;
```
yorkerlin: @vigsterkr 
The unit test will fail if int32_t `get_hash_value(key)%  hash_size;`

src/shogun/lib/Map.h
```diff
@@ -292,7 +292,7 @@ IGNORE_IN_CLASSLIST template<class K, class T> class CMap: public CSGObject
 	 */
 	int32_t hash(const K& key)
 	{
-		return CHash::MurmurHash3((uint8_t*)(&key), sizeof(key), 0xDEADBEEF) % hash_size;
+		return get_hash_value(key)% (uint32_t) hash_size;
```
yorkerlin: The unit test will fail if `int32_t get_hash_value(key)% hash_size;`

src/shogun/lib/StringMap.h
```diff
@@ -0,0 +1,72 @@
...
+
+protected:
+	/** Get the hash of a given string key
+	 * @param key a given string key
+	 * @return hash of the key
+	 */
+	virtual uint32_t get_hash_value(const std::string& key)
+	{
+		const char* k=key.c_str();
+		return CHash::MurmurHash3((uint8_t*)(k), key.length(), 0xDEADBEEF);
```
yorkerlin: I use the content instead of the memory address to generate a hash value

src/shogun/lib/Map.h
```diff
@@ -441,6 +440,15 @@ IGNORE_IN_CLASSLIST template<class K, class T> class CMap: public CSGObject
 	}
 
 protected:
+	/** Get the hash of a given key
+	 * @param key a given key
+	 * @return hash of the key
+	 */
+	virtual uint32_t get_hash_value(const K& key)
+	{
+		return CHash::MurmurHash3((uint8_t*)(&key), sizeof(key), 0xDEADBEEF);
```
iglesias: A comment on why we are using this value for the third argument would be helpful.

vigsterkr: @iglesias you mean `0xDEADBEEF` ? :) that's a good seed, no? at least easy to remember :)

src/shogun/lib/StringMap.h
```diff
@@ -0,0 +1,72 @@
...
+namespace shogun
+{
+
+#define IGNORE_IN_CLASSLIST
+IGNORE_IN_CLASSLIST template<class T> class CStringMap: public CMap<std::string, T>
+{
+public:
+	/** Custom constructor */
+	CStringMap(int32_t size=41, int32_t reserved=128, bool tracable=true)
+		:CMap<std::string, T>(size, reserved, tracable) {}
```
iglesias: Why does it make sense to have the specialisation for string explicitly in a new class?

tests/unit/lib/StringMap_unittest.cc
```diff
@@ -0,0 +1,104 @@
...
+
+using namespace shogun;
+
+TEST(StringMap, test1)
+{
+	CStringMap<SGVector<float64_t> > smap;
+
+	const char* name="stringmapunittests";
+	index_t len=strlen(name)+1;
+	char *c=SG_MALLOC(char, len);
```
iglesias: Can we use higher level language features instead of this explicit memory management?

src/shogun/lib/Map.h
iglesias: No idea it was just a seed :-) I suppose any value would work, then.

yorkerlin: @iglesias 
I just use the same seed from `CMap`

src/shogun/lib/StringMap.h
yorkerlin: @iglesias 
`CMap` uses the memory address/pointer to generate a hash value.
I need a Map use the content instead of the memory address to generate a hash value.
I need the StringMap for this:
https://github.com/shogun-toolbox/shogun/issues/2903

tests/unit/lib/StringMap_unittest.cc
yorkerlin: The `Map` use the memory address to generate a hash value while `StringMap` uses the content.
I need different memory addresses to test `StringMap`.

yorkerlin: I think `std::string key="name"` should work too.

yorkerlin:
@iglesias 
Let me know once you think it is good to merge

iglesias:
I only have the remark against using malloc and free. Feel free to merge if you feel it is right though.

tests/unit/lib/StringMap_unittest.cc
vigsterkr: yeah try to use `std:string` instead of simple `char*`
you can always convert it by `.c_str()`

yorkerlin:
errors are not related to this PR.
I merge this PR.

PR #2906, opened 2015-08-27
rename init() method in GP

yorkerlin:
I rename the method to keep the method in the GP modular consistent


yorkerlin:
Errors are not related to this PR.
I merge this.

yorkerlin:
it seems the renaming will cause a warning.
I revert this this PR.

PR #2913, opened 2015-08-31
support sparse penalty (eg, L1)

yorkerlin:


yorkerlin:
@lisitsyn 
This PR is about implementing truncated gradient for L1 penalty.
paper:
www.jmlr.org/papers/volume10/langford09a/langford09a.pdf
code:
https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/linear_model/sgd_fast.pyx

src/shogun/optimization/L1Penalty.h
```diff
@@ -0,0 +1,129 @@
...
+	/** Given the value of a target variable,
+	 * this method returns the penalty of the variable 
+	 *
+	 * @param variable value of the variable
+	 * @return penalty of the variable
+	 */
+	virtual float64_t get_penalty(float64_t variable) {return CMath::abs(variable);}
+
+	virtual float64_t get_penalty_gradient(float64_t variable,
+		float64_t gradient_of_variable) {return 0.0;}
```
yorkerlin: For smooth penalty, we return the gradient related to the penalty.
For non-smooth penalty, we can return a subgradient.

For sparse penalty (eg, L1), a subgradient does not guarantee sparsity.
Instead, we do the follwoing things:
* return 0 in  `get_penalty_gradient(...)`
* use `update_sparse_variable(...)` to enforece sparsity

yorkerlin: for elastic net (a combination of L1 and L2), 
we can the follwoing things:
* return gradient wrt `L2 penalty` in `get_penalty_gradient(...)`
* use update_sparse_variable(...) to enforce the `L1` sparsity constraint

src/shogun/optimization/L1Penalty.h
```diff
@@ -0,0 +1,129 @@
...
+
+	virtual void set_rounding_eplison(float64_t eplison)
+	{
+		REQUIRE(eplison>=0,"");
```
lisitsyn: Message please ;)

src/shogun/optimization/L1Penalty.h
lisitsyn: I see

lisitsyn:
@yorkerlin sorry I am not really into your pace, this one looks ok

tests/unit/optimization/StochasticMinimizers_unittest.cc
```diff
@@ -1190,4 +1193,325 @@ TEST(AdaptMomentumCorrection, test1)
...
+	opt2->set_number_passes(1);
+	opt2->set_learning_rate(rate2);
+	opt2->load_from_context(context);
+	delete context;
+
+	opt2->minimize();
+	//the loss in the reference program is log(1+exp(-y*w*x))
+	//However, the loss in our implementation is log(1+exp(y*w*x))
+	//
+	//result from sklearn.linear_model.SGDClassifier (v.0.16.1)
```
yorkerlin: @lisitsyn 
@karlnapf 
I think the minimizer framework is able to reproduce the same results (L1,L2, and elastic net)  used in sklearn (SGD_classifier).

If we scale up the framework (eg, extend to support `spare` vectors, parallel speedup (openMP ?), GPU speedup), we maybe able to good alternative to sklearn.

@lambday 
 Does the `linalg` support switch from CPU computation to GPU computation for `dense` vector ?

How about `sparse` vector?  Does the `linalg` support switch from CPU computation to GPU computation for `sparse` vector ?

tests/unit/optimization/StochasticMinimizers_unittest.cc
```diff
@@ -1190,4 +1193,325 @@ TEST(AdaptMomentumCorrection, test1)
...
+	delete updater2;
+	delete penalty_type2;
+	delete bb;
+}
+
+TEST(ElasticNetPenalty, test1)
+{
+	SGVector<float64_t> y(20);
+	SGMatrix<float64_t> x(2,20);
+	x(0,0)=-0.731271511775; x(1,0)=0.694867473874;
```
lisitsyn: Can we put this matrix outside of test so its not repeated?

yorkerlin: see the update

tests/unit/optimization/StochasticMinimizers_unittest.cc
```diff
@@ -356,18 +359,90 @@ bool ClassificationForTestCostFunction2::next_sample()
 	return true;
 }
 
-TEST(SGDMinimizer,test1)
+struct ClassificationCase
```
lisitsyn: That's better! I have naming suggestion though, it is usually called 'fixture'

tests/unit/optimization/StochasticMinimizers_unittest.cc
```diff
@@ -356,18 +359,90 @@ bool ClassificationForTestCostFunction2::next_sample()
...
+{
+	static SGVector<float64_t> y;
+	static SGMatrix<float64_t> x;
+	static void init();
+};
+
+SGVector<float64_t> RegressionCase::y=SGVector<float64_t>();
+SGMatrix<float64_t> RegressionCase::x=SGMatrix<float64_t>();
+
+void RegressionCase::init()
```
lisitsyn: Why static and init? 

I think simple constructor would be better so that it is not possible to miss initialization.

yorkerlin:
errors are not related to this PR.
I merge this

PR #2920, opened 2015-09-10
support proximal gradient methods as proximal penalty

yorkerlin:
consider the following problem
`min f(x)+h(x)`
where 
* f(x) is convex, differentiable, `domain of f(x)=R^n`  
* h(x) is closed, convex, possibly non-differentiable

Proximal gradient method:
at iteration k+1
* x^{k+1/2} = x^{k} - learning_rate * d_f(x^{k})
* x^{k+1} = prox_h (x^{k+1/2}) 

where `prox_h` is a proximal penalty for `h(x)`

for example, `h(x)=L1_weight * |x|_1`


yorkerlin:
errors are not related to this PR
merging

PR #2932, opened 2015-12-01
fix a typo

yorkerlin:
@iglesias

### DIFF
--- a/src/shogun/optimization/L1Penalty.h
+++ b/src/shogun/optimization/L1Penalty.h
@@ -112,11 +112,9 @@ public:
 	  if (CMath::abs(variable)<m_rounding_eplison)
 		  variable=0.0;
 	  return variable;
-		return 0;
 	}
 protected:
 	float64_t m_rounding_eplison;
-
 
 private:
 	void init()


iglesias:
Would be nice if you can also fix indentation starting in line 100. The first level of indentation of the method has two spaces, which does not correspond to the first level of the other methods in the same file.

src/shogun/optimization/L1Penalty.h
```diff
@@ -112,12 +112,10 @@ class L1Penalty: public SparsePenalty
 	  if (CMath::abs(variable)<m_rounding_eplison)
 		  variable=0.0;
 	  return variable;
-		return 0;
 	}
```
iglesias: Leave a newline also here if so.

iglesias:
Merging. Travis failures not caused by this change.

karlnapf:
Hi guys,

such patches should be labelled properly. Rather than "fix a typo", it should be "add forgotten return 0 statement"

It is impossible to sort things later otherwise, or track down bugs from commit logs in the buildbot

iglesias:
You are definitely right. Thanks for the heads up!

PR #2937, opened 2015-12-15

yorkerlin:


yorkerlin:
I only change the documents.
merging

PR #2940, opened 2015-12-15
attempt to fix csharp build via hiding implementation from SWIG

karlnapf:


vigsterkr:
no bueno! we should fix the typemaps in the swig definition of csharp

karlnapf:
sure, I agree, but this patch [here](https://github.com/shogun-toolbox/shogun/commit/7f15f5e1ca1fef6507e2a1ce800b9a21366d0240) [broke](http://buildbot.shogun-toolbox.org/builders/deb3%20-%20modular_interfaces/builds/2643/steps/csharp%20modular/logs/stdio) csharp modular.

I just try to get the build working again for the release. 
Happy to fix that properly after.

What is your suggestion with the typemap?

vigsterkr:
why after and not before? :)

karlnapf:
I don't know how to do it and nobody else does it.
Or can you prove me wrong? :)

vigsterkr:
well let's at least give it a go before just signing off of not doing it :)

karlnapf:
ok cool. It's one of the last things before we can release I think

karlnapf:
Also, the patch did not work. closing.

karlnapf:
@sonney2k told me this comes from having duplicate method signatures, btw

PR #2948, opened 2015-12-16

yorkerlin:


yorkerlin:
only docs are changed.
merging

karlnapf:
great! Getting there!

karlnapf:
Now you see why I always am so insisting on proper docs in PRs. We have to do it anyways and doing it before a release is quite a pain ;)

PR #2952, opened 2015-12-17
add modshogun.rb: make sure narray is loaded before modshogun.so

besser82:
Bugfix for some Ruby-installations, that require to have rubygems and narray loaded before modshogun.so


besser82:
Ruby testsuite is passing; Python-errors are unrelated.  Merging…

PR #2953, opened 2015-12-18
Revert "remove migration methods from all classes and blacklist depending libshogun examples"

besser82:
SerializationASCII heavily depends on migration-framework, so we cannot drop that easy, unfortunately.

This reverts commit 5af68edff0f7ea7100c142d280307a0e7acaf350.


karlnapf:
I don't think that ASCII serialisation heavily depends on migration framework.
I think there is another reason why things broke. Will have a look and then send the patch again.

karlnapf:
Before you merge this, give me a moment to take a look at the cause

karlnapf:
(in the plane back tomorrow)

karlnapf:
Should be fixed in #2955

besser82:
Going to merge #2955 after Travis is ready…

PR #2955, opened 2015-12-19
Fix serialisation segfaults

karlnapf:
* Load parameters properly again after having removed migration framework
 * more migration related code removals


karlnapf:
unit tests did not segfault, merging

Issue #2957, opened 2015-12-20
Make NeuralConvLayer use linalg convolution


sanuj:
- Implement convolution in linalg with `NATIVE` backend.
- Change `CConvolutionalFeatureMap::convolve` in `ConvolutionalFeatureMap.cpp` to use `linalg::convolve`

@karlnapf I'll send a PR soon. Tell me if I'm missing something.

karlnapf:
There is already something in place I saw. Better check that before you start....

karlnapf:
[here](http://www.shogun-)toolbox.org/doc/en/latest/structshogun_1_1linalg_1_1implementation_1_1convolve_3_01Backend_1_1EIGEN3_00_01Matrix_01_4.html

karlnapf:
Making the native backend use openmp would be an improvement

sanuj:
@karlnapf I had seen the `linalg/implementations/convolve.h` file and was talking about adding `NATIVE` backend to this.

`HAVE_LINALG_LIB` variable is not always set by cmake as you can see [here](https://github.com/shogun-toolbox/shogun/blob/develop/CMakeLists.txt#L752). Should I also take care of this by using `#ifdef HAVE_LINALG_LIB` and keeping the existing convolution code if `HAVE_LINALG_LIB` is not set?

sanuj:
`CConvolutionalFeatureMap.cpp` would look something like:
```
#ifdef HAVE_LINALG_LIB
#include <shogun/mathematics/linalg/linalg.h>

using namespace linalg;
#endif // HAVE_LINALG_LIB
...
...
...
#if defined HAVE_LINALG_LIB
    // use linalg::convolve
#else
    // The existing convolution code goes here
#endif // HAVE_LINALG_LIB
```

sanuj:
Why is [this code](https://github.com/shogun-toolbox/shogun/blob/d8eb73dd89f47e0da28f07163c4f635b96d0ec00/src/shogun/mathematics/linalg/internal/implementation/Convolve.h#L115) encapsulated in `#ifdef HAVE_EIGEN3 .... #endif`? It can also work without Eigen3.

karlnapf:
@lambday should answer here

karlnapf:
Any code from linalg should only be available if the linalg lib is enabled.

The code that you referenced indeed does not depend on eigen, but @lambday should get back to you about this...

PR #2961, opened 2015-12-23
Make ConvolutionalFeatureMap use linalg convolution

sanuj:
Will push more commits to this. All unit tests are passing. Yet to add `NATIVE` implementation to linalg convolution. #2957

### DIFF
--- a/src/shogun/neuralnets/ConvolutionalFeatureMap.cpp
+++ b/src/shogun/neuralnets/ConvolutionalFeatureMap.cpp
@@ -37,6 +37,10 @@
 #include <shogun/lib/SGVector.h>
 #include <shogun/lib/SGMatrix.h>
 #include <shogun/mathematics/Math.h>
+#ifdef HAVE_LINALG_LIB
+#include <shogun/mathematics/linalg/linalg.h>
+#endif // HAVE_LINALG_LIB
+
 
 using namespace shogun;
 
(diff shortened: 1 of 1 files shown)


src/shogun/neuralnets/ConvolutionalFeatureMap.cpp
```diff
@@ -283,6 +287,32 @@ void CConvolutionalFeatureMap::convolve(
 			outputs.matrix+i*outputs.num_rows + outputs_row_offset,
 			m_output_height, m_output_width, false);
 
+#if defined(HAVE_LINALG_LIB)
```
sanuj: @karlnapf This is one way to take care of `m_autoencoder_position` but is inefficient. A better way would be to edit linalg convolution to tackle `m_autoencoder_position` like it's done below in the existing code. If we do the other way then I'll have to edit `VIENNACL` implementation as well.

src/shogun/neuralnets/ConvolutionalFeatureMap.cpp
```diff
@@ -283,6 +287,32 @@ void CConvolutionalFeatureMap::convolve(
...
+		if(m_autoencoder_position == NLAP_NONE)
+		{
```
karlnapf: minor: we do not use ``` {} ``` for single line loops/if

src/shogun/neuralnets/ConvolutionalFeatureMap.cpp
karlnapf: I would go for the more efficient one.

src/shogun/neuralnets/ConvolutionalFeatureMap.cpp
```diff
@@ -309,6 +339,7 @@ void CConvolutionalFeatureMap::convolve(
 				result(res_y,res_x) = sum;
 			}
 		}
+#endif
```
karlnapf: So this makes our NNs dependent on linalg, right?
@lisitsyn do you object?

src/shogun/neuralnets/ConvolutionalFeatureMap.cpp
yorkerlin: Using gist for the wiki thing?

karlnapf: @yorkerlin ???

yorkerlin: Sorry. Wrong email

karlnapf:
This is way too old

PR #2966, opened 2016-01-12
fixed the typo about epsilon

yorkerlin:


iglesias:
lgtm. For next time maybe you can squash such commits into one (there is no need to have two separate commits for a change like this).

Is the error in the second Travis job unrelated?

yorkerlin:
the error is not related to this PR.
I merge this.

Issue #2967, opened 2016-01-12
Implementing rmsprop stochastic optimisation for neural nets


arasuarun:
A brief description of rmsprop, which is used in speeding up mini batch learning: http://www.cs.toronto.edu/~tijmen/csc321/slides/lecture_slides_lec6.pdf 
(3rd from last slide)

The “maths” part has already been written in /lib/optimisation/RmsPropUpdater.cpp and the training algorithm has to be implemented in the Neural Networks package

arasuarun:
The current plan I have is to add data members required for rmsProp to NeuralNetwork.cpp with default values such that RmsProp doesn't make any difference.
The parameters required for rmsprop are (i.e. creating an RmspropUpdate object) :
1. learning rate (already in NN.cpp)
2. epsilon
3. decay_factor

In this way, there will always be an RmsProp update object in the training function regardless of whether it is being used or not. Would this be okay or would another way be better?

arasuarun:
From the suggestions of @yorkerlin it makes more sense to have a general implementation, instead of just RmsProp. Hence, we can use AdaDeltaUpdater, AdaGradUpdater, RmsPropUpdater, etc... without changing the code in neural nets.

@iglesias @lisitsyn @yorkerlin My understanding so far is that they all derive from DescendUpdaterWithCorrection. The optimisation methods that are already there (and to come) have the same interface and (I think) the same data members. 
Please correct me if I'm wrong. Is there an even more general implementation possible?

iglesias:
It makes sense to me.

arasuarun:
So, Descend Updater is the base class for all these optimisation methods, which all have the "update_variable" method with the same parameters.
I have a doubt about the implementation: how exactly would the user choose which optimisation method he wants? Is it like we have a DescendUpdater *object as a data member in neuralnetwork.cpp* and the exact derived class can be chosen by constructing that? 
@yorkerlin

yorkerlin:
From user' view,  users only need to pass a string (enum class) to use
existing optimizers.
For implementation, we may use factory pattern or strategy pattern.
Any other ideas?

On Thursday, January 21, 2016, Arasu Arun <notifications@github.com> wrote:

> So, Descend Updater is the base class for all these optimisation methods,
> which all have the "update_variable" method with the same parameters.
> I have a doubt about the implementation: how exactly would the user choose
> which optimisation method he wants? Is it like we have a DescendUpdater *object
[...]

PR #2969, opened 2016-01-19
RmsProp stochastic optimization in Neural Nets

arasuarun:
From issue #2967 

Implemented RmsProp training algorithm from optimization/RmsPropUpdater.h to Neural Networks, in NeuralNetwork.cpp.
Default values of the hyper parameters are the standard default values: epsilon=1e-6 and decay_factor=0.9.

I also wrote a simple unit test to make sure rmsprop works, based off the regression unit test.

### DIFF
--- a/src/shogun/neuralnets/NeuralNetwork.cpp
+++ b/src/shogun/neuralnets/NeuralNetwork.cpp
@@ -34,6 +34,7 @@
 #include <shogun/neuralnets/NeuralNetwork.h>
 #include <shogun/mathematics/Math.h>
 #include <shogun/optimization/lbfgs/lbfgs.h>
+#include <shogun/optimization/RmsPropUpdater.h>
 #include <shogun/features/DenseFeatures.h>
 #include <shogun/lib/DynamicObjectArray.h>
 #include <shogun/neuralnets/NeuralLayer.h>
@@ -265,6 +266,10 @@ bool CNeuralNetwork::train_gradient_descent(SGMatrix<float64_t> inputs,
 		"Gradient descent learning rate (%f) must be > 0\n", gd_learning_rate);
 	REQUIRE(gd_momentum>=0,
 		"Gradient descent momentum (%f) must be > 0\n", gd_momentum);
+  REQUIRE(rms_epsilon>=0,
+    "RmsProp epsilon (%f) must be non-negative\n", rms_epsilon);
+  REQUIRE(rms_decay_factor>=0 && rms_decay_factor<1.0,
+    "RmsProp decay factor (%f) must be in range [0,1)\n", rms_decay_factor);
 
 	int32_t training_set_size = inputs.num_cols;
 	if (gd_mini_batch_size==0) gd_mini_batch_size = training_set_size;
@@ -276,6 +281,12 @@ bool CNeuralNetwork::train_gradient_descent(SGMatrix<float64_t> inputs,
 	// needed for momentum
 	SGVector<float64_t> param_updates(n_param);
 	param_updates.zero();
+
+  // RmsPropUpdater object 
+  RmsPropUpdater rmsprop_updater;
+  rmsprop_updater.set_learning_rate(gd_learning_rate);
+  rmsprop_updater.set_epsilon(rms_epsilon);
+  rmsprop_updater.set_decay_factor(rms_decay_factor);
 
 	float64_t error_last_time = -1.0, error = -1.0;
 
@@ -309,7 +320,6 @@ bool CNeuralNetwork::train_gradient_descent(SGMatrix<float64_t> inputs,
 
 			float64_t e = compute_gradients(inputs_batch, targets_batch, gradients);
 
-
 			for (int32_t k=0; k<m_num_layers; k++)
 			{
 				SGVector<float64_t> layer_gradients = get_section(gradients, k);
@@ -331,9 +341,10 @@ bool CNeuralNetwork::train_gradient_descent(SGMatrix<float64_t> inputs,
 			{
 				param_updates[k] = gd_momentum*param_updates[k]
 						-alpha*gradients[k];
-
-				m_params[k] -= alpha*gradients[k];
 			}
+
+      // updating params
+      rmsprop_updater.update_variable(m_params,gradients,alpha);
 
 			if (error_last_time!=-1.0)
 			{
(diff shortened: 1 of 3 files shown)


arasuarun:
In the gradient_descent unit test, I set the values of epsilon and decay_factor as 1.0 and 0.99 respective. This ensures that rmsprop has minimal effect in learning and doesn't interfere.

Ideally, decay_factor should be 1.0, but that's not permitted. I will later send a PR proposing to change its range from [0,1) to [0,1], if you guys will allow that.

src/shogun/neuralnets/NeuralNetwork.cpp
```diff
@@ -265,6 +266,10 @@ bool CNeuralNetwork::train_gradient_descent(SGMatrix<float64_t> inputs,
 		"Gradient descent learning rate (%f) must be > 0\n", gd_learning_rate);
 	REQUIRE(gd_momentum>=0,
 		"Gradient descent momentum (%f) must be > 0\n", gd_momentum);
+  REQUIRE(rms_epsilon>=0,
```
lisitsyn: Here and through the changes, new code's indentation doesn't match original one.

arasuarun: I think all of those should be fixed now. My tab length in vim was too small and I didn't notice the tab/space issue.

src/shogun/neuralnets/NeuralNetwork.h
```diff
@@ -442,6 +442,18 @@ friend class CDeepBeliefNetwork;
 	 * default value is -1
 	 */
 	float64_t gd_error_damping_coeff;
+
+	/** Used in RmsProp stochastic optimization
```
iglesias: I **think** you need to use [this syntax](https://www.stack.nl/~dimitri/doxygen/manual/grouping.html#memgroup) to group the documentation of more than one member together.

src/shogun/neuralnets/NeuralNetwork.h
```diff
@@ -442,6 +442,18 @@ friend class CDeepBeliefNetwork;
...
 	 */
 	float64_t gd_error_damping_coeff;
+
+	/** Used in RmsProp stochastic optimization 
+	* RmsProp scales each gradient down by its running average.
+	* It then updates the parameters acc. to the learning_rate.
+	*     scale=decay_factor*scale+(1-decay_factor)*gradient*gradient
+	*     param=param-learning_rate*gradient/sqrt(scale+epsilon)
+	* @epsilon: default=1e-6
+	* @decay_factor: default=0.9
```
iglesias: You don't need the @'s here, I believe.

src/shogun/neuralnets/NeuralNetwork.h
```diff
@@ -442,6 +442,18 @@ friend class CDeepBeliefNetwork;
 	 * default value is -1
 	 */
 	float64_t gd_error_damping_coeff;
+
+	/** Used in RmsProp stochastic optimization 
+	* RmsProp scales each gradient down by its running average.
+	* It then updates the parameters acc. to the learning_rate.
+	*     scale=decay_factor*scale+(1-decay_factor)*gradient*gradient
```
iglesias: Can we get these equations with Latex? See for instance the doc of check_gradients in this file.

arasuarun: Okay. I'll fix all of those right away.

src/shogun/neuralnets/NeuralNetwork.cpp
```diff
@@ -264,7 +265,11 @@ bool CNeuralNetwork::train_gradient_descent(SGMatrix<float64_t> inputs,
 	REQUIRE(gd_learning_rate>0,
 		"Gradient descent learning rate (%f) must be > 0\n", gd_learning_rate);
 	REQUIRE(gd_momentum>=0,
-		"Gradient descent momentum (%f) must be > 0\n", gd_momentum);
+    		"Gradient descent momentum (%f) must be > 0\n", gd_momentum);
```
iglesias: I am not sure why the change in this line. Anyway, maybe we can fix the > to >= as the actual require is doing :-)

arasuarun: I actually wanted to change it. But since it was completely unrelated to what I was actually doing, I decided to save it for another PR.

iglesias:
If would be great if you can squash the indentation commits into the main one.

arasuarun:
I fixed up:
1. proper documentation for grouping members together
2. using Latex for the formulae
3. remove some unnecessary '@'s

I'm sorry for the trouble - it's my first PR and I was careless. 
I'll now get onto squashing the 5 commits into 1. 
Thanks :)

iglesias:
Thanks to you for getting back with the fixes swiftly :-)

src/shogun/neuralnets/NeuralNetwork.h
```diff
@@ -442,6 +442,20 @@ friend class CDeepBeliefNetwork;
 	 * default value is -1
 	 */
 	float64_t gd_error_damping_coeff;
+
+	//@{
```
lisitsyn: I am not sure about this thing - what is it for?

yorkerlin:
@lisitsyn @iglesias @arasuarun 
We should use an interface for optimizers instead of a specific optimizer (eg, `rmsprop`). 
Please consider the following case:
if users want to use ADAM/ADGRAD optimizer, it seems that we have to modify the code according to this PR. 
@arasuarun  correct me if I am wrong.

src/shogun/neuralnets/NeuralNetwork.h
arasuarun: Grouping similar members together. Both of those are only used for RmsProp.  https://www.stack.nl/%7Edimitri/doxygen/manual/grouping.html#memgroup

arasuarun:
@yorkerlin That's completely right. Now that you mention it, since they all derive from the class DescendUpdaterWithCorrection and have the same members/interface, it makes more sense to have a general implementation.

I raised #2967 some time back (and didn't get any response, so I went ahead with it :P ). I think we can continue the discussion there.

karlnapf:
news?

arasuarun:
Sorry for the late response @karlnapf, I missed your ping among the various the notifications. I now get a taste of how you guys must feel. :) 

Reg. this, considering we have PR #2973, this is no longer required. I'll go ahead and close this. Thanks.

PR #2971, opened 2016-01-24
WIP added DescendUpdater training to Neural Nets

arasuarun:
This implements a general DescendUpdater training in nn.cpp
It currently has no functionality to set the data members for the DescendUpdater object, so GD goes forth with the default values (the gradient_descent unit test is passed).

Is this implementation design correct? I'll modify it or complete it accordingly.

### DIFF
--- a/src/shogun/neuralnets/NeuralNetwork.cpp
+++ b/src/shogun/neuralnets/NeuralNetwork.cpp
@@ -34,6 +34,11 @@
 #include <shogun/neuralnets/NeuralNetwork.h>
 #include <shogun/mathematics/Math.h>
 #include <shogun/optimization/lbfgs/lbfgs.h>
+#include <shogun/optimization/DescendUpdater.h>
+#include <shogun/optimization/AdaDeltaUpdater.h>
+#include <shogun/optimization/AdaGradUpdater.h>
+#include <shogun/optimization/GradientDescendUpdater.h>
+#include <shogun/optimization/RmsPropUpdater.h>
 #include <shogun/features/DenseFeatures.h>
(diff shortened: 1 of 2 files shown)


arasuarun:
Some thoughts:
* if we had parameterised constructors for all DescendUpdater objects, it'll be easy to set their parameters. But now, it will require creating a temporary object, more lines of code and a **static_cast** (that's why I didn't write it down, yet)
* We need to include all the DescendUpdater subclass's headers in NN.cpp. Is there any way we can avoid that? 
* I wrote the following line in NN.h. Is this fine or is it not okay?
```
class DescendUpdater;
```

arasuarun:
@yorkerlin?

src/shogun/neuralnets/NeuralNetwork.cpp
```diff
@@ -354,6 +366,27 @@ bool CNeuralNetwork::train_gradient_descent(SGMatrix<float64_t> inputs,
 	return true;
 }
 
+void CNeuralNetwork::initialize_descend_updater()
+{
+	switch(descend_updater_method)
+	{
+		case 0:
```
karlnapf: minor: Can you use the enum definitions here please?

arasuarun:
Something got messed up and it wouldn't allow me to push. I ended up accidentally deleting my branch and closing this PR. I opened up an identical one here: #2973 
I'm really really sorry about the mess. Won't happen again. 
Thanks.

src/shogun/neuralnets/NeuralNetwork.cpp
arasuarun: Done. It's on #2973

PR #2972, opened 2016-01-25
Added parameterized constructors to DescendUpdater derived classes

arasuarun:
Added parameterised constructors to the following classes:
* AdaDeltaUpdater
* AdaGradUpdater
* RmsPropUpdater

This will make initialising those objects in other classes easier. Another advantage of avoiding using the *set_xyz* functions explicitly is that they are not present in the DU class and hence, we can work with just a DU object that is actually cast from a subclass of it.
(I'm not sure if there is a way around it even without a param ctor)

I modified some of the unit tests to use the new constructors for testing.

### DIFF
--- a/src/shogun/optimization/AdaDeltaUpdater.cpp
+++ b/src/shogun/optimization/AdaDeltaUpdater.cpp
@@ -39,6 +39,15 @@ AdaDeltaUpdater::AdaDeltaUpdater()
 	init();
 }
 
+AdaDeltaUpdater::AdaDeltaUpdater(float64_t learning_rate,float64_t epsilon,float64_t decay_factor)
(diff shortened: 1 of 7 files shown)


karlnapf:
hey @yorkerlin 
Could you take care of this? We should merge such small things asap if generally ok

@arasuarun Lets wait for Wu to get  back, otherwise I can also merge it. But Wu is the expert

yorkerlin:
@arasuarun 
I will look at your PRs next weekend :)

yorkerlin:
BTW, I will add the ADAMUpdater next weekend since the method is stable now.

yorkerlin:
it seems errors are not related to this PR.
@arasuarun  thanks!

PR #2973, opened 2016-01-26
[WIP] Added DescendUpdater training to Neural Nets

arasuarun:
This implements a general DescendUpdater training in nn.cpp
It currently has no functionality to set the data members for the DescendUpdater object, so GD goes forth with the default values (the gradient_descent unit test is passed).

Is this implementation design correct? I'll modify it or complete it accordingly.

### DIFF
--- a/src/shogun/neuralnets/NeuralNetwork.cpp
+++ b/src/shogun/neuralnets/NeuralNetwork.cpp
@@ -35,10 +35,7 @@
 #include <shogun/mathematics/Math.h>
 #include <shogun/optimization/lbfgs/lbfgs.h>
 #include <shogun/optimization/DescendUpdater.h>
-#include <shogun/optimization/AdaDeltaUpdater.h>
-#include <shogun/optimization/AdaGradUpdater.h>
 #include <shogun/optimization/GradientDescendUpdater.h>
-#include <shogun/optimization/RmsPropUpdater.h>
 #include <shogun/features/DenseFeatures.h>
 #include <shogun/lib/DynamicObjectArray.h>
 #include <shogun/neuralnets/NeuralLayer.h>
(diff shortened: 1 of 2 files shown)


lisitsyn:
@arasuarun it is a good thing to do! Maybe we can drop the enum and just pass the object instead?

arasuarun:
@yorkerlin thought it would be better to do it this way #2967. 

If we used an object, the user would have to create one separately himself and then pass it while creating the neural net. It seems like extra effort on his part... or is there a better way to do it?

arasuarun:
And as of now, there is no way to set the data members of the DU object (hence WIP). I opened #2972 to make this easier.

arasuarun:
@yorkerlin What about this one? :)

src/shogun/neuralnets/NeuralNetwork.cpp
```diff
@@ -354,6 +366,27 @@ bool CNeuralNetwork::train_gradient_descent(SGMatrix<float64_t> inputs,
 	return true;
 }
 
+void CNeuralNetwork::initialize_descend_updater()
```
lisitsyn: Could we pass an object instead? It is much cleaner and we won't have any need to maintain the enums

lisitsyn: Oops, sorry, I already said that ;)

arasuarun:
Okay, I'll remove the enum and just keep a DescendUpdater object as a data member for NeuralNetworks. I think that'll easily wrap up this entire training algorithm.

arasuarun:
@lisitsyn @yorkerlin I've now changed it to the user setting an object instead of passing an enum.

arasuarun:
@yorkerlin @lisitsyn @anyone? :P

arasuarun:
Ping :)

karlnapf:
ping :)

arasuarun:
@yorkerlin @lisitsyn , please review. With the user passing the DescendUpdater object, it's not actually a [WIP] anymore, unless you guys would like to have anything else added.

PR #2978, opened 2016-02-07
updated docs to fix  python_modular warnings

yorkerlin:
@karlnapf 
this PR should fix all  python_modular warnings related to the optimization framework.


yorkerlin:
I merge this since only docs are changed.

PR #2979, opened 2016-02-08
remove python_modular warnings caused by obsolete doxygen fields

karlnapf:


karlnapf:
Merging as only doxygen comments were modifed

Issue #2987, opened 2016-02-14
k-Means is very slow, scales badly with k


kno10:
This variant of k-means (which is overly complicated) scales very badly.

Some observations:
* at the beginning of each iteration, the mean is recomputed (in `if (!fixed_centers)`). Afterwards, it is incrementally updated, but that data is never used. All use of `mus` in the second loop can be deleted.
* in each iteration, each vector will be retrieved three times, and freed three times, where one should be enough ( `get_feature_vector`).

But that still does not explain my runtime measurements (using 3.2.0 - I only now realized that the version preinstalled with Linux distributions is much older - albeit there does not appear to be any relevant difference):
k=10 5.91 seconds, k=100 137.009 seconds, k=1000 2409.55 seconds.

Where usually, the performance would increase by ~7.5 when increasing k by 10 (because of fewer iterations I guess), it increases by ~20 in Shogun with k, making it the slowest k-means I benchmarked except for flexclust (pure R), sklearn 0.14 (fixed since, 100x faster now), and scipy.cluster.vq. In these, the interpreter is probably to blame. For the record, the best K-means-variant I benchmarked on this data is 0.80, 3.76, 7.35; the best Lloyd implementation is 2.15, 42.16, 202.19... so Shogun is 12x slower than necessary, and better algorithms are over 300x faster.

Above observation (duplicate computation of centers) would explain a factor of 2 or 3 maybe. That would be enough to make k=10 competitive, but there is a superlinear cost with increasing k that I cannot discover in Shogun source code.

Note that #1263 suggests to drop legacy code like this completely.

kno10:
Update: I found another issue with this k-means. It is also initialized very badly. Every point is randomly assigned to a cluster, which will cause cluster centers to be very close to each other. The more common random initialization is to simply choose k random instances, and use these as centers.
The k-means++ initialization does not appear to be available using the command line interface.

kno10:
Ouch. It also chooses `n` random instances during each iteration, instead of using each one time? That only means it randomly ignores about 1/3rd of instances in each iteration:
`const int32_t Pat=CMath::random(0, XSize-1);`

vigsterkr:
@kno10 yeah indeed when i started some benchmarking of shogun, i've started with k-means and currently i do agree that's it's in a rather horrible state. :(

kno10:
[...]

PR #2988, opened 2016-02-14
Prune k-means, remove random-sampling during iteration

kno10:
1. The means are updated at the beginning of the loop. No need to update them inbetween (the result will not be used).

2. Instead of sampling n times from n objects, use all n objects once (can be reassigned only once anyway!)

3. `fixed_centers` will do the same every iteration, so only iterate once. This codepath should be removed completely!

Not fixed: A) no need to store `dists`. B) remove `fixed_centers` C) bad initialization D) allow kmeans++ from commandline.

*I have never compiled this code.* I don't use Shogun, but I have been looking at different k-means implementations and Shogun stood out as very slow, so I investigated the reasons for this. See also: #2987


karlnapf:
closing as never meant to be merged

PR #2989, opened 2016-02-17
fix gTest segfaults with GCC >= 6.0.0.

besser82:
In some cases gTest segfaults with GCC >= 6.0.0.  This is a dirty fix.
TODO: Update to gTest-release with proper support for GCC >= 6.0.0.
See:  https://github.com/google/googletest/issues/705


besser82:
Testsuite failures with Clang are not related.  Merging…

PR #2990, opened 2016-02-17
Make Java and CSharp install-dir configurable.

besser82:
add optional CMake-switches to configure:
* CSHARP_INSTALL_DIR
* JAVA_INSTALL_DIR
* JNI_INSTALL_DIR

Comes in handy for distro-packaging…


besser82:
Travis is not needed for this…  No change in code nor behaviour of CMake.  Merging…

PR #2994, opened 2016-02-19
fix Wstrict-overflow

besser82:
as said in title.


besser82:
Build fails are not related…  Merging…

PR #3007, opened 2016-02-24
Remove 'using namespace std' from header files

yozw:


besser82:
Same goes here…  If Travis is fine with it, I'll merge this.

lisitsyn:
Travis failure is unrelated, flapping test.

PR #3009, opened 2016-02-25
Remove unnesecary dependencies translation rule on C# target. Fix slo…

sorig:
…w and fast parser to generate consistent output. Remove automatically generated files by PLY.


karlnapf:
Thanks a ton. Travis also agreed, so I will merge.

Could you have another look at the csharp json for the types?
But I guess any error will pop up as a csharp runtime error when the types are used wrongly.

PR #3010, opened 2016-02-26
run python meta example generator tests using nosetests

karlnapf:


karlnapf:
@vigsterkr let me know if this is appropriate and if so, hit merge

karlnapf:
Ok thanks. Will push directly then with a bunch of other changes

karlnapf:
ah sorry will do

PR #3012, opened 2016-02-26
added ruby to meta examples, cookbook, and test build

karlnapf:
fully working locally


karlnapf:
ruby built was green, merging

PR #3014, opened 2016-02-26
hacky draft of added cpp to meta examples and the test build.

karlnapf:


karlnapf:
This works locally, but needs some more love. I need comments by @sorig @vigsterkr @besser82

examples/meta/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
...
+# (not generated yet so have to fake filenames from META_EXAMPLES list)
+FOREACH(META_EXAMPLE ${META_EXAMPLES})
+    # assume a structure <target_language>/<category>/listing.sg
+	STRING(REGEX REPLACE ".*/(.*).sg" "\\1" EXAMPLE_NAME ${META_EXAMPLE})
+	STRING(REGEX REPLACE ".*/(.*/.*).sg" "\\1" EXAMPLE_NAME_WITH_DIR ${META_EXAMPLE})
+	STRING(REGEX REPLACE "/" "-" EXAMPLE_NAME_WITH_DIR ${EXAMPLE_NAME_WITH_DIR})
+    STRING(REGEX REPLACE ".*/(.*)/.*.sg" "\\1" EXAMPLE_REL_DIR ${META_EXAMPLE})
+
+	ADD_CUSTOM_COMMAND(OUTPUT ${EXAMPLE_NAME_WITH_DIR}
+			  COMMAND g++ -std=c++11 -I /home/heiko/git/shogun/shogun_develop/shogun/src -L /home/heiko/git/shogun/shogun_develop/shogun/build/src/shogun  -o ${CMAKE_CURRENT_BINARY_DIR}/${EXAMPLE_REL_DIR}/${EXAMPLE_NAME} ${CMAKE_CURRENT_BINARY_DIR}/${EXAMPLE_REL_DIR}/${EXAMPLE_NAME}.cpp -lshogun
```
karlnapf: @besser82 @vigsterkr how to do this in the proper was of using the g++ environment set in cmake?

examples/meta/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
...
+	LIST(APPEND CPP_EXAMPLES ${EXAMPLE_NAME_WITH_DIR})
+
+    # run test in source dir, to have access to "data" folder
+    # run class from binary dir though as it is generated there
+    # add class location to classpath to do that
+	add_test(generated_cpp-${EXAMPLE_NAME_WITH_DIR} ${CMAKE_CURRENT_BINARY_DIR}/${EXAMPLE_REL_DIR}/${EXAMPLE_NAME})
+	set_property(TEST generated_cpp-${EXAMPLE_NAME_WITH_DIR} PROPERTY
+		WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
+	LIST(APPEND INSTALL_EXAMPLES ${CMAKE_CURRENT_BINARY_DIR}/${EXAMPLE_REL_DIR}/${EXAMPLE_NAME})
+	# TODO: Need some way to have finished building libshogun.so before this
```
karlnapf: I need some way to ensure that linshogun.so is already linked when running this with ```make -j```. How?

examples/meta/generator/targets/cpp.json
```diff
@@ -0,0 +1,61 @@
+{
+    "Program": "#include <shogun/base/init.h>\n#include <shogun/io/SGIO.h>\n#include <shogun/features/DenseFeatures.h>\n#include <shogun/labels/MulticlassLabels.h>\n#include <shogun/multiclass/KNN.h>\n#include <shogun/io/CSVFile.h>\n#include <shogun/distance/EuclideanDistance.h>\n\n${dependencies}\n\nusing namespace shogun;\n\nint main(int, char*[]) {\n    init_shogun_with_defaults();\n\n$program\n    exit_shogun();\n    return 0;\n}\n",
```
karlnapf: @sorig you mentioned the explicit imports in python. We need the same thing for c++.
We have a problem here though: the directory structure of shogun. Basically, we somehow need to infer the include path from the type. Any ideas for doing this in an elegant way?

examples/meta/generator/targets/cpp.json
```diff
@@ -0,0 +1,61 @@
...
+        "AllDependencies": "$enumDependencies\n\n",
+        "EnumDependencies": "$enums",
+        "DependencyListElementClass": "$element",
+        "DependencyListElementEnum": "import static org.shogun.$type.$value;",
+        "DependencyListSeparator": "\n"
+    },
+    "Statement": "    $statement;\n",
+    "Comment": "    //$comment\n",
+    "Init": {
+        "Construct": "C$type* $name = new C$type($arguments);\n    SG_REF($name)",
```
karlnapf: I think we need to distribuish constructors for SGObject (with Cprefix) and other objects

examples/meta/generator/targets/cpp.json
```diff
@@ -0,0 +1,61 @@
...
+        "Copy": "$type $name = $expr"
```
karlnapf: Same here. I had to comment out the ```LongRealVector output = test_labels.get_values()``` below. What are your thoughts on this problem?

examples/meta/generator/targets/cpp.json
```diff
@@ -0,0 +1,61 @@
...
+        "BoolLiteral": {
+            "True": "true",
+            "False": "false"
+        },
+        "NumberLiteral": "$number",
+        "MethodCall": "$object->$method($arguments)",
+        "Identifier": "$identifier",
+        "Enum":"$value"
+    },
+    "Print": "SG_SPRINT($expr)",
```
karlnapf: This does not make any sense for cpp unfortunately, as the print functions do not take SGObjects or vectors.
We can of course just remove this, or define a print function for vectors only maybe?

examples/meta/src/classifier/knn.sg
```diff
@@ -17,9 +17,8 @@ KNN knn(3, distance, labels)
 
 #![train_and_apply]
 knn.train()
-Labels test_labels = knn.apply(feats_test)
```
karlnapf: This works via "copy" above, so I cannot use it as the C is not prefixed

examples/meta/generator/targets/cpp.json
```diff
@@ -0,0 +1,61 @@
...
+        "RealFeatures": "DenseFeatures<float64_t>",
+        "Default": "$type",
+        "bool": "boolean",
+        "string": "String",
+        "BoolVector": "bool*",
+        "CharVector": "char*",
+        "ByteVector": "byte*",
+        "WordVector": "word*",
+        "ShortVector": "short*",
+        "IntVector": "int*",
```
sorig: Why are these native types and not e.g. `SGVector<int32_t>` like the translation rules in `/src/interfaces/modular/Library.i`?

examples/meta/cpp/CMakeLists.txt
karlnapf: Just saying: add_executable does not work. The .cpp files do not exist when cmake is at this point. Maybe via some dependency construction this might be changed

examples/meta/generator/targets/cpp.json
sorig: Hmm these imports are annoying. We can define import statements in the `"Dependencies"` field, but I don't know where we'd get the path information from (like `"shogun/features/DenseFeatures.h"`). 

The meta-language just knows the name of the classes that are being used, not where they're located. Looks like we need are more complex class scraper that includes that info. Would that be fragile to files being moved around in different releases of shogun? What do you think?

examples/meta/generator/targets/cpp.json
sorig: Well the easiest thing for now would be to just add a comment telling users to take a look themselves at whatever we want to print in the examples.

If we want to make a print statement for vectors only, we need to some more complex analysis of the parsed programs so we keep track of what type each variable has. I think it's possible, but would of course require some thought and work. Would any other language need this feature?

examples/meta/generator/targets/cpp.json
sorig: Oh sorry, just saw your comment about this. The github UI is a bit confusing

examples/meta/cpp/CMakeLists.txt
lisitsyn: You need something like 
```
SET_SOURCE_FILES_PROPERTIES(${CMAKE_CURRENT_BINARY_DIR}/${EXAMPLE_REL_DIR}/${EXAMPLE_NAME}.cpp PROPERTIES GENERATED 1)
```

examples/meta/cpp/CMakeLists.txt
lisitsyn: I think this could be resolved once `add_executable` is used.

lisitsyn:
Can't commit to your repo, so use something like that:

```cmake
# add test case for each generated example
# (not generated yet so have to fake filenames from META_EXAMPLES list)
FOREACH(META_EXAMPLE ${META_EXAMPLES})
    # assume a structure <target_language>/<category>/listing.sg
        STRING(REGEX REPLACE ".*/(.*).sg" "\\1" EXAMPLE_NAME ${META_EXAMPLE})
        STRING(REGEX REPLACE ".*/(.*/.*).sg" "\\1" EXAMPLE_NAME_WITH_DIR ${META_EXAMPLE})
        STRING(REGEX REPLACE "/" "-" EXAMPLE_NAME_WITH_DIR ${EXAMPLE_NAME_WITH_DIR})
    STRING(REGEX REPLACE ".*/(.*)/.*.sg" "\\1" EXAMPLE_REL_DIR ${META_EXAMPLE})

    SET(GENERATED_CPP ${CMAKE_CURRENT_BINARY_DIR}/${EXAMPLE_REL_DIR}/${EXAMPLE_NAME}.cpp)
    SET(GENERATED_CPP_TARGET ${EXAMPLE_NAME_WITH_DIR}-meta-cpp)
    SET(GENERATED_CPP_TEST ${EXAMPLE_NAME_WITH_DIR}-meta-cpp-test)

    ADD_EXECUTABLE(${GENERATED_CPP_TARGET} ${GENERATED_CPP})
    ADD_DEPENDENCIES(${GENERATED_CPP_TARGET} meta_examples shogun)
    TARGET_LINK_LIBRARIES(${GENERATED_CPP_TARGET} shogun)
    SET_SOURCE_FILES_PROPERTIES(${GENERATED_CPP} PROPERTIES GENERATED 1)

    LIST(APPEND CPP_EXAMPLES ${GENERATED_CPP_TARGET})

    # run test in source dir, to have access to "data" folder
    # run class from binary dir though as it is generated there
    # add class location to classpath to do that
    ADD_TEST(NAME ${GENERATED_CPP_TEST}
             COMMAND ${CMAKE_CURRENT_BINARY_DIR}/${GENERATED_CPP_TARGET}
             WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
    LIST(APPEND INSTALL_EXAMPLES ${CMAKE_CURRENT_BINARY_DIR}/${GENERATED_CPP_TARGET})

ENDFOREACH()

add_custom_target(build_cpp_meta_examples ALL
                                 DEPENDS ${CPP_EXAMPLES}
                                 COMMENT "Compiled generated cpp examples")

INSTALL(PROGRAMS ${INSTALL_EXAMPLES}
        DESTINATION share/shogun/examples/libshogun
        COMPONENT libshogun-examples)
```

examples/meta/generator/targets/cpp.json
```diff
@@ -0,0 +1,61 @@
...
+        "AllDependencies": "$enumDependencies\n\n",
+        "EnumDependencies": "$enums",
+        "DependencyListElementClass": "$element",
+        "DependencyListElementEnum": "import static org.shogun.$type.$value;",
+        "DependencyListSeparator": "\n"
+    },
+    "Statement": "$statement;\n",
+    "Comment": "//$comment\n",
+    "Init": {
+        "Construct": "C$type* $name = new C$type($arguments);\nSG_REF($name)",
```
lisitsyn: Try adding `#include <shogun/lib/some.h>` and then using `auto $name = some<C$type>($arguments)` :wink:

examples/meta/generator/targets/cpp.json
```diff
@@ -0,0 +1,61 @@
...
+        "AllDependencies": "$enumDependencies\n\n",
+        "EnumDependencies": "$enums",
+        "DependencyListElementClass": "$element",
+        "DependencyListElementEnum": "import static org.shogun.$type.$value;",
+        "DependencyListSeparator": "\n"
+    },
+    "Statement": "$statement;\n",
+    "Comment": "//$comment\n",
+    "Init": {
+        "Construct": "auto $name = some<C$type>($arguments)",
```
karlnapf: @lisitsyn and away are the memory leaks! Good stuff I forgot about that. I do not need to SG_UNREF ever?

lisitsyn: @karlnapf probably yes. If there is any issue it should be possible to fix that.

examples/meta/generator/targets/cpp.json
karlnapf: No ok, I think we just remove the print.
Do you think we should remove it from the grammar all together?

examples/meta/generator/targets/cpp.json
karlnapf: Done in https://github.com/shogun-toolbox/shogun/commit/c220dac005387df3a32fd374198649053a57f1fe

PR #3016, opened 2016-02-27
autogenerate modshogun.rb with correct module-suffix

besser82:
As said in title.  Should not break anything.


besser82:
Travis is fine… Merging…

PR #3021, opened 2016-02-28
crude/incomplete/working draft of how multiple matrices/vectors can b…

karlnapf:
…e stored from meta examples


karlnapf:
merging as works locally, avoid the branch to diverge more

PR #3026, opened 2016-03-01
random initialization of cluster centroids

manantomar:
Instead of randomly allocating each feature vector to a cluster centroid
,it will be better to choose k random feature vectors and consider them
as the initial centroids so that they are well separated.


src/shogun/clustering/KMeans.cpp
```diff
@@ -80,32 +80,30 @@ void CKMeans::set_random_centers(SGVector<float64_t> weights_set, SGVector<int32
...
 	{
-		const int32_t Cl=CMath::random(0, k-1);
-		weights_set[Cl]+=1.0;
-		ClList[i]=Cl;
+		do
+		{
+			const int32_t Cl=CMath::random(0, XSize-1);
 
-		int32_t vlen=0;
+		} while(ClList[Cl]=Cl);
```
karlnapf: 1. Could you do this without a do-while?
2. Something is weird with the condition. You create random numbers until "hitting" something???
3. Please when changing things, make sure they are formatted according to our guidelines (even if the KMeans class itself is not)

src/shogun/clustering/KMeans.cpp
```diff
@@ -80,32 +80,30 @@ void CKMeans::set_random_centers(SGVector<float64_t> weights_set, SGVector<int32
...
-			mus.matrix[Cl*dimensions+j] += vec[j];
+			mus.matrix[i*dimensions+j] += vec[j];
 
 		lhs->free_feature_vector(vec, i, vfree);
+
 	}
 
 	SG_UNREF(lhs);
 
-	for (int32_t i=0; i<k; i++)
```
karlnapf: why this?

karlnapf:
Thanks for the patch!

But I am a bit confused whether this actually helps. Could you both explain and clean up?

manantomar:
So previously the loop creates random numbers from one to k and assigns
each value to the corresponding feature vector .However if the same number
is generated again the cluster centroid is set as the mean of both the
feature vectors.The problem with this might be that :

1.There is no guarantee that each centroid is initialized .

2.Since the mean(of randomly generated numbers which correspond to a
feature vector) is being generated ,it will concentrate most centroids to
the center of the data whereas we want them to be as separate as possible.

So instead if we choose k random points from the data set and assign them
as our centroids ,we can get well separated centers.

The do-while  just checks if we are generating distinct numbers or not.

I hope this makes better sense and is useful for the initialization .And I
will definitely try to follow the coding guideline from next time (sorry
for that one).

Thanks
Manan Tomar

karlnapf:
I see.

This is not a good way of choosing k in N without repetition. Better way is to create a permutation from 1,...,N and then select the first k values.

`CMath` has a method to permute, and `SGVector` a method `range_fill` which you can use for that.

A comment saying that "initialise the cluster centroids randomly among all data" would be helpful as well. Also make sure to update the class documentation if you change this.

Thanks and looking forward to the update

manantomar:
Also the ipython notebook for clustering does not explain how random initialization works.So ,should I add this method to it ?
Thanks for the feedback.
Hopefully the pr can be merged successfully. :)

karlnapf:
yes, it would be cool to add this in the notebook -- it is a nice and small update. Send it in a separate PR.

src/shogun/clustering/KMeans.cpp
```diff
@@ -77,35 +77,31 @@ void CKMeans::set_initial_centers(SGMatrix<float64_t> centers)
 
 void CKMeans::set_random_centers(SGVector<float64_t> weights_set, SGVector<int32_t> ClList, int32_t XSize)
 {
+	/*initialise the cluster centroids randomly among all data*/
```
karlnapf: we use ```/* comment */``` i.e. with spaces

src/shogun/clustering/KMeans.cpp
```diff
@@ -77,35 +77,31 @@ void CKMeans::set_initial_centers(SGMatrix<float64_t> centers)
...
 	CDenseFeatures<float64_t>* lhs=
 		CDenseFeatures<float64_t>::obtain_from_generic(distance->get_lhs());
 
-	for (int32_t i=0; i<XSize; i++)
+	SGVector<int32_t> Perm=SGVector<int32_t>(XSize);
```
karlnapf: we usually name variables with lower cases.
Also I suggest to call it random_inds

src/shogun/clustering/KMeans.cpp
```diff
@@ -77,35 +77,31 @@ void CKMeans::set_initial_centers(SGMatrix<float64_t> centers)
...
-		const int32_t Cl=CMath::random(0, k-1);
+		const int32_t Cl=Perm[i];
 		weights_set[Cl]+=1.0;
-		ClList[i]=Cl;
+		ClList[Cl]=Cl;
 
 		int32_t vlen=0;
 		bool vfree=false;
-		float64_t* vec=lhs->get_feature_vector(i, vlen, vfree);
+		float64_t* vec=lhs->get_feature_vector(Cl, vlen, vfree);
```
karlnapf: can you replace the get_feature_vector and free_feature_vector calls with the `SGVector` version. This is old code before we had `SGVector`, and the new version just works as ```SGVector<float64_t> vec = lhs->get_feature_vector(Cl)``` and doesnt require cleaning up afterwards

src/shogun/clustering/KMeans.cpp
```diff
@@ -77,35 +77,31 @@ void CKMeans::set_initial_centers(SGMatrix<float64_t> centers)
...
 
 		for (int32_t j=0; j<dimensions; j++)
-			mus.matrix[Cl*dimensions+j] += vec[j];
+			mus.matrix[i*dimensions+j] += vec[j];
```
karlnapf: SGMatrix has the () operator which you can give row and column, could you use this here? makes the code easier to read. You see you have to clean up other peoples mess ;)

src/shogun/clustering/KMeans.cpp
karlnapf: vfree can go then too

src/shogun/clustering/KMeans.cpp
```diff
@@ -77,35 +77,31 @@ void CKMeans::set_initial_centers(SGMatrix<float64_t> centers)
 
 void CKMeans::set_random_centers(SGVector<float64_t> weights_set, SGVector<int32_t> ClList, int32_t XSize)
 {
+	/*initialise the cluster centroids randomly among all data*/
 	CDenseFeatures<float64_t>* lhs=
```
karlnapf: we have a new feature in the code that does 
```auto lhs = ....```
You can use this here to make the code easier to read. It is mostly syntax sugar
@lisitsyn we can use this in Shogun right? Any checks needed....?

**Comment by karlnapf** [comment]
Getting there!

**Comment by karlnapf** [comment]
Oh btw, can you also double check the doxygen documentation of this method? And correct if necessary?

**Comment by manantomar** [comment]
Sorry for the delay.I have made the above changes and the doxygen documentation seems fine as it is to me.
Thanks for the feedback

File: src/shogun/clustering/KMeans.cpp [hunk 8]
```diff
@@ -77,35 +77,29 @@ void CKMeans::set_initial_centers(SGMatrix<float64_t> centers)
 
 void CKMeans::set_random_centers(SGVector<float64_t> weights_set, SGVector<int32_t> ClList, int32_t XSize)
 {
+	/* initialise the cluster centroids randomly among all data */
 	CDenseFeatures<float64_t>* lhs=
 		CDenseFeatures<float64_t>::obtain_from_generic(distance->get_lhs());
```
karlnapf: need q REQUIRE that this pointer is not NULL

karlnapf: See https://github.com/shogun-toolbox/shogun/wiki/Assertions

src/shogun/clustering/KMeans.cpp
```diff
@@ -77,35 +77,29 @@ void CKMeans::set_initial_centers(SGMatrix<float64_t> centers)
...
+		ClList[Cl]=Cl;
 
-		int32_t vlen=0;
-		bool vfree=false;
-		float64_t* vec=lhs->get_feature_vector(i, vlen, vfree);
+		SGVector<float64_t> vec=lhs->get_feature_vector(Cl);
 
 		for (int32_t j=0; j<dimensions; j++)
-			mus.matrix[Cl*dimensions+j] += vec[j];
+			mus.operator()(j,i) += vec[j];
```
karlnapf: ehm, could you use this operator properly? mu(j,i) ?

karlnapf:
Great.
Now what about adding a unit test for this random initialising method?
2 tests: It should check that all clusters are initialised to one of the data points, and that there are no 2 clusters that are the same
For a very small dataset (3 points) and small (nontrivial) k (2)

karlnapf:
Also, it looks like you broke something. 

https://travis-ci.org/shogun-toolbox/shogun/jobs/113943490#L3923

*Always* run tests locally before sending a PR. At least the libshogun and the python examples

manantomar:
Hi,
First of all sorry for all the extra commits and the time delay.
I have made the above changes and added a unit test for the same.
Also can you please give me some advice regarding how to run these builds locally (This might be a very silly question but i am quite stuck at this and any hint will be really helpful )
Thanks for the feedback and looking forward to successfully merging this PR. :)

src/shogun/clustering/KMeans.cpp
```diff
@@ -77,35 +77,30 @@ void CKMeans::set_initial_centers(SGMatrix<float64_t> centers)
 
 void CKMeans::set_random_centers(SGVector<float64_t> weights_set, SGVector<int32_t> ClList, int32_t XSize)
 {
+	/* initialise the cluster centroids randomly among all data */
 	CDenseFeatures<float64_t>* lhs=
 		CDenseFeatures<float64_t>::obtain_from_generic(distance->get_lhs());
+	REQUIRE(lhs!=NULL, "lhs Pointer should not be NULL");
```
karlnapf: This error message does not help anyone.
BUT I just realised the cast will always work as CEuclideandistance is type safe
You can make it an assert, or just remove it

src/shogun/clustering/KMeans.cpp
```diff
@@ -197,8 +190,8 @@ void CKMeans::compute_cluster_variances()
 				for (l=0; l<dimensions; l++)
 				{
 					dist+=CMath::sq(
-							mus.matrix[i*dimensions+l]
-									-mus.matrix[j*dimensions+l]);
+							mus(l,i)
```
karlnapf: can you format this properly?

src/shogun/clustering/unittest_random_initialization.cc
```diff
@@ -0,0 +1,52 @@
+/*
```
karlnapf: The filename is not conform to the way we do it in Shogun.

src/shogun/clustering/unittest_random_initialization.cc
```diff
@@ -0,0 +1,52 @@
...
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
```
karlnapf: Can you use BSD?
See [here](https://gist.github.com/karlnapf/a535d810b7d78deaee81)
Put your name. Check other files for how to do it

src/shogun/clustering/unittest_random_initialization.cc
```diff
@@ -0,0 +1,52 @@
...
+    tri(0,2)=10;
+    tri(1,0)=0;
+    tri(1,1)=5;
+    tri(1,2)=0;
+
+    CDenseFeatures<float64_t>* features=new CDenseFeatures<float64_t>(tri);
+    CEuclideanDistance* distance=new CEuclideanDistance(features, features);
+    CKMeans* clustering=new CKMeans(2, distance);
+
+    SGVector<int32_t> ClList=SGVector<int32_t>(3);
```
karlnapf: This is neither a list, nor do we use CamelCases names.
Just call it `classes`

src/shogun/clustering/unittest_random_initialization.cc
```diff
@@ -0,0 +1,52 @@
...
+    ClList.zero();
+    SGVector<float64_t> weights_set=SGVector<float64_t>(2);
```
karlnapf: weights_set is not a good name either

src/shogun/clustering/unittest_random_initialization.cc
```diff
@@ -0,0 +1,52 @@
...
+    weights_set.zero();
+
+    clustering->set_random_centers(weights_set, ClList, 3);
+
+    CDenseFeatures<float64_t>* randomised_centers=CDenseFeatures<float64_t>::obtain_from_generic(distance->get_lhs());
+    SGMatrix<float64_t> randomised_centers_matrix=randomised_centers->get_feature_matrix();
+
+    EXPECT_TRUE(!((randomised_centers_matrix(0,0)==randomised_centers_matrix(0,1))&&(randomised_centers_matrix(1,0)==randomised_centers_matrix(1,1))));
```
karlnapf: Make this 3 separate checks

src/shogun/clustering/unittest_random_initialization.cc
```diff
@@ -0,0 +1,52 @@
...
+
+    for(int32_t i=0; i<2; i++)
+        EXPECT_TRUE(weights_set[i]!=0);
+
+    SG_UNREF(randomised_centers);
+    SG_UNREF(clustering);
```
karlnapf: this will likely cause memory errors.
Just unref clustering (which takes care of the rest)
Run  with valgrind to check for leaks

karlnapf: OR, just use `some` and `auto`.

tests/unit/clustering/unittest_random_initialization.cc
```diff
@@ -0,0 +1,52 @@
+/*
```
karlnapf: why two files?

karlnapf:
run tests via passing `-DENABLE_TESTING=On` to cmake, and then `make unit-test`.
You can run single tests via the shogun-unit-test binary and passing options. See the docs

karlnapf:
any news?

src/shogun/clustering/unittest_random_initialization.cc
manantomar: Sorry but I didn't get what you meant by 3 checks .We have to check if both coordinates are not same at the same time, so how can we split it in 3 ?

manantomar:
So I have updated with above comments .However I am still unable to figure out why the code is breaking .I am completely on it and will try to do it as soon as I can.
Thanks for the feedback.

P.S. : Just this morning my Linux os  crashed ,so it might take me little more time to start debugging again.

tests/unit/clustering/RandomInitialization_unittest.cc
```diff
@@ -0,0 +1,78 @@
...
+    /* vector for size of each cluster */
+    SGVector<float64_t> cluster_size=SGVector<float64_t>(2);
+    cluster_size.zero();
+
+    clustering->set_random_centers(cluster_size, classes, 3);
+
+    CDenseFeatures<float64_t>* randomised_centers=CDenseFeatures<float64_t>::obtain_from_generic(distance->get_lhs());
+    SGMatrix<float64_t> randomised_centers_matrix=randomised_centers->get_feature_matrix();
+
+    EXPECT_FALSE((randomised_centers_matrix(0,0)==randomised_centers_matrix(0,1)) &&
```
karlnapf: please dont do booleans, but just call EXPECT_FALSE a few times.

BTW there is EXPECT_FLOAT_EQ ...
See https://github.com/google/googletest/blob/master/googletest/docs/AdvancedGuide.md

karlnapf:
Travis is horribly broken, it doesnt compile!
Please dont send PRs with code that does not compile

tests/unit/clustering/RandomInitialization_unittest.cc
```diff
@@ -0,0 +1,78 @@
+/*
```
karlnapf: This is a unit test for KMeans, and not for random initialisation. The file should be named accordingly

tests/unit/clustering/RandomInitialization_unittest.cc
```diff
@@ -0,0 +1,78 @@
...
+    /* vector for storing cluster center of each point */
+    SGVector<int32_t> classes=SGVector<int32_t>(3);
+    classes.zero();
+    /* vector for size of each cluster */
+    SGVector<float64_t> cluster_size=SGVector<float64_t>(2);
+    cluster_size.zero();
+
+    clustering->set_random_centers(cluster_size, classes, 3);
+
+    CDenseFeatures<float64_t>* randomised_centers=CDenseFeatures<float64_t>::obtain_from_generic(distance->get_lhs());
```
karlnapf: 80 character margin wants to be respected :)

tests/unit/clustering/RandomInitialization_unittest.cc
```diff
@@ -0,0 +1,78 @@
...
+    CKMeans* clustering=new CKMeans(2, distance);
+
+    /* vector for storing cluster center of each point */
+    SGVector<int32_t> classes=SGVector<int32_t>(3);
+    classes.zero();
+    /* vector for size of each cluster */
+    SGVector<float64_t> cluster_size=SGVector<float64_t>(2);
+    cluster_size.zero();
+
+    clustering->set_random_centers(cluster_size, classes, 3);
```
karlnapf: this method is private.
You can either test it [like this](https://github.com/google/googletest/blob/master/googletest/docs/AdvancedGuide.md#testing-private-code)

or you test it through the public interface, i.e. you make the KMeans to run the random center, and then check the centers, extracted via the public interface

The second way is much preferred

manantomar:
Updated with the above comments.
So the unit test is working fine however there is a segfault while running libshogun which I am not able debug currently.
Also, I realized that saurabh has made changes to kmeans (great work done by him)which are quite similar to the ones I made (the random initialization part) ,so can you please guide me as if what I should be doing next .Should I :  
1.Stick on what I was doing 
2.Since we are targeting the same issue ,I should work with him and get kmeans cleaned up completely.
3. or move on to some other entrance task.
Thanks a lot for the feedback and looking forward to a better and improved shogun. :)

karlnapf:
I guess we will just use @Saurabh7 's changes. This is a pitty, we would like to avoid such redundant work in the future. But at least you learned a bit.

@Saurabh7 please have a look in here and if there is anything that you can re-use, please do.

It would be cool to get KMeans into mlpacks benchmark system (or at least run the new version to compare shogun with other libs). So maybe one of you can do that? Remember in open-source communities, collaboration is key! :) But you can also move to another entrance task if you prefer

Saurabh7:
Yes I agree, unfortunately I had fit together most of the major parts together.
KMeans is there in https://github.com/zoq/benchmarks ,  I ran it with some of the datasets in it and got better results in comparison with scikit with the newer version for bigger of the datasets .. 
although I am not sure everything is a fair comparison in it right now (parameters like `tol` need to be set to zero in scikit) and also initialization methods. @manantomar you could have a look if you are interested :)

manantomar:
Thanks for the comments . I will try the benchmarking now .Hopefully we can get the Kmeans running smoothly :)

PR #3030, opened 2016-03-02
add docstrings to the classification notebook.

youssef-emad:
@karlnapf I have just added the docstrings. check [here](https://github.com/youssef-emad/shogun/blob/develop/doc/ipython-notebooks/classification/Classification.ipynb).


karlnapf:
Thanks!

PR #3038, opened 2016-03-04
Added Neural Tanh Layer and corresponding unit test

arasuarun:
The Tanh squasher is a very popular alternative to the sigmoid layer. It's widely used, although not as much as before thanks to relu's. It's definitely important for a complete neural nets package. :)

### DIFF
--- a/src/shogun/neuralnets/NeuralTanhLayer.cpp
+++ b/src/shogun/neuralnets/NeuralTanhLayer.cpp
@@ -1,0 +1,127 @@
+/*
+ * Copyright (c) 2016, Shogun Toolbox Foundation
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+
+ * 1. Redistributions of source code must retain the above copyright notice,
+ * this list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ *
+ * 3. Neither the name of the copyright holder nor the names of its
+ * contributors may be used to endorse or promote products derived from this
+ * software without specific prior written permission.
+
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ *
+ * Written (W) 2016 Arasu Arun
+ */
+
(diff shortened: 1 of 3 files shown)


src/shogun/neuralnets/NeuralTanhLayer.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+	SGMatrix<float64_t> W(parameters.vector+m_num_neurons,
+		m_num_neurons, num_inputs, false);
+
+	float64_t contraction_term = 0;
+	for (int32_t i=0; i<m_num_neurons; i++)
+	{
+		float64_t sum_j = 0;
+		for (int32_t j=0; j<num_inputs; j++)
+			sum_j += W(i,j)*W(i,j);
```
vigsterkr: it's not 80s anymore. please use appropriate SIMD operations to do these things... or just use eigen.

src/shogun/neuralnets/NeuralTanhLayer.cpp
```diff
@@ -0,0 +1,127 @@
...
+float64_t CNeuralTanhLayer::compute_contraction_term(
+	SGVector< float64_t > parameters)
+{
+	int32_t num_inputs = SGVector<int32_t>::sum(m_input_sizes.vector, m_input_sizes.vlen);
+
+	SGMatrix<float64_t> W(parameters.vector+m_num_neurons,
+		m_num_neurons, num_inputs, false);
+
+	float64_t contraction_term = 0;
+	for (int32_t i=0; i<m_num_neurons; i++)
```
vigsterkr: specify openmp pragmas

vigsterkr:
this is a vanilla implementation. please consider using our linalg library and openmp at the least.

arasuarun:
I was wondering why you guys didn't use the linalg library for the logistic layer so I just stuck with the same implementation. I'll modify this one to use it now. Thanks.

src/shogun/neuralnets/NeuralTanhLayer.h
```diff
@@ -0,0 +1,124 @@
...
+
+#ifndef __NEURALTANHLAYER_H__
+#define __NEURALTANHLAYER_H__
+
+#include <shogun/neuralnets/NeuralLinearLayer.h>
+
+namespace shogun
+{
+/** @brief Neural layer with linear neurons, with a [tanh activation
+ * function](http://en.wikipedia.org/wiki/Hyperbolic_function). can be used as a
```
karlnapf: Can be used .... Capital C

Can you also put a reference to a paper here, where there is some evidence that this is useful? Thanks

tests/unit/neuralnets/NeuralTanhLayer_unittest.cc
```diff
@@ -0,0 +1,167 @@
...
+
+	// compute the layer's activations
+	input->compute_activations(x);
+	layer.compute_activations(params, layers);
+	SGMatrix<float64_t> A = layer.get_activations();
+
+	// manually compute the layer's activations
+	SGMatrix<float64_t> A_ref(layer.get_num_neurons(), x.num_cols);
+
+	float64_t* biases = params.vector;
```
karlnapf: Same as before, pls dont use pointers

tests/unit/neuralnets/NeuralTanhLayer_unittest.cc
```diff
@@ -0,0 +1,167 @@
...
+	float64_t* weights = biases + layer.get_num_neurons();
+
+	for (int32_t i=0; i<A_ref.num_rows; i++)
+	{
+		for (int32_t j=0; j<A_ref.num_cols; j++)
+		{
+			A_ref(i,j) = biases[i];
+
+			for (int32_t k=0; k<x.num_rows; k++)
+				A_ref(i,j) += weights[i+k*A_ref.num_rows]*x(k,j);
```
karlnapf: SGMatrix has the () operator

tests/unit/neuralnets/NeuralTanhLayer_unittest.cc
```diff
@@ -0,0 +1,167 @@
...
+#include <shogun/lib/SGMatrix.h>
+#include <shogun/mathematics/Math.h>
+#include <gtest/gtest.h>
+
+using namespace shogun;
+
+/** Compares the activations computed using the layer against manually computed
+ * activations
+ */
+TEST(NeuralTanhlayer, compute_activations)
```
karlnapf: Same as before, these unit tests are too big! 
Try to test much smaller units .... split more tests. Unit tests should mostly just cover a single method, not the output of a full chain of operations as here. Try to isolate things more 
You can keep the big one as well though ....

karlnapf:
Nice patch.
I agree with @vigsterkr please use linalg, eigen, and openmp to make this stuff fast

karlnapf:
news?

arasuarun:
@karlnapf if you are okay with it, I think you can merge it now. When refactoring the nn package with linalg and openmp, I think it'd be better if all the modules are refactored at once. What do you think?

karlnapf:
Few things that block this being merged.
 * travis is failing, your code does not compile
 * Shogun's NN should be refactored 1by1, not all at once (small PRs!)
 * Put all the open mp stuff already in. Once something is merged, it usually is unlikely that it is touched again, so we want to make it as good as we can in the first merge

PR #3039, opened 2016-03-05
write output with utf-8 encoding

besser82:
Doxy2Swig should write UTF-8 encoded strings to the output-file.


besser82:
Travis is fine.  Merging…

PR #3042, opened 2016-03-05
add dot products and precomputed norms to CEuclideanDistance

Saurabh7:


src/shogun/distance/EuclideanDistance.cpp
```diff
@@ -54,6 +55,32 @@ float64_t CEuclideanDistance::compute(int32_t idx_a, int32_t idx_b)
 		get_feature_vector(idx_b, blen, bfree);
 	ASSERT(alen==blen)
 
+	if(m_dot_enabled)
+	{
+		result+=CMath::dot(avec, bvec, alen);
```
vigsterkr: linalg or sgvect not cmath plz

Saurabh7:
w.r.t discussion in #2987 .

- distance computed using dot products as ` dist(x, y) = sqrt(dot(x, x) - 2 * dot(x, y) + dot(y, y))`
- precompute lhs and rhs squared norms
- unit tests

src/shogun/distance/EuclideanDistance.cpp
```diff
@@ -54,6 +55,32 @@ float64_t CEuclideanDistance::compute(int32_t idx_a, int32_t idx_b)
...
+		result*=-2;
+		if(m_rhs_sq_norms)
+		{
```
karlnapf: minor: we dont use `{ ... }` if there is only a single line of code inside

src/shogun/distance/EuclideanDistance.h
```diff
@@ -37,7 +37,12 @@ namespace shogun
  * \f[\displaystyle
  *  d({\bf x},{\bf x'})= \sum_{i=0}^{n}|{\bf x_i}-{\bf x'_i}|^2
  * \f]
- *
+ * 
+ * Alternatively can be computed as :
```
karlnapf: I think this should be the default, not an option. Don't you agree?

src/shogun/distance/EuclideanDistance.h
```diff
@@ -125,6 +161,21 @@ class CEuclideanDistance: public CRealDistance
 	protected:
 		/** if application of sqrt on matrix computation is disabled */
 		bool disable_sqrt;
+
+		/** if squared norms of rhs features are used */
+		bool m_rhs_sq_norms;
```
karlnapf: please remove these bools. We should always precompute the norms, no?

src/shogun/distance/EuclideanDistance.h
```diff
@@ -125,6 +161,21 @@ class CEuclideanDistance: public CRealDistance
...
+
+		/** if squared norms of lhs features are used */
+		bool m_lhs_sq_norms;
+
+		/** i dot products are used in distance computations */
+		bool m_dot_enabled;
+
+		/** squared norms from features of right hand side to be used if dot enabled */
+		SGVector<float64_t> m_rhs_squared_norms;
```
karlnapf: if you make sure that the rhs norms are always computed when the rhs is changed, these are always up to date, so no need for booleans

src/shogun/distance/EuclideanDistance.cpp
```diff
@@ -66,10 +93,63 @@ float64_t CEuclideanDistance::compute(int32_t idx_a, int32_t idx_b)
...
+
+	m_rhs_squared_norms=rhs_sq;
+}
+
+void CEuclideanDistance::precompute_lhs_squared_norms()
+{
+	m_lhs_sq_norms=true;
+	SGVector<float64_t>lhs_sq=SGVector<float64_t>(lhs->get_num_vectors());
+
+	for(int32_t idx_i=0; idx_i<lhs->get_num_vectors(); idx_i++)
```
karlnapf: open mp can be used here. Might be good to check that

src/shogun/distance/EuclideanDistance.cpp
```diff
@@ -66,10 +93,63 @@ float64_t CEuclideanDistance::compute(int32_t idx_a, int32_t idx_b)
...
+	{
+		SGVector<float64_t> tempvec=((CDenseFeatures<float64_t>*) lhs)->get_feature_vector(idx_i);
+		float64_t temp=linalg::dot(tempvec, tempvec);
```
karlnapf: combine this and the next line

src/shogun/distance/EuclideanDistance.cpp
```diff
@@ -66,10 +93,63 @@ float64_t CEuclideanDistance::compute(int32_t idx_a, int32_t idx_b)
...
+		lhs_sq[idx_i]=temp;
+		((CDenseFeatures<float64_t>*) lhs)->free_feature_vector(tempvec, idx_i);
```
karlnapf: Can you investigate the `free_feature_vector` mechanism a bit and see whether this call is really needed?

src/shogun/distance/EuclideanDistance.cpp
```diff
@@ -66,10 +93,63 @@ float64_t CEuclideanDistance::compute(int32_t idx_a, int32_t idx_b)
...
+
+	m_lhs_squared_norms=lhs_sq;
+}
+
+void CEuclideanDistance::set_dot_enabled(bool dot)
+{
+	m_dot_enabled=dot;
+}
+
+bool CEuclideanDistance::get_dot_enabled()
```
karlnapf: all these methods should go

tests/unit/distance/EuclideanDistance_unittest.cc
```diff
@@ -63,6 +63,49 @@ TEST(EuclideanDistance,distance)
 	SG_UNREF(euclidean); // the features are unref-ed here as well
 }
 
+TEST(EuclideanDistance, distance_dot)
```
karlnapf: good test, I guess we just need one of them if the bools are removed

src/shogun/distance/EuclideanDistance.cpp
Saurabh7: found a performance issue with linalg::dot and sgvector for smaller dimensions. https://gist.github.com/lambday/7cbd77eb0530ad41d9f5 . So we might have to use cmath::dot for time being till we have updates as this is inside `compute()`

src/shogun/distance/EuclideanDistance.h
Saurabh7: I wasnt completely sure of implications on other algorithms so decided to start off by keeping it optional.

src/shogun/distance/EuclideanDistance.h
Saurabh7: - Decided to make `precompute_xxx_squared_norms()` public since we don't know how often the rhs is changing.
for example: 
 - if rhs changes and we end up computing pairwise distances between say _all_ vectors of rhs and lhs this is worth.
 - But if rhs changes quite often and we  need pairwise distances using  only _some_  vectors , computing squared norms each time would be inefficient ?

src/shogun/distance/EuclideanDistance.cpp
karlnapf: OK! Great catch.
Even better would be to fix this and then use linalg

src/shogun/distance/EuclideanDistance.h
karlnapf: But the results will not change at all, this is just precomputing things

src/shogun/distance/EuclideanDistance.h
karlnapf: I see your point here.
In fact, I think it does not make sense to precompute the rhs squared norms at all. Distance is only computed once no?
Maybe only precompute lhs?

Saurabh7: We are not calculating the whole distance matrix. It is just between 2 vectors from lhs and rhs  `distance (index a, index b)`
So if we know that either lhs or rhs or both are fixed over some number of computations we can just use `rhs_norm[ index b]` instead of computing `b.b` for every computation. That was the idea.

src/shogun/distance/EuclideanDistance.h
Saurabh7: Yes, i was thinking performance wise, but now that i think of it cant see a reason not to, will change it then.

src/shogun/distance/EuclideanDistance.h
karlnapf: What about:
storing the lhs norms always
not storing the rhs norms

Offer a method that computes all pairwise distance between lhs and rhs, and precompute the rhs norms in there?

H

karlnapf: Actually I had another thought about this.

It should be like this (and this is close to what you did initially)

By default, nothing is precomputed (we dont know which distances user wants to access)
 and both the rhs and lhs norms are just empty vectors
There is an optional method "precompute" (plus two method precompute_lhs and precompute_rhs that precomputes both sides
Whenever rhs or lhs are changed, the norm vectors are reset.

There is a method to compute the distance for *all* pairs of lhs/rhs. This one obviously calls precompute inside

Then you can manually call the precompute in KMeans and/or ask for many distances

Saurabh7: Do you mean combining both precompute in one method ? That would be inefficient as in KMeans where one is fixed always but the other is changing over iterations (but fixed inside one iteration)

karlnapf: No I mean to have a method for each rhs and lhs, and one wrapper for both

Saurabh7: So the wrapper should call one of the or both methods depending on what has changed. Am i right ?

If i am right : the only way to keep track of which side changed is in `replace_lhs` / rhs which happens to be non virtual in `CDistance`

karlnapf: The wrapper just calls both methods, so that you dont have to call them manually. You can also just skip the wrapper. Not important.

I dont know where the rhs/lhs are changed, you have to check that. But you can definitely overload these methods, reset the norm vectors and then call the base class method

Saurabh7:
Updated to remove the bools.

Also @karlnapf I did check there is no way to know for sure when rhs/lhs is modified. `replace_rhs( )` only changes the `CFeature` object. but we cannot know when its feature matrix is modified , as is happening in KMeans too. So we may have to rely on public precompute only to compute new norms.

src/shogun/distance/EuclideanDistance.cpp
```diff
@@ -54,9 +55,19 @@ float64_t CEuclideanDistance::compute(int32_t idx_a, int32_t idx_b)
 		get_feature_vector(idx_b, blen, bfree);
 	ASSERT(alen==blen)
 
-	for (int32_t i=0; i<alen; i++)
-		result+=CMath::sq(avec[i] - bvec[i]);
+	result+=CMath::dot(avec, bvec, alen);
+	result*=-2;
+	
+	if(m_rhs_squared_norms.size()>0)
```
karlnapf: I would rather use if(m_rhs_squared_norms.vec)

karlnapf:
Ok I see, you are right. I still am unsure whether this addition is really a good idea as people might forget about it and then get wrong results without a warning or an error.
The final option I see is to add a method to compute pairwise distances for all features rhs/lhs. In this case, we can definitely precompute the norms internally. I think this would also fit into KMeans wouldnt it?
You can use subsets on the features to specify which indices you want to use.

Let me know what you think about that!

Saurabh7:
Yes that could lead to problems if people forget to do it right actually. 
computing pairwise distance for all is good idea but then we have to store extra  numofsamples*numofcenters size matrix per iteration
The way scikit do it is:
- have percomute_distance argument which if true and numofsamples*numofcenters < some number (so not too much memeory used) they will store all pairwise distances in matrix.
- They actually let the user compute norms externally and pass them as arguments. This norms are then used to compute multipe pairwise distances.

I think we can have a function to compute multiple distances (which could compute single pairwise distance too if only 2 indices passed as args). We can use user calculated norms passed as arguments so things dont go wrong.

karlnapf:
Storing this matrix is a bad idea indeed.
I think let's maybe keep it as it is now, no need to let users pass things ...
But make sure to add reset() calls in all places where the features do get changed, to avoid problems described above.

src/shogun/distance/EuclideanDistance.cpp
```diff
@@ -66,10 +77,44 @@ float64_t CEuclideanDistance::compute(int32_t idx_a, int32_t idx_b)
...
+	m_lhs_squared_norms=SGVector<float64_t>();
+	m_rhs_squared_norms=SGVector<float64_t>();
+}
+
 void CEuclideanDistance::init()
 {
 	disable_sqrt=false;
-
+	reset_squared_norms();
 	m_parameters->add(&disable_sqrt, "disable_sqrt", "If sqrt shall not be applied.");
```
karlnapf: need to register the squared norm members here as well

src/shogun/distance/EuclideanDistance.h
```diff
@@ -37,6 +37,11 @@ namespace shogun
  * \f[\displaystyle
  *  d({\bf x},{\bf x'})= \sum_{i=0}^{n}|{\bf x_i}-{\bf x'_i}|^2
  * \f]
+ * 
+ * Distance is computed as :  
+ * \sqrt{{\bf x}\cdot {\bf x} - 2{\bf x}\cdot {\bf x'} + {\bf x'}\cdot {\bf x'}}
+ *  
+ * Squared norms for left hand side and right hand side features can be precomputed.
```
karlnapf: but a big warning here that the user has to track these calls when features are changed

src/shogun/distance/EuclideanDistance.h
```diff
@@ -112,6 +117,21 @@ class CEuclideanDistance: public CRealDistance
 		 *  @return distance value or upper_bound
 		 */
 		virtual float64_t distance_upper_bounded(int32_t idx_a, int32_t idx_b, float64_t upper_bound);
+		
+		/**
+		 * Precompute squared norms from features of right hand side
```
karlnapf: another warning should be put here

src/shogun/distance/EuclideanDistance.h
```diff
@@ -112,6 +117,21 @@ class CEuclideanDistance: public CRealDistance
...
+
+		/**
+		 * Precompute squared norms from features of left hand side
+		 */
+		virtual void precompute_lhs_squared_norms();
+	
+		/**
+		 * Reset squared norms for features of both sides
+		 */
+		virtual void reset_squared_norms();
```
karlnapf: Should contain instructions that this needs to be called in case of user changes the features

karlnapf:
Good to merge apart from the minors. Can you also squash all this into a single commit?

Saurabh7:
Done. btw does travis still check integration tests ? :)

karlnapf:
Nope, we disabled them. Soon to be replaced though

karlnapf:
No, dont bother

karlnapf:
travis is fine, merging

Saurabh7:
hmm  - http://buildbot.shogun-toolbox.org/builders/precise%20-%20libshogun/builds/1140/steps/compile/logs/stdio . Is this because of  guards for linalg ?

karlnapf:
yes, please fix this soon
Also will make some minor comments above

src/shogun/distance/EuclideanDistance.cpp
```diff
@@ -66,11 +77,47 @@ float64_t CEuclideanDistance::compute(int32_t idx_a, int32_t idx_b)
...
+	m_parameters->add(&m_rhs_squared_norms, "m_rhs_squared_norms", "squared norms from features of right hand side");
```
karlnapf: Capital S in "squared"

tests/unit/distance/EuclideanDistance_unittest.cc
```diff
@@ -63,6 +63,28 @@ TEST(EuclideanDistance,distance)
...
+	euclidean->precompute_rhs_squared_norms();	
+
+	// check distances computed one by one
+	EXPECT_EQ(euclidean->distance(0,0), 2);
+	EXPECT_EQ(euclidean->distance(0,1), 2);
+	EXPECT_EQ(euclidean->distance(1,0), 5);
+	EXPECT_EQ(euclidean->distance(1,1), 5);
+
+	// release memory
+	SG_UNREF(euclidean); // the features are unref-ed here as well
```
karlnapf: no need for this second comment, also we dont really do comments in same line

tests/unit/distance/EuclideanDistance_unittest.cc
```diff
@@ -63,6 +63,28 @@ TEST(EuclideanDistance,distance)
...
+	euclidean->precompute_lhs_squared_norms();
+	euclidean->precompute_rhs_squared_norms();	
+
+	// check distances computed one by one
+	EXPECT_EQ(euclidean->distance(0,0), 2);
+	EXPECT_EQ(euclidean->distance(0,1), 2);
+	EXPECT_EQ(euclidean->distance(1,0), 5);
+	EXPECT_EQ(euclidean->distance(1,1), 5);
+
+	// release memory
```
karlnapf: this comment can go, the code speaks for itself

src/shogun/distance/EuclideanDistance.cpp
```diff
@@ -66,11 +77,47 @@ float64_t CEuclideanDistance::compute(int32_t idx_a, int32_t idx_b)
...
+	{
+		SGVector<float64_t> tempvec=((CDenseFeatures<float64_t>*) rhs)->get_feature_vector(idx_i);
+		rhs_sq[idx_i]=linalg::dot(tempvec, tempvec);
+		((CDenseFeatures<float64_t>*) rhs)->free_feature_vector(tempvec, idx_i);
+	}
+
+	m_rhs_squared_norms=rhs_sq;
+}
+
+void CEuclideanDistance::precompute_lhs_squared_norms()
```
iglesias: Sorry about joining late :-)

To avoid code duplication, I'd say it is worth to make an auxiliary method that computes the squared norms given the appropriate CDenseFeatures<float64_t> object.

src/shogun/distance/EuclideanDistance.h
```diff
@@ -37,6 +37,13 @@ namespace shogun
  * \f[\displaystyle
  *  d({\bf x},{\bf x'})= \sum_{i=0}^{n}|{\bf x_i}-{\bf x'_i}|^2
  * \f]
+ * 
+ * Distance is computed as :  
+ * \sqrt{{\bf x}\cdot {\bf x} - 2{\bf x}\cdot {\bf x'} + {\bf x'}\cdot {\bf x'}}
+ *  
+ * Squared norms for left hand side and right hand side features can be precomputed.
+ * WARNING : Make sure to reset squared norms using reset_squared_norms() when features
+ * or feature matrix are changed.
```
iglesias: I think that this warning only the doc is very dangerous. We should handle this is in the code as well.

src/shogun/distance/EuclideanDistance.cpp
karlnapf: Good point :) I agree

src/shogun/distance/EuclideanDistance.h
karlnapf: It is not 100% possible -- some algorithms change the feature matrix. We had a long discussion here. If you have a solution, share it!
Though we should definitely add this calls whenever we *know* that the features are changed 
@Saurabh7 can you take care?

iglesias: Would be it possible using a `precomputed` boolean attribute? If set to true, the distances are precomputed, and updated whenever a feature matrix is changed.

karlnapf: We dont know when feature matrix is changed....
For the cases we know, we can (and should) just call reset.
But you are right it is not yet optimal. I leave it to @Saurabh7 to come up with a better solution

src/shogun/distance/EuclideanDistance.cpp
Saurabh7: yes will add this

Saurabh7:
Thanks for the comments
well the options are straight forward, i will summarize again:

- we handle precompute in distance class: No way to know when feature matrix changes. hence cant automatically reset. User has to reset precomputations

- let user pass precomputed norms: user has to reset here too.

- or calculate mutiple distance pairs : only in this case can we precompute and reset safely.
  - But this comes at a cost:
  - have to store a results in a matrix
  - some unnnecessary computations if only one feature side is changing.

karlnapf:
Let us take option 1, reset the norms whenever we know that the distance is changed (we know often), add a warning for the cases we dont know

PR #3059, opened 2016-03-11
minor fixes HAVE_LINALG

Saurabh7:
some fixes after https://github.com/shogun-toolbox/shogun/pull/3042.

### DIFF
--- a/src/shogun/distance/EuclideanDistance.cpp
+++ b/src/shogun/distance/EuclideanDistance.cpp
@@ -11,7 +11,10 @@
 #include <shogun/lib/common.h>
 #include <shogun/io/SGIO.h>
 #include <shogun/distance/EuclideanDistance.h>
+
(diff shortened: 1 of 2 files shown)


karlnapf:
Thanks!

Saurabh7:
Hmm I didnt know about this HAVE_LINALG change  @karlnapf 
https://github.com/shogun-toolbox/shogun/commit/7339764920061d35581287f2483c24cc6187901e#diff-af3b638bc2a3e6c650974192a53c7291L760
 so its set to 1 only if VIENNACL is found ??

karlnapf:
Yes, we recently made eigen3 a hard requirement

Saurabh7:
ah then this guard wouldnt make sense here, it will fail even if eigen is found ?

karlnapf:
HAVE_LINALG is only set if c++11 is available, that is why you need to use it

PR #3062, opened 2016-03-12
eigen3 correction

karlnapf:
@vigsterkr this is both a fix of HAVE_LINALG not being defined earlier.
In addition, I made the individual linalg options proper string options, before they were just ON/OFF, which means they were ignored.
Also, I renamed them to start with LINALG_ so that they are grouped if ppl check the options and not scattered all over without it being clear that they change linag behaviour.

 This might interfere with some builds. Let me know

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -780,16 +780,15 @@ ELSE ()
 ENDIF ()
 
 # Linear algebra default global backend setups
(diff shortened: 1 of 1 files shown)


karlnapf:
sorry forgot to remove this test file, --reopening

PR #3063, opened 2016-03-12
eigen3 correction

karlnapf:
@vigsterkr UPDATE: I applied the fix to develop as it is definitely necessary. The rest is optional for cleaner cmake options.

I renamed them to start with LINALG_ so that they are grouped if ppl check the options and not scattered all over without it being clear that they change linag behaviour. Also, ccmake which shows the cached options now does't show a binary ON/OFF, but the string. Just like CMAKE_BUILD_TYPE

 This might interfere with some builds. Let me know

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -779,11 +779,15 @@ ELSE ()
 ENDIF ()
 
 # Linear algebra default global backend setups
-OPTION (LinAlgBackend "Set global linear algebra backend library for all modules")
-IF (LinAlgBackend STREQUAL "EIGEN3")
+if( NOT LINALG_DEFAULT_BACKEND )
+  set( LINALG_DEFAULT_BACKEND EIGEN3 CACHE STRING
+       "Set global linear algebra backend library for all modules: EIGEN3, VIENNACL"
+       FORCE )
+endif()
+IF (LINALG_DEFAULT_BACKEND STREQUAL "EIGEN3")
 	SET(USE_EIGEN3 1)
 	MESSAGE("-- Eigen3 set as default global linear algebra backend library")
-ELSEIF (LinAlgBackend STREQUAL "VIENNACL")
+ELSEIF (LINALG_DEFAULT_BACKEND STREQUAL "VIENNACL")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -773,22 +773,22 @@ ENDIF()
 
 # Check if supported linear algebra backend is found
 IF (HAVE_CXX0X OR HAVE_CXX11)
-	IF (VIENNACL_FOUND AND ENABLE_VIENNACL)
-		SET(HAVE_LINALG_LIB 1)
-		MESSAGE("-- Supported backend(s) for linear algebra FOUND and ENABLED")
-	ELSE ()
-		MESSAGE("-- Supported backend(s) for linear algebra NOT FOUND/NOT ENABLED")
-	ENDIF ()
+	SET(HAVE_LINALG_LIB 1)
```
karlnapf: HAVE_LINALG was not enabled previously, which was an error

karlnapf:
mmh, no travis-ci somehow

karlnapf:
travis is [green](https://travis-ci.org/shogun-toolbox/shogun/builds/115528764)

karlnapf:
The viennaCL buildbot requires to change `DLinAlgBackend=VIENNACL ..` to `LINALG_DEFAULT_BACKEND=VIENNACL ..`

vigsterkr:
@karlnapf any reason why not to merge this?

karlnapf:
needs a  buildbot change, see comment above. I would prefer this.
Not important as it is refactored anyways. Merge if it is easy to change buildbot scripts, otherwise just close

vigsterkr:
i've merged and updated the buildbot params for the viennacl job, i'm getting the following output from cmake
```
-- ViennaCL set as default global linear algebra backend library
-- Eigen3 set as default core module (linalg) backend library
-- Eigen3 set as default reduction module (linalg) backend library
-- Eigen3 set as default eigen solver module (linalg) backend library
```
@karlnapf why Eigen3 is still the eigen solver module?

karlnapf:
This has to be fixed

karlnapf:
I guess this can wait until we re-structure no?

PR #3065, opened 2016-03-13
Avoid compiler warnings in LinearMachine

youssef-emad:
@karlnapf 
I compiled and there were no warnings.
I am using gcc  4.8.4

### DIFF
--- a/src/shogun/machine/LinearMachine.cpp
+++ b/src/shogun/machine/LinearMachine.cpp
@@ -22,11 +22,11 @@ CLinearMachine::CLinearMachine(): CMachine()
 	init();
 }
 
-CLinearMachine::CLinearMachine(bool compute_bias): CMachine()
+CLinearMachine::CLinearMachine(bool comput_bias): CMachine()
 {
 	init();
 
-	m_compute_bias = compute_bias;
+	set_compute_bias(comput_bias);
 }
 
 CLinearMachine::CLinearMachine(CLinearMachine* machine) : CMachine()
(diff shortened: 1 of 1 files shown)


src/shogun/machine/LinearMachine.cpp
```diff
@@ -35,6 +35,7 @@ CLinearMachine::CLinearMachine(CLinearMachine* machine) : CMachine()
 
 	set_w(machine->get_w().clone());
 	set_bias(machine->get_bias());
+	set_compute_bias(machine->get_compute_bias());
```
karlnapf: Might be good to unit test this. We forgot about it initially ;)

src/shogun/machine/LinearMachine.cpp
```diff
@@ -22,11 +22,11 @@ CLinearMachine::CLinearMachine(): CMachine()
 	init();
 }
 
-CLinearMachine::CLinearMachine(bool compute_bias): CMachine()
+CLinearMachine::CLinearMachine(bool comput_bias): CMachine()
```
karlnapf: dont really like the new name, since it has a typo, but that might be just me being OCD ;)

karlnapf:
Thanks!

src/shogun/machine/LinearMachine.cpp
youssef-emad: hahaha :smiley:

youssef-emad:
urw :smile:

PR #3068, opened 2016-03-13
general update of knn example, in particular integration test

karlnapf:


karlnapf:
All relevant (that is strongly typed languages) travis builds are green. Merging

PR #3067, opened 2016-03-13
kmeans update

Saurabh7:
This has most of the primary changes to Kmeans based on https://github.com/shogun-toolbox/shogun/issues/2987 and some general cleanups.
Still have to do some aesthetic changes.

- Changed random initialization
  - now chooses `k` random points as centers rather than assigning random centers to all points.
  - Removed unnecessary paramters, cluster_list, weight_set, etc. This should be calculated in the Lloyds implementation.
- changed manual initialization
  - again Removed unnecessary paramters
  - removed calculation of  distances and assigning of clusters from here, this along wth the parameters now takes place in assignment step of lloyds.

- made clear separation for assigment step and update step in lloyds impl.
- removed random sampling of instances to be updated in each iteration, each sample should be updated. 
- used precomputed squared norms for euclidean distances
- removed various unnecessary computations in lloyds impl
  - no need to change center values in assigment steps now takes place only in update step.
  - no need to store distances in iterations, now calculated in place (scope for parallelisation) 
- some new unit-tests

PR #3073, opened 2016-03-15
add precompute to base Distance

Saurabh7:
Adding precompute to base distance to overload in sub class.
Discussed in PR https://github.com/shogun-toolbox/shogun/pull/3067 based on https://github.com/shogun-toolbox/shogun/issues/2987

### DIFF
--- a/src/shogun/distance/Distance.h
+++ b/src/shogun/distance/Distance.h
@@ -120,6 +120,29 @@ class CDistance : public CSGObject
 			return distance(idx_a, idx_b);
 		}
 
+		/**  
+		 * Precomputation related to features of right hand side
+		 * WARNING : Make sure to reset computations using reset_precompute()
+		 * when features or feature matrix are changed.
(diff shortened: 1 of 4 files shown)


src/shogun/distance/Distance.h
```diff
@@ -120,6 +120,29 @@ class CDistance : public CSGObject
 			return distance(idx_a, idx_b);
 		}
 
+		/**  
+		 * Precomputation related to features of right hand side
+		 * WARNING : Make sure to reset computations using reset_precompute()
```
karlnapf: This needs a bit more documentation.
Put the warnings you have in Euclidean distance here as well. Also mention this method in the class description on top.

karlnapf:
Nice thanks. I will merge this, please address the minor documentation issue in the next patch

karlnapf:
Merging as this is only affects interfaces

PR #3076, opened 2016-03-15
fix FC22 buildbot (hopefully)

karlnapf:
http://buildbot.shogun-toolbox.org/builders/FC22%20-%20libshogun/builds/150/steps/test/logs/stdio

Avoid using REGEXP for untangling filenames in meta examples, but rather leave this to cmake. Hoping this fixes the problem on buildbot


karlnapf:
lets see if it worked :)

Issue #3072, opened 2016-03-14
Remove NATIVE parts from linalg


karlnapf:
As we now fully depend on eigen3, we dont need our own implementations in linalg anymore....

sanuj:
@karlnapf shall i entirely remove `NATIVE` from linalg or just the implementations?

karlnapf:
we don't need the backend itself anymore. And yes, all the implementation will also go.

BUT before you delete them, make sure that the eigen3 counterpart is implemented (and also has unit test coverage)

@lambday should comment

sanuj:
@lambday  waiting for your comment :)

sanuj:
@karlnapf This is done for now as we can't remove NATIVE backend. It is still being used in elementwise operations which eigen3 is bad at. @lambday can confirm.

PR #3081, opened 2016-03-15
remove native scale and vector sum in linalg

sanuj:
For #3072.


karlnapf:
Just to make sure: these are all covered by an eigen3 implementation yes?

PR #3082, opened 2016-03-16
add get_w to lrr cookbook page

sanuj:
Shows api to extract weights in the lrr cookbook page


karlnapf:
I think you might have forgotten data

sanuj:
@karlnapf I've updated, now travis should not complain. Everything passing on my local.

PR #3084, opened 2016-03-17
Added HAVE_EIGEN3 setting

OXPHOS:
- ref issue#3083
- Temporal fix in CMakeList.txt and config.h.in to include Eigen3 library


vigsterkr:
@OXPHOS thnx for the patch but lets rather fix the unit tests... we really would like to have eigen as a hard dependency.

OXPHOS:
Sure I was just having trouble testing my code. I would like to know how to enable Eigen3 in unit tests if possible. Thanks!

vigsterkr:
ok so you were having problem when you included eigen headers? could you point me to an error, i.e. copy paste to pastebin or something?

karlnapf:
Sorry I just only saw this so my other comments might be outdated

PR #3085, opened 2016-03-17
Implement Mean in Linalg - Vector Mean

OXPHOS:
- Ref issue ##3051
- Added VectorMean (and tests) for Eigen3


vigsterkr:
where's the problem with `EIGEN3`? i let's try to fix that instead of adding it back to the macros

OXPHOS:
It won't run any test/module with `ifdef HAVE_EIGEN3`

vigsterkr:
yea but there should not be any `#ifdef HAVE_EIGEN3` anymore anywhere

OXPHOS:
Seriously..I forked the base this afternoon and I think it is everywhere..

OXPHOS:
Okay I'll rebase and remove any `ifdef HAVE_EIGEN3` I run into.

vigsterkr:
@OXPHOS ok i see what you mean i just did a:
```
cd tests/unit
git grep HAVE_EIGEN
```
all those `#ifdef HAVE_EIGEN` should be out...

OXPHOS:
@vigsterkr Thanks for letting me know...learned a lot about cmake during the struggle. Will update the code later.

karlnapf:
Ill open an entrance task to remove them all.

CMakeLists.txt
```diff
@@ -565,8 +565,10 @@ FIND_PACKAGE(Eigen3 ${EIGEN_VERSION_MINIMUM})
 IF(NOT EIGEN3_FOUND)
 	include(external/Eigen3)
 	LIST(APPEND SYSTEM_INCLUDES ${EIGEN_INCLUDE_DIR})
+	SET(HAVE_EIGEN3 1)
```
karlnapf: I think this should go...we rather should fix #3088 
@vigsterkr agree?

examples/meta/generator/translate.py
```diff
@@ -1,17 +1,22 @@
 import json
 import sys
 from string import Template
-from sets import Set
 import os.path
 import argparse
 import re
+# import set in a python 2->3 safe way
```
karlnapf: oh this is a patch I recently sent. 
You have to
`git pull --rebase upstream develop`
to avoid this showing up

CMakeLists.txt
vigsterkr: yes yes this should never come back :)

src/shogun/mathematics/linalg/internal/implementation/VectorMean.h
```diff
@@ -0,0 +1,134 @@
...
+
+	/**
+	 * Method that computes the mean of SGVectors using Eigen3
+	 *
+	 * @param vec a vector whose mean has to be computed
+	 * @return the vector mean \f$\mean_i a_i\f$
+	 */
+	static T compute(SGVector<T> vec)
+	{
+		REQUIRE(vec.vlen > 0, "Vector can not be empty!\n");
```
karlnapf: typo: cannot

src/shogun/mathematics/linalg/internal/implementation/VectorMean.h
```diff
@@ -0,0 +1,134 @@
...
+
+	/**
+	 * Method that computes the mean of SGVectors using Eigen3
+	 *
+	 * @param a vector whose mean has to be computed
+	 * @return the vector mean \f$\mean_i a_i\f$
+	 */
+	static T compute(CGPUVector<T> vec)
+	{
+		SG_SERROR("The operation VectorMean() on a vector is currently not supported\n");
```
karlnapf: on a GPUVector

karlnapf: If it is not supported, then why put the implementation ?
@lambday what is the motivation for this?

src/shogun/mathematics/linalg/internal/modules/Redux.h
```diff
@@ -36,6 +36,7 @@
 #include <shogun/mathematics/linalg/internal/implementation/Sum.h>
```
karlnapf: ehm, please never change file permissions in pull requests.... dont know what happened here

tests/unit/mathematics/linalg/VectorMean_unittest.cc
```diff
@@ -0,0 +1,83 @@
...
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * The views and conclusions contained in the software and documentation are those
+ * of the authors and should not be interpreted as representing official policies,
+ * either expressed or implied, of the Shogun Development Team.
+ */
+
+#include <shogun/lib/config.h>
+
+#if defined(HAVE_CXX0X) || defined(HAVE_CXX11)
```
karlnapf: This should be
HAVE_LINALG_LIB instead

tests/unit/mathematics/linalg/VectorMean_unittest.cc
```diff
@@ -0,0 +1,83 @@
...
+
+#include <shogun/mathematics/linalg/linalg.h>
+#include <shogun/lib/SGVector.h>
+#include <gtest/gtest.h>
+
+#ifdef HAVE_EIGEN3
```
karlnapf: these can go

tests/unit/mathematics/linalg/VectorMean_unittest.cc
```diff
@@ -0,0 +1,83 @@
...
+#include <shogun/lib/GPUVector.h>
+#endif
+
+using namespace shogun;
+
+#ifdef HAVE_LINALG_LIB
+#ifdef HAVE_EIGEN3
+TEST(VectorMean, SGVector_explicit_eigen3_backend)
+{
+	const index_t size=10;
```
karlnapf: this is not a good unit test.
Rather have values 0,1,2,....,9 in the vector and check the sum.
This can pass for all sorts of weird bugs (just taking the mean over a subset for example)

tests/unit/mathematics/linalg/VectorMean_unittest.cc
```diff
@@ -0,0 +1,83 @@
...
+
+	float64_t result=linalg::vector_mean<linalg::Backend::EIGEN3>(a);
+
+	EXPECT_NEAR(result, 1.0, 1E-15);
+}
+
+TEST(VectorMean, Eigen3_dynamic_explicit_eigen3_backend)
+{
+	const index_t size=10;
+	Eigen::VectorXd a=Eigen::VectorXd::Constant(size, 1);
```
karlnapf: same here

tests/unit/mathematics/linalg/VectorMean_unittest.cc
```diff
@@ -0,0 +1,83 @@
...
+
+	EXPECT_NEAR(linalg::vector_mean<linalg::Backend::EIGEN3>(a), 1.0, 1E-15);
+}
+
+#endif // HAVE_EIGEN3
+
+#ifdef HAVE_VIENNACL
+
+TEST(VectorMean, viennacl_backend)
```
karlnapf: no need tu put empty tests

karlnapf:
Much nicer if the patch is small.

There are some issues with the way you used git. The rest is fine to merge :)

src/shogun/mathematics/linalg/internal/implementation/VectorMean.h
```diff
@@ -0,0 +1,134 @@
...
+	/**
+	 * Method that computes the mean of SGVectors using Eigen3
+	 *
+	 * @param vec a vector whose mean has to be computed
+	 * @return the vector mean \f$\mean_i a_i\f$
+	 */
+	static T compute(SGVector<T> vec)
+	{
+		REQUIRE(vec.vlen > 0, "Vector can not be empty!\n");
+                return (vector_sum<Backend::EIGEN3, SGVector<T> >::compute(vec)
```
lambday: Few points

- I think this method should return `float64_t` or `complex128_t`. Imagine when `T = int`.
- I think we should just create a map of the vector and call eigen3 `mean()`.
- We don't need two different methods for vector and matrices. Only one should be able to handle both. Let me know if you have questions about this :)

src/shogun/mathematics/linalg/internal/implementation/VectorMean.h
lambday: @karlnapf these should not be there. Implementing `mean` should not be too hard on viennacl. I'd rather remove this whole specialization as of now, and add this only when it has a working implementation.

IMO the only other place we did is for block-wise sum (viennacl implementations absent). Those should be gone as well

lambday:
@OXPHOS this is nice. Can you please have a look at my comments and address those?

@karlnapf for the `linalg` to work as we want it, we first need to remove the support for setting `linalg` backends shogun-wise globally. There should not be a `viennacl` build. If this thing is gone, one can easily add implementation for *only* `eigen3` or *only* `viennacl` without having to implement both each time. This would eliminate the need to have that `SG_NOTIMPLEMENTED` things (as in this PR).

src/shogun/mathematics/linalg/internal/implementation/VectorMean.h
OXPHOS: I left an error message because I didn't know how to implement ViennaCL but wanted to set a reminder for people following-up. But I agree we can do MeanEigen3 and MeanViennaCL.

tests/unit/mathematics/linalg/VectorMean_unittest.cc
OXPHOS: Sure. I'll add more tests.

src/shogun/mathematics/linalg/internal/implementation/VectorMean.h
OXPHOS: - I thought about the first point but I was worried I would lose precision for some cases. Will do if this is not a concern.
- I was wondering why we have so many Sum implementations but just to be cautious I decided to follow all the routines from VectorSum and Sum. I'm not sure about the methods of SGVector and SGMatrix but I can check out and try to squash both in one.

OXPHOS:
Thanks guys for your comments! I'll do the followings:
- Remove HAVE_EIGEN3
- Revise Mean_Eigen3.h for general SGVector/SGMatrix mean
- Revise Mean_Eigen3_unittest.cc

src/shogun/mathematics/linalg/internal/implementation/VectorMean.h
lambday: Hi @OXPHOS. Thanks for sharing your thoughts.
- Actually I think the return type should be whatever eigen `mean` returns for this specialization. But it should never be just `T`. If `T` is of any integer type, it give wrong results. Do you think it would be doable using any of C++11 magic? I am talking about that 
```c++
auto foo() -> decltype(...)
```
thing. Something like [this](https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/mathematics/linalg/internal/modules/ElementwiseOperations.h#L56)
- You're right, there are just too many of them. Although all of them serve different purposes. But yes we'd like to clean things up a bit, hopefully, in this year's GSoC :)
About handling `SGVector` and `SGMatrix`, I'd say that we can just create an `Eigen::Map` of matrix type (the vector is just a matrix with 1 column) and then call `mean()` on that method. All we need to achieve this is to specify the data pointer, number of rows and cols. The method then can take a template instead. Kinda like [this one here](https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/mathematics/linalg/internal/implementation/ElementwiseUnaryOperation.h#L95).

The reason I am excited to have just one method is that, while using the API, I won't need to specify `linalg::vector_mean(...)` or `linalg::matrix_mean(...)` each time. I'll just call `linalg::mean(...)` and expect that it knows how to handle this.

OXPHOS: Oh yes `decltype` should work! But for integers if we just do `/` the mean would still be `int` right? All I can think of is add an `if` for whatever Int type and do some tricks but it looks lame..

lambday: @OXPHOS yes it would look lame. What I was thinking that - say, have a trait.
```c++
template <typename T>
struct some_trait
{
    using Scalar = T;
};
```
and then specialize it as 
```c++
template <>
struct some_trait<int32_t> // similarly for other int types
{
    using Scalar = float64_t;
};
```
so your method signature becomes
```c++
template <class T>
typename some_trait<typename T::Scalar>::Scalar mean(T op);
// T = SGVector or SGMatrix, both have a Scalar defined
```
Yes we are not using `decltype` for this one. But this is one way we can achieve this - no ugly `if-else`. It still solves the return type problem for *all* the types without losing precision. Let me know if there are better/other ways you can think of to handle this.

I am not sure whether this is the way to go though. Say, if I have a int vector with 1s and 0s, as long as there is a single 0, I'd get 0 mean every time if the return type is just int and not some floating type (which would be the case if we didn't do type-cast while `/`-ing). Is that the expected behavior? Not sure, just thinking aloud.

OXPHOS: I'll try this way. I think returning float mean from int matrix makes more sense.

karlnapf:
@lambday can I  leave merging this to you? :)

PR #3090, opened 2016-03-17
Fix typo in tests/meta/CMakeLists.txt

Xbar:
The condition should be ```IF (EXISTS <FileName>)``` or ```IF (EXISTS <DirName>/)```.
Missing the trailing ```/``` will return FALSE for dir in cmake version 3.5.0


karlnapf:
Thanks!

Issue #3095, opened 2016-03-18
use eigen3 in kernel ridge regression


karlnapf:
1. Use eigen3 to solve the linear system. Put this into a seperate method `solve_krr_system` (to be overloaded in approximate KRR classes like #3508 #3057 #3024 )
2. Remove the Iterative Gauss Seidel method
3. Optional Add a subclass of KRR, which solves the linear system using Gauss seidel and early stopping. However, the iterative linear solver should not be part of this class,  but part of linalg. Conjugate gradient is also an option (and already implemented)

Step 1 will affect the integration test for the [meta examples](https://github.com/shogun-toolbox/shogun/tree/develop/doc/cookbook), which has to be updated accordingly.

karlnapf:
I think this will also fix the [buildbot error](http://buildbot.shogun-toolbox.org/builders/deb1%20-%20libshogun) on kernel ridge regression

sanuj:
@karlnapf I'm almost done with this. I wanted to discuss some points before I send a PR:
* All viennacl calls are in pseudo inverse method. Do we want to keep the pseudo inverse method, if yes then how can i replace the viennacl calls?
* Will there be any case when the kernel matrix is not a square matrix? For LLT, I have assumed that kernel matrix will always be a square matrix.

LLT in KRR is working, I have tested it. The integration test fail due to some slight changes in decimals of alpha so I need to update the data file. Also, there are no unit test for KRR (shall i add one?).

karlnapf:
* no need to do viennacl calls, just use eigen3 itself (if you add a LLT solve to linalg, please send a seperate PR, and then send another PR for using this linalg llt solve in KRR)
 * for KRR training, the kernel matrix is always square symmetric psd
 * Unit tests are very important, make sure to test a few corner cases

PR #3098, opened 2016-03-19
trian krr by using eigen3 llt

sanuj:
For #3095. KRR uses Eigen3's LLT now.
I haven't added any unit tests as other regression unit tests have taken reference data from some other ml toolbox. Shall I also prepare small data and run it on scikit learn to use it's results in our unit tests?
Apart from that, there are 2 tests related to KRR. One is `generated_cpp-kernel_ridge_regression` and the other one is `evaluation_cross_validation_regression`. Both are passing.

### DIFF
--- a/src/shogun/regression/KernelRidgeRegression.cpp
+++ b/src/shogun/regression/KernelRidgeRegression.cpp
@@ -11,13 +11,13 @@
 
 #include <shogun/lib/config.h>
 
-#ifdef HAVE_LAPACK
 #include <shogun/regression/KernelRidgeRegression.h>
-#include <shogun/mathematics/lapack.h>
 #include <shogun/mathematics/Math.h>
 #include <shogun/labels/RegressionLabels.h>
+#include <shogun/mathematics/eigen3.h>
 
 using namespace shogun;
+using namespace Eigen;
 
 CKernelRidgeRegression::CKernelRidgeRegression()
 : CKernelMachine()
@@ -25,7 +25,7 @@ CKernelRidgeRegression::CKernelRidgeRegression()
 	init();
 }
 
-CKernelRidgeRegression::CKernelRidgeRegression(float64_t tau, CKernel* k, CLabels* lab, ETrainingType m)
+CKernelRidgeRegression::CKernelRidgeRegression(float64_t tau, CKernel* k, CLabels* lab)
 : CKernelMachine()
 {
 	init();
@@ -33,7 +33,6 @@ CKernelRidgeRegression::CKernelRidgeRegression(float64_t tau, CKernel* k, CLabel
 	m_tau=tau;
 	set_labels(lab);
 	set_kernel(k);
-	m_train_func=m;
 }
 
 void CKernelRidgeRegression::init()
@@ -43,82 +42,27 @@ void CKernelRidgeRegression::init()
 	SG_ADD(&m_tau, "tau", "Regularization parameter", MS_AVAILABLE);
 }
 
-bool CKernelRidgeRegression::train_machine_pinv()
+bool CKernelRidgeRegression::solve_krr_system()
 {
-	// Get kernel matrix
-	SGMatrix<float64_t> kernel_matrix=kernel->get_kernel_matrix<float64_t>();
-	int32_t n = kernel_matrix.num_cols;
-	int32_t m = kernel_matrix.num_rows;
-	ASSERT(kernel_matrix.matrix && m>0 && n>0)
+	SGMatrix<float64_t> kernel_matrix(kernel->get_kernel_matrix());
+	int32_t n = kernel_matrix.num_rows;
+	SGVector<float64_t> y = ((CRegressionLabels*)m_labels)->get_labels();
 
-	for(int32_t i=0; i < n; i++)
-		kernel_matrix.matrix[i+i*n]+=m_tau;
+	for(index_t i=0; i<n; i++)
+		kernel_matrix(i,i) += m_tau;
 
-	/* re-set alphas of kernel machine */
-	m_alpha=((CRegressionLabels*) m_labels)->get_labels_copy();
+	Map<MatrixXd> eigen_kernel_matrix(kernel_matrix.matrix, n, n);
+	Map<VectorXd> eigen_alphas(m_alpha.vector, n);
+	Map<VectorXd> eigen_y(y.vector, n);
 
-	/* tell kernel machine that all alphas are needed as'support vectors' */
-	m_svs=SGVector<index_t>(m_alpha.vlen);
-	m_svs.range_fill();
-
-	if (get_alphas().vlen!=n)
+	LLT<MatrixXd> llt;
+	llt.compute(eigen_kernel_matrix);
+	if (llt.info() != Eigen::Success)
 	{
-		SG_ERROR("Number of labels does not match number of kernel"
-				" columns (num_labels=%d cols=%d\n", m_alpha.vlen, n);
+		SG_WARNING("Features covariance matrix was not positive definite\n");
+		return false;
 	}
-
-	clapack_dposv(CblasRowMajor,CblasUpper, n, 1, kernel_matrix.matrix, n,
-			m_alpha.vector, n);
-
-	return true;
-}
-
-bool CKernelRidgeRegression::train_machine_gs()
-{
-	int32_t n = kernel->get_num_vec_rhs();
-	int32_t m = kernel->get_num_vec_lhs();
-	ASSERT(m>0 && n>0)
-
-	// re-set alphas of kernel machine
-	SGVector<float64_t> b;
-	float64_t alpha_old;
-
-	b=((CRegressionLabels*) m_labels)->get_labels_copy();
-	m_alpha=((CRegressionLabels*) m_labels)->get_labels_copy();
-	m_alpha.zero();
-
-	// tell kernel machine that all alphas are needed as 'support vectors'
-	m_svs=SGVector<index_t>(m_alpha.vlen);
-	m_svs.range_fill();
-
-	if (get_alphas().vlen!=n)
-	{
-		SG_ERROR("Number of labels does not match number of kernel"
-				" columns (num_labels=%d cols=%d\n", m_alpha.vlen, n);
-	}
-
-	// Gauss-Seidel iterative method
-	float64_t sigma, err, d;
-	bool flag=true;
-	while(flag)
-	{
-		err=0.0;
-		for(int32_t i=0; i<n; i++)
-		{
-			sigma=b[i];
-			for(int32_t j=0; j<n; j++)
-				if (i!=j)
-					sigma-=kernel->kernel(j, i)*m_alpha[j];
-			alpha_old=m_alpha[i];
-			m_alpha[i]=sigma/(kernel->kernel(i, i)+m_tau);
-			d=fabs(alpha_old-m_alpha[i]);
-			if(d>err)
-				err=d;
-		}
-		if (err<=m_epsilon)
-			flag=false;
-	}
-
+	eigen_alphas = llt.solve(eigen_y);
 	return true;
 }
 
@@ -138,18 +82,22 @@ bool CKernelRidgeRegression::train_machine(CFeatures *data)
 	}
 	ASSERT(kernel && kernel->has_features())
 
-	switch (m_train_func)
+	if (m_labels->get_num_labels() != kernel->get_num_vec_rhs())
 	{
-		case PINV:
-			return train_machine_pinv();
-			break;
-		case GS:
-			return train_machine_gs();
-			break;
-		default:
-			return train_machine_pinv();
-			break;
+		SG_ERROR("Number of labels does not match number of kernel"
+			" columns (num_labels=%d cols=%d\n", m_labels->get_num_labels(), kernel->get_num_vec_rhs());
 	}
+
+	// allocate alpha vector
+	set_alphas(SGVector<float64_t>(m_labels->get_num_labels()));
+
+	if(!solve_krr_system())
+		return false;
+
+	/* tell kernel machine that all alphas are needed as'support vectors' */
+	m_svs = SGVector<index_t>(m_alpha.vlen);
+	m_svs.range_fill();
+	return true;
 }
 
 bool CKernelRidgeRegression::load(FILE* srcfile)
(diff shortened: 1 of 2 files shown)


src/shogun/regression/KernelRidgeRegression.cpp
```diff
@@ -43,82 +42,44 @@ void CKernelRidgeRegression::init()
...
-	/* re-set alphas of kernel machine */
-	m_alpha=((CRegressionLabels*) m_labels)->get_labels_copy();
-
-	/* tell kernel machine that all alphas are needed as'support vectors' */
-	m_svs=SGVector<index_t>(m_alpha.vlen);
-	m_svs.range_fill();
-
-	if (get_alphas().vlen!=n)
+    SGMatrix<float64_t> kernel_matrix(kernel->get_kernel_matrix());
+	if (kernel_matrix.num_rows!=kernel_matrix.num_cols)
```
lisitsyn: Something wrong with indentation

src/shogun/regression/KernelRidgeRegression.cpp
```diff
@@ -43,82 +42,44 @@ void CKernelRidgeRegression::init()
...
-	m_alpha=((CRegressionLabels*) m_labels)->get_labels_copy();
-
-	/* tell kernel machine that all alphas are needed as'support vectors' */
-	m_svs=SGVector<index_t>(m_alpha.vlen);
-	m_svs.range_fill();
-
-	if (get_alphas().vlen!=n)
+	SGMatrix<float64_t> kernel_matrix(kernel->get_kernel_matrix());
+	if (kernel_matrix.num_rows!=kernel_matrix.num_cols)
+	{
```
karlnapf: no {} for single line if statements.

karlnapf: I think this check should not be necessary. In fact, the algorithm (or the base class have to check) should make sure that the kernel is initialised with the training data on both sides.....

src/shogun/regression/KernelRidgeRegression.cpp
```diff
@@ -43,82 +42,44 @@ void CKernelRidgeRegression::init()
...
+		SG_ERROR("Kernel Matrix should be a square matrix.\n");
+	}
+	int32_t n = kernel_matrix.num_rows;
+	SGVector<float64_t> y = ((CRegressionLabels*)m_labels)->get_labels_copy();
```
karlnapf: why a copy?

src/shogun/regression/KernelRidgeRegression.cpp
```diff
@@ -43,82 +42,44 @@ void CKernelRidgeRegression::init()
...
+	if (y.vlen!=n)
 	{
 		SG_ERROR("Number of labels does not match number of kernel"
-				" columns (num_labels=%d cols=%d\n", m_alpha.vlen, n);
+			" columns (num_labels=%d cols=%d\n", m_alpha.vlen, n);
```
karlnapf: these checks should be done in the constructor/setter/ or even better in the base class. This method should assume that all is set up (you should put these assumptions in the doxygen docs of it)

src/shogun/regression/KernelRidgeRegression.cpp
```diff
@@ -43,82 +42,44 @@ void CKernelRidgeRegression::init()
...
 	}
 
-	clapack_dposv(CblasRowMajor,CblasUpper, n, 1, kernel_matrix.matrix, n,
-			m_alpha.vector, n);
-
-	return true;
-}
+	SGVector<float64_t> tau_vector(n);
```
karlnapf: can you do this via a loop on the diagonal?
Allocating more memory is not necessary here.

src/shogun/regression/KernelRidgeRegression.h
```diff
@@ -135,17 +122,11 @@ class CKernelRidgeRegression : public CKernelMachine
...
 		void init();
 
-		/** train regression using Gauss-Seidel iterative method
-		 *
-		 * @return whether training was successful
-		 */
-		bool train_machine_gs();
-
-		/** train regression using pinv
+		/** train regression using Eigen-3's LLT(Cholesky decomposition)
```
karlnapf: dont mention eigen3 here.
Just say it is done via Cholesky

karlnapf: btw can you change the interface a bit.
This method should be protected and just return the alphas.
Then the train method itself should prepare everything, call the method, and then set the returned alphas.
This way, we can change the solver *only* without having to re-write all the checks etc

karlnapf:
Great, thanks for the patch.
We need to do a few iterations to merge, but that is totally fine

karlnapf:
I wont merge the data for now, so that any changed can be put in there too

src/shogun/regression/KernelRidgeRegression.cpp
sanuj: because i don't want to change m_labels by calling `eigen_y = llt.solve(eigen_y);` below, so i copied them in a new vector.

karlnapf: no need for that, just store the linear solve result in m_alphas straight away

src/shogun/regression/KernelRidgeRegression.cpp
```diff
@@ -43,83 +42,27 @@ void CKernelRidgeRegression::init()
...
+SGVector<float64_t> CKernelRidgeRegression::solve_krr_system()
 {
-	// Get kernel matrix
-	SGMatrix<float64_t> kernel_matrix=kernel->get_kernel_matrix<float64_t>();
-	int32_t n = kernel_matrix.num_cols;
-	int32_t m = kernel_matrix.num_rows;
-	ASSERT(kernel_matrix.matrix && m>0 && n>0)
+	SGMatrix<float64_t> kernel_matrix(kernel->get_kernel_matrix());
+	int32_t n = kernel_matrix.num_rows;
+	SGVector<float64_t> y = ((CRegressionLabels*)m_labels)->get_labels_copy();
```
sanuj: copy is inevitable since m_alpha is NULL.

src/shogun/regression/KernelRidgeRegression.cpp
```diff
@@ -138,18 +81,21 @@ bool CKernelRidgeRegression::train_machine(CFeatures *data)
 	}
 	ASSERT(kernel && kernel->has_features())
 
-	switch (m_train_func)
+	if (m_labels->get_num_labels() != kernel->get_num_vec_rhs())
```
sanuj: adding this check here and not in constructor/setter/base class because
* in constructor or base class labels or kernel may not be defined.
* setters like `set_kernel()` and `set_labels()` can be called in any order so one of these variables can be NULL.
* `train_machine()` does all the checking when it's required.

sanuj:
@karlnapf updated :) data pr remains the same.

src/shogun/regression/KernelRidgeRegression.cpp
karlnapf: Initialise in base class once the training features are passed

src/shogun/regression/KernelRidgeRegression.cpp
karlnapf: What about using a setter for this, that is also called from the constructors ...The setter might also initialise the alphas
IF you want them here, take all checks out of the matrix inversion method at least

karlnapf:
Ok will merge data and then this one once travis is green.
Thanks!

karlnapf:
I restarted travis after the data merge, it was failing....

sanuj:
It will take me some time to update the changes.... i m away from my laptop.

sanuj:
@karlnapf if i add a setter for m_alpha then why do we need to return the alpha vector in `solve_krr_system()`? The vector can directly be updated inside the method itself.

karlnapf:
I think the method can just take a preallocated alpha vector as a parameter. Then populate and return it. The train method then calls the setter. This way everything is very modular

src/shogun/regression/KernelRidgeRegression.cpp
```diff
@@ -43,83 +42,27 @@ void CKernelRidgeRegression::init()
 	SG_ADD(&m_tau, "tau", "Regularization parameter", MS_AVAILABLE);
 }
 
-bool CKernelRidgeRegression::train_machine_pinv()
+SGVector<float64_t> CKernelRidgeRegression::solve_krr_system()
```
karlnapf: take alpha vector as input

src/shogun/regression/KernelRidgeRegression.cpp
```diff
@@ -43,86 +42,60 @@ void CKernelRidgeRegression::init()
...
 	return true;
 }
 
-bool CKernelRidgeRegression::train_machine_gs()
+bool CKernelRidgeRegression::train_machine(CFeatures *data)
 {
-	int32_t n = kernel->get_num_vec_rhs();
-	int32_t m = kernel->get_num_vec_lhs();
-	ASSERT(m>0 && n>0)
+	check_and_set_alphas(data);
```
sanuj: I'm calling this method here and not in the constructors because checks should only run when both labels and kernel are set which is not necessary in a constructor.

src/shogun/regression/KernelRidgeRegression.cpp
```diff
@@ -43,86 +42,60 @@ void CKernelRidgeRegression::init()
...
-	return true;
+bool CKernelRidgeRegression::save(FILE* dstfile)
+{
+	SG_SET_LOCALE_C;
+	SG_RESET_LOCALE;
+	return false;
 }
 
-bool CKernelRidgeRegression::train_machine(CFeatures *data)
+void CKernelRidgeRegression::check_and_set_alphas(CFeatures *data)
```
karlnapf: Why a new method for this?
This can be just done in the base class `CKernelMachine` and in the setter for features of it

src/shogun/regression/KernelRidgeRegression.h
```diff
@@ -132,20 +119,23 @@ class CKernelRidgeRegression : public CKernelMachine
...
 		virtual bool train_machine(CFeatures* data=NULL);
 
-	private:
-		void init();
-
-		/** train regression using Gauss-Seidel iterative method
+		/** Train regression using Cholesky decomposition
 		 *
-		 * @return whether training was successful
+		 * @param alphas vector of alpha already initialized
```
karlnapf: If the alphas are already prepared, this method can just fill them and doesnt need to taken them as an argument

karlnapf: BTW and for future comments: the vector here it is not initialized, but pre-allocated

src/shogun/regression/KernelRidgeRegression.cpp
karlnapf: I thought about this again.
Just allocate m_alphas in the train method before you call the solve method. The solve method assumes that memory is already allocated .... done

src/shogun/regression/KernelRidgeRegression.cpp
sanuj: @karlnapf if you look at line [73](https://github.com/shogun-toolbox/shogun/pull/3098/files#diff-429ee3465120e3c46c4432c53977a7cfR73), the first check can't go in setters/constructors for this use-case:
```
krr = KernelRidgeRegression()
krr.set_kernel(kernel)
krr.set_labels(labels)
krr.train()
```
The second check is explicit for krr. Last check can only be performed when `m_labels` is not NULL.

sanuj: The matrix inversion method only has one check which is dependent on eigen's llt so can't be taken out of it.

karlnapf: I see

sanuj:
I have updated, looking forward to your comments :)

src/shogun/regression/KernelRidgeRegression.cpp
```diff
@@ -43,82 +42,29 @@ void CKernelRidgeRegression::init()
...
-	/* tell kernel machine that all alphas are needed as'support vectors' */
-	m_svs=SGVector<index_t>(m_alpha.vlen);
-	m_svs.range_fill();
-
-	if (get_alphas().vlen!=n)
+	for(index_t i=0; i<n; i++)
 	{
-		SG_ERROR("Number of labels does not match number of kernel"
-				" columns (num_labels=%d cols=%d\n", m_alpha.vlen, n);
+		m_alpha[i] = y[i];
```
karlnapf: this is not really needed, you can just have another eigen3 map

src/shogun/regression/KernelRidgeRegression.cpp
```diff
@@ -43,82 +42,29 @@ void CKernelRidgeRegression::init()
...
-			m_alpha[i]=sigma/(kernel->kernel(i, i)+m_tau);
-			d=fabs(alpha_old-m_alpha[i]);
-			if(d>err)
-				err=d;
-		}
-		if (err<=m_epsilon)
-			flag=false;
-	}
-
+	eigen_alphas = llt.solve(eigen_alphas);
```
karlnapf: eigen_alphas = llt.solve(eigen_y)

src/shogun/regression/KernelRidgeRegression.cpp
```diff
@@ -138,18 +84,21 @@ bool CKernelRidgeRegression::train_machine(CFeatures *data)
...
-			return train_machine_gs();
-			break;
-		default:
-			return train_machine_pinv();
-			break;
+		SG_ERROR("Number of labels does not match number of kernel"
+			" columns (num_labels=%d cols=%d\n", m_labels->get_num_labels(), kernel->get_num_vec_rhs());
 	}
+
+	set_alphas(SGVector<float64_t>(m_labels->get_num_labels()));
```
karlnapf: good!
maybe comment "allocate alpha vector"

src/shogun/regression/KernelRidgeRegression.h
```diff
@@ -132,20 +119,16 @@ class CKernelRidgeRegression : public CKernelMachine
 		 */
 		virtual bool train_machine(CFeatures* data=NULL);
 
-	private:
-		void init();
-
-		/** train regression using Gauss-Seidel iterative method
+		/** Train regression using Cholesky decomposition.
+		 * Memory for vector of alphas should be allocated before calling this method.
```
karlnapf: Assumes that m_alphas is already allocated

karlnapf:
Ok, we are there. Last two minor changes and we can merge.

sanuj:
@karlnapf Travis is green!

karlnapf:
Party:)

Issue #3074, opened 2016-03-15
add linear solve to linalg


karlnapf:
* psd matrix (LLT) -- most important
 * Free form (QR)
 * least squares (SVD)

This would also involve investigating how for example np.linalg.solve does the same job to get inspiration.

Obviously, good documentation about the cases is key.

karlnapf:
All linear solves should have the possibility to *precompute* the linear solve. That is factorize the matrix and offer a method that computes the solve given the output of the factors. See numpy again for this. For example, for LLT, this will be a triangular solve (one can do this with eigen). For the others it works imilarly

c4goldsw:
Hey there!  I just found out about GSoC on Friday and I'm a little caught up with school work.  I'm in the second half of a specialist linear algebra course and this would be a great way to put my knowledge to use.  After tomorrow, I'll have a lot more time on my hands to get started on this.

karlnapf:
Great!

PR #3101, opened 2016-03-20
LLT linear solver using eigen3

curiousguy13:
As given in #3074, implemented LLT linear solver in linalg using eigen3 for Dense Matrices to solve linear systems of the form (Ax=b) with a method to precompute the cholesky factor i.e. the lower triangular matrix L and a triangular solve method to solve the linear system given L and b as the input.


karlnapf:
This is a great initiative, but we mostly want this solver to be part of linalg.
@lambday can you comment a bit

vigsterkr:
closing...

PR #3103, opened 2016-03-20
added first version of bigtesting framework

lambday:
WIP merging [flash](https://github.com/lambday/flash) with Shogun. Code is not following Shogun's standard yet.


lambday:
feature branch. merging.

PR #3112, opened 2016-03-22
add cookbook page for linear svm

sanuj:
Looking forward to feedback. I'll send the data PR once this gets finalized.


karlnapf:
Nice! :)

doc/cookbook/source/examples/classifier/linear_svm.rst
```diff
@@ -0,0 +1,60 @@
+=============================
+Linear Support Vector Machine
+=============================
+
+Linear Support Vector Machine is a binary classifier which finds a hyper-plane such that the margins between the two classes are maximized. These margins can be described by the following equations:
```
karlnapf: Missing "The"

doc/cookbook/source/examples/classifier/linear_svm.rst
```diff
@@ -0,0 +1,60 @@
...
+
+.. math::
+    {\bf w}^\top{\bf x} + b = 1
```
karlnapf: in terms of math, please only state the loss function that is minimized

doc/cookbook/source/examples/classifier/linear_svm.rst
```diff
@@ -0,0 +1,60 @@
...
+
+    {\bf w}^\top{\bf x} + b = -1
+
+where :math:`{\bf w}` is vector of weights, :math:`{\bf x}` is feature vector and :math:`b` is bias.
+
+The distance between these two margins is :math:`\frac{2}{\|\mathbf{w}\|}`, so we want to minimize :math:`\|\mathbf{w}\|`. By using Lagrange Multipliers, we can show the solution to be:
```
karlnapf: Avoid Lagrange multipliers. Keep in mind this is not to explain the algorithm or math, but just show how to use the API in Shogun. You can say "The solution takes the form ...."

doc/cookbook/source/examples/classifier/linear_svm.rst
karlnapf: You could also write the minimization problem. But we only want one major equation

doc/cookbook/source/examples/classifier/linear_svm.rst
```diff
@@ -0,0 +1,60 @@
...
+
+.. math::
+
+    \mathbf{w} = \sum_i \alpha_i y_i \mathbf{x}_i
+
+here most :math:`\alpha_i` turn out to be zero, which means that the solution is a sparse linear combination of the training features. The training features for which :math:`\alpha_i` are not zero are called support vectors.
```
karlnapf: No need to be this lengthy. Just say that the alphas are sparse and thats it

doc/cookbook/source/examples/classifier/linear_svm.rst
```diff
@@ -0,0 +1,60 @@
...
+
+-------
+Example
+-------
+
+Imagine we have files with training and test data. We create CDenseFeatures (here 64 bit floats aka RealFeatures) and :sgclass:`CBinaryLabels` as
+
+.. sgexample:: linear_svm.sg:create_features
+
+In order to run :sgclass:`CLibLinear`, we need to initialize some parameters like regularization coefficient and epsilon.
```
karlnapf: C is coming from the loss function, so you can refer to it in math mode. `:math:`C`
epsilon is the residual convergence parameter of the solver, state that.
Also state that there are other solver types, referring to the class docs of CLibLinear

karlnapf: L2R_L2LOSS_SVC stands for something, state that here as well

doc/cookbook/source/examples/classifier/linear_svm.rst
```diff
@@ -0,0 +1,60 @@
...
+
+.. sgexample:: linear_svm.sg:set_parameters
+
+We create an instance of the :sgclass:`CLibLinear` classifier by passing it regularization coefficient, features and labels. We also need to set a solver and epsilon.
```
karlnapf: The second sentence should be covered above

doc/cookbook/source/examples/classifier/linear_svm.rst
```diff
@@ -0,0 +1,60 @@
...
+
+.. sgexample:: linear_svm.sg:create_instance
+
+Then we train and apply it to test data, which here gives :sgclass:`CBinaryLabels`.
+
+.. sgexample:: linear_svm.sg:train_and_apply
+
+We can extract :math:`{\bf w}` and :math:`b`.
```
karlnapf: rather than extracting b, one would extact the alphas here.
Is the b not always set to zero for svm?

doc/cookbook/source/examples/classifier/linear_svm.rst
```diff
@@ -0,0 +1,60 @@
...
+.. sgexample:: linear_svm.sg:evaluate_accuracy
+
+----------
+References
+----------
+:wiki:`Support_vector_machine`
+
+:wiki:`Lagrange_multiplier`
+
+.. bibliography:: ../../references.bib
```
karlnapf: we should cite the Libliear paper here and reference the website

doc/cookbook/source/examples/classifier/linear_svm.rst
```diff
@@ -0,0 +1,60 @@
+=============================
+Linear Support Vector Machine
```
karlnapf: Mention that this is Liblinear specific

karlnapf:
Needs one or two iterations, but already looks nice. Thanks for that! :)

karlnapf:
some travis builds fail. I actually never used enums from meta examples. It might be that you have to replace it with the int value for now -- and I need to fix this

doc/cookbook/source/examples/classifier/linear_svm.rst
sanuj: @karlnapf There is no function to extract the alphas. `CLibLinear` is derived from `CLinearMachine` and `get_alpha()` is available in `CKernelMachine`. In the ipython notebook for SVMs `w` and `b` are extracted to draw the decision boundary and `b` is non zero.

doc/cookbook/source/examples/classifier/linear_svm.rst
sanuj: to reference the website, shall i just put the link? I think that's not the correct way to do it.

sanuj:
I tried `Int L2R_L2LOSS_SVC = 2`. Its not working. generate.py is giving some error. `Real L2R_L2LOSS_SVC = 2` generates the cookbook but then cpp complains.

doc/cookbook/source/examples/classifier/linear_svm.rst
karlnapf: Interesting.....We definitely want the alphas for liblinear -- it is an svm after all. Where are they stored? *are* they stored somewhere? Or just w?
Can you investigate and open an issue if we cannot extract them?
But fine for the cookbook here

doc/cookbook/source/references.bib
```diff
@@ -4,3 +4,20 @@ @book{barber2012bayesian
...
+}
+
+@article{fan2008liblinear,
+  title={LIBLINEAR: A Library for Large Linear Classification},
+  author={Fan, Rong-En and Chang, Kai-Wei and Hsieh, Cho-Jui and Wang, Xiang-Rui and Lin, Chih-Jen},
+  journal={The Journal of Machine Learning Research},
+  volume={9},
+  pages={1871-1874},
+  year={2008},
+  publisher={JMLR.org}
```
karlnapf: remove publisher

doc/cookbook/source/references.bib
```diff
@@ -4,3 +4,20 @@ @book{barber2012bayesian
...
+@book{cristianini2000introduction,
+  title={An Introduction To Support Vector Machines And Other Kernel-Based Learning Methods},
+  author={N. Cristianini and J. Shawe-Taylor},
+  year={2000},
+  publisher={Cambridge University Press}
+}
+
+@article{fan2008liblinear,
+  title={LIBLINEAR: A Library for Large Linear Classification},
+  author={Fan, Rong-En and Chang, Kai-Wei and Hsieh, Cho-Jui and Wang, Xiang-Rui and Lin, Chih-Jen},
```
karlnapf: Can you abbreviate first names. See the bibtex entry I put in initially

doc/cookbook/source/references.bib
```diff
@@ -4,3 +4,20 @@ @book{barber2012bayesian
...
+  journal={The Journal of Machine Learning Research},
+  volume={9},
+  pages={1871-1874},
```
karlnapf: usually pages are separated with double dash, --

doc/cookbook/source/references.bib
```diff
@@ -4,3 +4,20 @@ @book{barber2012bayesian
   year={2012},
   publisher={Cambridge University Press}
 }
+
+@book{cristianini2000introduction,
```
karlnapf: this is good

examples/meta/src/classifier/linear_svm.sg
```diff
@@ -0,0 +1,40 @@
...
+#![create_features]
+
+#![set_parameters]
+Real C = 1
+Real epsilon = 0.001
+#![set_parameters]
+
+#![create_instance]
+LibLinear svm(C, features_train, labels_train)
+svm.set_liblinear_solver_type(enum LIBLINEAR_SOLVER_TYPE.L2R_L2LOSS_SVC)
```
karlnapf: this works now?

doc/cookbook/source/examples/classifier/linear_svm.rst
```diff
@@ -0,0 +1,61 @@
...
+=============================
+
+Linear Support Vector Machine is a binary classifier which finds a hyper-plane such that the margins between the two classes are maximized. The loss function that needs to be minimized is:
+
+.. math::
+
+    \min_{\bf w} \frac{1}{2}{\bf w}^\top{\bf w} + C\sum_{i=1}^{N}\xi({\bf w};{\bf x_i}, y_i)
+
+where :math:`{\bf w}` is vector of weights, :math:`{\bf x_i}` is feature vector, :math:`y_i` is the corresponding label, :math:`C>0` is a penalty parameter and :math:`N` is the number of training samples.
```
karlnapf: you need to say what the \xi is

doc/cookbook/source/examples/classifier/linear_svm.rst
```diff
@@ -0,0 +1,61 @@
...
+
+The solution takes the following form:
+
+.. math::
+
+    \mathbf{w} = \sum_i \alpha_i y_i \mathbf{x}_i
+
+here :math:`\alpha_i` are sparse.
```
karlnapf: where

doc/cookbook/source/examples/classifier/linear_svm.rst
```diff
@@ -0,0 +1,61 @@
...
+
+Imagine we have files with training and test data. We create CDenseFeatures (here 64 bit floats aka RealFeatures) and :sgclass:`CBinaryLabels` as
+
+.. sgexample:: linear_svm.sg:create_features
+
+In order to run :sgclass:`CLibLinear`, we need to initialize some parameters like :math:`C` and epsilon which is the residual convergence parameter of the solver.
+
+.. sgexample:: linear_svm.sg:set_parameters
+
+We create an instance of the :sgclass:`CLibLinear` classifier by passing it regularization coefficient, features and labels. We need to set the solver type to L2R_L2LOSS_SVC which stands for L2 regularized SVM with L2-loss. There are many other solver types in :sgclass:`CLibLinear` to choose from.
```
karlnapf: dont explicitly mention the L2R_L2LOSS_SVC here but just say "We here set the solver type to L2 regularized classification. There are many other solver types in :sgclass:`CLibLinear` to choose from."

doc/cookbook/source/examples/classifier/linear_svm.rst
```diff
@@ -0,0 +1,61 @@
...
+.. sgexample:: linear_svm.sg:evaluate_accuracy
+
+----------
+References
+----------
+:wiki:`Support_vector_machine`
+
+:wiki:`Lagrange_multiplier`
+
+http://www.csie.ntu.edu.tw/~cjlin/liblinear/
```
karlnapf: Can you make this a link and do something like
[LibLinear website](http://www.csie.ntu.edu.tw/~cjlin/liblinear/)

karlnapf:
Ok, this is almost ready to be merged. A few minor updates. Let me fix the enum issue and then you can rebase and we can merge

examples/meta/src/classifier/linear_svm.sg
sorig: See #3120

karlnapf:
Looks like we need to do the same for [ruby](https://travis-ci.org/shogun-toolbox/shogun/jobs/118819598)
@sanuj can you send a separate patch for that, and then rebase against it here.
You can test this PR locally via enabling ruby ´-DRubyModular=On`

sanuj:
@karlnapf I pushed a temporary commit (ruby modular doesn't build on my local) to check if changing the json file fixes the problem but it's not working. See [here](https://travis-ci.org/shogun-toolbox/shogun/jobs/118973198).

karlnapf:
Just restarted travis. If it still fails, try to rebase against latest develop. I just merged your other PR. Thanks for that!

sanuj:
@karlnapf build fails after rebasing :(

karlnapf:
will check later

sorig:
This should fix it: #3126

sorig:
I'm pretty sure Travis is going to fail for Lua, in which case the fix should be #3127

karlnapf:
Ah finally :)

Issue #3107, opened 2016-03-22
Fix broken LDA unit test


OXPHOS:
LDAtest will be enabled once HAVE_EIGEN3 flag is removed from LDA_unittest.cc
And running LDAtest will lead to the failure of the test. 

> 260/379 Test #256: unit-LDATest .......................................................***Failed    0.04 sec
> Note: Google Test filter = LDATest.*
> [==========] Running 4 tests from 1 test case.
> [----------] Global test environment set-up.
> [----------] 4 tests from LDATest
> [ RUN      ] LDATest.CheckEigenvectors_FLD
> [       OK ] LDATest.CheckEigenvectors_FLD (1 ms)
> [ RUN      ] LDATest.CheckProjection_FLD
> /opt/shogun/tests/unit/classifier/LDA_unittest.cc:141: Failure
> The difference between -304.80621346 and projection_FLD[0] is 6.4366247722805383, which exceeds epsilon, where
> -304.80621346 evaluates to -304.80621345999998,
> projection_FLD[0] evaluates to -311.24283823228052, and
> epsilon evaluates to 1e-08.
> /opt/shogun/tests/unit/classifier/LDA_unittest.cc:142: Failure
> The difference between -281.12285949 and projection_FLD[1] is 6.4366247711261053, which exceeds epsilon, where
> -281.12285949 evaluates to -281.12285949,
> projection_FLD[1] evaluates to -287.5594842611261, and
> epsilon evaluates to 1e-08.
> /opt/shogun/tests/unit/classifier/LDA_unittest.cc:143: Failure
> The difference between -228.60266985 and projection_FLD[2] is 6.4366247721223147, which exceeds epsilon, where
> -228.60266985 evaluates to -228.60266985000001,
> projection_FLD[2] evaluates to -235.03929462212233, and
> epsilon evaluates to 1e-08.
> /opt/shogun/tests/unit/classifier/LDA_unittest.cc:144: Failure
> The difference between -300.38571766 and projection_FLD[3] is 6.4366247723569927, which exceeds epsilon, where
> -300.38571766 evaluates to -300.38571766000001,
> projection_FLD[3] evaluates to -306.822342432357, and
> epsilon evaluates to 1e-08.
> /opt/shogun/tests/unit/classifier/LDA_unittest.cc:145: Failure
> The difference between -258.89964153 and projection_FLD[4] is 6.4366247778518186, which exceeds epsilon, where
> -258.89964153 evaluates to -258.89964153,
[...]

PR #3113, opened 2016-03-23
Fix broken LDA unit test

theaverageguy:
Fixes https://github.com/shogun-toolbox/shogun/issues/3107

### DIFF
--- a/tests/unit/classifier/LDA_unittest.cc
+++ b/tests/unit/classifier/LDA_unittest.cc
@@ -35,7 +35,7 @@
 
 #ifdef HAVE_EIGEN3
 using namespace shogun;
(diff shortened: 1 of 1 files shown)


karlnapf:
This is not a fix

karlnapf:
I will merge as this test is broken
@OXPHOS  if you rebase, then your linalg patch should work

theaverageguy:
Could you help me with this @karlnapf ? 
I believe I didn't understand it properly. The link you shared did disable the class.

karlnapf:
Sure, it is now disabled. Better would be to *fix* the test and activate it again. Next step. Find out why it failed in the first place.

Issue #2247, opened 2014-05-22
bug in CProbitLikelihood::get_log_probability_derivative_f


yorkerlin:
The bug will occur due to the numerical issue, which comes from GPML 3.1
This bug is fixed in GPML 3.4.
I will fix it this week.

yorkerlin:
Done

PR #3121, opened 2016-03-26
added normal_cdf

lambday:
@karlnapf sending a PR because I wanted you to have a look. `erfc(...)` is a C-function so no C++11 is needed for this one.

### DIFF
--- a/src/shogun/mathematics/Statistics.cpp
+++ b/src/shogun/mathematics/Statistics.cpp
@@ -15,8 +15,9 @@
 #include <shogun/lib/SGVector.h>
 #include <shogun/lib/SGSparseMatrix.h>
 #include <shogun/lib/SGSparseVector.h>
-
+#include <cmath>
 #include <shogun/mathematics/eigen3.h>
+
 using namespace Eigen;
 
 using namespace shogun;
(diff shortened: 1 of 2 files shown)


src/shogun/mathematics/Statistics.cpp
```diff
@@ -573,9 +574,7 @@ float64_t CStatistics::erfc8_weighted_sum(float64_t x)
 
 float64_t CStatistics::normal_cdf(float64_t x, float64_t std_dev)
 {
-	SG_SERROR("NOT IMPLEMENTED");
-	return 0;
-//	return 0.5*(error_function_complement(-x/std_dev/1.41421356237309514547));
```
yorkerlin: Make sure that you do not re-introduce the same bug I fixed before.
https://github.com/lambday/shogun/commit/b7e98db219d909ee0c61b359ee83e1687a79368e

yorkerlin: please add  the unit tests at https://github.com/lambday/shogun/commit/b7e98db219d909ee0c61b359ee83e1687a79368e

lambday: @yorkerlin I think since this is in the standard, we're good to go. Have a look at the discussion [here](http://stackoverflow.com/questions/631629/erfx-and-math-h) and the implementation [here](http://www.netlib.org/fdlibm/s_erf.c). Looks like they take quite a lot of measures for different cases. I'll add the unit tests though.

One question : are these numbers (for the `lnormal_cdf` case)
```c++
t[0]=0.197197002817524946749;
t[1]=0.0056939283141627128337;
t[2]=11.17348207854067787537;
t[3]=0.513878566283254256675;
t[4]=4.34696415708135575073;
t[5]=-8.69392831416271150147;
t[6]=-17.38785662832542300293;
t[7]=-34.42874909956949380785;
t[8]=-67.46964157081356461276;
t[9]=-136.67410392703391153191;
```
just chosen randomly while writing the unit-test or there is some source that I should be looking at?

yorkerlin: @lambday 
I found a bug when I worked on Gaussian processes.
These inputs are used to re-produce the bug.

yorkerlin: Ref
https://github.com/shogun-toolbox/shogun/issues/2247

PR #3122, opened 2016-03-27
fix enum rule for ruby

sanuj:
@karlnapf Couldn't build it on my local. Got this error:
```
Linking CXX shared module modshogun.so
/usr/bin/ld: /usr/local/lib/libruby-static.a(array.o): relocation R_X86_64_PC32 against symbol `rb_ary_free' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: final link failed: Bad value
```


karlnapf:
Thanks, should be good. Travis error was unrelated. Merging :)

Issue #3118, opened 2016-03-24
Remove CStatistic::mean/variance/covariance 


karlnapf:
And replace all their usage with linalg implementation.
In addition, respect Shoguns way of defining vectors, which is as column vectors, not row vectors. A couple of unit tests and other places in Shogun depend on this.

Mean/Std/Var:
 * partly already in linalg
 * should be possible to compute elementwise, rowwise, and columnwise

Covariance:
 * not in linalg yet, but should re-use mean computation

See also #3117

c4goldsw:
I'm trying to find a standard deviation or variance function in **linalg**, but I'm unable to find one - could you please point me in the right direction? Or, have I misunderstood you, in that it has yet to be implemented (because of "partly already in **linalg**")?

c4goldsw:
Also, I've noticed that there are different implementations of functions for vectors and matrices - I've replaced all usages of CStatistics::mean() with linalg::mean() (#3096), but I haven't done anything with CStatistics::matrix_mean() .  Does this issue involving replacing both or only vector implementations?

karlnapf:
The idea is to
1. Add all functionality for mean/variance/etc to linalg. Those that are not yet there, such as variance
2. Make Shogun use them rather than CStatistics
3. Remove them from CStatistics

Start one by one. As a first step, you could grep for `CStatistics::matrix_mean` etc and replace them with linalg calls

PR #3123, opened 2016-03-28
Refactoring for issue #3118

c4goldsw:
Refactored code in reference to #3118 .  Made tests where needed, or used existing tests to see if functions could still work after being refactored.


c4goldsw:
@karlnapf 4 out of the 6 build failures are being caused by the inability to find the symbol "ZN6shogun11CStatistics8varianceENS_8SGVectorIdEE" when trying to compile builds for the modular interfaces.  How can this be resolved?  The other two are being caused by an example test failure, which should be easy to fix.

c4goldsw:
Closing for fixing.

karlnapf:
Hi
Please never send PRs of this size. 2-3 files max, as few lines of code as possible.
We can re-factor one by one, which is much less error prone

c4goldsw:
@karlnapf I'll keep that in  mind going forward then.  I'll stash the changes I've made and submit PRs individually.  I would still appreciate it if you could look at the build errors, as most of them were caused by the same thing, as mentioned in my second comment - I don't know how to resolve the issue.

karlnapf:
First patch should just add the row/colwise addition -- with unit test. Then the patches with using it should come

PR #3124, opened 2016-03-28
parallelize Kmeans computations

Saurabh7:
A way to parallelize computations in KMeans using OpenMP .
We can discuss here.

### DIFF
--- a/src/shogun/clustering/KMeansLloydImpl.cpp
+++ b/src/shogun/clustering/KMeansLloydImpl.cpp
@@ -54,7 +54,10 @@ void CKMeansLloydImpl::Lloyd_KMeans(int32_t k, CDistance* distance, int32_t max_
 		rhs_mus->copy_feature_matrix(mus);
 
 		distance->precompute_rhs();
-		
+
+#pragma omp parallel for firstprivate(lhs_size, dimensions, k) \
+		shared(mus, cluster_assignments, weights_set) \
+		reduction(+:changed) if (!fixed_centers)
 		/* Assigment step : Assign each point to nearest cluster */
 		for (int32_t i=0; i<lhs_size; i++)
 		{ 
@@ -77,7 +80,9 @@ void CKMeansLloydImpl::Lloyd_KMeans(int32_t k, CDistance* distance, int32_t max_
 			if (min_cluster!=cluster_assignments_i)
 			{
 				changed++;
+#pragma omp atomic
 				weights_set[min_cluster]+= 1.0;
+#pragma omp atomic
 				weights_set[cluster_assignments_i]-= 1.0;
 
 				if(fixed_centers)


Saurabh7:
I think we already have CMake setting omp flags if suitable compilers. right?

Btw here are the benchmark framework results before and after:

```
                 mlpack  matlab    scikit  mlpy    shogun  weka 
mnist               -       -  79.298909    -  47.606928     - 
corel-histogram     -       -   1.316512    -  2.281468     - 
USCensus1990        -       -  0.009211     -  0.075297     - 
isolet              -       -  5.541788     -  4.694107     - 
covtype             -       -  23.472009    -  31.512816     - 
```
After:
```
                  mlpack  matlab     scikit  mlpy     shogun  weka 
mnist                  -       -  78.776041     -  28.317168     - 
corel-histogram        -       -   1.320681     -   1.644450     - 
USCensus1990           -       -  0.009345      -  0.039751     - 
isolet                 -       -  5.537824      -  3.577482     - 
covtype                -       -  23.354638     -  22.171795     - 
```

karlnapf:
Wow nice!
How many cores did you use?

Can you explain a bit on how you parallelized? I see the code, but what is the rationale? Are the operations in KMeans thread safe (important)? 

Why is Kmeans on one core still so much slower on certain datasets?

Yes openmp is handled in Cmake

Saurabh7:
yeah :)
My machine has 4 , so omp should be using all by default.

- The for loop updates each sample for a given iteration. I am parallelizing this loop.
- lhs_size, dimensions are already initialized so i made them firstprivate.
- Shared variables like mus, weights_set, assignments i have made them shared for all threads.
- To avoid problematic updating of these shared variables,  i have given atomic directive for their read/write. This will make them thread safe.
- `changed` variable is the no. of assignments changed. I am applying (+) reduction to this so it will be summed at last for all threads to give overall number of change.
- parallel for `fixed center` version is disabled as it requires much more complications and atomic pragmas.

For the `covtype` dataset its slow. How it is different from the other large dataset is that it has lower dimensions (52) and huge number of samples 500k ,( others have ~500 dimension) . So this may be the `linalg::dot` thing poping up again, since that dataset involves lot of computations. I will try to investigate with other and update on this.

Saurabh7:
The script in benchmark framework was actually bugged, which made shogun resort to random intialization. I have sent a PR: https://github.com/zoq/benchmarks/pull/13
This improves the results for that dataset, its still a bit slow and this happens only for this set of initial centroids.
before:
  ```
                mlpack  matlab     scikit  mlpy     shogun  weka 
mnist                  -       -  79.235335    -   41.759917     - 
corel-histogram        -       -   1.318852    -    1.895212     - 
USCensus1990           -       -   0.010071     -    0.009292     - 
isolet                 -       -   5.534171     -  3.003076     - 
covtype                -       -  23.331519     -  25.492907     - 

```
after openmp :
 ```
                mlpack  matlab     scikit  mlpy     shogun  weka 
mnist                  -       -  79.060508     -  23.927803     - 
corel-histogram        -       -   1.320199     -  1.267014     - 
USCensus1990           -       -  0.009464      -  0.012043     - 
isolet                 -       -  5.642191      -  1.620542     - 
covtype                -       -  23.268746     -  17.921144     - 

```

karlnapf:
What about adding a benchmark for the k=7 as well?

karlnapf:
Will merge. Nice one!

Saurabh7:
thanks ! 
k=7 is the one already there , did you mean another k?
Since they test only for manual initialization (for fairness between all the libraries) , current data for that dataset is only for 7 centers
There are other issues too, i cannot set tol=0.0 just for comparing scikit and shogun, since other libraries also use tolerance criteria.
So its not going to be completely fair anyways i believe but a good estimate.

karlnapf:
BTW this should be added to NEWS

Saurabh7:
Yes sure will add it in next patch . I was thinking about refactor the Lloyd and mini batch into separate classes. what do you think  ?

vigsterkr:
@Saurabh7 yeah sounds like a plan. see ELKI's k-means family: http://elki.dbs.ifi.lmu.de/wiki/Algorithms

Saurabh7:
Alright Just found out our kmeans++ initialization is doing nothing, just using the first `k` samples from dataset as centers ;)
Will fix that first then.

PR #3127, opened 2016-03-29
Meta-Language: Fix Lua enum rule

sorig:
I think this is the last one. The enums have already been tested in Python, R, Octave, and Java.


karlnapf:
It actually worked before, but this seems cleaner. 
Thanks!

karlnapf:
Ah! Lua is currently disabled -- overloading Shogun methods doesnt work in there. This is why the build did not fail. So this patch then solves the problem we did not have before :)

PR #3128, opened 2016-03-29
removed compilation errors, added libcdf

lambday:
- added incomplete_beta from libcdf (wrapper in CStatistics)
- also, added unit-tests, asserted against values computed from octave

@karlnapf please have a look!

### DIFF
--- a/examples/undocumented/libshogun/evaluation_cross_validation_classification.cpp
+++ b/examples/undocumented/libshogun/evaluation_cross_validation_classification.cpp
@@ -105,7 +105,7 @@ void test_cross_validation()
 			splitting, eval_crit);
 
 	cross->set_num_runs(10);
-	cross->set_conf_int_alpha(0.05);
+//	cross->set_conf_int_alpha(0.05);
 
 	/* actual evaluation */
 	CCrossValidationResult* result=(CCrossValidationResult*)cross->evaluate();
--- a/examples/undocumented/libshogun/mathematics_confidence_intervals.cpp
+++ b/examples/undocumented/libshogun/mathematics_confidence_intervals.cpp
@@ -17,6 +17,7 @@ using namespace shogun;
 
 void test()
 {
+	/*
 	SGVector<float64_t> data(10);
 	SGVector<float64_t>::range_fill_vector(data.vector, data.vlen, 1.0);
 
@@ -26,6 +27,7 @@ void test()
 
 	SG_SPRINT("sample mean: %f. True mean lies in [%f,%f] with %f%%\n",
 			mean, low, up, 100*(1-error_prob));
+	*/
 }
 
 int main(int argc, char **argv)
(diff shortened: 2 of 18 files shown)


src/shogun/mathematics/Statistics.cpp
```diff
@@ -754,17 +755,43 @@ float64_t CStatistics::inverse_gamma_cdf(float64_t p, float64_t a,
...
-//	SG_SERROR("NOT IMPLEMENTED");
-//	return 0;
-//}
+float64_t CStatistics::incomplete_beta(float64_t a, float64_t b, float64_t x)
+{
+	float64_t y=1-x;
+	float64_t result=0;
+	float64_t result2=0;
+	int error_code=0;
+	beta_inc(&a, &b, &x, &y, &result, &result2, &error_code);
```
lambday: @karlnapf there is a look-up table or sort for some predefined a, b and x in cdflib. We can use that. Also, maybe for  `x < 0.5`, we can use symmetry - will check if that is faster.

examples/undocumented/libshogun/evaluation_cross_validation_classification.cpp
```diff
@@ -105,7 +105,7 @@ void test_cross_validation()
 			splitting, eval_crit);
 
 	cross->set_num_runs(10);
-	cross->set_conf_int_alpha(0.05);
+//	cross->set_conf_int_alpha(0.05);
```
karlnapf: These guys can be deleted. They are useless anyways :)

examples/undocumented/libshogun/mathematics_confidence_intervals.cpp
```diff
@@ -17,6 +17,7 @@ using namespace shogun;
 
 void test()
 {
```
karlnapf: the whole file should go :)

examples/undocumented/libshogun/modelselection_combined_kernel_sub_parameters.cpp
```diff
@@ -162,7 +162,7 @@ void modelselection_combined_kernel()
 
 	/* larger number of runs to have tighter confidence intervals */
```
karlnapf: the comment should be changed to "to have less noise"

src/shogun/mathematics/Statistics.cpp
```diff
@@ -754,17 +755,43 @@ float64_t CStatistics::inverse_gamma_cdf(float64_t p, float64_t a,
...
+	if (error_code!=0)
+	{
+		switch (error_code)
+		{
+			case 1:
+				SG_SERROR("Params a (%f) or b (%f) cannot be negative!\n", a, b);
```
karlnapf: I think these guys should be checked via REQUIRE before, but then the error codes can basically be ignored right?

karlnapf: btw we should always print the value of the given numbers as in the REQUIRE elsehwere

karlnapf:
I think this is good. I will merge and continue to edit myself

yorkerlin:
@karlnapf @lambday 
Does this PR can pass the unit test about `lnormal_cdf` ?

karlnapf:
yes, the test was never touched and passes

PR #3130, opened 2016-03-29
Refactoring for issue #3118

c4goldsw:
This PR contains linalg::variance and linalg::matrix_variance.  Contains three simple unit tests.


src/shogun/mathematics/linalg/internal/implementation/Variance.h
```diff
@@ -0,0 +1,236 @@
...
+ * either expressed or implied, of the Shogun Development Team.
+ */
+
+#ifndef SRC_SHOGUN_MATHEMATICS_LINALG_INTERNAL_IMPLEMENTATION_VARIANCE_H_
+#define SRC_SHOGUN_MATHEMATICS_LINALG_INTERNAL_IMPLEMENTATION_VARIANCE_H_
+
+#include <shogun/lib/config.h>
+#include <shogun/lib/SGVector.h>
+#include <shogun/lib/SGMatrix.h>
+//#include <shogun/mathematics/linalg/internal/implementation/MeanEigen3.h> To be included when matrix mean is implemented
```
karlnapf: can you remove such temp comments in PRs? Can always re-add later

src/shogun/mathematics/linalg/internal/implementation/Variance.h
```diff
@@ -0,0 +1,236 @@
...
+	 * \f$\{x_1, ..., x_m\}\f$, this is
+	 * \f$\frac{1}{m-1}\sum_{i=1}^m (x-\bar{x})^2\f$ where
+	 * \f$\bar x=\frac{1}{m}\sum_{i=1}^m x_i\f$
+	 *
+	 * @param a vector of values
+	 * @return variance of given values
+	 */
+	static T compute(SGVector<T> vec)
+	{
+		ASSERT(vec.vlen>1)
```
karlnapf: REQUIRE please.
See https://github.com/shogun-toolbox/shogun/wiki/Assertions

src/shogun/mathematics/linalg/internal/implementation/Variance.h
```diff
@@ -0,0 +1,236 @@
...
+		ASSERT(vec.vector)
+
+		float64_t mean = implementation::mean<Backend::EIGEN3, Vector>::compute(vec);
+
+		float64_t sum_squared_diff=0;
+		for (index_t i=0; i<vec.vlen; ++i)
```
karlnapf: This can be done using SIMD in eigen3. No loop please

src/shogun/mathematics/linalg/internal/implementation/Variance.h
```diff
@@ -0,0 +1,236 @@
...
+	 * Computes the variance for each row/col of matrix
+	 *
+	 * @param values vector of values
+	 * @param col_wise if true, every column vector will be used, row vectors
+	 * otherwise
+	 * @return variance of given values
+	 */
+	static ReturnTypeVec compute(SGMatrix<T> m, bool col_wise){
+
+		ASSERT(m.num_rows>0)
```
karlnapf: REQUIRE and proper message please

src/shogun/mathematics/linalg/internal/implementation/Variance.h
```diff
@@ -0,0 +1,236 @@
...
+		ASSERT(m.num_cols>0)
+		ASSERT(m.matrix)
+
+		ReturnTypeVec variance;
+
+		if(col_wise){
```
karlnapf: we put { in new line

src/shogun/mathematics/linalg/internal/implementation/Variance.h
```diff
@@ -0,0 +1,236 @@
...
+
+			variance = ReturnTypeVec(m.num_cols);
+			SGVector<T> tempVec(m.num_rows);
+
+			for (index_t i=0; i< m.num_cols; ++i){
+
+				//It'd be nice if SGMatrix offered a get_column_vector function that retrurned a SGVector
```
karlnapf: please avoid such comments (but just send a patch :) )

BUT as sbove, can you please do this with eigen without writing loops

src/shogun/mathematics/linalg/internal/implementation/Variance.h
```diff
@@ -0,0 +1,236 @@
...
+	}
+
+	/**
+	 * Method that computes the element-wise variance of a matrix
+	 *
+	 * @param m the matrix whose element-wise variance we want to
+	 * compute
+	 * @return the element-wise variance of m
+	 *
+	static T compute(SGMatrix<T> m){
```
karlnapf: please remove any commented out code in PRs

tests/unit/mathematics/Statistics_unittest.cc
```diff
@@ -539,3 +540,89 @@ TEST(Statistics, vector_mean_overflow_test)
...
+{
+	Eigen::VectorXd a(5);
+	a << 600, 470, 170, 430, 300;
+	EXPECT_EQ(linalg::variance(a), 21704);
+}
+
+TEST(Statistics, matrix_variance_test_colwise)
+{
+	SGMatrix<float64_t> A(5, 5);
+	A(0,0) = 35.000000;
```
karlnapf: why not just 2x2? makes an easier to read unit test

karlnapf:
Good patch! Needs some minor changes, but in principle this is soon ready to be merged :)

tests/unit/mathematics/Statistics_unittest.cc
iglesias: What about a more concise method to input the matrix instead of one line per element? I find all these lines to initialise a small matrix an overkill.

src/shogun/mathematics/linalg/internal/implementation/Variance.h
```diff
@@ -0,0 +1,236 @@
...
+ * @brief Generic class variance which provides a static compute method. This class
+ * is specialized for different types of vectors.
+ */
+template <enum Backend, class Vector>
+struct variance
+{
+	/** Scalar type */
+	typedef typename Vector::Scalar T;
+
+	/** Calculates unbiased empirical variance estimator of given values. Given
```
iglesias: estimator -> estimate

src/shogun/mathematics/linalg/internal/implementation/Variance.h
```diff
@@ -0,0 +1,236 @@
...
+	 * \f$\{x_1, ..., x_m\}\f$, this is
+	 * \f$\frac{1}{m-1}\sum_{i=1}^m (x-\bar{x})^2\f$ where
+	 * \f$\bar x=\frac{1}{m}\sum_{i=1}^m x_i\f$
```
iglesias: I would find it more consistent if the symbol for the vector in the equations (x) is the same as the variable name (a).

src/shogun/mathematics/linalg/internal/implementation/Variance.h
```diff
@@ -0,0 +1,236 @@
...
+	 *
+	 * @param a vector of values
```
iglesias: @param vec

iglesias: The comments above regarding the method documentation also apply here and below.

src/shogun/mathematics/linalg/internal/implementation/Variance.h
```diff
@@ -0,0 +1,236 @@
...
+		ASSERT(vec.vlen>1)
+		ASSERT(vec.vector)
+
+		float64_t mean = implementation::mean<Backend::EIGEN3, Vector>::compute(vec);
+
+		float64_t sum_squared_diff=0;
+		for (index_t i=0; i<vec.vlen; ++i)
+			sum_squared_diff+=CMath::pow(vec.vector[i]-mean, 2);
+
+		return sum_squared_diff / (vec.vlen);
```
iglesias: According to the method documentation, should it be divided by `vec.vlen` or `vec.vlen-1`?

tests/unit/mathematics/Statistics_unittest.cc
```diff
@@ -539,3 +540,89 @@ TEST(Statistics, vector_mean_overflow_test)
...
 }
+
+TEST(Statistics, variance_test)
+{
+	Eigen::VectorXd a(5);
+	a << 600, 470, 170, 430, 300;
+	EXPECT_EQ(linalg::variance(a), 21704);
+}
+
+TEST(Statistics, matrix_variance_test_colwise)
```
iglesias: Let's use `col_wise` and `row_wise` for the test names for consistency with the implementation above.

iglesias: I see there are other things in `Redux.h` called `rowwise` so consistency is a bit compromised already. I have no strong feelings about this any more, then.

src/shogun/mathematics/linalg/internal/modules/Redux.h
```diff
@@ -95,6 +96,44 @@ typename implementation::int2float<typename Vector::Scalar>::floatType mean(Vect
...
+ * Wrapper method for internal implementation of matrix variance of values that works
+ * with generic matricies.  Computes variance elementwise
+ * @param a matrix whose variance has to be computed
+ * @return the vector variance \f$\variance_i a_i\f$
+ *
+template <Backend backend=linalg_traits<Redux>::backend, class Matrix>
+typename Matrix::Scalar matrix_variance(Matrix m)
+{
+	return implementation::matrix_variance<backend,Matrix>::compute(m);
+}*/
```
iglesias: Why keeping this method commented?

src/shogun/mathematics/linalg/internal/modules/Redux.h
```diff
@@ -95,6 +96,44 @@ typename implementation::int2float<typename Vector::Scalar>::floatType mean(Vect
...
+
+/**
+ * Wrapper method for internal implementation of matrix variance of values that works
+ * with generic matricies.  Computes variance row-wise and column-wise
+ * @param a matrix whose variance has to be computed
```
iglesias: param m

iglesias: Documentation for the second argument.

iglesias:
Thanks for the patch! It needs a bit polishing here and there. Lookout for instances of the stuff we have pointed out even in places where we have not said anything explicitly.

karlnapf:
Thanks for reviewing @iglesias :)

src/shogun/mathematics/linalg/internal/implementation/Variance.h
c4goldsw: @karlnapf , I searched around a bit for doing SIMD operations or coefficeint-wise operations and came [across this](http://stackoverflow.com/questions/21728341/coefficient-wise-custom-functions-in-eigen), though it doesn't seem this is appropriate as it doesn't allow for in place operations.  Could you please point me in the right direction for carrying out SIMD operations in eigen3?

iglesias: Hey @c4goldsw. I believe @karlnapf was referring to something along the line of what is discussed in the second link in your message.

For inspiration, have a look to how this is done in other places in linalg for other operations. I have checked out a couple for you:
- Dot product between two vectors: https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/mathematics/linalg/internal/implementation/Dot.h#L102.
- Element-wise product of matrices: https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/mathematics/linalg/internal/implementation/ElementwiseProduct.h#L123

Hope that helps!

c4goldsw: @iglesias  Thanks!

src/shogun/mathematics/linalg/internal/implementation/Variance.h
c4goldsw: @iglesias I should have brought this up with someone else earlier - `vec.vlen - 1` [makes no sense](https://en.wikipedia.org/wiki/Variance#Discrete_random_variable), especially because the mean is computed using all values in the vector.    I had tested it as it is and it doesn't break current tests (though I'm wondering why previous tests passed with using `vec.vlen -1`).

iglesias: What do you mean it makes no sense? The link you have provided is to compute the variance of a discrete random variable from its probability mass function. That is not the same as using data to estimate the variance of an underlying model. The latter is in fact what is described in the documentation of the `compute` methods here.

If you are interested in understanding why in variance estimation the N-1 makes sense, I suggest you a quick read: http://dawenl.github.io/files/mle_biased.pdf.

c4goldsw: Well, thank you for providing with that.

c4goldsw:
I ran into some difficulties with merging recent changes so I decided to reset my local copy.  I'll create a new PR for this (I'll try not to let this happen again).

PR #3125, opened 2016-03-29
Add matrix_mean in linalg using Eigen3

OXPHOS:
Refer #3051 and PR #3070 

An issue with current solution is one has to have `no_diag` parameter when computing the mean of a matrix. 
Maybe a better way is to map `vector` to `matrix` in `Redux.h`?

Also I am wondering maybe the test file is too long.

Issue #3100, opened 2016-03-20
Make linalg easy again


lambday:
DISCLAIMER : Any similarity to anyone's presidential campaign slogan is purely coincidental.

tl;dr: 
(1) what the hell is going on 
(2) now the days have changed 
(3) get rid of that ugly {{...}} guy. 
(4) It would be a great opportunity

Guys, me and @karlnapf had a discussion to clean up linalg lib a bit. Here are a few bits and pieces of that discussion
- We **have** to stop supporting linalg back-end switch Shogun-wise. Earlier it seemed like a useful idea. But now it looks like a **{{insert swear words}}** feature to keep supporting. It's super easy to break the build with this because it's super easy to overlook the technical details that are involved to bring this feature alive. It calls for nasty bugs (examples #2801, #2825) that makes one go like "what the hell is going on" (1). And in reality, we hardly think that it would cater for any greater good for the lib.
- We **have** to stop automatic type conversion from CPU-type to GPU-type in the linalg methods. The data transfer involves some cost, which, if hidden under the hood, can go unnoticed. The devs should know when they want a GPU type.
- We **have** to separate hardware requirements (whether GPU is available or not) from software requirements (which third-party library is available). Earlier Eigen3 wasn't a hard requirement, so in designing libalg, we had to assume the possibility that Eigen3 is not there. But now the days have changed (2). It's time we rethink some of these stuffs. We may also give support for CUDA in future.

So, here are the plans for how we're gonna do it.
- A user of linalg would ask a GPU factory for a GPU type (data transfer from CPU --> GPU happens here).

``` c++
// cpu_type = SGMatrix<T> or SGVector<T>
// in future, maybe even SGSparseVector<T> and SGSparseMatrix<T>
auto gpu_type = linalg::GPUFactory::get_gpu_type(cpu_type);
auto result = linalg::foo(gpu_type);
```

  This makes it explicit - the dev knows that there is a cost involved (we could make it more explicit by naming it `pretty_please_put_data_on_gpu_if_possible_and_yes_I_know_it_is_costly(...)`). Plus, it makes it easier to code for this thing. Plus it helps us to get rid of that ugly `linalg::Backend::FOOBAR` guy (3) at the time of calling. (if we want anything to be under the hood, it's this guy).
- In the above example, we can consider the following cases:
  - GPU is available and ViennaCL is there (our default GPU lib) : returns a ViennaCL type
[...]

PR #3070, opened 2016-03-13
Mean computation in linalg - VectorMean

OXPHOS:
Issue #3051 
- Added VectorMean and Mean implementation in Redux Module. Modified from and implemented VectorSum.h and Sum.h
- Can return means for whole matrix, symmetric matrix, matrix blocks and return columnwise and rowwise means. 
- Diag values can be excluded for mean calculation in Eigen3 and in square matrices in Viennacl. 
- Empty matrices/vectors considered.
- Compile test passed

- Corrected typos in Redux.h (comments) 

I should have run tests before pulling. However, I find it a little bit difficult for me. I will be appreciated if someone can tell me how to run tests. Thanks!

Issue #3051, opened 2016-03-07
Add mean computation to linalg


youssef-emad:
The goal is to extend [linalg](http://shogun-toolbox.org/doc/en/latest/dir_0d103e974bf2461595574c32f82fb9be.html) to have mean computation operations.

karlnapf:
Can collect patches that need this here:

#3044

Ahmkel:
I am interested in this.
Could you please explain what's needed in more details?

karlnapf:
Check the linalg readme and read the code and unit tests. Then try to add the operation. There is already a lot of summing in the redux module so this is a very simple task

OXPHOS:
Hi, 
I had a look at the code. Since linalg does all the matrix computation, I guess I need to implement a Mean module independent of all current modules in linalg.h, and calculate the mean.row and mean.col of matrices? Did I take this too easy?
I had no experience in coding large projects and I really would like to learn! Thanks!

karlnapf:
There already is a sum redux module. But yes

theaverageguy:
Is this one done @karlnapf ? Can be closed I guess.

PR #3136, opened 2016-03-31
Add rowwise_mean to linalg using Eigen3

OXPHOS:
Refer #3051 and PR #3070, #3125 
- I didn't implement `colwise_mean` at the same time because I'm not sure how this implement looks like. Apparently I cannot copy `SGVector<T>` to `SGVector<int2float type>` so I just used `for`-loops. Also, I think we are supposed to use EIGEN3 methods, which I have not tried (but used `for`-loop instead). 
- According to #3100, I think we can get rid of `<enum Backend>` generic type? I can re-factor this after `rowwise`/`colwise` are both done.
- There are also `Block` and `symmetric matrix` calculation for `matrixsum`. I am wondering whether we need both for `matrixmean` as while?

- I saw the comparison highlighted the changes I made in PR #3125 (merged). I had no idea why this is happening...

### DIFF
--- a/src/shogun/mathematics/linalg/internal/implementation/MeanEigen3.h
+++ b/src/shogun/mathematics/linalg/internal/implementation/MeanEigen3.h
@@ -79,9 +79,7 @@ struct int2float<int64_t>
 };
  
 /**
- * @brief Generic class mean which provides a static compute method. This class
- * is specialized for different types of vectors and matrices, providing a mean
- * to deal with various vectors and matrices directly.
+ * @brief Generic class mean which provides a static compute method. 
  */
 template <enum Backend, class Matrix>
 struct mean
@@ -96,7 +94,7 @@ struct mean
 	 * Method that computes the vector mean
 	 *
 	 * @param a vector whose mean has to be computed
-	 * @return the vector mean \f$\mean_i a_i\f$
+	 * @return the vector mean \f$\bar a_i\f$
 	 */
 	static ReturnType compute(Matrix a);
 
@@ -108,6 +106,37 @@ struct mean
 	 * @return the matrix mean \f$\1/N^2\sum_{i,j=1}^N m_{i,j}\f$
 	 */
 	static ReturnType compute(Matrix a, bool no_diag=false);
+};
+
+/**
+ * @brief Generic class rowwise_mean which provides a static compute method.
+ */
+template <enum Backend, class Matrix>
+struct rowwise_mean
+{
+	/** Scalar type */
+	typedef typename Matrix::Scalar T;
+
+	/** int2float type */
+	typedef typename int2float<T>::floatType ReturnDataType;
+
+	/**
+	 * Method that computes the row wise sum of co-efficients of SGMatrix using Eigen3
+	 *
+	 * @param m the matrix whose rowwise sum of co-efficients has to be computed
+	 * @param no_diag if true, diagonal entries are excluded from the sum
+	 * @return the rowwise sum of co-efficients computed as \f$s_i=\sum_{j}m_{i,j}\f$
+	 */
+	static SGVector<ReturnDataType> compute(SGMatrix<T> m, bool no_diag);
+
+	/**
+	 * Method that computes the row wise sum of co-efficients of SGMatrix using Eigen3
+	 *
+	 * @param m the matrix whose rowwise sum of co-efficients has to be computed
+	 * @param no_diag if true, diagonal entries are excluded from the sum
+	 * @param result Pre-allocated vector for the result of the computation
+	 */
+	static void compute(SGMatrix<T> mat, SGVector<ReturnDataType> result, bool no_diag);
 };
 
 /**
@@ -127,7 +156,7 @@ struct mean<Backend::EIGEN3, Matrix>
 	 * Method that computes the mean of SGVectors using Eigen3
 	 *
 	 * @param a vector whose mean has to be computed
-	 * @return the vector mean \f$\mean_i a_i\f$
+	 * @return the vector mean \f\bar a_i\f$
 	 */
 	static ReturnType compute(SGVector<T> vec)
 	{
@@ -141,7 +170,7 @@ struct mean<Backend::EIGEN3, Matrix>
 	 *
 	 * @param a matrix whose mean has to be computed
 	 * @param no_diag if true, diagonal entries are excluded from the mean (default - false) 
-	 * @return the matrix mean \f$\mu_{i,j}m_{i,j}\f$
+	 * @return the matrix mean \f$\1/N^2 \sum_{i,j=1}^N m_{i,j}\f$
 	 */
         static ReturnType compute(SGMatrix<T> mat, bool no_diag=false)
 	{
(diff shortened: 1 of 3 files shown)


Saurabh7:
Regarding last point , I think you are missing a rebase on updated develop. https://github.com/shogun-toolbox/shogun/wiki/README_developer , its explained under "versioning scheme"

src/shogun/mathematics/linalg/internal/implementation/MeanEigen3.h
```diff
@@ -93,15 +91,55 @@ struct mean
 	typedef typename int2float<T>::floatType ReturnType;
 
 	/**
-	 * Method that computes the vector/matrix mean
+	 * Method that computes the vector mean
 	 *
-	 * @param a vector/matrix whose mean has to be computed
-	 * @return the vector/matrix mean \f$\mean_i a_i\f$
+	 * @param a vector whose mean has to be computed
+	 * @return the vector mean \f$\mean_i a_i\f$
```
karlnapf: I am pretty sure `\mean` is not a latex macro

src/shogun/mathematics/linalg/internal/implementation/MeanEigen3.h
```diff
@@ -126,8 +164,112 @@ struct mean<Backend::EIGEN3, Matrix>
...
+		REQUIRE(mat.num_cols > 0, "Matrix column number cannot be zero!\n");
+
+		Eigen::Map<MatrixXt> m = mat;
+		SGVector<T> temp(m.rows());
+		temp = rowwise_sum<Backend::EIGEN3, SGMatrix<T> >::compute(mat, no_diag);
+
+		if (!no_diag)
+		{
+			for (index_t i = 0; i < m.rows(); ++i)
+			{
```
karlnapf: minor: we dont use { } for single statement `for` or `if`

src/shogun/mathematics/linalg/internal/modules/Redux.h
```diff
@@ -190,15 +178,58 @@ typename implementation::rowwise_sum<backend,Matrix>::ReturnType rowwise_sum(
...
+	return implementation::mean<backend,Matrix>::compute(m, no_diag);
+}
+
+/**
+ * Wrapper method for internal implementation of matrix rowwise mean of values 
+ * that works with generic dense matrices
+ *
+ * @param m the matrix whose rowwise mean has to be computed
+ * @param no_diag if true, diagonal entries are excluded from the mean (default - false)
+ * @return the rowwise mean computed as \f$\mean_i a_i\f$
```
karlnapf: latex

tests/unit/mathematics/linalg/MeanEigen3_unittest.cc
```diff
@@ -82,5 +82,180 @@ TEST(MeanEigen3, Eigen3_dynamic_explicit_eigen3_backend_int)
...
 }
 
+TEST(MeanEigen3, SGMatrix_asymmetric_eigen3_backend_with_diag_float)
+{
+	const index_t m = 2;
+	const index_t n = 3;
+	SGMatrix<float64_t> mat(m, n);
+
+	for (index_t i = 0; i < m; ++i)
+	{
```
karlnapf: btw it would be nicer if one could "read off" the matrix in such unit tests -- so that I can directly check what is going on. This loop is hard to decipher.

karlnapf:
I think for now we dont need to blocks. We might use later @lambday when we work more on the block mmd

src/shogun/mathematics/linalg/internal/implementation/MeanEigen3.h
```diff
@@ -126,8 +164,112 @@ struct mean<Backend::EIGEN3, Matrix>
...
+				result[i] = temp[i] / ReturnDataType(m.cols()); 
+			}
+		}
+		else if (m.rows() <= m.cols())
+		{
+			for (index_t i = 0; i < m.rows(); ++i)
```
karlnapf: yeah eigen3 simd would be better here. is that not possible?

karlnapf:
@Saurabh7 is correct, you just have to rebase against develop.

Whenever you pull, use the rebase flag, so avoid this

OXPHOS:
Thanks I'll try rebasing. I am not familiar with SIMD so I might need a little bit reading. I was out of town for the last few days so I was not able to work on this.

karlnapf:
Just use Eigen operations rather than loops.

OXPHOS:
Update: I cannot bypass loops now because of the type convert problem during mean calculation.

I tried:

```
/** int2float type */
typedef typename int2float<T>::floatType ReturnDataType;
 /** Eigen vector type - T */
typedef Eigen::Matrix<T,Eigen::Dynamic,1> VectorXt;
 /** Eigen vector type - int2float Type */
typedef Eigen::Matrix<ReturnDataType,Eigen::Dynamic,1> VectorXr;

SGVector<T> sum = rowwise_sum<Backend::EIGEN3, SGMatrix<T> >::compute(mat, no_diag);
Eigen::Map<VectorXt> s = sum;
Eigen::Map<VectorXr> r1 = result; // SGVector<ReturnDataType> 
r1 = s / mat.num_cols; // **Error: MIX DATA TYPE**
Vector Xt r2;
r2 = s / mat.num_cols; // Correct
r2 = s / ReturnDataType(mat.num_cols); // **Error: MIX DATA TYPE**
```

I also tried using SGVector, but it doesn't support copy from one type to another

```
SGVector<T> sum = rowwise_sum<Backend::EIGEN3, SGMatrix<T> >::compute(mat, no_diag);
SGVector<double> mean(sum); // **Error:  no matching constructor for initialization of 'SGVector<double>'**
```

karlnapf:
Ok then, I see.
Sorry I did not realise this problem when I asked for that.
Well then -- if you rebase, squash, and address the comments above, we can merge this
Thanks!

OXPHOS:
@karlnapf Learned something more about cast...Thought it would be easy in the first place. But I guess it's nice to have `int` -> `float` copy method in `SGVector` and 'SGMatrix` ?

I'll see how things work out this time. I'll squash once everything looks good.
And thanks for the tips on rebasing! @Saurabh7 Cute invoker : )

Saurabh7:
@OXPHOS Glorious invocation :) )

karlnapf:
@OXPHOS I think such a method should be in linalg itself. Feel free to add....

data
```diff
@@ -1 +1 @@
-Subproject commit c964e86dbcc984d4d87ae94378ccf81eaa2cbb37
+Subproject commit 77cba34872e1cf4c4eabd4f9540daa4e0ea76b06
```
karlnapf: I think this one slipped in by accident. Has to be removed

karlnapf:
Fine to merge apart from the data version

OXPHOS:
...May I ask how to access that data version thing? I cannot even find it. And the change was in my very first commit on this branch so I could not revert/reset. Rebase didn't seem to work either..

OXPHOS:
Did it..But still no idea where that data version change came from.

karlnapf:
Data is changed whenever the git submodule data is changed (might happen if you run tests locally, or otherwise). It just stored the sha1 hash of the current version of the data submodule that shogun depends upon

karlnapf:
Thanks for the patch! :)

Issue #3140, opened 2016-03-31
Compilation error using newer Eigen library


arianepaola:
The current shogun develop branch and also earlier versions (e.g. shogun_4.1.0) do not compile with newer Eigen library versions.

Below is the output of Ubuntu 16.04 LTS using libeigen3-dev (3.3~beta1-2)

<div>
<pre>
ubuntu@1604:~/shogun/build$ cmake -GNinja .. && ninja
-- The C compiler identification is GNU 5.3.1
-- The CXX compiler identification is GNU 5.3.1
-- Check for working C compiler using: Ninja
-- Check for working C compiler using: Ninja -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler using: Ninja
-- Check for working CXX compiler using: Ninja -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Found CCache: /usr/bin/ccache  
-- Performing Test _HAS_CXX11_FLAG
-- Performing Test _HAS_CXX11_FLAG - Success
-- Checking C++11 support for "__func__"
-- Checking C++11 support for "__func__": works
-- Checking C++11 support for "atomic"
-- Checking C++11 support for "atomic": works
-- Checking C++11 support for "auto"
-- Checking C++11 support for "auto": works
-- Checking C++11 support for "auto_ret_type"
-- Checking C++11 support for "auto_ret_type": works
-- Checking C++11 support for "class_override_final"
-- Checking C++11 support for "class_override_final": not supported
-- Checking C++11 support for "constexpr"
-- Checking C++11 support for "constexpr": works
-- Checking C++11 support for "cstdint"
-- Checking C++11 support for "cstdint": works
-- Checking C++11 support for "decltype"
-- Checking C++11 support for "decltype": works
-- Checking C++11 support for "initializer_list"
-- Checking C++11 support for "initializer_list": works
-- Checking C++11 support for "lambda"
-- Checking C++11 support for "lambda": works
-- Checking C++11 support for "long_long"
-- Checking C++11 support for "long_long": works
-- Checking C++11 support for "nullptr"
-- Checking C++11 support for "nullptr": works
-- Checking C++11 support for "regex"
-- Checking C++11 support for "regex": works
-- Checking C++11 support for "rvalue-references"
[...]

PR #3142, opened 2016-04-01
Workaround for #3140 Compilation error using newer Eigen library.

arianepaola:
This pull request is a workaround for #3140 causing a compilation error, when using Eigen 3.3 versions.
It adds an additional check to verify that the Eigen version is between 3.1.0 and 3.2.5. (excluding >= 3.2.6)

### DIFF
--- a/src/shogun/mathematics/linalg/ratapprox/logdet/opfunc/DenseMatrixExactLog.cpp
+++ b/src/shogun/mathematics/linalg/ratapprox/logdet/opfunc/DenseMatrixExactLog.cpp
@@ -11,7 +11,7 @@
 
 #include <shogun/mathematics/eigen3.h>
 
-#if EIGEN_VERSION_AT_LEAST(3,1,0)
+#if EIGEN_VERSION_AT_LEAST(3,1,0) && (EIGEN_WORLD_VERSION == 3 && EIGEN_MAJOR_VERSION == 2 && EIGEN_MINOR_VERSION < 6 )
 #include <unsupported/Eigen/MatrixFunctions>
 #endif // EIGEN_VERSION_AT_LEAST(3,1,0)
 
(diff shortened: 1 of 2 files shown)


vigsterkr:
@arianepaola good, but would be better if we could actually solve this by actually have a solution for 3.3+ :) because now as far as i understand this will just not use eigen 3.3+, or?

src/shogun/mathematics/linalg/ratapprox/logdet/opfunc/DenseMatrixExactLog.cpp
```diff
@@ -11,7 +11,7 @@
 
 #include <shogun/mathematics/eigen3.h>
 
-#if EIGEN_VERSION_AT_LEAST(3,1,0)
+#if EIGEN_VERSION_AT_LEAST(3,1,0) && (EIGEN_WORLD_VERSION == 3 && EIGEN_MAJOR_VERSION == 2 && EIGEN_MINOR_VERSION < 6 )
```
karlnapf: Can you use the CMake function `VERSION_LESS`
See e.g. `examples/meta/cpp/CMakeLists.txt`
It is a bit cleaner

karlnapf:
I agree with @vigsterkr on fixing this somehow rather than excluding new eigen versions.
However, I am fine to merge this as a temp. solution as we can just bundle eigen if the system version is too recent. I leave it to @vigsterkr to decide

vigsterkr:
@karlnapf no temp merge. i *hate* temp merges :dancer:

vigsterkr:
i mean i really appreciate the effort here! i do! but let's try to fix it so that 3.3+ can play as well. should not be that hard

vigsterkr:
btw i should have written after the :dancer: *caramba!*

karlnapf:
hahahah ;)

karlnapf:
I would merge this one. Requiring a stable eigen version (and bundling if not available) is fine in my eyes. Once the new eigen is out, we can change that

vigsterkr:
CARAMBA! :do_not_litter: 
be more patient :) i'll add patches here and then let's merge it plz.

arianepaola:
@vigsterkr @karlnapf The comments make sense. For example Ubuntu 16.04 will be released this month and they often release with beta packages. The Eigen 3.3 beta also uses a 3.2.92 version.

I have modified my pull request to take out the version numbers from the source file and to rely only on one preprocessor definition. This moves the minimum Eigen version to a preprocessor definition and also resolves a mismatch between the source files (e.g. at least 3.1.0) and the CMake required Eigen 3.1.2.

EIGEN_VERSION_DEPENDABLE surrounds the previous code block.

In my opinion it is not a good idea to force the download of earlier Eigen versions, such as the defined 3.2.8 in cmake/external/Eigen3.cmake as it will hide bugs such as #3140 and will hinder testing shogun on newer versions.

The modified pull request will still trigger the error in #3140 and provide a temporary work around for #3142.

After providing a final work around for #3142, the CMake IF clause can be modified to just use at least Eigen 3.1.2.

Thanks, Ariane

PR #3143, opened 2016-04-01
Fixes small grammar error in the cmake message output.

arianepaola:
This fixes a small grammar error in the cmake message output.


iglesias:
Thanks! Since the line with the second change is actually a comment, can we just get rid of it?

karlnapf:
Please leave it. It should be there but was a hotfix to make the build work. I am still trying to figure out how to do this properly (and being able to give this error message)

arianepaola:
Thanks @iglesias @karlnapf

PR #3144, opened 2016-04-03
added cholesky decomposition in linalg

curiousguy13:
added cholesky decomposition with eigen3 backend in linalg.

### DIFF
--- a/src/shogun/mathematics/linalg/internal/implementation/Cholesky.h
+++ b/src/shogun/mathematics/linalg/internal/implementation/Cholesky.h
@@ -1,0 +1,118 @@
+/*
+ * Copyright (c) The Shogun Machine Learning Toolbox
+ * Written (w) 2016 Kunal Arora
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * The views and conclusions contained in the software and documentation are those
+ * of the authors and should not be interpreted as representing official policies,
+ * either expressed or implied, of the Shogun Development Team.
+ */
+
+#ifndef CHOLESKY_IMPL_H_
+#define CHOLESKY_IMPL_H_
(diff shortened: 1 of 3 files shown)


src/shogun/mathematics/linalg/internal/implementation/Cholesky.h
```diff
@@ -0,0 +1,118 @@
...
+ * @brief Generic class which is specialized for different backends to compute the cholesky decomposition of a dense matrix
+ */
+template <enum Backend, class Matrix>
+struct cholesky
+{
+
+	/**Scalar type */
+	typedef typename Matrix::Scalar T;
+
+	/**Compute the cholesky decomposition A=L L* or A=U* U of a Hermitian positive definite matrix
```
karlnapf: Should mention this is lower Cholesky. And what about using latex for the doxygen?

src/shogun/mathematics/linalg/internal/implementation/Cholesky.h
```diff
@@ -0,0 +1,118 @@
...
+	 * @param A - the matrix whose cholesky decomposition is to be computed
+	 * @param lower - whether to compute the uper or lower triangular Cholesky factorization (default:upper)
```
karlnapf: I would vote for default: lower

src/shogun/mathematics/linalg/internal/implementation/Cholesky.h
```diff
@@ -0,0 +1,118 @@
...
+
+		//compute matrix L or U
+		if(lower==false)
+			map_cho= llt.matrixU();
+		else
+			map_cho= llt.matrixL();
+
+		// checking for success
+		if (llt.info()==Eigen::NumericalIssue)
+			SG_SWARNING("Matrix is not Hermitian positive definite!\n");
```
karlnapf: This should be an error I think

tests/unit/mathematics/linalg/Cholesky_unittest.cc
```diff
@@ -0,0 +1,82 @@
...
+
+using namespace shogun;
+using namespace Eigen;
+
+TEST(LLTLinearSolver, compute_cholesky_lower)
+{
+	const index_t size=2;
+	SGMatrix<float64_t> m(size, size);
+
+	// LLT doesn't work on non-symmetric matrices
```
karlnapf: I don't get the comment.

karlnapf:
Exactly what we were looking for! Thanks.
Please check my comments and squash all commits to one.
If somebody else agrees, fine to merge from my side then

tests/unit/mathematics/linalg/Cholesky_unittest.cc
curiousguy13: The matrix m used in the test case is a symmetric matrix because LLT does not work on non-symmetric matrices.
Should I change/remove the comment?

karlnapf: yeah maybe just remove it. 
It is by definition that Cholesky only works on sym psd matrices

src/shogun/mathematics/linalg/internal/implementation/Cholesky.h
```diff
@@ -0,0 +1,118 @@
...
+	/** Eigen3 matrix type */
+	typedef Eigen::Matrix<T,Eigen::Dynamic,Eigen::Dynamic> MatrixXt;
+
+	/**Compute the cholesky decomposition A=L L* or A=U* U of a Hermitian positive definite matrix
+	 * @param A - the matrix whose cholesky decomposition is to be computed
+	 * @param lower - whether to compute the uper or lower triangular Cholesky factorization (default:upper)
+	 */
+	static SGMatrix<T> compute(SGMatrix<T> A, bool lower)
+	{
+
```
iglesias: Minor remark: this newline is not needed here.

iglesias:
+1 to Heiko's comments and it looks good to me too.

curiousguy13:
@karlnapf thanks for the review. I made the changes you asked and squashed the commits.

karlnapf:
Thanks and merging. Next step would be a triangular solve to solve linear systems with a precomputed Cholesky. 

@yorkerlin

karlnapf:
http://buildbot.shogun-toolbox.org/builders/trusty%20-%20libshogun%20-%20viennacl/builds/609/steps/compile/logs/stdio

Can you please fix that? This build is an old ubuntu ...

curiousguy13:
@karlnapf hmm, that's weird. I'll look into it.

yorkerlin:
@curiousguy13  @karlnapf 
The LLT solver use eigen3.
However, the `trusty` build uses `viennacl` instead of eigen3.

curiousguy13:
@yorkerlin yes, but many other linalg implementations like rangeFill and mean have only their eigen3 backend implemented and i don't think this trusty-viennacl fails on those.

yorkerlin:
@curiousguy13 
In that case, feel free to use the `viennacl` backend and test them locally.
Hope it helps

yorkerlin:
@curiousguy13 
A more thing:
```
yes, but many other linalg implementations like rangeFill and mean have only their eigen3 backend implemented and i don't think this trusty-viennacl fails on those.
```
First of all, make sure that you unit tests are guarded.
Secondly, make sure that there are unit tests using `rangeFill` and `mean`. 
Finally, make sure that locally you can reproduce the same error.
Hope it helps :) 

I had some similar experience when I implemented `GaussianARDKernel` using `linalg`.
All builds are fine expect the `trusty` build using `clang`.

karlnapf:
@lambday this is another reason to refactor the linalg stuff soon :)

PR #3145, opened 2016-04-05
fix kmeanspp initialization

Saurabh7:


Saurabh7:
I will post some details/results here later tonight  and any specific comments.

NEWS
```diff
@@ -3,6 +3,7 @@
 	* SHOGUN Release version 4.2.0 (libshogun 17.2, data 0.11, parameter 1)
 	* Features:
 	    - Add new Shogun cookbook [Heiko Strathmann, Sergey Lisitsyn]
+	    - Add new improved KMeans++ initialization [Saurabh Mahindre]
```
karlnapf: new improved is a bit vague... can you be concrete like "fixed bug in KMeans++" or "Improved efficiency of KMeans++"
It is also not a new feature but an efficiency improvement, there is a separate category for that

NEWS
```diff
@@ -12,6 +13,8 @@
 	    - Add meta examples to test build and integration test [Heiko Strathmann]
 	    - Drop ALGLIB (GPL) dependency in CStatistics and ship CDFLIB (public domain) instead [Heiko Strathmann]
 	    - Drop p-value estimation in model-selection [Heiko Strathmann]
+	    - Major efficiency improvments and bugfixes for KMeans [Saurabh Mahindre]
```
karlnapf: These two should go together

karlnapf: BTW can you please send a separate patch for the readme update?

src/shogun/clustering/KMeans.cpp
```diff
@@ -171,7 +175,11 @@ bool CKMeans::train_machine(CFeatures* data)
 
 	/* if kmeans++ to be used */
 	if (use_kmeanspp)
+	{
+#ifdef HAVE_LINALG_LIB
 		mus_initial=kmeanspp();
```
karlnapf: what happens if linalg is not available?

karlnapf:
What would a unit test for this initialization look like?

karlnapf:
Thanks for that! :)

src/shogun/clustering/KMeans.cpp
Saurabh7: There will be no initial centers and it will resort to random initialization.

Saurabh7:
Previous kmeans++ was not kmeans++ ;)  all it did was choose first `k` samples from data as initial centers.

```
	for (int32_t c_m=0;c_m<k;c_m++)
	{
		SGVector<float64_t> feature=lhs->get_feature_vector(c_m);
		for (int32_t r_m=0;r_m<dim;r_m++)
			mat(r_m,c_m)=feature[r_m];
	}
	SG_UNREF(lhs);
	return mat;
```
so I thought I could mention this as addition.

Saurabh7:
We already have a unit-test for this initialization but KMeans will run as long as we supply `k` centers. 

This initialization should improve quality of convergence (sum of squares), 
for example I was able to reproduce an example from the kmeans++ paper (http://ilpubs.stanford.edu:8090/778/1/2006-13.pdf) for cloud dataset:

```
Final sum: 15020418.176736 with KMeans++
Final sum: 28159168.675626  with KMeans

```
which is big improvement in quality.
For non synthetic meaningful dataset this init gives good convergence for all dataset i tested, but for bigger non meaningful dataset this can lead to slightly increased runtime but always better quality.
btw I think we could add a method to get the final within cluster sum?

karlnapf:
Ok, so the readme should definitely say "fix". Great catch btw! :)
What about putting up a benchmark for this initialisation in the benchmark framwork, or at least put this on a list.

karlnapf:
Restarted travis -- timeout error. Let's see
Otherwise I am fine to merge this.

karlnapf:
BTW the example can be added to the notebook. That would be very nice

Saurabh7:
I have removed changes to NEWS, will send another small PR after this.

Saurabh7:
I was thinking about putting this in benchmark framework too, but i believe only scikit has kmeans++  and its not really a fair comparison with another initialization methods for other libraries ?? 

Yes adding the example to notebook is good idea , I will add a method like `get_sum()` to KMeans for that. Then compare it with normal initialization.

karlnapf:
sklearn has kmeans++ so yes definitely compare it!

karlnapf:
Fine to merge from my side

iglesias:
Hey, @Saurabh7. I think that the previous kmeans++ has indeed implemented the kmeans++: https://github.com/shogun-toolbox/shogun/commit/05408fa731cda5f32e300b80f3ed6fcce1b944f9#diff-d738d5845382c912de5f74730c54329bR471.

Can you please elaborate?

> Previous kmeans++ was not kmeans++ ;) all it did was choose first k samples from data as initial centers. (which is worse than random init)

Saurabh7:
@iglesias Its only returning first `k` samples as initial centers : 

```
	for (int32_t c_m=0;c_m<k;c_m++)
	{
		SGVector<float64_t> feature=lhs->get_feature_vector(c_m);
		for (int32_t r_m=0;r_m<dim;r_m++)
			mat(r_m,c_m)=feature[r_m];
	}
	SG_UNREF(lhs);
	return mat;
```

iglesias:
I think you are right. There is a small bug in the previous code, the first line in the loop above

`SGVector<float64_t> feature=lhs->get_feature_vector(c_m);`

should be

`SGVector<float64_t> feature=lhs->get_feature_vector(mu_index[c_m]);`

Does that make sense to you?

Saurabh7:
Yes, I think that makes sense if we are storing indices. Thanks!

Saurabh7:
So How do we proceed here , what do you guys think?
 the  changes in this PR will be improving the current version then I think :

- Replace `while` loop with for
- cleanups using `CMath` and `linalg` operations.
- using newly added precomputations in distance.
- parallel computations for distances.
- Using local tries to improve the randomly chosen candidates (scikit uses this too, which is ported form the original authors implementation i believe)

iglesias:
I say it is ready to merge from my side too. Thanks for the work and awesome you caught the bug!

Is the benchmark mentioned above coming next?

src/shogun/clustering/KMeans.cpp
```diff
@@ -331,57 +339,84 @@ void CKMeans::store_model_features()
...
+	SGVector<float64_t> min_dist=SGVector<float64_t>(lhs_size);
+	min_dist.zero();
+
+	/* First center is chosen at random */
+	int32_t mu=CMath::random((int32_t) 0, lhs_size-1);
+	SGVector<float64_t> mu_first=lhs->get_feature_vector(mu);	
+	for(int32_t j=0; j<dimensions; j++)
+		centers(j, 0)=mu_first[j];
+
+	distance->precompute_lhs();
```
iglesias: I think this name can lead to confusion because it is not distances that are pre-computed but the squared norms of the vectors. Why did the name change in #3073, @Saurabh7?

karlnapf:
https://github.com/shogun-toolbox/shogun/commit/69f1f2e17d14863e70129077bba2e390cdddc8d1

src/shogun/clustering/KMeans.cpp
Saurabh7: @iglesias It was added to base distance to make it generic and avoid calls to specific distance specialization . Depending on the derived class the precomputations are handled specifically ( e.g. squared norms for EuclideanDistance)

iglesias: I see. Are there any other examples currently implemented where `precompute` is calculating something different to squared norms?

Saurabh7: Not currently i dont think so . But this may avoid adding functionality in code that depends only on one special class.

Saurabh7:
@karlnapf will take care 

 btw heres the output of benchmark with kmeans++ : 

```
                mlpack  matlab    scikit  mlpy    shogun  weka 
wine                -       -  0.002446     -  0.004367     - 
USCensus1990        -       -  0.041973     -  0.031639     -
isolet              -       -  3.655900     -  1.371127     - 
covtype             -       -  21.970030    -  17.281851     -
mnist               -       -  39.379816    -  19.649297     - 
corel-histogram     -       -    1.848748   -   1.648231     - 
```

Also I checked final convergence, so even while we are fast, quality is good : 

```
           shogun                                        scikit
mnist     Final sum: 178432244999.232086              inertia 178432548991.517
isolet    Final sum: 472147.971                       inertia 467640.877
covtype   Final sum: 448279298246.616                 inertia 448279298610.849
corel-h   Final sum: 6705.182                         inertia 6760.769
```

iglesias:
Time is looking nice (btw, I am assuming that time is in the first table, measured in seconds).

Can you provide more context about the second table (what are final sum and intertia)? Also, out of curiosity, why a couple of data sets of the first table are missing in the second?

Does it make sense to include these results in the benchmark framework?

Saurabh7:
The first one is results from the benchmark framework , yes its runtime in seconds.

Second one I generated using my own scripts on the bigger of the datasets. 
Final sum/ inertia is essentially the sum of squared distances to final chosen closest centers. 

The benchmark framework already has comparison for lloyds implementation for chosen initial centers. I think we already get understanding of how fast the implementation is using that.

iglesias:
All righty, thanks for the explanation :-)

karlnapf:
Nice work. Looking forward to have that in the benchmark framework. BTW what about these linalg overhead problems? What is the state there?

Saurabh7:
@karlnapf `linalg::dot` was slower compared to `CMath::dot` for smaller dimensions , we haven't had any updates there.

But this reminded me of something so I ran some benchmarks:
script: https://gist.github.com/Saurabh7/c0318f5656d6ca2799df969e4582e237
output: https://gist.github.com/Saurabh7/44ecad25e829f19faefb209143d09b02

@lambday  what do you think? I think this was tried , but not for smaller vectors .. 
Passing `SGVector` is slower compared to say passing pointers.

Saurabh7:
@karlnapf @lambday 
I think I understood the cause for performance hit. 

- I changed the compute in dot.h to pass by const reference
```
static T compute(shogun::SGVector<T> a, shogun::SGVector<T> b)
``` 
 to 
```
static T compute(const shogun::SGVector<T>& a, const shogun::SGVector<T>& b)
````

- And make direct call to 
`linalg::implementation::dot<linalg::Backend::EIGEN3, SGVector<float64_t>>::compute(d.v1, d.v2)`

now the results are same for `CMath::dot` and `linalg::dot`

- So the two issues were:

  - Pass by value for `SGVector` caused a bit of performance hit. This was not noticed for higher dimensions but is noticeable for lower dimensions.
  - Wrapper function is causing a bit of overhead compared to direct call to compute.

**Comment by Saurabh7** [comment]
Script to reproduce after change to const reference : https://gist.github.com/Saurabh7/d4776a4243fd435b56d89113ae0e00bf

output : https://gist.github.com/Saurabh7/981de598b744dbaced2be34e15f529eb

**Comment by Saurabh7** [comment]
@karlnapf @lambday 
wow I just checked by changing the wrapper method to pass by reference too:

```
typename Vector::Scalar dot(Vector a, Vector b)
```
changed to
```
typename Vector::Scalar dot(Vector& a, Vector& b)
```
along with changing the compute to pass const reference : 
```
static T compute(const shogun::SGVector<T>& a, const shogun::SGVector<T>& b)
```

and the results are now same as in CMath.
So looks like the problem was due to `SGVector` being passed by value in both wrapper method and in addition in the compute method leading to additional overhead.

iglesias:
Nice analysis and findings!

Saurabh7:
Thanks! I have not tested this for other operations and types but there is chance of similar cases for say smaller vectors. I can send a PR if this looks viable , but this is more of a temporary solution right ?

iglesias:
I think @lambday can give you a good answer to that.

karlnapf:
Great analysis. The reason why this is only noticeable in small dimensions simply will be the fact that for large dimensions, the dot product time dominates the call-by-value overhead. I don't see any reason for not using const references by default everywhere in linalg (apart from some of the in-place methods). Certainly, we have to do something about that, as the impact in KMeans was significant, which is not good.
@lambday @vigsterkr can you comment?

Issue #3046, opened 2016-03-06
Integrate python graphical examples in the test build


karlnapf:
Every now and then, we have to re-factor our graphical examples, e.g. #3045

Currently we have no way of checking whether they work.
While we don't want to (Python) unit test these scripts, it would be great if we could at least *execute* the examples once in our test build. Just like the other python examples

This task is to add the python graphical examples to the test build. This works via creating a CMakeLists.txt file in the appropriate directory, and adding this directory in cmake.

In order to run the graphical examples, the graphical output needs to be re-directed. Otherwise travis and the buildbot will fail. A hack to do this is to append something as
```Python
import matplotlib; matplotlib.use('Agg')
``` 
to the python example before running it. Google for what it does [Here](https://github.com/karlnapf/kernel_exp_family/blob/master/kernel_exp_family/examples/run_all.sh) is a hacky way to do this.
@vigsterkr or @lisitsyn  might have a better way in their sleves.

theaverageguy:
I am on it.!

theaverageguy:
Just checking @karlnapf :
1. shogun/examples/undocumented/python_modular/graphical/ This would contain the CMakeLists.txt . 

2. Add this directory in cmake .
3. Appending `import matplotlib; matplotlib.use('Agg')` to every test python file.

karlnapf:
something like this. We just want to make sure that all the graphical scripts at least run.
Some care will have to be taken when they go into infinite loops
I suggest googling how to test software with graphical interfaces ... The 'agg' is just a hacky way to do it

theaverageguy:
I am sending a PR. Please take a look when free @karlnapf . :) 
Can I break it into smaller PRs maybe?

karlnapf:
Please do. Send a PR for a single example first as a prototype. Then we can apply this iteratively to all examples

PR #3146, opened 2016-04-05
start building python graphical tests

sanuj:
For #3046.
I'm in the process of learning CMake. Please review.

### DIFF
--- a/examples/undocumented/python_modular/graphical/CMakeLists.txt
+++ b/examples/undocumented/python_modular/graphical/CMakeLists.txt
@@ -1,17 +1,14 @@
 INCLUDE(PythonEnvironment)
 GET_PYTHON_ENV()
 
-FILE(GLOB PYTHON_EXAMPLES *.py)
+# FILE(GLOB PYTHON_EXAMPLES *.py)
+SET(PYTHON_EXAMPLES classifier_gaussian_process_binary_classification.py)
 FOREACH(EXAMPLE ${PYTHON_EXAMPLES})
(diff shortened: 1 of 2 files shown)


examples/undocumented/python_modular/graphical/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
+INCLUDE(PythonEnvironment)
+GET_PYTHON_ENV()
+
+# FILE(GLOB PYTHON_EXAMPLES *.py)
```
sanuj: We can uncomment this to build all the graphical examples but work needs to be done on the graphical examples to make them build without error. Currently only `classifier_gaussian_process_binary_classification.py` is tested.

sanuj:
[Dunno why it's failing.](https://travis-ci.org/shogun-toolbox/shogun/jobs/120876406#L3143) Passes on my local.

karlnapf:
As said, check the travis log, search for "error" and you will see that matplotlib is not there

examples/undocumented/python_modular/graphical/classifier_gaussian_process_binary_classification.py
```diff
@@ -7,25 +7,27 @@
 # Written (W) 2013 Roman Votyakov
 #
 
-from pylab import *
-from numpy import *
-from itertools import *
+import matplotlib
+matplotlib.use('Agg')   # Comment this to see the plot
```
karlnapf: this should not be in the code file. It should only be put there when the script is called from the cmake test

karlnapf:
I close this for now....Feel free to re-open if you wanna continue, but it is not really there yet

PR #3147, opened 2016-04-07
update docker config to install matplotlib

sanuj:
Installs matplotlib in travis docker builds. For #3146


karlnapf:
@vigsterkr you ok with that? I am

vigsterkr:
@sanuj @karlnapf is this for graphical examples?
we really want them to be run as tests?

karlnapf:
@vigsterkr the idea was mostly to check basic things such as imports, method names etc. So in a way, yes. But just executing, not asserting results. The problem currently is that they outdated (as in not executable) over time, and it takes time to fix.

vigsterkr:
i mean it's just a docker image and it will not hurt anybody to have matplotlib in the container :)

PR #3150, opened 2016-04-08
fix for PR #3144

curiousguy13:
added a missing definition of ReturnType in Cholesky.h


karlnapf:
Lets see :)

curiousguy13:
@karlnapf since the trusty-viennacl build i still failing I am setting up viennacl and will try to reproduce the error locally in order to investigate what is going wrong.

PR #3151, opened 2016-04-08
Using the NLOPT wrapper in the optimization framework (1)

yorkerlin:
@karlnapf

### DIFF
--- a/src/shogun/machine/gp/SingleSparseInferenceBase.cpp
+++ b/src/shogun/machine/gp/SingleSparseInferenceBase.cpp
@@ -316,6 +316,20 @@ double CSingleSparseInferenceBase::nlopt_function(unsigned n, const double* x, d
 void CSingleSparseInferenceBase::enable_optimizing_inducing_features(bool is_optmization)
 {
 	m_opt_inducing_features=is_optmization;
+#ifdef USE_GPL_SHOGUN
+#ifdef HAVE_NLOPT
+	if (m_opt_inducing_features)
+	{
+		check_fully_sparse();
+		REQUIRE(m_fully_sparse,"Please use a kernel which has the functionality about optimizing inducing features\n");
+	}
+#else
+	SG_WARNING("For this functionality we require NLOPT library\n");
+#endif //HAVE_NLOPT
+#else 
+	SG_WARNING("For this functionality we require NLOPT (GPL License) library\n");
+#endif //USE_GPL_SHOGUN
+
 }
 
 void CSingleSparseInferenceBase::optimize_inducing_features()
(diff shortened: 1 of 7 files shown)


src/shogun/machine/gp/SingleSparseInferenceBase.cpp
```diff
@@ -270,64 +328,22 @@ void CSingleSparseInferenceBase::optimize_inducing_features()
...
-	// set absolute argument tolearance
-	nlopt_set_xtol_abs1(opt, m_ind_tolerance);
-	nlopt_set_ftol_abs(opt, m_ind_tolerance);
-
-	nlopt_set_min_objective(opt, CSingleSparseInferenceBase::nlopt_function, this);
+	CSingleSparseInferenceCostFunction *cost_fun=new CSingleSparseInferenceCostFunction();
+	SG_REF(this);
+	cost_fun->set_target(this);
+    
+	NLOPTMinimizer* opt=new NLOPTMinimizer(cost_fun);
```
yorkerlin: @karlnapf 
we can use another optimizer which supports box constraints.

yorkerlin:
merging

src/shogun/machine/gp/SingleSparseInferenceBase.cpp
```diff
@@ -29,23 +29,81 @@
...
+        void set_target(CSingleSparseInferenceBase *obj)
+	{
+		if(obj!=m_obj)
+		{
+			SG_REF(obj);
+			SG_UNREF(m_obj);
+			m_obj=obj;
+			REQUIRE(m_obj,"Object not set\n");
+			m_obj->check_fully_sparse();
+			REQUIRE(m_obj->m_fully_sparse,"Can not compute gradient\n");
```
karlnapf: Proper error message. State why it cannot do that, what is given and what is expected

yorkerlin: @karlnapf 
```
void CSingleSparseInferenceBase::check_fully_sparse()
{
        REQUIRE(m_kernel, "Kernel must be set first\n")
        if (strstr(m_kernel->get_name(), "SparseKernel")!=NULL)
                m_fully_sparse=true;
        else
        {
                SG_WARNING( "The provided kernel does not support to optimize inducing features\n");
                m_fully_sparse=false;
        }
}
```

karlnapf: Why put the warning in the method implicitly?
I would rather `REQUIRE(m_obj->m_fully_sparse,"Can not compute gradient, the provided kernel (%s) does not support to optimize inducing features\n", kernel->get_name());` to be fully explicit

PR #3153, opened 2016-04-09
minor fix

yorkerlin:


karlnapf:
@yorkerlin could you put a bit more informative commit messages? :)

Issue #3156, opened 2016-04-11
compilation error in methods with no viennacl implementation


curiousguy13:
linalg currently contains methods like range_fill, mean, cholesky etc. which only have their eigen3 implementation and when viennacl is given as backend , it gives a compilation error.

the following is an example of such an error in cholesky.h when viennacl is given as backend in cholesky_unittest.cc.
http://buildbot.shogun-toolbox.org/builders/trusty%20-%20libshogun%20-%20viennacl/builds/616/steps/compile/logs/stdio

In my opinion there should be a way to deal with such situation without producing error.

karlnapf:
I absolutely agree, this is far from optimal. In particular the HAVE_VIENNACL guard to resolve this  doesnt make any sense.
We had some discussions about a re-factored linalg. We will put that to life soon

PR #3157, opened 2016-04-11
explicitly mentioned eigen3 backend in Cholesky_unittest and a minor change in Cholesky.h

curiousguy13:
fix for the failing build with viennacl backend : http://buildbot.shogun-toolbox.org/builders/trusty%20-%20libshogun%20-%20viennacl/builds/616

I have also opened issue #3156 regarding the general handling of viennacl backend by methods that don't have their viennacl backends implemented in linalg yet.

also, a minor change for the compute method to use the general ReturnType instead of SGMatrix


karlnapf:
In that case, the unit test should have an eigen3 suffix

karlnapf:
Fine to merge otherwise. Please update the name, squash the commits, and then I will merge.
Thanks for that!

curiousguy13:
done. :)

karlnapf:
Great thanks!
Will check the build whether it worked

PR #3159, opened 2016-04-11
Using the LBFGS wrapper in the optimization framework (1)

yorkerlin:


karlnapf:
How is the speed difference when compared to nlopt?

Issue #2558, opened 2014-10-25
Clean up KMeans


karlnapf:
The class is very messy. Since the mini batch extension, things got worse
- Do a proper object oriented distinction between kmeans and mini batch  kmeans rather than handling everything in one class. Its two _different_ algorithms, and therefore should be in different classes (where one inherits from the other and only specialises some internals)
- Fix error messages and requires, see [here](https://github.com/shogun-toolbox/shogun/wiki/Assertions) for a guide on how to write proper require messages
- Update documentation and describe the different methods properly, mini batch is not mentioned in class docs
- Write unit tests (if not there yet)

PR #3166, opened 2016-04-14
refactor KMeans

Saurabh7:


Saurabh7:
Refactor `KMeans` to have different classes for `KMeans` (Lloyds) and `KMeansMiniBatch` inheriting from `KMeansBase` .

So basically:
```
KMeans.cpp -> KMeansBase.cpp
KMeansLloydImpl.cpp -> KMeans.cpp
KMeansMiniBatchImpl.cpp -> KMeansMiniBatch.cpp
```
Along with adding some helper functions for training and some doc changes.

Github comparison is making it look messy , doesnt detect renames :(

src/shogun/clustering/KMeans.cpp
```diff
@@ -4,445 +4,187 @@
...
+	weights_set[0]=lhs_size;
 
-	SG_UNREF(lhs);
+	distance->precompute_lhs();
 
-}
+	int32_t changed=1;
+	int32_t iter;
 
-void CKMeans::compute_cluster_variances()
```
Saurabh7: This `compute_cluster_variances` I cannot make sense of it. 
But this is leading to additional computations since this is called at end of `train_machine` . and causes performance hit for some case like this for dataset `corel-histogram shape=(68040, 32) for k=10`
```
2.77 s with compute_cluster_variances
1.31 s without compute_cluster_variances
```

src/shogun/clustering/KMeans.cpp
```diff
@@ -4,445 +4,187 @@
...
+					
 				}
+
+				cluster_assignments[i] = min_cluster;
 			}
 		}
+		if(changed==0)
+			break;
 
-		R.vector[i]=(0.7*CMath::sqrt(rmin1)+0.3*CMath::sqrt(rmin2));
```
Saurabh7: I cannot make sense of this computation nor find any reference. Any hints ?

karlnapf: this is basically crap ;) so definitely remove it

src/interfaces/modular/Clustering.i
```diff
@@ -19,11 +19,15 @@
 %rename(DistanceMachine) CDistanceMachine;
 %rename(Hierarchical) CHierarchical;
 %rename(KMeans) CKMeans;
+%rename(KMeansBase) CKMeansBase;
```
karlnapf: does the base class need to be exposed via swig?

karlnapf:
It is impossible to review this. A few high level comments:

 * the `impl` classes are made to hide c++ code from SWIG, which will only see the main interface class. Are you sure you want to remove them. See some other method, e.g. the metric learning does this
 * similarly, the base class for sure would not be exposed via SWIG
 * A class diagram would help, can you draw one so that we can discuss these changes?
 * make sure that things remain working, i.e. the notebook, the examples, etc
 * are there any speed implications coming from this?
 * if you want, I can put up a feature branch for you. There you can develop a bit more freely and then we can merge the whole thing....thoughts

We need another dev to give an opinion here before merging. @vigsterkr @lisitsyn ?

Saurabh7:
This idea was discussed a bit here: https://github.com/shogun-toolbox/shogun/issues/2558

- Both KMeans have one `protected` method with the alogrithm implementations say `Lloyd_KMeans()` . I can still keep them in `impl` if it is going to slow down swig. But we need impl files for each algo then.  
- Tried removing Base class from `clustering_inlcudes.i` but  it doesn't work since we are inheriting from it. Other base classes like base `distance.h` are also exposed.
-  class diagram is simple like : 
```
			CKMeansBase
	| ---------- | ----------- | 
CKMeans  CKMeansMiniBatch  CKMeansElkan(or something else in future maybe)   
```
- I havent included speedups or changes for now , this is just restructuring.

karlnapf:
Thanks for the comments. I think it should be ok to merge then....

karlnapf:
@vigsterkr from my side this can be merged. What do you think? Shall we maybe put this in a feature branch to allow for some more checks (on buildbot for example?) Or merge from here? I think the new structure is fine and travis seems to be fine as well. ....

karlnapf:
(has to be rebased against #3217 though before merge)

karlnapf:
Can you rebase? There will be some minor conflicts from #3217
I will ping @vigsterkr to review so that we can merge soon

Saurabh7:
Ok I resolved conflicts. Lets see travis again.

src/shogun/clustering/KMeans.cpp
```diff
@@ -4,446 +4,187 @@
...
-			int32_t new_center=0;		
-			float64_t prob=CMath::random(0.0, 1.0);
-			prob=prob*sum;
+				for (int32_t j=0; j<dim; j++)
+					centers(j, cluster_i) += vec[j];
+				lhs->free_feature_vector(vec, i);
+			}
 		
-			for(int32_t j=0; j<lhs_size; j++)
+			for (int32_t i=0; i<num_centers; i++)
```
vigsterkr: coulnd't we omp pragma this?

src/shogun/clustering/KMeans.cpp
```diff
@@ -4,446 +4,187 @@
...
 			{
-				temp_sum+=min_dist[j];
-				if (prob <= temp_sum)
+				if (weights_set[i]!=0.0)
```
vigsterkr: shouldn't we at least allow a little bit of error/epsilon from 0.0 ?

src/shogun/clustering/KMeans.cpp
```diff
@@ -4,446 +4,187 @@
...
 				{
-					new_center=j;
-					break;
+					for (int32_t j=0; j<dim; j++)
```
vigsterkr: mmm this is basically a scaling of the centers_i column with weights_set[i];
you should use a vector operation on this.

src/shogun/clustering/KMeansBase.cpp
```diff
@@ -0,0 +1,381 @@
...
+#include <shogun/clustering/KMeansBase.h>
+#include <shogun/distance/Distance.h>
+#include <shogun/distance/EuclideanDistance.h>
+#include <shogun/labels/Labels.h>
+#include <shogun/features/DenseFeatures.h>
+#include <shogun/mathematics/Math.h>
+#include <shogun/base/Parallel.h>
+#include <shogun/mathematics/eigen3.h>
+
+#ifdef HAVE_PTHREAD
```
vigsterkr: i know that this is just the code moved from the original implementation, but what's the use of including pthread? couldn't find any pthread calls...

src/shogun/clustering/KMeansBase.cpp
```diff
@@ -0,0 +1,381 @@
...
+	dimensions=lhs->get_num_features();
+	const int32_t centers_size=dimensions*k;
+
+	REQUIRE(lhs_size>0, "Lhs features should not be empty");
+	REQUIRE(dimensions>0, "Lhs features should have more than zero dimensions");
+
+	/* if kmeans++ to be used */
+	if (use_kmeanspp)
+	{
+#ifdef HAVE_LINALG_LIB
```
vigsterkr: what happens if we dont have linalg lib?
i mean in that case we should at least warn the user that kmeans++ wont be used or something...

vigsterkr:
lets address the comments + rebase with the latest develop and see how travis behaves :)

src/shogun/clustering/KMeans.cpp
karlnapf: I think it is not worth it as the loop is ultra cheap

src/shogun/clustering/KMeansBase.cpp
karlnapf: Yes, and mention this in the class interface docs

src/shogun/clustering/KMeans.cpp
Saurabh7: Linalg didnt have colwise scale so had skipped this.

karlnapf: in that case either add it, or use eigen3 to the rescue

vigsterkr: we should add it if it isn't added yet, but `centers` is an `SGMatrix` right? you can easily just define the region of area where the column vector is and do the scaling, or?

karlnapf:
Can we push this in parallel to the other stuff you are doing @Saurabh7 ?

src/shogun/clustering/KMeans.cpp
Saurabh7: Yes, this whole updating part can be done with eigen operations. Wasn't so used to eigen back then :) changing ...

src/shogun/clustering/KMeans.cpp
Saurabh7: This is basically integer, fixed.

vigsterkr: not sure if i understand :)

Saurabh7: The weight vector should be int type. This is just checking if we don't divide by zero.

vigsterkr: ah but if it's an int type why do you force a type cast with `0.0`? :)

vigsterkr: @Saurabh7 are you sure it's an int type:
```SGVector<float64_t> weights_set``` ?

Comment (Saurabh7) [inline src/shogun/clustering/KMeans.cpp]: Its not right now. I am changing it, since its number of points assigned to cluster.

**Comment by Saurabh7** [comment]
Ok updated :)

File: src/shogun/clustering/KMeans.cpp [hunk 9]
```diff
@@ -4,446 +4,189 @@
...
 
-		for (int32_t j=0; j<dimensions; j++)
-			mus(j,i)=vec[j];
+	SGVector<int32_t> cluster_assignments=SGVector<int32_t>(lhs_size);
+	cluster_assignments.zero();
 
-		lhs->free_feature_vector(vec, cluster_center_i);
-	}
+	/* Weights : Number of points in each cluster */
+	SGVector<int32_t> weights_set(num_centers);
```
vigsterkr: is `2^31 = 2147483648` i.e. 2.1 billion points enough? :) i mean in case of a big data set it could be more than that or?
i'm just suggesting that maybe using `uint32_t` or rather `int64_t` or `uint64_t` would be better choice, or?

src/shogun/clustering/KMeans.cpp
```diff
@@ -4,446 +4,189 @@
...
-					rmin1=dist;
-					rmin2=dist;
-					first_round=false;
-				}
-				else
+			if (min_cluster!=cluster_assignments_i)
+			{
+				changed++;
+#pragma omp atomic
+				weights_set[min_cluster]+= 1;
```
vigsterkr: this should be rather `++weights_set[min_cluster]`

src/shogun/clustering/KMeans.cpp
```diff
@@ -4,446 +4,189 @@
...
+#pragma omp atomic
+				weights_set[cluster_assignments_i]-= 1;
```
vigsterkr: just as above:
```--weights_set[cluster_assignments_i]```

File: src/shogun/clustering/KMeans.cpp [hunk 12]
```diff
@@ -4,446 +4,189 @@
...
+
+				if(fixed_centers)
 				{
-					if ((dist<rmin2) && (dist>=rmin1))
-						rmin2=dist;
+					SGVector<float64_t>vec=lhs->get_feature_vector(i);
```
vigsterkr: i know that it should be a different PR, but maybe it'd be good to think about whether we really only wanna support vectors of `SGVector<float64_t>`, i.e. why not `SGVector<float32_t>` or `SGVector<int32_t>` etc.

src/shogun/clustering/KMeans.cpp
```diff
@@ -4,446 +4,189 @@
...
+					
+					/* mu_new = mu_old + (x - mu_old)/(w) */					
+					for (j=0; j<dim; j++)
+					{
+						centers(j, min_cluster)+=
+							(vec[j]-centers(j, min_cluster)) / weights_set[min_cluster];
```
vigsterkr: only thing: a `/` operator (division) is usually much more costly than multiplication...
so precomputing `x = 1/weights_set[min_cluster]` and then just do
```
vec[j]-x*centers(j, min_cluster))
```
of course it'd be good to check what does the compiler do when you do `-O3`

src/shogun/clustering/KMeans.cpp
```diff
@@ -4,446 +4,189 @@
...
+					if (weights_set[cluster_assignments_i]!=0)
 					{
-						rmin2=rmin1;
-						rmin1=dist;
+						SGVector<float64_t>vec1=lhs->get_feature_vector(i);
+
+						for (j=0; j<dim; j++)
+						{
+							centers(j, cluster_assignments_i)-=
+								(vec1[j]-centers(j, cluster_assignments_i)) / weights_set[cluster_assignments_i];
```
vigsterkr: see above (diff between `/` and `*` operation)

src/shogun/clustering/KMeans.cpp
Saurabh7: Yes in the case if 2.1 billion points happen to be assigned to one center, it might be short :) should i change to int64_t then ?

vigsterkr: isn't `weights_set[i] = the number of elements assigned to the ith centre`?

Saurabh7: yes it is.

vigsterkr: ah ok yeah then rather use int64_t

Saurabh7:
Updated

vigsterkr:
ALL GREEN! good job @Saurabh7 let's merge it!

PR #3172, opened 2016-04-17
Implement the Nystrom approximate algorithm for KRR

cfjhallgren:
Hey here is my PR.

#TODO's in the code should be attended to or removed.

Added a normal example in the meantime.

### DIFF
--- a/src/shogun/regression/KRRNystrom.cpp
+++ b/src/shogun/regression/KRRNystrom.cpp
@@ -1,0 +1,136 @@
+
+/*
+ * Copyright (c) The Shogun Machine Learning Toolbox
+ * Written (W) 2016 Fredrik Hallgren
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * The views and conclusions contained in the software and documentation are those
+ * of the authors and should not be interpreted as representing official policies,
+ * either expressed or implied, of the Shogun Development Team.
+ */
+
+#include <limits>
+#include <shogun/regression/KRRNystrom.h>
+#include <shogun/labels/RegressionLabels.h>
+#include <shogun/mathematics/eigen3.h>
+#include <shogun/mathematics/Math.h>
+
+using namespace shogun;
+using namespace Eigen;
+
+CKRRNystrom::CKRRNystrom() : CKernelRidgeRegression()
+{
+	init();
+}
+
+CKRRNystrom::CKRRNystrom(float64_t tau, int32_t m, CKernel* k, CLabels* lab)
+: CKernelRidgeRegression(tau, k, lab)
+{
+	init();
+
+	m_num_rkhs_basis=m;
+
+	int32_t n=kernel->get_num_vec_lhs();
+
+	REQUIRE(m_num_rkhs_basis <= n, "Number of sampled rows (%d) must be \
+less than number of data points (%d)\n", m_num_rkhs_basis, n);
+}
+
+void CKRRNystrom::init()
+{
+	m_num_rkhs_basis=100;
+}
+
+SGVector<int32_t> CKRRNystrom::subsample_indices()
+{
+	int32_t n=kernel->get_num_vec_lhs();
+	SGVector<int32_t> temp(n);
+	temp.range_fill();
+	CMath::permute(temp);
+	SGVector<int32_t> col(m_num_rkhs_basis);
+	for (auto i=0; i<m_num_rkhs_basis; ++i)
+		col[i]=temp[i];
+	CMath::qsort(col.vector, m_num_rkhs_basis);
+
+	return col;
+}
+
+bool CKRRNystrom::solve_krr_system()
+{
+	int32_t n=kernel->get_num_vec_lhs();
+
+	SGVector<float64_t> y=((CRegressionLabels*)m_labels)->get_labels();
+	if (y==NULL)
+		SG_ERROR("Labels not set.\n");
+	SGVector<int32_t> col=subsample_indices();
+	SGMatrix<float64_t> K_mm(m_num_rkhs_basis, m_num_rkhs_basis);
+	SGMatrix<float64_t> K_nm(n, m_num_rkhs_basis);
+	#pragma omp parallel for
+	for (index_t j=0; j<m_num_rkhs_basis; ++j)
+	{
+		for (index_t i=0; i<n; ++i)
+			K_nm(i,j)=kernel->kernel(i,col[j]);
+	}
+	#pragma omp parallel for
+	for (index_t i=0; i<m_num_rkhs_basis; ++i)
+		memcpy(K_mm.matrix+i*m_num_rkhs_basis, K_nm.get_row_vector(col[i]), m_num_rkhs_basis*sizeof(float64_t));
+
+	Map<MatrixXd> K_mm_eig(K_mm.matrix, m_num_rkhs_basis, m_num_rkhs_basis);
+	Map<MatrixXd> K_nm_eig(K_nm.matrix, n, m_num_rkhs_basis);
+	MatrixXd K_mn_eig = K_nm_eig.transpose();
+	Map<VectorXd> y_eig(y.vector, n);
+	VectorXd alphas_eig(m_num_rkhs_basis);
+
+	/* Calculate the Moore-Penrose pseudoinverse */
+	MatrixXd Kplus=K_mn_eig*K_nm_eig+m_tau*K_mm_eig;
+	SelfAdjointEigenSolver<MatrixXd> solver(Kplus);
+	if (solver.info()!=Success)
+	{
+		SG_WARNING("Eigendecomposition failed.\n")
+		return false;
+	}
+
+	/* Solve the system for alphas */
+	MatrixXd D=solver.eigenvalues().asDiagonal();
+	MatrixXd eigvec=solver.eigenvectors();
+	float64_t dbl_epsilon=std::numeric_limits<float64_t>::epsilon();
+	const float64_t tolerance=m_num_rkhs_basis*dbl_epsilon*D.maxCoeff();
+	for (index_t i=0; i<m_num_rkhs_basis; ++i)
+	{
+		if (D(i,i)<tolerance)
+			D(i,i)=0;
+		else
+			D(i,i)=1/D(i,i);
+	}
+	MatrixXd pseudoinv=eigvec*D*eigvec.transpose();
+	alphas_eig=pseudoinv*K_mn_eig*y_eig;
+
+	/* Expand alpha with zeros to size n */
+	SGVector<float64_t> alpha_n(n);
+	alpha_n.zero();
+	for (index_t i=0; i<m_num_rkhs_basis; ++i)
+		alpha_n[col[i]]=alphas_eig[i];
+	m_alpha=alpha_n;
+
+	return true;
+}
--- a/src/shogun/regression/KRRNystrom.h
+++ b/src/shogun/regression/KRRNystrom.h
@@ -1,0 +1,128 @@
+
+/*
+ * Copyright (c) The Shogun Machine Learning Toolbox
+ * Written (W) 2016 Fredrik Hallgren
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * The views and conclusions contained in the software and documentation are those
+ * of the authors and should not be interpreted as representing official policies,
+ * either expressed or implied, of the Shogun Development Team.
+ */
+
+#ifndef _KRRNYSTROM_H__
+#define _KRRNYSTROM_H__
+
+#include <shogun/regression/KernelRidgeRegression.h>
+
+namespace shogun {
+
+/** @brief Class KRRNystrom implements the Nyström method for kernel ridge
+ * regression, using a low-rank approximation to the kernel matrix.
+ *
+ * The method is equivalent to ordinary kernel ridge regression, but through
+ * projecting the data on a subset of the data points, the full
+ * kernel matrix does not need to be computed, and the resulting system of
+ * equations for the alphas is cheaper to solve.
+ *
+ * Instead of the original linear system, the following approximate system
+ * is solved
+ *
+ * \f[
+ * {\bf \alpha} = (\tau K_{m,m} + K_{m,n}K_{n,m})^+K_{m,n} {\bf y}
+ * ]
+ *
+ * where \f$K_{n,m}\f$ is a submatrix containing all n rows and the m columns
+ * corresponding to the m chosen training examples, \f$K_{m,n}\f$ is its
+ * transpose and \f$K_{m,m} is the submatrix with the m rows and columns
+ * corresponding to the training examples chosen. \f$+\f$ indicates the
+ * Moore-Penrose pseudoinverse. The complexity is \f$O(m^2n)\f$.
+ *
+ * Several ways to subsample columns/rows have been proposed. Here they are
+ * subsampled uniformly. To implement another sampling method one has to
+ * override the method 'subsample_indices'.
+ */
+class CKRRNystrom : public CKernelRidgeRegression
+{
+public:
+	MACHINE_PROBLEM_TYPE(PT_REGRESSION);
+
+	/** Default constructor */
+	CKRRNystrom();
+
+	/** Constructor
+	 *
+	 * @param tau regularization parameter tau
+	 * @param m number of rows/columns to choose
+	 * @param k kernel
+	 * @param lab labels
+	 */
+	CKRRNystrom(float64_t tau, int32_t m, CKernel* k, CLabels* lab);
+
+	/** Default destructor */
+	virtual ~CKRRNystrom() {}
+
+	/** Set the number of columns/rows to choose
+	 *
+	 * @param m new m
+	 */
+	inline void set_num_rkhs_basis(int32_t m)
+	{
+		m_num_rkhs_basis=m;
+
+		if (kernel!=NULL)
+		{
+			int32_t n=kernel->get_num_vec_lhs();
+
+			REQUIRE(m_num_rkhs_basis<=n, "Number of sampled rows (%d) must be \
+less than number of data points (%d)\n", m_num_rkhs_basis, n);
+		}
+
+	};
+
+	/** @return object name */
+	virtual const char* get_name() const { return "KRRNystrom"; }
+
+protected:
+	/** Train regression using the Nyström method.
+	 *
+	 * @return boolean to indicate success
+	 */
+	virtual bool solve_krr_system();
+
+	/** Sample indices to pick rows/columns from kernel matrix
+	 *
+	 * @return SGVector<int32_t> with sampled indices
+	 */
+	SGVector<int32_t> subsample_indices();
+
+	/** Number of columns/rows to be sampled */
+	int32_t m_num_rkhs_basis;
+
+private:
+	void init();
+
+};
+
+}
+
+#endif // _KRRNYSTROM_H__
--- a/src/shogun/regression/KernelRidgeRegression.h
+++ b/src/shogun/regression/KernelRidgeRegression.h
@@ -74,7 +74,7 @@ class CKernelRidgeRegression : public CKernelMachine
 		 *
 		 * @param tau new tau
 		 */
-		inline void set_tau(float64_t tau) { m_tau = tau; };
+		inline virtual void set_tau(float64_t tau) { m_tau = tau; };
 
 		/** set convergence precision for gauss seidel method
 		 *
@@ -125,15 +125,16 @@ class CKernelRidgeRegression : public CKernelMachine
 		 *
 		 * @return boolean to indicate success
 		 */
-		bool solve_krr_system();
+		virtual bool solve_krr_system();
 
 	private:
 		void init();
 
-	private:
+	protected:
 		/** regularization parameter tau */
 		float64_t m_tau;
 
+	private:
 		/** epsilon constant */
 		float64_t m_epsilon;
 
--- a/tests/unit/regression/krrnystrom_unittest.cc
+++ b/tests/unit/regression/krrnystrom_unittest.cc
@@ -1,0 +1,159 @@
+/*
+* Copyright (c) The Shogun Machine Learning Toolbox
+* Written (w) 2016 Fredrik Hallgren
+* All rights reserved.
+*
+* Redistribution and use in source and binary forms, with or without
+* modification, are permitted provided that the following conditions are met:
+*
+* 1. Redistributions of source code must retain the above copyright notice, this
+* list of conditions and the following disclaimer.
+* 2. Redistributions in binary form must reproduce the above copyright notice,
+* this list of conditions and the following disclaimer in the documentation
+* and/or other materials provided with the distribution.
+*
+* THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+* ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+* WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+* DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+* ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+* (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+* LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+* ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+* (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+* SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*
+* The views and conclusions contained in the software and documentation are those
+* of the authors and should not be interpreted as representing official policies,
+* either expressed or implied, of the Shogun Development Team.
+*/
+
+#include <shogun/features/DenseFeatures.h>
+#include <shogun/labels/RegressionLabels.h>
+#include <shogun/kernel/GaussianKernel.h>
+#include <shogun/regression/KRRNystrom.h>
+#include <shogun/regression/KernelRidgeRegression.h>
+#include <shogun/base/some.h>
+#include <gtest/gtest.h>
+
+#ifdef HAVE_CXX11
+
+using namespace shogun;
+
+/**
+ * Test the main algorithm by comparison of the alphas and the predictions,
+ * using all columns in the approximation to make sure results are very similar
+ */
+TEST(KRRNystrom, apply_and_compare_to_KRR_with_all_columns)
+{
+	/* data matrix dimensions */
+	index_t num_vectors=10;
+	index_t num_features=1;
+
+	/* training label data */
+	SGVector<float64_t> lab(num_vectors);
+
+	/* fill data matrix and labels */
+	SGMatrix<float64_t> train_dat(num_features, num_vectors);
+	SGMatrix<float64_t> test_dat(num_features, num_vectors);
+	for (index_t i=0; i<num_vectors; ++i)
+	{
+		/* labels are linear plus noise */
+		lab.vector[i]=i+CMath::normal_random(0, 1.0);
+		train_dat.matrix[i]=i;
+		test_dat.matrix[i]=i;
+	}
+
+	/* training features */
+	auto features=some<CDenseFeatures<float64_t>>(train_dat);
+	auto features_krr=some<CDenseFeatures<float64_t>>(train_dat);
+
+	/* testing features */
+	auto test_features=some<CDenseFeatures<float64_t>>(test_dat);
+
+	/* training labels */
+	auto labels=some<CRegressionLabels>(lab);
+	auto labels_krr=some<CRegressionLabels>(lab);
+
+	/* kernel */
+	auto kernel=some<CGaussianKernel>(features, features, 10, 0.5);
+	auto kernel_krr=some<CGaussianKernel>(features_krr, features_krr, 10, 0.5);
+
+	/* kernel ridge regression and the nystrom approximation */
+	float64_t tau=0.01;
+	auto nystrom=some<CKRRNystrom>(tau, num_vectors, kernel, labels);
+	auto krr=some<CKernelRidgeRegression>(tau, kernel_krr, labels_krr);
+
+	nystrom->train();
+	krr->train();
+
+	SGVector<float64_t> alphas=nystrom->get_alphas();
+	SGVector<float64_t> alphas_krr=krr->get_alphas();
+
+	for (index_t i=0; i<num_vectors; ++i)
+		EXPECT_NEAR(alphas[i], alphas_krr[i], 1E-1);
+
+	auto result=nystrom->apply_regression(test_features);
+	auto result_krr=krr->apply_regression(test_features);
+
+	for (index_t i=0; i<num_vectors; ++i)
+		EXPECT_NEAR(result->get_label(i), result_krr->get_label(i), 1E-5);
+}
+
+/**
+ * Test the main algorithm by comparison of the alphas and the predictions,
+ * using a subset of the columns.
+ */
+TEST(KRRNystrom, apply_and_compare_to_KRR_with_column_subset)
+{
+	/* data matrix dimensions */
+	index_t num_vectors=100;
+	index_t num_features=1;
+	index_t num_basis_rkhs=50;
+
+	/* training label data */
+	SGVector<float64_t> lab(num_vectors);
+
+	/* fill data matrix and labels */
+	SGMatrix<float64_t> train_dat(num_features, num_vectors);
+	SGMatrix<float64_t> test_dat(num_features, num_vectors);
+	for (index_t i=0; i<num_vectors; ++i)
+	{
+		/* labels are linear plus noise */
+		float64_t point=(float64_t)i*10/num_vectors;
+		lab.vector[i]=point+CMath::normal_random(0, 1.0);
+		train_dat.matrix[i]=point;
+		test_dat.matrix[i]=point;
+	}
+
+	/* training features */
+	auto features=some<CDenseFeatures<float64_t>>(train_dat);
+	auto features_krr=some<CDenseFeatures<float64_t>>(train_dat);
+
+	/* testing features */
+	auto test_features=some<CDenseFeatures<float64_t>>(test_dat);
+
+	/* training labels */
+	auto labels=some<CRegressionLabels>(lab);
+	auto labels_krr=some<CRegressionLabels>(lab);
+
+	/* kernel */
+	auto kernel=some<CGaussianKernel>(features, features, 10, 0.5);
+	auto kernel_krr=some<CGaussianKernel>(features_krr, features_krr, 10, 0.5);
+
+	/* kernel ridge regression and the nystrom approximation */
+	float64_t tau=0.01;
+	auto nystrom=some<CKRRNystrom>(tau, num_basis_rkhs, kernel, labels);
+	auto krr=some<CKernelRidgeRegression>(tau, kernel_krr, labels_krr);
+
+	nystrom->train();
+	krr->train();
+
+	auto result=nystrom->apply_regression(test_features);
+	auto result_krr=krr->apply_regression(test_features);
+
+	for (index_t i=0; i<num_vectors; ++i)
+		EXPECT_NEAR(result->get_label(i), result_krr->get_label(i), 1E-1);
+}
+
+#endif /* HAVE_CXX11 */


karlnapf:
There is a problem with this implementation: It doesn't save any computation ;)
I will comment in-line

src/shogun/regression/KrrNystrom.cpp
```diff
@@ -0,0 +1,79 @@
...
+void CKrrNystrom::init()
+{
+	// Initialize parameters
+	m_m=1000;  // TODO change to what seems to make sense
+	// TODO check that less than n
+}
+
+bool CKrrNystrom::solve_krr_system()
+{
+	SGMatrix<float64_t> kernel_matrix(kernel->get_kernel_matrix());
```
karlnapf: For Nystrom, the point is *not* to compute the full kernel matrix, but only the subset you subsample below. Otherwise there is no gain in memory usage

src/shogun/regression/KrrNystrom.cpp
```diff
@@ -0,0 +1,79 @@
...
+		kernel_matrix(i,i)+=m_tau;
+
+	// Create random submatrix
+	SGVector<int32_t> col(m_m);
+	col.random(0,n-m_m+1);
+	CMath::qsort(col.vector, m_m);
+	SGMatrix<float64_t> K_mm(m_m,m_m);
+	SGMatrix<float64_t> K_nm(n,m_m);
+	for (index_t i=0; i<m_m*m_m; ++i)
+		K_mm[i]=kernel_matrix(col[i/m_m]+i/m_m,col[i%m_m]+i%m_m);
```
karlnapf: Choosing the indices for the random sub-matrix should come from a separate method. Then we can overload it and use different ways to do this: Incomplete Cholesky, uniform (as you do here), and leverage scores. All with a simple one-liner in a new class

src/shogun/regression/KrrNystrom.cpp
```diff
@@ -0,0 +1,79 @@
...
+
+	// Create Eigen3 objects
+	Map<MatrixXd> K_mm_eigen(K_mm.matrix, m_m, m_m);
+	Map<MatrixXd> K_nm_eigen(K_nm.matrix, n, m_m);
+	Map<VectorXd> y_eigen(y.vector, n);
+	Map<VectorXd> alphas_eigen(m_alpha.vector, n);
+
+	// Perform eigendecomposition of K_mm and calculate approximate
+	// eigendecomposition (U*D*U^T) of original kernel matrix
+	SelfAdjointEigenSolver<MatrixXd> solver(K_mm_eigen);
```
karlnapf: Is computing the Eigendecomposition really the way to go?
Check eq5 of this paper: http://arxiv.org/pdf/1507.04717v6.pdf
This is how I would do it ....

src/shogun/regression/KrrNystrom.cpp
```diff
@@ -0,0 +1,79 @@
...
+	// TODO add check for success
+	MatrixXd D=solver.eigenvalues().asDiagonal();
+	MatrixXd eigvec=solver.eigenvectors(); // TODO confirm normalized and vectors along columns
+	MatrixXd U=K_nm_eigen*eigvec*D;
+
+	// Solve system
+	alphas_eigen=U*D.inverse()*U.transpose()*y_eigen; // TODO make sure diagonal inverse efficient
```
karlnapf: If you do it this way, the alphas you get will be m-dimensional, which makes predictions cheaper

examples/undocumented/libshogun/regression_krr_nystrom.cpp
```diff
@@ -0,0 +1,79 @@
...
+#include <shogun/regression/KrrNystrom.h>
+
+using namespace shogun;
+
+void print_message(FILE* target, const char* str)
+{
+	fprintf(target, "%s", str);
+}
+
+void test_krr_nystrom()
```
karlnapf: rather than writing examples, I think we should start with unit tests

karlnapf: this is good for testing internally though.
You could also do a meta example instead ...

src/shogun/regression/KrrNystrom.cpp
karlnapf: Sorry, I just realised you are exactly doing the Moore Pensrose pseudo-inverse. It is a pitty that eigen3 doesnt have that in natively

src/shogun/regression/KrrNystrom.cpp
karlnapf: You need to tell Shogun what the "support vectors" of this solution are then, the alpha vector will be sparse. It is a bit like in SVM, where the alpha vector is sparse as well. This needs to be carefully unit tested in order to make sure it really produces the correct results

src/shogun/regression/KrrNystrom.cpp
```diff
@@ -0,0 +1,79 @@
+
+/*
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Written (W) 2016 Fredrik Hallgren
```
karlnapf: we use the BSD license these days, pick from any existing file

src/shogun/regression/KrrNystrom.cpp
```diff
@@ -0,0 +1,79 @@
...
+	col.random(0,n-m_m+1);
+	CMath::qsort(col.vector, m_m);
+	SGMatrix<float64_t> K_mm(m_m,m_m);
+	SGMatrix<float64_t> K_nm(n,m_m);
+	for (index_t i=0; i<m_m*m_m; ++i)
+		K_mm[i]=kernel_matrix(col[i/m_m]+i/m_m,col[i%m_m]+i%m_m);
+	for (index_t j=0; j<n*m_m; ++j) // TODO Assuming row-first indexing
+		K_nm[j]=kernel_matrix(j/m_m,col[j%m_m]+j%m_m);
+
+	// Create Eigen3 objects
```
karlnapf: this comment is useless, we try to avoid useless source code comments in shogun :)

src/shogun/regression/KrrNystrom.cpp
```diff
@@ -0,0 +1,79 @@
...
+	SGVector<float64_t> y=((CRegressionLabels*)m_labels)->get_labels();
+
+	// Add tau parameter
+	for(index_t i=0; i<n; i++)
+		kernel_matrix(i,i)+=m_tau;
+
+	// Create random submatrix
+	SGVector<int32_t> col(m_m);
+	col.random(0,n-m_m+1);
+	CMath::qsort(col.vector, m_m);
```
karlnapf: we have range_fill

src/shogun/regression/KrrNystrom.cpp
```diff
@@ -0,0 +1,79 @@
...
+}
+
+bool CKrrNystrom::solve_krr_system()
+{
+	SGMatrix<float64_t> kernel_matrix(kernel->get_kernel_matrix());
+	int32_t n=kernel_matrix.num_rows;
+	SGVector<float64_t> y=((CRegressionLabels*)m_labels)->get_labels();
+
+	// Add tau parameter
+	for(index_t i=0; i<n; i++)
```
karlnapf: this should not be added on the fly I guess, but on the mm matrix see again http://arxiv.org/pdf/1507.04717v6.pdf

src/shogun/regression/KrrNystrom.cpp
```diff
@@ -0,0 +1,79 @@
...
+: CKernelRidgeRegression(tau, k, lab)
+{
+	init();
+
+	m_m=m;
+}
+
+void CKrrNystrom::init()
+{
+	// Initialize parameters
```
karlnapf: not a useful comment :)

src/shogun/regression/KrrNystrom.h
```diff
@@ -0,0 +1,58 @@
...
+#define _KRRNYSTROM_H__
+
+#include <shogun/regression/KernelRidgeRegression.h>
+
+namespace shogun {
+
+class CKrrNystrom : public CKernelRidgeRegression
+{
+public:
+	MACHINE_PROBLEM_TYPE(PT_REGRESSION); // Seems need to be set for all classes
```
karlnapf: Yes, so  the comment can go :)

src/shogun/regression/KrrNystrom.h
```diff
@@ -0,0 +1,58 @@
+/*
+ * This program is free software; you can redistribute it and/or modify
```
karlnapf: bsd please

src/shogun/regression/KrrNystrom.h
```diff
@@ -0,0 +1,58 @@
...
+#include <shogun/regression/KernelRidgeRegression.h>
+
+namespace shogun {
+
+class CKrrNystrom : public CKernelRidgeRegression
+{
+public:
+	MACHINE_PROBLEM_TYPE(PT_REGRESSION); // Seems need to be set for all classes
+
+	// Default constructor
```
karlnapf: For later iterations: please make this doxygen comformant, so that class docs can be generated automatically, see http://www.shogun-toolbox.org/doc/en/3.0.0/classshogun_1_1CKernelRidgeRegression.html

src/shogun/regression/KrrNystrom.h
```diff
@@ -0,0 +1,58 @@
...
+ */
+
+#ifndef _KRRNYSTROM_H__
+#define _KRRNYSTROM_H__
+
+#include <shogun/regression/KernelRidgeRegression.h>
+
+namespace shogun {
+
+class CKrrNystrom : public CKernelRidgeRegression
```
karlnapf: Please add a description with the basic math in the class description.
See http://www.shogun-toolbox.org/doc/en/3.0.0/classshogun_1_1CKernelRidgeRegression.html

karlnapf: I think it should be `CKRRNystrom`

src/shogun/regression/KrrNystrom.h
karlnapf: Sorry this one renders correctly:
http://www.shogun-toolbox.org/doc/en/latest/classshogun_1_1CKernelRidgeRegression.html

src/shogun/regression/KrrNystrom.cpp
karlnapf: See also http://eigen.tuxfamily.org/index.php?title=FAQ#Is_there_a_method_to_compute_the_.28Moore-Penrose.29_pseudo_inverse_.3F
BTW you need to set a reasonable precision here as well, see https://en.wikipedia.org/wiki/Moore%E2%80%93Penrose_pseudoinverse#Singular_value_decomposition_.28SVD.29

cfjhallgren:
In Seeger et al the alphas have dimension n but the methods appear equivalent. I'll write the maths in the documentation.

However I noticed a mistake in the code, I have yet to solve the final system of equations after doing the pseudo inverse. I will fix this and also attempt to refactor to not use the full kernel matrix for memory efficiency.

cfjhallgren:
Happy to use equation 5 in Rudi et al too.

src/shogun/regression/KrrNystrom.h
cfjhallgren: Changed this :)

src/shogun/regression/KrrNystrom.h
cfjhallgren: It's gone :)

cfjhallgren:
Pushed some changes but seems the rebase didn't work, will try again

cfjhallgren:
Ok now it seems the rebase worked.

Kernel approximation is given by  K_{nm}K^+_{mm}K_{mn} and I solve the system with the woodbury formula (+ denotes pseudo-inverse)

Next up:
- don't calculate full kernel matrix
- checks for m_m < n
- move sampling to separate function
- doxygen compliant comments
- more documentation
- BSD

cfjhallgren:
Seems the checks pass now as well :)

cfjhallgren:
A couple 'TODO's still left in the code. I think I addressed most of your previous comments. Currently in the process of writing some unit tests.

cfjhallgren:
Caveat: haven't tested that the class docs render correctly

cfjhallgren:
Also the separate sampling method is WIP

examples/undocumented/libshogun/regression_krr_nystrom.cpp
```diff
@@ -0,0 +1,100 @@
...
+{
```
karlnapf: BTW for the meta example, have a look here for inspiration
http://shogun.ml/cookbook/latest/examples/regression/kernel_ridge_regression.html
The two files from which is generated are here:
https://github.com/shogun-toolbox/shogun/blob/develop/doc/cookbook/source/examples/regression/kernel_ridge_regression.rst
https://github.com/shogun-toolbox/shogun/blob/develop/examples/meta/src/regression/kernel_ridge_regression.sg

karlnapf: And the test file for integration testing is here:
https://github.com/shogun-toolbox/shogun-data/blob/master/testsuite/meta/regression/kernel_ridge_regression.dat
This is a separate repo and has to be sent in a separate patch.
See the Readme of the meta examples

src/shogun/regression/KRRNystrom.cpp
```diff
@@ -0,0 +1,106 @@
...
+void CKRRNystrom::init()
+{
+	m_m=1000;
+}
+
+SGVector<int32_t> CKRRNystrom::sample_index()
+{
+	int32_t n=kernel->get_num_vec_lhs();
+	SGVector<int32_t> col(m_m);
+	col.random(0,n-m_m);
```
karlnapf: I think this one here samples with replacement.
Why not just create a range_fill [0, ...., n-1] and then permute?

src/shogun/regression/KRRNystrom.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+	m_m=m;
+}
+
+void CKRRNystrom::init()
+{
+	m_m=1000;
+}
+
+SGVector<int32_t> CKRRNystrom::sample_index()
```
karlnapf: maybe a better name is `subsample_indices`?

src/shogun/regression/KRRNystrom.cpp
```diff
@@ -0,0 +1,106 @@
...
+	CMath::qsort(col.vector, m_m);
+	for (index_t i=0; i<m_m; ++i)
+		col[i] += i;
+
+	return col;
+}
+
+bool CKRRNystrom::solve_krr_system()
+{
+	int32_t n=kernel->get_num_vec_lhs(); // TODO ever different from rhs?
```
karlnapf: For training, the lhs is equal to rhs

src/shogun/regression/KRRNystrom.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+	ASSERT(m_m <= n);
```
karlnapf: Can you use REQUIRE here and write a proper error message?

src/shogun/regression/KRRNystrom.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+	SGVector<float64_t> y=((CRegressionLabels*)m_labels)->get_labels();
+	SGVector<int32_t> col=sample_index();
+	SGMatrix<float64_t> K_mm(m_m,m_m);
+	SGMatrix<float64_t> K_nm(n,m_m);
+	for (index_t i=0; i<m_m*m_m; ++i)
+		K_mm[i]=kernel->kernel(col[i/m_m],col[i%m_m]);
```
karlnapf: You can index matrices like `K_mm[row, col]` with a double nested loop here.
But actually it would be better if you set a column subset on the features, and then just ask the kernel to give you the K_nm kernel matrix (which will be of the correct size, and computed using multiple cores). Then you simpy copy the values to K_mm rather than re-computing it.

src/shogun/regression/KRRNystrom.cpp
```diff
@@ -0,0 +1,106 @@
...
+	MatrixXd Kplus=m_tau*K_mm_eig + K_mn_eig*K_nm_eig;
+	SelfAdjointEigenSolver<MatrixXd> solver(Kplus);
+	if (solver.info()!=Success)
+	{
+		SG_WARNING("Eigendecomposition failed.")
+		return false;
+	}
+	/* Solve the system for alphas */
+	MatrixXd D=solver.eigenvalues().asDiagonal();
+	MatrixXd eigvec=solver.eigenvectors(); // TODO confirm normalized and vectors along columns
```
karlnapf: I think you should set a tolerance level here, have a look at the wikipedia entry where they describe how it is say done in numpys pinv

src/shogun/regression/KRRNystrom.cpp
```diff
@@ -0,0 +1,106 @@
...
+	MatrixXd pseudoinv=eigvec*D.inverse()*eigvec.transpose(); // TODO confirm diagonal inverse efficient
+	alphas_eig=1.0/m_tau*(y_eig-K_nm_eig*pseudoinv*K_mn_eig*y_eig); // TODO maybe split up
```
karlnapf: this is fine, no need to split

src/shogun/regression/KRRNystrom.h
```diff
@@ -0,0 +1,134 @@
...
+
+namespace shogun {
+
+/** @brief Class KRRNystrom implements the Nyström method for kernel ridge
+ * regression, using a low-rank approximation to the kernel matrix.
+ *
+ * The method is equivalent to ordinary kernel ridge regression, but through a
+ * low-rank approximation to the original kernel matrix, the full kernel matrix
+ * does not need to be computed, and the resulting system of equations for the
+ * alphas is cheaper to solve.
```
karlnapf: Maybe state the Big-O complexity here?

src/shogun/regression/KRRNystrom.h
```diff
@@ -0,0 +1,134 @@
...
+ * transpose and \f$K_{m,m} is the submatrix with the m rows and columns
+ * corresponding to the training examples chosen. \f$+\f$ indicates the
+ * Moore-Penrose pseudoinverse.
+ *
+ * The resulting system of equations for the alphas can then be solved through
+ * the Woodbury formula, which gives the following equation for the alphas
+ *
+ * \f[
+ * {\bf \alpha} = \frac{1}{\tau} ({\bf y} - K_{n,m} (\tau K_{m,m} + K_{m,n}K_{n,m})^+K_{m,n} {\bf y}
+ * ]
```
karlnapf: Maybe state that there are multiple ways to select columns. And that overloading the "subsample_indices" method can alter that

karlnapf: And that we are uniformly sub-sampling in this class

src/shogun/regression/KRRNystrom.h
```diff
@@ -0,0 +1,134 @@
...
+	 *
+	 * @param m new m
+	 */
+	inline void set_m(int32_t m) { m_m = m; };
+
+	/** Set the regularization parameter
+	 *
+	 * @param tau new tau
+	 */
+	inline void set_tau(float64_t tau) { m_tau = tau; };
```
karlnapf: This shoud be inherited from the base class no?

src/shogun/regression/KRRNystrom.h
```diff
@@ -0,0 +1,134 @@
...
+	virtual const char* get_name() const { return "KRRNystrom"; }
+
+	/**
+	 * Get the classifier type
+	 *
+	 * @return classifier type KernelRidgeRegression
+	 */
+	EMachineType get_classifier_type()
+	{
+		return CT_KERNELRIDGEREGRESSION;
```
karlnapf: this is not classification and it is not pure ridge regression. You can add a new type if you want here. Or just remove this method as we can also just use dynamic_cast

src/shogun/regression/KRRNystrom.h
```diff
@@ -0,0 +1,134 @@
...
+	/** Sample indices to pick rows/columns from kernel matrix
+	 *
+	 * @return SGVector<int32_t> with sampled indices
+	 */
+	SGVector<int32_t> sample_index();
+private:
+	void init();
+
+	/** Number of columns/rows to be sampled */
+	int32_t m_m;
```
karlnapf: This should be protected as well: sub-classes might want to read it

src/shogun/regression/KRRNystrom.h
```diff
@@ -0,0 +1,134 @@
...
+	{
+		return CT_KERNELRIDGEREGRESSION;
+	}
+
+protected:
+	/** Train regression using the Nyström method.
+	 *
+	 * @return boolean to indicate success
+	 */
+	bool solve_krr_system();
```
karlnapf: This method needs to be virtual btw

src/shogun/regression/KernelRidgeRegression.h
```diff
@@ -125,15 +125,16 @@ class CKernelRidgeRegression : public CKernelMachine
 		 *
 		 * @return boolean to indicate success
 		 */
-		bool solve_krr_system();
+		virtual bool solve_krr_system();
```
karlnapf: Ah there we go :) Nevermind the above comment

src/shogun/regression/KernelRidgeRegression.h
```diff
@@ -125,15 +125,16 @@ class CKernelRidgeRegression : public CKernelMachine
...
 
 	private:
 		void init();
 
-	private:
+	protected:
```
karlnapf: yep! See also above comment

karlnapf:
Cool stuff! This is almost ready!

Nevermind the different sub-sampling techniques for now.
Next steps:
1.) Clean up things (resolve all TODOs etc, remove the example code from the patch for now)
2.) Write unit test (at least two cases: ensuring that for m=n the results are the same, and that for a smaller m, the results are at least similar)
3.) Squash all commits into a single one
Then we can merge. Then
4.) Meta example similar to the one I posted
5.) Data patch with integration test
6.) Cookbook page (if you want)

cfjhallgren:
Cool!

I have been away in Sweden for a few days but I'm back in London tomorrow :)

In the meantime I removed the example code and did a unit test and the code seems to work. Initial results (alphas in particular) are very similar to the original KRR method. When I add a tolerance level they differ more (will investigate).

I addressed most of your comments above - will push soon :)

src/shogun/regression/KRRNystrom.h
cfjhallgren: Good point :)

src/shogun/regression/KRRNystrom.h
cfjhallgren: Good point

src/shogun/regression/KRRNystrom.h
cfjhallgren: Added :)

src/shogun/regression/KRRNystrom.cpp
cfjhallgren: I first sample with replacement then adjust by sorting and adding an increasing sequence of numbers from 0 to m-1. Maybe permuting is cleaner.

src/shogun/regression/KRRNystrom.cpp
cfjhallgren: Yes sounds better :)

cfjhallgren:
Just pushed  some changes and unit tests. 

In the main unit test I use a simple linear example with noise and RBF kernel and compare nystrom with the original using 10 data points each (so results should be the same).

The first four alphas are as follows:

nystrom                            krr
-1379.9329053440129    -1380.0338757362974
2773.579320582402        2773.4428957324544, and
-3522.11899622485        -3522.2550535189807
2745.6471890153457       2745.4287106676047

First four predicted labels:

nystrom                             krr
0.88485147755566351     0.52979802894051109
0.81868948607437175     0.30734515172343291
2.7678414033546024       2.0999106183636242
4.7506759400829015       3.9516436781741486

I will continue investigasting.

cfjhallgren:
Notice some checks didn't pass - maybe the DBL_EPSILON constant is not available

src/shogun/regression/KRRNystrom.h
cfjhallgren: I removed it.

src/shogun/regression/KRRNystrom.h
cfjhallgren: Good idea, added it.

cfjhallgren:
The subsample_indices is protected so I didn't test it, but when I printed the indices for the example above (i.e. sampling all indices) the result was as expected with indices 1,2, ...., 10. Will try to test some more.

cfjhallgren:
The results become more similar as the tau parameter is increased (both absolute and relative difference) - which is not strange perhaps since it would make the system more well-conditioned. This might also hint at that the cause of the difference might be related to the accuracy of the respective solution methods although I'm mostly guessing here.

cfjhallgren:
Actually I'm not sure about the relative difference.

I'm thinking if there is some other way to solve the nystrom system...

cfjhallgren:
I'll continue investigating.

src/shogun/regression/KRRNystrom.cpp
```diff
@@ -0,0 +1,126 @@
...
+void CKRRNystrom::init()
+{
+	m_m=1000;
+}
+
+SGVector<int32_t> CKRRNystrom::subsample_indices()
+{
+	int32_t n=kernel->get_num_vec_lhs();
+	SGVector<int32_t> col(m_m);
+	col.random(0,n-m_m);
```
karlnapf: can you do
`col.range_fill();
CMath::permute(col);`

src/shogun/regression/KRRNystrom.cpp
```diff
@@ -0,0 +1,126 @@
...
+	REQUIRE(m_m <= n, "Number of sampled rows/columns must be less than total\n");
+
+	SGVector<float64_t> y=((CRegressionLabels*)m_labels)->get_labels();
+	SGVector<int32_t> col=subsample_indices();
+	SGMatrix<float64_t> K_mm(m_m,m_m);
+	SGMatrix<float64_t> K_nm(n,m_m);
+	for (index_t i=0; i<m_m; ++i)
+	{
+		for (index_t j=0; j<m_m; ++j)
+		{
```
karlnapf: minor, we dont do {} for single lines

src/shogun/regression/KRRNystrom.cpp
```diff
@@ -0,0 +1,126 @@
...
+{
+	int32_t n=kernel->get_num_vec_lhs();
+
+	REQUIRE(m_m <= n, "Number of sampled rows/columns must be less than total\n");
+
+	SGVector<float64_t> y=((CRegressionLabels*)m_labels)->get_labels();
+	SGVector<int32_t> col=subsample_indices();
+	SGMatrix<float64_t> K_mm(m_m,m_m);
+	SGMatrix<float64_t> K_nm(n,m_m);
+	for (index_t i=0; i<m_m; ++i)
```
karlnapf: Can you actually use the kernel->get_kernel_matrix() interface for this? It is all parallelised etc.
You can set a subset to the features using your `col` variable before

src/shogun/regression/KRRNystrom.cpp
```diff
@@ -0,0 +1,126 @@
...
+	MatrixXd Kplus=m_tau*K_mm_eig+K_mn_eig*K_nm_eig;
+	SelfAdjointEigenSolver<MatrixXd> solver(Kplus);
+	if (solver.info()!=Success)
+	{
+		SG_WARNING("Eigendecomposition failed.")
+		return false;
+	}
+
+	/* Solve the system for alphas */
+	MatrixXd D=solver.eigenvalues().asDiagonal();
```
karlnapf: We should have a method `pinv` in CMath....
I wrote the same in a feature branch and it is not good to have duplicate code.
I will take care of this later, no action required here for now

tests/unit/regression/krrnystrom_unittest.cc
```diff
@@ -0,0 +1,133 @@
...
+using namespace shogun;
+
+/**
+ * Test that the expected interface is available.
+ */
+TEST(KRRNystrom, test_interface)
+{
+
+	/* Use default constructor */
+	CKRRNystrom krr;
```
karlnapf: please always allocate Shogun objects on the heap, i.e. use `new` and `SG_REF/SG_UNREF`

Alternatively, you can use `some`
`auto kernel = some<CGaussianKernel>();`
For that, you need to include it and guard the code with `#ifdef HAVE_CXX11`

tests/unit/regression/krrnystrom_unittest.cc
```diff
@@ -0,0 +1,133 @@
...
+	for (index_t i=0; i<num_vectors; ++i)
+	{
+		/* labels are linear plus noise */
+		lab.vector[i]=i+CMath::normal_random(0, 1.0);
+		train_dat.matrix[i]=i;
+		test_dat.matrix[i]=i;
+	}
+
+	/* training features */
+	CDenseFeatures<float64_t>* features=
```
karlnapf: you can also use C++11 and some here if you want, allows for easier reference managment

tests/unit/regression/krrnystrom_unittest.cc
```diff
@@ -0,0 +1,133 @@
...
+	CKernelRidgeRegression* krr=new CKernelRidgeRegression(tau, kernel_krr, labels_krr);
+
+	nystrom->train();
+	krr->train();
+
+	SGVector<float64_t> alphas=nystrom->get_alphas();
+	SGVector<float64_t> alphas_krr=krr->get_alphas();
+
+	for (index_t i=0; i<num_vectors; ++i) {
+		EXPECT_NEAR(alphas[i], alphas_krr[i], 1E-1);
```
karlnapf: this should be 1e-5 at least in my experience if all of the matrix is used (only difference should be the numerics of using the SVD to invert the kernel matrix)

tests/unit/regression/krrnystrom_unittest.cc
```diff
@@ -0,0 +1,133 @@
...
+	CKRRNystrom* nystrom=new CKRRNystrom(tau, num_vectors, kernel, labels);
+	CKernelRidgeRegression* krr=new CKernelRidgeRegression(tau, kernel_krr, labels_krr);
+
+	nystrom->train();
+	krr->train();
+
+	SGVector<float64_t> alphas=nystrom->get_alphas();
+	SGVector<float64_t> alphas_krr=krr->get_alphas();
+
+	for (index_t i=0; i<num_vectors; ++i) {
```
karlnapf: minor { } is always in a new line, when there is only one line, we dont use it

tests/unit/regression/krrnystrom_unittest.cc
```diff
@@ -0,0 +1,133 @@
...
+TEST(KRRNystrom, test_get_classifier_type)
+{
+	CKRRNystrom krr;
+	EXPECT_EQ(krr.get_classifier_type(), CT_KERNELRIDGEREGRESSION);
+}
+
+/**
+ * Test the main algorithm by comparison of the alphas and predictions
+ */
+TEST(KRRNystrom, compare_to_KRR)
```
karlnapf: name the test so that it states that it compares using *all* columns
Then, there should be another test that uses only say half of the columns (with a lower accuracy)

tests/unit/regression/krrnystrom_unittest.cc
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	CRegressionLabels* result=(CRegressionLabels *)nystrom->apply(test_features);
+	CRegressionLabels* result_krr=(CRegressionLabels *)krr->apply(test_features);
+
+	for (index_t i=0; i<num_vectors; ++i) {
+		EXPECT_NEAR(result->get_label(i), result_krr->get_label(i),1E-1);
+	}
+
+	SG_UNREF(nystrom);
```
karlnapf: not needed if you use `some`

src/shogun/regression/KRRNystrom.h
```diff
@@ -0,0 +1,134 @@
...
+
+	/** Sample indices to pick rows/columns from kernel matrix
+	 *
+	 * @return SGVector<int32_t> with sampled indices
+	 */
+	SGVector<int32_t> subsample_indices();
+
+	/** Number of columns/rows to be sampled */
+	int32_t m_m;
```
karlnapf: BTW sorry for only bringing this up now. But maybe the name "m_num_rkhs_basis" is better than m. Minor though

karlnapf:
Hi,
sorry about the delay.

About the results.
I would solve the KRR system to get an alpha that has dimension m. This makes evaluating the regression function cheaper (kernel is computed against less point).
For that you need to either set the support vectors and translate the alpha (filling zeros where needed), or overload the apply method of the base class. I have the feeling that your error is coming from something related to this

karlnapf:
Let me know if you keep on having problems, I can pull your feature branch and potentially help

Nice one, we are getting there, the code is soon ready :)

cfjhallgren:
Hi sorry for the delay.

I was over at gatsby today for the journal club :)

I will look into the apply function and try calculating the alphas with dimension m.

karlnapf:
No worries,

I was out for work today. 
Let's get this merged, it is close :)

cfjhallgren:
Sounds good :)

src/shogun/regression/KRRNystrom.cpp
cfjhallgren: Will do.

src/shogun/regression/KRRNystrom.cpp
cfjhallgren: Will do.

tests/unit/regression/krrnystrom_unittest.cc
cfjhallgren: Fixed.

src/shogun/regression/KRRNystrom.cpp
cfjhallgren: Fixed.

tests/unit/regression/krrnystrom_unittest.cc
cfjhallgren: Fixed.

src/shogun/regression/KRRNystrom.h
cfjhallgren: Good name :)

tests/unit/regression/krrnystrom_unittest.cc
cfjhallgren: Will try using 'some'.

cfjhallgren: Now using 'some' in the whole test file. I saw SG_REF/SG_UNREF is disappearing?

cfjhallgren:
Implemented the Rudi et al method in 'solve_krr_system', which seems to give somewhat better results for m = n. Both methods are somewhat sensitive the tau and differences become smaller as tau increases.

Will now update the 'apply' function(s) for when m < n.

cfjhallgren:
Initially had planned to set 'm_svs' to the smaller set of alphas but found no good place to do this, so set 'm_alpha' to size 'n' and filled with zeros. Almost done.

tests/unit/regression/krrnystrom_unittest.cc
karlnapf: Yes you can avoid the SG_REFs in that case

karlnapf:
It doesnt really matter how it is done, important is that it is consistent :)
Cool this is really almost done now

cfjhallgren:
Now it's done and seems to work. Tests pass with all columns and half of the columns (the latter a bit sensitive to the number of training samples in the test).

As mentioned some sensitivity to the tau parameter, if I increase tau from 0.01 to e.g. 0.001 tests fail, but only just.

So now possibly the only things that are left are to update the documentation to reflect more the rudi et al methodology, and possibly add another test that is a bit different.

Oh also haven't done the change you suggested on calculating the kernel matrix, need to look into this.

Will push tomorrow, had some merging issues.

cfjhallgren:
Sorry that should be 'decrease' from 0.01 to 0.001

karlnapf:
Great stuff! Looking forward to merge this. You can also add an entry to the NEWS giving yourself credit.

cfjhallgren:
Me too :)

Just pushed the rudi method changes and other fixes. Will do the pending items above.

src/shogun/regression/KRRNystrom.cpp
```diff
@@ -0,0 +1,128 @@
...
+	CMath::qsort(col.vector, m_num_rkhs_basis);
+
+	return col;
+}
+
+bool CKRRNystrom::solve_krr_system()
+{
+	int32_t n=kernel->get_num_vec_lhs();
+
+	REQUIRE(m_num_rkhs_basis <= n, "Number of sampled rows/columns must be less than total\n");
```
karlnapf: Number of subsampled columns (%d) must be less than number of data (%d).

I.e. give the user the information what exactly is wrong

src/shogun/regression/KRRNystrom.cpp
```diff
@@ -0,0 +1,128 @@
...
+
+	SGVector<float64_t> y=((CRegressionLabels*)m_labels)->get_labels();
```
karlnapf: Either make sure this cast is valid (dynamic cast, check for NULL), or put a comment that this is ensured somewhere else

src/shogun/regression/KRRNystrom.cpp
```diff
@@ -0,0 +1,128 @@
...
+	SGVector<int32_t> col=subsample_indices();
+	SGMatrix<float64_t> K_mm(m_num_rkhs_basis, m_num_rkhs_basis);
+	SGMatrix<float64_t> K_nm(n, m_num_rkhs_basis);
+	for (index_t i=0; i<m_num_rkhs_basis; ++i)
```
karlnapf: #pragma omp parallel for

makes this run on multiple cores

src/shogun/regression/KRRNystrom.cpp
```diff
@@ -0,0 +1,128 @@
...
+	{
+		for (index_t j=0; j<m_num_rkhs_basis; ++j)
```
karlnapf: exploit symmetry here, `for index_t j=0; j<=i; j++)` and then set both `K_mm(i,j)` and `K_mm(j,i)`

src/shogun/regression/KRRNystrom.cpp
```diff
@@ -0,0 +1,128 @@
...
+			K_mm(i,j)=kernel->kernel(col[i],col[j]);
+		for (index_t j=0; j<n; ++j)
```
karlnapf: if you flip the loop (i.e. you traverse the matrix column-wise rather than row-wise), it will be more CPU-cache friendly and therefore significantly faster

src/shogun/regression/KRRNystrom.h
```diff
@@ -0,0 +1,116 @@
...
+ * Instead of the original linear system, the following approximate system
+ * is solved
+ *
+ * \f[
+ * {\bf \alpha} = (\tau K_{m,m} + K_{m,n}K_{n,m})^+K_{m,n} {\bf y}
+ * ]
+ *
+ * where \f$K_{n,m}\f$ is a submatrix containing all n rows and the m columns
+ * corresponding to the m chosen training examples, \f$K_{m,n}\f$ is its
+ * transpose and \f$K_{m,m} is the submatrix with the m rows and columns
```
karlnapf: great description :)

src/shogun/regression/KRRNystrom.h
karlnapf: Otherwise subclasses can't override it

karlnapf: Maybe also put the keyword here, so that I wont moan like 10 more times because I forgot you changed the base class ;)

tests/unit/regression/krrnystrom_unittest.cc
```diff
@@ -0,0 +1,186 @@
+/*
```
karlnapf: Usually, our naming scheme is a bit different, see the other files.

Minor ...

tests/unit/regression/krrnystrom_unittest.cc
```diff
@@ -0,0 +1,186 @@
...
+ */
+TEST(KRRNystrom, test_setters)
+{
+	/* Use default constructor */
+	auto krr=some<CKRRNystrom>();
+
+	/* Test setters */
+	krr->set_tau(0.1);
+	krr->set_epsilon(0.1);
+	krr->set_num_rkhs_basis(100);
```
karlnapf: if you really want to test those, you should call the getter afterwards and ensure it returns the same you set

tests/unit/regression/krrnystrom_unittest.cc
```diff
@@ -0,0 +1,186 @@
...
+	nystrom->train();
+	krr->train();
+
+	SGVector<float64_t> alphas=nystrom->get_alphas();
+	SGVector<float64_t> alphas_krr=krr->get_alphas();
+
+	for (index_t i=0; i<num_vectors; ++i)
+		EXPECT_NEAR(alphas[i], alphas_krr[i], 1E-1);
+
+	auto result=(CRegressionLabels *)nystrom->apply(test_features);
```
karlnapf: use `apply_regression` which will directly give you regression labels and you can avoid the cast

karlnapf:
Travis complains about your get_name unit test:
https://travis-ci.org/shogun-toolbox/shogun/jobs/133729677#L3755

From my side, you can as well just remove it

karlnapf:
Ok, this is 1 step from being merged

 * fix the remaining minor things, in particular the travis
 * squash all commits into a single one
 * we can merge :)

Next steps:
 * meta example (would be cool, ask for help if you wanna write one)
 * a little benchmark against full KRR
 * potentially an update of the ipython notebook where KRR is introduced
 * you choose :)

cfjhallgren:
Great :)

cfjhallgren:
One guess is that the travis error is the DBL_EPSILON constant - I included <cfloat> which I hoped would help but seems it didn't.

src/shogun/regression/KRRNystrom.cpp
```diff
@@ -0,0 +1,128 @@
...
+	if (solver.info()!=Success)
+	{
+		SG_WARNING("Eigendecomposition failed.")
+		return false;
+	}
+
+	/* Solve the system for alphas */
+	MatrixXd D=solver.eigenvalues().asDiagonal();
+	MatrixXd eigvec=solver.eigenvectors();
+	const float64_t tolerance=m_num_rkhs_basis*DBL_EPSILON*D.maxCoeff();
```
karlnapf: I think this is (or should be) defined in CMath ... it is not portable this way. Might be a reason for an error

cfjhallgren: Great will check in CMath

src/shogun/regression/KRRNystrom.h
cfjhallgren: Thanks :)

src/shogun/regression/KRRNystrom.cpp
cfjhallgren: Done.

karlnapf: Dont see it
I mean this
`REQUIRE(m_num_rkhs_basis <= n, "Number of sampled columns (%d) cannot be larger than number of data (%d).\n", m_num_rkhs_basis, n);`

src/shogun/regression/KRRNystrom.cpp
cfjhallgren: Good idea.

src/shogun/regression/KRRNystrom.cpp
cfjhallgren: Cool

src/shogun/regression/KRRNystrom.cpp
cfjhallgren: Done.

tests/unit/regression/krrnystrom_unittest.cc
cfjhallgren: I removed the test.

tests/unit/regression/krrnystrom_unittest.cc
cfjhallgren: Changed names a bit.

tests/unit/regression/krrnystrom_unittest.cc
cfjhallgren: Great.

src/shogun/regression/KRRNystrom.cpp
cfjhallgren: Added a check

src/shogun/regression/KRRNystrom.cpp
```diff
@@ -0,0 +1,139 @@
...
+	CMath::qsort(col.vector, m_num_rkhs_basis);
+
+	return col;
+}
+
+bool CKRRNystrom::solve_krr_system()
+{
+	int32_t n=kernel->get_num_vec_lhs();
+
+	REQUIRE(m_num_rkhs_basis <= n, "Number of sampled rows/columns must be less than number of data\n");
```
karlnapf: Should really print out both m_num_rkhs_basis and n here. Otherwise users might be confused.
Also, the check should be done in the constructor and setter rather than here.

cfjhallgren: Good point.

src/shogun/regression/KRRNystrom.cpp
```diff
@@ -0,0 +1,139 @@
...
+
+	SGVector<float64_t> y=((CRegressionLabels*)m_labels)->get_labels();
+	if (y==NULL)
+		SG_ERROR("Labels not set.");
```
cfjhallgren: Maybe I should raise a SG_WARNING and return 'false' instead like below.

Also I saw this cast in a couple of other places, so maybe it is checked somewhere like you mentioned that I didn't see.

karlnapf: No an error is fine, but it needs a newline at the end

src/shogun/regression/KRRNystrom.cpp
```diff
@@ -0,0 +1,139 @@
...
+	MatrixXd K_mn_eig = K_nm_eig.transpose();
+	Map<VectorXd> y_eig(y.vector, n);
+	VectorXd alphas_eig(m_num_rkhs_basis);
+
+	/* Calculate the Moore-Penrose pseudoinverse */
+	MatrixXd Kplus=K_mn_eig*K_nm_eig+m_tau*K_mm_eig;
+	SelfAdjointEigenSolver<MatrixXd> solver(Kplus);
+	if (solver.info()!=Success)
+	{
+		SG_WARNING("Eigendecomposition failed.")
```
karlnapf: newline missing

src/shogun/regression/KRRNystrom.cpp
```diff
@@ -0,0 +1,139 @@
...
+		for (index_t j=0; j<=i; ++j)
+		{
+			K_mm(i,j)=kernel->kernel(col[i],col[j]);
+			K_mm(j,i)=K_mm(i,j);
+		}
+	}
+	#pragma omp parallel for
+	for (index_t i=0; i<n; ++i)
+	{
+		for (index_t j=0; j<m_num_rkhs_basis; ++j)
```
karlnapf: why not swap the loop order here too for cache friendliness?

src/shogun/regression/KRRNystrom.cpp
```diff
@@ -0,0 +1,139 @@
...
+			K_nm(i,j)=kernel->kernel(i,col[j]);
```
karlnapf: You are doing redundant computation here:
I suggest
 * first compute K_nm
 * then copy the data into K_mm using `memcpy` and `SGMatrix::get_column_vector`

karlnapf:
Ok travis is fine. So as soon as we have the kernel matrix computation sorted and done efficiently, we can squash the commits and merge.

src/shogun/regression/KRRNystrom.cpp
cfjhallgren: Done.

src/shogun/regression/KRRNystrom.cpp
cfjhallgren: Added it to the constructor and setter instead (for the latter after checking "kernel!=NULL"). Also printed out the numbers.

karlnapf: Did you push? Dont see the update here

cfjhallgren:
Had some issues with memcpy but pushed the other changes in the meanwhile.

karlnapf:
so when all is done, the commits need to be squashed and we can merge.

cfjhallgren:
Great will try to complete it today :)

cfjhallgren:
Now memcpy works :) Will squash commits later today.

cfjhallgren:
Squashed commits :)

cfjhallgren:
Will start looking at meta example while awaiting comments :)

cfjhallgren:
Will try running example from rudi or seeger first I think.

cfjhallgren:
Think I'll try to reproduce the examples in rudi. Is there somewhere to put C++ examples or are all meta examples in python?

cfjhallgren:
At the same time I'll do performance comparison

karlnapf:
Ok, fine to merge from my side.
There are a few minor things (I will comment) that you can address in a separate patch.
E.g. the NEWS file should be updated with this.

You can try to add a meta example next, see all the other open PRs.
See the readme [here](https://github.com/shogun-toolbox/shogun/tree/develop/doc/cookbook)

Then yes, performance checks, we usually to this in [this framework](https://github.com/zoq/benchmarks) and mostly compare against sklearn. There are lots of examples where @Saurabh7 did similar benchmark comparisons.

karlnapf:
@vigsterkr feel free to merge once reviewed

src/shogun/regression/KRRNystrom.h
vigsterkr: mmm why isn't this `virtual` if in the parent class it is `virtual`?

vigsterkr:
cool stuff @cfjhallgren !
i only wonder about that `virtual` other than that we could merge this!

src/shogun/regression/KRRNystrom.h
cfjhallgren: Yes perhaps better to have the keyword here too.

cfjhallgren:
Thanks :) I added the virtual keyword now.

cfjhallgren:
Cool will start with meta example.

vigsterkr:
ok, great! let's see what travis says and then i'll merge it!

Issue #3174, opened 2016-04-18
one issue  about using Shogun's optimizers in target languages


yorkerlin:
@karlnapf 
In `CInference` class
```
virtual void register_minimizer(Minimizer* minimizer);
```
In `Minimizer` class
```
#ifndef MINIMIZER_H
#define MINIMIZER_H
#include <shogun/lib/config.h>
namespace shogun
{

/** @brief The minimizer base class.
 *
 */
class Minimizer
{
public: 
        /** Do minimization and get the optimal value 
         * 
         * @return optimal value
         */
        virtual float64_t minimize()=0;

        virtual ~Minimizer() {}
};

}
#endif
```

Note that
* `CInference` is a sub-class of `SGObject`
* `LBFGSMinimizer` class is a sub-class of Minimizer
* `CSingleLaplaceInferenceMethod` is a sub-class of CInference
* `Minimizer` is `NOT` a sub-class of `CSGObject`
 
The following lines of C++ code work.

```
CSingleLaplaceInferenceMethod* inf = new CSingleLaplaceInferenceMethod();
LBFGSMinimizer* opt=new LBFGSMinimizer();
inf->register_minimizer(opt);
```

However, the following lines of Python code do not work 
```
inf=SingleLaplaceInferenceMethod()
opt = LBFGSMinimizer()
inf.register_minimizer(opt)
```
Error output:
```
TypeError: in method 'Inference_register_minimizer', argument 2 of type 'Minimizer *'
```

yorkerlin:
In `src/interfaces/modular/Minimizer.i`
```
%include <shogun/optimization/Minimizer.h>
%include <shogun/optimization/FirstOrderMinimizer.h>
%include <shogun/optimization/lbfgs/lbfgscommon.h>
%include <shogun/optimization/lbfgs/LBFGSMinimizer.h>
```

In `src/interfaces/modular/Minimizer_includes.i`
```
%{
 #include <shogun/optimization/Minimizer.h>
 #include <shogun/optimization/demo.h>
 #include <shogun/optimization/FirstOrderMinimizer.h>
 #include <shogun/optimization/lbfgs/lbfgscommon.h>
 #include <shogun/optimization/lbfgs/LBFGSMinimizer.h>
%}
```

karlnapf:
You can only pass sub-classes of SGObject through swig

yorkerlin:
@karlnapf 
Why can I pass the minimizer to a non-SGObject through swig?  

[...]

PR #3175, opened 2016-04-22
fixed 3174 issue

yorkerlin:
@karlnapf @iglesias @vigsterkr 

https://github.com/shogun-toolbox/shogun/issues/3174


src/interfaces/modular/modshogun.i
```diff
@@ -77,8 +77,8 @@
 %include "Statistics_includes.i"
 %include "Latent_includes.i"
 %include "Metric_includes.i"
-%include "GaussianProcess_includes.i"
 %include "Minimizer_includes.i"
```
yorkerlin: The i files about minimizers muse be defined before the i files about GP .

iglesias:
Order with the includes can mess with anyone very badly! Cool that you detected the issue :-)

Why did you change the using namespace to the block? I thought the convention we were employing the convention of `using namespace` in cpp implementation files.

iglesias:
Oh, now I realized you already merged this directly. It is all right, bu please get back on the namespace query.

PR #3176, opened 2016-04-23
refactor the GP framework 6

yorkerlin:


src/shogun/machine/gp/KLDualInferenceMethod.h
```diff
@@ -39,9 +39,52 @@
...
+protected:
+	/** Init before minimization */
+	virtual void init_minimization()
+	{
+		REQUIRE((m_linesearch == ELBFGSLineSearch::BACKTRACKING_ARMIJO) ||
+			(m_linesearch == ELBFGSLineSearch::BACKTRACKING_WOLFE) ||
+			(m_linesearch == ELBFGSLineSearch::BACKTRACKING_STRONG_WOLFE),
+			"The provided line search method is not supported. Please use backtracking line search methods\n");
+		LBFGSMinimizer::init_minimization();
+	}
```
iglesias: Is there a strong reason to keep implementation of these methods in the header? Also for the ones below.

yorkerlin: No. I will correct that and add comments

PR #3177, opened 2016-04-23
updated the GPC notebook

yorkerlin:


yorkerlin:
@karlnapf 
The notebook is fine now.

PR #3179, opened 2016-04-25
fix an issue related to enum class for the precise build

yorkerlin:
@karlnapf 
The compiler in the precise build does not support `enum` class.


yorkerlin:
ref
http://buildbot.shogun-toolbox.org/builders/precise%20-%20libshogun/builds/1206/steps/compile/logs/stdio

PR #3184, opened 2016-05-11
Implement CParameterCombination::obtain_from_generic()

lkuchenb:
Iterating through parameter combinations in python as indicated in `modelselection_parameter_tree_modular.py` doesn't work since `get_element()` returns a `CSGObject` pointer.

This patch adds an `obtain_from_generic()` function to `CParamterCombination` that performs the type cast. The example was adopted.

### DIFF
--- a/src/shogun/modelselection/ParameterCombination.h
+++ b/src/shogun/modelselection/ParameterCombination.h
@@ -222,10 +222,16 @@ public:
 	static CParameterCombination* obtain_from_generic(
 			CSGObject* param_combination)
 	{
-		if (!param_combination)
+		if (param_combination)
+		{
+			CParameterCombination* casted = dynamic_cast<CParameterCombination*>(param_combination);
+			REQUIRE(casted, "CParameterCombination::obtain_from_generic(): Error, provided object"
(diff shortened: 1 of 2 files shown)


src/shogun/modelselection/ParameterCombination.h
```diff
@@ -215,6 +215,25 @@ friend class CModelSelectionParameters;
...
+	 *
+	 * @param param_combination its dynamic type must be CParameterCombination
+	 */
+	static CParameterCombination* obtain_from_generic(
+			CSGObject* param_combination)
+	{
+		if (param_combination)
+		{
+			CParameterCombination* casted = dynamic_cast<CParameterCombination*>(param_combination);
+			REQUIRE(casted, "CParameterCombination::obtain_from_generic(): Error, provided object"
```
karlnapf: Can remove the "CParameterCombination::obtain_from_generic(): ".
We add that automatically now

karlnapf:
Thanks for that! There is a minor thing. If you have a minute, it would be great if you can update. Otherwise I will merge soon

karlnapf:
Great thanks!

PR #3185, opened 2016-05-14
Fix buildbot mathematics/ warnings

OXPHOS:
Is this the correct way to do it? - 

Fixed documentation warnings in `mathematics/` in http://buildbot.shogun-toolbox.org/builders/deb3%20-%20modular_interfaces/builds/2853

### DIFF
--- a/src/shogun/mathematics/linalg/internal/implementation/MeanEigen3.h
+++ b/src/shogun/mathematics/linalg/internal/implementation/MeanEigen3.h
@@ -5,7 +5,6 @@
  *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions are met:
- *
  * 1. Redistributions of source code must retain the above copyright notice, this
  *    list of conditions and the following disclaimer.
  * 2. Redistributions in binary form must reproduce the above copyright notice,
@@ -44,17 +43,19 @@ namespace shogun
 
 namespace linalg
 {
-    
+
 namespace implementation
 {
-    
+
+#ifndef SWIG // SWIG should skip this part
 /**
- * @brief Generic class int2float which converts different types of integer 
+ * @brief Generic class int2float which converts different types of integer
  * into float64 type.
  */
 template <typename inputType>
 struct int2float
 {
+	/** orininal type **/
 	using floatType = inputType;
 };
 
@@ -64,6 +65,7 @@ struct int2float
 template <>
 struct int2float<int32_t>
 {
+	/** float64_t type **/
 	using floatType = float64_t;
 };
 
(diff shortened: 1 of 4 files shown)


src/shogun/mathematics/linalg/internal/implementation/MeanEigen3.h
```diff
@@ -55,6 +55,7 @@ namespace implementation
 template <typename inputType>
 struct int2float
 {
+	/** orininal type **/
```
karlnapf: this should just be swig/doxygen ignored

karlnapf:
Thats the way :)
great!

src/shogun/mathematics/linalg/internal/implementation/MeanEigen3.h
OXPHOS: Like:
```
#ifndef SWIG
      /*..*/
      template<..>
      struct int2float {..}
#endif
```

?

karlnapf: yep  `// SWIG should skip this part`

OXPHOS:
While it seems now the int2float structure is required : (
Maybe we can leave it there since mean calculation is going to be refactored anyway?

karlnapf:
None of this should be required / or exposed to SWIG. It is all internal.
It is refactored anyways, you are right. But should be fixed for release anyways. Should not be to hard to hide all this. But maybe if it not easy, don't spend too much time on it

OXPHOS:
@karlnapf May I ask why cookbook fails?

karlnapf:
No idea, this happened in another PR as well and @vigsterkr restarted it. But doesnt work here.

OXPHOS:
@karlnapf it worked : /

src/shogun/mathematics/linalg/internal/implementation/MeanEigen3.h
```diff
@@ -73,11 +75,24 @@ struct int2float<int32_t>
 template <>
 struct int2float<int64_t>
 {
+	/** float64_t type **/
 	using floatType = float64_t;
 };
- 
+#else
```
karlnapf: I am not sure I understand the logic of this preprocessor if then else.
Don't we just want to hide things from swig? Why the else case?

OXPHOS: @karlnapf aha sorry I missed this..Because int2float type is used in Redux.h and MeanEigen3.h several times as return type. If I hide int2float from SWIG, it's going to be undefined.

karlnapf: Does cause a build error?
The SWIG #ifdefs are not really used in compilation mode, but only for the swig wrapper (that should not need this type anyways)

OXPHOS:
@karlnapf You were right about SWIG..Should be ready now

karlnapf:
Thanks!

PR #3201, opened 2016-05-23
Fixed missing software-properties-common package error.

### LINKED ISSUE #3200: Docker missing package error.
Running Docker (configs/Dockerfile) crashes due to a missing software-properties-common package error.

arianepaola:
Fixed missing software-properties-common package error.

Fixes #3200 Docker missing package error.


karlnapf:
Doesn't hurt so I merge (even though I don't really see why something would crash otherwise, but well ;) )

PR #3203, opened 2016-05-25
Properly handle basic types in meta examples

karlnapf:


src/shogun/lib/common.h
```diff
@@ -56,7 +56,8 @@ typedef long double floatmax_t;
 #define STRING_LEN_STR             "256"
 typedef char                       string_t[STRING_LEN];
 
-typedef int                        machine_int_t;
+typedef int                        int32_t;
```
vigsterkr: ⭕

PR #3209, opened 2016-05-27
fixed a bug about memory leak

yorkerlin:
@vigsterkr @karlnapf


yorkerlin:
there is one issue. I will fix it in this weekend.

vigsterkr:
@yorkerlin i have a fix for this, coming already to develop, so can we close this?
which issue do you refer to?

PR #3210, opened 2016-05-27
parallel xval with svm

Saurabh7:
NOT merge ready

### DIFF
--- a/src/shogun/evaluation/CrossValidation.cpp
+++ b/src/shogun/evaluation/CrossValidation.cpp
@@ -7,7 +7,8 @@
  * Written (W) 2011-2012 Heiko Strathmann
  * Copyright (C) 2011 Berlin Institute of Technology and Max-Planck-Society
  */
-
+#include <shogun/machine/KernelMachine.h>
+#include <shogun/features/DenseFeatures.h>
 #include <shogun/evaluation/CrossValidation.h>
 #include <shogun/machine/Machine.h>
 #include <shogun/evaluation/Evaluation.h>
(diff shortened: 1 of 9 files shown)


Saurabh7:
This is specific to LibSVM * , since I am creating its different objects in each fold.

For working with say `Machine*` , I was thinking on the lines of making a virtual method `create_shallow_copy` to create Machine objects with similar parameters.

Similar as above for Features. Currently using a `create_shallow_copy` for `DenseFeatures<float64_t>`

src/shogun/evaluation/CrossValidation.cpp
```diff
@@ -264,8 +272,27 @@ float64_t CCrossValidation::evaluate_one_run()
 		m_machine->set_store_model_features(true);
 
 		/* do actual cross-validation */
+#pragma omp parallel for
```
karlnapf: I think you want to parallelise over both folds and runs

src/shogun/evaluation/CrossValidation.cpp
```diff
@@ -264,8 +272,27 @@ float64_t CCrossValidation::evaluate_one_run()
...
 		for (index_t i=0; i <num_subsets; ++i)
 		{
+			CMachine* machine;
```
karlnapf: OK
I suggest for now: `CSGObject::clone`, which gives a deep copy.
Try that and see how bad it is ;)

karlnapf: We copy the feature matrix this way, but that is OK for problems that are not too big. Let's see whether this works.
We can then start optimising further

PR #3183, opened 2016-05-05
Add kmeans page to cookbook

OXPHOS:
- There's no parameter `CLabels` in class `CKMeans`, so I can't find a way to use `apply_*`, `eval.evaluate`, so as to compare test and training dataset.
- Though I didn't see why `CKMeans` cannot have `CLabels` - we can just label the clusters 1..N.

- I thought about evaluating the clustering performance by calculating the Euclidean distances between centers of training dataset and test dataset, but there's no handy method for now.

- I didn't see the difference between dataset `fm_train_real.dat` and `classifier_binary__2d_linear_features_train.dat`but I think it doesn't really matter which one to use..?

PR #3217, opened 2016-05-28
renamed kmeans methods to be ruby friendly

karlnapf:
@OXPHOS The old kmeans methods for some reason were ruby unfriendly. I rename them here, after which the ruby example works for me locally. You have to rebase after merge and then adjust the kmeans method calls in your example #3183


karlnapf:
Plus some REQUIRE cleanups

PR #3222, opened 2016-05-28
fix meta example language to allow the "0" literal

karlnapf:
@OXPHOS this allows to use
`int variable = 0` in the meta language


OXPHOS:
Thanks!

PR #3224, opened 2016-05-29
parallel xval with clone

Saurabh7:
NOT merge ready

### DIFF
--- a/src/shogun/evaluation/CrossValidation.cpp
+++ b/src/shogun/evaluation/CrossValidation.cpp
@@ -264,11 +264,23 @@ float64_t CCrossValidation::evaluate_one_run()
 		m_machine->set_store_model_features(true);
 
 		/* do actual cross-validation */
+		#pragma omp parallel for
 		for (index_t i=0; i <num_subsets; ++i)
 		{
+			CMachine* machine;
+			CFeatures* features;
+			CLabels* labels;
+
+			if (get_global_parallel()->get_num_threads()==1)
+				machine=m_machine;
+			else
+				machine=(CMachine*)m_machine->clone();
+
 			/* evtl. update xvalidation output class */
 			CCrossValidationOutput* current=(CCrossValidationOutput*)
 					m_xval_outputs->get_first_element();
+			#pragma omp critical
+			{
 			while (current)
 			{
 				current->update_fold_index(i);
@@ -276,20 +288,25 @@ float64_t CCrossValidation::evaluate_one_run()
 				current=(CCrossValidationOutput*)
 						m_xval_outputs->get_next_element();
 			}
+			}
 
 			/* set feature subset for training */
 			SGVector<index_t> inverse_subset_indices=
 					m_splitting_strategy->generate_subset_inverse(i);
-			m_features->add_subset(inverse_subset_indices);
-			for (index_t p=0; p<m_features->get_num_preprocessors(); p++)
-			{
-				CPreprocessor* preprocessor = m_features->get_preprocessor(p);
-				preprocessor->init(m_features);
-				SG_UNREF(preprocessor);
-			}
+			
+			if (get_global_parallel()->get_num_threads()==1)
+				features=m_features;
+			else
+				features=(CFeatures*)m_features->clone();
+
+			features->add_subset(inverse_subset_indices);
 
 			/* set label subset for training */
-			m_labels->add_subset(inverse_subset_indices);
+			if (get_global_parallel()->get_num_threads()==1)
+				labels=m_labels;
+			else
+				labels=machine->get_labels();
+			labels->add_subset(inverse_subset_indices);
 
 			SG_DEBUG("training set %d:\n", i)
 			if (io->get_loglevel()==MSG_DEBUG)
@@ -300,30 +317,33 @@ float64_t CCrossValidation::evaluate_one_run()
 
 			/* train machine on training features and remove subset */
 			SG_DEBUG("starting training\n")
-			m_machine->train(m_features);
+			machine->train(features);
 			SG_DEBUG("finished training\n")
 
 			/* evtl. update xvalidation output class */
+			#pragma omp critical
+			{
 			current=(CCrossValidationOutput*)m_xval_outputs->get_first_element();
 			while (current)
 			{
 				current->update_train_indices(inverse_subset_indices, "\t");
-				current->update_trained_machine(m_machine, "\t");
+				current->update_trained_machine(machine, "\t");
 				SG_UNREF(current);
 				current=(CCrossValidationOutput*)
 						m_xval_outputs->get_next_element();
 			}
+			}
 
-			m_features->remove_subset();
-			m_labels->remove_subset();
+			features->remove_subset();
+			labels->remove_subset();
 
 			/* set feature subset for testing (subset method that stores pointer) */
 			SGVector<index_t> subset_indices =
 					m_splitting_strategy->generate_subset_indices(i);
-			m_features->add_subset(subset_indices);
+			features->add_subset(subset_indices);
 
 			/* set label subset for testing */
-			m_labels->add_subset(subset_indices);
+			labels->add_subset(subset_indices);
 
 			SG_DEBUG("test set %d:\n", i)
 			if (io->get_loglevel()==MSG_DEBUG)
@@ -334,33 +354,42 @@ float64_t CCrossValidation::evaluate_one_run()
 
 			/* apply machine to test features and remove subset */
 			SG_DEBUG("starting evaluation\n")
-			SG_DEBUG("%p\n", m_features)
-			CLabels* result_labels=m_machine->apply(m_features);
+			SG_DEBUG("%p\n", features)
+			CLabels* result_labels=machine->apply(features);
 			SG_DEBUG("finished evaluation\n")
-			m_features->remove_subset();
+			features->remove_subset();
 			SG_REF(result_labels);
 
 			/* evaluate */
-			results[i]=m_evaluation_criterion->evaluate(result_labels, m_labels);
+			results[i]=m_evaluation_criterion->evaluate(result_labels, labels);
 			SG_DEBUG("result on fold %d is %f\n", i, results[i])
 
 			/* evtl. update xvalidation output class */
+			#pragma omp critical
+			{
 			current=(CCrossValidationOutput*)m_xval_outputs->get_first_element();
 			while (current)
 			{
 				current->update_test_indices(subset_indices, "\t");
 				current->update_test_result(result_labels, "\t");
-				current->update_test_true_result(m_labels, "\t");
+				current->update_test_true_result(labels, "\t");
 				current->post_update_results();
 				current->update_evaluation_result(results[i], "\t");
 				SG_UNREF(current);
 				current=(CCrossValidationOutput*)
 						m_xval_outputs->get_next_element();
 			}
+			}
 
 			/* clean up, remove subsets */
+			labels->remove_subset();
+			if (get_global_parallel()->get_num_threads()!=1)
+			{
+				SG_UNREF(machine);
+				SG_UNREF(features);
+				SG_UNREF(labels);
+			}
 			SG_UNREF(result_labels);
-			m_labels->remove_subset();
 		}
 
 		SG_DEBUG("done unlocked evaluation\n", get_name())
--- a/tests/unit/evaluation/CrossValidation_multithread_unittest.cc
+++ b/tests/unit/evaluation/CrossValidation_multithread_unittest.cc
@@ -1,0 +1,168 @@
+/*
+ * Copyright (c) The Shogun Machine Learning Toolbox
+ * Written (w) 2016 Saurabh Mahindre
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * The views and conclusions contained in the software and documentation are those
+ * of the authors and should not be interpreted as representing official policies,
+ * either expressed or implied, of the Shogun Development Team.
+ *
+ */
+
+#include <shogun/base/init.h>
+#include <shogun/features/DenseFeatures.h>
+#include <shogun/labels/BinaryLabels.h>
+#include <shogun/kernel/GaussianKernel.h>
+#include <shogun/classifier/svm/LibSVM.h>
+#include <shogun/evaluation/CrossValidation.h>
+#include <shogun/evaluation/StratifiedCrossValidationSplitting.h>
+#include <shogun/evaluation/ContingencyTableEvaluation.h>
+#include <gtest/gtest.h>
+#include <shogun/labels/MulticlassLabels.h>
+#include <shogun/multiclass/KNN.h>
+#include <shogun/evaluation/MulticlassAccuracy.h>
+#include <shogun/distance/EuclideanDistance.h>
+
+using namespace shogun;
+
+void generate_data(SGMatrix<float64_t>& mat, SGVector<float64_t> &lab)
+{
+	int32_t num=lab.size();
+
+	for (index_t i=0; i<num; i++)
+	{
+		mat(0,i)=i<num/2 ? 0+(CMath::randn_double()*4) : 100+(CMath::randn_double()*4)	;
+		mat(1,i)=i;
+	}
+
+	for (index_t i=0; i<num; ++i)
+		lab.vector[i]=i<num/2 ? 0 : 1;
+
+}
+
+TEST(CrossValidation_multithread, LibSVM_unlocked)
+{
+	int32_t num=100;
+	SGMatrix<float64_t> mat(2, num);
+
+	/* training labels +/- 1 for each cluster */
+	SGVector<float64_t> lab(num);
+
+	/*create simple linearly separable data*/
+	generate_data(mat, lab);
+
+	for (index_t i=0; i<num/2; ++i)
+		lab.vector[i]-=1;
+	CBinaryLabels* labels=new CBinaryLabels(lab);
+
+	CDenseFeatures<float64_t>* features=
+			new CDenseFeatures<float64_t>(mat);
+	SG_REF(features);	
+
+	int32_t width=100;
+	CGaussianKernel* kernel=new CGaussianKernel(width);
+	kernel->init(features, features);
+
+	/* create svm via libsvm */
+	float64_t svm_C=1;
+	CLibSVM* svm=new CLibSVM(svm_C, kernel, labels);
+
+	CContingencyTableEvaluation* eval_crit=
+			new CContingencyTableEvaluation(ACCURACY);
+
+	index_t n_folds=4;
+	CStratifiedCrossValidationSplitting* splitting=
+			new CStratifiedCrossValidationSplitting(labels, n_folds);
+
+	CCrossValidation* cross=new CCrossValidation(svm, features, labels,
+			splitting, eval_crit);
+
+	cross->set_autolock(false);
+	cross->set_num_runs(4);
+	cross->parallel->set_num_threads(1);	
+
+	CCrossValidationResult* result1=(CCrossValidationResult*)cross->evaluate();
+	float64_t mean1 = result1->mean;
+	
+	cross->parallel->set_num_threads(3);
+
+	CCrossValidationResult* result2=(CCrossValidationResult*)cross->evaluate();
+	float64_t mean2 = result2->mean;
+
+	EXPECT_EQ(mean1, mean2);
+
+	/* clean up */
+	SG_UNREF(result1);
+	SG_UNREF(result2);
+	SG_UNREF(cross);
+	SG_UNREF(features);
+}
+
+TEST(CrossValidation_multithread, KNN)
+{
+	int32_t num=100;
+	SGMatrix<float64_t> mat(2, num);
+
+	SGVector<float64_t> lab(num);
+
+	/*create simple linearly separable data*/
+	generate_data(mat, lab);
+	CMulticlassLabels* labels=new CMulticlassLabels(lab);
+
+	CDenseFeatures<float64_t>* features=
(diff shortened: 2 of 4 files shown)


Saurabh7:
I tried this for svm, krr, multiclassliblinear, KNN.

Saurabh7:
a simple `#pragma omp parallel for` around runs too is segfaulting for now

src/shogun/evaluation/CrossValidation.cpp
```diff
@@ -86,8 +86,9 @@ CEvaluationResult* CCrossValidation::evaluate()
 
 	if (m_autolock)
 	{
+		int32_t flag=0;
 		/* if machine supports locking try to do so */
-		if (m_machine->supports_locking())
+		if (m_machine->supports_locking() && flag)
```
karlnapf: what does this flag do?
In general, use variable names that are self explaining :)

src/shogun/evaluation/CrossValidation.cpp
```diff
@@ -280,16 +286,22 @@ float64_t CCrossValidation::evaluate_one_run()
 			/* set feature subset for training */
 			SGVector<index_t> inverse_subset_indices=
 					m_splitting_strategy->generate_subset_inverse(i);
-			m_features->add_subset(inverse_subset_indices);
+
+			CFeatures* features=CFeatures::create_shallow_copy(m_features);
```
karlnapf: keep in mind that cloning the machine above also will clone the features (including the matrix), if the features are set.
I would make use of that and not separately clone the features here. I get it that you want to be efficient here, but let's postpone that for a bit and get the other things right first.

src/shogun/evaluation/CrossValidation.cpp
```diff
@@ -280,16 +286,22 @@ float64_t CCrossValidation::evaluate_one_run()
...
 			{
 				CPreprocessor* preprocessor = m_features->get_preprocessor(p);
-				preprocessor->init(m_features);
+				preprocessor->init(features);
 				SG_UNREF(preprocessor);
 			}
 
 			/* set label subset for training */
-			m_labels->add_subset(inverse_subset_indices);
+			CLabels* labels=(CLabels*)m_labels->clone();
```
karlnapf: same as with features, use the cloned machine

src/shogun/evaluation/CrossValidation.cpp
```diff
@@ -264,8 +266,12 @@ float64_t CCrossValidation::evaluate_one_run()
...
 
 		/* do actual cross-validation */
+#pragma omp parallel for
 		for (index_t i=0; i <num_subsets; ++i)
 		{
+			CMachine* machine=(CMachine*)m_machine->clone();
+			SG_REF(machine);
+			machine->set_store_model_features(true);
 			/* evtl. update xvalidation output class */
 			CCrossValidationOutput* current=(CCrossValidationOutput*)
```
karlnapf: keep in mind that your x-validation output currently is shared across the threads.
That is fine actually (updating it is fast), but you might want to do the update of it atomic so that the different threads dont conflict here

src/shogun/evaluation/CrossValidation.cpp
```diff
@@ -280,16 +286,22 @@ float64_t CCrossValidation::evaluate_one_run()
...
-			m_features->add_subset(inverse_subset_indices);
+
+			CFeatures* features=CFeatures::create_shallow_copy(m_features);
+			SG_REF(features);
+			features->add_subset(inverse_subset_indices);
 			for (index_t p=0; p<m_features->get_num_preprocessors(); p++)
 			{
 				CPreprocessor* preprocessor = m_features->get_preprocessor(p);
-				preprocessor->init(m_features);
+				preprocessor->init(features);
```
karlnapf: I think this should be done on the cloned features

karlnapf: Actually, I think this is wrong anyways --- for now remove all the preprocessors from the xvalidation code, we can do that later

src/shogun/evaluation/CrossValidation.cpp
```diff
@@ -280,16 +286,22 @@ float64_t CCrossValidation::evaluate_one_run()
 			/* set feature subset for training */
```
karlnapf: is this call thread safe? Should be, but we better make sure it is

src/shogun/features/Features.cpp
```diff
@@ -357,4 +358,73 @@ bool CFeatures::get_feature_class_compatibility(EFeatureClass rhs) const
 	if (this->get_feature_class()==rhs)
 		return true;
 	return false;
+} 
+
+CFeatures* CFeatures::create_shallow_copy(CFeatures* other)
+{
+	SG_SDEBUG("Entering!\n");
```
karlnapf: let's avoid this one for now (save it for later)

src/shogun/features/Features.cpp
```diff
@@ -357,4 +358,73 @@ bool CFeatures::get_feature_class_compatibility(EFeatureClass rhs) const
...
+	}
+	else
+		SG_SNOTIMPLEMENTED;
+	SG_SDEBUG("Leaving!\n");
+	return shallow_copy;
+}
+
+void CFeatures::clone_subset_stack(CFeatures* src, CFeatures* dst)
+{
+	SG_SDEBUG("Entering!\n");
```
karlnapf: this should be covered by "clone"

karlnapf:
Cool.

What about a unit test that ensures that a single thread version produces the same as a multithreaded version and there are no memory errors?

src/shogun/evaluation/CrossValidation.cpp
Saurabh7: Yes I had tried to do this initially but no there is no `get_features` or `set_features` in `CMachine`

src/shogun/evaluation/CrossValidation.cpp
Saurabh7: Just using this to skip locked training temporarily and test the unlocked version.

src/shogun/evaluation/CrossValidation.cpp
karlnapf: I see. The problem with cloning the features by hand is that *if* they had already been assigned to the machine in some way (via kernel e.g.), then they *will* already have been cloned. But OK, let's postpone that as well.
Can you add all these things to a TODO list so that we wont forget?

karlnapf: Definitely no shallow copy here for now

Saurabh7:
Question: Do we always wants this to go for clone and parallel or make it user defined ?
e.g when I test parallel+clone vs normal output ?

karlnapf:
Single core computation should not clone anything, i.e. set `machine=m_machine`

karlnapf:
You can ask Shogun for the number of cores used via `CParallel`

I think currently this mechanism kind of ignores openmp
@lambday can you send a fix for develop?

lambday:
@karlnapf sure. I'll do that tomorrow. Gotta add a HAVE_OPENMP check in the cmake for that.

src/shogun/machine/KernelMachine.cpp
```diff
@@ -486,7 +486,10 @@ bool CKernelMachine::train_locked(SGVector<index_t> indices)
 		SG_ERROR("CKernelMachine::train_locked() call data_lock() before!\n")
 
 	/* this is asusmed here */
-	ASSERT(m_custom_kernel==kernel)
+	//ASSERT(m_custom_kernel==kernel)
+	SG_UNREF(kernel);
+	kernel=m_custom_kernel;
```
Saurabh7: clone creates different copy for each.
And since clone happens after data lock, the assert fails. Had to do this.

karlnapf: Let us focus on the non-locked case for now.

src/shogun/evaluation/CrossValidation.cpp
```diff
@@ -122,6 +122,7 @@ CEvaluationResult* CCrossValidation::evaluate()
 
 	/* perform all the x-val runs */
 	SG_DEBUG("starting %d runs of cross-validation\n", m_num_runs)
+	#pragma omp parallel for
```
karlnapf: lets only parallelise the folds for now. This comes back later

src/shogun/evaluation/CrossValidation.cpp
```diff
@@ -264,32 +264,60 @@ float64_t CCrossValidation::evaluate_one_run()
...
+
+			if (get_global_parallel()->get_num_threads()==1)
+			{
+				machine=m_machine;
+				SG_REF(machine);
+			}
+			else
+			{
+				machine=(CMachine*)m_machine->clone();
+				machine->set_store_model_features(true);
```
karlnapf: wouldnt this also be applied if not threads are active?

src/shogun/evaluation/CrossValidation.cpp
```diff
@@ -264,32 +264,60 @@ float64_t CCrossValidation::evaluate_one_run()
...
 		for (index_t i=0; i <num_subsets; ++i)
 		{
+			CMachine* machine;
+			CFeatures* features;
+			CLabels* labels;
+
+			if (get_global_parallel()->get_num_threads()==1)
+			{
+				machine=m_machine;
+				SG_REF(machine);
```
karlnapf: why SG_REF here and not below?

src/shogun/evaluation/CrossValidation.cpp
```diff
@@ -334,33 +365,39 @@ float64_t CCrossValidation::evaluate_one_run()
...
 				current->update_evaluation_result(results[i], "\t");
 				SG_UNREF(current);
 				current=(CCrossValidationOutput*)
 						m_xval_outputs->get_next_element();
 			}
+			}
 
 			/* clean up, remove subsets */
+			labels->remove_subset();		
+			SG_UNREF(machine);
```
karlnapf: I suggest you only unref if num threads >1

karlnapf: Then you can avoid SG_REFing the m_machine (which doesnt really make sense)
And then instead, you can ref the output of the clones

tests/unit/evaluation/CrossValidation_multithread_unittest.cc
```diff
@@ -0,0 +1,107 @@
...
+#include <shogun/evaluation/StratifiedCrossValidationSplitting.h>
+#include <shogun/evaluation/ContingencyTableEvaluation.h>
+#include <gtest/gtest.h>
+
+using namespace shogun;
+
+TEST(CrossValidation_multithread, LibSVM_unlocked)
+{
+	int32_t num=100;
+	/*create simple linearly separable data*/
```
karlnapf: can you put this into an external function of the unit test?
There is a setup mechanism for that

tests/unit/evaluation/CrossValidation_multithread_unittest.cc
```diff
@@ -0,0 +1,107 @@
...
+	/* evaluation criterion */
+	CContingencyTableEvaluation* eval_crit=
+			new CContingencyTableEvaluation(ACCURACY);
+
+	/* splitting strategy */
+	index_t n_folds=10;
+	CStratifiedCrossValidationSplitting* splitting=
+			new CStratifiedCrossValidationSplitting(labels, n_folds);
+
+	/* cross validation instance, 10 runs, 95% confidence interval */
```
karlnapf: no confidence intervals, that is copy paste waste

tests/unit/evaluation/CrossValidation_multithread_unittest.cc
```diff
@@ -0,0 +1,107 @@
...
+	cross->parallel->set_num_threads(cross->parallel->get_num_cpus());
+
+	result=(CCrossValidationResult*)cross->evaluate();
+	float64_t mean2 = result->mean;
+
+	EXPECT_EQ(mean1, mean2);
+
+	/* clean up */
+	SG_UNREF(result);
+	SG_UNREF(cross);
```
karlnapf: cool!

karlnapf:
What does valgrind say to this? unit test?

tests/unit/evaluation/CrossValidation_multithread_unittest.cc
```diff
@@ -0,0 +1,107 @@
...
+	index_t n_folds=10;
+	CStratifiedCrossValidationSplitting* splitting=
+			new CStratifiedCrossValidationSplitting(labels, n_folds);
+
+	/* cross validation instance, 10 runs, 95% confidence interval */
+	CCrossValidation* cross=new CCrossValidation(svm, features, labels,
+			splitting, eval_crit);
+
+	cross->set_autolock(false);
+	cross->set_num_runs(5);
```
karlnapf: make it 3 runs to save time

tests/unit/evaluation/CrossValidation_multithread_unittest.cc
```diff
@@ -0,0 +1,107 @@
...
+	float64_t svm_eps=0.0001;
+	CLibSVM* svm=new CLibSVM(svm_C, kernel, labels);
+	svm->set_epsilon(svm_eps);
+
+	/* evaluation criterion */
+	CContingencyTableEvaluation* eval_crit=
+			new CContingencyTableEvaluation(ACCURACY);
+
+	/* splitting strategy */
+	index_t n_folds=10;
```
karlnapf: make 3 folds to save time

karlnapf:
can you squash the commits after the rebase?

src/shogun/evaluation/CrossValidation.cpp
```diff
@@ -264,32 +264,57 @@ float64_t CCrossValidation::evaluate_one_run()
...
 		/* do actual cross-validation */
+		#pragma omp parallel for if (get_global_parallel()->get_num_threads()!=1)
 		for (index_t i=0; i <num_subsets; ++i)
 		{
+			CMachine* machine;
+			CFeatures* features;
+			CLabels* labels;
+
+			if (get_global_parallel()->get_num_threads()==1)
+			{
```
karlnapf: no {} for single instructions please

src/shogun/evaluation/CrossValidation.cpp
```diff
@@ -264,32 +264,57 @@ float64_t CCrossValidation::evaluate_one_run()
...
+				machine=m_machine;
+			}
+			else
+			{
+				machine=(CMachine*)m_machine->clone();
```
karlnapf: You probably want to REF them here

src/shogun/evaluation/CrossValidation.cpp
```diff
@@ -264,32 +264,57 @@ float64_t CCrossValidation::evaluate_one_run()
...
+			
+			if (get_global_parallel()->get_num_threads()==1)
 			{
-				CPreprocessor* preprocessor = m_features->get_preprocessor(p);
-				preprocessor->init(m_features);
-				SG_UNREF(preprocessor);
+				features=m_features;
 			}
+			else
+				features=(CFeatures*)m_features->clone();
```
karlnapf: and REF

src/shogun/evaluation/CrossValidation.cpp
```diff
@@ -264,32 +264,57 @@ float64_t CCrossValidation::evaluate_one_run()
...
+			
+			features->add_subset(inverse_subset_indices);
 
 			/* set label subset for training */
-			m_labels->add_subset(inverse_subset_indices);
+			if (get_global_parallel()->get_num_threads()==1)
+			{
```
karlnapf: no {}

src/shogun/evaluation/CrossValidation.cpp
```diff
@@ -264,32 +264,57 @@ float64_t CCrossValidation::evaluate_one_run()
...
+				labels=m_labels;
+			}
+			else
+				labels=machine->get_labels();
```
karlnapf: wrong tab indent. This automatically REFs, no need to decrease refcounter at the end

src/shogun/evaluation/CrossValidation.cpp
```diff
@@ -334,33 +362,42 @@ float64_t CCrossValidation::evaluate_one_run()
...
 				current=(CCrossValidationOutput*)
 						m_xval_outputs->get_next_element();
 			}
+			}
 
 			/* clean up, remove subsets */
+			labels->remove_subset();
+			if (get_global_parallel()->get_num_threads()!=1)
+			{
+				SG_UNREF(machine);
```
karlnapf: you are playing with fire here as you did not ref things when they were cloned (only the labels) ;)

tests/unit/evaluation/CrossValidation_multithread_unittest.cc
```diff
@@ -0,0 +1,119 @@
...
+#include <shogun/evaluation/StratifiedCrossValidationSplitting.h>
+#include <shogun/evaluation/ContingencyTableEvaluation.h>
+#include <gtest/gtest.h>
+
+using namespace shogun;
+
+void generate_data(SGMatrix<float64_t>& mat, SGVector<float64_t> &lab)
+{
+	int32_t num=lab.size();
+	SG_SPRINT("%d", num);
```
karlnapf: please no prints in tests

tests/unit/evaluation/CrossValidation_multithread_unittest.cc
```diff
@@ -0,0 +1,119 @@
...
+
+	for (index_t i=0; i<num; i++)
+	{
+		mat(0,i)=i<num/2 ? 0 : 100;
```
karlnapf: can you make this noisy? Most methods behave themselves better for gaussian like input

tests/unit/evaluation/CrossValidation_multithread_unittest.cc
```diff
@@ -0,0 +1,119 @@
...
+TEST(CrossValidation_multithread, LibSVM_unlocked)
+{
+	int32_t num=100;
+	SGMatrix<float64_t> mat(2, num);
+
+	/* training labels +/- 1 for each cluster */
+	SGVector<float64_t> lab(num);
+
+	/*create simple linearly separable data*/
+	generate_data(mat, lab);
```
karlnapf: this method could return a std::pair<CFeatures*, CLabels*> with newly allocated data, so this code gets slimmer

tests/unit/evaluation/CrossValidation_multithread_unittest.cc
```diff
@@ -0,0 +1,119 @@
...
+
+	CBinaryLabels* labels=new CBinaryLabels(lab);
+
+	CDenseFeatures<float64_t>* features=
+			new CDenseFeatures<float64_t>(mat);
+	SG_REF(features);	
+
+	int32_t kernel_cache=100;
```
karlnapf: use the new cache less constructor #3237

tests/unit/evaluation/CrossValidation_multithread_unittest.cc
```diff
@@ -0,0 +1,119 @@
...
+	int32_t width=10;
+	CGaussianKernel* kernel=new CGaussianKernel(kernel_cache, width);
+	kernel->init(features, features);
+
+	/* create svm via libsvm */
+	float64_t svm_C=10;
+	float64_t svm_eps=0.0001;
+	CLibSVM* svm=new CLibSVM(svm_C, kernel, labels);
+	svm->set_epsilon(svm_eps);
```
karlnapf: no need to set epsilon here

tests/unit/evaluation/CrossValidation_multithread_unittest.cc
```diff
@@ -0,0 +1,119 @@
...
+			new CDenseFeatures<float64_t>(mat);
+	SG_REF(features);	
+
+	int32_t kernel_cache=100;
+	int32_t width=10;
+	CGaussianKernel* kernel=new CGaussianKernel(kernel_cache, width);
+	kernel->init(features, features);
+
+	/* create svm via libsvm */
+	float64_t svm_C=10;
```
karlnapf: set C to 1

tests/unit/evaluation/CrossValidation_multithread_unittest.cc
```diff
@@ -0,0 +1,119 @@
...
+	CCrossValidationResult* result1=(CCrossValidationResult*)cross->evaluate();
+	float64_t mean1 = result1->mean;
+	
+	cross->parallel->set_num_threads(cross->parallel->get_num_cpus());
+
+	CCrossValidationResult* result2=(CCrossValidationResult*)cross->evaluate();
+	float64_t mean2 = result2->mean;
+
+	EXPECT_EQ(mean1, mean2);
+
```
karlnapf: so this is valgrind clean?

Saurabh7: Yes it clean. Hence wondering about the Ref comments.

src/shogun/evaluation/CrossValidation.cpp
Saurabh7: Wondering why this worked.
https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/base/SGObject.cpp#L721
clone REFs the copy. So no REF needed ?

karlnapf: You *do* need them. Just because it worked doesnt mean it will always work. Basically, behaviour is undefined if the thing is not refed and you pass it around. Example

```
bar=foo->clone(); // ref-counter is 0
method_call(bar) // assume this method increases the counter once at the beginning, and then decreases afterwards -- it becomes 0 and the object is deleted
bar->method() // crash
```

tests/unit/evaluation/CrossValidation_multithread_unittest.cc
karlnapf: Cool -- if it is clean. Then please polish the PR and we can go on.

Before I merge, Can you put some benchmark numbers?

Saurabh7:
abalone7

```
Multiclassliblinear
Data size: (2785 11) , Single Threaded: 3155 , MultiThreaded: 973

KNN
Data size: (2785 11) , Single Threaded: 2489 , MultiThreaded: 2089
```

karlnapf:
The benchmark should go to ryancurtins framework btw

src/shogun/evaluation/CrossValidation.cpp
```diff
@@ -264,32 +264,55 @@ float64_t CCrossValidation::evaluate_one_run()
...
+			CFeatures* features;
+			CLabels* labels;
+
+			if (get_global_parallel()->get_num_threads()==1)
+				machine=m_machine;
+			else
+			{
+				machine=(CMachine*)m_machine->clone();
+				SG_REF(machine);
+				machine->set_store_model_features(true);
```
karlnapf: sorry, can you explain why this is necessary after cloning but not without cloning? machine should be exactly the same as m_machine here....or am I getting this wrong?

karlnapf:
We can merge as soon as you address the comments on the unit tests.
Also, if you write more benchmarks, put them in as unit tests as well.
Let's get this merged by tomorrow early ok?

karlnapf:
The times are interesting, I guess KNN is already multicore that is why doing multiple of them doesnt help?

Saurabh7:
Dont think KNN is multicore.

libSVM
`Data size: 1000 , Single Threaded: 40792 , MultiThreaded: 25174`
Ok will add this to framework.

src/shogun/evaluation/CrossValidation.cpp
Saurabh7: Changed this btw. Thanks for explaining.

src/shogun/evaluation/CrossValidation.cpp
Saurabh7: I think you are right, since m_store_model_features is a param, it should be same for clone. Removed.

Saurabh7:
Ok one other thing I have to mention here `cross->parallel->get_num_cpus()` returns 1 for me always. 

So   `set_num_threads(cross->parallel->get_num_cpus())` wasnt actually running in parallel here. So for the numbers I used `set_num_threads(3).`

tests/unit/evaluation/CrossValidation_multithread_unittest.cc
```diff
@@ -0,0 +1,168 @@
...
+			splitting, eval_crit);
+
+	cross->set_autolock(false);
+	cross->set_num_runs(4);
+	cross->parallel->set_num_threads(1);	
+
+	CCrossValidationResult* result1=(CCrossValidationResult*)cross->evaluate();
+	float64_t mean1 = result1->mean;
+	
+	cross->parallel->set_num_threads(cross->parallel->get_num_cpus());
```
karlnapf: set to 3 manually here

karlnapf:
As discussed
 * unit test should manually set number of threads to 3 (and still be memory clean with valgrind)
 * update @brief of the x-validation class: say that it runs with the current number of threads, and that it duplicates all objects (might be changed later)
 * update NEWS
 * squash commits

Saurabh7:
updated :)

karlnapf:
Great!
This is good to go once travis is green

We will re-visit for
 * parallel across runs
 * sharing feature matrices (needs some const cleanups)

src/shogun/evaluation/CrossValidation.cpp
lisitsyn: Is this comment ^ already addressed somehow?

Saurabh7: Unit tests are valgrind clean here.

karlnapf:
Ok. Great!
This is a nice improvement.....we should get back to the shared features at some (later) point

PR #3225, opened 2016-05-29
test lda meta example on travis

karlnapf:


karlnapf:
This works locally (with java enabled). Testing whether travis fails, as in #3219

karlnapf:
travis worked

PR #3226, opened 2016-05-30
Cereal serialization - SGVector

OXPHOS:
I cannot directly add methods to SGVector because `Math.h` calls `SGVector.load()` during compiling and somehow it confuses cereal `load` and the old `load`, and leads to error.

In `CerealTest.h`, **only line 467 - 483** are cereal serialization. others are just copy-paste and name refactor from `SGVector.h`.
`CerealTest.cpp` is just the name refactor version of `SGVector.cpp`.
`Cereal_unittests.cc` is the save/load test case.

@vigsterkr

### DIFF
--- a/examples/meta/CMakeLists.txt
+++ b/examples/meta/CMakeLists.txt
@@ -8,6 +8,7 @@
 # execute_process(COMMAND sh ${CMAKE_CURRENT_SOURCE_DIR}/generator/types/get_type_list.sh > ${CMAKE_CURRENT_SOURCE_DIR}/generator/types/typelist)
 
 file(MAKE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/parser_files)
+include_directories(${CEREAL_INCLUDE_DIRS})
 
 # Run example generation script
 add_custom_target(meta_examples
--- a/src/shogun/lib/SGVector.h
+++ b/src/shogun/lib/SGVector.h
@@ -474,6 +474,28 @@ template<class T> class SGVector : public SGReferencedData
 		 * @return matrix
 		 */
 		static void convert_to_matrix(T*& matrix, index_t nrows, index_t ncols, const T* vector, int32_t vlen, bool fortran_order);
+
+		template<class Archive>
+		void cereal_save(Archive & ar) const
+		{
+			ar(cereal::base_class<SGReferencedData>(this));
+			ar(cereal::make_nvp("length", vlen));
+			for (index_t i = 0; i < vlen; ++i)
+				ar(vector[i]);
+		}
+
+		template<class Archive>
+		void cereal_load(Archive & ar)
+		{
+			unref();
+			ar(cereal::base_class<SGReferencedData>(this));
+
+			ar(vlen);
+			vector = SG_MALLOC(T, vlen);
+			for (index_t i = 0; i < vlen; ++i)
+				ar(vector[i]);
+		}
+
 #endif // #ifndef SWIG // SWIG should skip this part
 	protected:
 		/** needs to be overridden to copy data */
--- a/tests/unit/io/Cereal_unittest.cc
+++ b/tests/unit/io/Cereal_unittest.cc
@@ -1,0 +1,79 @@
+#include <shogun/lib/common.h>
+#include <gtest/gtest.h>
+
+#include <shogun/lib/SGVector.h>
+#include <cereal/archives/json.hpp>
+
+#include <shogun/io/SGIO.h>
+#include <cstring>
+#include <fstream>
+
+using namespace shogun;
+
+TEST(Cereal, Json_SGVector_FLOAT64_load_equals_saved)
+{
+	const index_t size = 5;
+	SGVector<float64_t> a(size);
+	SGVector<float64_t> b;
+	a.range_fill(1.0);
+
+	std::string filename = std::tmpnam(nullptr);
+
+	try
+	{
+		{
+			std::ofstream os(filename.c_str());
+			cereal::JSONOutputArchive archive(os);
+			archive(a);
+		}
+
+		{
+			std::ifstream is(filename.c_str());
+			cereal::JSONInputArchive archive(is);
+			archive(b);
+		}
+	}
+	catch (std::exception& e)
+		SG_SINFO("Error code: %s \n", e.what());
+
+	EXPECT_EQ(a.size(), b.size());
+	EXPECT_EQ(a.ref_count(), b.ref_count());
+	for (index_t i = 0; i < size; i++)
+		EXPECT_NEAR(a[i], b[i], 1E-15);
+
+	remove(filename.c_str());
+}
+
+TEST(Cereal, Json_SGVector_load_equals_saved_refcounting_false)
+{
+	const index_t size = 5;
+	SGVector<float64_t> a(size, false);
+	SGVector<float64_t> b;
+	a.range_fill(1.0);
+
+	std::string filename = std::tmpnam(nullptr);
+
+	try
+	{
+		{
+			std::ofstream os(filename.c_str());
+			cereal::JSONOutputArchive archive(os);
+			archive(a);
+		}
+
+		{
+			std::ifstream is(filename.c_str());
+			cereal::JSONInputArchive archive(is);
+			archive(b);
+		}
+	}
+	catch (std::exception& e)
+		SG_SINFO("Error code: %s \n", e.what());
+
+	EXPECT_EQ(a.size(), b.size());
+	EXPECT_EQ(-1, b.ref_count());
+	for (index_t i = 0; i < size; i++)
+		EXPECT_NEAR(a[i], b[i], 1E-15);
+
+	remove(filename.c_str());
+}
--- a/src/shogun/lib/SGReferencedData.h
+++ b/src/shogun/lib/SGReferencedData.h
@@ -10,12 +10,11 @@
 #define __SGREFERENCED_DATA_H__
 
 #include <shogun/lib/config.h>
-
 #include <shogun/lib/common.h>
+#include <shogun/lib/RefCount.h>
 
 namespace shogun
 {
-class RefCount;
 
 /** @brief shogun reference count managed data */
 class SGReferencedData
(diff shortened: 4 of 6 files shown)


src/shogun/lib/CerealTest.cpp
```diff
@@ -0,0 +1,958 @@
+/*
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
```
karlnapf: For later:
Use BSD license and your own name
See https://gist.github.com/karlnapf/24763634d5142f92f5213d282cf67e0f

src/shogun/lib/CerealTest.h
```diff
@@ -0,0 +1,509 @@
...
+		 *	false - row-major order (C, Python)
+		 * @return matrix
+		 */
+		static void convert_to_matrix(T*& matrix, index_t nrows, index_t ncols, const T* vector, int32_t vlen, bool fortran_order);
+#endif // #ifndef SWIG // SWIG should skip this part
+
+		template<class Archive>
+		void save(Archive & ar) const
+		{
+			ar(vlen);
```
karlnapf: this seems OK

OXPHOS:
@vigsterkr Hey I just cleaned up this SGVector-cereal prototype. Wanna have a look at it? I guess I'll have to generalize it as next step.

OXPHOS:
Error message when add cereal save/load methods in SGVector.h:

https://gist.github.com/OXPHOS/b9c363c2de6129029428e270f3bc815c

OXPHOS:
The name reload works so easily..

OXPHOS:
@vigsterkr  updated

CMakeLists.txt
```diff
@@ -590,6 +590,14 @@ IF (LAPACK_FOUND AND NOT (DARWIN AND RModular))
 	ENDIF()
 ENDIF()
 
+FIND_PACKAGE(Cereal)
+IF(NOT CEREAL_FOUND)
```
karlnapf: Can we have a separate patch for the cmake changes ?

src/shogun/lib/SGVector.h
```diff
@@ -474,6 +477,25 @@ template<class T> class SGVector : public SGReferencedData
 		 * @return matrix
 		 */
 		static void convert_to_matrix(T*& matrix, index_t nrows, index_t ncols, const T* vector, int32_t vlen, bool fortran_order);
+
+		template<class Archive>
```
karlnapf: This seems ok but what we really need is a patch that serialises the parameter map of the tag framework....thoughts?

src/shogun/lib/SGVector.h
```diff
@@ -474,6 +477,25 @@ template<class T> class SGVector : public SGReferencedData
...
+		void cereal_save(Archive & ar) const
+		{
+			ar(vlen);
+			for (index_t i = 0; i < vlen; ++i)
+				ar(vector[i]);
+		}
+
+		template<class Archive>
+		void cereal_load(Archive & ar)
```
karlnapf: GPU matrix?

src/shogun/lib/SGVector.h
```diff
@@ -474,6 +477,25 @@ template<class T> class SGVector : public SGReferencedData
...
+		{
+			unref();
```
karlnapf: We need to resolve the refcounting thing here before merging
@vigsterkr ?

karlnapf:
I am OK mostly here (since it is feature branch)

src/shogun/lib/SGVector.h
OXPHOS: With discussions in #3375.
I think we need the unref() - because we allocated new space for loaded data and the old ones are gone.
For the same reason I think we need a new RefCount.
However, I cannot do it now without appropriate reset method or serialization method in SGReferenceData.

The old load method was like:
```
	SGVector<T> vec;
	loader->get_vector(vec.vector, vec.vlen);
	copy_data(vec);
	copy_refcount(vec);
	ref();
```

tests/unit/io/Cereal_unittest.cc
```diff
@@ -0,0 +1,31 @@
...
+
+	{
+		std::ifstream is("cereatltest_sgvector.cereal");
+		cereal::JSONInputArchive archive(is);
+		archive(b);
+	}
+
+	for (index_t i = 0; i < size; i++)
+		EXPECT_NEAR(a[i], b[i], 1E-15);
+}
```
vigsterkr: please make sure that you delete the `cereatltest_sgvector.cereal` file before exiting this function:
 * anticipate IO exceptions...
 * please use temporary files (see the other serialization unittests) to avoid any possible inconsistent state...

tests/unit/io/Cereal_unittest.cc
```diff
@@ -0,0 +1,52 @@
...
+		cereal::JSONOutputArchive archive(os);
+		archive(a);
+	}
+
+	{
+		std::ifstream is("cerealtest_sgvector_false.cereal");
+		cereal::JSONInputArchive archive(is);
+		archive(b);
+	}
+	EXPECT_NEAR(b.ref_count(), -1, 1E-15);
```
vigsterkr: use `EXPECT_EQ` as the `ref_count` is an `int32_t`

tests/unit/io/Cereal_unittest.cc
```diff
@@ -0,0 +1,52 @@
...
+		cereal::JSONOutputArchive archive(os);
+		archive(a);
+	}
+
+	{
+		std::ifstream is("cerealtest_sgvector.cereal");
+		cereal::JSONInputArchive archive(is);
+		archive(b);
+	}
+	for (index_t i = 0; i < size; i++)
```
vigsterkr: check the ref_count value

tests/unit/io/Cereal_unittest.cc
```diff
@@ -0,0 +1,52 @@
...
+		EXPECT_NEAR(a[i], b[i], 1E-15);
+}
+
+TEST(Cereal, Json_SGVector_load_equals_saved_refcounting_false)
```
vigsterkr: i would add a test that creates a vector call `SG_REF` on it 2-3 times, serialize it and then check the deserialized vectors ref_count value...

tests/unit/io/Cereal_unittest.cc
```diff
@@ -2,38 +2,55 @@
...
+#include <cstdio>
+
 #include <shogun/lib/SGVector.h>
 #include <cereal/archives/json.hpp>
-#include <fstream>
 
 using namespace shogun;
 
 TEST(Cereal, Json_SGVector_load_equals_saved)
 {
```
OXPHOS: @vigsterkr I added `tmpnam` and `std::ios_base::failure` to this test. Do you want to have a look? It doesn't look very pleasant but it's working at local (I'll try to figure out why travis cannot find cereal files)

tests/unit/io/Cereal_unittest.cc
```diff
@@ -2,38 +2,55 @@
...
+	SGVector<float64_t> a(size);
+	SGVector<float64_t> b;
 	a.range_fill(1.0);
 
-	{
-		std::ofstream os("cerealtest_sgvector.cereal");
-		cereal::JSONOutputArchive archive(os);
-		archive(a);
-	}
+	char name[256] = "cerealtest_sgvector.cereal";
```
vigsterkr: do we really care the name of the file? no

tests/unit/io/Cereal_unittest.cc
```diff
@@ -2,38 +2,55 @@
...
+	char* file = tmpnam (name);
```
vigsterkr: * use it as in the example of the api i gave you:
```
std::string filename = std::tmpnam(nullptr);
```

tests/unit/io/Cereal_unittest.cc
```diff
@@ -2,38 +2,55 @@
...
+		EXPECT_EQ(b.ref_count(), a.ref_count());
+		for (index_t i = 0; i < size; i++)
+			EXPECT_NEAR(a[i], b[i], 1E-15);
+
+		std::remove(file);
 	}
-	
-	for (index_t i = 0; i < size; i++)
-		EXPECT_NEAR(a[i], b[i], 1E-15);
+	catch (const std::ios_base::failure& e)
```
vigsterkr: when an exception happens, the `std::remove` will not be called, so you will need to call it in the catch section as well.
no need to narrow down to io exceptions only, catch all of them by:
```catch (std::exception& e)```

File: tests/unit/io/Cereal_unittest.cc [hunk 15]
```diff
@@ -2,38 +2,55 @@
...
 
 TEST(Cereal, Json_SGVector_load_equals_saved_refcounting_false)
 {
 	const index_t size = 5;
-	SGVector<float64_t> a(size, false), b;
+	SGVector<float64_t> a(size, false);
+	SGVector<float64_t> b;
 	a.range_fill(1.0);
 
 	{
```
vigsterkr: change here the filename to `std::tempnam` as well, as well as wrap it around with try/catch

tests/unit/io/Cereal_unittest.cc
```diff
@@ -2,38 +2,55 @@
...
+			cereal::JSONOutputArchive archive(os);
+			archive(a);
+		}
+
+		{
+			std::ifstream is(file);
+			cereal::JSONInputArchive archive(is);
+			archive(b);
+		}
+
```
vigsterkr: btw end here the try block... dont catch the GTest related exceptions...

src/shogun/lib/SGVector.h
karlnapf: ok!

karlnapf:
Needs rebase
@vigsterkr from my side this is fine to merge

vigsterkr:
@karlnapf lot of updates are still pending... :) we'll get there sooner or later

examples/meta/CMakeLists.txt
```diff
@@ -8,6 +8,7 @@
 # execute_process(COMMAND sh ${CMAKE_CURRENT_SOURCE_DIR}/generator/types/get_type_list.sh > ${CMAKE_CURRENT_SOURCE_DIR}/generator/types/typelist)
 
 file(MAKE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/parser_files)
+include_directories(${CEREAL_INCLUDE_DIRS})
```
vigsterkr: ok i understand this... buuuuut and here comes the *pain* part: in the top level CMakeLists.txt you do this:
https://github.com/OXPHOS/shogun/blob/86425a27566deec4f8eb065af5526f23571a031c/CMakeLists.txt#L595

a.k.a. what we did for eigen. if you end up doing the internal dependency to have this working (above) you would need to ship cereal with shogun in the header directory... which is quite a problem.... only way to solve this if we could add all the cereal related implementations/definitions (declaration can go into headers as that can be easily solved) into .cpp

OXPHOS:
Merged via #3375

PR #3230, opened 2016-05-30
Linalg Refactor - CPU dot

OXPHOS:
- I'll separate classes into different .h and .cpp later
- Not much work in GPU part yet.
- Works for Eigen backend, dot(a, b)

@lambday @vigsterkr @karlnapf


karlnapf:
Cool!
I guess next step would be to test a GPU dot product, and then two cases:
  * GPU lib is available (and dot is computed on GPU after a transfer)
 * GPU lib is not available (fall back to CPU)

@lambday should also comment

src/shogun/mathematics/linalg/linalgRefactor.h
```diff
@@ -0,0 +1,171 @@
...
+    {
+        return false;
+    }
+};
+
+template <class T>
+struct CPU_Vector : public BaseVector<T>
+{
+    //<SGVector<T>>* CPUptr;
+    SGVector<T> vec;
```
lambday: can we just hold a reference to the SGVector instead instead? We won't be passing this CPUVector around via any API, so an extra refcount is not really needed I think.

src/shogun/mathematics/linalg/linalgRefactor.h
```diff
@@ -0,0 +1,171 @@
...
+    BaseVector(const SGVector<T> vector){}
+
+    virtual bool onGPU()
+    {
+        return false;
+    }
+};
+
+template <class T>
+struct CPU_Vector : public BaseVector<T>
```
lambday: Drop the underscore '_' from the class name. CamelCase should be enough

src/shogun/mathematics/linalg/linalgRefactor.h
```diff
@@ -0,0 +1,171 @@
...
+#include <shogun/mathematics/eigen3.h>
+#include <iostream>
+
+using namespace shogun;
+
+template <class T>
+struct BaseVector
+{
+    BaseVector(){}
+    BaseVector(const SGVector<T> vector){}
```
lambday: We don't need this constructor. Remember, an instance of BaseVector is never really created. It is always either CPUVector or GPUVector.

src/shogun/mathematics/linalg/linalgRefactor.h
```diff
@@ -0,0 +1,171 @@
...
+
+    virtual bool onGPU()
```
lambday: Make this pure virtual. The base class doesn't really know whether the data is on CPU or GPU.

src/shogun/mathematics/linalg/linalgRefactor.h
```diff
@@ -0,0 +1,171 @@
...
+    bool onGPU()
+    {
+        return false;
+    }
+};
+
+template <typename T>
+struct GPU_Vector : public BaseVector<T>
+{
+#ifdef HAVE_VIENNACL
```
lambday: We gotta make this pimpl. No hard-dependency on ViennaCL in the headers :) Maybe when we split the header and the cpp, you can do that?

src/shogun/mathematics/linalg/linalgRefactor.h
```diff
@@ -0,0 +1,171 @@
...
+class GPUBackend
+{
+public:
+ #ifdef HAVE_VIENNACL
+    template <class T>
+    T dot(GPU_Vector<T> a, GPU_Vector<T> b)
+    {
+        // Dereference a.GPUptr and b.GPUptr to vcl_vector?
+        // viennacl::linalg::inner_prod(vcl_vector_a, vcl_vector_b);
+        // Transfer back to CPU end???
```
lambday: Just do what the present viennacl dot is doing.

src/shogun/mathematics/linalg/linalgRefactor.h
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+   // similarly, other methods
+};
+
+//template <typename T>
+class GPUBackend
+{
+public:
+ #ifdef HAVE_VIENNACL
```
lambday: Remember to keep the header free of this dependency when you split your code into cpp and headers. Only the cpp file should have this if-def checks, if any.

src/shogun/mathematics/linalg/linalgRefactor.h
```diff
@@ -0,0 +1,171 @@
...
+  #else
+   template <class T>
+   T dot(GPU_Vector<T> a, GPU_Vector<T> b)
+   {
+       throw std::runtime_error("user did not register GPU backend");
+   }
+ #endif
+ };
+
+class LinalgRefactor
```
lambday: Why does this class needs to exist?

tests/unit/mathematics/linalg/LinalgRefactor_unittest.cc
```diff
@@ -0,0 +1,46 @@
...
+    a.set_const(1);
+    b.set_const(2);
+
+    CPU_Vector<int32_t> a_CPU(a);
+    CPU_Vector<int32_t> b_CPU(b);
+
+    CPUBackend eigenBackend;
+    LinalgRefactor lr(&eigenBackend);
+
+    int32_t result = lr.dot<int32_t>(&a_CPU, &b_CPU);
```
lambday: The compiler can deduce the type T. Can you try without specifying it explicitly.

src/shogun/mathematics/linalg/linalgRefactor.h
lambday: Oh I see. My bad.

tests/unit/mathematics/linalg/LinalgRefactor_unittest.cc
```diff
@@ -0,0 +1,46 @@
...
+    const index_t size=10;
+    SGVector<int32_t> a(size), b(size);
+    a.set_const(1);
+    b.set_const(2);
+
+    CPU_Vector<int32_t> a_CPU(a);
+    CPU_Vector<int32_t> b_CPU(b);
+
+    CPUBackend eigenBackend;
+    LinalgRefactor lr(&eigenBackend);
```
lambday: This is okay for now. But keep in mind that the creation of this object has to be done in `init_shogun()` call, since this object better be global. We don't want to create it every-time. We should be able to use this line `sg_linalg->dot(..., ...)`, in a similar fashion we use `sg_io` etc.

lambday:
Alright this looks okay! Needs some more polishing.

Apart from what @karlnapf already mentioned,

- Split the files, make sure it works after splitting
- Use the pimpl pattern to remove the dependency from ViennaCL in the headers
- Draft a small POC of how the Factory class should work. This class would be responsible for creating the instances of CPUVectors or GPUVectors. Things to keep in mind - we want the construction to be as lightweight as possible, i.e. avoid reference-counting if it is not really needed, etc
- Change the name LinalgRefactor to something that's gonna stay. We need a global instance of that class. The methods should be state-less and thread-safe. Make sure that they are.
-  Benchmark **MUST**. Write a small example in which (a) we do the dot using eigen3 directly and (b) we use this linalg dot (you use benchmarking libraries google-benchmark/hayai etc) - take small to large vectors, run 1000 times, check how much we're paying for all this flexibility.

src/shogun/mathematics/linalg/linalgRefactor.h
OXPHOS: Thanks for your comments! @lambday  I am using underscore because there's already CPUVector (or GPUVector or both) in shogun namespace. I'll look into the other points now.

OXPHOS:
@vigsterkr @lambday There're many ways to put `ifdef` in .cpp for `dot()`. I currently didn't use opaque pointer because I can only think of Class - Member Function - Class, kinda complicated..?

But on a second thought I don't think my current implementation is interchangeable..

lambday:
@OXPHOS please put one class per file. So there will be

1. BaseVector.h
2. CPUVector.h
3. CPUVector.cpp
4. GPUVector.h
5. GPUVector.cpp
6. Linalg.h
7. Linalg.cpp
8. CPUBackend.h
9. CPUBackend.cpp
10. GPUBackend.h
11. GPUBackend.cpp

When you split things like that, it would be easier to refactor things with opaque ptrs.

OXPHOS:
@lambday please check. thx!

src/shogun/mathematics/linalgrefactor/BaseVector.h
```diff
@@ -0,0 +1,20 @@
+#include <shogun/lib/config.h>
+#include <shogun/lib/SGVector.h>
+
+#ifndef _BASEVECTOR_H__
```
lambday: get rid of the leading '_'. those are reserved for system headers. We should just use `BASEVECTOR_H__`

src/shogun/mathematics/linalgrefactor/CPUBackend.h
```diff
@@ -0,0 +1,27 @@
...
+
+namespace shogun
+{
+
+/** Backend Class **/
+class CPUBackend
+{
+public:
+    template <typename T>
+    T dot(const CPUVector<T> &a, const CPUVector<T> &b)
```
lambday: Move the implementation to the `.cpp`

src/shogun/mathematics/linalgrefactor/GPU_Vector.h
```diff
@@ -0,0 +1,26 @@
...
+namespace shogun
+{
+
+template <class T>
+struct GPU_Vector : public BaseVector<T>
+{
+private:
+    struct GPUArray;
+public:
+    std::unique_ptr<GPUArray> gpuarray;
```
lambday: This should not be public

data
```diff
@@ -1 +1 @@
-Subproject commit 77cba34872e1cf4c4eabd4f9540daa4e0ea76b06
+Subproject commit ee04cf3387575dbefb16d8f1d60605e041a90b21
```
lambday: Why the data commit?

src/shogun/mathematics/linalgrefactor/CPUBackend.h
```diff
@@ -0,0 +1,27 @@
+#include <shogun/lib/config.h>
+#include <shogun/mathematics/eigen3.h>
+#include <shogun/mathematics/linalgrefactor/CPUVector.h>
+
+#ifndef _CPUBACKEND_H__
+#define _CPUBACKEND_H__
```
lambday: Same as above

src/shogun/mathematics/linalgrefactor/CPUVector.h
```diff
@@ -0,0 +1,25 @@
+#include <shogun/lib/config.h>
+#include <shogun/lib/SGVector.h>
+#include <shogun/mathematics/linalgrefactor/BaseVector.h>
+
+#ifndef _CPUVECTOR_H__
+#define _CPUVECTOR_H__
```
lambday: Same as above

src/shogun/mathematics/linalgrefactor/GPUBackend.h
```diff
@@ -0,0 +1,20 @@
+#include <shogun/lib/config.h>
+#include <shogun/mathematics/eigen3.h>
+#include <shogun/mathematics/linalgrefactor/GPU_Vector.h>
+
+#ifndef _GPUBACKEND_H__
+#define _GPUBACKEND_H__
```
lambday: Same as above

src/shogun/mathematics/linalgrefactor/GPUBackend.cpp
```diff
@@ -0,0 +1,31 @@
...
+            * @param a first vector
+            * @param b second vector
+            * @return the dot product of \f$\mathbf{a}\f$ and \f$\mathbf{b}\f$, computed
+            * as \f$\sum_i a_i b_i\f$
+            */
+           T compute(GPU_Vector<T> a, GPU_Vector<T> b)
+           {
+               return viennacl::linalg::inner_prod(*(a.GPUptr), *(b.GPUptr));
+           }
+       }
```
lambday: Indentation :)

src/shogun/mathematics/linalgrefactor/GPU_Vector.h
```diff
@@ -0,0 +1,26 @@
+#include <shogun/lib/config.h>
+#include <shogun/lib/SGVector.h>
+#include <shogun/mathematics/linalgrefactor/BaseVector.h>
+#include <memory>
+
+#ifndef _GPU_VECTOR_H__
+#define _GPU_VECTOR_H__
```
lambday: Same as above

src/shogun/mathematics/linalgrefactor/linalgRefactor.h
```diff
@@ -0,0 +1,102 @@
...
+#include <shogun/lib/config.h>
+#include <shogun/mathematics/eigen3.h>
+#include <shogun/mathematics/linalgrefactor/CPUBackend.h>
+#include <shogun/mathematics/linalgrefactor/GPUBackend.h>
+#include <shogun/mathematics/linalgrefactor/CPUVector.h>
+#include <shogun/mathematics/linalgrefactor/GPU_Vector.h>
+#include <memory>
+
+#ifndef _LINALGR_H__
+#define _LINALGR_H__
```
lambday: Same as above

src/shogun/mathematics/linalgrefactor/linalgRefactor.h
```diff
@@ -0,0 +1,102 @@
...
+
+    LinalgRefactor(CPUBackend* cpu_backend):cpubackend(cpu_backend), gpubackend(nullptr){}
+
+    LinalgRefactor(GPUBackend* gpu_backend):cpubackend(nullptr), gpubackend(gpu_backend){}
+
+    LinalgRefactor(CPUBackend* cpu_backend, GPUBackend* gpu_backend)
+    :cpubackend(cpu_backend), gpubackend(gpu_backend){}
+
+    template <class T>
+    T dot(BaseVector<T>* a, BaseVector<T>* b)
```
lambday: Move the implementation to `.cpp`

OXPHOS:
@lambday dunno why it's failing some tests..also this doesn't seem to be thread safe?

src/shogun/mathematics/linalgrefactor/GPUBackend.cpp
```diff
@@ -0,0 +1,30 @@
...
+    * Method that computes the dot product using ViennaCL
+    *
+    * @param a first vector
+    * @param b second vector
+    * @return the dot product of \f$\mathbf{a}\f$ and \f$\mathbf{b}\f$, computed
+    * as \f$\sum_i a_i b_i\f$
+    */
+    T compute(GPU_Vector<T> a, GPU_Vector<T> b)
+    {
+       return viennacl::linalg::inner_prod(*(a.GPUptr), *(b.GPUptr));
```
lambday: What is this GPUptr? It's not declared or defined in GPU_Vector class.

lambday: Shouldn't it be working with the VCLMemoryArray shared ptr thing that it has?

lambday:
@OXPHOS alright this looks better. Slowly getting there :)

Please fix the viennacl code. Make sure everything runs in your local shogun installation.

tests/unit/mathematics/linalg/LinalgRefactor_unittest.cc
```diff
@@ -0,0 +1,52 @@
...
+    const index_t size=10;
+    SGVector<int32_t> a(size), b(size);
+    a.set_const(1);
+    b.set_const(2);
+
+    CPUVector<int32_t> a_CPU(a);
+    CPUVector<int32_t> b_CPU(b);
+
+    CPUBackend eigenBackend;
+    Linalg lr;
```
lambday: No need to create the Linalg object here now. You have already added the global instance of it.

tests/unit/mathematics/linalg/LinalgRefactor_unittest.cc
```diff
@@ -0,0 +1,52 @@
...
+    lr.set_cpu_backend(&eigenBackend);
+
+    auto result = lr.dot<int32_t>(&a_CPU, &b_CPU);
```
lambday: Use the global linalg object.

lambday:
@OXPHOS it should be thread safe. Imagine the use-case. You have a bunch of vectors. In multiple threads, you'll be calling the dot method for those in the global linalg instance. Since the method is stateless (takes two vectors at a time, returns a scalar), it shouldn't be a problem I suppose.

OXPHOS:
I have to put GPUBackend and GPUVector in the same file, or GPUBackend cannot compile because it calls method within "opaque" struct GPUArray.

Also I have the following errors when using ViennaCL. Seems like big trouble..

https://gist.github.com/OXPHOS/70e7fd0e50c7e794c374067100f8413c

lambday:
@OXPHOS hi

- I see. Should have seen it coming. Anyway, you probably won't need to put those in the same file. Create a separate file for the impl, say, `GPUVectorImpl` (ViennaCL ifdefs can be used in its header), and then in your `GPUVector` class, you forward declare it and hold a pointer to the impl. And then in your `GPUBackend` class cpp, include that `GPUVectorImpl` header and you're done. See what I mean?

- Probably it's not that big of a trouble as you think. Your graphics card is not that good. You can still use `float32_t`.

OXPHOS:
Sure I see!
Guess I have to give up on some unit-tests then..was thinking about passing them all

lambday:
@OXPHOS sure they will work if you install Intel's opencl sdk thingi that works with your main memory and not graphics card memory. I remember I did it quite a long time back, as my graphics card sucks as well (double precision operations not supported). 

Google it a bit. If you can't find it maybe I can look for it again.

Also, install `clinfo`. Useful.

lambday:
oh, that only works if you have intel's integrated graphics card. not sure which one you have though.

lambday:
@OXPHOS any updates on this one? :)

OXPHOS:
@lambday 
- Added some new methods and constructors. 
- More unit-tests
- `sg_linalg` Global variable
- ViennaCL backend functional

src/shogun/mathematics/linalgrefactor/BaseVector.h
```diff
@@ -0,0 +1,21 @@
...
+#define BASEVECTOR_H__
+
+namespace shogun
+{
+
+/** Vector structure **/
+template <class T>
+struct BaseVector
+{
+    BaseVector(){}
```
lambday: newline :)
```c++
BaseVector()
{
}
```

lambday: Please make similar changes in other places where you've used this

src/shogun/mathematics/linalgrefactor/CPUBackend.cpp
```diff
@@ -0,0 +1,16 @@
+#include <shogun/mathematics/linalgrefactor/CPUBackend.h>
+
+namespace shogun
+{
+
+template <typename T> T CPUBackend::dot(const CPUVector<T> &a, const CPUVector<T> &b)
```
lambday: Usually in shogun the style is to put a newline after the `template <typename T>`. Make similar changes in other places where you've used this

src/shogun/mathematics/linalgrefactor/CPUVector.cpp
```diff
@@ -0,0 +1,19 @@
+#include <shogun/mathematics/linalgrefactor/CPUVector.h>
+
+namespace shogun
+
+{
+template <class T>
+CPUVector<T>::CPUVector():CPUptr(nullptr), vlen(0) { }
```
lambday: newline

src/shogun/mathematics/linalgrefactor/CPUVector.cpp
```diff
@@ -0,0 +1,19 @@
...
+
+template <class T>
+CPUVector<T>::CPUVector(const SGVector<T> &vector)
+: CPUptr(vector.vector), vlen(vector.vlen) { }
```
lambday: newline

src/shogun/mathematics/linalgrefactor/CPUVector.cpp
```diff
@@ -0,0 +1,19 @@
...
+
+template <class T>
+CPUVector<T>::CPUVector(const CPUVector<T> &vector)
+: CPUptr(vector.CPUptr), vlen(vector.vlen) { }
```
lambday: newline

src/shogun/mathematics/linalgrefactor/GPUArray.cpp
```diff
@@ -0,0 +1,46 @@
...
+
+    viennacl::backend::memory_write(*GPUptr, 0, vlen*sizeof(T),
+        vector.vector);
+}
+
+template <class T>
+GPU_Vector<T>::GPUArray::GPUArray(const GPU_Vector<T>::GPUArray &array)
+//:GPUptr(new VCLMemoryArray()), vlen(array.vlen), offset(0)
+{
+    GPUptr = array.GPUptr;
```
lambday: indentation

src/shogun/mathematics/linalgrefactor/GPU_Vector.cpp
```diff
@@ -0,0 +1,63 @@
+#include <shogun/mathematics/linalgrefactor/GPU_Vector.h>
+#include <shogun/mathematics/linalgrefactor/GPUArray.h>
+
+namespace shogun
+{
+
+template <class T>
+GPU_Vector<T>::GPU_Vector()
```
lambday: No underscores in class names :)

src/shogun/mathematics/linalgrefactor/GPU_Vector.cpp
```diff
@@ -0,0 +1,63 @@
...
+{
+    vlen = 0;
+    offset = 0;
+}
+
+template <class T>
+GPU_Vector<T>& GPU_Vector<T>::operator=(const GPU_Vector<T> &other)
+{
+    // check for self-assignment
+    if(&other == this)
```
lambday: Nice! :)

src/shogun/mathematics/linalgrefactor/GPU_Vector.cpp
```diff
@@ -0,0 +1,63 @@
...
+        return *this;
+
+    // reuse storage when possible
+    gpuarray.reset(new GPUArray(*(other.gpuarray)));
+    vlen = other.vlen;
+    return *this;
+}
+
+template <class T> GPU_Vector<T>::~GPU_Vector() { }
```
lambday: newline

src/shogun/mathematics/linalgrefactor/GPU_Vector.h
```diff
@@ -0,0 +1,49 @@
...
+
+    GPU_Vector(const GPU_Vector<T> &vector);
+
+    ~GPU_Vector();
+
+    void init();
+
+    GPU_Vector<T>& operator=(const GPU_Vector<T> &other);
+
+    bool onGPU() { return true; }
```
lambday: make it 
```c++
inline bool onGPU() const
{
    return true;
}
```

src/shogun/mathematics/linalgrefactor/GPU_Vector.h
```diff
@@ -0,0 +1,49 @@
...
+
+public:
+    /** Vector length */
+    index_t vlen;
+
+    /** Offset for the memory segment, i.e the data of the vector
+	 * starts at vector+offset
+	 */
+	index_t offset;
```
lambday: indentation

src/shogun/mathematics/linalgrefactor/linalgRefactor.cpp
```diff
@@ -0,0 +1,77 @@
+#include <shogun/mathematics/linalgrefactor/linalgRefactor.h>
+
+namespace shogun
+{
+
+Linalg::Linalg():m_cpubackend(&eigenBackend), m_gpubackend(nullptr){}
```
lambday: newline

src/shogun/mathematics/linalgrefactor/linalgRefactor.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+Linalg::Linalg(CPUBackend* cpubackend):m_cpubackend(cpubackend), m_gpubackend(nullptr){}
```
lambday: newline

src/shogun/mathematics/linalgrefactor/linalgRefactor.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+Linalg::Linalg(GPUBackend* gpubackend):m_cpubackend(&eigenBackend), m_gpubackend(gpubackend){}
```
lambday: newline

src/shogun/mathematics/linalgrefactor/linalgRefactor.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+Linalg::Linalg(CPUBackend* cpubackend, GPUBackend* gpubackend)
+:m_cpubackend(cpubackend), m_gpubackend(gpubackend){}
```
lambday: newline

src/shogun/mathematics/linalgrefactor/linalgRefactor.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+template <class T>
+T Linalg::dot(BaseVector<T> *a, BaseVector<T> *b)
+{
+    if (a->onGPU() && b->onGPU())
+    {
+        if (this->hasGPUBackend())
+        {
+            // do the gpu backend dot product
+            // you shouldn't care whether it's viennacl or some other GPU backend.
```
lambday: remove the comment

src/shogun/mathematics/linalgrefactor/linalgRefactor.cpp
```diff
@@ -0,0 +1,77 @@
...
+            return m_gpubackend->dot<T>(static_cast<GPU_Vector<T>&>(*a),
+                                            static_cast<GPU_Vector<T>&>(*b));
+        } else {
```
lambday: newline before and after `else`

src/shogun/mathematics/linalgrefactor/linalgRefactor.cpp
```diff
@@ -0,0 +1,77 @@
...
+            SG_SERROR("User did not register GPU backend. \n");
+            return -1;
+        }
+    }
+    else {
```
lambday: newline

src/shogun/mathematics/linalgrefactor/linalgRefactor.cpp
```diff
@@ -0,0 +1,77 @@
...
+        // take care that the matricies are on the same backend
+        if (a->onGPU()){
```
lambday: newline

src/shogun/mathematics/linalgrefactor/linalgRefactor.cpp
```diff
@@ -0,0 +1,77 @@
...
+            SG_SERROR("User did not register GPU backend. \n");
+            return -1;
+        }
+        else if (b->onGPU()) {
```
lambday: newline

src/shogun/mathematics/linalgrefactor/linalgRefactor.h
```diff
@@ -0,0 +1,88 @@
...
+#ifndef LINALGR_H__
+#define LINALGR_H__
+
+namespace shogun
+{
+/** Linalg Class **/
+class Linalg
+{
+    CPUBackend eigenBackend;
+    CPUBackend* m_cpubackend;
```
lambday: `unique_ptr` :)

src/shogun/mathematics/linalgrefactor/linalgRefactor.h
```diff
@@ -0,0 +1,88 @@
...
+    GPUBackend* m_gpubackend;
```
lambday: same as above

tests/unit/mathematics/LinalgRefactor_unittest.cc
```diff
@@ -0,0 +1,80 @@
...
+    CPUVector<int32_t> a_CPU(a);
+    CPUVector<int32_t> b_CPU(b);
+
+    auto result = sg_linalg->dot(&a_CPU, &b_CPU);
+
+    EXPECT_NEAR(result, 20.0, 1E-15);
+}
+
+#ifdef HAVE_VIENNACL
+
```
lambday: add test for GPUVector convert.

src/shogun/mathematics/linalgrefactor/CPUBackend.h
```diff
@@ -0,0 +1,21 @@
...
+
+namespace shogun
+{
+
+/** Backend Class **/
+class CPUBackend
+{
+public:
+    template <typename T>
+    T dot(const CPUVector<T> &a, const CPUVector<T> &b);
```
lambday: Use const method.
```c++
template <typename T>
T dot(const CPUVector<T> &a, const CPUVector<T> &b) const;
```

src/shogun/mathematics/linalgrefactor/CPUVector.h
```diff
@@ -0,0 +1,28 @@
...
+    T* CPUptr;
+    index_t vlen;
+
+    CPUVector();
+
+    CPUVector(const SGVector<T> &vector);
+
+    CPUVector(const CPUVector<T> &vector);
+
+    bool onGPU() { return false; }
```
lambday:
```c++
inline bool onGPU() const
{
    return false;
}
```

src/shogun/mathematics/linalgrefactor/GPUBackend.h
```diff
@@ -0,0 +1,26 @@
...
+namespace shogun
+{
+
+class GPUBackend
+{
+template <class T> friend struct GPU_Vector;
+
+public:
+    template <typename T>
+    T dot(const GPU_Vector<T> &a, const GPU_Vector<T> &b);
```
lambday:
```c++
template <typename T>
T dot(const GPU_Vector<T> &a, const GPU_Vector<T> &b) const;
```

src/shogun/mathematics/linalgrefactor/linalgRefactor.h
```diff
@@ -0,0 +1,88 @@
...
+    void set_cpu_backend(CPUBackend* cpubackend);
+
+    CPUBackend* get_cpu_backend();
+
+    void set_gpu_backend(GPUBackend* gpubackend);
+
+    GPUBackend* get_gpu_backend();
+
+    template <class T>
+    T dot(BaseVector<T> *a, BaseVector<T> *b);
```
lambday: make it
```c++
template <class T>
T dot(BaseVector<T> *a, BaseVector<T> *b) const;
```

OXPHOS:
- I think Heiko @karlnapf showed me a header different from 
https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/statistics/HypothesisTest.h
last time but I cannot find it for now.. so I didn't add the header yet.

- For unit-test case: GPU convert (i.e. test whether SGVector can be convert to GPU), it requires element access [] which needs more work

lambday:
Hi @OXPHOS 

- Doesn't matter. Use whichever you want. But license has to be there.
- Can you bypass requiring the `operator[]` by casting it to viennacl vector type and using it's access operator?

OXPHOS:
@lambday both done.

src/shogun/base/init.cpp
```diff
@@ -37,6 +38,7 @@ namespace shogun
 	Version* sg_version=NULL;
 	CMath* sg_math=NULL;
 	CRandom* sg_rand=NULL;
+	std::unique_ptr<SGLinalg> sg_linalg(nullptr);
```
lambday: can you please try using shogun's Unique here? if that works then we won't need c++11 guard around all these

src/shogun/base/init.cpp
```diff
@@ -66,6 +68,8 @@ namespace shogun
 			sg_math = new shogun::CMath();
 		if (!sg_rand)
 			sg_rand = new shogun::CRandom();
+		if (!sg_linalg)
+			sg_linalg = std::unique_ptr<SGLinalg>(new shogun::SGLinalg());
```
lambday: probably then you won't need to create an instance here (if you use `Unique`)

tests/unit/mathematics/LinalgRefactor_unittest.cc
```diff
@@ -0,0 +1,103 @@
...
+    for (index_t i = 0; i < size; ++i)
+        EXPECT_NEAR(a[i], (a_CPU.CPUptr)[i], 1E-15);
+}
+
+TEST(SGLinalg, CPU_Vector_copy)
+{
+    const index_t size = 10;
+    SGVector<int32_t> a(size);
+    for (index_t i = 0; i < size; ++i) a[i] = i;
```
lambday: `a[i] = i` in a new line please :)

src/shogun/base/init.h
```diff
@@ -108,6 +109,18 @@ void set_global_rand(CRandom* rand);
  */
 CRandom* get_global_rand();
 
+/** set the global linalg library object
+ *
+ * @param linalg linalg object to use
+ */
+void set_global_linlg(SGLinalg* linalg);
```
lambday: Is it defined somewhere? I see definition of get_global_linalg() in the cpp but not this one. 

Also, maybe this is not needed. Just remove this for now :)

src/shogun/mathematics/linalg/BaseVector.h
```diff
@@ -0,0 +1,63 @@
...
+template <class T>
+struct BaseVector
+{
+	/** Default Constructor */
+	BaseVector()
+	{
+	}
+
+	/** Data Storage */
+	virtual bool onGPU() = 0 ;
```
lambday: make it const :)

lambday: oh for these methods that return something, you should use `@return` and then just write the description :)

src/shogun/mathematics/linalg/CPUVector.h
```diff
@@ -0,0 +1,75 @@
...
+	CPUVector();
+
+	/** Wrap SGVector */
+	CPUVector(const SGVector<T> &vector);
+
+	/** Copy Constructor*/
+	CPUVector(const CPUVector<T> &vector);
+
+	/** Data Storage */
+	inline bool onGPU()
```
lambday: const

src/shogun/mathematics/linalg/CPUVector.h
```diff
@@ -0,0 +1,75 @@
...
+{
+
+/** CPU vector structure */
+template <class T>
+struct CPUVector : public BaseVector<T>
+{
+	/** vector  */
+	T* CPUptr;
+	/** length of vector  */
+	index_t vlen;
```
lambday: Maybe put the data members at the bottom :)

src/shogun/mathematics/linalg/GPUArray.cpp
```diff
@@ -0,0 +1,89 @@
...
+{
+	return VCLVector(GPUVector<T>::GPUArray::GPUvec());
+}
+
+#else // HAVE_VIENNACL
+
+template <class T>
+GPUVector<T>::GPUArray::GPUArray(const SGVector<T> &vector)
+{
+	SG_SERROR("User did not register GPU backend. \n");
```
lambday: Nice :)

src/shogun/mathematics/linalg/GPUArray.h
```diff
@@ -0,0 +1,100 @@
...
+	/** Memory segment holding the data for the vector */
+	std::shared_ptr<VCLMemoryArray> GPUptr;
+
+	/** Vector length */
+	index_t vlen;
+
+	/** Offset for the memory segment, i.e the data of the vector
+	 * starts at vector+offset
+	 */
+	index_t offset;
```
lambday: data members at the bottom :)

src/shogun/mathematics/linalg/GPUVector.h
```diff
@@ -0,0 +1,96 @@
...
+	/** Overload operator "=" */
+	GPUVector<T>& operator=(const GPUVector<T> &other);
+
+	/** Data Storage */
+	inline bool onGPU()
+	{
+		return true;
+	}
+
+public:
```
lambday: If it is a struct, you don't need to declare anything public :)

src/shogun/mathematics/linalg/GPUVector.h
```diff
@@ -0,0 +1,96 @@
...
+	~GPUVector();
+
+	/** needs to be overridden to initialize empty data */
+	void init();
+
+	/** Overload operator "=" */
+	GPUVector<T>& operator=(const GPUVector<T> &other);
+
+	/** Data Storage */
+	inline bool onGPU()
```
lambday: const :)

src/shogun/mathematics/linalg/GPUVector.h
```diff
@@ -0,0 +1,96 @@
...
+
+/** GPU vector wrapper structure */
+template <class T>
+struct GPUVector : public BaseVector<T>
+{
+	/** Structure for different types of GPU vectors */
+	struct GPUArray;
+
+	/** Opaque pointer of GPUArray */
+	std::unique_ptr<GPUArray> gpuarray;
```
lambday: data members at the bottom :)

src/shogun/mathematics/linalg/SGLinalg.cpp
```diff
@@ -0,0 +1,125 @@
...
+
+template <class T>
+T SGLinalg::dot(BaseVector<T> *a, BaseVector<T> *b) const
+{
+	if (a->onGPU() && b->onGPU())
+	{
+		if (this->hasGPUBackend())
+		{
+			// do the gpu backend dot product
+			// you shouldn't care whether it's viennacl or some other GPU backend.
```
lambday: remove these comments :)

src/shogun/mathematics/linalg/SGLinalg.h
```diff
@@ -0,0 +1,116 @@
...
+{
+
+/** Linalg Class **/
+class SGLinalg
+{
+	/** pointer to cpubackend */
+	std::shared_ptr<CPUBackend> m_cpubackend;
+
+	/** pointer to gpubackend */
+	std::shared_ptr<GPUBackend> m_gpubackend;
```
lambday: data members at the bottom :)

src/shogun/mathematics/linalg/SGLinalg.h
```diff
@@ -0,0 +1,116 @@
...
+	 * @param a first vector
+	 * @param b second vector
+	 * @return the dot product of \f$\mathbf{a}\f$ and \f$\mathbf{b}\f$, represented
+	 * as \f$\sum_i a_i b_i\f$
+	 */
+	template <class T>
+	T dot(BaseVector<T> *a, BaseVector<T> *b) const;
+
+	/** Check whether gpubackend is registered by user */
+	bool hasGPUBackend() const;
```
lambday: Nice :)

lambday:
@OXPHOS really nice job with this :) Let me merge this for now. Travis fails seem unrelated, Plus this is a feature branch.

Check my minor comments. Maybe you can send a patch for those one later.

PR #3236, opened 2016-05-31
added openmp set num threads in CParallel

lambday:
@karlnapf @vigsterkr @lisitsyn @sonney2k please have a look.


src/shogun/base/Parallel.cpp
```diff
@@ -43,14 +54,14 @@ Parallel::~Parallel()
 int32_t Parallel::get_num_cpus() const
 {
 #if defined(LINUX) && defined(_SC_NPROCESSORS_ONLN)
-		return sysconf( _SC_NPROCESSORS_ONLN );
```
karlnapf: usually, we try to avoid whitespace changes in such patches, and would send a separate patch

karlnapf:
Fine with me

karlnapf:
Feel free to merge once travis is happy

tests/unit/base/Parallel_unittest.cc
```diff
@@ -0,0 +1,73 @@
...
+#include <shogun/base/Parallel.h>
+#include <gtest/gtest.h>
+#include <omp.h>
+
+using namespace shogun;
+
+TEST(Parallel, openmp_get_num_threads)
+{
+	int32_t omp_num_threads=omp_get_num_threads();
+	ASSERT_EQ(1, omp_num_threads);
```
karlnapf: why is this 1 by default?

karlnapf: why not num_cpus?

karlnapf: UPDATE sorry I got it now

tests/unit/base/Parallel_unittest.cc
```diff
@@ -0,0 +1,73 @@
...
+
+	int32_t sg_num_threads=get_global_parallel()->get_num_threads();
```
karlnapf: @Saurabh7 this is the call you will need

karlnapf:
Travis is green everywhere it matters, merging

PR #3241, opened 2016-06-02
Meta Language fix static call import

sorig:
Fixes issue with imports of classes which are used for static methods calls #3183


OXPHOS:
Thanks! This should make life easier ;)

karlnapf:
I get it, this simply collects all classes ever touched (instantiated or static calls) in the dependencies. Thanks!

PR #3243, opened 2016-06-03
Lars update

Saurabh7:
Not merge ready.

### DIFF
--- a/src/shogun/regression/LeastAngleRegression.cpp
+++ b/src/shogun/regression/LeastAngleRegression.cpp
@@ -21,7 +21,9 @@
 #include <shogun/mathematics/lapack.h>
 #include <shogun/regression/LeastAngleRegression.h>
 #include <shogun/labels/RegressionLabels.h>
+#include <shogun/mathematics/eigen3.h>
 
+using namespace Eigen;
 using namespace shogun;
 using namespace std;
 
@@ -80,6 +82,8 @@ CLeastAngleRegression::CLeastAngleRegression(bool lasso) :
 	CLinearMachine(), m_lasso(lasso),
 	m_max_nonz(0), m_max_l1_norm(0)
 {
+	m_epsilon = CMath::MACHINE_EPSILON;
+	SG_ADD(&m_epsilon, "epsilon", "Epsilon for early stopping", MS_AVAILABLE);
 	SG_ADD(&m_max_nonz, "max_nonz", "Max number of non-zero variables", MS_AVAILABLE);
 	SG_ADD(&m_max_l1_norm, "max_l1_norm", "Max l1-norm of estimator", MS_AVAILABLE);
 }
@@ -127,25 +131,26 @@ bool CLeastAngleRegression::train_machine(CFeatures* data)
 	fill(m_is_active.begin(), m_is_active.end(), false);
 
 	SGVector<float64_t> y = ((CRegressionLabels*) m_labels)->get_labels();
+	Map<VectorXd> map_y(y.vector, y.size());
 	SGMatrix<float64_t> Xorig = feats->get_feature_matrix();
 
 	// transpose(X) is more convenient to work with since we care
 	// about features here. After transpose, each row will be a data
 	// point while each column corresponds to a feature
-	SGMatrix<float64_t> X(n_vec, n_fea, true);
-	for (int32_t i=0; i < n_vec; ++i)
-	{
-		for (int32_t j=0; j < n_fea; ++j)
-			X(i,j) = Xorig(j,i);
-	}
+	SGMatrix<float64_t> X (n_vec, n_fea);
+	Map<MatrixXd> map_Xr(Xorig.matrix, n_fea, n_vec);
+	Map<MatrixXd> map_X(X.matrix, n_vec, n_fea);
+	map_X = map_Xr.transpose();
+	
+	SGMatrix<float64_t> X_active(n_vec, n_fea);
 
 	// beta is the estimator
 	vector<float64_t> beta = make_vector(n_fea, 0);
 
 	vector<float64_t> Xy = make_vector(n_fea, 0);
+	Map<VectorXd> map_Xy(&Xy[0], n_fea);
 	// Xy = X' * y
-	cblas_dgemv(CblasColMajor, CblasTrans, n_vec, n_fea, 1, X.matrix, n_vec,
-		y.vector, 1, 0, &Xy[0], 1);
+	map_Xy=map_Xr*map_y;
 
 	// mu is the prediction
 	vector<float64_t> mu = make_vector(n_vec, 0);
@@ -172,13 +177,14 @@ bool CLeastAngleRegression::train_machine(CFeatures* data)
 	// main loop
 	//========================================
 	int32_t nloop=0;
-	while (m_num_active < max_active_allowed && max_corr > CMath::MACHINE_EPSILON && !stop_cond)
+	while (m_num_active < max_active_allowed && max_corr/n_vec > m_epsilon && !stop_cond)
 	{
 		// corr = X' * (y-mu) = - X'*mu + Xy
-		copy(Xy.begin(), Xy.end(), corr.begin());
-		cblas_dgemv(CblasColMajor, CblasTrans, n_vec, n_fea, -1,
-			X.matrix, n_vec, &mu[0], 1, 1, &corr[0], 1);
-
+		Map<VectorXd> map_corr(&corr[0], n_fea);
+		Map<VectorXd> map_mu(&mu[0], n_vec);
+		
+		map_corr = map_Xy - (map_Xr*map_mu);
+		
 		// corr_sign = sign(corr)
 		for (uint32_t i=0; i < corr.size(); ++i)
 			corr_sign[i] = CMath::sign(corr[i]);
@@ -189,63 +195,67 @@ bool CLeastAngleRegression::train_machine(CFeatures* data)
 		if (!lasso_cond)
 		{
 			// update Cholesky factorization matrix
-			R=cholesky_insert(X, R, i_max_corr);
+			if (m_num_active == 0)
+			{ 
+				// R isn't allocated yet
+				R=SGMatrix<float64_t>(1,1);
+				float64_t diag_k = map_X.col(i_max_corr).dot(map_X.col(i_max_corr));
+				R(0,0) = CMath::sqrt(diag_k);
+			}
+			else
+				R=cholesky_insert(X, X_active, R, i_max_corr, m_num_active);
 			activate_variable(i_max_corr);
 		}
 
-		// corr_sign_a = corr_sign[m_active_set]
-		vector<float64_t> corr_sign_a(m_num_active);
+		// Active variables
+		Map<MatrixXd> map_Xa(X_active.matrix, n_vec, m_num_active);
+		if (!lasso_cond)
+			map_Xa.col(m_num_active-1)=map_X.col(i_max_corr);
+		
+		SGVector<float64_t> corr_sign_a(m_num_active);
 		for (int32_t i=0; i < m_num_active; ++i)
 			corr_sign_a[i] = corr_sign[m_active_set[i]];
 
-		// GA1 = R\(R'\corr_sign_a)
-		vector<float64_t> GA1(corr_sign_a);
-		cblas_dtrsm(CblasColMajor, CblasLeft, CblasUpper, CblasTrans, CblasNonUnit,
-			m_num_active, 1, 1, R.matrix, m_num_active, &GA1[0], m_num_active);
-		cblas_dtrsm(CblasColMajor, CblasLeft, CblasUpper, CblasNoTrans, CblasNonUnit,
-			m_num_active, 1, 1, R.matrix, m_num_active, &GA1[0], m_num_active);
+		Map<VectorXd> map_corr_sign_a(corr_sign_a.vector, corr_sign_a.size());
+		Map<MatrixXd> map_R(R.matrix, R.num_rows, R.num_cols);
+		VectorXd solve = map_R.transpose().triangularView<Lower>().solve<OnTheLeft>(map_corr_sign_a);
+		VectorXd GA1 = map_R.triangularView<Upper>().solve<OnTheLeft>(solve);
 
 		// AA = 1/sqrt(GA1' * corr_sign_a)
-		float64_t AA = cblas_ddot(m_num_active, &GA1[0], 1, &corr_sign_a[0], 1);
+		float64_t AA = GA1.dot(map_corr_sign_a);
 		AA = 1/CMath::sqrt(AA);
 
-		// wA = AA*GA1
-		vector<float64_t> wA(GA1);
-		for (int32_t i=0; i < m_num_active; ++i)
-			wA[i] *= AA;
+		VectorXd wA = AA*GA1;
 
 		// equiangular direction (unit vector)
 		vector<float64_t> u = make_vector(n_vec, 0);
-		// u = X[:,m_active_set] * wA
-		for (int32_t i=0; i < m_num_active; ++i)
-		{
-			// u += wA[i] * X[:,m_active_set[i]]
-			cblas_daxpy(n_vec, wA[i],
-				X.get_column_vector(m_active_set[i]), 1, &u[0], 1);
-		}
+		Map<VectorXd> map_u(&u[0], n_vec);
+		
+		map_u = map_Xa*wA;
 
-		// step size
 		float64_t gamma = max_corr / AA;
 		if (m_num_active < n_fea)
 		{
+			#pragma omp parallel for shared(gamma)
 			for (int32_t i=0; i < n_fea; ++i)
 			{
 				if (m_is_active[i])
 					continue;
 
 				// correlation between X[:,i] and u
-				float64_t dir_corr = cblas_ddot(n_vec,
-					X.get_column_vector(i), 1, &u[0], 1);
+				float64_t dir_corr = map_u.dot(map_X.col(i));
 
 				float64_t tmp1 = (max_corr-corr[i])/(AA-dir_corr);
 				float64_t tmp2 = (max_corr+corr[i])/(AA+dir_corr);
+				#pragma omp critical
+				{
 				if (tmp1 > CMath::MACHINE_EPSILON && tmp1 < gamma)
 					gamma = tmp1;
 				if (tmp2 > CMath::MACHINE_EPSILON && tmp2 < gamma)
 					gamma = tmp2;
+				}
 			}
 		}
-
 		int32_t i_kick=-1;
 		int32_t i_change=i_max_corr;
 		if (m_lasso)
@@ -256,7 +266,7 @@ bool CLeastAngleRegression::train_machine(CFeatures* data)
 
 			for (int32_t i=0; i < m_num_active; ++i)
 			{
-				float64_t tmp = -beta[m_active_set[i]] / wA[i];
+				float64_t tmp = -beta[m_active_set[i]] / wA(i);
 				if (tmp > CMath::MACHINE_EPSILON && tmp < lasso_bound)
 				{
 					lasso_bound = tmp;
@@ -273,12 +283,11 @@ bool CLeastAngleRegression::train_machine(CFeatures* data)
 		}
 
 		// update prediction: mu = mu + gamma * u
-		cblas_daxpy(n_vec, gamma, &u[0], 1, &mu[0], 1);
+		map_mu += gamma*map_u;
 
 		// update estimator
 		for (int32_t i=0; i < m_num_active; ++i)
-			beta[m_active_set[i]] += gamma * wA[i];
-
+			beta[m_active_set[i]] += gamma * wA(i);
 		// early stopping on max l1-norm
 		if (m_max_l1_norm > 0)
 		{
@@ -301,11 +310,16 @@ bool CLeastAngleRegression::train_machine(CFeatures* data)
 		// if lasso cond, drop the variable from active set
 		if (lasso_cond)
 		{
-			beta[i_change] = 0; // ensure it be zero
-
-			// update Cholesky factorization
+			beta[i_change] = 0; 
 			R=cholesky_delete(R, i_kick);
 			deactivate_variable(i_kick);
+
+			// Remove column from active set
+			int32_t numRows = map_Xa.rows();
+			int32_t numCols = map_Xa.cols()-1;
+			if( i_kick < numCols )
+				map_Xa.block(0, i_kick, numRows, numCols-i_kick) = 
+					map_Xa.block(0, i_kick+1, numRows, numCols-i_kick).eval();			
 		}
 
 		nloop++;
@@ -318,8 +332,9 @@ bool CLeastAngleRegression::train_machine(CFeatures* data)
 		// early stopping with max number of non-zero variables
 		if (m_max_nonz > 0 && m_num_active >= m_max_nonz)
 			stop_cond = true;
+		SG_DEBUG("Added : %d , Dropped %d, Active set size %d max_corr %.17f \n", i_max_corr, i_kick, m_num_active, max_corr);
 
-	} // main loop
+	}
 
 	// assign default estimator
 	w.vlen = n_fea;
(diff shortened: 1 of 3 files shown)


src/shogun/regression/LeastAngleRegressionImpl.h
```diff
@@ -0,0 +1,51 @@
...
+#include <shogun/lib/config.h>
+#include <vector>
+#include <shogun/lib/common.h>
+#include <shogun/lib/SGMatrix.h>
+#include <shogun/mathematics/eigen3.h>
+
+namespace shogun
+{
+/** Implementation class for Least Angle Regression */
+class CLeastAngleRegressionImpl
```
lisitsyn: Why to even have a class with the only static method?

Saurabh7: Going to add more.
Wanted to make them public and unit-test.
Can change it back later too.

src/shogun/regression/LeastAngleRegression.cpp
```diff
@@ -127,6 +131,7 @@ bool CLeastAngleRegression::train_machine(CFeatures* data)
 	fill(m_is_active.begin(), m_is_active.end(), false);
 
 	SGVector<float64_t> y = ((CRegressionLabels*) m_labels)->get_labels();
+	Map<VectorXd> map_y(y.vector, y.size());
```
karlnapf: pls dont call them map_* but rather eigen_*

src/shogun/regression/LeastAngleRegression.cpp
```diff
@@ -174,10 +180,13 @@ bool CLeastAngleRegression::train_machine(CFeatures* data)
...
 	{
+		// c_k = x_t*r_k
 		// corr = X' * (y-mu) = - X'*mu + Xy
 		copy(Xy.begin(), Xy.end(), corr.begin());
-		cblas_dgemv(CblasColMajor, CblasTrans, n_vec, n_fea, -1,
-			X.matrix, n_vec, &mu[0], 1, 1, &corr[0], 1);
+		Map<VectorXd> map_corr(&corr[0], n_fea);
+		Map<VectorXd> map_mu(&mu[0], n_vec);
+		
+		map_corr=map_Xy-(map_X.transpose()*map_mu);
```
karlnapf: nice -- much easier to read

src/shogun/regression/LeastAngleRegressionImpl.cpp
```diff
@@ -0,0 +1,83 @@
...
+	VectorXd solve = Rmat.transpose().triangularView<Lower>().solve<OnTheLeft>(col_vec);
+	SGVector<float64_t> R_k=SGVector<float64_t>(solve);
+
+	float64_t R_kk = CMath::sqrt(diag_k - solve.dot(solve));
+
+	// new_R = [R R_k; zeros(...) R_kk]
+	SGMatrix<float64_t> nR(num_active+1, num_active+1);
+	for (int32_t i=0; i < num_active; ++i)
+		for (int32_t j=0; j < num_active; ++j)
+			nR(i,j) = R(i,j);
```
karlnapf: maybe  put a newline here?

src/shogun/regression/LeastAngleRegressionImpl.cpp
```diff
@@ -0,0 +1,83 @@
...
+	SGVector<float64_t> col_k(num_active);
+	for (int32_t i=0; i < num_active; ++i)
+	{
+		// col_k[i] = X[:,i_max_corr]' * X[:,m_active_set[i]]
+		Map<VectorXd> active_i(X.get_column_vector(active_set[i]), X.num_rows);
+		col_k[i]=x_vec.dot(active_i);
+	}
+
+	// R' * R_k = (X' * X)_k = col_k, solving to get R_k
+	Map<VectorXd> col_vec(col_k.vector, col_k.size());
```
karlnapf: These are updates of Cholesky factors. There are eigen3 methods for this...

How they are embedded within pseudo code, an example is here:
https://github.com/jbowlan/pyl1ls/blob/master/lars.py

Eigen3 methods are here:
https://eigen.tuxfamily.org/dox/classEigen_1_1LLT.html#a67b49e986da81c3c90fe1793681e70ad

It would be good if those would be put into linalg (later). But definitely not do this manually ....

karlnapf:
Cool, looks good so far....
Only should use eigen3 for these cholesky updates


Can you give me an update?

src/shogun/regression/LeastAngleRegressionImpl.cpp
karlnapf: rankUpdate

Saurabh7:
```
           mlpack    scikit  mlpy    shogun 
diabetes        -  0.040068     -  0.000867 
cosExp          -  0.011672     -  0.001960 
arcene          -  0.305375     -  0.018512 
madelon         -  0.087029     -  0.051801 
mnist           -         -     -  2.312541 
```

src/shogun/regression/LeastAngleRegressionImpl.h
```diff
@@ -0,0 +1,51 @@
...
+{
+	public:
+		/**   */
```
vigsterkr: dont forget the documentation of that method

tests/unit/regression/lars_unittest.cc
```diff
@@ -228,4 +234,98 @@ TEST(LeastAngleRegression, lars_N_LessThan_D)
...
+
+	for (index_t i=0; i<6; i++)
+	{	
+		for (index_t j=0; j<4; j++)
+		{
+			mat(i,j)=CMath::random(0.0,1.0);
+			matn(i,j)=mat(i,j);
+		}
+	}
+	for (index_t i=0 ;i<6; i++)
```
vigsterkr: please use variables instead of magical constants like 6 and 4...
so lie define:
```
int32_t mat_rows = 6;
int32_t mat_cols = 6;
...
```
after that you can refer to those variables + changing the unit test from 6 to 60 would be easy :P

src/shogun/regression/LeastAngleRegression.cpp
```diff
@@ -138,14 +143,15 @@ bool CLeastAngleRegression::train_machine(CFeatures* data)
 		for (int32_t j=0; j < n_fea; ++j)
```
vigsterkr: this transposition of the input matrix should be handled by eigen or linalg... doing this with a double nested loops is soo... 101 computer science

src/shogun/regression/LeastAngleRegression.cpp
```diff
@@ -138,14 +143,15 @@ bool CLeastAngleRegression::train_machine(CFeatures* data)
...
 
 	// beta is the estimator
 	vector<float64_t> beta = make_vector(n_fea, 0);
 
 	vector<float64_t> Xy = make_vector(n_fea, 0);
+	Map<VectorXd> map_Xy(&Xy[0], n_fea);
 	// Xy = X' * y
-	cblas_dgemv(CblasColMajor, CblasTrans, n_vec, n_fea, 1, X.matrix, n_vec,
-		y.vector, 1, 0, &Xy[0], 1);
+	map_Xy=map_X.transpose()*map_y;
```
vigsterkr: so here basically we X^T^T, i.e. double transpose the original input..... so we could just use the input itself right? :)

src/shogun/regression/LeastAngleRegression.cpp
vigsterkr: that `map_X.transpose()` = Xorig (the original input)

Saurabh7:
Tried to make similar:
https://gist.github.com/Saurabh7/b07c65df21a4fe498284a8086fe617b8

src/shogun/regression/LeastAngleRegression.cpp
karlnapf: this is the reason why I want get rid of those lapack calls, they totally hide such things.

vigsterkr: but that lapack thing wasn't even doing this... it was that manual nested loop in the beginning of `train_machine` that created this :P

Saurabh7:
Ok, I have updated almost all lapack calls, also using eigen's column accessors now everywhere.
Parallelized over loops where dot products are needed (since cant slice matrices).

I tried different combination of paramters for scikit and shogun extensively. 
Also changed Shogun's early stopping criterion to `max_corr/n_vec > CMath::MACHINE_EPSILON`  (similar to scikit btw).   But Shogun still goes for the full path unnecessarily sometimes.

Saurabh7:
Changed the loops for transpose. If we want to remove it completely will have to specify row major eigen matrices since further we have to loop over features (which are cols right now) multiple times.

src/shogun/regression/LeastAngleRegression.cpp
```diff
@@ -187,7 +181,7 @@ bool CLeastAngleRegression::train_machine(CFeatures* data)
 		Map<VectorXd> map_mu(&mu[0], n_vec);
 		
 		map_corr=map_Xy-(map_X.transpose()*map_mu);
```
vigsterkr: why don't use just have 2 maps?
`map_X` and `map_Xt`

src/shogun/regression/LeastAngleRegressionImpl.cpp
```diff
@@ -40,23 +40,23 @@ using namespace Eigen;
 
 namespace shogun
 {
-SGMatrix<float64_t> CLeastAngleRegressionImpl::cholesky_insert_helper(SGMatrix<float64_t> X, SGMatrix<float64_t> R, VectorXd x_vec, SGVector<float64_t> active_set)
+SGMatrix<float64_t> CLeastAngleRegressionImpl::cholesky_insert_helper(Ref<MatrixXd> map_X, SGMatrix<float64_t>& R, int32_t i_max_corr, SGVector<float64_t> active_set)
```
vigsterkr: can't we just pass around SGVector/SGMatrix instead of going full into eigen?

Saurabh7: Then I have to transpose the matrix inside.

Saurabh7:
Ok when I avoid the transpose at the start and use `map.transpose().col(i)` , this leads to larger runtimes.
https://gist.github.com/Saurabh7/ab36961c51cd5e69213096527b4d2810

src/shogun/regression/LeastAngleRegressionImpl.h
karlnapf: We also do not want eigen in the interface.
Even if it is impl class, put a SGVector, etc
Creating Map is free

src/shogun/regression/LeastAngleRegression.cpp
karlnapf: +1

src/shogun/regression/LeastAngleRegressionImpl.cpp
karlnapf: why not create a Map inside and transpose it?

karlnapf:
Interesting.

The only reason I can think of is that this might be due to cache effects and the prefetcher.
Can you maybe check that hypothesis?
If so, then the work of the double transpose might is worth it

karlnapf:
It is important that you document these things, or even put them in the
benchmarking framework.
For later.

What are the current results?

2016-06-08 12:56 GMT+01:00 Saurabh Mahindre <notifications@github.com>:

karlnapf:
oh also, *when* does shognu do the full path unnecessarily? Would be good
to identify these cases

2016-06-08 12:56 GMT+01:00 Saurabh Mahindre <notifications@github.com>:

Saurabh7:
with current version: https://gist.github.com/Saurabh7/d2d4ed03b7433f8c33459b106bdf5375

with `map.transpose().col(i)`: https://gist.github.com/Saurabh7/da64a4fcecbc78dcf6fd8f17b5fd80a4

Saurabh7:
Shogun's epsilon was very small ` CMath::MACHINE_EPSILON` which lead to continuing the iters even when max corelation was very low.
Scikit uses eps as `np.finfo(np.float32).eps` which comes out to be 1.1920929e-07 here.
I used this for now.

some values
 ```
  N     D     SG time    SK time       SG mse       SK mse    SG iters    SK iters
----  ----  ----------  ---------  -----------  -----------  ----------  ----------
2000   100  0.00801992  0.0135689  9.25081e+07  9.25081e+07          35          33
2000   500  0.339843    0.240674   2.52726e+09  2.52726e+09         167         165
2000  1000  1.5217      7.07059    8.8488e+09   9.34494e+09         352        1024
5000  1000  3.13794     2.2291     2.3638e+10   2.3638e+10          333         331
5000  2000  18.5837     68.6767    1.04143e+11  1.05728e+11         679        2026
```

File: src/shogun/regression/LeastAngleRegressionImpl.cpp (hunk 11 above)
Comment (Saurabh7) [inline src/shogun/regression/LeastAngleRegressionImpl.cpp]: Leads to additional overhead if I save the transpose. If I use `map.transpose().col(i)` too many cache misses. So options might be pass Eigen things or manually tranpose SGMatrix initally or any suggestions ?

Comment (Saurabh7) [inline src/shogun/regression/LeastAngleRegressionImpl.cpp]: Or mayb use directly, no helper method.

Comment (vigsterkr) [inline src/shogun/regression/LeastAngleRegressionImpl.cpp]: @Saurabh7 maybe this is the reason why in the initial implementation there was that transposition of the input matrix. but it'd be good to actually use a more optimal solution for matrix transposition than nested loops.
all lapack/blas functions has `TRANS` function arg that will let you do the operation on the original or the transposed version of the matrix. would be interesting to see the cache misses there ... :)

**Comment by vigsterkr** [comment]
@Saurabh7 thnx for the benchmarks... any idea(s) why that diff between SG and SK in case of `N=5000` and `D=1000`? or that's the cost of that two extra iteration?

File: src/shogun/regression/LeastAngleRegressionImpl.cpp (hunk 11 above)
Comment (Saurabh7) [inline src/shogun/regression/LeastAngleRegressionImpl.cpp]: Develop version: [output](https://gist.github.com/Saurabh7/65c920f3813908358ded0c4ffde5b38b)
where lapack routines are used.
Other than that we need the transpose to extract columns and take ddots

**Comment by Saurabh7** [comment]
OK heres comparison of latest version for different number of relevant features:
(setting max iters for scikit at some places as it fails to stop)

NUM FEATS * 0.8 : https://gist.github.com/Saurabh7/f0258a66b331728ce995bc96d0f528f7
NUM FEATS* 0.5 : https://gist.github.com/Saurabh7/5ebd29f17313a335ced279db0a999b1e
NUM FEATS* 0.3: https://gist.github.com/Saurabh7/45b2a1ee0524cdda88e6af405c849e3e

File: src/shogun/regression/LeastAngleRegression.cpp [hunk 12]
```diff
@@ -127,25 +130,26 @@ bool CLeastAngleRegression::train_machine(CFeatures* data)
...
-	{
-		for (int32_t j=0; j < n_fea; ++j)
-			X(i,j) = Xorig(j,i);
-	}
+	SGMatrix<float64_t> X (n_vec, n_fea);
+	Map<MatrixXd> map_Xr(Xorig.matrix, n_fea, n_vec);
+	Map<MatrixXd> map_X(X.matrix, n_vec, n_fea);	
+   	map_X = map_Xr.transpose();
+	
+	float64_t* X_active = SG_MALLOC(float64_t, n_vec*n_fea);
```
vigsterkr: this sounds like a `SGVector<float64_t>` no?

src/shogun/regression/LeastAngleRegression.cpp
vigsterkr: shouldn't this be then `map_Xr*map_y`?

src/shogun/regression/LeastAngleRegression.cpp
```diff
@@ -172,80 +176,86 @@ bool CLeastAngleRegression::train_machine(CFeatures* data)
 	// main loop
 	//========================================
 	int32_t nloop=0;
-	while (m_num_active < max_active_allowed && max_corr > CMath::MACHINE_EPSILON && !stop_cond)
+	while (m_num_active < max_active_allowed && max_corr/n_vec > 1.192E-07 && !stop_cond)
```
vigsterkr: mmm use a macro or a static const for magical constant values

src/shogun/regression/LeastAngleRegression.cpp
```diff
@@ -172,80 +176,86 @@ bool CLeastAngleRegression::train_machine(CFeatures* data)
...
 	{
 		// corr = X' * (y-mu) = - X'*mu + Xy
-		copy(Xy.begin(), Xy.end(), corr.begin());
-		cblas_dgemv(CblasColMajor, CblasTrans, n_vec, n_fea, -1,
-			X.matrix, n_vec, &mu[0], 1, 1, &corr[0], 1);
-
+		Map<VectorXd> map_corr(&corr[0], n_fea);
+		Map<VectorXd> map_mu(&mu[0], n_vec);
+		
+		map_corr = map_Xy - (map_X.transpose()*map_mu);
```
vigsterkr: again why not to use `map_Xr*map_mu` here instead of `map_X.transpose()*map_mu`?

src/shogun/regression/LeastAngleRegression.h
```diff
@@ -16,6 +16,7 @@
 #ifdef HAVE_LAPACK
 #include <vector>
 #include <shogun/machine/LinearMachine.h>
+#include <shogun/mathematics/eigen3.h>
```
vigsterkr: do we really need this?

src/shogun/regression/LeastAngleRegression.cpp
karlnapf: this should be a parameter

src/shogun/regression/LeastAngleRegression.cpp
```diff
@@ -172,80 +176,86 @@ bool CLeastAngleRegression::train_machine(CFeatures* data)
...
+		Map<VectorXd> map_u(&u[0], n_vec);
+		
+		map_u=map_Xa*wA;
 
-		// step size
 		float64_t gamma = max_corr / AA;
 		if (m_num_active < n_fea)
 		{
+			#pragma omp parallel for shared(gamma)
 			for (int32_t i=0; i < n_fea; ++i)
```
karlnapf: this parallel helps?

src/shogun/regression/LeastAngleRegressionImpl.h
vigsterkr: doxygen is still missing

src/shogun/regression/LeastAngleRegression.cpp
karlnapf: yes

src/shogun/regression/LeastAngleRegression.cpp
```diff
@@ -127,25 +130,26 @@ bool CLeastAngleRegression::train_machine(CFeatures* data)
...
-	SGMatrix<float64_t> X(n_vec, n_fea, true);
-	for (int32_t i=0; i < n_vec; ++i)
-	{
-		for (int32_t j=0; j < n_fea; ++j)
-			X(i,j) = Xorig(j,i);
-	}
+	SGMatrix<float64_t> X (n_vec, n_fea);
+	Map<MatrixXd> map_Xr(Xorig.matrix, n_fea, n_vec);
+	Map<MatrixXd> map_X(X.matrix, n_vec, n_fea);	
+   	map_X = map_Xr.transpose();
```
karlnapf: you you either use spaces or tab stops (minor)

src/shogun/regression/LeastAngleRegression.cpp
```diff
@@ -127,25 +130,26 @@ bool CLeastAngleRegression::train_machine(CFeatures* data)
 	fill(m_is_active.begin(), m_is_active.end(), false);
 
 	SGVector<float64_t> y = ((CRegressionLabels*) m_labels)->get_labels();
+	Map<VectorXd> map_y(y.vector, y.size());
 	SGMatrix<float64_t> Xorig = feats->get_feature_matrix();
```
vigsterkr: any reason for example why LARS shouldn't be able to work on `SGMatrix<float32_t>` ?
but yeah this should be addressed in a consequent PR ;)

src/shogun/regression/LeastAngleRegression.cpp
```diff
@@ -127,25 +130,26 @@ bool CLeastAngleRegression::train_machine(CFeatures* data)
...
-	}
+	SGMatrix<float64_t> X (n_vec, n_fea);
+	Map<MatrixXd> map_Xr(Xorig.matrix, n_fea, n_vec);
+	Map<MatrixXd> map_X(X.matrix, n_vec, n_fea);	
+   	map_X = map_Xr.transpose();
+	
+	float64_t* X_active = SG_MALLOC(float64_t, n_vec*n_fea);
 
 	// beta is the estimator
 	vector<float64_t> beta = make_vector(n_fea, 0);
```
karlnapf: why using vector here?

src/shogun/regression/LeastAngleRegressionImpl.cpp
```diff
@@ -0,0 +1,69 @@
...
+ * The views and conclusions contained in the software and documentation are those
+ * of the authors and should not be interpreted as representing official policies,
+ * either expressed or implied, of the Shogun Development Team.
+ *
+ */
+#include <shogun/lib/config.h>
+
+#include <shogun/mathematics/Math.h>
+#include <shogun/regression/LeastAngleRegressionImpl.h>
+#include <shogun/mathematics/eigen3.h>
```
karlnapf: this line has to go

src/shogun/regression/LeastAngleRegressionImpl.cpp
```diff
@@ -0,0 +1,69 @@
...
+
+	// R' * R_k = (X' * X)_k = col_k, solving to get R_k	
+	map_R.transpose().triangularView<Lower>().solveInPlace<OnTheLeft>(R_k);
+	float64_t R_kk = CMath::sqrt(diag_k - R_k.dot(R_k));
+
+	SGMatrix<float64_t> R_new(num_active+1, num_active+1);
+	Map<MatrixXd> map_R_new(R_new.matrix, R_new.num_rows, R_new.num_cols);
+
+	map_R_new.block(0, 0, num_active, num_active) = map_R;
+	for (int32_t i=0; i < num_active; ++i)
```
karlnapf: you can do this with memset

src/shogun/regression/LeastAngleRegressionImpl.cpp
```diff
@@ -0,0 +1,69 @@
...
+		map_R_new(i, num_active) = R_k[i];
+	for (int32_t i=0; i < num_active; ++i)
```
karlnapf: same here

src/shogun/regression/LeastAngleRegressionImpl.cpp
```diff
@@ -0,0 +1,69 @@
...
+#include <shogun/regression/LeastAngleRegressionImpl.h>
+#include <shogun/mathematics/eigen3.h>
+
+using namespace std;
+using namespace shogun;
+using namespace Eigen;
+
+namespace shogun
+{
+SGMatrix<float64_t> CLeastAngleRegressionImpl::cholesky_insert_helper(SGMatrix<float64_t>& X, float64_t* X_active, SGMatrix<float64_t>& R, int32_t i_max_corr , int32_t num_active)
```
karlnapf: some of the matrices can be const references

src/shogun/regression/LeastAngleRegressionImpl.h
```diff
@@ -0,0 +1,51 @@
...
+ */
+
+#ifndef _LARSIMPL_H_
+#define _LARSIMPL_H_
+
+#include <shogun/lib/config.h>
+#include <vector>
+#include <shogun/lib/common.h>
+#include <shogun/lib/SGMatrix.h>
+#include <shogun/mathematics/eigen3.h>
```
karlnapf: please no eigen3 includes in .h files (even if this is impl)

src/shogun/regression/LeastAngleRegressionImpl.h
```diff
@@ -0,0 +1,51 @@
...
+ * of the authors and should not be interpreted as representing official policies,
+ * either expressed or implied, of the Shogun Development Team.
+ *
+ */
+
+#ifndef _LARSIMPL_H_
+#define _LARSIMPL_H_
+
+#include <shogun/lib/config.h>
+#include <vector>
```
karlnapf: this one should go as well

src/shogun/regression/LeastAngleRegression.cpp
Saurabh7: Havent checked exactly how much it helps. Its just parallelizing the dot products.

Saurabh7:
Updated and squashed .

vigsterkr:
well, it's time to merge! :)

tests/unit/regression/lars_unittest.cc
```diff
@@ -228,4 +233,113 @@ TEST(LeastAngleRegression, lars_N_LessThan_D)
 	SG_UNREF(features);
 	SG_UNREF(labels);
 }
+
+TEST(LeastAngleRegression, cholesky_insert)
+{
+	class lars_helper: public CLeastAngleRegression
```
karlnapf: usually, these are called "mock" or so, but ok

karlnapf:
Great that this is merged :)

karlnapf:
Now it is time for the final benchmarks! Looking forward to those

PR #3247, opened 2016-06-06
Linalg benchmark

OXPHOS:
- compare CPU.dot and GPU.dot of current linalg and refactored linalg
- ref #3230 
- cannot use global variable: `sg_linalg` (segmentation fault 11)

### DIFF
--- a/benchmarks/linalg_refactor_benchmark.cpp
+++ b/benchmarks/linalg_refactor_benchmark.cpp
@@ -1,0 +1,149 @@
+#include <shogun/base/init.h>
+#include <shogun/lib/SGVector.h>
+#include <shogun/mathematics/linalgrefactor/linalgRefactor.h>
+
+#include <shogun/mathematics/eigen3.h>
+#include <viennacl/linalg/inner_prod.hpp>
+#include <viennacl/vector.hpp>
+
+#include <algorithm>
+#include <memory>
+#include <hayai/hayai.hpp>
+
+#include <iostream>
+
+using namespace shogun;
+
+/**
+ * Instructions :
+ * 1. Install benchmarking toolkit "hayai" (https://github.com/nickbruun/hayai)
+ * 2. Compile against libhayai_main, e.g.
+ * 		g++ -O3 -std=c++11 linglg_refactor_benchmark.cpp -I/usr/include/eigen3 \
+ *		-lshogun -lhayai_main -lOpenCL -o benchmark
+ * 3. ./benchmark
+ */
+
+/** Generate data only once */
+typedef float32_t T;
+typedef viennacl::vector_base<T, std::size_t, std::ptrdiff_t> VCLVectorBase;
+
+template<typename value_type>
+struct Data
+{
+	typedef viennacl::backend::mem_handle VCLMemoryArray;
+
+	Data()
+	{
+		num_rows = 100;
+		init();
+	}
+
+	Data(index_t num_rows)
+	{
+		this->num_rows = num_rows;
+		init();
+	}
+
+	~Data()
+	{
+		exit_shogun();
+	}
+
+	void init()
+	{
+		A = init_sg(1);
+		B = init_sg(2);
+		Av = init_v(A);
+		Bv = init_v(B);
+		Ac = std::unique_ptr<BaseVector<value_type>> (init_c(A));
+		Bc = std::unique_ptr<BaseVector<value_type>> (init_c(B));
+		Ag = std::unique_ptr<BaseVector<value_type>> (init_g(A));
+		Bg = std::unique_ptr<BaseVector<value_type>> (init_g(B));
+
+		init_shogun_with_defaults();
+		GPUBackend viennaclBackend;
+		sg_linalg->set_gpu_backend(&viennaclBackend);
+	}
+
+	/** SGVector **/
+	SGVector<value_type> init_sg(value_type begin)
+	{
+		SGVector<value_type> m(num_rows);
+		m.range_fill(begin);
+
+		return m;
+	}
+
+	/** ViennaCL Vector for test **/
+	VCLVectorBase init_v(SGVector<value_type> m)
+	{
+		VCLVectorBase mv;
+		std::shared_ptr<VCLMemoryArray> vector(new VCLMemoryArray());
+		viennacl::backend::memory_create(*vector, sizeof(value_type)*num_rows,
+			viennacl::context());
+		viennacl::backend::memory_write(*vector, 0, num_rows*sizeof(value_type),
+			m.vector);
+		mv = VCLVectorBase(*vector, num_rows, 0, 1);
+
+		return mv;
+	}
+
+	/** CPUVector derived from BaseVector **/
+	std::unique_ptr<BaseVector<value_type>> init_c(SGVector<value_type> m)
+	{
+		std::unique_ptr<CPUVector<value_type>> mc(new CPUVector<value_type>(m));
+		return std::move(mc);
+	}
+
+	/** GPUVector derived from BaseVector **/
+	std::unique_ptr<BaseVector<value_type>> init_g(SGVector<value_type> m)
+	{
+		std::unique_ptr<GPU_Vector<value_type>> mg(new GPU_Vector<value_type>(m));
+		return std::move(mg);
+	}
+
+	SGVector<value_type> A;
+	SGVector<value_type> B;
+	VCLVectorBase Av;
+	VCLVectorBase Bv;
+	std::unique_ptr<BaseVector<value_type>> Ac;
+	std::unique_ptr<BaseVector<value_type>> Bc;
+	std::unique_ptr<BaseVector<value_type>> Ag;
+	std::unique_ptr<BaseVector<value_type>> Bg;
+
+	index_t num_rows;
+};
+
+BENCHMARK_P(CPUVector, dot_explict_eigen3, 10, 1000,
+	(const SGVector<T> &A, const SGVector<T> &B))
+{
+	typedef Eigen::Matrix<T, Eigen::Dynamic, 1> VectorXt;
+	Eigen::Map<VectorXt> vec_A(A.vector, A.vlen);
+	Eigen::Map<VectorXt> vec_B(B.vector, B.vlen);
+
+    	auto C = vec_A.dot(vec_B);
+}
+
+BENCHMARK_P(CPUVector, dot_eigen3, 10, 1000,
+	(BaseVector<T> *A, BaseVector<T> *B))
+{
+	auto C = sg_linalg->dot(A, B);
+}
+
+BENCHMARK_P(GPU_Vector, dot_explict_viennacl, 10, 1000,
+	(const VCLVectorBase &A, const VCLVectorBase &B))
+{
+	auto C = viennacl::linalg::inner_prod(A, B);
+}
+
+BENCHMARK_P(GPU_Vector, dot_viennacl, 10, 1000,
+	(BaseVector<T> *A, BaseVector<T> *B))
+{
+	auto C = sg_linalg->dot(A, B);
+}
+
+Data<T> data(1000000);
+BENCHMARK_P_INSTANCE(CPUVector, dot_explict_eigen3, (data.A, data.B));
+BENCHMARK_P_INSTANCE(CPUVector, dot_eigen3, (data.Ac.get(), data.Bc.get()));
+BENCHMARK_P_INSTANCE(GPU_Vector, dot_explict_viennacl, (data.Av, data.Bv));
+BENCHMARK_P_INSTANCE(GPU_Vector, dot_viennacl, (data.Ag.get(), data.Bg.get()));


OXPHOS:
I can only do:
`clang++ -O3 -std=c++11 ../benchmarks/linalg_benchmark.cpp -I/usr/include/eigen3 -lshogun -lhayai_main -framework OpenCL -o benchmark`

g++ and `-lOpenCL` didn't work for me.

For `-lOpenCL`, the error was: cannot find OpenCL.
For g++ and `-framework OpenCL`, I got error: https://gist.github.com/OXPHOS/dc59556267bc24614ebceb0af9ba8218

lambday:
Hi @OXPHOS 

The point of the benchmark is to compare the performance between direct library function call and shogun linalg method call, NOT between CPU and GPU backend. The two benchmarks should be something like

- create SGVector and CPUVector from it. Now compare between (a) creating Eigen3 map directly and calling eigen3 dot (b) calling shogun linalg dot on CPUVector that uses Eigen3 backend.
- create SGVector and GPUVector from it. Now compare between (a) calling viennacl inner_prod on the viennacl vector (b) calling shogun linalg dot on GPUVector.

Run this experiment with vectors of size 1000, 10000, 100000 and so on. We want to check whether using our version of dot has any overhead, and if yes, how much. See the point?

There might be an issue with your OpenCL installation. Should not be too hard to fix it.

benchmarks/linalg_benchmark.cpp
```diff
@@ -0,0 +1,71 @@
...
+		B = init(2);
+		Av = init(A);
+		Bv = init(B);
+	}
+
+	SGVector<int32_t> init(int32_t begin)
+	{
+		index_t n = num_rows;
+		std::vector<int32_t> mem(n);
+		std::iota(mem.data(), mem.data() + n, begin);
```
lambday: Why create `std::vector` first?

benchmarks/linalg_benchmark.cpp
```diff
@@ -0,0 +1,71 @@
...
+	CGPUVector<int32_t> Av;
+	CGPUVector<int32_t> Bv;
+
+	static constexpr index_t num_rows = 100;
+};
+
+BENCHMARK_P(SGVector, dot_product_eigen3, 10, 1000,
+		(SGVector<int32_t> A, SGVector<int32_t> B))
+{
+	auto C = linalg::dot<linalg::Backend::EIGEN3>(A, B);
```
lambday: We need to benchmark the new code that you added, not the old code - we're gonna trash that anyway. That's the plan :)

lambday: Oh I see. You wanted to compare between the old code and the new one. I think this is not required.

benchmarks/linalg_refactor_benchmark.cpp
```diff
@@ -0,0 +1,94 @@
...
+		Bc = init_c(B);
+        Ag = init_g(A);//GPU_Vector<int32_t>(A);
+        Bg = GPU_Vector<int32_t>(B);
+	}
+
+	SGVector<int32_t> init(int32_t begin)
+	{
+
+        index_t n = num_rows;
+        std::vector<int32_t> mem(n);
```
lambday: Why `std::vector` and then `memcpy`? You can directly do that with `SGVector`.

benchmarks/linalg_refactor_benchmark.cpp
```diff
@@ -0,0 +1,94 @@
...
+};
+
+BENCHMARK_P(CPUVector, dot_int32_eigen3, 10, 1000,
+		(CPUVector<int32_t> A, CPUVector<int32_t> B))
+{
+    CPUBackend eigenBackend;
+    //sg_linalg->cpubackend = &eigenBackend;
+
+    Linalg linalg(&eigenBackend);
+    auto C = linalg.dot<int32_t>(&A, &B);
```
lambday: We **have** to make the global `sg_linalg` work :)

benchmarks/linalg_refactor_benchmark.cpp
OXPHOS: I'll checkout. thx!
BTW there's a similar step in max_product_benchmark, but there're some bugs..It copies an empty vector to SGMatrix and returned a deleted SGMatrix. Not sure whether you want to fix it or I should.

lambday: Oh I see. Don't bother for now. We're excited about this benchmark for now :D

OXPHOS:
(not sure where to put this)
graph:
https://drive.google.com/file/d/0B6YkOSoC_E5raktfWFRweXgwa1U/view
vector size from 10 - 10^7,
left panel compares Eigen3 dot product with explicit method(blue) and CPUVector wrapper method(black).
right panel compares ViennaCL dot product with explicit method(blue) and GPU_Vector wrapper method(black).
GPU_Vector - GPUBackend is extremely slow..

lambday:
Hi @OXPHOS!

Great job with the benchmarks :) This is exactly what we were looking for. Interesting issues will be visible from these benchmarks.

Let me comment on the code first and then we'll discuss

benchmarks/linalg_refactor_benchmark.cpp
```diff
@@ -0,0 +1,151 @@
...
+
+	Data()
+	{
+		num_rows = 100;
+		A = init(1);
+		B = init(2);
+		Ac = init_c(A);
+		Bc = init_c(B);
+        Ag = init_g(A);
+        Bg = init_g(B);
```
lambday: indentation :)

benchmarks/linalg_refactor_benchmark.cpp
```diff
@@ -0,0 +1,151 @@
...
+		A = init(1);
+		B = init(2);
+		Ac = init_c(A);
+		Bc = init_c(B);
+		Ag = init_g(A);
+		Bg = init_g(B);
+		Av = init_v(A);
+		Bv = init_v(B);
+		init_shogun_with_defaults();
+	}
```
lambday: You could have just created a `init()` method that does all those stuff, and then could have called that method from both the constructor. Anyway, not a big deal

benchmarks/linalg_refactor_benchmark.cpp
```diff
@@ -0,0 +1,151 @@
...
+
+	SGVector<int32_t> init(int32_t begin)
+	{
+		SGVector<int32_t> m(num_rows);
+		m.range_fill(begin);
+
+		return m;
+	}
+
+    CPUVector<int32_t> init_c(SGVector<int32_t> m)
```
lambday: Indentation

benchmarks/linalg_refactor_benchmark.cpp
```diff
@@ -0,0 +1,151 @@
...
+			viennacl::context());
+		viennacl::backend::memory_write(*vector, 0, num_rows*sizeof(int32_t),
+			m.vector);
+		mv = VCLVectorBase(*vector, num_rows, 0, 1);
+		return mv;
+	}
+
+	SGVector<int32_t> A;
+	SGVector<int32_t> B;
+    CPUVector<int32_t> Ac;
```
lambday: Indentation

benchmarks/linalg_refactor_benchmark.cpp
```diff
@@ -0,0 +1,151 @@
...
+	Eigen::Map<VectorXt> vec_B(B.vector, B.vlen);
+    //Eigen::Map<VectorXt> vec_B = B;
+    auto C = vec_A.dot(vec_B);
+}
+
+BENCHMARK_P(CPUVector, dot_int32_eigen3, 10, 1000,
+		(CPUVector<int32_t> A, CPUVector<int32_t> B))
+{
+	CPUBackend eigenBackend;
+	sg_linalg->set_cpu_backend(&eigenBackend);
```
lambday: Backend should be registered in the `Data::init()` itself. Also, I think there should **always** already be a CPU-backend (Eigen). We shouldn't have to set it every time.

benchmarks/linalg_refactor_benchmark.cpp
```diff
@@ -0,0 +1,151 @@
...
+		(VCLVectorBase A, VCLVectorBase B))
+{
+    auto C = viennacl::linalg::inner_prod(A, B);
+}
+
+BENCHMARK_P(GPU_Vector, dot_int32_viennacl, 10, 1000,
+		(GPU_Vector<int32_t> A, GPU_Vector<int32_t> B))
+{
+    GPUBackend viennaclBackend;
+    sg_linalg->set_gpu_backend(&viennaclBackend);
```
lambday: Register the backend in `Data::init()`. Don't have to do it in each iteration.

benchmarks/linalg_refactor_benchmark.cpp
```diff
@@ -0,0 +1,151 @@
...
+
+    auto C = sg_linalg->dot<int32_t>(&A, &B);
```
lambday: No need to pass the template arg `int32_t`.

benchmarks/linalg_refactor_benchmark.cpp
```diff
@@ -0,0 +1,151 @@
...
+    auto C = vec_A.dot(vec_B);
+}
+
+BENCHMARK_P(CPUVector, dot_int32_eigen3, 10, 1000,
+		(CPUVector<int32_t> A, CPUVector<int32_t> B))
+{
+	CPUBackend eigenBackend;
+	sg_linalg->set_cpu_backend(&eigenBackend);
+
+    auto C = sg_linalg->dot<int32_t>(&A, &B);
```
lambday: No need to pass the template arg `int32_t`.

benchmarks/linalg_refactor_benchmark.cpp
```diff
@@ -0,0 +1,151 @@
...
+{
+	typedef Eigen::Matrix<int32_t, Eigen::Dynamic, 1> VectorXt;
+	Eigen::Map<VectorXt> vec_A(A.vector, A.vlen);
+	Eigen::Map<VectorXt> vec_B(B.vector, B.vlen);
+    //Eigen::Map<VectorXt> vec_B = B;
+    auto C = vec_A.dot(vec_B);
+}
+
+BENCHMARK_P(CPUVector, dot_int32_eigen3, 10, 1000,
+		(CPUVector<int32_t> A, CPUVector<int32_t> B))
```
lambday: Pass these as `BaseVector<int32_t>*`, as they are supposed to be used.

lambday: Maybe even `BaseVector<int32_t> const * const`.

benchmarks/linalg_refactor_benchmark.cpp
```diff
@@ -0,0 +1,151 @@
...
+}
+
+BENCHMARK_P(GPU_Vector, dot_int32_explict_viennacl, 10, 1000,
+		(VCLVectorBase A, VCLVectorBase B))
+{
+    auto C = viennacl::linalg::inner_prod(A, B);
+}
+
+BENCHMARK_P(GPU_Vector, dot_int32_viennacl, 10, 1000,
+		(GPU_Vector<int32_t> A, GPU_Vector<int32_t> B))
```
lambday: Pass it using pointers. See comment above.

benchmarks/linalg_refactor_benchmark.cpp
```diff
@@ -0,0 +1,151 @@
...
+		(CPUVector<int32_t> A, CPUVector<int32_t> B))
+{
+	CPUBackend eigenBackend;
+	sg_linalg->set_cpu_backend(&eigenBackend);
+
+    auto C = sg_linalg->dot<int32_t>(&A, &B);
+}
+
+BENCHMARK_P(GPU_Vector, dot_int32_explict_viennacl, 10, 1000,
+		(VCLVectorBase A, VCLVectorBase B))
```
lambday: Pass it using const ref.

benchmarks/linalg_refactor_benchmark.cpp
```diff
@@ -0,0 +1,151 @@
...
+};
+
+BENCHMARK_P(CPUVector, dot_int32_explict_eigen3, 10, 1000,
+		(SGVector<int32_t> A, SGVector<int32_t> B))
+{
+	typedef Eigen::Matrix<int32_t, Eigen::Dynamic, 1> VectorXt;
+	Eigen::Map<VectorXt> vec_A(A.vector, A.vlen);
+	Eigen::Map<VectorXt> vec_B(B.vector, B.vlen);
+    //Eigen::Map<VectorXt> vec_B = B;
+    auto C = vec_A.dot(vec_B);
```
lambday: Indentation

benchmarks/linalg_refactor_benchmark.cpp
```diff
@@ -0,0 +1,151 @@
...
+		return mv;
+	}
+
+	SGVector<int32_t> A;
+	SGVector<int32_t> B;
+    CPUVector<int32_t> Ac;
+    CPUVector<int32_t> Bc;
+    GPU_Vector<int32_t> Ag;
+    GPU_Vector<int32_t> Bg;
+	VCLVectorBase Av;
```
lambday: Indentation

benchmarks/linalg_refactor_benchmark.cpp
```diff
@@ -0,0 +1,151 @@
...
+ * 2. Compile against libhayai_main, e.g.
+ * 		g++ -O3 -std=c++11 linglg_refactor_benchmark.cpp -I/usr/include/eigen3 \
+ *		-lshogun -lhayai_main -lOpenCL -o benchmark
+ * 3. ./benchmark
+ */
+
+/** Generate data only once */
+typedef viennacl::vector_base<int32_t, std::size_t, std::ptrdiff_t> VCLVectorBase;
+
+struct Data
```
lambday: If you create this as a template, then we can easily check with the tests with different data-types. For example, the real use cases would mostly be `float64_t` and `float32_t`. can you please do that?

lambday:
Hey there! Please check the comments. We need to make sure that no extra copying is happening under the hood. The `dot` interface we designed is supposed to work with base vector pointers. So that's exactly what the test-cases should deal with. Maybe changing those to pointers have an effect on the result?

Good job by the way! Also, please check the [coding style](https://github.com/shogun-toolbox/shogun/wiki/Code-style) for shogun. It is better to follow a uniform style everywhere.

Let me know when you fix the issues. Would be interesting to see how it performs in other machines.

benchmarks/linalg_refactor_benchmark.cpp
```diff
@@ -0,0 +1,151 @@
...
+		viennacl::backend::memory_write(*vector, 0, num_rows*sizeof(value_type),
+			m.vector);
+		mv = VCLVectorBase(*vector, num_rows, 0, 1);
+
+		return mv;
+	}
+
+	SGVector<value_type> A;
+	SGVector<value_type> B;
+	CPUVector<value_type> Ac;
```
lambday: Just have a `BaseVector<value_type>* Ac` here. Same for others. Should be allocated in the heap.

benchmarks/linalg_refactor_benchmark.cpp
```diff
@@ -0,0 +1,151 @@
...
+{
+	auto C = sg_linalg->dot(A, B);
+}
+
+//index_t vectorsize[] = {10, 100, 1000, 10000, 100000};
+//for (index_t i = 0; i < 5; ++i) {}
+
+Data<T> data(10);
+BENCHMARK_P_INSTANCE(CPUVector, dot_explict_eigen3, (data.A, data.B));
+BENCHMARK_P_INSTANCE(CPUVector, dot_eigen3, (&data.Ac, &data.Bc));
```
lambday: We can directly pass `data.Ac` here then. This is how we'll use it in shogun code.

benchmarks/linalg_refactor_benchmark.cpp
```diff
@@ -0,0 +1,158 @@
...
+		m.range_fill(begin);
+
+		return m;
+	}
+
+	/** CPUVector derived from BaseVector **/
+	std::unique_ptr<BaseVector<value_type>> init_c(SGVector<value_type> m)
+	{
+			//CPUVector<value_type> mc(m);
+			std::unique_ptr<CPUVector<value_type>> mc(new CPUVector<value_type>(m));
```
lambday: indendation went wrong again

benchmarks/linalg_refactor_benchmark.cpp
```diff
@@ -0,0 +1,158 @@
...
+        	return std::move(mc);
+	}
+
+	/** GPUVector derived from BaseVector **/
+	GPU_Vector<value_type> init_g(SGVector<value_type> m)
+	{
+        	GPU_Vector<value_type> mg(m);
```
lambday: Make the GPUVector live on the heap as well

lambday: Make the GPUVector live on the heap as well

lambday:
benchmark code, travis check not needed. merging.

PR #3249, opened 2016-06-06
Added variance & rowwise variance (for matricies) and vector variance

c4goldsw:
Submitting for issue #3118 .

### DIFF
--- a/src/shogun/mathematics/linalg/internal/implementation/Variance.h
+++ b/src/shogun/mathematics/linalg/internal/implementation/Variance.h
@@ -96,14 +96,12 @@ struct variance<Backend::EIGEN3, Matrix>
 		REQUIRE(x.num_rows > 0, "Please ensure that m has more than 0 rows.\n")
 		REQUIRE(x.num_cols > 0, "Please ensure that m has more than %d columns.\n")
 
-		SGMatrix<T> squaredResult(x.num_rows, x.num_cols);
 		Eigen::Map<MatrixXt> eigX = x;
-		Eigen::Map<MatrixXt> eigSquaredResult = squaredResult;
+		MatrixXt eigSquaredResult(x.num_rows, x.num_cols);
 
 		T meanVal = mean<Backend::EIGEN3, SGMatrix<T>>::compute(x);		
 		eigSquaredResult.fill(meanVal);
-		eigSquaredResult = eigX - eigSquaredResult;
-		eigSquaredResult = eigSquaredResult.array() * eigSquaredResult.array();
+		eigSquaredResult = (eigX - eigSquaredResult).array().square();
 
 		return ((T) 1 / (x.num_rows*x.num_cols - 1)) * eigSquaredResult.sum();
 	}
@@ -155,15 +153,12 @@ struct vector_variance<Backend::EIGEN3, Vector>
 	{
 		REQUIRE(x.vlen>1, "Please ensure that vector length is greater than 1.\n")
 
-		SGVector<T> squaredResult(x.vlen);
 		Eigen::Map<VectorXt> eigX = x;
-		Eigen::Map<VectorXt> eigSquaredResult = squaredResult;
+		VectorXt eigSquaredResult(x.vlen);
 
 		T meanVal = mean<Backend::EIGEN3, SGVector<T>>::compute(x);
 		eigSquaredResult.fill(meanVal);
-		eigSquaredResult = eigX - eigSquaredResult;
-		eigSquaredResult = eigSquaredResult.array() * eigSquaredResult.array();
-
+		eigSquaredResult = (eigX - eigSquaredResult).array().square();
 		return ((T) 1 / (x.vlen - 1)) * eigSquaredResult.sum();
 	}
 };
@@ -290,21 +285,19 @@ struct rowwise_variance<Backend::EIGEN3, Matrix>
 		REQUIRE(x.num_cols > 0, "Please ensure that x has more than 0 columns.\n")
 
 		SGVector<T> tempVec; //used to store multiple results
-		SGMatrix<T> squaredResult(x.num_rows, x.num_cols);
 		Eigen::Map<MatrixXt> eigX = x;
-		Eigen::Map<MatrixXt> eigSquaredResult = squaredResult;
+		MatrixXt eigSquaredResult(x.num_rows, x.num_cols);
 
 		tempVec = rowwise_mean<Backend::EIGEN3, SGVector<T>>::compute(x, false);
-		Eigen::Map<VectorXt> eigMean = tempVec;
+		Eigen::Map<VectorXt> eigTempVec = tempVec;
 		for(int i = 0; i < x.num_cols; ++i)
 		{
-			eigSquaredResult.col(i) = eigMean;
+			eigSquaredResult.col(i) = eigTempVec;
 		}
-		eigSquaredResult = eigX - eigSquaredResult;
-		eigSquaredResult = eigSquaredResult.array() * eigSquaredResult.array();
+		eigSquaredResult = (eigX - eigSquaredResult).array().square();
 
-		tempVec = rowwise_sum<Backend::EIGEN3, SGMatrix<T>>::compute(squaredResult, false);
-		tempVec.scale((T) 1 / (x.num_cols -1));
+		eigTempVec = eigSquaredResult.rowwise().sum();
+		eigTempVec *= ((T) 1 / (x.num_cols -1));
 		return  tempVec;
 	}
 };
--- a/src/shogun/mathematics/linalg/internal/implementation/Variance.h
+++ b/src/shogun/mathematics/linalg/internal/implementation/Variance.h
@@ -1,0 +1,319 @@
+/*
+ * Copyright (c) The Shogun Machine Learning Toolbox
+ * Written (w) 2016 Chris Goldsworthy
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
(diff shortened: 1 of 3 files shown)


src/shogun/mathematics/linalg/internal/implementation/Variance.h
```diff
@@ -0,0 +1,319 @@
...
+		SGVector<T> tempVec; //used to store multiple results
+		SGMatrix<T> squaredResult(x.num_rows, x.num_cols);
+		Eigen::Map<MatrixXt> eigX = x;
+		Eigen::Map<MatrixXt> eigSquaredResult = squaredResult;
+
+		tempVec = rowwise_mean<Backend::EIGEN3, SGVector<T>>::compute(x, false);
+		Eigen::Map<VectorXt> eigMean = tempVec;
+		for(int i = 0; i < x.num_cols; ++i)
+		{
+			eigSquaredResult.col(i) = eigMean;
```
c4goldsw: I don't know if there's a better way to do this - I'm trying to make all of the columns of eigMean be equal to tempVec (which holds the rowwise mean for each row)

c4goldsw:
It's also worth nothing that I haven't implemented a columnwise computation as there's no columnwise mean yet - I can do that in a later PR.

src/shogun/mathematics/linalg/internal/implementation/Variance.h
```diff
@@ -0,0 +1,319 @@
...
+		REQUIRE(x.num_rows > 0, "Please ensure that m has more than 0 rows.\n")
+		REQUIRE(x.num_cols > 0, "Please ensure that m has more than %d columns.\n")
+
+		SGMatrix<T> squaredResult(x.num_rows, x.num_cols);
+		Eigen::Map<MatrixXt> eigX = x;
+		Eigen::Map<MatrixXt> eigSquaredResult = squaredResult;
+
+		T meanVal = mean<Backend::EIGEN3, SGMatrix<T>>::compute(x);		
+		eigSquaredResult.fill(meanVal);
+		eigSquaredResult = eigX - eigSquaredResult;
```
lambday: This line could be problematic. Did you make sure that it works?

Also, since this squared result thing is not being returned by the API, it can just be a Eigen3 matrix that you use internally for this method. As long as the API doesn't take or return a non SG type, we're good. See what I mean?

src/shogun/mathematics/linalg/internal/implementation/Variance.h
```diff
@@ -0,0 +1,319 @@
...
+		eigSquaredResult = eigSquaredResult.array() * eigSquaredResult.array();
```
lambday: Can you use `eigSquaredResult.array().square()` ? I think it does the same.

src/shogun/mathematics/linalg/internal/implementation/Variance.h
```diff
@@ -0,0 +1,319 @@
...
+	{
+		REQUIRE(x.vlen>1, "Please ensure that vector length is greater than 1.\n")
+
+		SGVector<T> squaredResult(x.vlen);
+		Eigen::Map<VectorXt> eigX = x;
+		Eigen::Map<VectorXt> eigSquaredResult = squaredResult;
+
+		T meanVal = mean<Backend::EIGEN3, SGVector<T>>::compute(x);
+		eigSquaredResult.fill(meanVal);
+		eigSquaredResult = eigX - eigSquaredResult;
```
lambday: Same as above.

src/shogun/mathematics/linalg/internal/implementation/Variance.h
```diff
@@ -0,0 +1,319 @@
...
+		eigSquaredResult = eigSquaredResult.array() * eigSquaredResult.array();
```
lambday: Same as above.

src/shogun/mathematics/linalg/internal/implementation/Variance.h
```diff
@@ -0,0 +1,319 @@
...
+	 * Computes the variance for each column of a matrix
+	 *
+	 * @param m matrix of values
+	 * @return the column-wise variance of a matrix
+	 */
+	static ReturnType compute(SGMatrix<T> m)
+	{
+		REQUIRE(m.num_rows > 0, "Please ensure that m has more than 0 rows.\n")
+		REQUIRE(m.num_cols > 0, "Please ensure that m has more than %d columns.\n")
+
```
lambday: Impl?

src/shogun/mathematics/linalg/internal/implementation/Variance.h
c4goldsw: It works on some small unit test - what problem could arise?

In regards the squaredResult matrix, are you saying I don't need it?  That makes sense.  But what should I replace it with?  I thought all Maps have to be initialised with something anyway.

src/shogun/mathematics/linalg/internal/implementation/Variance.h
c4goldsw: I had mentioned this above, but there currently is no implementation for a columnwise mean, so I'm leaving this unimplemented for now.

src/shogun/mathematics/linalg/internal/implementation/Variance.h
lambday: Maybe we can avoid having the temporary matrix `eigSquaredResult` altogether.

c4goldsw:
Hey, I've realised that further improvements can be made, so hold off on looking at this until I make another commit.

src/shogun/mathematics/linalg/internal/implementation/Variance.h
```diff
@@ -0,0 +1,312 @@
...
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * The views and conclusions contained in the software and documentation are those
+ * of the authors and should not be interpreted as representing official policies,
+ * either expressed or implied, of the Shogun Development Team.
+ */
+
+#ifndef SRC_SHOGUN_MATHEMATICS_LINALG_INTERNAL_IMPLEMENTATION_VARIANCE_H_
```
karlnapf: ehm.....this might be a bit long, no? :)

src/shogun/mathematics/linalg/internal/implementation/Variance.h
```diff
@@ -0,0 +1,312 @@
...
+#define SRC_SHOGUN_MATHEMATICS_LINALG_INTERNAL_IMPLEMENTATION_VARIANCE_H_
+
+#include <shogun/lib/config.h>
+#include <shogun/lib/SGVector.h>
+#include <shogun/lib/SGMatrix.h>
+#include <shogun/mathematics/linalg/internal/implementation/MeanEigen3.h>
+#include <shogun/mathematics/eigen3.h>
```
karlnapf: we never want eigen3 includes in header files, messes up things quite a bit

src/shogun/mathematics/linalg/internal/implementation/Variance.h
```diff
@@ -0,0 +1,312 @@
...
+
+#include <iostream>
```
karlnapf: also, no stdlib includes in header files.
If you do this namespace thing here, it will be set for all of shogun, which is not wanted

src/shogun/mathematics/linalg/internal/implementation/Variance.h
```diff
@@ -0,0 +1,312 @@
...
+	 * \f$\frac{1}{m*n-1}\sum_{i=1}^m\sum_{j=1}^n (x_{ij}-\bar{x})^2\f$ where
+	 * \f$\bar x=\frac{1}{mn}\sum_{i=1}^m\sum_{j=1}^n x_{ij}\f$
+	 *
+	 * @param x matrix of values
+	 * @return variance of given values
+	 */
+	static T compute(SGMatrix<T> x)
+	{
+		REQUIRE(x.num_rows > 0, "Please ensure that m has more than 0 rows.\n")
+		REQUIRE(x.num_cols > 0, "Please ensure that m has more than %d columns.\n")
```
karlnapf: this might segfault as there is nothing passed to %d

src/shogun/mathematics/linalg/internal/implementation/Variance.h
```diff
@@ -0,0 +1,312 @@
...
+	 * \f$x\f$ with entries \f$\{x_{11}, ..., x_{mn}\}\f$, this is
+	 * \f$\frac{1}{m*n-1}\sum_{i=1}^m\sum_{j=1}^n (x_{ij}-\bar{x})^2\f$ where
+	 * \f$\bar x=\frac{1}{mn}\sum_{i=1}^m\sum_{j=1}^n x_{ij}\f$
+	 *
+	 * @param x matrix of values
+	 * @return variance of given values
+	 */
+	static T compute(SGMatrix<T> x)
+	{
+		REQUIRE(x.num_rows > 0, "Please ensure that m has more than 0 rows.\n")
```
karlnapf: Should always state what was given by the user, see our wiki on error messages

src/shogun/mathematics/linalg/internal/implementation/Variance.h
```diff
@@ -0,0 +1,312 @@
...
+		REQUIRE(x.num_cols > 0, "Please ensure that m has more than %d columns.\n")
+
+		Eigen::Map<MatrixXt> eigX = x;
+		MatrixXt eigSquaredResult(x.num_rows, x.num_cols);
+
+		T meanVal = mean<Backend::EIGEN3, SGMatrix<T>>::compute(x);		
+		eigSquaredResult.fill(meanVal);
+		eigSquaredResult = (eigX - eigSquaredResult).array().square();
+
+		return ((T) 1 / (x.num_rows*x.num_cols - 1)) * eigSquaredResult.sum();
```
karlnapf: I think there should be a flag degrees of freedom that can be used to change normalisation to 1/n, 1/(n-1) , etc

tests/unit/mathematics/Statistics_unittest.cc
```diff
@@ -718,3 +719,37 @@ TEST(Statistics, vector_mean_overflow_test)
 	a.set_const(std::numeric_limits<float64_t>::max());
 	EXPECT_EQ(std::numeric_limits<float64_t>::max(), CStatistics::mean(a));
 }
+
+TEST(Statistics, variance_test)
```
karlnapf: these tests would rather be at the same place as all the other linalg tests

karlnapf:
Just saw your last comment and leaving this for now.
We are also currently refactoring linalg, which will have an impact on this thing

src/shogun/mathematics/linalg/internal/implementation/Variance.h
c4goldsw: What do you mean by that?  What will be set for all of Shogun?  Just curious.

karlnapf: the std namespace will be active everywhere you include this file (which is for example happening in the class list, so everywhere)

PR #3251, opened 2016-06-07
meta examples: fixes for java & R

karlnapf:
Aiming to fix the java meta example problems with static calls involving `Math`.

This does all imports in java explicit (cleaner anyways) and calls `Math` as `org.shogun.Math`

Running through travis to check whether old examples work. If green, anyone feel free to merge

@sanuj @OXPHOS


karlnapf:
Added a fix for static calls in R

karlnapf:
Fixed the last error, merging

PR #3252, opened 2016-06-08
linalg benchmark update

OXPHOS:
@lambday Compatible with the cleaned-up linalg (Class name, unique_ptr)

### DIFF
--- a/benchmarks/linalg_benchmark.cpp
+++ b/benchmarks/linalg_benchmark.cpp
@@ -1,6 +1,38 @@
+/*
+ * Copyright (c) 2016, Shogun-Toolbox e.V. <shogun-team@shogun-toolbox.org>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ *  1. Redistributions of source code must retain the above copyright notice,
+ *     this list of conditions and the following disclaimer.
+ *
+ *  2. Redistributions in binary form must reproduce the above copyright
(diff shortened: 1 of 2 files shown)


lambday:
Hey there!

So, as it turned out, the speed-up was due to inlining in the main caller. The correct way to benchmark this would be to compare between

- a shogun class, calling a sg_linalg method (e.g. dot) over and over again
- another shogun class, calling native eigen3 dot over and over again

We compile both these into shared libraries, and then from main, we call methods from both these classes and compare. I did that [here](https://github.com/lambday/derby/tree/master/benchmarks/eigenvslinalg).

Here are few samples runs:
```
case : large dimension, dynamic sized vector (VectorXd)
dim = 1,000,000
runs = 100
sg_linalg: 2.23658
direct eigen: 2.25466
```

```
case : small dimension, dynamic sized vector map (VectorXd)
dim = 3
runs = 10,000,000
linalg: 0.754815
eigen: 0.645714
```

```
case : small dimension, static sized vector map (Vector3d)
dim = 3
runs = 10,000,000
linalg: 0.746212
eigen: 0.597646
```

You can have a look and try running it on your machine. Good news is, ours is not too slow :)
So, maybe we'll come back to benchmarking once again, when we'll port shogun algorithms to sg_linalg. Would make sense then to compare with (a) the existing implementation with native eigen3 (b) new implementation replacing those calls with sg_linalg.

I am merging this. But let's move on to adding more methods in linalg (the existing ones first) :).

OXPHOS:
Thanks! So surprised to see the difference..so theoretically we should dump all stuff in one file..

PR #3257, opened 2016-06-08
Make DoubleMatrix type available within cookbooks.

arianepaola:
This adds the DoubleMatrix type and makes it available within cookbooks.

PR #3259, opened 2016-06-08
Make DoubleMatrix type available within cookbooks.

arianepaola:
This adds the DoubleMatrix type and makes it available within cookbooks.


vigsterkr:
just closed #3257 because it's a duplicate... let me know if you don't agree

PR #3277, opened 2016-06-09
Linalg - Vector Class

OXPHOS:


src/shogun/mathematics/linalg/SGLinalg.cpp
```diff
@@ -33,6 +33,8 @@
 
 #include <shogun/mathematics/linalg/SGLinalg.h>
 
+#include<iostream>
```
lambday: This should not be required :)

src/shogun/mathematics/linalg/SGLinalg.cpp
```diff
@@ -118,8 +118,31 @@ bool SGLinalg::hasGPUBackend() const
 	return m_gpubackend != nullptr;
 }
 
+template <class T>
+std::shared_ptr<BaseVector<T>> SGLinalg::convert_vector(const SGVector<T> &vector) const
+{
+	std::shared_ptr<BaseVector<T>> basevector;
+
+#ifdef GPU_CALC
```
lambday: Where is this defined? Maybe I am missing something here!

src/shogun/mathematics/linalg/SGLinalg.h
```diff
@@ -99,9 +98,23 @@ class SGLinalg
...
+	/** Check whether gpubackend is registered by user
+	 * @return whether gpubackend is registered by user
+	 */
 	bool hasGPUBackend() const;
 
+	/** Wrap SGVector with class CPUVector and GPUVector
+	 * @return pointer of BaseVector
+	 */
+	template <class T>
+	std::shared_ptr<BaseVector<T>> convert_vector(const SGVector<T> &vector) const;
```
lambday: Maybe call it `get_gpu_type`. It is not very clear from the name that you want to `convert` it into gpu type. Also, we want this to be overloaded so that the users just use the same name while converting both matrix and vector. Maybe you have a better name in mind?

tests/unit/mathematics/SGLinalg_unittest.cc
```diff
@@ -60,6 +60,21 @@ TEST(SGLinalg, CPUBackend_dot)
 	EXPECT_NEAR(result, 20.0, 1E-15);
 }
 
+TEST(SGLinalg, SGVector_CPUBackend_dot)
```
lambday: Well, when you change the behavior of this `convert_vector` thing (return gpu type if there is any, otherwise go on with cpu type), then this test should work even outside `HAVE_VIENNACL` guard.

src/shogun/mathematics/linalg/BaseVector.h
```diff
@@ -54,7 +54,7 @@ struct BaseVector
 	}
 
 	/** Data Storage */
-	virtual bool onGPU() = 0 ;
+	virtual bool onGPU() const = 0 ;
```
vigsterkr: forgive my ignorance but if we have `CPUVector` and `GPUVector` then why do we need `onGPU` function. since the class could identify the storage or?

lambday: well, this was decided because dynamic cast is thought to be slower compared to one virtual call + static cast. But I see what you mean. We probably won't need to have two different classes, only one class should suffice.

lambday: @OXPHOS what was the issue again with putting both these thing inside a single class? I think it is doable.

vigsterkr: well instead of dynamic casting see the SGObject `get_name` virtual function. that's a more general way to identify the type of the object without dynamic casting.

lambday: Okay thanks. But maybe we won't have to do the casting, if we can do what you discussed earlier in the gist @OXPHOS made - have it in the same class and call onGPU method. I can't seem to remember what was stopping us from doing that.

vigsterkr: yeah that would be more ideal indeed.

src/shogun/mathematics/linalg/SGLinalg.cpp
```diff
@@ -118,8 +116,32 @@ bool SGLinalg::hasGPUBackend() const
...
+	std::shared_ptr<BaseVector<T>> basevector;
+
+#ifdef HAVE_VIENNACL
+	if hasGPUBackend()
+	{
+		basevector = std::shared_ptr<BaseVector<T>>(new GPUVector<T>(vector));
+	}
+	else
+	{
+		SG_SINFO("User did not register GPU backend. Will use CPU backend instead. \n");
```
karlnapf: See my other comments about this error message. Suboptimal to write this everytime
@lambday @vigsterkr

src/shogun/mathematics/linalg/SGLinalg.h
```diff
@@ -75,15 +70,19 @@ class SGLinalg
 	 */
 	void set_cpu_backend(std::shared_ptr<CPUBackend> cpubackend);
 
-	/** get CPU backend */
+	/** get CPU backend
+	 * @return cpubackend smart pointer
```
karlnapf: why mention the return type in docs, it will be printed anyways

tests/unit/mathematics/SGLinalg_unittest.cc
```diff
@@ -60,6 +60,21 @@ TEST(SGLinalg, CPUBackend_dot)
...
+{
+	const index_t size=10;
+	SGVector<int32_t> a(size), b(size);
+	a.set_const(1);
```
karlnapf: use range_fill please, not constant. Covers many more cases where things can go wrong

OXPHOS:
ref #3278 
The vector class implementation based on my understand.

So the user will convert SGVector -> LinalgVector for both CPU and GPU as the first step.
Vector can be transferred to GPU if user explicitly calls `LinalgVector.transfer_to_GPU()`.

I'll check #3278 'comments for more details tomorrow as it is kinda late now.
@karlnapf @lambday @vigsterkr

src/shogun/mathematics/linalg/GPUVectorImpl.h
```diff
@@ -0,0 +1,100 @@
...
+#include <shogun/lib/SGVector.h>
+#include <shogun/mathematics/linalg/linalgVector.h>
+#include <memory>
+
+#ifndef GPU_VECTOR_IMPL_H__
+#define GPU_VECTOR_IMPL_H__
+
+#ifdef HAVE_CXX11
+#ifdef HAVE_VIENNACL
+#include <viennacl/vector.hpp>
```
OXPHOS: This class should be able to transfer vector from CPU to GPU via multiple GPU libraries (and use whatever GPU library the user assigns) in the future

src/shogun/mathematics/linalg/GPUVectorImpl.cpp
```diff
@@ -0,0 +1,87 @@
...
+LinalgVector<T>::GPUVectorImpl::GPUVectorImpl()
+{
+}
+
+template<class T>
+LinalgVector<T>::GPUVectorImpl::GPUVectorImpl(T* data, index_t len)
+:m_GPUptr(new VCLMemoryArray()), m_len(len), m_offset(0)
+{
+
+	{
```
vigsterkr: why the double section?

src/shogun/mathematics/linalg/linalgVector.h
```diff
@@ -0,0 +1,108 @@
...
+	 */
+	inline const T& operator[](index_t index) const
+	{
+		return m_data[index];
+	}
+
+private:
+	alignas(CPU_CACHE_LINE_SIZE) bool m_onGPU;
+	alignas(CPU_CACHE_LINE_SIZE) index_t m_len;  // same logic
+	alignas(CPU_CACHE_LINE_SIZE) T* m_data;      // non-owning ptr, referring to the SGVector.vector
```
lambday: The comment `// non-owning ptr, referring to the SGVector.vector` is no longer valid, since this is now a deep copy. Can we please have shared-ptr for m_data here instead?

src/shogun/mathematics/linalg/linalgVector.cpp
```diff
@@ -0,0 +1,168 @@
...
+	{
+		m_onGPU = true;
+		m_gpu_impl = std::unique_ptr<GPUVectorImpl>(new GPUVectorImpl(*(vector.m_gpu_impl)));
+	}
+}
+
+template<class T>
+LinalgVector<T>::~LinalgVector()
+{
+	free(m_data);
```
lambday: Does it work properly? Did you valgrind?

src/shogun/mathematics/linalg/linalgVector.cpp
```diff
@@ -0,0 +1,168 @@
...
+		m_gpu_impl.reset(new GPUVectorImpl(*(vector.m_gpu_impl)));
+	}
+	return *this;
+
+}
+
+template<class T>
+LinalgVector<T>::operator SGVector<T>() const
+{
+	return SGVector<T>(m_data, m_len);
```
lambday: What about the case when data is on GPU? Say, you transfer the data on GPU once and then did a bunch of in-place operations. Then the data pointed by m_data is no-longer what the user would want. 

I think if the data is on GPU, it should copy that to m_data first. Then it should create a new SGVector, deep copy m_data to the vector of that SGVector and then return. This thing above will lead to segfault. For example:
```c++
SGVector<float64_t> v;
if (some_condition)
{
    // create a LinalgVector instance, vec, and do some linalg, finally assign that to the above SGVector
    v  = vec;
    // vec instance dies, along with its m_data
}
// try to use v, bam! dangling pointer, v.vector :(
```

src/shogun/mathematics/linalg/linalgVector.cpp
```diff
@@ -0,0 +1,168 @@
...
+index_t LinalgVector<T>::size() const
+{
+	return m_len;
+}
+
+template<class T>
+void LinalgVector<T>::transferToGPU()
+{
+#ifdef HAVE_VIENNACL
+	m_gpu_impl = std::unique_ptr<GPUVectorImpl>(new GPUVectorImpl(m_data, m_len));
```
lambday: Thread safety?

src/shogun/mathematics/linalg/linalgVector.cpp
```diff
@@ -0,0 +1,168 @@
...
+	SG_SERROR("User did not register GPU backend. \n");
+#endif
+}
+
+template<class T>
+void LinalgVector<T>::transferToCPU()
+{
+	if (m_gpu_impl != nullptr)
+	{
+		m_gpu_impl.release();
```
lambday: what happens to the data on GPU? Shouldn't it be copied to the m_data?

src/shogun/mathematics/linalg/linalgVector.cpp
lambday: actually, when you fix the `transferToCPU` method, just call it here if onGPU() is true, before you create the SGVector.

src/shogun/mathematics/linalg/linalgVector.h
```diff
@@ -0,0 +1,108 @@
...
+#ifdef HAVE_CXX11
+
+namespace shogun
+{
+
+ /** Vector structure
+  * for both CPU and GPU vector
+  */
+template <class T>
+class LinalgVector
```
lambday: Could you please call it `Vector`? Please change the file name also accordingly. `LinalgVector` is a bad name :D

lambday:
Hi @OXPHOS !

Looks good. Please check my comments. Also, please squash all commits together. We should be able to merge this soon!

lambday:
Also, please add unit-tests to all the methods you added. Specially transferToCPU, transferToGPU, cast operators, etc etc. Make sure everything works with valgrind as well :)

src/shogun/mathematics/linalg/linalgVector.cpp
vigsterkr: @lambday why wouldn't it be working? you are using malloc; the alignedness is only about the size nothing else...
@OXPHOS please use `SG_FREE` and not direct memory functions like `free`

OXPHOS:
There's still a problem - say we have both CPU and GPU storage for the same vector. Now every time when we re extracting the Vector, we update the Vector based on GPU information. However, if someone transferred the vector to GPU, but have the `elementwise.product` done on CPU, GPU is actually behind CPU. In this case, current `SGVector<T>()` -> `transfertoCPU()` doesn't make sense at all.

I can think of 1). If `onGPU()==true` and the calculation is done on CPU, we synchronize GPU vector everytime. This will be slow. 2). Ask the user to transfer/synchronize GPU/CPU data before using Vector instance to initialize SGVector

lambday:
Aren't we performing the operations on GPU if onGPU returns true by design? If you check the dot method that you implemented (other methods will be implemented in a similar fashion), then there won't be any case when GPU data is present but the operation is performed on CPU. Does it solve the issue?

src/shogun/mathematics/linalg/Vector.cpp
```diff
@@ -0,0 +1,181 @@
...
+Vector<T>::Vector()
+{
+	init();
+}
+
+template<class T>
+Vector<T>::Vector(SGVector<T> const &vector)
+{
+	init();
+	m_data = std::shared_ptr<T>(reinterpret_cast<T*>(SG_MALLOC(aligned_t, vector.vlen)), free);
```
lambday: Please use `SG_FREE` instead of `free` since you have used `SG_MALLOC` :)

src/shogun/mathematics/linalg/Vector.cpp
```diff
@@ -0,0 +1,181 @@
...
+	std::copy(vector.vector, vector.vector + vector.vlen, m_data.get());
+	m_len = vector.vlen;
+}
+
+template<class T>
+Vector<T>::Vector(Vector<T> const &vector)
+{
+	init();
+
+	m_data = std::shared_ptr<T>(reinterpret_cast<T*>(SG_MALLOC(aligned_t, vector.size())), free);
```
lambday: Please use `SG_FREE` instead of `free` since you have used `SG_MALLOC` :)

src/shogun/mathematics/linalg/Vector.cpp
```diff
@@ -0,0 +1,181 @@
...
+	std::copy(vector.data(), vector.data() + vector.size(), m_data.get());
+	m_len = vector.m_len;
+
+	if (vector.onGPU())
+	{
+		m_onGPU = true;
+		m_gpu_impl = std::unique_ptr<GPUVectorImpl>(new GPUVectorImpl(*(vector.m_gpu_impl)));
+		transferToCPU();
```
lambday: If we do the transfer to CPU here, then the above `std::copy` is no more required :)

src/shogun/mathematics/linalg/Vector.cpp
```diff
@@ -0,0 +1,181 @@
...
+
+template<class T>
+Vector<T>::~Vector()
+{
+}
+
+template<class T>
+Vector<T>& Vector<T>::operator=(SGVector<T> const &vector)
+{
+	m_data = std::shared_ptr<T>(reinterpret_cast<T*>(SG_MALLOC(aligned_t, vector.vlen)), free);
```
lambday: Please use `SG_FREE` instead of `free` since you have used `SG_MALLOC` :)

src/shogun/mathematics/linalg/Vector.cpp
```diff
@@ -0,0 +1,181 @@
...
+	m_len = vector.vlen;
+	m_onGPU = false;
+	m_gpu_impl.release();
+	return *this;
+}
+
+template<class T>
+Vector<T>& Vector<T>::operator=(Vector<T> const &vector)
+{
+	m_data = std::shared_ptr<T>(reinterpret_cast<T*>(SG_MALLOC(aligned_t, vector.size())), free);
```
lambday: Please use `SG_FREE` instead of `free` since you have used `SG_MALLOC` :)

src/shogun/mathematics/linalg/Vector.cpp
```diff
@@ -0,0 +1,181 @@
...
+	std::copy(vector.data(), vector.data() + vector.size(), m_data.get());
+	m_len = vector.m_len;
+	m_onGPU = vector.m_onGPU;
+
+	if (vector.onGPU())
+	{
+		m_onGPU = true;
+		m_gpu_impl.reset(new GPUVectorImpl(*(vector.m_gpu_impl)));
+		transferToCPU();
```
lambday: If we do the transfer to CPU here, then the above `std::copy` is no more required :)

src/shogun/mathematics/linalg/Vector.h
```diff
@@ -0,0 +1,123 @@
...
+	void transferToCPU();
+
+	/** Operator overload for vector read only access
+	 *
+	 * @param index dimension to access
+	 *
+	 */
+	inline const T& operator[](index_t index) const
+	{
+		return m_data.get()[index];
```
lambday: This should return the data from GPU memory if onGPU is true :)

src/shogun/mathematics/linalg/Vector.h
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	/** Operator overload for vector read only access
+	 *
+	 * @param index dimension to access
+	 *
+	 */
+	inline T& operator[](index_t index)
+	{
+		return m_data.get()[index];
```
lambday: This should return the data from GPU memory if onGPU is true :)
Maybe move these to cpp? Otherwise you'll have to include GPUVectorImpl here which is something we want to avoid.

src/shogun/mathematics/linalg/Vector.h
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+private:
+	void init();
+
+public: // Has to be public for unit-tests to use
+	class GPUVectorImpl;
+	alignas(CPU_CACHE_LINE_SIZE) std::unique_ptr<GPUVectorImpl> m_gpu_impl;
```
lambday: err.. this should not be public. Unit-testing can be done via the following:
- create an instance of Vector class, keep the data on cpu.
- create another instance of Vector class, transfer it to gpu.
- assert against values read index by index (will work once you fix the `operator[]`)

tests/unit/mathematics/linalg/LinalgVector_unittest.cc
```diff
@@ -0,0 +1,118 @@
...
+TEST(LinalgVector, deepcopy_constructor_from_SGVector)
+{
+	const index_t size = 10;
+	SGVector<int32_t> a(size);
+	a.range_fill(0);
+
+	Vector<int32_t> a_copy(a);
+	Vector<int32_t> a_vec(a);
+
+	a.range_fill(1);
```
lambday: Curious about this line. What you wanted to do here?

src/shogun/mathematics/linalg/Vector.cpp
```diff
@@ -0,0 +1,226 @@
...
+// Doesn't seem to work
+/*
+#ifdef HAVE_VIENNACL
+	if (onGPU())
+		return (*m_gpu_impl)[index];
+#endif
+*/
+	if (onGPU())
+		transferToCPU();
+	return m_data.get()[index];
```
OXPHOS: I was using (*m_gpu_impl)[index] to access GPU vector element in unit-tests.
But I made some changes here. Ideally, I want [] to return scalar type from GPU when`onGPU()==true` and from CPU when `onGPU()==false`.
So, the `m_gpu_impl` can stay private while the unit-test can still access the element.

However,
- if I do it as the way I commented out, apparently the two types cannot be converted easily.
- if I do `transferToCPU()`, a - it's extremely slow; b - i cannot use it for `operator[](index_t index) const`; c - I cannot change the value of the element (may not required by users? I was changing the value in unit-tests).

So I don't have very good idea to implement the method here.

src/shogun/mathematics/linalg/Vector.cpp
```diff
@@ -0,0 +1,226 @@
...
+{
+	if (onGPU())
+		transferToCPU();
+	return m_data.get();
+}
+
+template<class T>
+T const * Vector<T>::data() const
+{
+	return m_data.get();
```
OXPHOS: I have a copy constructor `Vector<T>::Vector(Vector<T> const &vector)`, which requires this const method.

However, the CPU and GPU cannot be synchronized within const method.

This might not be a big problem, because if the data in onGPU, the new Vector will have synced GPU and CPU vector. But the old vector won't and they'll be different. Also the const and non-const methods are incoherent.

lambday: See, we should not sync the data in everything that we do! GPU data, if present, should be copied to CPU *only when* (a) transferToCPU method is called explicitly or (b) an SGVector is created from this object. Otherwise we'll be ending up doing the transfer all the time! That's not very good!

The point of these `data()` methods were to access the CPU data by CPU backend (say, when you want to create an eigen map, which will only be used when no data is there in the GPU). The GPU backend anyway accesses the gpu data (it should not call these `data()` methods. See what I mean?

Just make the data methods as it was before. In addition to that, let's make onGPU `false` explicitly when non-const `data()` is called. The reason is that, non-const access means that we're modifying it, in which case GPU data should become obsolete.

src/shogun/mathematics/linalg/Vector.cpp
lambday: Try this
```c++
template<class T>
T& Vector<T>::operator[](index_t index)
{
    REQUIRE(index>=0 && index<m_len; "The index (%d) should be within range [0, %d]\n", index, m_len-1);
    T value;
    if (onGPU())
        value = m_gpu_impl[index]; // viennacl::entry_proxy<T> to T transfer (GPU -> CPU happens here)
    else
        value = m_data[index];
    return value;
}
```

lambday: Oh it should not be `T&`. Maybe just add the read-only (const) version for now?

src/shogun/mathematics/linalg/GPUVectorImpl.cpp
```diff
@@ -0,0 +1,103 @@
...
+namespace shogun
+{
+
+template<class T>
+Vector<T>::GPUVectorImpl::GPUVectorImpl()
+{
+}
+
+template<class T>
+Vector<T>::GPUVectorImpl::GPUVectorImpl(T* data, index_t len)
```
karlnapf: Can we have a constructor that just creates an empty structure on the GPU? So that I can allocate giant heaps of memory directly in GPU without the need to transfer it?

karlnapf: Actually for both vectors, matrices, etc

src/shogun/mathematics/linalg/GPUVectorImpl.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+template <class T>
+void Vector<T>::GPUVectorImpl::transferToCPU(T* data)
+{
+	viennacl::backend::memory_read(*m_GPUptr, m_offset*sizeof(T), m_len*sizeof(T),
+		data);
+}
+
+template <class T>
+viennacl::const_entry_proxy<T> Vector<T>::GPUVectorImpl::operator[](index_t index) const
```
karlnapf: This is a dangerous operations, as expensive. Remove.
For unit tests, we can just transfer the full structure back to main memory.
There is no need to access single elements of gpu structures, just do linalg operations on them

src/shogun/mathematics/linalg/GPUVectorImpl.cpp
```diff
@@ -0,0 +1,103 @@
...
+{
+	return viennacl::const_entry_proxy<T>(m_offset+index, *m_GPUptr);
+}
+
+template <class T>
+viennacl::entry_proxy<T> Vector<T>::GPUVectorImpl::operator[](index_t index)
```
karlnapf: same here

src/shogun/mathematics/linalg/GPUVectorImpl.cpp
```diff
@@ -0,0 +1,103 @@
...
+{
+	return viennacl::entry_proxy<T>(m_offset+index, *m_GPUptr);
+}
+
+template class Vector<int32_t>;
```
karlnapf: what about 64 bit?
Some backends dont support it I guess? Does yours?
IF the backend supports, it should be offered here as well

src/shogun/mathematics/linalg/GPUVectorImpl.h
```diff
@@ -0,0 +1,119 @@
...
+
+	typedef viennacl::backend::mem_handle VCLMemoryArray;
+	typedef viennacl::vector_base<T, std::size_t, std::ptrdiff_t> VCLVectorBase;
+	typedef viennacl::vector<T> VCLVector;
+
+public:
+	/** Default constructor */
+	GPUVectorImpl();
+
+	/** Creates a gpu vector with Vector */
```
karlnapf: Please write clearer API docs. What does this mean? :)

src/shogun/mathematics/linalg/GPUVectorImpl.h
```diff
@@ -0,0 +1,119 @@
...
+	GPUVectorImpl(T* data, index_t len);
+
+	/** Copy Constructor */
+	GPUVectorImpl(const Vector<T>::GPUVectorImpl &array);
+
+	/** Returns a ViennaCL vector wrapped around the data of this vector. Can be
```
karlnapf: This one is good

src/shogun/mathematics/linalg/GPUVectorImpl.h
```diff
@@ -0,0 +1,119 @@
...
+	VCLVector vector();
+
+	/** Read only memory access. Note that this is very slow as it copies the
+	 * element from the GPU to the CPU
+	 *
+	 * @param index Element index
+	 */
+	viennacl::const_entry_proxy<T> operator[](index_t index) const;
+
+	/** Read/write memory access. Note that this is very slow as it copies the
```
karlnapf: Remove

src/shogun/mathematics/linalg/Vector.h
```diff
@@ -0,0 +1,134 @@
...
+
+#ifndef LINALG_VECTOR_H__
+#define LINALG_VECTOR_H__
+
+#ifdef HAVE_CXX11
+
+namespace shogun
+{
+
+/** Vector structure
```
karlnapf: Needs better documentation (can be done later) in the `@brief`
In particular, what is this thing used for and maybe an example of a typical case

src/shogun/mathematics/linalg/Vector.h
```diff
@@ -0,0 +1,134 @@
...
+	/** Construct and copy from SGVector */
+	Vector(SGVector<T> const &vector);
+
+	/** Copy Constructor */
+	Vector(Vector<T> const &vector);
+
+	/** Destructor */
+	~Vector();
+
+	/** overload operator = */
```
karlnapf: Be consistent with Capital letters at sentence starts

src/shogun/mathematics/linalg/Vector.h
```diff
@@ -0,0 +1,134 @@
...
+	Vector& operator=(SGVector<T> const &vector);
+
+	/** overload operator = */
+	Vector& operator=(Vector<T> const &vector);
+
+	/** Converts the vector into an SGVector */
+	operator SGVector<T>();
+
+	/** Data Storage */
```
karlnapf: Better comment

src/shogun/mathematics/linalg/Vector.h
```diff
@@ -0,0 +1,134 @@
...
+	bool onGPU() const;
+
+	/** Return vector data */
```
karlnapf: `@return`

src/shogun/mathematics/linalg/Vector.h
```diff
@@ -0,0 +1,134 @@
...
+	T* data();
+
+	/** Return vector data. Read only */
```
karlnapf: No need to specify "read only", it has the const keyword

src/shogun/mathematics/linalg/Vector.h
```diff
@@ -0,0 +1,134 @@
...
+	T const * data() const;
+
+	/** Return vector size. Read only */
+	index_t size() const;
+
+	/** Transfer vector from CPU to GPU */
+	void transferToGPU();
```
karlnapf: BTW, we dont use `transferToGPU` but `transfer_to_gpu`
This is true for all of this patch

karlnapf: Can you put disclaimers at these methods that the calls are expensive

src/shogun/mathematics/linalg/Vector.h
```diff
@@ -0,0 +1,134 @@
...
+
+	/** Transfer vector from GPU to CPU */
+	void transferToCPU();
+
+	/** Release vector from GPU */
+	void releaseFromGPU();
```
karlnapf: The naming here is not really good, what does "release" mean? Free memory on GPU?

src/shogun/mathematics/linalg/Vector.h
```diff
@@ -0,0 +1,134 @@
...
+
+	/** Operator overload for vector read only access
+	 *
+	 * @param index dimension to access
+	 *
+	 */
+	T& operator[](index_t index);
```
karlnapf: Should go away. If one needs elementwise access, one should cast this guy to SGVector

src/shogun/mathematics/linalg/Vector.h
```diff
@@ -0,0 +1,134 @@
...
+
+	/** Operator overload for vector read only access
+	 *
+	 * @param index dimension to access
+	 *
+	 */
+	const T& operator[](index_t index) const;
```
karlnapf: remove

src/shogun/mathematics/linalg/Vector.h
```diff
@@ -0,0 +1,134 @@
...
+
+private:
+	void init();
+
+	/** Store the position of the data */
+	alignas(CPU_CACHE_LINE_SIZE) bool m_onGPU;
```
karlnapf: not sure about all these aligns
@vigsterkr  can you enlighten me?

src/shogun/mathematics/linalg/Vector.cpp
```diff
@@ -0,0 +1,227 @@
...
+Vector<T>::Vector()
+{
+	init();
+}
+
+template<class T>
+Vector<T>::Vector(SGVector<T> const &vector)
+{
+	init();
+	m_data = std::shared_ptr<T>(reinterpret_cast<T*>(SG_MALLOC(aligned_t, vector.vlen)),
```
karlnapf: If we do things like this, we need 100% unit test coverage (that is checked for mem errors)

src/shogun/mathematics/linalg/Vector.cpp
```diff
@@ -0,0 +1,227 @@
...
+									[](T* ptr) {SG_FREE(ptr);});
+	m_len = vector.m_len;
+
+
+	if (vector.onGPU())
+	{
+#ifdef HAVE_VIENNACL
+		m_onGPU = true;
+		m_gpu_impl = std::unique_ptr<GPUVectorImpl>(new GPUVectorImpl(*(vector.m_gpu_impl)));
+		transferToCPU();
```
karlnapf: why do we need this transfer? It seems wrong

src/shogun/mathematics/linalg/Vector.cpp
```diff
@@ -0,0 +1,227 @@
...
+#endif
+	}
+	else
+	{
+		std::copy(vector.data(), vector.data() + vector.size(), m_data.get());
```
karlnapf: this does deep copy only if vector is on CPU?
Why?
No deep copies via copy constructor. There can be a clone or so for that

src/shogun/mathematics/linalg/Vector.cpp
```diff
@@ -0,0 +1,227 @@
...
+{
+	init();
+	m_data = std::shared_ptr<T>(reinterpret_cast<T*>(SG_MALLOC(aligned_t, vector.vlen)),
+									[](T* ptr) {SG_FREE(ptr);});
+	std::copy(vector.vector, vector.vector + vector.vlen, m_data.get());
+	m_len = vector.vlen;
+}
+
+template<class T>
+Vector<T>::Vector(Vector<T> const &vector)
```
karlnapf: Shallow copy for this one

src/shogun/mathematics/linalg/Vector.cpp
```diff
@@ -0,0 +1,227 @@
...
+Vector<T>::~Vector()
+{
+}
+
+template<class T>
+Vector<T>& Vector<T>::operator=(SGVector<T> const &vector)
+{
+	m_data = std::shared_ptr<T>(reinterpret_cast<T*>(SG_MALLOC(aligned_t, vector.vlen)),
+									[](T* ptr) {SG_FREE(ptr);});
+	std::copy(vector.vector, vector.vector+vector.vlen, m_data.get());
```
karlnapf: There is a problem here:
We definitely dont want the data to be copied here. We want the same data to be re-used. Otherwise using linalg operations will always cause a memory copy.
Therefore, if copying from a SGVector, we need to increase the reference counting in order to avoid problems. This means that `Vector` here and the base class `SGReferencesData` need to share the ref counter business.
NOW, that is a problem if we are doing all this in a separate class (no idea how to resolve atm) 
A further problem is the alignment of the data, because we need to copy data in order to re-use memory from SGVector
@vigsterkr @lambday discussion needed

karlnapf: If we did not align things, then the shared reference counting would be easier. We can just make this `Vector` a sub-class of `SGReferencedData` and good.
There are also further questions of how reference counting behaves when data is transferred to GPU

src/shogun/mathematics/linalg/Vector.cpp
```diff
@@ -0,0 +1,227 @@
...
+	m_len = vector.vlen;
+	m_onGPU = false;
+	m_gpu_impl.release();
```
karlnapf: why this release call?

src/shogun/mathematics/linalg/Vector.cpp
```diff
@@ -0,0 +1,227 @@
...
+	m_data = std::shared_ptr<T>(reinterpret_cast<T*>(SG_MALLOC(aligned_t, vector.size())),
+	 							[](T* ptr) {SG_FREE(ptr);});
+	m_len = vector.m_len;
+
+	if (vector.onGPU())
+	{
+#ifdef HAVE_VIENNACL
+		m_onGPU = true;
+		m_gpu_impl.reset(new GPUVectorImpl(*(vector.m_gpu_impl)));
+		transferToCPU();
```
karlnapf: why this?

src/shogun/mathematics/linalg/Vector.cpp
karlnapf: Ok some more thoughts, more explicit

 * *Definitely* no copying. It has to be possible that we pass two `SGVector`s to `linalg::dot` which causes an operation that is just like doing the dot-product with say eigen3
 * This means there is a problem with shared memory managment between `SGVector` (used to create this `Vector`) and this `Vector`. Deleting the `SGVector` might free memory and `Vector` causes a segfault. Dont want that. So we need to share the reference counting. This means that `Vector` *has* to be a subclass of `SGReferencedData`. 
 * Finally, there is a problem if data is transfered to GPU, because now, the original `SGVector` memory *can* safely be freed. So once we transfer data, reference counter can be decreased. This means transferring data "gives up" on data in main memory. Transferring back creates a *new* `SGVector` that is again reference counted. Maybe we should call the transfer in fact *move_to_gpu* (as original data might be freed)
* Need to think about alignment. Might have to be remove for now (and then later introduce when we align `SGVector` as well. I have limited understanding of this though.

My two cents. Let me know

karlnapf: On the question whether to have a separate `linalg::Vector` as compared to putting things into `SGVector`, the fact that we really dont want to have element-wise access for GPU data speaks for having a separate class.

src/shogun/mathematics/linalg/Vector.cpp
```diff
@@ -0,0 +1,227 @@
...
+	init();
+
+	m_data = std::shared_ptr<T>(reinterpret_cast<T*>(SG_MALLOC(aligned_t, vector.size())),
+									[](T* ptr) {SG_FREE(ptr);});
+	m_len = vector.m_len;
+
+
+	if (vector.onGPU())
+	{
+#ifdef HAVE_VIENNACL
```
vigsterkr: why is thins ifdef still here???????
only check whether gpubackend is registered.... nothing else

src/shogun/mathematics/linalg/Vector.cpp
```diff
@@ -0,0 +1,227 @@
...
+template<class T>
+Vector<T>& Vector<T>::operator=(Vector<T> const &vector)
+{
+	m_data = std::shared_ptr<T>(reinterpret_cast<T*>(SG_MALLOC(aligned_t, vector.size())),
+	 							[](T* ptr) {SG_FREE(ptr);});
+	m_len = vector.m_len;
+
+	if (vector.onGPU())
+	{
+#ifdef HAVE_VIENNACL
```
vigsterkr: wie so?

src/shogun/mathematics/linalg/Vector.h
vigsterkr: https://nativecoding.wordpress.com/2015/06/19/multithreading-multicore-programming-and-false-sharing-benchmark/
http://stackoverflow.com/questions/18236603/cache-lines-false-sharing-and-alignment
http://protheus.net.au/cache-in-a-multi-core-environment/

vigsterkr:
http://www.stack.nl/~dimitri/doxygen/manual/docblocks.html

src/shogun/mathematics/linalg/Vector.cpp
OXPHOS: @vigsterkr I didn't get your idea here.. If I remove the #ifdef I cannot even finish compiling as m_gpu_impl is initialized here.

PR #3278, opened 2016-06-10
linalg sum method

OXPHOS:


src/shogun/mathematics/linalg/GPUBackend.cpp
```diff
@@ -60,9 +61,23 @@ T GPUBackend::dot(const GPUVector<T> &a, const GPUVector<T> &b) const
 #endif
 }
 
+template <typename T>
+T GPUBackend::sum(const GPUVector<T> &vec) const
+{
+#ifdef HAVE_VIENNACL
+	return viennacl::linalg::sum(vec.gpuarray->GPUvec());
+#else
+	SG_SERROR("User did not register GPU backend. \n");
```
karlnapf: I don't like it if we would have to put this error message in *every* implementation.

Minor: No space after period.

src/shogun/mathematics/linalg/GPUBackend.h
```diff
@@ -66,6 +66,15 @@ class GPUBackend
 	 */
 	template <typename T>
 	T dot(const GPUVector<T> &a, const GPUVector<T> &b) const;
+
+	/**
+	 * Method that computes the sum of SGVectors using Eigen3
```
karlnapf: not Eigen3.
I suggest to remove the library from the doc. Also above
@lambday thoughts?

src/shogun/mathematics/linalg/CPUBackend.h
```diff
@@ -65,6 +65,15 @@ class CPUBackend
 	 */
 	template <typename T>
 	T dot(const CPUVector<T> &a, const CPUVector<T> &b) const;
+
+	/**
+	 * Method that computes the sum of SGVectors using Eigen3
```
karlnapf: I would remove "with Eigen3" here

src/shogun/mathematics/linalg/SGLinalg.cpp
```diff
@@ -113,13 +113,37 @@ T SGLinalg::dot(BaseVector<T> *a, BaseVector<T> *b) const
 	}
 }
 
+template <class T>
+T SGLinalg::sum(BaseVector<T> *vec) const
+{
+	if (vec->onGPU())
+	{
+		if (this->hasGPUBackend())
+		{
```
karlnapf: we dont use {} for single line statements

src/shogun/mathematics/linalg/SGLinalg.cpp
```diff
@@ -113,13 +113,37 @@ T SGLinalg::dot(BaseVector<T> *a, BaseVector<T> *b) const
...
+			return m_gpubackend->sum<T>(static_cast<GPUVector<T>&>(*vec));
+		}
+		else
+		{
+			SG_SERROR("User did not register GPU backend. \n");
```
karlnapf: Why was there an error message in the sum implementation if there is already one here?

Also, can we just call some method that throws the error? We dont want to re-write this all the time.
Finally, can you please provide a better error message? Like "Tried to call GPU method %s without registering a GPU backend first. Register as blablabla"

src/shogun/mathematics/linalg/SGLinalg.cpp
```diff
@@ -113,13 +113,37 @@ T SGLinalg::dot(BaseVector<T> *a, BaseVector<T> *b) const
...
+ 			return -1;
```
karlnapf: -1 ?

tests/unit/mathematics/SGLinalg_unittest.cc
```diff
@@ -60,6 +60,19 @@ TEST(SGLinalg, CPUBackend_dot)
 	EXPECT_NEAR(result, 20.0, 1E-15);
 }
 
+TEST(SGLinalg, CPUBackend_sum)
+{
+	const index_t size=10;
```
karlnapf: This is not an optimal unit test.
Why don't you use range_fill, then at least we guard against a few more cases

tests/unit/mathematics/SGLinalg_unittest.cc
```diff
@@ -106,6 +119,23 @@ TEST(SGLinalg, GPUBackend_dot)
...
 
+TEST(SGLinalg, GPUBackend_sum)
+{
+	const index_t size=10;
+	SGVector<int32_t> a(size);
+	a.set_const(2);
+
+	GPUVector<int32_t> a_GPU(a);
+
+	std::shared_ptr<GPUBackend> ViennaCLBackend;
```
karlnapf: this needs to be guarded, no?

tests/unit/mathematics/SGLinalg_unittest.cc
```diff
@@ -106,6 +119,23 @@ TEST(SGLinalg, GPUBackend_dot)
...
+	ViennaCLBackend = std::shared_ptr<GPUBackend>(new GPUBackend);
+
+	sg_linalg->set_gpu_backend(ViennaCLBackend);
```
karlnapf: I suggest you keep the GPU backend test in a different file, or at least move registering the backend to a method. To avoid having to register it for every test

tests/unit/mathematics/SGLinalg_unittest.cc
```diff
@@ -106,6 +119,23 @@ TEST(SGLinalg, GPUBackend_dot)
 	EXPECT_NEAR(result, 20.0, 1E-15);
 }
 
+TEST(SGLinalg, GPUBackend_sum)
+{
+	const index_t size=10;
+	SGVector<int32_t> a(size);
+	a.set_const(2);
+
+	GPUVector<int32_t> a_GPU(a);
```
karlnapf: this guy now lives on CPU memory if no GPU is available?

src/shogun/mathematics/linalg/GPUBackend.cpp
vigsterkr: why do we generate at all `GPUBackend` class if there's no gpu backend? i'm sorry but i'm very very confused what's happening here...

you should implement a `GPUBackend` using for example viannacl IF it's available. if not you just don't create one. like many other models, that if LAPACK was not present, the implementation was not built.

src/shogun/mathematics/linalg/GPUBackend.cpp
```diff
@@ -37,6 +37,7 @@
 #ifdef HAVE_VIENNACL
 #include <viennacl/vector.hpp>
 #include <viennacl/linalg/inner_prod.hpp>
+#include <viennacl/linalg/sum.hpp>
 #endif
```
vigsterkr: why close here the `#ifdef HAVE_VIENNACL` ?
i mean if you don't have viennacl this whole implementation is just foobar.
do it like
```
#ifdef HAVE_VIENNACL

... the whole GPUBackend implementation...
#endif
```

src/shogun/mathematics/linalg/GPUBackend.cpp
```diff
@@ -37,6 +37,7 @@
 #ifdef HAVE_VIENNACL
 #include <viennacl/vector.hpp>
 #include <viennacl/linalg/inner_prod.hpp>
+#include <viennacl/linalg/sum.hpp>
```
vigsterkr: why close here the `#ifdef HAVE_VIENNACL` ?
i mean if you don't have viennacl this whole implementation is just foobar.
do it like
```
#ifdef HAVE_VIENNACL

... the whole GPUBackend implementation...
#endif
```

src/shogun/mathematics/linalg/GPUBackend.cpp
vigsterkr: why close here the `#ifdef HAVE_VIENNACL` ?
i mean if you don't have viennacl this whole implementation is just foobar.
do it like
```
#ifdef HAVE_VIENNACL

... the whole GPUBackend implementation...
#endif
```

src/shogun/mathematics/linalg/SGLinalg.cpp
vigsterkr: `NaN` in best case....

src/shogun/mathematics/linalg/GPUBackend.cpp
lambday: Agreed! The creation of the class itself should fail if viennacl is not found.

src/shogun/mathematics/linalg/GPUBackend.h
lambday: Agreed!

src/shogun/mathematics/linalg/SGLinalg.cpp
```diff
@@ -113,13 +113,37 @@ T SGLinalg::dot(BaseVector<T> *a, BaseVector<T> *b) const
...
 }
 
+template <class T>
+T SGLinalg::sum(BaseVector<T> *vec) const
+{
+	if (vec->onGPU())
+	{
+		if (this->hasGPUBackend())
+		{
+			return m_gpubackend->sum<T>(static_cast<GPUVector<T>&>(*vec));
```
lambday: don't need to pass `T` in `sum<T>` I believe.

lambday:
@OXPHOS hey!

Can you quickly change the implementation of vector class based on what we had discussed? It is better to be done in the initial phase before we add too many methods (easy to change now than later). So the idea is that there will only be one class for vector, something like
```c++
class GPUVectorImpl; // this is the GPUArray thing that you have

template <typename T>
class Vector
{
public:
    Vector(SGVector<T> const &);
    Vector& operator=(SGVector<T> const &);
    operator SGVector<T>() const;
    bool onGPU() const;
    T* data();
    T const * data() const;
    index_t size() const;
private:
    bool m_onGPU;   // this variable would be assessed in every linalg call,
                    // makes sense to put this first
    index_t m_len;  // same logic
    T* m_data;      // non-owning ptr, referring to the SGVector.vector
    unique_ptr<GPUVectorImpl> m_gpu_impl;
    void transferToGPU();
    void transferToCPU();
};
```

So, every time a call to `transferToGPU()` is made (by SGLinalg class), the flag `onGPU` is set to be `true`. The main thing to keep in mind here: we need to make sure that the CPU and GPU data are in sync.

- We need to ensure that we invalidate the GPU data if the CPU data is updated. Say, someone transfers the data to GPU, then (s)he assigns another `SGVector` to it (`m_data` points to a different memory now). Then the data on GPU is no more valid. Subsequent calls to linalg methods should use the CPU data then. 
```c++
// vec1 and vec2 are SGVectors
auto v = sg_linalg->transferToGPU(vec1);
// do something with v
v = vec2;
// v.onGPU() == ?, should be false
```
To achieve this, all we have to do is the set `onGPU` is false every time the non-const method `data()` is invoked. Then as long as we implement the assignment operator with the non const `data()` method we're good. 

- Similarly, if the data is on GPU and we perform some in-place operations, then the CPU data is obsolete. We need to make sure we call the `transferToCPU()` method in the operator that returns a `SGVector` then (after all the linalg operations we need to return a `SGVector` finally anyway).

- The linalg methods that take a `Vector<T> const &` as parameters (e.g. dot, sum, etc), will make calls to the const version of `data` so it's all good. Methods that performs in-place operations will simply pass a non-const reference. Notice that we won't pass the vector as a ptr anymore. It can live on stack (just like SGVector).

Also, think a bit about thread safety while updating the `onGPU` variable. Please let me know if you have any questions about this. This should be much cleaner than what we have now - hate the name "BaseVector" for a class name :D 

@karlnapf @vigsterkr @lisitsyn please share your thoughts and suggestions.

lambday:
Maybe it is a good idea to keep the refcount as well?

PR #3279, opened 2016-06-11
Fix evaluation xval

Saurabh7:


Saurabh7:
Might lead to rare issues with say `ContingencyTableEvaluation` which used stored members to evaluate.
http://buildbot.shogun-toolbox.org/builders/CentOS7%20-%20libshogun/builds/465/steps/test/logs/stdio

karlnapf:
Thanks!

karlnapf:
BTW.
Can you please check that *all* calls inside the omp are thread safe?
Otherwise we might run into more problems later

PR #3282, opened 2016-06-12
cookbook - CARTree -  classification tree

OXPHOS:
- CART takes a SGVector in parameter list. What I had in the code is the most reasonable way I found to initialize a SGVector in meta - but it didn't work. I got error:

```
/Users/zora/Github/shogun/build/examples/meta/cpp/classifier/randomforest.cpp:32:16: error: 
      no template named 'CSGVector'; did you mean 'SGVector'?
auto ft = some<CSGVector<bool>>(2);
               ^~~~~~~~~
               SGVector
/Users/zora/Github/shogun/src/shogun/lib/SGString.h:22:25: note: 'SGVector'
      declared here
template<class T> class SGVector;
```
- Need more work on the introduction of cart and I didn't check the style yet as `make cookbook` fails because of the changes I made for the first point.


karlnapf:
The trick is to change the json for cpp

karlnapf:
It should be easier to use a bool vector...

karlnapf:
@OXPHOS currently the meta language does not support
 * instantiating `SG*` types
 * accessing elements within

I think we should add both. @sorig do you have an idea how much work that is? We would need to adjust the include path stuff (similar to what @OXPHOS did in this patch), and then wrapping the `[]` and `()` operators might be nice as well. What do you think?

@OXPHOS Let's see what @sorig says before we continue here

OXPHOS:
I rebased based on #3285 but got the same error here and at local.

I removed `parser.out`, `parsetab.py`, `parsetab.pyc`, `lextab.py` and `lextab.pyc` in examples/meta/generator.

karlnapf:
1. Remove the build dir
2. Clean up the `examples/meta/generator/` parser output

It works on my machine. Ask travis, it might confirm

sorig:
Let me know if you need help with this.

Java error is caused by [java.json#L14](https://github.com/shogun-toolbox/shogun/blob/develop/examples/meta/generator/targets/java.json#L14) not using the correct type. You need to figure out how `BoolVector`s are mapped to Java and use the appropriate type here.

C# is also an error with mapped types. Take the generated C# file, figure out how to make it compile, and update the `csharp.json` file.
```
cartree.cs(26,22): error CS1502: The best overloaded method match for `CARTree.CARTree(BoolVector, EProblemType, int, bool)' has some invalid arguments
/opt/shogun/build/src/interfaces/csharp_modular/modshogun.dll (Location of the symbol related to previous error)
cartree.cs(26,34): error CS1503: Argument `#1' cannot convert `bool[]' expression to type `BoolVector'
```

The errors in the other languages are probably similar, but need investigation.

OXPHOS:
@sorig Hey I have got some results, but not solutions...
So I need to use a new `BoolVector` for `cartree`, but I looked into the errors in Java and found that there's no `BoolVector` type in `jblas`. I think this is why Heiko commented out [this](https://github.com/shogun-toolbox/shogun/commit/1909ded8a6987806f34330d9003adce035af5da3).
If we just use `0` and `1` for `false` and `true`,  apparently `csharp` doesn't like it.  And use `IntVector` instead of `BoolVector` won't work either. So I am not sure how to deal with the problem.

There's another problem in Java, even if the above one is solved, also appeared in #3303:
```java
//![set_attribute_types]
DoubleMatrix ft = new DoubleMatrix(2);
ft.put(0, 0);
ft.put(1, 0);
//![set_attribute_types]
```
If I create a vector of length 2 in meta language (and translated in `cartree.java` as shown in the above code), and pass the vector to the constructor of `CCarTree`:
```java
//![create_instance]
CARTree classifier = new CARTree(ft, PT_MULTICLASS, 5, true);
//![create_instance]
```
and output the length of the vector in CARTREE constructor:
```cpp
CCARTree::CCARTree(SGVector<bool> attribute_types, EProblemType prob_type, int32_t num_folds, bool cv_prune): CTreeMachine<CARTreeNodeData>()
{
    std::cerr << "vector length: " << attribute_types.vlen << std::endl;
}
```
I will get: `vector length: 1`, when `ft` has size of 2.
I have tried initializing a matrix of size(1, 2) or size(2, 1), and I got the same error.
So in summary whatever size vector/matrix I pass to the constructor of CCarTree as param, it will end up with size of 1. I have no idea why this is happening. I am not sure whether you could help look into this? Thanks!

OXPHOS:
@karlnapf any thoughts? : )

karlnapf:
@OXPHOS I can look at this, shall I just copy the meta example in this PR? Does it reproduce the errors.
With Java, there is a problem with strong types and implicit numerical downcasting ... Ill look intot his

OXPHOS:
@karlnapf Maybe you could reproduce the error by using the code in #3303, where the error is simpler: When passing a vector
```
//![set_attribute_types]
DoubleMatrix ft = new DoubleMatrix(2);
ft.put(0, 0);
ft.put(1, 0);
//![set_attribute_types]
```
to the constructor:
```
CCHAIDTree(int32_t dependent_vartype, SGVector<int32_t> feature_types, int32_t num_breakpoints=0);
```
The vector will have size of 1.

karlnapf:
The "other" problem should be solved in #3451 
Looking into the other tomorrow or so.

karlnapf:
There is another problem with csharp and boolean vectors --- we don't have a typemap, which means we cannot pass boolean vectors as input in the csharp interface.

This, sadly, means that for now, we have to change this example to not using boolean vectors that are hand-constructed as parameters. Instead, we can create an instance of the (mapped) `BoolVector` as an object and then use a method (say `zero`) to populate it (hacky)

karlnapf:
Can you rebase? Should work now

karlnapf:
I guess this needs an updated data version and a squash. It is all ready otherwise

karlnapf:
Can you please run the tests locally (at least the cpp version) before putting things into the PR. This would have been detected locally.

Is done with
`ctest -R integration_meta_cpp-multiclass_classifier-cartree -V`

karlnapf:
See the error is coming from a missing shogun-data update

The generated test passed:
https://travis-ci.org/shogun-toolbox/shogun/jobs/175608246#L3717

But checking against reference data failed:
https://travis-ci.org/shogun-toolbox/shogun/jobs/175608246#L4322

karlnapf:
And it even tells you the line where it differs:
https://travis-ci.org/shogun-toolbox/shogun/jobs/175608246#L4427

OXPHOS:
@karlnapf Thanks..! I saw only some of the tests failed so I thought it should be some specific errors without checking closely.. I updated the test dataset and it worked for me at local now.

karlnapf:
So let's get it in then :)

OXPHOS:
`Ruby` and `CSharp` don't like `0` as `False` value, and it might also be true for `Octave`. But it is weird that in `R` tests the `integration_meta_cpp-multiclass_classifier-cartree` failed as the generated output didn't match ref. I'll check whether there is chance to generate different results from the algorithm.

karlnapf:
Ah sigh, I thought I had solved that. But we have a mechanic to fix this...checking  ...

karlnapf:
@OXPHOS rebase against develop, then it should work (I tried locally for all interfaces that failed in the last build here)

karlnapf:
@OXPHOS You should always try these things locally before you push. At least in one modular language.

So here is the listing I used:
https://gist.github.com/karlnapf/6cf4186dc77861681ceba938f397f2c0

Note that meta language syntax is `True` `False`

Also, check out the docs for `git commit --amend`, which allows you to update your commit, so that the PR doesnt have like 5 commits, but just one

karlnapf:
I dont understand the failure in octave. Checking again. The rest looks fine

karlnapf:
Ok so works locally for oactave, I restarted the build
Just need to squash the commits, then we should be able to merge (fingers crossed travis is fine)

karlnapf:
This seems to be specific to the travis setup, I don't have this problem locally ....

karlnapf:
@OXPHOS solved the problem with octave. Man, this is a beast to get merged.
But I think you can squash the next time before you push ( after rebasing against develop once #3559 is merged), since all the rest works

OXPHOS:
@karlnapf It finally works! Thanks for the advice and sorry that I skipped the local test because of laziness.

karlnapf:
Nice one, this was a big effort to get these Boolean vectors working. But it really helped in figuring out corner cases of the whole system. Thanks for the patience :)

doc/cookbook/source/examples/multiclass_classifier/cartree.rst
```diff
@@ -0,0 +1,47 @@
+==================================
+Classification And Regression Tree
+==================================
+
+Decision tree learning uses a decision tree as a predictive model which maps observations about an item to conclusions about the item's target value.
```
karlnapf: Decition tree learning is based on trees as predictive models. (Remove the second sentence)

doc/cookbook/source/examples/multiclass_classifier/cartree.rst
```diff
@@ -0,0 +1,47 @@
...
+
+Decision trees are mostly used as the following two types:
```
karlnapf: There are two types of decision trees:

karlnapf: Or even better: There are decision trees for both classification (integer-valued) and regression (real-valued).

doc/cookbook/source/examples/multiclass_classifier/cartree.rst
```diff
@@ -0,0 +1,47 @@
...
+
+- Classification tree, where the predicted outcome is the class to which the data belongs.
+- Regression tree, where predicted outcome can be considered a real number.
+
+Classification And Regression Tree (CART) algorithm is an umbrella method that can be applied to generate both classification tree and regression tree.
```
karlnapf: The `classification and regression tree` (CART) algorithm is an umbrella method ....

doc/cookbook/source/examples/multiclass_classifier/cartree.rst
```diff
@@ -0,0 +1,47 @@
...
+
+In this example, we showed how to apply CART algorithm to multi-class dataset and predict the labels with classification tree.
```
karlnapf: In this example, we show how to apply CART to multi-class datasets.

Remove the rest and fix typos

doc/cookbook/source/examples/multiclass_classifier/cartree.rst
```diff
@@ -0,0 +1,47 @@
...
+
+-------
+Example
+-------
+
+Imagine we have files with training and test data. We create CDenseFeatures (here 64 bit floats aka RealFeatures) and :sgclass:`CMulticlassLabels` as
+
+.. sgexample:: cartree.sg:create_features
+
+We set the type of each predictive attribute (true for nominal, false for ordinal/continuous)
```
karlnapf: These words you can also use above when you talk about regression/classification values

doc/cookbook/source/examples/multiclass_classifier/cartree.rst
```diff
@@ -0,0 +1,47 @@
...
+
+.. sgexample:: cartree.sg:set_attribute_types
+
+We create an instance of the :sgclass:`CCARTree` classifier by passting it the attribute types and the tree type.
+We can also set the number of subsets used in cross-valiation and whether to use cross-validation pruning.
```
karlnapf: A reference to a wiki article would be good here

PR #3246, opened 2016-06-05
add cookbook for neural net classification

sanuj:
Integration data for this cookbook also keeps on changing.

PR #3283, opened 2016-06-12
add %newobject in neuralnet interface

sanuj:
For #3246. Fixes segfault while doing:
```
layers = NeuralLayers()
layers = layers.input(num_feats)
layers = layers.linear(50)
layers = layers.softmax(2)
```

### DIFF
--- a/src/shogun/neuralnets/NeuralLayers.h
+++ b/src/shogun/neuralnets/NeuralLayers.h
@@ -47,6 +47,9 @@ namespace shogun
  * Uses builder-style syntax to describe layers
  * one by one.
  *
+ * While adding method for a new layer, also add its prototype
+ * in NeuralNets.i prefixed by "%newobject" since these methods return
+ * a reference to the same object.
  *
  */
 class CNeuralLayers : public CSGObject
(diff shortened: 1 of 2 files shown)


karlnapf:
Can you valgrind your mini example for leaks?

lisitsyn:
:+1:

sanuj:
[Valgrind output](https://gist.github.com/sanuj/052df069becd588315a2407e6ea274cc) using `valgrind --tool=memcheck --suppressions=valgrind-python.supp python ./feedforward_net_classification.py` by following [this](http://stackoverflow.com/questions/3982036/how-can-i-use-valgrind-with-python-c-extensions) link.

karlnapf:
So this works now?

lisitsyn:
One more thing!

Lets put a warning in the doc of these methods.

src/shogun/neuralnets/NeuralLayers.h
```diff
@@ -47,6 +47,7 @@ namespace shogun
  * Uses builder-style syntax to describe layers
  * one by one.
  *
+ * While adding a new layer, also add it's prototype in SWIG with %newobject.
```
karlnapf: This both contains a grammar error "it's -> its", and also doesnt make any sense to me :)

rather remove, or replace with something meaningful. Recall that users of Shogun dont know what SWIG is. Rather say that every methods returns a reference to the same object or so

PR #3285, opened 2016-06-13
Meta language: Modified parser to handle SGVector, SGMatrix accessing

sorig:
WIP

### DIFF
--- a/examples/meta/.gitignore
+++ b/examples/meta/.gitignore
@@ -1,3 +1,0 @@
-generator/parsetab.py
-generator/parser.out
-generator/lextab.py
--- a/examples/meta/generator/parse.py
+++ b/examples/meta/generator/parse.py
@@ -1,8 +1,8 @@
 import sys
 import json
-import os
 import argparse
 import ply
+
 
 # The FastParser parses input using PLY
 class FastParser:
@@ -10,12 +10,8 @@ class FastParser:
         from ply import lex
         from ply import yacc
 
-        # Add all shogun types to reserved identifiers
-        # (this is commented out as we allow any identifier to be shogun one)
-        #self.addShogunTypes(self.reserved)
-
         # Build the lexer and the parser
-        self.lexer = lex.lex(module=self,optimize=1)
+        self.lexer = lex.lex(module=self, optimize=1)
         self.parser = yacc.yacc(module=self)
 
     def parse(self, programString, filePath="Unknown"):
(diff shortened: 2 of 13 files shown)


examples/meta/.gitignore
```diff
@@ -1,3 +0,0 @@
-generator/parsetab.py
```
karlnapf: +1

examples/meta/generator/parse.py
```diff
@@ -85,7 +77,8 @@ def addShogunTypes(self, dictionary):
     t_EQUALS = "="
     t_LPAREN = "\("
     t_RPAREN = "\)"
-    t_COMMENT = r"\#.*\n"
```
karlnapf: comment removed?

karlnapf:
Cool thanks for that. Once we have a json update, I can take over

examples/meta/generator/parse.py
sorig: There was a bug with the parsing error messages not counting comments as a line, so I've moved this rule into one with the side effect of incrementing the line count like the rule for the NEWLINE token.

sorig:
@karlnapf ready for review when travis has finished 

These changes include:
- Support for element access and element assignment for SGVectors and SGMatrices
- Refactor of `translate.py`: dependency extraction and automatic injection of variable storing is now mostly decoupled from translation
- Support for native construction of SGMatrix/SGVector equivalents in target languages. This is implemented for Python. It is an optional feature so doesn't break the other languages.

Does not include:
- Checking that variables being accessed (e.g. `myMatrix[0,2]`) are actually of type SGVector/SGMatrix

sorig:
Here is a [test file](https://gist.github.com/sorig/fa501825e97d33719fbe741d9d988a36) that I used. Would be nice to get a meta example using these features merged, so we can test all the languages.

examples/meta/generator/targets/python.json
```diff
@@ -11,9 +11,33 @@
     "Comment": "#$comment\n",
     "Init": {
         "Construct": "$name = $type($arguments)",
-        "Copy": "$name = $expr"
+        "Copy": "$name = $expr",
+        "BoolVector": "$name = np.zeros( ($arguments), dtype='bool')",
+        "CharVector": "$name = np.zeros( ($arguments), dtype='|S1')",
+        "ByteVector": "$name = np.zeros( ($arguments), dtype='uint8')",
+        "WordVector": "$name = np.zeros( ($arguments), dtype='uint16')",
+        "ShortVector": "$name = np.zeros( ($arguments), dtype=???????)",
```
sorig: I can't import some of these from the Python interface, so can't check what type they should have. Maybe they just need to be removed?

karlnapf:
Let me know when the vector / matrix access stuff is there.
In the test file, it would be good to check both accessing matrix and vector

sorig:
@karlnapf All green. Ready to merge

sorig:
The test file has both a vector and a matrix :)

karlnapf:
Great thanks

@oxphos this will soon allow for vector element access in meta examples, after I added some minor adjustments

karlnapf:
@OXPHOS you should be able to use [something like this](https://gist.github.com/sorig/fa501825e97d33719fbe741d9d988a36) now

Can you try?

karlnapf:
keep in mind to remove old parser output

PR #3287, opened 2016-06-14
cookbook - shareboost

OXPHOS:


examples/meta/src/classifier/shareboost.sg
```diff
@@ -0,0 +1,30 @@
...
+
+#![create_features]
+RealFeatures features_train(f_feats_train)
+RealFeatures features_test(f_feats_test)
+MulticlassLabels labels_train(f_labels_train)
+MulticlassLabels labels_test(f_labels_test)
+#![create_features]
+
+#![create_instance]
+ShareBoost shareboost(features_train, labels_train, 30)
```
OXPHOS: The last argument should be `min(fm_train_real.shape[0]-1, 30)`. I dont know where the `30` came from but as the training dataset has 80 rows I just wrote down 30 here.

examples/meta/src/classifier/shareboost.sg
```diff
@@ -0,0 +1,30 @@
...
+#![create_instance]
+
+#![train_and_apply]
+shareboost.train()
+RealSubsetFeatures features_test_sub(features_test, shareboost.get_activeset())
```
OXPHOS: `RealSubsetFeatures` cannot be identified.

karlnapf: you need to add it to the cpp json, `generator/targets/cpp.json` in the same style as RealFeatures etc

examples/meta/src/classifier/shareboost.sg
karlnapf: ??
Please read the docs (or the code) for what this thing does. It should be apparent.
And then mention it in the cookbook. No need to take min here, you can just decide on any number (that works for this dataset)

OXPHOS: It's nonzeros features but it didn't say where the number `30` came from. 
My guess is this is the number from the paper.

karlnapf: yeah just some subset of features. You can just say that you pick some number of data here

OXPHOS:
@karlnapf I am not 100% sure it's a bug - because it makes sense to have nonzero features number < total feature number - but for this dataset, I have two use all (**both**) features to get correct output (classification).

With the orig codes, if I do `ShareBoost shareboost(features_train, labels_train, 1)`, I'll get binary classification; with `ShareBoost shareboost(features_train, labels_train, 2)`, I'll get error: `It doesn't make sense to use ShareBoost with num non-zero features > num features in the data\n`

OXPHOS:
I can send in a mini patch separately if necessary

src/shogun/multiclass/ShareBoost.cpp
```diff
@@ -53,8 +53,8 @@ bool CShareBoost::train_machine(CFeatures* data)
 
 	if (m_nonzero_feas <= 0)
 		SG_ERROR("Set a valid (> 0) number of non-zero features to seek before training\n")
-	if (m_nonzero_feas >= dynamic_cast<CDenseFeatures<float64_t>*>(m_features)->get_num_features())
-		SG_ERROR("It doesn't make sense to use ShareBoost with num non-zero features >= num features in the data\n")
+	if (m_nonzero_feas > dynamic_cast<CDenseFeatures<float64_t>*>(m_features)->get_num_features())
```
karlnapf: Please put this into a new PR
Also, if you are touching this already, please do
´"Number of non-zero features (%d) cannot be larger than number of features (%d) in the data\n"`
I.e. pass the information to user

Furthermore, you need to check that the result of the dynamic cast is not NULL, in which case you would segfault here

karlnapf:
I think your forgot the `.rst` :)

karlnapf:
Also needs integration test data

OXPHOS:
@karlnapf updated

doc/cookbook/source/references.bib
```diff
@@ -79,3 +79,10 @@ @inproceedings{gao2011discriminative
   year={2011},
   organization={IEEE}
 }
+@inproceedings{shalev2011shareboost,
+  title={Shareboost: Efficient multiclass learning with feature sharing},
+  author={Shalev-Shwartz, Shai and Wexler, Yonatan and Shashua, Amnon},
```
karlnapf: pls use the same style as the other bibtex files. I.e. abbreviate first names

doc/cookbook/source/examples/multiclass_classifier/shareboost.rst
```diff
@@ -0,0 +1,33 @@
+==========
+ShareBoost
+==========
+
+ShareBoost algorithm learns a multiclass predictor from a subset of shared features of the samples with forward greedy selection approach.
```
karlnapf: The ShareBoost algorithm

karlnapf:
Minor comments, pls send a separate fix. Merging. Thanks!

PR #3290, opened 2016-06-15
lars max l1 norm test

Saurabh7:


vigsterkr:
@Saurabh7 can you just put this test into the other branch PR, that way travis will test everything at once...

PR #3291, opened 2016-06-15
Fix Python 2 to 3 exception conversion.

arianepaola:
Fixes exception to be Python 2 and 3 compatible.


arianepaola:
@karlnapf ready to merge

PR #3305, opened 2016-06-21
refactored GaussianKernel as a subclass of CShiftInvariantKernel

lambday:
@karlnapf please have a  look!

### DIFF
--- a/src/shogun/kernel/GaussianKernel.cpp
+++ b/src/shogun/kernel/GaussianKernel.cpp
@@ -7,53 +7,43 @@
  * Written (W) 1999-2010 Soeren Sonnenburg
  * Written (W) 2011 Abhinav Maurya
  * Written (W) 2012 Heiko Strathmann
+ * Written (W) 2016 Soumyajit De
  * Copyright (C) 1999-2009 Fraunhofer Institute FIRST and Max-Planck-Society
  * Copyright (C) 2010 Berlin Institute of Technology
  */
 
 #include <shogun/lib/common.h>
-#include <shogun/base/Parameter.h>
 #include <shogun/kernel/GaussianKernel.h>
 #include <shogun/features/DotFeatures.h>
-#include <shogun/io/SGIO.h>
+#include <shogun/distance/EuclideanDistance.h>
 #include <shogun/mathematics/Math.h>
 
 using namespace shogun;
 
-void CGaussianKernel::set_width(float64_t w)
+CGaussianKernel::CGaussianKernel() : CShiftInvariantKernel()
 {
-	REQUIRE(w>0, "width (%f) must be positive\n",w);
-	m_log_width=CMath::log(w/2.0)/2.0;
+	register_params();
 }
 
-float64_t CGaussianKernel::get_width() const
+CGaussianKernel::CGaussianKernel(float64_t w) : CShiftInvariantKernel()
 {
-	return CMath::exp(m_log_width*2.0)*2.0;
-}
-
-CGaussianKernel::CGaussianKernel() : CDotKernel()
-{
-	init();
-}
-
-CGaussianKernel::CGaussianKernel(float64_t w) : CDotKernel()
-{
-	init();
+	register_params();
 	set_width(w);
 }
 
-CGaussianKernel::CGaussianKernel(int32_t size, float64_t w) : CDotKernel(size)
+CGaussianKernel::CGaussianKernel(int32_t size, float64_t w) : CShiftInvariantKernel()
 {
-	init();
+	register_params();
 	set_width(w);
+	set_cache_size(size);
 }
 
-CGaussianKernel::CGaussianKernel(CDotFeatures* l, CDotFeatures* r,
-		float64_t w, int32_t size) : CDotKernel(size)
+CGaussianKernel::CGaussianKernel(CDotFeatures* l, CDotFeatures* r, float64_t w, int32_t size) : CShiftInvariantKernel()
 {
-	init();
+	register_params();
 	set_width(w);
-	init(l,r);
+	set_cache_size(size);
+	init(l, r);
 }
 
 CGaussianKernel::~CGaussianKernel()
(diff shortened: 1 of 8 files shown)


src/shogun/kernel/GaussianKernel.cpp
```diff
@@ -63,13 +53,9 @@ CGaussianKernel::~CGaussianKernel()
 
 CGaussianKernel* CGaussianKernel::obtain_from_generic(CKernel* kernel)
 {
-	if (kernel->get_kernel_type()!=K_GAUSSIAN)
-	{
-		SG_SERROR("CGaussianKernel::obtain_from_generic(): provided kernel is "
-				"not of type CGaussianKernel!\n");
-	}
+	REQUIRE(kernel->get_kernel_type()==K_GAUSSIAN,
+		"Provided kernel (%s) must be of type CGaussianKernel!\n", kernel->get_name());
```
karlnapf: maybe use self->name here rather than CGaussianKernel?

src/shogun/kernel/GaussianKernel.cpp
```diff
@@ -82,86 +68,55 @@ CSGObject *CGaussianKernel::shallow_copy() const
...
-	REQUIRE(lhs && rhs, "Features not set!\n")
+	REQUIRE(lhs, "Left hand side features must be set!\n")
+	REQUIRE(rhs, "Rightt hand side features must be set!\n")
 
 	if (!strcmp(param->m_name, "log_width"))
 	{
 		SGMatrix<float64_t> derivative=SGMatrix<float64_t>(num_lhs, num_rhs);
-
 		for (int j=0; j<num_lhs; j++)
+		{
```
karlnapf: openmp maybe?
But this is not the point here I guess

tests/unit/base/SGObject_unittest.cc
```diff
@@ -208,7 +208,7 @@ TEST(SGObject,equals_complex_equal)
 	SG_UNREF(file);
 
 	/* now compare */
-	floatmax_t accuracy=1E-15;
+	floatmax_t accuracy=1E-10;
```
karlnapf: ?

karlnapf:
Fine to merge from my side once trabis is green

src/shogun/kernel/GaussianKernel.cpp
lambday: hehe tried that.. then realized that this is a static method :D

src/shogun/kernel/GaussianKernel.cpp
```diff
@@ -82,86 +68,55 @@ CSGObject *CGaussianKernel::shallow_copy() const
...
 			for (int k=0; k<num_rhs; k++)
 			{
-				float64_t element=distance(j,k);
-				derivative(j,k)=exp(-element)*element*2.0;
+				float64_t element=distance(j, k);
```
lambday: actually this is slow if precomputed distance matrix is provided. maybe I'll try changing the order in a separate patch (making sure that col-wise traversal is actually beneficial)

src/shogun/kernel/GaussianKernel.cpp
lambday: agreed

src/shogun/kernel/GaussianKernel.cpp
```diff
@@ -82,86 +68,55 @@ CSGObject *CGaussianKernel::shallow_copy() const
...
-	if (lhs==rhs)
-		sq_rhs=sq_lhs;
-	else
-		precompute_squared_helper(sq_rhs, (CDotFeatures*) rhs);
+	return CMath::exp(m_log_width*2.0)*2.0;
 }
 
-SGMatrix<float64_t> CGaussianKernel::get_parameter_gradient(
-		const TParameter* param, index_t index)
+SGMatrix<float64_t> CGaussianKernel::get_parameter_gradient(const TParameter* param, index_t index)
```
vigsterkr: why does it only can be a `float64_t`? and not a `float32_t` or a `floatmax_t`?

src/shogun/kernel/GaussianKernel.cpp
```diff
@@ -82,86 +68,55 @@ CSGObject *CGaussianKernel::shallow_copy() const
...
 		SGMatrix<float64_t> derivative=SGMatrix<float64_t>(num_lhs, num_rhs);
-
 		for (int j=0; j<num_lhs; j++)
+		{
 			for (int k=0; k<num_rhs; k++)
 			{
-				float64_t element=distance(j,k);
-				derivative(j,k)=exp(-element)*element*2.0;
+				float64_t element=distance(j, k);
+				derivative(j, k)=exp(-element)*element*2.0;
```
vigsterkr: for fun and profit: http://stackoverflow.com/questions/10552280/fast-exp-calculation-possible-to-improve-accuracy-without-losing-too-much-perfo/10792513#10792513

src/shogun/kernel/GaussianKernel.cpp
```diff
@@ -171,15 +126,33 @@ SGMatrix<float64_t> CGaussianKernel::get_parameter_gradient(
...
 {
-	return (sq_lhs[idx_a]+sq_rhs[idx_b]-2*CDotKernel::compute(idx_a,idx_b))/get_width();
+	CKernel::load_serializable_post();
+	if (lhs && rhs)
+		m_distance->init(lhs, rhs);
+}
+
+float64_t CGaussianKernel::distance(int32_t idx_a, int32_t idx_b) const
+{
+	return CShiftInvariantKernel::distance(idx_a, idx_b)/get_width();
```
vigsterkr: mmm will clang/gcc in `-O3` optimize this into `1/get_width()` being a const and just do a `CShiftInvariantKernel::distance(idx_a, idx_b)*<that const>`?

src/shogun/kernel/GaussianKernel.cpp
```diff
@@ -82,86 +68,55 @@ CSGObject *CGaussianKernel::shallow_copy() const
...
-	if (!lhs || !rhs)
-		return;
-
-	precompute_squared_helper(sq_lhs, (CDotFeatures*) lhs);
-
-	if (lhs==rhs)
-		sq_rhs=sq_lhs;
-	else
-		precompute_squared_helper(sq_rhs, (CDotFeatures*) rhs);
+	return CMath::exp(m_log_width*2.0)*2.0;
```
vigsterkr: see below about fastexp

src/shogun/kernel/GaussianKernel.cpp
lambday: good point. it is coming from the fact that the distance we calculate is always `float64_t` no matter what. we should change that later. me and @karlnapf had a discussion on similar things a few days back. how about (just like in CKernel), we have a templated version of `CDistance::distance`, and then a non-templated version that (by default) returns float64_t.

src/shogun/kernel/GaussianKernel.cpp
lambday: good point. not sure whether it does that. better to be explicit.

lambday:
updated the PR with a few suggested changes. travis is green. merging this one.

src/shogun/kernel/GaussianKernel.cpp
karlnapf: Yeah this is not good. But not the point of this PR and has been there before.
This patch is just some mini changes that we need for research code.
chris can deal with this, he is almost done with LARS

PR #3307, opened 2016-06-21
fix uninitialized memory read error in CEuclideanDistance and a typo

lambday:


lambday:
minor, travis is green. merging.

PR #3313, opened 2016-06-26
Updated LARS so that it can support float32

c4goldsw:


src/interfaces/python_modular/CMakeLists.txt
```diff
@@ -66,4 +66,4 @@ configure_file(${CMAKE_CURRENT_BINARY_DIR}/main_init.py.in main_init.py COPYONLY
 INSTALL(FILES ${CMAKE_CURRENT_BINARY_DIR}/main_init.py
 	DESTINATION ${PYTHON_PACKAGES_PATH}/shogun
   COMPONENT python
-	RENAME __init__.py)
+	RENAME __init__.py)
```
vigsterkr: why this whitespace change? :)

src/shogun/features/DenseFeatures.cpp
```diff
@@ -352,6 +355,12 @@ template<class ST> void CDenseFeatures<ST>::copy_feature_matrix(SGMatrix<ST> src
 	initialize_cache();
 }
 
+template<class ST> Eigen::Map < Eigen::Matrix< ST, -1, -1, 0,-1,-1>, 0, Eigen::Stride<0,0>> CDenseFeatures<ST>::get_eigen_map()
```
vigsterkr: you could define a `typedef` for this, like `CDenseFeaturesEigenMap`:
```
typdef Eigen::Map < Eigen::Matrix< ST, -1, -1, 0,-1,-1>, 0, Eigen::Stride<0,0>> CDenseFeaturesEigenMap<ST>
```
and shoudn't we return a reference?

src/shogun/regression/LeastAngleRegression.h
```diff
@@ -13,12 +13,26 @@
...
 #include <shogun/machine/LinearMachine.h>
 
 namespace shogun
 {
+
+template <typename inputType>
+struct EigenTypes
+{
+	using ST = inputType;
+	typedef Eigen::Matrix<ST,-1,1,0,-1,1> EigenVectorXt;
```
vigsterkr: i reckon this is typedef-ed in SGVector:
https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/lib/SGVector.h#L40

you could use that from there, it's a public definition.

src/shogun/regression/LeastAngleRegression.h
```diff
@@ -72,14 +86,19 @@ class CFeatures;
 class CLeastAngleRegression: public CLinearMachine
 {
 public:
+	
+	const unsigned int Lower = 1;
```
vigsterkr: use `uint32_t` please

src/shogun/regression/LeastAngleRegression.h
```diff
@@ -154,6 +175,7 @@ class CLeastAngleRegression: public CLinearMachine
 	 * become invalid if train is called *again*. So make a copy if you want to
 	 * call train multiple times.
 	 */
+	template <typename ST>
```
vigsterkr: why the template param if you dont use it below?

src/shogun/regression/LeastAngleRegression.h
```diff
@@ -183,13 +205,385 @@ class CLeastAngleRegression: public CLinearMachine
...
+		}
+	}
+
+	template <typename ST>
+	static void find_max_abs(const vector<ST> &vec, const vector<bool> &ignore_mask,
+		int32_t &imax, ST& vmax)
+	{
+		imax = -1;
+		vmax = -1;
+		for (uint32_t i=0; i < vec.size(); ++i)
```
vigsterkr: use `index_t`

vigsterkr:
@c4goldsw good stuff!
i've added couple of comments, but there's a major problem: why did you have to move the whole implementation out of the `.cpp` into the header?

src/shogun/regression/LeastAngleRegression.h
```diff
@@ -183,13 +205,385 @@ class CLeastAngleRegression: public CLinearMachine
...
+			}
+		}
+	}
+
+	template <typename ST>
+	bool train_machine(CFeatures* data)
+	{
+		if (!m_labels)
+			SG_ERROR("No labels set\n")
+		if (m_labels->get_label_type() != LT_REGRESSION)
```
karlnapf: Can you pls print the provided  features
`"Provided Labels (%s) are of type (%d) where they should be regression labels (%d)\n", m_labels->get_name(), m_labels->get_label_type(), LT_REGRESSION)`

src/shogun/regression/LeastAngleRegression.h
```diff
@@ -183,13 +205,385 @@ class CLeastAngleRegression: public CLinearMachine
...
+		if (!data)
+			data=features;
+
+		if (!data)
+			SG_ERROR("No features set\n")
+
+		if (m_labels->get_num_labels() != data->get_num_vectors())
+			SG_ERROR("Number of training vectors does not match number of labels\n")
+
+		if (data->get_feature_class() != C_DENSE)
```
karlnapf: See above

src/shogun/regression/LeastAngleRegression.h
```diff
@@ -183,13 +205,385 @@ class CLeastAngleRegression: public CLinearMachine
...
+		if (m_labels->get_label_type() != LT_REGRESSION)
+			SG_ERROR("Expected RegressionLabels\n")
+
+		if (!data)
+			data=features;
+
+		if (!data)
+			SG_ERROR("No features set\n")
+
+		if (m_labels->get_num_labels() != data->get_num_vectors())
```
karlnapf: See above, please print the numbers in there

src/shogun/regression/LeastAngleRegression.h
karlnapf: What is data is NULL? Segfault? Needs a check I guess

src/shogun/regression/LeastAngleRegression.h
```diff
@@ -183,13 +205,385 @@ class CLeastAngleRegression: public CLinearMachine
...
+			SG_ERROR("Number of training vectors does not match number of labels\n")
+
+		if (data->get_feature_class() != C_DENSE)
+			SG_ERROR("Expected Simple Features\n")
+
+		//if (data->get_feature_type() != F_DREAL)
```
karlnapf: no commented out code pls :)

src/shogun/regression/LeastAngleRegression.h
```diff
@@ -183,13 +205,385 @@ class CLeastAngleRegression: public CLinearMachine
...
+		//	SG_ERROR("Expected Real Features\n")
+
+		std::vector<std::vector<ST>> m_beta_path_t;		
+
+		CDenseFeatures<ST>* feats=(CDenseFeatures<ST>*) data;
```
karlnapf: yesss!

src/shogun/regression/LeastAngleRegression.h
```diff
@@ -183,13 +205,385 @@ class CLeastAngleRegression: public CLinearMachine
...
+
+		// init facilities
+		m_beta_idx.clear();
+		m_beta_path_t.clear();
+		m_num_active = 0;
+		m_active_set.clear();
+		m_is_active.resize(n_fea);
+		fill(m_is_active.begin(), m_is_active.end(), false);
+
+		SGVector<ST> y = ((CRegressionLabels*) m_labels)->template get_labels_t<ST>();
```
karlnapf: This creates a copy if the type doesnt match ST?

tests/unit/regression/lars_unittest.cc
```diff
@@ -91,6 +91,30 @@ void generate_data_n_less_d(SGMatrix<float64_t> &data, SGVector<float64_t> &lab)
 	lab[2]=-0.553422897279663;
 }
 
+//Generate the Data that N is less than D
+void generate_data_n_less_d(SGMatrix<float32_t> &data, SGVector<float64_t> &lab)
+{
```
karlnapf: Can't you re-use the existing code in the sense that you generate the float64 data, and then manually copy it elementwise to float32 structures?

src/shogun/regression/LeastAngleRegression.h
```diff
@@ -183,13 +205,385 @@ class CLeastAngleRegression: public CLinearMachine
...
+			if (CMath::abs(vec[i]) > vmax)
+			{
+				vmax = CMath::abs(vec[i]);
+				imax = i;
+			}
+		}
+	}
+
+	template <typename ST>
+	bool train_machine(CFeatures* data)
```
karlnapf: I dont think this should be templated.
The type should always be inferred from the provided features.
So maybe accept CDenseFeatures<T> here and get T from there?

src/shogun/regression/LeastAngleRegression.h
```diff
@@ -210,7 +605,7 @@ class CLeastAngleRegression: public CLinearMachine
 	int32_t m_max_nonz;  //!< max number of non-zero variables for early stopping
 	float64_t m_max_l1_norm; //!< max l1-norm of beta (estimator) for early stopping
 
-	std::vector<std::vector<float64_t> > m_beta_path;
+	std::vector<std::vector<float64_t>> m_beta_path;
```
karlnapf: pls dont do that (yet=)

tests/unit/regression/lars_unittest.cc
```diff
@@ -210,10 +234,10 @@ TEST(LeastAngleRegression, lars_N_LessThan_D)
 	SG_REF(labels);
 	CLeastAngleRegression* lars=new CLeastAngleRegression(false);
 	lars->set_labels(labels);
-	lars->train(features);
+	lars->train_machine<float64_t>(features);
```
karlnapf: no template here pls

tests/unit/regression/lars_unittest.cc
```diff
@@ -210,10 +234,10 @@ TEST(LeastAngleRegression, lars_N_LessThan_D)
...
 
-	SGVector<float64_t> active2=SGVector<float64_t>(lars->get_w_for_var(2));
-	SGVector<float64_t> active1=SGVector<float64_t>(lars->get_w_for_var(1));
+	SGVector<float64_t> active2=SGVector<float64_t>(lars->get_w_for_var<float64_t>(2));
```
karlnapf: I wouldnt template this

tests/unit/regression/lars_unittest.cc
```diff
@@ -233,6 +257,41 @@ TEST(LeastAngleRegression, lars_N_LessThan_D)
 	SG_UNREF(labels);
 }
 
+TEST(LeastAngleRegression, lars_N_LessThan_D_32_feature_test)
```
karlnapf: please dont mix CamelCase_with_underscores

tests/unit/regression/lars_unittest.cc
```diff
@@ -233,6 +257,41 @@ TEST(LeastAngleRegression, lars_N_LessThan_D)
...
+	SGVector<float64_t> lab(3);
+	generate_data_n_less_d(data,lab);
+
+	CDenseFeatures<float32_t>* features=new CDenseFeatures<float32_t>(data);
+	SG_REF(features);
+	CRegressionLabels* labels=new CRegressionLabels(lab);
+	SG_REF(labels);
+	CLeastAngleRegression* lars=new CLeastAngleRegression(false);
+	lars->set_labels(labels);
+	lars->train_machine<float32_t>(features);
```
karlnapf: great that this works now.
Now let's clean up the interface a bit in the sense that we dont have templated train methods.

karlnapf:
+1 Good stuff.
Though the typename should always come from provided features, currently the templated methods somehow make things redundant

src/shogun/regression/LeastAngleRegression.h
c4goldsw: I realised that it's superfluous, fixed.

src/interfaces/python_modular/CMakeLists.txt
c4goldsw: I had generated interface code and accidentally committed it, so I removed all the files from that directory, restored the ones from the head (that were there by default) and re-commited them.  I forgot to restore this file, and after struggling to get it from the remote repository, I just recreated it and did a cut and paste.

**tdlr** Still learning about git.

src/shogun/features/DenseFeatures.cpp
c4goldsw: Why should we return a reference?  If I think about it, it would be a good idea if we're dealing with a lot of features - the object could be allocated on the heap/free store and a reference could be passed back to LARs.  This would avoid having the whole thing made on the stack (as it currently is), just to have it copied on returning it.

c4goldsw: The more I think about this, the more confused I am.  I also realise that everything I said above may not be correct - don't EigenMaps just wrap around SG structures?  What would the advantage be of getting a reference to the map?  Would I have to free that reference once it's no longer needed?  I've only learned C and I'm still learning C++, so I appreciate the help.

c4goldsw:
@vigsterkr I moved methods from the .cpp file into the .h file that had to be templated - isn't that standard practice?  But, I'll be moving a large portion of the methods back in .cpp file as they don't have to be templated.

src/shogun/regression/LeastAngleRegression.h
c4goldsw: Do what?  This was already implemented.

tests/unit/regression/lars_unittest.cc
c4goldsw: I realised this was superfluous, taken out.

tests/unit/regression/lars_unittest.cc
c4goldsw: Will do.

src/shogun/regression/LeastAngleRegression.h
c4goldsw: That's already handled by lines 271 to 275 (tested too).

src/shogun/regression/LeastAngleRegression.h
c4goldsw: Still learning c++ : is there an elegant way to go about this (i.e. getting the type of CDenseFeatures and passing that type onto other things which rely off it)?

src/shogun/regression/LeastAngleRegression.h
c4goldsw: No - I realise that's what we discussed doing, I'll implement that now (and I believe you mean float64_t and not ST :) ).

vigsterkr:
@c4goldsw check how it is done in case of `DenseFeatures`

src/shogun/features/DenseFeatures.cpp
vigsterkr: yeah but even if it's a map the content is copied if it's not a reference.... you dont want to copy anything since you just want the one obj to be passed around, or?

src/shogun/regression/LeastAngleRegression.h
```diff
@@ -183,13 +205,385 @@ class CLeastAngleRegression: public CLinearMachine
 	/** @return object name */
 	virtual const char* get_name() const { return "LeastAngleRegression"; }
 
+
+	template <typename ST>
+	static vector<ST> make_vector(int32_t size, ST val)
+	{
```
karlnapf: Pls replace this by http://en.cppreference.com/w/cpp/algorithm/iota

src/shogun/regression/LeastAngleRegression.h
```diff
@@ -183,13 +205,385 @@ class CLeastAngleRegression: public CLinearMachine
...
+		vector<ST> result(size);
+		fill(result.begin(), result.end(), val);
+		return result;
+	}
+
+	template <typename ST>
+	static void plane_rot(ST x0, ST x1,
+		ST &y0, ST &y1, SGMatrix<ST> &G)
+	{
+		G.zero();
```
karlnapf: can you memset to zero manually here? We dont want to use this interface really

src/shogun/regression/LeastAngleRegression.h
karlnapf: do this change, removing the space between the > >

src/shogun/regression/LeastAngleRegression.h
karlnapf: I am actually not sure about the most elegant way at this moment.
We might have to do some explicit checking of types (somewhere under the hood of shogun)
@vigsterkr whats you suggestion?

src/shogun/regression/LeastAngleRegression.h
c4goldsw: @karlnapf Also, I won't be adding additional error info to this line or line 284 yet, since CFeatures can't adequately get the feature / scalar type yet (I can do that in a later PR).

src/shogun/regression/LeastAngleRegression.h
c4goldsw: Why would we want to increment the values of a vector (as the entries change)?  Whoever implemented this is only using it to make vectors with all 0 entries, so I'll just replace those lines with a memset to zero.

vigsterkr: @c4goldsw :+1: for that

c4goldsw:
@vigsterkr So, I've changed everything except for making ```CDenseLabels::get_eigen_map()``` return a reference to the eigen map it creates, as I still don't know why it should return a reference (though I'm no longer confused about what I had asked above).

src/shogun/labels/DenseLabels.h
```diff
@@ -115,19 +136,59 @@ class CDenseLabels : public CLabels
...
-		/** Getter for labels
+		/** get copy of labels.
 		 *
-		 * @return labels, a copy if a subset is set
+		 * possible with subset
+		 *
+		 * @return labels
 		 */
-		SGVector<float64_t> get_labels();
+		 SGVector<float64_t> get_labels_copy();
```
vigsterkr: no need for that little space there.

src/shogun/regression/LeastAngleRegression.cpp
```diff
@@ -75,96 +58,131 @@ static void find_max_abs(const vector<float64_t> &vec, const vector<bool> &ignor
...
 
 	if (!data)
-		data=features;
+	{
+		if(!features)
+			SG_ERROR("Both data and LinearMachine's features is null! Please ensure that one of them is set\n")
 
-	if (!data)
-		SG_ERROR("No features set\n")
+		SG_INFO("Data is null, so using LinearMachine's features instead.\n")
```
karlnapf: No need for that INFO message

src/shogun/regression/LeastAngleRegression.cpp
```diff
@@ -75,96 +58,131 @@ static void find_max_abs(const vector<float64_t> &vec, const vector<bool> &ignor
...
+
+		if(features->get_feature_class() != C_DENSE)
+			SG_ERROR("LinearMachine's features' feature-type is not C_DENSE. Please ensure that it is\n")
```
karlnapf: Pls dont use classnames in error messages.
Just say "Feature type (%d) must be C_DENSE (%d)\n", ....

src/shogun/regression/LeastAngleRegression.cpp
```diff
@@ -75,96 +58,131 @@ static void find_max_abs(const vector<float64_t> &vec, const vector<bool> &ignor
...
-		SG_ERROR("Expected Simple Features\n")
+	//check for type of CFeatures, then call the appropriate template method
+	if(data->get_feature_type() == F_DREAL)
+		return CLeastAngleRegression::train_machine_i((CDenseFeatures<float64_t> *) data);
+	else//data->get_feature_type() == F_DREAL
+		return CLeastAngleRegression::train_machine_i((CDenseFeatures<float32_t> *) data);
+}
+
+template <typename ST>
+bool CLeastAngleRegression::train_machine_i(CDenseFeatures<ST> * data)
```
karlnapf: whats the i_i suffix supposed to mean

src/shogun/regression/LeastAngleRegression.cpp
```diff
@@ -75,96 +58,131 @@ static void find_max_abs(const vector<float64_t> &vec, const vector<bool> &ignor
...
-		SG_ERROR("Number of training vectors does not match number of labels\n")
+	if (data->get_feature_type() != F_DREAL && data->get_feature_type() != F_SHORTREAL)
+		SG_ERROR("Expected Real Features!\n")
 
-	if (data->get_feature_class() != C_DENSE)
-		SG_ERROR("Expected Simple Features\n")
+	//check for type of CFeatures, then call the appropriate template method
+	if(data->get_feature_type() == F_DREAL)
+		return CLeastAngleRegression::train_machine_i((CDenseFeatures<float64_t> *) data);
+	else//data->get_feature_type() == F_DREAL
```
karlnapf: ehm???

karlnapf: You basically need to switch over all types that CFeatures can have here, and throw an error message otherwise

karlnapf: @vigsterkr 
Can't we somehow do these type of checks in a base class? This will appear over and over again if we start porting more algos to using multiple memory widths

src/shogun/regression/LeastAngleRegression.cpp
```diff
@@ -175,75 +193,76 @@ bool CLeastAngleRegression::train_machine(CFeatures* data)
 	//========================================
 	int32_t nloop=0;
 	while (m_num_active < max_active_allowed && max_corr/n_vec > m_epsilon && !stop_cond)
-	{
+	{
```
karlnapf: these tabs are not necessary

src/shogun/regression/LeastAngleRegression.cpp
```diff
@@ -175,75 +193,76 @@ bool CLeastAngleRegression::train_machine(CFeatures* data)
...
 		
 		// corr_sign = sign(corr)
-		for (uint32_t i=0; i < corr.size(); ++i)
+		for (index_t i=0; i < corr.size(); ++i)
 			corr_sign[i] = CMath::sign(corr[i]);
 
 		// find max absolute correlation in inactive set
 		find_max_abs(corr, m_is_active, i_max_corr, max_corr);
-
+
```
karlnapf: can you check your PR to avoid such whitespace changes, in particular if you just add random tabs to lines

tests/unit/regression/lars_unittest.cc
```diff
@@ -233,6 +233,48 @@ TEST(LeastAngleRegression, lars_N_LessThan_D)
...
+	generate_data_n_less_d(data_64,lab);
+
+	//copy data_64 into a float32_t SGMatrix
+	SGMatrix<float32_t> data(5,3);
+	for(index_t c = 0; c < data_64.num_cols; ++c)
+		for(index_t r = 0; r < data_64.num_rows; ++r)
+			data(r, c) = (float32_t) data_64(r, c);
+
+
+	CDenseFeatures<float32_t>* features = new CDenseFeatures<float32_t>(data);
```
karlnapf: so now we need tests
 * for all other float types
 *`and for all other types (checking that errors are thrown)

@vigsterkr can we have a unit test that checks the train method for multiple feature types? And then we can just add the the classes that support multiple to that test, and specify which types it should work on

```
test_features_train_type(lars_instance, {float32_t, float64_6, floatmax_}) // checks that lars works with all those types and throws an error otherwise
```

karlnapf: or we could even have a template for unit tests that just exchanges the features in train and results.
Like automating the test that @c4goldsw wrote here based on the 64 bit version

src/shogun/regression/LeastAngleRegression.cpp
c4goldsw: "_i" was for "implementation", but I've changed it to "_templated now".

src/shogun/regression/LeastAngleRegression.cpp
c4goldsw: Why though must I check for all types?  If you look at lines 119 and 120, I throw and error if the types are not float32_t or float64_t.  Isn't that sufficient, given those are the only types that make sense?

src/shogun/regression/LeastAngleRegression.cpp
c4goldsw: Will do.

src/shogun/regression/LeastAngleRegression.cpp
c4goldsw: Why though must I check for all types? If you look at lines 119 and 120, I throw and error if the types are not float32_t or float64_t. Isn't that sufficient, given those are the only types that make sense?

tests/unit/regression/lars_unittest.cc
c4goldsw: I'll try to template the test

src/interfaces/python_modular/CMakeLists.txt
karlnapf: can you remove this change?

src/shogun/features/DenseFeatures.cpp
```diff
@@ -352,6 +353,12 @@ template<class ST> void CDenseFeatures<ST>::copy_feature_matrix(SGMatrix<ST> src
 	initialize_cache();
 }
 
+template<class ST> typename SGMatrix<ST>::EigenMatrixXtMap CDenseFeatures<ST>::get_eigen_map()
```
karlnapf: This should not be here but in SGVector

src/shogun/features/DenseFeatures.h
```diff
@@ -25,6 +25,8 @@
 
 #include <shogun/lib/SGMatrix.h>
 
+using namespace Eigen;
```
karlnapf: we dont want eigen3 to be in header files if possible

src/shogun/labels/DenseLabels.h
karlnapf: whitespace cleanup needed

src/shogun/regression/LeastAngleRegression.cpp
karlnapf: Yeah you are right!

src/interfaces/python_modular/CMakeLists.txt
karlnapf: pls remove this change

src/shogun/features/DenseFeatures.h
```diff
@@ -23,8 +23,6 @@
 #include <shogun/features/StringFeatures.h>
 #include <shogun/lib/DataType.h>
 
-#include <shogun/lib/SGMatrix.h>
```
karlnapf: why this?

src/shogun/labels/DenseLabels.h
```diff
@@ -106,6 +112,21 @@ class CDenseLabels : public CLabels
 		 */
 		float64_t get_label(int32_t idx);
 
+		/** get label, templated
```
karlnapf: No need to say it is templated, the code speaks for itself, and the interface

src/shogun/labels/DenseLabels.h
```diff
@@ -106,6 +112,21 @@ class CDenseLabels : public CLabels
...
+		 *
+		 * possible with subset
+		 *
+		 * @param idx index of label to get
+		 * @return value of label
+		 */
+		template<typename ST>
+		ST get_label_t(int32_t idx)
```
karlnapf: is the "_t" suffix the right one  here?
Why not "_i"?

src/shogun/labels/DenseLabels.h
```diff
@@ -115,19 +136,59 @@ class CDenseLabels : public CLabels
 		 */
 		int32_t get_int_label(int32_t idx);
 
-		/** Getter for labels
+		/** get copy of labels.
```
karlnapf: Start sentences with capital letters

src/shogun/labels/DenseLabels.h
```diff
@@ -115,19 +136,59 @@ class CDenseLabels : public CLabels
...
-		 * @return labels, a copy if a subset is set
+		 * possible with subset
+		 *
+		 * @return labels
 		 */
-		SGVector<float64_t> get_labels();
+		SGVector<float64_t> get_labels_copy();
 
-		/** get copy of labels.
+ 		/** Getter for labels which is templated
```
karlnapf: temove templated

src/shogun/labels/DenseLabels.h
```diff
@@ -115,19 +136,59 @@ class CDenseLabels : public CLabels
...
+				return get_labels_copy_t<ST>();
+
+			SGVector<ST> labels_copy(m_labels.vlen);
+			for(int i = 0; i < m_labels.vlen; ++i)
+				labels_copy[i] = (ST) m_labels[i];
+
+			return labels_copy;
+		}
+
+		/** get copy of labels, templated.
```
karlnapf: see other comments

src/shogun/regression/LeastAngleRegression.cpp
```diff
@@ -75,96 +58,132 @@ static void find_max_abs(const vector<float64_t> &vec, const vector<bool> &ignor
...
+		G(0,1) = sx1;
+		G(1,1) = sx0;
 
+		y0 = r;
+		y1 = 0;
+	}
 }
 
-bool CLeastAngleRegression::train_machine(CFeatures* data)
+bool CLeastAngleRegression::train_machine(CFeatures * data)
```
karlnapf: no need for this change

src/shogun/regression/LeastAngleRegression.cpp
```diff
@@ -75,96 +58,132 @@ static void find_max_abs(const vector<float64_t> &vec, const vector<bool> &ignor
...
 {
+	//error checking
```
karlnapf: useless comment

src/shogun/regression/LeastAngleRegression.cpp
```diff
@@ -75,96 +58,132 @@ static void find_max_abs(const vector<float64_t> &vec, const vector<bool> &ignor
...
 	if (!m_labels)
 		SG_ERROR("No labels set\n")
-	if (m_labels->get_label_type() != LT_REGRESSION)
-		SG_ERROR("Expected RegressionLabels\n")
 
-	if (!data)
-		data=features;
+	if (m_labels->get_label_type() != LT_REGRESSION)
+		SG_ERROR("Provided labels (%s) are of type (%d), where they shold be regressionlabels (%d)\n",
```
karlnapf: typo: should
typo: regression labels.
Missing period

src/shogun/regression/LeastAngleRegression.cpp
```diff
@@ -75,96 +58,132 @@ static void find_max_abs(const vector<float64_t> &vec, const vector<bool> &ignor
...
+			m_labels->get_name(), m_labels->get_label_type(), LT_REGRESSION, m_labels->get_label_type())
 
 	if (!data)
-		SG_ERROR("No features set\n")
+	{
+		if(!features)
+			SG_ERROR("Both data and LinearMachine's features is null! Please ensure that one of them is set\n")
```
karlnapf: "No features provided.\n"

src/shogun/regression/LeastAngleRegression.cpp
```diff
@@ -75,96 +58,132 @@ static void find_max_abs(const vector<float64_t> &vec, const vector<bool> &ignor
...
-	if (m_labels->get_num_labels() != data->get_num_vectors())
-		SG_ERROR("Number of training vectors does not match number of labels\n")
+		if(features->get_feature_class() != C_DENSE)
+			SG_ERROR("Feature-type (%d) must be of type C_DENSE (%d)\n", features->get_feature_class(), C_DENSE)
+			
+		data = features;
+	}
+	else
+		if (data->get_feature_class() != C_DENSE)
+			SG_ERROR("Expected Simple Features (C_DENSE)")
```
karlnapf: say what we got instead

src/shogun/regression/LeastAngleRegression.cpp
```diff
@@ -75,96 +58,132 @@ static void find_max_abs(const vector<float64_t> &vec, const vector<bool> &ignor
...
+
+	if (data->get_num_vectors() != m_labels->get_num_labels())
+		SG_ERROR("Number of training vectors (%d) does not match number of labels (%d)\n", 
+			data->get_num_vectors(), m_labels->get_num_labels())
+
+	if (data->get_feature_type() != F_DREAL && data->get_feature_type() != F_SHORTREAL &&
+		data->get_feature_type() != F_LONGREAL)
+		SG_ERROR("Expected Real Features!\n")
```
karlnapf: say what we got, give more details on what was given.
These error messages are for users, so dont make their life hard :)

src/shogun/regression/LeastAngleRegression.cpp
```diff
@@ -75,96 +58,132 @@ static void find_max_abs(const vector<float64_t> &vec, const vector<bool> &ignor
...
+
+	//check for type of CFeatures, then call the appropriate template method
+	if(data->get_feature_type() == F_DREAL)
+		return CLeastAngleRegression::train_machine_templated((CDenseFeatures<float64_t> *) data);
+	else if(data->get_feature_type() == F_SHORTREAL)
+		return CLeastAngleRegression::train_machine_templated((CDenseFeatures<float32_t> *) data);
+	else //data->get_feature_type() == F_LONGREAL
```
karlnapf: remove this comment, put in another else if, and bail on else

src/shogun/regression/LeastAngleRegression.cpp
```diff
@@ -75,96 +58,132 @@ static void find_max_abs(const vector<float64_t> &vec, const vector<bool> &ignor
...
 
-	if (data->get_feature_class() != C_DENSE)
-		SG_ERROR("Expected Simple Features\n")
+template <typename ST>
+bool CLeastAngleRegression::train_machine_templated(CDenseFeatures<ST> * data)
+{
 
-	if (data->get_feature_type() != F_DREAL)
-		SG_ERROR("Expected Real Features\n")
+	std::vector<std::vector<ST>> m_beta_path_t;
```
karlnapf: why the suffix?

src/shogun/regression/LeastAngleRegression.cpp
```diff
@@ -195,55 +214,56 @@ bool CLeastAngleRegression::train_machine(CFeatures* data)
...
+				R=SGMatrix<ST>(1,1);
+				ST diag_k = map_X.col(i_max_corr).dot(map_X.col(i_max_corr));
+				R(0,0) = CMath::sqrt( diag_k);
 			}
 			else
 				R=cholesky_insert(X, X_active, R, i_max_corr, m_num_active);
 			activate_variable(i_max_corr);
 		}
-
+
```
karlnapf: this whitespace is unnecessary

src/shogun/features/DenseFeatures.h
c4goldsw: I believe I had placed that in there whilst trying to use EigenMatrixMapXt - I just took it out as I'm no longer using that.

src/shogun/regression/LeastAngleRegression.cpp
```diff
@@ -278,31 +299,31 @@ bool CLeastAngleRegression::train_machine(CFeatures* data)
...
 
-				Map<VectorXd> map_beta(&beta[0], n_fea);
-				Map<VectorXd> map_beta_prev(&m_beta_path[nloop][0], n_fea);
+				typename SGVector<ST>::EigenVectorXtMap map_beta(&beta[0], n_fea);
+				typename SGVector<ST>::EigenVectorXtMap map_beta_prev(&m_beta_path_t[nloop][0], n_fea);
 				map_beta = (1-s)*map_beta_prev + s*map_beta;
 			}
 		}
-
+
```
karlnapf: I wont comment on more of those, but please dont do such changes

src/shogun/regression/LeastAngleRegression.cpp
```diff
@@ -317,9 +338,9 @@ bool CLeastAngleRegression::train_machine(CFeatures* data)
 				map_Xa.block(0, i_kick, numRows, numCols-i_kick) = 
 					map_Xa.block(0, i_kick+1, numRows, numCols-i_kick).eval();			
 		}
-
+		
 		nloop++;
-		m_beta_path.push_back(beta);
+		m_beta_path_t.push_back(beta);
```
karlnapf: ah now I get the template suffix thing, it is a different variable

src/shogun/regression/LeastAngleRegression.h
```diff
@@ -72,14 +73,19 @@ class CFeatures;
 class CLeastAngleRegression: public CLinearMachine
 {
 public:
+	
+	/*const unsigned int Lower = 1;
```
karlnapf: ?

src/shogun/regression/LeastAngleRegression.h
```diff
@@ -120,14 +126,16 @@ class CLeastAngleRegression: public CLinearMachine
 	 *
 	 * @param num_variable number of non-zero coefficients
 	 */
+	
 	void switch_w(int32_t num_variable)
 	{
 		if (w.vlen <= 0)
-			SG_ERROR("cannot swith estimator before training")
+			SG_ERROR("cannot switch estimator before training")
```
karlnapf: Better error message!

src/shogun/regression/LeastAngleRegression.h
```diff
@@ -184,20 +192,50 @@ class CLeastAngleRegression: public CLinearMachine
...
+
+	template <typename ST>
+	static void find_max_abs(const std::vector<ST> &vec, const std::vector<bool> &ignore_mask,
+		int32_t &imax, ST& vmax);
 
 private:
+	/**
+	* An encapsulated, templated training method.  Can be called by calling train_machine,
+	* the interface method that encapsulates this method.
+	* @param data the data being passed to LARs to be trained on
```
karlnapf: Just say this is a templated specialization of the train method

tests/unit/regression/lars_unittest.cc
```diff
@@ -38,6 +38,8 @@
 #include <shogun/preprocessor/PruneVarSubMean.h>
 #include <shogun/preprocessor/NormOne.h>
 
+#include "stdio.h"
```
karlnapf: remove!

tests/unit/regression/lars_unittest.cc
```diff
@@ -234,6 +236,123 @@ TEST(LeastAngleRegression, lars_N_LessThan_D)
...
 }
 
+template <typename ST>
+void lars_n_less_than_d_feature_test_templated()
+{
+	SGMatrix<float64_t> data_64(5,3);
+	SGVector<float64_t> lab(3);
+	generate_data_n_less_d(data_64,lab);
+
+	std::cout << "Copying contents into matrix.\n";
```
karlnapf: remove

tests/unit/regression/lars_unittest.cc
```diff
@@ -234,6 +236,123 @@ TEST(LeastAngleRegression, lars_N_LessThan_D)
...
+{
+	lars_n_less_than_d_feature_test_templated<float32_t>();
+}
+
+TEST(LeastAngleRegression, lars_template_test_float64)
+{
+	lars_n_less_than_d_feature_test_templated<float64_t>();
+}
+
+TEST(LeastAngleRegression, lars_template_test_floatmax)
```
karlnapf: nice and covering all cases

src/shogun/labels/DenseLabels.h
c4goldsw: "_t" is for "_templated".  Should I have something else there instead?

c4goldsw:
@karlnapf not done yet, I hadn't seen your other comments.

src/shogun/regression/LeastAngleRegression.cpp
c4goldsw: Do you also understand it within the context of the templated functions?

src/shogun/regression/LeastAngleRegression.h
c4goldsw: Left over when I was attempting to template it the train method, forgot to remove.

src/shogun/regression/LeastAngleRegression.h
c4goldsw: Any suggestions?

src/shogun/labels/DenseLabels.h
karlnapf: No it's good, I just got confused ;)

src/shogun/regression/LeastAngleRegression.h
karlnapf: @Saurabh7 probably know a good one

karlnapf:
Time to squash all commits into one and then I'm happy to merge

src/shogun/regression/LeastAngleRegression.h
Saurabh7: "Train model before updating model parameters"  .. something like this sounds good ?

c4goldsw: @Saurabh7 Thanks!

src/shogun/labels/DenseLabels.h
```diff
@@ -106,6 +112,21 @@ class CDenseLabels : public CLabels
...
+		 * possible with subset
+		 *
+		 * @param idx index of label to get
+		 * @return value of label
+		 */
+		template<typename ST>
+		ST get_label_t(int32_t idx)
+		{
+			int32_t real_num=m_subset_stack->subset_idx_conversion(idx);
+			ASSERT(m_labels.vector && idx<get_num_labels())
```
karlnapf: a REQUIRE would be better here, with a proper error message for users

src/shogun/labels/DenseLabels.h
```diff
@@ -115,19 +136,59 @@ class CDenseLabels : public CLabels
...
+		 * set or if ST is not of type float64_t
+		 */ 
+		template<typename ST>
+		SGVector<ST> get_labels_t()
+		{
+			if (m_subset_stack->has_subsets())
+				return get_labels_copy_t<ST>();
+
+			SGVector<ST> labels_copy(m_labels.vlen);
+			for(int i = 0; i < m_labels.vlen; ++i)
```
karlnapf: you can do a memcpy here, no need for a loop

karlnapf: index_t for the loop counter pls

src/shogun/labels/DenseLabels.h
```diff
@@ -115,19 +136,59 @@ class CDenseLabels : public CLabels
...
 		 * @return labels
 		 */
-		SGVector<float64_t> get_labels_copy();
+		template<typename ST>
+		SGVector<ST> get_labels_copy_t()
+		{
+			if (!m_subset_stack->has_subsets())
+			{
+				SGVector<ST> labels_copy(m_labels.vlen);
+				for(int i = 0; i < m_labels.vlen; ++i)
```
karlnapf: index_t and memcpy

src/shogun/regression/LeastAngleRegression.cpp
```diff
@@ -24,45 +24,28 @@ using namespace Eigen;
...
-		y0 = r;
-		y1 = 0;
-	}
+	
 }
 
-static void find_max_abs(const vector<float64_t> &vec, const vector<bool> &ignore_mask,
-	int32_t &imax, float64_t& vmax)
+template <typename ST>
+void CLeastAngleRegression::find_max_abs(const std::vector<ST> &vec, const std::vector<bool> &ignore_mask,
```
karlnapf: is this method hidden from SWIG?
@Saurabh7 it contains a std::vector which we usually try to avoid in interfaces

src/shogun/regression/LeastAngleRegression.cpp
```diff
@@ -75,96 +58,130 @@ static void find_max_abs(const vector<float64_t> &vec, const vector<bool> &ignor
...
-	if (!data)
-		data=features;
+	if (m_labels->get_label_type() != LT_REGRESSION)
+		SG_ERROR("Provided labels (%s) are of type (%d) - they should be regression labels (%d) instead.\n",
+			m_labels->get_name(), m_labels->get_label_type(), LT_REGRESSION, m_labels->get_label_type())
 
 	if (!data)
-		SG_ERROR("No features set\n")
+	{
+		if(!features)
```
karlnapf: REQUIRE saves the if

src/shogun/regression/LeastAngleRegression.h
```diff
@@ -120,14 +122,16 @@ class CLeastAngleRegression: public CLinearMachine
 	 *
 	 * @param num_variable number of non-zero coefficients
 	 */
+	
 	void switch_w(int32_t num_variable)
 	{
 		if (w.vlen <= 0)
-			SG_ERROR("cannot swith estimator before training")
+			SG_ERROR("Please train the model before updating its parameters")
```
karlnapf: use REQUIRE rather than if and SG_ERROR

src/shogun/regression/LeastAngleRegression.h
```diff
@@ -184,20 +188,49 @@ class CLeastAngleRegression: public CLinearMachine
 	virtual const char* get_name() const { return "LeastAngleRegression"; }
 
 protected:
-	virtual bool train_machine(CFeatures* data=NULL);
+	/**
+	* An interface method to - this is called by the superclass's (CLinearMachine)
```
karlnapf: The sentence seems to be starting weirdly "An interface method to -- this is called ..."

karlnapf: No need to explicitly mention the base class name here (i.e. remove CLinearMachine)

src/shogun/regression/LeastAngleRegression.h
```diff
@@ -184,20 +188,49 @@ class CLeastAngleRegression: public CLinearMachine
...
+	* train() method.  This method checks to see if data is a dense feature vector,
```
karlnapf: pls doxygen explicitly reference methods you refer to, so that doxygen generates links in the html ouput

src/shogun/regression/LeastAngleRegression.h
```diff
@@ -184,20 +188,49 @@ class CLeastAngleRegression: public CLinearMachine
...
+	* and that it's elements are of type float64_t or float32_t.  It then calls
```
karlnapf: try to avoid explicitly mentioning the supported types since they easily get outdated. I.e. here oyu talk about 32 and 64, but not about the recently added floatmat. So better completely avoid such things

src/shogun/regression/LeastAngleRegression.h
```diff
@@ -184,20 +188,49 @@ class CLeastAngleRegression: public CLinearMachine
...
 
 protected:
-	virtual bool train_machine(CFeatures* data=NULL);
+	/**
+	* An interface method to - this is called by the superclass's (CLinearMachine)
+	* train() method.  This method checks to see if data is a dense feature vector,
+	* and that it's elements are of type float64_t or float32_t.  It then calls 
+	* train_machine_templated with the appropriate template parameters (either float64_t or
+	* float32_t)
+	* @param data the data being passed to LARs to be trained on
```
karlnapf: typo LARS

karlnapf: The comment is also way to wordy:
`@data training data`

karlnapf:
Man, what an effort. Well done!
I hope you are happy with the result, we certainly are :)

I will merge this now. Please
 * send a clean up patch for my latest set of (minor) comments
 * add to the NEWS file your contribution
 * think about what you learned and how we can generalise this to other Shogun classes, and which.

c4goldsw:
@karlnapf It is good to see this finally in!  Things will be much faster now.  I'll send in the clean up patch soon.

src/shogun/labels/DenseLabels.h
c4goldsw: How am I meant to 'memcpy' if the array types differ?

src/shogun/labels/DenseLabels.h
c4goldsw: Same issue with `memcpy` as above.

src/shogun/regression/LeastAngleRegression.cpp
c4goldsw: I wouldn't be able to tell you, will discus in IRC.

src/shogun/regression/LeastAngleRegression.h
c4goldsw: Will have to ask about this too

c4goldsw: @karlnapf would it be `@see classname::method` ?

src/shogun/labels/DenseLabels.h
karlnapf: You can't. My bad :)

src/shogun/regression/LeastAngleRegression.h
karlnapf: Google it, I dont know by heart

PR #3322, opened 2016-06-29
Parallel RF

Saurabh7:
WIP

### DIFF
--- a/src/shogun/features/DenseFeatures.cpp
+++ b/src/shogun/features/DenseFeatures.cpp
@@ -639,7 +639,6 @@ CFeatures* CDenseFeatures<ST>::copy_dimension_subset(SGVector<index_t> dims)
 template<class ST>
 CFeatures* CDenseFeatures<ST>::shallow_subset_copy()
 {
-	SG_SDEBUG("Entering!\n");
 	CFeatures* shallow_copy_features=nullptr;
 	
 	SG_SDEBUG("Using underlying feature matrix with %d dimensions and %d feature vectors!\n", num_features, num_vectors);
@@ -648,7 +647,6 @@ CFeatures* CDenseFeatures<ST>::shallow_subset_copy()
 	if (m_subset_stack->has_subsets())
 		shallow_copy_features->add_subset(m_subset_stack->get_last_subset()->get_subset_idx());
 	
-	SG_SDEBUG("Leaving!\n");
 	return shallow_copy_features;
 }
 
--- a/src/shogun/features/DenseFeatures.cpp
+++ b/src/shogun/features/DenseFeatures.cpp
@@ -636,6 +636,7 @@ CFeatures* CDenseFeatures<ST>::copy_dimension_subset(SGVector<index_t> dims)
 	return result;
 }
 
+#ifndef SWIG // SWIG should skip this part
 template<class ST>
 CFeatures* CDenseFeatures<ST>::shallow_subset_copy()
 {
@@ -650,6 +651,7 @@ CFeatures* CDenseFeatures<ST>::shallow_subset_copy()
 	
 	return shallow_copy_features;
 }
+#endif
 
 template<class ST> ST* CDenseFeatures<ST>::compute_feature_vector(int32_t num, int32_t& len,
 		ST* target)
--- a/src/shogun/features/DenseFeatures.cpp
+++ b/src/shogun/features/DenseFeatures.cpp
@@ -642,7 +642,7 @@ CFeatures* CDenseFeatures<ST>::shallow_subset_copy()
 	CFeatures* shallow_copy_features=nullptr;
 	
 	SG_SDEBUG("Using underlying feature matrix with %d dimensions and %d feature vectors!\n", num_features, num_vectors);
-	SGMatrix<ST> shallow_copy_matrix = SGMatrix<ST>(feature_matrix.matrix, num_features, num_vectors, false);
+	SGMatrix<ST> shallow_copy_matrix(feature_matrix);
 	shallow_copy_features=new CDenseFeatures<ST>(shallow_copy_matrix);
 	if (m_subset_stack->has_subsets())
 		shallow_copy_features->add_subset(m_subset_stack->get_last_subset()->get_subset_idx());
--- a/src/shogun/features/DenseFeatures.cpp
+++ b/src/shogun/features/DenseFeatures.cpp
@@ -636,7 +636,6 @@ CFeatures* CDenseFeatures<ST>::copy_dimension_subset(SGVector<index_t> dims)
 	return result;
 }
 
-#ifndef SWIG // SWIG should skip this part
 template<class ST>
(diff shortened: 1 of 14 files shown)


src/shogun/features/DenseFeatures.cpp
```diff
@@ -636,6 +636,20 @@ CFeatures* CDenseFeatures<ST>::copy_dimension_subset(SGVector<index_t> dims)
 	return result;
 }
 
+template<class ST>
+CFeatures* CDenseFeatures<ST>::shallow_subset_copy()
+{
+	CFeatures* shallow_copy_features=nullptr;
+	
+	SG_SDEBUG("Using underlying feature matrix with %d dimensions and %d feature vectors!\n", num_features, num_vectors);
+	SGMatrix<ST> shallow_copy_matrix = SGMatrix<ST>(feature_matrix.matrix, num_features, num_vectors, false);
```
karlnapf: You can use the copy constructor of SGMatrix

src/shogun/features/DenseFeatures.cpp
```diff
@@ -636,6 +636,20 @@ CFeatures* CDenseFeatures<ST>::copy_dimension_subset(SGVector<index_t> dims)
...
+	shallow_copy_features=new CDenseFeatures<ST>(shallow_copy_matrix);
+	if (m_subset_stack->has_subsets())
```
karlnapf: seems good like this

src/shogun/labels/DenseLabels.cpp
```diff
@@ -203,3 +203,4 @@ int32_t CDenseLabels::get_num_labels() const
 	return m_subset_stack->has_subsets()
 			? m_subset_stack->get_size() : m_labels.vlen;
 }
+
```
karlnapf: please dont do such whitespace changes

src/shogun/labels/MulticlassLabels.cpp
```diff
@@ -134,3 +134,16 @@ int32_t CMulticlassLabels::get_num_classes()
 	SGVector<float64_t> unique=get_unique_labels();
 	return unique.vlen;
 }
+
+CLabels* CMulticlassLabels::shallow_subset_copy()
+{
+	CLabels* shallow_copy_labels=nullptr;
+	SGVector<float64_t> shallow_copy_vector = SGVector<float64_t>(m_labels.vector, m_labels.size(), false);
```
karlnapf: copy constr

src/shogun/labels/MulticlassLabels.cpp
```diff
@@ -134,3 +134,16 @@ int32_t CMulticlassLabels::get_num_classes()
 	SGVector<float64_t> unique=get_unique_labels();
 	return unique.vlen;
 }
+
+CLabels* CMulticlassLabels::shallow_subset_copy()
```
karlnapf: BTW for both of those methods: You either need to hide them from SWIG or put a `%newobject` directive in the interface file

src/shogun/multiclass/tree/CARTree.cpp
```diff
@@ -603,8 +603,6 @@ int32_t CCARTree::compute_best_attribute(const SGMatrix<float64_t>& mat, const S
 	{
 		SGVector<float64_t> feats(num_vecs);
 		SGVector<index_t> sorted_args(num_vecs);
```
karlnapf: are the changes in this file related to parallel? If not, pls separate pr

tests/unit/features/DenseFeatures_unittest.cc
```diff
@@ -119,3 +119,33 @@ TEST(DenseFeaturesTest, copy_dimension_subset_with_subsets)
 	SG_UNREF(features);
 	SG_UNREF(f_reduced);
 }
+
+TEST(DenseFeaturesTest, shallow_copy_subset_data)
+{
+	index_t dim=5;
+	index_t n=10;
+
+	SGMatrix<float64_t> data(dim, n);
```
karlnapf: Use iota on data.matrix. One line of code

tests/unit/features/DenseFeatures_unittest.cc
```diff
@@ -119,3 +119,33 @@ TEST(DenseFeaturesTest, copy_dimension_subset_with_subsets)
...
+	for (index_t i=0; i<dim; ++i)
+		for (index_t j=0; j<n; ++j)	
+			data(i,j)=CMath::random(0, n-1);
+	
+	SGVector<index_t> inds(n/2);
```
karlnapf: we have a permutation methods
range_fill (or iota) and then permute

tests/unit/features/DenseFeatures_unittest.cc
```diff
@@ -119,3 +119,33 @@ TEST(DenseFeaturesTest, copy_dimension_subset_with_subsets)
 	SG_UNREF(features);
 	SG_UNREF(f_reduced);
 }
+
+TEST(DenseFeaturesTest, shallow_copy_subset_data)
```
karlnapf: Need more tests.
Pls add some tests where you ensure that the objects can be used within threads. 
I.e. modifying the shallow_subset copy doesnt change another one...
etc

karlnapf:
So this works?
Whats the gain with increasing number of cores?

Saurabh7:
Yes, it works.
Although calculating oob error is leading to change in values for test. 
I suspected this is due to chaning of order in which we append oob indices (multi thread) to dynamic array. So I made it critical for the thread, but still there are some changes.

outputs: https://gist.github.com/Saurabh7/40889d51c856f73c33337408eb3a7fd1

karlnapf:
I guess you can force a single thread for the result tests.
Multithread tests can be done on the feature subset shallow copy alone, independently of rf

Saurabh7:
Ok so few things for this:
currently this Mock test which uses mock labels fails: https://github.com/shogun-toolbox/shogun/blob/develop/tests/unit/multiclass/BaggingMachine_unittest.cc#L35. because this shallow_subset_copy method is not implemented in Labels base class.

Regarding If objects are safe to use in other algorithms, `get_feature_matrix()` returns a copy if subset are present so thats not a problem.
other access methods like `get_feature_vector` and `set_feature_vector` will change the data if written using it.

Saurabh7:
Also I found the changes in oob error were because, randomly generated indices in multithreaded version keep chaning even thought I set seed in sg_rand.

src/shogun/labels/MulticlassLabels.cpp
karlnapf: this is still a blocker

karlnapf:
And merge :)

karlnapf:
ah crap, it wasnt squashed ;)
Can you always make sure you squash your commits before asking to merge them?
I know I should check, but there is a lot of stuff going on atm

Saurabh7:
@karlnapf ah ok sorry :) Next time I will put an explicit comment when final.

PR #3325, opened 2016-06-29
Fix c++0x error

cfjhallgren:


src/shogun/regression/KRRNystrom.cpp
```diff
@@ -68,7 +68,7 @@ SGVector<int32_t> CKRRNystrom::subsample_indices()
 	temp.range_fill();
 	CMath::permute(temp);
 	SGVector<int32_t> col(m_num_rkhs_basis);
-	for (auto i=0; i<m_num_rkhs_basis; ++i)
+	for (index_t i=0; i<m_num_rkhs_basis; ++i)
```
karlnapf: guarding would have done it as well, but this is OK

PR #3329, opened 2016-07-03
add create() in class_list

sanuj:
@lisitsyn shall i add swig interface?

### DIFF
--- a/src/interfaces/modular/SGBase.i
+++ b/src/interfaces/modular/SGBase.i
@@ -107,6 +107,7 @@ public void readExternal(java.io.ObjectInput in) throws java.io.IOException, jav
  #include <shogun/base/Version.h>
  #include <shogun/base/Parallel.h>
  #include <shogun/base/SGObject.h>
+ #include <shogun/base/class_list.h>
 
  extern void sg_global_print_message(FILE* target, const char* str);
  extern void sg_global_print_warning(FILE* target, const char* str);
@@ -325,6 +326,7 @@ namespace std {
 %include <shogun/io/SGIO.h>
 %include <shogun/base/Version.h>
 %include <shogun/base/Parallel.h>
+%include <shogun/base/class_list.h>
 
 #ifdef SWIGPYTHON
 namespace shogun
@@ -508,3 +510,7 @@ copy_reg._reconstructor=_sg_reconstructor
 
 
 #endif /* SWIGPYTHON  */
+
+using namespace shogun;
+
+%template(createKernel) create<CKernel>;
--- a/src/shogun/base/class_list.h
+++ b/src/shogun/base/class_list.h
@@ -12,7 +12,6 @@
 #define __SG_CLASS_LIST_H__
 
 #include <shogun/lib/config.h>
-
 #include <shogun/lib/DataType.h>
 
 namespace shogun {
@@ -24,6 +23,20 @@ namespace shogun {
 	 */
 	CSGObject* new_sgserializable(const char* sgserializable_name,
 										EPrimitiveType generic);
+
+    /** Creates an instance of a class identified by input name
+     * that is wrapped with a shared pointer like structure @ref Some
+     *
+     * @param name name of class
+     * @param type type for template class
+     * @return Some object that holds created instance of @ref T
+     */
+    template<class T>
+    T* create(const char* name,
+        EPrimitiveType type=PT_NOT_GENERIC)
+    {
+        return dynamic_cast<T*>(new_sgserializable(name, type));
+    }
 }
 
 #endif /* __SG_CLASS_LIST_H__  */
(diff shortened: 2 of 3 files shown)


tests/unit/base/class_list_unittest.cc
```diff
@@ -0,0 +1,26 @@
...
+#include <shogun/base/class_list.h>
+#include <shogun/kernel/GaussianKernel.h>
+#include <gtest/gtest.h>
+
+using namespace shogun;
+
+TEST(class_list, create_gaussian_kernel)
+{
+    const char* class_name = "GaussianKernel";
+    auto gk = create<CGaussianKernel>(class_name);
```
lisitsyn: `create<CKernel>`, this one is important for plugins switch

src/shogun/base/class_list.h
```diff
@@ -24,6 +24,18 @@ namespace shogun {
...
+    /** Returns an object of input class
+     * @param name name of class
+     * @param type type for template class
+     * @return object of the input class
+     */
+    template<class T>
+    T* create(const char* name,
+        EPrimitiveType type=PT_NOT_GENERIC)
+    {
+        return (T*)new_sgserializable(name, type);
```
lisitsyn: Please use `dynamic_cast` explicitly

src/shogun/base/class_list.h
```diff
@@ -24,6 +24,18 @@ namespace shogun {
...
 	CSGObject* new_sgserializable(const char* sgserializable_name,
 										EPrimitiveType generic);
+
+    /** Returns an object of input class
+     * @param name name of class
+     * @param type type for template class
+     * @return object of the input class
+     */
+    template<class T>
+    T* create(const char* name,
```
lisitsyn: Let it use `Some` instead of pointers

tests/unit/base/class_list_unittest.cc
```diff
@@ -0,0 +1,26 @@
...
+    auto another_gk = new CGaussianKernel();
+
+    EXPECT_EQ(strcmp(gk->get_name(), class_name), 0);
+    EXPECT_TRUE(gk->equals(another_gk));
+    delete another_gk;
+}
```
lisitsyn: No newline

src/interfaces/modular/SGBase.i
```diff
@@ -508,3 +510,7 @@ copy_reg._reconstructor=_sg_reconstructor
 
 
 #endif /* SWIGPYTHON  */
+
+using namespace shogun;
+
+%template(createKernel) create<CKernel>;
```
karlnapf: is this a prototype or why is this kernel only?

src/shogun/base/class_list.h
```diff
@@ -12,7 +12,6 @@
 #define __SG_CLASS_LIST_H__
 
 #include <shogun/lib/config.h>
-
```
karlnapf: pls dont do those. send separate patches for whitespace changes

src/shogun/base/class_list.h
```diff
@@ -24,6 +23,20 @@ namespace shogun {
 	 */
 	CSGObject* new_sgserializable(const char* sgserializable_name,
 										EPrimitiveType generic);
+
+    /** Creates an instance of a class identified by input name
+     * that is wrapped with a shared pointer like structure @ref Some
```
karlnapf: shared-pointer-like

tests/unit/base/class_list_unittest.cc
```diff
@@ -0,0 +1,26 @@
...
+#include <shogun/base/SGObject.h>
+#include <shogun/base/class_list.h>
+#include <shogun/kernel/GaussianKernel.h>
+#include <gtest/gtest.h>
+
+using namespace shogun;
+
+TEST(class_list, create_gaussian_kernel)
+{
+    const char* class_name = "GaussianKernel";
```
karlnapf: great this works!
This needs to be tested for *all* shogun classes.
See the unit tests for clone and equals, which we test for all classes as well

sanuj: @karlnapf, @lisitsyn wanted this. I think it's a proof of concept. `createX()` will be used when we will have plugins. It will load the concerned plugin and return an object. Currently we don't have plugins so I used `new_sgserializable`. @lisitsyn wanted to see if it works in python, and it is working. I plan to include this in my tags-swig PR (#3332) because the SWIG interface will be similar to set/get/has.

lisitsyn: @karlnapf @sanuj ok as it works lets remove swig part because we need that mighty list of base classes that should come from other PR. Not sure about all shogun classes.

sanuj:
Closing this, will send a PR against `feature/tags`.

PR #3330, opened 2016-07-03
Updated interface for dependency translation rules. No more duplicate…

sorig:
… imports.

Also adds documentation for translation rules.

### DIFF
--- a/examples/meta/generator/targets/doc.js
+++ b/examples/meta/generator/targets/doc.js
@@ -1,0 +1,239 @@
+/** Target files specify rules for translation from meta-language to a target
+ *  programming language. Each rule specifies the syntax of a specific language
+ *  construct in the target programming language. The rules are given in 
+ *  Python's template string syntax. Each keyword, prefixed with a "$", is
+ *  substituted by the associated translation which is obtained by recursive
+ *  applications of the translation rules. Some rules are optional.
+ *
+ *  This document lists all rules and the keywords available for each rule.
+ */
+{
+    /** "Root" translation rule
+     * Keywords:
+     * $program: String of all translated statements and comments
+     * $dependencies: Dependency string as specified by the rules in "Dependencies"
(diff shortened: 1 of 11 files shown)


examples/meta/generator/translate.py
```diff
@@ -209,67 +209,45 @@ def dependenciesString(self, allClasses, interfacedClasses, enums):
...
+        elif self.targetDict["Dependencies"].get("IncludeInterfacedClasses"):
+            dependencies = dependencies.union(interfacedClasses)
+        elif self.targetDict["Dependencies"].get("IncludeEnums"):
+            dependencies = dependencies.union(enums)
 
-        # Elements are just formatted as their names as default
-        elementTemplate = Template("$className")
-        if "DependencyListElementClass" in self.targetDict["Dependencies"]:
-            elementTemplate = Template(self.targetDict["Dependencies"]["DependencyListElementClass"])
+        translations = set(map(self.translateDependencyElement, dependencies))
```
sorig: @karlnapf Here I'm making sure that each translated dependency is unique. I.e. if two different dependencies (an enum and a class) result in the same import translation, then they're squashed.

examples/meta/generator/targets/doc.js
```diff
@@ -0,0 +1,239 @@
+/** Target files specify rules for translation from meta-language to a target
```
karlnapf: what type of file is this?
.js ?

karlnapf:
Great!
Lets merge once travis is green

examples/meta/generator/targets/doc.js
sorig: JavaScript. JSON does not support comments, so syntax highlighters make this hard to read. I could also change to markdown and wrap this in a code segment?

karlnapf: Lets discuss tomorrow

Issue #3328, opened 2016-07-01
Tests / illustration for meta language


karlnapf:
This task to to add a few meta language examples that serve the purpose of
 * testing the possible language constructs (like the ones I put for vectors)
 * illustrating what the meta language can do

In that sense it is both test and language illustration.

 * rename `tests` folder to `example` or `syntax_examples`
 * put in files with tests for everything we do in the meta langauge
  * enums
  * literals
  * Instantiating objects via constructors
  * assignment
  * ...

Think of some clever structure

Note to GSoC students. You can send such an example test patch instead a normal coobook if you wish. 
@sanuj @OXPHOS @Saurabh7

sanuj:
@karlnapf Can you please link to your vectors test, I can't seem to find it.  Which `tests` folder has to be renamed to `example` or `syntax_examples`? Were you talking about `shogun/examples/meta/generator/tests`?

karlnapf:
It is here. I put some tests already but need to make things work with travis before merging them.
You can also send PRs here to increase passing tests

https://travis-ci.org/shogun-toolbox/shogun/builds/145334402
https://github.com/shogun-toolbox/shogun/compare/develop...738612febb82

karlnapf:
@sorig do we have all meta language constructs covered, or do we need more?

PR #3332, opened 2016-07-04
add SWIG interface for tags

sanuj:
A few questions:
- How to test this:question: Shall I add meta-language code for testing:question:
- Where to include MockObject in SWIG:question: This will be required for testing.
- Commented out `SG_INSTA` are abstract templates in shogun. @lisitsyn wanted to discuss something about these with @karlnapf or @vigsterkr.

`SG_INSTA` stands for `shogun instantiate`. Can change the name if someone proposes a better one. :smile: 

This code works in Python. [Here](https://gist.github.com/sanuj/e24d003b5af990fee1b772129ace75fa) is some old working example but it won't work with latest changes like registering arbitrary parameters is not allowed now. Will update this example as soon as I know where to put MockObject in SWIG.

### DIFF
--- a/src/interfaces/modular/SGBase.i
+++ b/src/interfaces/modular/SGBase.i
@@ -1,6 +1,7 @@
 /* base includes required by any module */
 %include "stdint.i"
 %include "exception.i"
+%include "std_string.i"
 
 %feature("ref")   shogun::CSGObject "SG_REF($this);"
 %feature("unref") shogun::CSGObject "SG_UNREF($this);"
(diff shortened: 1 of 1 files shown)


sanuj:
Come on ruby! :gem:

karlnapf:
You can add meta language examples for testing, yes.
If you find that something is incomplete, let me know.
Pls do this in the way I described in #3328 i.e. create a file that is a test and an api illustration for tags at the same time

karlnapf:
No idea about the mock object .... @lisitsyn  ?

karlnapf:
The classes in there are *way* to specific. I dont understand at all why this is done? Can you explain a bit?

sanuj:
@karlnapf These are abstract classes in shogun i.e. they have a pure virtual function. Now the plan has changed. @lisitsyn and I decided to create a new jinja file which will have SWIG interface for set/get/has/Tag and this file will be included in `SGBase.i`. The jinja file will read the names (X) of classes from `shogun-base.txt` and generate code for setX() / getX() / hasX() / TagX in SWIG. Initially we will have 3-4 classes in `shogun-base.txt` and we can keep adding on classes later. `shogun-base.txt` will have list of classes that we want to keep in base shogun.

sanuj:
:eyes: See `add swig for tags using Jinja` #3357.

PR #3221, opened 2016-05-28
Implement tag-based parameters in SGObject

sanuj:
Temporary PR for @lisitsyn.

PR #3333, opened 2016-07-04
change method names, fix bug and docs

sanuj:
Includes updates for latest review of #3221 and a small bug fix which I found while testing for SWIG interface.


karlnapf:
Can be merged once travis is green

PR #3334, opened 2016-07-04
LinalgRefactor-mean

OXPHOS:


karlnapf:
Fine to merge from my side
@lambday @vigsterkr I leave to you

OXPHOS:
@karlnapf @lambday updated

src/shogun/mathematics/linalg/LinalgBackendBase.h
```diff
@@ -95,6 +95,20 @@ class LinalgBackendBase
 	#undef BACKEND_GENERIC_DOT
 
 	/**
+	* Wrapper method that computes mean of SGVectors and SGMatrices.
+	*
+	* @see linalg::mean
+	*/
+	#define BACKEND_GENERIC_MEAN(Type, Container) \
+	virtual Type mean(const Container<Type>& a) const \
```
vigsterkr: so are we sure that the mean of a `SGVector<int32_t>` should be expressed as an `int32_t` and not at least a `float32_t`? because that's not really a mean...

OXPHOS: @vigsterkr updated - return float mean result from int vectors/matrices

vigsterkr:
cant we just simply have mean that returns always float64_t?

vigsterkr:
i.e. i dont really get why we really need that enum story.

src/shogun/mathematics/linalg/LinalgBackendBase.h
vigsterkr: yeah but why not just:
`virtual float64_t mean(const Container<Type>& a) const`?

OXPHOS:
so all types will be converted to float64_t? I don't think it will work for the complex.

vigsterkr:
in that case you could use something like:
```
template < typename = typename std::enable_if< std::is_same<T, complex_t>::type >::type >
complex_t mean(T);

template < T >
float64_t mean(T);
```

OXPHOS:
Ah cool! Lemme try

karlnapf:
Updates?

OXPHOS:
I'm having some trouble using `typename std::enable_if<std::is_same<Type, complex128_t>::type, complex128_t>::type` in macros.
I did:
https://gist.github.com/OXPHOS/8f2d048e71df1dfc98bb7e2f0953c30e
but I got several errors:
http://pastebin.com/EeYvAay1
I'm not sure why this is happening.

vigsterkr:
you cannot have this:
https://gist.github.com/OXPHOS/8f2d048e71df1dfc98bb7e2f0953c30e#file-linalgnamespace-h-L10

src/shogun/mathematics/linalg/LinalgNamespace.h
```diff
@@ -129,19 +129,34 @@ T dot(const SGVector<T>& a, const SGVector<T>& b)
...
+ * @return the vector mean \f$\bar{a}_i\f$ or matrix mean \f$\bar{m}_{i,j}\f$
+ */
+template <typename T, template <typename> class Container>
+typename std::enable_if<!std::is_same<T, complex128_t>::value, float64_t>::type
+mean(const Container<T>& a)
+{
+	REQUIRE(a.size() > 0, "Vector/Matrix cannot be empty!\n");
+	return std::abs(infer_backend(a)->mean(a));
+}
+
```
OXPHOS: What I had here is calculating mean with only one method and cast the result to different type. But I realized that ViennaCL doesn't work that well with complex data and we'll have problem with colwise/rowwise mean operation.

Another way is to have independent `complex_mean` and `float_mean` in the backends implementation.

vigsterkr: no you are getting this a bit wrong. no need for casting or anything, every interface (complex vs rest) should just call the right function. so in other words when you have:
```
template <typename T, template <typename> class Container>
complex128_t mean_impl(const Container<T>& a) const;
```

then it should actually call:
```
complex64_t sum_impl(Container<complex128_t>& a)
```
or use `enabled_if`, but don't enable this for *every* `T`.

you should not define a generic:
```
template <typename T, template <typename> class Container>
complex128_t mean_impl(const Container<T>& a) const;
```
it should only be defined for `T = complex128_t`.

OXPHOS:
@vigsterkr @karlnapf  updated

tests/unit/mathematics/linalg/operations/Viennacl_operations_unittest.cc
```diff
@@ -51,6 +51,35 @@ TEST(LinalgBackendViennaCL, dot)
...
+
+	EXPECT_NEAR(result, 2.5, 1E-15);
+}
+
+TEST(LinalgBackendViennaCL, SGMatrix_mean)
+{
+	sg_linalg->set_gpu_backend(new LinalgBackendViennaCL());
+
+	const index_t nrows = 2, ncols = 3;
+	SGMatrix<int32_t> mat(nrows, ncols), mat_gpu(nrows, ncols);
```
karlnapf: pls dont do 2 variables per line

karlnapf:
Fine to merge from my side

karlnapf:
Ill leave it to @vigsterkr 
@OXPHOS ping me if not merged by later today then Ill do it

OXPHOS:
squashed

OXPHOS:
@karlnapf wants to merge this?

PR #3336, opened 2016-07-04
LinalgRefactor-SGVector&SGMatrix-max

OXPHOS:


src/shogun/mathematics/linalg/LinalgBackendViennacl.h
```diff
@@ -40,6 +40,7 @@
 #include <viennacl/vector.hpp>
 #include <viennacl/linalg/inner_prod.hpp>
 #include <shogun/mathematics/linalg/GPUMemoryViennaCL.h>
+#include <shogun/mathematics/linalg/internal/opencl_util.h>
```
OXPHOS: Not sure whether this file needs to be moved around/refactored

src/shogun/mathematics/linalg/LinalgBackendViennacl.h
```diff
@@ -115,6 +186,27 @@ class LinalgBackendViennaCL : public LinalgBackendGPUBase
...
+		GPUMemoryViennaCL<T>* vector_gpu =
+			static_cast<GPUMemoryViennaCL<T>*>(vec.gpu_vector.get());
+
+		viennacl::ocl::enqueue(kernel(vector_gpu->data(vec.size()),
+			cl_int(vec.size()), cl_int(vector_gpu->m_offset),
+			result_gpu->data(1)));
+
+		SGVector<T> result_cpu = linalg::from_gpu(new SGVector<T>(result_gpu, 1));
+		return result_cpu[0];
+	}
```
OXPHOS: - `GPUMemoryViennaCL<T>` can only be accessed (initialized) from existed `SGVector<T>`, which makes the code complicated. To save the result, I need construct a CPU vector with fixed length and value, and then transfer it to the GPU. Things will be easier if there's variable in `GPUMemoryViennaCL<T>` to hold the vector length.

- There's problem with `SGVector<T>(GPUMemoryBase<T>* vector, index_t len)` constructor. As the cpu vector pointer is not allocated, it cannot be freed. (`malloc: *** error for object 0x7fe0c2c18320: pointer being freed was not allocated`). I still need to work on it.

src/shogun/mathematics/linalg/LinalgBackendViennacl.h
```diff
@@ -115,6 +186,27 @@ class LinalgBackendViennaCL : public LinalgBackendGPUBase
...
 
+	/** Return the largest element in the vector with ViennaCL library */
+	template <typename T>
+	T max_impl(const SGVector<T>& vec) const
+	{
+		viennacl::ocl::kernel& kernel = generate_kernel<T>();
+
+		SGVector<T> result(1);
+		result.set_const(-1);
+		GPUMemoryViennaCL<T>* result_gpu =
```
karlnapf: maybe we want a private helper method for this?

src/shogun/mathematics/linalg/LinalgBackendViennacl.h
```diff
@@ -115,6 +186,27 @@ class LinalgBackendViennaCL : public LinalgBackendGPUBase
...
 	}
 
+	/** Return the largest element in the vector with ViennaCL library */
+	template <typename T>
+	T max_impl(const SGVector<T>& vec) const
+	{
+		viennacl::ocl::kernel& kernel = generate_kernel<T>();
+
+		SGVector<T> result(1);
+		result.set_const(-1);
```
karlnapf: why this?

src/shogun/mathematics/linalg/LinalgBackendViennacl.h
```diff
@@ -115,6 +186,27 @@ class LinalgBackendViennaCL : public LinalgBackendGPUBase
 		return viennacl::linalg::inner_prod(a_gpu->data(a.size()), b_gpu->data(b.size()));
 	}
 
+	/** Return the largest element in the vector with ViennaCL library */
+	template <typename T>
+	T max_impl(const SGVector<T>& vec) const
+	{
+		viennacl::ocl::kernel& kernel = generate_kernel<T>();
```
karlnapf: @lambday so this kernel is compiled everytime we call this operation?
Is that good?

src/shogun/mathematics/linalg/LinalgBackendViennacl.h
karlnapf: should be ok but @lambday needs to comment as well

src/shogun/mathematics/linalg/LinalgBackendViennacl.h
OXPHOS: gives error:
error for object 0x7ff2bc0b9af0: pointer being freed was not allocated
in unit-test

vigsterkr:
@OXPHOS pingu

PR #3342, opened 2016-07-05
Shared features for xval

Saurabh7:
Not merge ready


Saurabh7:
@karlnapf  simple changes if one wants to use the shallow copy.

karlnapf:
Yep pretty easy.
Not sure what the default should be.

I have the feeling that shogun develiopers should decide this. So we can add a flag to CMachine that it only uses features read-only, and then set it to false by default and override where we know.

But I feel there is a better way
@vigsterkr @lisitsyn ?

karlnapf:
Features will be read only soon

PR #3350, opened 2016-07-06
LinalgRefactor - SGVector&SGMatrix - setconst

OXPHOS:


tests/unit/mathematics/linalg/operations/Viennacl_operations_unittest.cc
```diff
@@ -49,4 +49,23 @@ TEST(LinalgBackendViennaCL, dot)
...
+	for (index_t i = 0; i < size; ++i)
+		a[i] = value;
+
+	b_gpu = to_gpu(b);
+	set_const(b_gpu, value);
+	b_cpu = from_gpu(b_gpu);
+
+	for (index_t i = 0; i < size; ++i)
+		EXPECT_NEAR(a[i], b_cpu[i], 1E-15);
+}
```
OXPHOS: This doesn't look very right..
To set a gpu vector, ideally an empty gpu vector with malloced space should be provided.
However, there's no way to initiate such gpu vector from the SGVector interface.

Now, a cpu vector with unknown content is transferred to gpu and the gpu memory is reset by `value`.

src/shogun/mathematics/linalg/LinalgBackendEigen.h
```diff
@@ -111,6 +133,14 @@ class LinalgBackendEigen : public LinalgBackendBase
 		return (typename SGVector<T>::EigenVectorXtMap(a)).dot(typename SGVector<T>::EigenVectorXtMap(b));
 	}
 
+	/** Set const method */
+	template <typename T>
+	void set_const_impl(SGVector<T>& vec, T value) const
+	{
+		for (index_t i = 0; i < vec.size(); ++i)
+			vec[i] = value;
```
lambday: isn't `std::fill` faster?

tests/unit/mathematics/linalg/operations/Viennacl_operations_unittest.cc
lambday: do you get "reading uninitialized values" error on valgrind?
however, for this unit-test's purpose, you can set the value const to 0 for the cpu vector, transfer to gpu, set the value to something else, transfer back and test. that will get rid of this error.

src/shogun/mathematics/linalg/LinalgBackendBase.h
```diff
@@ -104,6 +104,33 @@ class LinalgBackendBase
 	#undef BACKEND_GENERIC_DOT
 
 	/**
+	 * Wrapper method that set const values to vectors.
+	 *
+	 * @see linalg::set_const
+	 */
+	#define BACKEND_GENERIC_SET_CONST(Type) \
```
karlnapf: This will also be available for both vector and matrix I guess?

karlnapf:
Generally fine, but pls also apply the changes discussed in the matrix PR here

OXPHOS:
@vigsterkr ready to merge

PR #3356, opened 2016-07-07
Automatically genereated parser files are now stored in the build dir…

sorig:
…ectory.

`parse.py` and `generate.py` now have the command line option `--parser_files_dir` that allows users to specify where automatically generated parser files should be stored. When running `make meta_examples` the files are stored under `build/examples/meta/parser_files/`. If `parse.py` is run without the new command line argument, the parser is run in unoptimised mode (slower) and no parser files are generated.


karlnapf:
This is Great! Cleaner!

PR #3360, opened 2016-07-10
rf reg cookbook

Saurabh7:


Saurabh7:
@karlnapf I will send data,
but I think its good to set num of threads here in some way, currently 

```
rand_forest.parallel.set_num_threads(1)
rand_forest.train(features_train)
```
gives

```
/shogun/build/examples/meta/cpp/regression/random_forest_regression.cpp:16:1: error: ‘rand_forest’ was not declared in this scope
 rand_forest->train(features_train);

```

karlnapf:
Ah yeah, I think we need some SWIG compatible wrapper method to set the number of threads. Like for init seed.
This thing you do is not compatible with the meta grammar.
I assume get_global_parallel doesnt work?

doc/cookbook/source/examples/regression/random_forest_regression.rst
```diff
@@ -0,0 +1,42 @@
+=============
+Random Forest
+=============
+
+A Random Forest is an ensemble learning method which implements multiple decision trees during training. It predicts by using a combination rule on the outputs of individual decision trees.
+
+See :cite:`Breiman2001` for a detailed introduction.
```
karlnapf: pls reference the classification cookbook

examples/meta/src/regression/random_forest_regression.sg
```diff
@@ -0,0 +1,36 @@
...
+RegressionLabels labels_train(f_labels_train)
+RegressionLabels labels_test(f_labels_test)
+#![create_features]
+
+#![create_combination_rule]
+MeanRule mean_rule()
+#![create_combination_rule]
+
+#![create_instance]
+RandomForest rand_forest(1,10)
```
karlnapf: can you make 1 and 10 explicit variables and mention them?

karlnapf:
Pretty good, this is mergable once the minor things are addressed,
we can hold the integration data back until we sorted out the thread setting...

Saurabh7:
updated. 
ah something else :hand:  `:doc:random_forest` doesnt link properly here just saw http://buildbot.shogun-toolbox.org/cookbook_pr/208f4b16af90cd66ef38fe5c434f3d781ad147e8/examples/regression/random_forest_regression.html.

PR #3372, opened 2016-07-16
fixed the variational notebook

yorkerlin:


yorkerlin:
@karlnapf 
From now on, I will focus on re-factoring the optimization framework.
For GP, the next new features will be structured GP from Andrew's works. I am waiting for his KISS-GP code.

vigsterkr:
@yorkerlin before jumping in alone, could we have like a design session (irc,gdocs, gist) where we can agree on the optimization framework refactoring?

yorkerlin:
yes. Maybe Monday after the GSOC meeting? 
Some ideas:
* using variadic templates
* add parallel coordinate descent solver from https://github.com/uclaopt/TMAC

vigsterkr:
my pain points:
 * kkt solver
 * use the linalg backend so we can use various linalg backends.
 * maybe instead of directly addign TMAC (looks great) work on a more general design?

yorkerlin:
I merge this PR.

PR #3373, opened 2016-07-17
fixed a memory leak bug

yorkerlin:


yorkerlin:
@karlnapf 
We should use c++11 soon.
The main reason of this issue is the `optimization` is not a `SGObject.`

yorkerlin:
One issue about changing Minimizer as a sub-class of SGObject:
Some minimizer uses some external lib (eg, LBFGS, and NLOPT) and some type from the external lib.
It seems to be hard to serialize this external type.  (eg, https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/optimization/NLOPTMinimizer.h#L128) 
Any idea about this issue? @vigsterkr  @lisitsyn  @lambday

karlnapf:
Make it a sub-class then....As I said, the motivation to not do this earlier is not existing anymore.

karlnapf:
If you just dont do SG_ADD, then varialbes will be ignored when serializing.....

yorkerlin:
I close this due to https://github.com/shogun-toolbox/shogun/pull/3380

PR #3375, opened 2016-07-18
Cereal serialization - SGObejct

OXPHOS:
class Any serialization.

- Changes in any.h and any.cpp
- Need rebase on #3226

### DIFF
--- a/src/shogun/base/SGObject.cpp
+++ b/src/shogun/base/SGObject.cpp
@@ -28,6 +28,13 @@
 #include <stdlib.h>
 #include <stdio.h>
 
+#include <cereal/cereal.hpp>
+#include <cereal/types/polymorphic.hpp>
+#include <cereal/archives/binary.hpp>
+#include <cereal/archives/json.hpp>
+#include <cereal/archives/xml.hpp>
+#include <fstream>
+
 #ifdef HAVE_CXX11
 #include <unordered_map>
 #else
@@ -211,6 +218,68 @@ int32_t CSGObject::unref()
 	}
 }
 #endif //USE_REFERENCE_COUNTING
+
+void CSGObject::save_binary(const char* filename) const
+{
+	std::ofstream os(filename);
+	cereal::BinaryOutputArchive archive(os);
+	archive(*this);
+}
+
+void CSGObject::save_json(const char* filename) const
+{
+	std::ofstream os(filename);
+	cereal::JSONOutputArchive archive(os);
+	archive(*this);
+}
+
+void CSGObject::save_xml(const char* filename) const
+{
+	std::ofstream os(filename);
+	cereal::XMLOutputArchive archive(os);
+	archive(*this);
+}
+
+void CSGObject::load_binary(const char* filename)
+{
+	std::ifstream is(filename);
+	cereal::BinaryInputArchive archive(is);
+	archive(*this);
+}
+
+void CSGObject::load_json(const char* filename)
+{
+	std::ifstream is(filename);
+	cereal::JSONInputArchive archive(is);
+	archive(*this);
+}
+
+void CSGObject::load_xml(const char* filename)
+{
+	std::ifstream is(filename);
+	cereal::XMLInputArchive archive(is);
+	archive(*this);
+}
+
+template<class Archive>
+void CSGObject::cereal_save(Archive& ar) const
+{
+	for (const auto& it:self->map)
+		ar(cereal::make_nvp(it.first.name(), it.second));
+}
+template void CSGObject::cereal_save<cereal::BinaryOutputArchive>(cereal::BinaryOutputArchive& ar) const;
+template void CSGObject::cereal_save<cereal::JSONOutputArchive>(cereal::JSONOutputArchive& ar) const;
+template void CSGObject::cereal_save<cereal::XMLOutputArchive>(cereal::XMLOutputArchive& ar) const;
+
+template<class Archive>
+void CSGObject::cereal_load(Archive& ar)
+{
+	for (auto& it:self->map)
+		ar(it.second);
+}
+template void CSGObject::cereal_load<cereal::BinaryInputArchive>(cereal::BinaryInputArchive& ar);
+template void CSGObject::cereal_load<cereal::JSONInputArchive>(cereal::JSONInputArchive& ar);
+template void CSGObject::cereal_load<cereal::XMLInputArchive>(cereal::XMLInputArchive& ar);
 
 #ifdef TRACE_MEMORY_ALLOCS
 #include <shogun/lib/Map.h>
--- a/src/shogun/base/SGObject.h
+++ b/src/shogun/base/SGObject.h
@@ -124,6 +124,58 @@ public:
 	/** destructor */
 	virtual ~CSGObject();
 
+	/** serializes the SGObject to a binary file
+	 *
+	 * @param filename Binary archive filename
+	 */
+	void save_binary(const char* filename) const;
+
+	/** serializes the SGObject to a JSON file
+	 *
+	 * @param filename JSON archive filename
+	 */
+	void save_json(const char* filename) const;
+
+	/** serializes the SGObject to a XML file
+	 *
+	 * @param filename XML archive filename
+	 */
+	void save_xml(const char* filename) const;
+
+	/** loads SGObject from a Binary file
+	 *
+	 * @param filename Binary archive filename
+	 */
+	void load_binary(const char* filename);
+
+	/** loads SGObject from a JSON file
+	 *
+	 * @param filename JSON archive filename
+	 */
+	void load_json(const char* filename);
+
+	/** loads SGObject from a XML file
+	 *
+	 * @param filename XML archive filename
+	 */
+	void load_xml(const char* filename);
+
+#ifndef SWIG // SWIG should skip this part
+	/** serializes SGObject parameters to Archive with Cereal
+	 *
+	 * @param ar Archive
+	 */
+	template<class Archive>
+	void cereal_save(Archive& ar) const;
+
+	/** loads SGObject parameters from Archive with Cereal
+	 *
+	 * @param ar Archive
+	 */
+	template<class Archive>
+	void cereal_load(Archive& ar);
+#endif // #ifndef SWIG // SWIG should skip this part
+
 #ifdef USE_REFERENCE_COUNTING
 	/** increase reference counter
 	 *
--- a/src/shogun/lib/SGReferencedData.cpp
+++ b/src/shogun/lib/SGReferencedData.cpp
@@ -1,6 +1,12 @@
 #include <shogun/lib/SGReferencedData.h>
 #include <shogun/lib/RefCount.h>
 
+#include <cereal/cereal.hpp>
+#include <cereal/types/polymorphic.hpp>
+#include <cereal/archives/binary.hpp>
+#include <cereal/archives/json.hpp>
+#include <cereal/archives/xml.hpp>
+
 using namespace shogun;
 
 namespace shogun {
@@ -112,4 +118,36 @@ int32_t SGReferencedData::unref()
 		return c;
 	}
 }
+
+template<class Archive>
+void SGReferencedData::cereal_save(Archive & ar) const
+{
+	if (m_refcount != NULL)
+		ar(cereal::make_nvp("ref_counting", true), m_refcount->ref_count());
+	else
+		ar(cereal::make_nvp("ref_counting", false));
+}
+template void SGReferencedData::cereal_save<cereal::BinaryOutputArchive>(cereal::BinaryOutputArchive& ar) const;
+template void SGReferencedData::cereal_save<cereal::JSONOutputArchive>(cereal::JSONOutputArchive& ar) const;
+template void SGReferencedData::cereal_save<cereal::XMLOutputArchive>(cereal::XMLOutputArchive& ar) const;
+
+template<class Archive>
+void SGReferencedData::cereal_load(Archive & ar)
+{
+	bool ref_counting;
+	ar(ref_counting);
+
+	if (ref_counting)
+	{
+		int32_t temp;
+		ar(temp);
+		m_refcount = new RefCount(temp);
+	}
+	else
+		m_refcount = NULL;
+}
+template void SGReferencedData::cereal_load<cereal::BinaryInputArchive>(cereal::BinaryInputArchive& ar);
+template void SGReferencedData::cereal_load<cereal::JSONInputArchive>(cereal::JSONInputArchive& ar);
+template void SGReferencedData::cereal_load<cereal::XMLInputArchive>(cereal::XMLInputArchive& ar);
+
 }
--- a/src/shogun/lib/SGReferencedData.h
+++ b/src/shogun/lib/SGReferencedData.h
@@ -43,6 +43,14 @@ class SGReferencedData
 		 */
 		int32_t ref_count();
 
+#ifndef SWIG // SWIG should skip this part
+		template<class Archive>
+		void cereal_save(Archive & ar) const;
+
+		template<class Archive>
+		void cereal_load(Archive & ar);
+#endif //#ifndef SWIG // SWIG should skip this part
+
 	protected:
 		/** copy refcount */
 		void copy_refcount(const SGReferencedData &orig);
--- a/src/shogun/lib/SGVector.h
+++ b/src/shogun/lib/SGVector.h
@@ -474,6 +474,12 @@ template<class T> class SGVector : public SGReferencedData
 		 * @return matrix
 		 */
 		static void convert_to_matrix(T*& matrix, index_t nrows, index_t ncols, const T* vector, int32_t vlen, bool fortran_order);
+
+		template<class Archive> void cereal_save(Archive & ar) const;
+
+		template<class Archive>
+		void cereal_load(Archive & ar);
+
 #endif // #ifndef SWIG // SWIG should skip this part
 	protected:
 		/** needs to be overridden to copy data */
--- a/src/shogun/lib/any.h
+++ b/src/shogun/lib/any.h
@@ -39,9 +39,83 @@
 #include <stdexcept>
 #include <typeinfo>
 #include <cxxabi.h>
+#include <shogun/io/SGIO.h>
 
 namespace shogun
 {
+
+#ifndef SWIG // SWIG should skip this part
+	namespace serial
+	{
+		enum EnumContainerType
+		{
+			CT_UNDEFINED,
+			CT_PRIMITIVE,
+			CT_SGVECTOR,
+			CT_SGMATRIX
+		};
+
+		enum EnumPrimitiveType
+		{
+			PT_UNDEFINED,
+			PT_INT_32,
+			PT_FLOAT_64,
+		};
+
+		/** cast data type to EnumContainerType and EnumPrimitiveType */
+		template<typename T>
+		struct Type2Enum
+		{
+			static constexpr EnumContainerType e_containertype = CT_UNDEFINED;
+			static constexpr EnumPrimitiveType e_primitivetype = PT_UNDEFINED;
+		};
+
+		template<>
+		struct Type2Enum<int32_t>
+		{
+			static constexpr EnumContainerType e_containertype = CT_PRIMITIVE;
+			static constexpr EnumPrimitiveType e_primitivetype = PT_INT_32;
+		};
+
+		template<>
+		struct Type2Enum<float64_t>
+		{
+			static constexpr EnumContainerType e_containertype = CT_PRIMITIVE;
+			static constexpr EnumPrimitiveType e_primitivetype = PT_FLOAT_64;
+		};
+
+		template<>
+		struct Type2Enum<SGVector<int32_t> >
+		{
+			static constexpr EnumContainerType e_containertype = CT_SGVECTOR;
+			static constexpr EnumPrimitiveType e_primitivetype = PT_INT_32;
+		};
+
+		template<>
+		struct Type2Enum<SGVector<float64_t> >
+		{
+			static constexpr EnumContainerType e_containertype = CT_SGVECTOR;
+			static constexpr EnumPrimitiveType e_primitivetype = PT_FLOAT_64;
+		};
+
+		/** @brief data structure that saves the EnumContainerType
+		 * and EnumPrimitiveType information of Any object
+		 */
+		struct DataType
+		{
+			EnumContainerType e_containertype;
+			EnumPrimitiveType e_primitivetype;
+
+			template<typename T>
+			void set()
+			{
+				e_containertype = Type2Enum<T>::e_containertype;
+				e_primitivetype = Type2Enum<T>::e_primitivetype;
+			}
+		};
+	}
+#endif //SWIG
+
 	/** Converts compiler-dependent name of class to
 	 * something human readable.
 	 * @return human readable name of class
@@ -172,12 +246,14 @@ namespace shogun
 		explicit Any(const T& v) : policy(select_policy<T>()), storage(nullptr)
 		{
 			policy->set(&storage, &v);
+			m_datatype.set<T>();
 		}
 
 		/** Copy constructor */
 		Any(const Any& other) : policy(other.policy), storage(nullptr)
 		{
 			policy->set(&storage, other.storage);
+			m_datatype = other.m_datatype;
 		}
 
 		/** Assignment operator
@@ -189,6 +265,7 @@ namespace shogun
 			policy->clear(&storage);
 			policy = other.policy;
 			policy->set(&storage, other.storage);
+			m_datatype = other.m_datatype;
 			return *(this);
 		}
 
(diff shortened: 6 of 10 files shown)


src/shogun/lib/any.cpp
```diff
@@ -0,0 +1,28 @@
+#include <shogun/lib/any.h>
+
+template<>
+Any::EnumType Any::type2enum<int>()
```
vigsterkr: you mean `int32_t`?

src/shogun/lib/any.cpp
```diff
@@ -0,0 +1,28 @@
...
+{
+	return TYPE_INT;
+}
+
+template<>
+Any::EnumType Any::type2enum<double>()
```
vigsterkr: `float64_t` ?

src/shogun/lib/any.cpp
```diff
@@ -0,0 +1,28 @@
...
+{
+	constexpr EnumType type = TYPE_DOUBLE;
+	return type;
+}
+
+template<>
+Any::EnumType Any::type2enum<SGVector<int>>()
```
vigsterkr: same as above

src/shogun/lib/any.cpp
```diff
@@ -0,0 +1,28 @@
...
+{
+	constexpr EnumType type = TYPE_SGVECTOR_INT;
+	return type;
+}
+
+template<>
+Any::EnumType Any::type2enum<SGVector<double>>()
```
vigsterkr: same as above

src/shogun/lib/any.h
```diff
@@ -212,6 +214,24 @@ namespace shogun
             policy->clear(&storage);
         }
 
+		template<class Archive>
+		void cereal_save(Archive& ar) const
+		{
+			switch (enumtype) {
+			case TYPE_INT:
+				ar(*(reinterpret_cast<int*>(storage))); break;
```
vigsterkr: please try to use the types that we use throughout in shogun...

src/shogun/lib/any.cpp
OXPHOS: oh i **was** trying to use `policy.type()` which only gives only int or double. I can be more specific if this strategy is okay

tests/unit/io/Cereal_unittest.cc
```diff
@@ -0,0 +1,51 @@
...
+#include <cereal/archives/json.hpp>
+#include <fstream>
+
+
+using namespace shogun;
+
+TEST(Cereal, Json_SGVector_FLOAT64_save_load_pair)
+{
+	SGVector<float64_t> a(2), b;
+	a.set_const(1.14263158);
```
karlnapf: pls never use const vectors in unit tests, just as in linalg

tests/unit/io/Cereal_unittest.cc
```diff
@@ -0,0 +1,77 @@
...
+	SGVector<float64_t> vec(5);
+	vec.range_fill(0);
+	Any a(vec);
+
+	{
+		std::ofstream os("cereatltest_save_anyobject.cereal");
+		cereal::JSONOutputArchive archive(os);
+		archive(a);
+	}
+}
```
OXPHOS: This test works and I got:
```
{
    "value0": {
        "value0": {
            "value0": 5,
            "value1": 0,
            "value2": 1,
            "value3": 2,
            "value4": 3,
            "value5": 4
        }
    }
}
```
in which value0 is the length of the vector
value1 - 5 are the entries of the sgvector

tests/unit/io/Cereal_unittest.cc
```diff
@@ -0,0 +1,77 @@
...
+	CSerializableJsonFile *file=new CSerializableJsonFile("cerealtest_param_save_cerealobject.json", 'w');
+	obj_save.save_serializable(file);
+	file->close();
+#endif
+
+	{
+		std::ofstream os("cereatltest_save_cerealobject.cereal");
+		cereal::JSONOutputArchive archive(os);
+		archive(obj_save);
+	}
```
OXPHOS: I got file:
```
{
    "value0": {
        "value0": {}
    }
}
```

tests/unit/io/Cereal_unittest.cc
```diff
@@ -0,0 +1,77 @@
...
+}
+
+
+TEST(Cereal, Json_CerealObject_save_method)
+{
+	CCerealObject obj_save;
+
+#ifdef HAVE_JSON
+	CSerializableJsonFile *file=new CSerializableJsonFile("cerealtest_param_save_cerealobject.json", 'w');
+	obj_save.save_serializable(file);
```
OXPHOS: I got 
```
{"filetype":"_SHOGUN_SERIALIZABLE_JSON_FILE_V_00_","vector":{"type":"SGVector<float64>","data":[]}}
```
which doesn't seem right

OXPHOS:
@karlnapf @vigsterkr please ignore the problems i mentioned a while ago - it's solved and now the saving is working.

I'll split the enum types and add load methods tomorrow.

tests/unit/base/CerealObject.h
```diff
@@ -0,0 +1,28 @@
...
+        init_params();
+    }
+
+    const char* get_name() const { return "CerealObject"; }
+
+protected:
+    void init_params()
+    {
+        SGVector<float64_t> vec(5);
+		vec.range_fill(0);
```
karlnapf: indentation

tests/unit/base/CerealObject.h
```diff
@@ -0,0 +1,28 @@
+#include <shogun/base/SGObject.h>
+#include <shogun/lib/SGVector.h>
+
+namespace shogun
+{
+
+/** @brief Used to test the SGObject serialization */
```
karlnapf: yep thats exactly how I imagined it :)

tests/unit/base/CerealObject.h
```diff
@@ -0,0 +1,28 @@
...
+    const char* get_name() const { return "CerealObject"; }
+
+protected:
+    void init_params()
+    {
+        SGVector<float64_t> vec(5);
+		vec.range_fill(0);
+
+		register_param("vector", vec);
+		SG_ADD(&vec, "vector", "test vector", MS_NOT_AVAILABLE);
```
karlnapf: I think no pointer here, but not sure

tests/unit/io/Cereal_unittest.cc
karlnapf: ah so this works

karlnapf: so goes this go over cereal? No I guess it uses the old code no?

tests/unit/io/Cereal_unittest.cc
karlnapf: might be the SG_ADD with pointer is wrong

src/shogun/base/SGObject.cpp
```diff
@@ -471,6 +473,23 @@ void CSGObject::save_serializable_post() throw (ShogunException)
 	m_save_post_called = true;
 }
 
+template<class Archive>
+void CSGObject::cereal_save(Archive& ar) const
```
karlnapf: I guess eventually, we just want to call this "save" and "load"

src/shogun/base/SGObject.h
```diff
@@ -124,6 +124,12 @@ class CSGObject
 	/** destructor */
 	virtual ~CSGObject();
 
+	template<class Archive>
```
karlnapf: just wondering now: are these called by users? If so, who creates the Archive?

src/shogun/lib/SGVector.h
```diff
@@ -474,6 +477,25 @@ template<class T> class SGVector : public SGReferencedData
...
+		{
+			ar(vlen);
+			for (index_t i = 0; i < vlen; ++i)
+				ar(vector[i]);
+		}
+
+		template<class Archive>
+		void cereal_load(Archive & ar)
+		{
+			unref();
```
karlnapf: why unref?

src/shogun/lib/any.h
```diff
@@ -111,23 +151,23 @@ namespace shogun
         {
             *(storage) = new T(*reinterpret_cast<T const*>(v));
         }
-        
+
```
karlnapf: I agree with this change, but it should be not in this PR, as it makes it harder to review

src/shogun/lib/any.h
```diff
@@ -189,6 +231,7 @@ namespace shogun
             policy->clear(&storage);
             policy = other.policy;
             policy->set(&storage, other.storage);
+			enumtype = other.enumtype;
```
karlnapf: indentation

src/shogun/lib/any.h
```diff
@@ -212,6 +255,24 @@ namespace shogun
...
+			case TYPE_INT_32:
+				ar(*(reinterpret_cast<int32_t*>(storage))); break;
+			case TYPE_FLOAT_64:
+				ar(*(reinterpret_cast<float64_t*>(storage))); break;
+			case TYPE_SGVECTOR_INT_32:
+				ar(*(reinterpret_cast<SGVector<int32_t>*>(storage))); break;
+			case TYPE_SGVECTOR_FLOAT_64:
+				ar(*(reinterpret_cast<SGVector<float64_t>*>(storage))); break;
+			case TYPE_UNDEFINED:
+				SG_SWARNING("Type error: undefined data type\n");
```
karlnapf: Give more information in such error messages

src/shogun/lib/any.h
```diff
@@ -259,18 +321,19 @@ namespace shogun
 
         BaseAnyPolicy* policy;
         void* storage;
+		EnumType enumtype;
```
karlnapf: indentation mess :)

src/shogun/lib/any.h
```diff
@@ -259,18 +321,19 @@ namespace shogun
...
     {
         void* lhs_storage = lhs.storage;
-        void* rhs_storage = rhs.storage;        
+        void* rhs_storage = rhs.storage;
         return lhs.policy == rhs.policy and
             lhs.policy->equals(&lhs_storage, &rhs_storage);
     }
 
     inline bool operator!=(const Any& lhs, const Any& rhs)
-    {
```
karlnapf: you should start using a proper texteditor, one that doesnt override things in files

tests/unit/features/DenseFeatures_unittest.cc
```diff
@@ -1,152 +0,0 @@
-/*
```
karlnapf: why is this happening?

tests/unit/io/Cereal_unittest.cc
```diff
@@ -0,0 +1,77 @@
...
+
+using namespace shogun;
+
+TEST(Cereal, Json_SGVector_FLOAT64_save_load_pair)
+{
+	SGVector<float64_t> a(2), b;
+	a.set_const(1.14263158);
+
+	{
+		std::ofstream os("cereatltest_sgvector.cereal");
```
karlnapf: this is basic testing I guess....but the user API should look a  bit different, right?

karlnapf:
looking forward for the update here.
You can clean up and squash, document already

src/shogun/base/SGObject.cpp
OXPHOS: yep..it'll be easy to do a grep with `cereal_save`

tests/unit/io/Cereal_unittest.cc
OXPHOS: This is what the examples is like in cereal docs..like:
```
int main()
{
  std::ofstream os("out.cereal", std::ios::binary);
  cereal::BinaryOutputArchive archive( os );

  SomeData myData;
  archive( myData );

  return 0;
}
```

src/shogun/base/SGObject.h
vigsterkr: it's the user's choice... we can create helper functions if we want, say `saveJSON`, `saveXML` etc. but the idea here is that the user could define it's own Archive type as well.

karlnapf: Weeeeellll.
I mean so internally OK.
But for shogun users, can we please have functions that only take filenames and an enum that tells the file type? SWIG compatible etc.
I

tests/unit/io/Cereal_unittest.cc
karlnapf: We dont want that in Shogun's interface to the outside world (SWIG)
We want the save method to take a filename string instead (and an enum for the type for serialization the user wants)

OXPHOS:
@karlnapf @vigsterkr 
Now we have a working serialization model of SGObject(with Primitive types and SGVectors), which can save and load SGObject parameters from and into the basetag/any parameter list. The serialization was tested in Cereal_unittest.cc

Problems:
- `SGVector` serialization is working but I need to fix the ref count. I think I have to serialize `m_refcount` (`RefCount::rc`) as well to set the counting right.
- lambday and I didn't find a good way to decrease the case listing in `Any::cereal_save`. So I just explicitly listed them all. Actually, using just one Enum can make it shorter than what I current have, but maybe less clear.
- To make the archive file more readable, I was trying to use `cereal::make_nvp` (name-value pair) to annotate each saved parameter. However, I got error when I did `#include <cereal/cereal.hpp>`: http://pastebin.com/CVJyKzfQ
- **Important: ** Can I make a pull to fix the whitespaces and indentations in these files? The indentions in `any.h` doesn't match what Shogun has. I have set the indentation size as Shogun's in vim so it's kinda hard to match `any.h`

src/shogun/base/SGObject.cpp
```diff
@@ -471,6 +473,22 @@ void CSGObject::save_serializable_post() throw (ShogunException)
...
+{
+	for (auto it = self->map.begin(); it != self->map.end(); ++it)
+		ar(it->second);
+}
+template void CSGObject::cereal_save<cereal::JSONOutputArchive>(cereal::JSONOutputArchive& ar) const;
```
vigsterkr: i wouldn't do it like this, as said i think in a chat or somewhere; i would still expose the `cereal_save` with template but as well give implementations like this:
`void saveJSON(char* filename) const;`, `void saveXML(char* filename) const;` etc. and have the corresponding pairs for loading. and you yourself create there `cereal::JSONOutputArchive` etc. so this way the simple user can just choose from one of the `save*` functions, whereas the more advanced user would still have the option to use the template...

OXPHOS:
@vigsterkr  updated

src/shogun/base/SGObject.h
```diff
@@ -124,6 +124,16 @@ class CSGObject
 	/** destructor */
 	virtual ~CSGObject();
 
+	void saveJSON(char* filename) const;
```
vigsterkr: great! so in the same manner, lets add support for all the default cereal output formats: xml, binary.

tests/unit/io/Cereal_unittest.cc
```diff
@@ -0,0 +1,71 @@
...
+	CCerealObject obj_save, obj_load;
+	SGVector<float64_t> A(5), B;
+	A.range_fill(0);
+
+	obj_save.set<SGVector<float64_t>>("test_vector", A);
+	obj_save.saveJSON("cereatltest_save_cerealobject.cereal");
+
+	obj_load.loadJSON("cereatltest_save_cerealobject.cereal");
+	B = obj_load.get<SGVector<float64_t>>("test_vector");
+
```
OXPHOS: I didn't find any available method in SGObject that can load parameters back to variables.

src/shogun/base/SGObject.cpp
```diff
@@ -34,6 +39,8 @@
 #include <map>
 #endif
 
+#include <cereal/archives/json.hpp>
```
vigsterkr: double include.... guessing by mistake

karlnapf:
As to your questions:
 * ref counts do not really need to be serialized -- if you de-serialize an object, its refcounter starts from scratch. The only difference is if a class contains two references to the same memory, which of course needs to be reflected in the counter. I am not sure Shogun currently does any of that.
 * The enum thing seems to be the only way out the type madness, so OK
 * please send a whitespace PR, of course
 * not sure about the readable archive error, maybe lets discuss in IRC

tests/unit/io/Cereal_unittest.cc
karlnapf: Can you explain what you mean by this?

 * de-serializatzion directly allocates member variable memory and loads the data there
 * setters (if exist) might allow to change members
 * otherwise no access

src/shogun/base/SGObject.cpp
```diff
@@ -471,6 +478,68 @@ void CSGObject::save_serializable_post() throw (ShogunException)
 	m_save_post_called = true;
 }
 
+void CSGObject::saveBinary(char* filename) const
```
karlnapf: can we please stay with Shogun's method naming convections? No camelCase

src/shogun/base/SGObject.cpp
```diff
@@ -471,6 +478,68 @@ void CSGObject::save_serializable_post() throw (ShogunException)
...
+{
+	std::ofstream os(filename);
+	cereal::BinaryOutputArchive archive(os);
+	archive(*this);
+}
+
+void CSGObject::saveJSON(char* filename) const
```
karlnapf: Not sure, I would find a single method that takes an enum cleaner here. But that is my taste
@vigsterkr any comments?

src/shogun/base/SGObject.cpp
```diff
@@ -471,6 +478,68 @@ void CSGObject::save_serializable_post() throw (ShogunException)
...
+		ar(it->second);
+}
+template void CSGObject::cereal_save<cereal::BinaryOutputArchive>(cereal::BinaryOutputArchive& ar) const;
+template void CSGObject::cereal_save<cereal::JSONOutputArchive>(cereal::JSONOutputArchive& ar) const;
+template void CSGObject::cereal_save<cereal::XMLOutputArchive>(cereal::XMLOutputArchive& ar) const;
+
+template<class Archive>
+void CSGObject::cereal_load(Archive& ar)
+{
+	for (auto it = self->map.begin(); it != self->map.end(); ++it)
```
karlnapf: nice and clean :)

src/shogun/base/SGObject.h
```diff
@@ -123,6 +123,60 @@ class CSGObject
 
 	/** destructor */
 	virtual ~CSGObject();
+        
+        /** cereal serialization saver wrapper.
```
karlnapf: This is *user documentation*
Please tune it more. Don't talk about wrappers or anything, just tell the user that this serializes the SGObject to a binary file. You can reference to other methods in doxygen for details

karlnapf: same for all other methods

src/shogun/base/SGObject.h
```diff
@@ -123,6 +123,60 @@ class CSGObject
...
+	void loadJSON(char* filename);
+        
+        /** cereal serialization loader wrapper.
+         * load SGObject registered parameters to XML archives
+         *
+         * @param filename XML archive filename
+         */
+	void loadXML(char* filename);
+    
+        /** cereal serailization save method
```
karlnapf: Needs to be hidden from SWIG.
Also please put more information for c++ users, and also for Shogun hackers in here

src/shogun/lib/any.h
```diff
@@ -35,13 +35,81 @@
...
 #include <cxxabi.h>
 
 namespace shogun
 {
+
+	namespace serial
+	{
+		enum EnumContainerType
+		{
+			CT_UNDEFINED,
```
karlnapf: can we have to patch to any separately, just as the one for SGVectpor

src/shogun/lib/any.h
```diff
@@ -212,6 +283,105 @@ namespace shogun
...
+					SG_SERROR("Type error: undefined data type cannot be serialized.\n");
+					break;
+				}
+				break;
+			case serial::EnumContainerType::CT_SGVECTOR:
+				switch (m_datatype.e_primitivetype) {
+				case serial::EnumPrimitiveType::PT_INT_32:
+					cereal_save_helper<Archive, SGVector<int32_t> >(ar);
+					break;
+				case serial::EnumPrimitiveType::PT_FLOAT_64:
```
karlnapf: its long to do it this way, but I don't really see another way, and it at least is very clear to read (in contrast to before where we have the same type of case distinction, but hidden deeply inside the spagetthi code.
So plus one!

karlnapf: BTW it would be cool if the whole serailization framework would have debug message that are very clear to read. They should include the parameter name, the type, etc

src/shogun/lib/any.h
```diff
@@ -212,6 +283,105 @@ namespace shogun
...
+					break;
+				}
+				break;
+
+			case serial::EnumContainerType::CT_SGVECTOR:
+				switch (m_datatype.e_primitivetype) {
+				case serial::EnumPrimitiveType::PT_INT_32:
+					cereal_load_helper<Archive, SGVector<int32_t>>(ar);
+					break;
+				case serial::EnumPrimitiveType::PT_FLOAT_64:
```
karlnapf: same here, please put debug messages in

karlnapf: Can we also at some point have a benchmark of how fast it is to store classes with the new vs the old framework?

tests/unit/features/DenseFeatures_unittest.cc
karlnapf: ping!

tests/unit/io/CerealObject.h
```diff
@@ -0,0 +1,33 @@
...
+		return m_vector;
+	}
+
+    const char* get_name() const { return "CerealObject"; }
+
+protected:
+    void init_params()
+    {
+		m_vector = SGVector<float64_t>(5);
+		m_vector.set_const(0);
```
karlnapf: no const in tests!!!

tests/unit/io/Cereal_unittest.cc
```diff
@@ -0,0 +1,71 @@
...
+using namespace shogun;
+
+TEST(Cereal, Json_SGVector_FLOAT64_save_load_pair)
+{
+	const index_t size = 5;
+	SGVector<float64_t> a(size), b;
+	a.range_fill(1.0);
+
+	{
```
karlnapf: why the {} ?

tests/unit/io/Cereal_unittest.cc
```diff
@@ -0,0 +1,71 @@
...
+#include <../tests/unit/io/CerealObject.h>
+
+#include <shogun/lib/SGVector.h>
+#include <cereal/archives/json.hpp>
+#include <fstream>
+
+
+using namespace shogun;
+
+TEST(Cereal, Json_SGVector_FLOAT64_save_load_pair)
```
karlnapf: remove "_pair"

src/shogun/base/SGObject.cpp
vigsterkr: @karlnapf that was my first idea as well, but then think about all those interpreted languages, like python :)
```from modshogun import Class, EnumForSerialization
a = Class()
a.save("to_file.json", EnumForSerialization.JSON)
```
seems a bit more shitty than
```from modshogun import Class
a = Class()
a.save_json("to_file.json")
```
or?

tests/unit/io/Cereal_unittest.cc
karlnapf: call it load_equals_saved

tests/unit/io/Cereal_unittest.cc
```diff
@@ -0,0 +1,71 @@
...
+		std::ifstream is("cereatltest_sgvector.cereal");
+		cereal::JSONInputArchive archive(is);
+		archive(b);
+	}
+
+	for (index_t i = 0; i < size; i++)
+		EXPECT_NEAR(a[i], b[i], 1E-15);
+}
+
+TEST(Cereal, Json_AnyObject_save_and_load_methods)
```
karlnapf: better name for test please (see above)

karlnapf:
Nice!

Lets clean up, split the patch, and merge it!

src/shogun/base/SGObject.cpp
karlnapf: yeah, ok!

src/shogun/lib/any.h
OXPHOS: Can you give an example where they should be?
We can output the `BaseTag`(parameter names) in `SGObject.cpp` and output the types in each `case` clause.

src/shogun/base/SGObject.cpp
lisitsyn: Why not `const auto& it : self->map`?

src/shogun/lib/any.h
```diff
@@ -172,12 +240,14 @@ namespace shogun
         explicit Any(const T& v) : policy(select_policy<T>()), storage(nullptr)
         {
             policy->set(&storage, &v);
+			m_datatype.set<T>();
```
lisitsyn: Indentation goes crazy

src/shogun/base/SGObject.cpp
vigsterkr: hohohoooooo! a goooood one :P welcome back amigo :)

src/shogun/base/SGObject.cpp
```diff
@@ -471,6 +478,68 @@ void CSGObject::save_serializable_post() throw (ShogunException)
...
+}
+
+void CSGObject::saveJSON(char* filename) const
+{
+	std::ofstream os(filename);
+	cereal::JSONOutputArchive archive(os);
+	archive(*this);
+}
+
+void CSGObject::saveXML(char* filename) const
```
vigsterkr: btw dont forget to decamelize these ones :)))

tests/unit/io/Cereal_unittest.cc
```diff
@@ -0,0 +1,71 @@
...
+		archive(b);
+	}
+
+	EXPECT_TRUE(a==b);
+}
+
+TEST(Cereal, Json_CerealObject_save_and_load_methods)
+{
+	CCerealObject obj_save, obj_load;
+	SGVector<float64_t> A(5), B;
```
lisitsyn: Please use one variable per line as this is error prone

src/shogun/lib/any.h
karlnapf: It should print every parameter (name and type) it serializes/loads. For Vectors even the size etc

src/shogun/lib/any.h
karlnapf: still

karlnapf:
Any updates here?
also needs a rebase

src/shogun/lib/any.h
OXPHOS: Yeah I thought I could update this after the SGVector part is done because SGObejct requires the serialization of SGVector anyway. But that one is harder than I thought..

karlnapf:
Not sure whats next here, as said above, this needs be cleaned up and then can be merged. Currently, travis is broken

src/shogun/lib/SGReferencedData.h
```diff
@@ -10,12 +10,11 @@
 #define __SGREFERENCED_DATA_H__
 
 #include <shogun/lib/config.h>
-
 #include <shogun/lib/common.h>
-
+#include <shogun/lib/RefCount.h>
+#include <cereal/types/polymorphic.hpp>
```
vigsterkr: you would need to move this as well :)

src/shogun/lib/common.h
```diff
@@ -15,6 +15,22 @@
...
+#endif
+#define CEREAL_SAVE_FUNCTION_NAME cereal_save
+
+#ifdef CEREAL_LOAD_FUNCTION_NAME
+#undef CEREAL_LOAD_FUNCTION_NAME
+#endif
+#define CEREAL_LOAD_FUNCTION_NAME cereal_load
+#endif // CEREAL_MACROS_HPP_
+
+#include <cereal/cereal.hpp>
```
vigsterkr: noup this should not be here :S

tests/unit/io/Cereal_unittest.cc
```diff
@@ -0,0 +1,136 @@
+#include <shogun/lib/common.h>
+#include <gtest/gtest.h>
+#include <../tests/unit/io/CerealObject.h>
```
vigsterkr: if ../ then please use `"CerealObject.h"`

src/shogun/lib/SGReferencedData.cpp
```diff
@@ -112,4 +118,39 @@ int32_t SGReferencedData::unref()
 		return c;
 	}
 }
+
+#ifndef SWIG // SWIG should skip this part
```
vigsterkr: no need for this in .cpp... swig uses headers only

tests/unit/io/Cereal_unittest.cc
```diff
@@ -0,0 +1,136 @@
+#include <shogun/lib/common.h>
+#include <gtest/gtest.h>
+#include "../tests/unit/io/CerealObject.h"
```
vigsterkr: "" makes includes relative... sooooooo `#include "CerealObject.h"`

src/shogun/lib/SGReferencedData.h
```diff
@@ -10,12 +10,10 @@
 #define __SGREFERENCED_DATA_H__
 
 #include <shogun/lib/config.h>
-
 #include <shogun/lib/common.h>
-
+#include <shogun/lib/RefCount.h>
```
vigsterkr: why do you need this? why the forward declaration was not enough?

src/shogun/base/SGObject.h
```diff
@@ -124,6 +124,58 @@ class CSGObject
...
+	 * @param filename XML archive filename
+	 */
+	void load_xml(const char* filename);
+
+#ifndef SWIG // SWIG should skip this part
+	/** serializes SGObject parameters to Archive with Cereal
+	 *
+	 * @param ar Archive
+	 */
+	template<class Archive>
```
vigsterkr: this works atm just because you include cereal header in `shogun/lib/common.h` which should be removed.... so you need to do a forward declaration of `Archive` there instead of the include to make this work.

vigsterkr:
@OXPHOS you still missed one of my comments where the cereal header needs to be eliminated:
```
/opt/shogun/src/shogun/lib/common.h:32:29: fatal error: cereal/cereal.hpp: No such file or directory

 #include <cereal/cereal.hpp>```

note the rest of my comments that in order to be able to do that you'll have to do some forward declarations.

File: tests/unit/io/Cereal_unittest.cc [hunk 49]
```diff
@@ -0,0 +1,174 @@
+#include <shogun/lib/common.h>
+#include <gtest/gtest.h>
+#include "CerealObject.h"
+
+#include <shogun/lib/SGVector.h>
+#include <cereal/archives/json.hpp>
```
vigsterkr: why do you need this?
or rather, why dont we use the `save_json` and `load_json` methods?

OXPHOS: `save_json` only works for SGObject. Actually the serialization of sgvector and any are not exposed to users. I just did the internal tests of SGVector and Any objects with the naked(?) save/load methods.

vigsterkr: yep yep here me being stupid again! :)

vigsterkr:
@OXPHOS SOOOOOOOOOOOOOOOOOOOOOOOOOOOOoooooooooooooooOOO? squash & merge?

OXPHOS:
@vigsterkr Cooooooooooooooool !
So do you want me to split the PR to SGVector/Any/SGObejct? Or just use this one?

vigsterkr:
@OXPHOS nono it's fine! just squash it into one commit :)

OXPHOS:
@vigsterkr done!

vigsterkr:
@OXPHOS ok, great! let's wait for travis to be green again (just to be on the very safe side) and then i'll merge! \o/

vigsterkr:
@OXPHOS btw just as a side-note here, we'll have to do more serialization implementation, see SGMatrix etc. i.e. none SGObject base classes :)

PR #3376, opened 2016-07-18
Update KNN

Saurabh7:
WIP

### DIFF
--- a/src/shogun/multiclass/KNN.cpp
+++ b/src/shogun/multiclass/KNN.cpp
@@ -19,6 +19,7 @@
 #include <shogun/lib/JLCoverTree.h>
 #include <shogun/lib/Time.h>
 #include <shogun/base/Parameter.h>
+#include <shogun/multiclass/tree/KDTree.h>
 
 //#define BENCHMARK_KNN
 //#define DEBUG_KNN
@@ -31,7 +32,7 @@ CKNN::CKNN()
 	init();
 }
 
-CKNN::CKNN(int32_t k, CDistance* d, CLabels* trainlab)
+CKNN::CKNN(int32_t k, CDistance* d, CLabels* trainlab, KNN_SOLVER knn_solver)
 : CDistanceMachine()
 {
 	init();
@@ -44,6 +45,7 @@ CKNN::CKNN(int32_t k, CDistance* d, CLabels* trainlab)
 	set_distance(d);
 	set_labels(trainlab);
 	m_train_labels.vlen=trainlab->get_num_labels();
+	m_knn_solver=knn_solver;
 }
 
 void CKNN::init()
@@ -54,15 +56,17 @@ void CKNN::init()
 
 	m_k=3;
 	m_q=1.0;
-	m_use_covertree=false;
 	m_num_classes=0;
+	m_leaf_size=1;
+	m_knn_solver=KNN_BRUTE;
 
 	/* use the method classify_multiply_k to experiment with different values
 	 * of k */
 	SG_ADD(&m_k, "m_k", "Parameter k", MS_NOT_AVAILABLE);
 	SG_ADD(&m_q, "m_q", "Parameter q", MS_AVAILABLE);
-	SG_ADD(&m_use_covertree, "m_use_covertree", "Parameter use_covertree", MS_NOT_AVAILABLE);
 	SG_ADD(&m_num_classes, "m_num_classes", "Number of classes", MS_NOT_AVAILABLE);
+	SG_ADD(&m_leaf_size, "m_leaf_size", "Leaf size for KDTree", MS_NOT_AVAILABLE);
+	SG_ADD((machine_int_t*) &m_knn_solver, "m_knn_solver", "Algorithm to solve knn", MS_NOT_AVAILABLE);
 }
 
 CKNN::~CKNN()
@@ -117,6 +121,8 @@ SGMatrix<index_t> CKNN::nearest_neighbors()
 	//pre-allocation of the nearest neighbors
 	SGMatrix<index_t> NN(m_k, n);
 
+	distance->precompute_lhs();
+
 	//for each test example
 	for (int32_t i=0; i<n && (!CSignal::cancel_computations()); i++)
 	{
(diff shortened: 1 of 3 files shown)


Saurabh7:
a few results https://gist.github.com/Saurabh7/05369cb213290b82060859e7324fccac
Current verison of kdtree helps for lower dimensions as it should but a bit slower in comparison.

vigsterkr:
@Saurabh7 quite big diff on the same data set (smaller dim) in case of SK... how many time do you run the tests? just to see what's the (mean, std) of runtimes....

Saurabh7:
The difference is consistently in the same range when ran multiple times.
code : https://gist.github.com/Saurabh7/88d09aa530257a17a6f73b0b8c47a7b7

Saurabh7:
Actually @vigsterkr there was something wrong. The leaf sizes were different for kdtree.
corrected version: https://gist.github.com/Saurabh7/b50a3043a98a1fa9b870ed5797486520

karlnapf:
Can we also parallelise thisi eventually?

src/shogun/multiclass/KNN.cpp
```diff
@@ -100,6 +102,12 @@ bool CKNN::train_machine(CFeatures* data)
 	m_min_label=min_class;
 	m_num_classes=max_class-min_class+1;
 
+	if (m_knn_solver == KNN_KDTREE)
+	{
+		m_kd_tree = new CKDTree(m_leaf_size);
```
karlnapf: this is a well tested shogun class?
I remember either kdtree or cover tree was buggy
Also, are both implementation BSD compatible?

src/shogun/multiclass/KNN.cpp
```diff
@@ -203,9 +213,10 @@ CMulticlassLabels* CKNN::apply_multiclass(CFeatures* data)
 #ifdef BENCHMARK_KNN
```
karlnapf: can we pls remove these horrible things? :)

src/shogun/multiclass/KNN.cpp
```diff
@@ -277,6 +288,25 @@ CMulticlassLabels* CKNN::apply_multiclass(CFeatures* data)
...
 		SG_PRINT(">>>> JL applied in %9.4f\n", tstart.cur_time_diff(false))
 #endif
+		break;
+	}
+	case KNN_KDTREE:
+	{
+		m_kd_tree->query_knn(dynamic_cast<CDenseFeatures<float64_t>*>(data), m_k);
+		SGMatrix<index_t> NN = m_kd_tree->get_knn_indices();
+		for (int32_t i=0; i<num_lab && (!CSignal::cancel_computations()); i++)
```
karlnapf: This thing is actually quite good to have. We should put this into other algos.
Wanna send a snother patch for that?
I think it doesnt work for openmp though

src/shogun/multiclass/KNN.cpp
```diff
@@ -31,7 +31,7 @@ CKNN::CKNN()
 	init();
 }
 
-CKNN::CKNN(int32_t k, CDistance* d, CLabels* trainlab)
+CKNN::CKNN(int32_t k, CDistance* d, CLabels* trainlab, KNN_SOLVER_TYPE knn_solver)
```
karlnapf: We should aim for mimicing sklearns setup and default parameters, as I think it is really well tuned

karlnapf:
Why are we slower than sklearn? Any hunches? Looked at their code?

Saurabh7:
Brute force should be fixed with some precomputations I think.

src/shogun/multiclass/KNN.cpp
Saurabh7: There are tests, more might be needed I will check

src/shogun/machine/DistanceMachine.cpp
```diff
@@ -56,59 +56,10 @@ void CDistanceMachine::distances_lhs(float64_t* result,int32_t idx_a1,int32_t id
...
-    if (num_threads < 2)
-    {
-        D_THREAD_PARAM param;
-        param.d=distance;
-        param.r=result;
-        param.idx_r_start=idx_a1;
-        param.idx_start=idx_a1;
-        param.idx_stop=idx_a2+1;
-        param.idx_comp=idx_b;
+#pragma omp parallel for
```
karlnapf: this is quite a big change.
Rationale?

karlnapf:
Mmmh this is getting too big, we need to chop it into pieces. Can you send smaller PRs with some explanation what you are doing?

Saurabh7:
Yep, I have added falconn lib external code here too (https://github.com/FALCONN-LIB/FALCONN), for LSH based approximate KNN. I will do that in separate PR then. 
Need to fix some tests and find a good way to set their parameters.  To be able to control tradeoff between accuracy and speed. 
for example for a particular set of parameters (not the best i think) I get on this data

```
Dataset    Shape            SG time    SK time    ACC sg    ACC sk
---------  -------------  ---------  ---------  --------  --------
toy        [10000, 1000]    1.19713    3.50962     0.943     0.987

```

karlnapf:
Cool stuff. 
But the KD tree is also necessary, and a general clean up.
Please do the patches one after another as small as possible so that we can review easier

Saurabh7:
Yes, KD tree is already in.

Saurabh7:
added tests too, could be good to go ..

src/shogun/multiclass/KNN.cpp
```diff
@@ -100,6 +102,14 @@ bool CKNN::train_machine(CFeatures* data)
 	m_min_label=min_class;
 	m_num_classes=max_class-min_class+1;
 
+	if (m_knn_solver == KNN_KDTREE)
+	{
+		CFeatures* lhs = distance->get_lhs();
+		m_kd_tree = new CKDTree(m_leaf_size);
+		m_kd_tree->build_tree(dynamic_cast<CDenseFeatures<float64_t>*>(lhs));
```
vigsterkr: our current kd tree only works with `Densefeature<float64_t>`?

Saurabh7: Yup looks like it.

src/shogun/machine/DistanceMachine.cpp
Saurabh7: I am removing this too, not absolutely necessary, can discuss later.

vigsterkr:
@Saurabh7 great! can you please squash this so i can merge it?
and in the next PR send in the falconn based update?

Saurabh7:
Squased! waiting for travis ..

vigsterkr:
@Saurabh7 ok cool as soon as it's green we can merge.

PR #3380, opened 2016-07-18
refactor OPT 1

yorkerlin:


yorkerlin:
@karlnapf 
Every class in the OPT framework is a sub-class of `SGObject`.

yorkerlin:
@karlnapf 
Is `travis` down?

karlnapf:
Nope Its just busy, I pushed some things today. Your patch is a minimal refactor right? Feel free to merge it once travisbis green

src/shogun/machine/gp/KLDualInferenceMethod.cpp
```diff
@@ -507,14 +534,20 @@ void CKLDualInferenceMethod::update_alpha()
...
 {
-	KLDualInferenceMethodMinimizer *minimizer=dynamic_cast<KLDualInferenceMethodMinimizer*>(m_minimizer);
-	REQUIRE(minimizer,"The minimizer must be an instance of KLDualInferenceMethodMinimizer\n"); 
+	CKLDualInferenceMethodMinimizer *minimizer=dynamic_cast<CKLDualInferenceMethodMinimizer*>(m_minimizer);
+	REQUIRE(minimizer,"The minimizer must be an instance of KLDualInferenceMethoMinimizer\n"); 
         KLDualInferenceMethodCostFunction* cost_fun=new  KLDualInferenceMethodCostFunction();
 	cost_fun->set_target(this);
+	bool cleanup=false;
+
+	if(this->ref_count()>1)
```
karlnapf: this can go I guess?

src/shogun/machine/gp/KLInference.cpp
```diff
@@ -324,6 +343,9 @@ float64_t CKLInference::optimization()
 {
         KLInferenceCostFunction *cost_fun=new KLInferenceCostFunction();
         cost_fun->set_target(this);
+	bool cleanup=false;
+	if(this->ref_count()>1)
```
karlnapf: all of them can go I guess?

src/shogun/optimization/NLOPTMinimizer.h
```diff
@@ -61,7 +61,7 @@ class CNLOPTMinimizer: public FirstOrderMinimizer
 	 *
 	 * @return name CNLOPTMinimizer
 	 */
-	virtual const char* get_name() const { return "CNLOPTMinimizer"; }
+	virtual const char* get_name() const { return "NLOPTMinimizer"; }
```
yorkerlin: @karlnapf 
The unit test of SGObject is based on get_name.

yorkerlin:
@karlnapf 
travis will fail when it comes to `SerializationJSON`.

yorkerlin:
@karlnapf 
Do we plan to drop JSON?  JSON cannot support `Double-precision floating-point` :(

src/shogun/machine/gp/KLDualInferenceMethod.cpp
yorkerlin: NO. let me give you an example

yorkerlin: I use this trick to deal with `circular reference`.

src/shogun/optimization/NLOPTMinimizer.h
karlnapf: yep!

karlnapf:
We dont plan to drop it, but to have a whole new serialization concept. THis might help at the point it is done.
I dont understand why only some tests fail. Any hacks to fix it temporarily?

src/shogun/machine/gp/KLDualInferenceMethod.cpp
karlnapf: Mmmh I see,
Not sure we would do that in  here. Where is the circular thing biting you?

yorkerlin:
The unit test will fail if some special double value (eg, INIFITY)  is used.

karlnapf:
unfortunately, http://stackoverflow.com/questions/13581843/php-how-to-encode-infinity-or-nan-numbers-to-json

karlnapf:
@OXPHOS @vigsterkr I guess we can change the en/decoder to use a string or so for these values? Should not be hard with cereal?

src/shogun/machine/gp/KLDualInferenceMethod.cpp
yorkerlin: Let me give you an example.
Case A: in KLDualInferenceMethod::method():
at time 1:  KLDualInferenceMethod: ref_count =0
at time 2:  cost_fun->set_target(this);   (KLDualInferenceMethod: ref_count =1)
at time 3:  SG_UNREF(cost_fun);  ((KLDualInferenceMethod: ref_count =0 and will delete KLDualInferenceMethod in a method in KLDualInferenceMethod  )



Case B: in KLDualInferenceMethod::method():
at time 1:  KLDualInferenceMethod: ref_count =1
at time 2:  cost_fun->set_target(this);   (KLDualInferenceMethod: ref_count =2)
at time 3:  SG_UNREF(cost_fun);  ((KLDualInferenceMethod: ref_count =1 and will NOT delete KLDualInferenceMethod in  )

yorkerlin: I have to deal with these two cases.

yorkerlin: The following code will produce these two cases.
Case A
```
int main()
{
KLDualInferenceMethod* obj=new KLDualInferenceMethod()
obj->method();
return 0;
}
```

Case B
```
void GPC::set_obj(KLDualInferenceMethod* obj)
{
SG_REF(obj)
m_obj=obj;
}
void GPC::doSomething()
{
m_obj->method();
}
```

yorkerlin:
@karlnapf 
what is more, JSON does not support small values. (eg, 1e-20)

src/shogun/optimization/lbfgs/LBFGSMinimizer.cpp
```diff
@@ -57,12 +56,48 @@ LBFGSMinimizer::LBFGSMinimizer(FirstOrderCostFunction *fun)
 	init();
 }
 
-void LBFGSMinimizer::init()
+void CLBFGSMinimizer::init()
 {
 	set_lbfgs_parameters();
+	m_min_step=1e-6;
```
yorkerlin: a hack to deal with issue related to JSON

yorkerlin:
travis is green now. I merge this.

PR #3387, opened 2016-07-21
LinalgRefactor - SGMatrix - add

OXPHOS:


src/shogun/mathematics/linalg/LinalgBackendEigen.h
```diff
@@ -127,6 +128,18 @@ class LinalgBackendEigen : public LinalgBackendBase
 		return c;
 	}
 
+	/** Eigen3 matrix C = alpha*A + beta*B method */
+	template <typename T>
+	SGMatrix<T> add_impl(const SGMatrix<T>& a, const SGMatrix<T>& b, T alpha, T beta) const
```
karlnapf: Again, I think all add and scale methods should be inplace. If a user wants a new matrix where things are added, he can allocate it himeself

src/shogun/mathematics/linalg/LinalgNamespace.h
```diff
@@ -115,6 +115,24 @@ Container<T> add(const Container<T>& a, const Container<T>& b, T alpha=1, T beta
...
+ * @param A first matrix
+ * @param B second matrix
+ * @param alpha constant to be multiplied by the first matrix
+ * @param beta constant to be multiplied by the second matrix
+ * @return The result matrix
+ */
+template <typename T>
+SGMatrix<T> add(const SGMatrix<T>& a, const SGMatrix<T>& b, T alpha=1, T beta=1)
+{
+	REQUIRE((a.num_rows == b.num_rows), "Matrices should have the same row number! "
```
karlnapf: Can we do this less verbosy?
"Number of rows of matrix a (%d) must match matrix b (%d)"

src/shogun/mathematics/linalg/LinalgBackendEigen.h
vigsterkr: true but think about performance as well... what about having the option for both?

karlnapf: I think that makes it a bit messy interface wise. Why performance?

vigsterkr: so think about
```
float64_t x[2048];
for (size_t i = 0; i < 2048; ++i) {
   x[i] = function(x[i]);
}
```
this is kind of the logic of inplace right? when you do this you do a `read x[i]`, `write x[i]` and then `read [i+1]` so forth and so on... since you do a `write x[i]`, it'll have to go into the RAM that will take:
 a) time
 b) will invalidate the L1 cache
so yeah it's gonna take much more time than:
```
float64_t in[2048];
float64_t out[2048];
for (size_t i = 0; i < 2048; ++i) {
   out[i] = function(in[i]);
}
```

karlnapf: @OXPHOS I discussed with @vigsterkr 
For all operations, that we might want to do inplace, we want something like [this](https://gist.github.com/karlnapf/28e1ea447d3d625f2205d303599fa4ff):

So that we can do the three cases described in the gist.
What would be even better was if we would not manually write the wrapper method for automatic allocation, but have an automatic template/macro way to generate it. That should be possible with c++11 templates, or even with the macro you have.

If we can do a proof of concept in this PR, then we can update the remaining operations. Ping us if you have questions

karlnapf: @OXPHOS Please add to your TODO list (in addition to deplying these updates) that we want to document things also in a readme, in particular the inplace example by @vigsterkr above should be in there to make the point that we can do inplace, but it will be slower, so we only wanna do it if memory blows up otherwise (e.g. element wise operations on gigantic matrices or so)....

vigsterkr:
@OXPHOS rebase? what's the status here?

OXPHOS:
@vigsterkr updated and rebased. ready to merge.

OXPHOS:
@vigsterkr wanna merge these?

PR #3390, opened 2016-07-25
add meta-example api tests for literals and enums

sanuj:
@karlnapf Not sure how to structure the `enum` api tests. Need feedback.
Shall i send a PR for literals.dat to shogun-data?


examples/meta/src/meta_api/enums.sg
```diff
@@ -0,0 +1 @@
+enum LIBLINEAR_SOLVER_TYPE.L2R_L2LOSS_SVC
```
karlnapf: structure is good!

karlnapf: @sorig can you guide this a bit while I am away?

sorig: Java and C# don't allow statements that have no side effects like this one. 

One solution could be to try putting it in a print statement `print enum LIBLINEAR_SOLVER_TYPE.L2R_L2LOSS_SVC`. The print statement doesn't work for arbitrary expressions because of C++, but I think it should work for something as simple as this. You may have to modify some of the `example/meta/generator/targets/*.json` files to get the this working (I think C++ needs an extra import statement for basic printing to work).

The other option is to find the simplest/most stable class in Shogun that has a method that accepts enums. The downside to this is that it couples the test with whatever class you choose which is not super clean.

sorig:
@sanuj This looks good! Just need to make the enum test work in Java and C#

examples/meta/src/meta_api/enums.sg
sorig: A third option and (probably the cleanest) is to try to get enum typed variables working. I.e. get the meta-language statement `LIBLINEAR_SOLVER_TYPE myVar = enum LIBLINEAR_SOLVER_TYPE.L2R_L2LOSS_SVC` to work. I actually don't think this would be very difficult - these types of statements already parse and translate. You just need to make sure that enum types like `LIBLINEAR_SOLVER_TYPE` are imported correctly in all the languages. That's just a matter of fixing the dependency translation rules in the target `.json` files. Try updating your example and see what travis says. Let me know if you get stuck.

sorig: Hey, I just had another look at this. I'm pretty sure the statement `LIBLINEAR_SOLVER_TYPE myVar = enum LIBLINEAR_SOLVER_TYPE.L2R_L2LOSS_SVC` is going to work in all languages without any modifications to the translation rules.

sanuj: @sorig I'm trying on my local now.

sorig:
Java error fixed in: #3408
C# error fixed in: #3410

sorig:
Note that you now need to specify the precision of floating point literals. E.g. `123.456` is a real (64 bit floating point) and `123.456f` is a float (32 bit). 

Wait for the two pull requests to be merged and then update your test to reflect this :)

sorig:
@sanuj my fixes are merged. This should pass if you update the example with the updated float literal.

sanuj:
@sorig i'm checking it on my local. Shall I put more enums in `enums.sg`? I can choose a few random enums existing in Shogun.

sorig:
Sure!
Could you also just add a few comments explaining the difference between the `real` and `float` literals and that example authors should use correctly typed literals? Would be great to have these meta-language tests also work as documentation for example authors

sorig:
Just realised that I accidentally undid the changes that fixed the Java error. See #3416

sorig:
C# should work if you update the example though

sanuj:
@sorig Will it be possible for you to come on the IRC. Wanted to discuss a few more things.

sanuj:
@sorig `C#` is passing but `Java` is still failing.

sorig:
@sanuj It seems like SWIG is doing something weird when exporting the enums. Java seems to treat the enum values as integers rather than the associated enum type. I don't currently have time to to investigate this further. Could you have a look?
There are several possible solutions:
* Figure out a way to make Java treat the enum values as type enum rather than int (ideal solution)
* If it turns out we HAVE to treat the enum values as `int` in Java then I see two possible solutions
     * Find a solution where we just cast the values to the appropriate type. E.g. translate `enum LIBLINEAR_SOLVER_TYPE.L2R_LR` to `(LIBLINEAR_SOLVER_TYPE)L2R_LR` or something like that (you will have to play around with the enum translation rule and the import statements in Java).
     * Create new syntax for enum variable instantiation. This is the last resort solution - would be quite ugly.

sorig:
@sanuj @vigsterkr See: http://www.swig.org/Doc1.3/Java.html#enumerations

Are we supporting JDK < 1.5? If not then we should switch the SWIG interface to use proper enums in Java.

vigsterkr:
forget jdk < 1.5... already 1.7 is EOL last april... @sorig can you do the change as part of this PR?

sorig:
@vigsterkr this is @sanuj's PR, but I have opened #3422

sorig:
@sanuj OK, #3422 seems to work. I tried to send the change as a PR to your fork, but your fork doesn't show up in the list for some reason.

Instead, can you cherry-pick [my commit](https://github.com/sorig/shogun/commit/06817c115d88c5bccce4ccd523151d7c918949e7) to this PR so these changes can be tested together?

I will close #3422 once you've done that.

PR #3391, opened 2016-07-25
LinalgRefactor -  SGVector - inplace add

OXPHOS:
Ref #3387

- I was trying to merge `void add` and `SGVector<T> add` into one method by doing ` template <typename T> auto add(SGVector<T>& a, SGVector<T>& b, T alpha=1, T beta=1, bool inplace=false) -> decltype((!inplace, SGVector<T>>))`, but I got error `expected ')' -> decltype((!inplace, SGVector<T>>))`. Not sure whether this is a solvable problem or whether there's other way to do it @vigsterkr 

- `TEST(LinalgBackendViennaCL, add_in_place)` is working well(`EXPECT_NEAR` are true), but I got error: `shogun-unit-test(9457,0x7fff7166a300) malloc: *** error for object 0x7f8a138c3590: pointer being freed was not allocated` at the end. I don't know which object was freed twice. 

- I think `add(A, B, alpha, beta, inplace=T)`  is more concise than `add(A, B, A, alpha, beta)` for in-place addition mentioned in [the gist](https://gist.github.com/karlnapf/28e1ea447d3d625f2205d303599fa4ff). If so, the manual allocating case would be kinda awkward. But my question is do we really need that? @karlnapf

### DIFF
--- a/src/shogun/mathematics/linalg/LinalgNamespace.h
+++ b/src/shogun/mathematics/linalg/LinalgNamespace.h
@@ -113,6 +113,29 @@ Container<T> add(const Container<T>& a, const Container<T>& b, T alpha=1, T beta
 }
 
 /**
+ * Performs the operation result = alpha*a + beta*b.
+ *
+ * @param a first vector
+ * @param b second vector
+ * @param result the vector that saves the result
+ * @param alpha constant to be multiplied by the first vector
+ * @param beta constant to be multiplied by the second vector
+ */
+template <typename T>
+void add(SGVector<T>& a, SGVector<T>& b, SGVector<T>& result, T alpha=1, T beta=1)
+{
+	REQUIRE(a.vlen == b.vlen, "Length of vector a (%d) doesn't match vector b (%d).\n", a.vlen, b.vlen);
+	REQUIRE(result.vlen == b.vlen, "Length of vector result (%d) doesn't match vector a (%d).\n", result.vlen, a.vlen);
+
+	REQUIRE(!(result.on_gpu()^a.on_gpu()),
+		"Cannot operate with vector result on_gpu (%d) and vector a on_gpu (%d).\n", result.on_gpu(), a.on_gpu());
+	REQUIRE(!(result.on_gpu()^b.on_gpu()),
+		"Cannot operate with vector result on_gpu (%d) and vector b on_gpu (%d).\n", result.on_gpu(), b.on_gpu());
+
+	infer_backend(a, b)->add(a, b, alpha, beta, result);
+}
+
+/**
  * Vector dot-product that works with generic vectors.
  *
  * @param a first vector
--- a/tests/unit/mathematics/linalg/operations/Viennacl_operations_unittest.cc
+++ b/tests/unit/mathematics/linalg/operations/Viennacl_operations_unittest.cc
@@ -34,6 +34,32 @@ TEST(LinalgBackendViennaCL, SGVector_add)
 		EXPECT_NEAR(alpha*A[i]+beta*B[i], result[i], 1e-15);
 }
 
+TEST(LinalgBackendViennaCL, add_in_place)
+{
+	sg_linalg->set_gpu_backend(new LinalgBackendViennaCL());
+
+	const float32_t alpha = 0.3;
+	const float32_t beta = -1.5;
+
+	SGVector<float32_t> A(9), B(9), C(9);
+	SGVector<float32_t> A_gpu, B_gpu;
+
+	for (index_t i = 0; i < 9; ++i)
+	{
+		A[i] = i;
+		B[i] = 0.5*i;
+		C[i] = i;
+	}
+	A_gpu = to_gpu(A);
+	B_gpu = to_gpu(B);
+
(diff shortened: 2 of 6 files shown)


tests/unit/mathematics/linalg/operations/Viennacl_operations_unittest.cc
```diff
@@ -28,12 +28,39 @@ TEST(LinalgBackendViennaCL, add)
...
+	for (index_t i = 0; i < 9; ++i)
+	{
+		A[i] = i;
+		B[i] = 0.5*i;
+		C[i] = i;
+	}
+	A_gpu = to_gpu(A);
+	B_gpu = to_gpu(B);
+
+	add(A_gpu, B_gpu, A_gpu, alpha, beta);
```
karlnapf: I agree this is an awkward interface, suggestions?

src/shogun/mathematics/linalg/LinalgNamespace.h
```diff
@@ -105,16 +105,36 @@ LinalgBackendBase* infer_backend(const Container<T>& a, const Container<T>& b)
...
 {
-	REQUIRE(a.vlen == b.vlen, "Vectors should have same length! "
-			"a(%d) vs b(%d)\n", a.vlen, b.vlen);
-
+	REQUIRE(a.vlen == b.vlen, "Length of vector a(%d) doesn't match vector b(%d).\n", a.vlen, b.vlen);
 	return infer_backend(a, b)->add(a, b, alpha, beta);
 }
 
 /**
+ * Performs the operation C = alpha*A + beta*B.
```
karlnapf: is that true?

src/shogun/mathematics/linalg/LinalgNamespace.h
```diff
@@ -105,16 +105,36 @@ LinalgBackendBase* infer_backend(const Container<T>& a, const Container<T>& b)
...
+ * @return The result vector
+ */
+template <typename T>
+void add(SGVector<T>& a, SGVector<T>& b, SGVector<T>& result, T alpha=1, T beta=1)
+{
+	REQUIRE(a.vlen == b.vlen, "Length of vector a(%d) doesn't match vector b(%d).\n", a.vlen, b.vlen);
+	REQUIRE(result.vlen == b.vlen, "Length of vector result(%d) doesn't match vector a(%d).\n", result.vlen, a.vlen);
+
+	REQUIRE(!(result.on_gpu()^a.on_gpu()),
+		"Cannot operate with vector result on_gpu()(%d) and vector a on_gpu()(%d).\n", result.on_gpu(), a.on_gpu());
```
karlnapf: remove the () after on_gpu maybe?

src/shogun/mathematics/linalg/LinalgNamespace.h
```diff
@@ -105,16 +105,36 @@ LinalgBackendBase* infer_backend(const Container<T>& a, const Container<T>& b)
...
+ * @param beta constant to be multiplied by the second vector
+ * @return The result vector
+ */
+template <typename T>
+void add(SGVector<T>& a, SGVector<T>& b, SGVector<T>& result, T alpha=1, T beta=1)
+{
+	REQUIRE(a.vlen == b.vlen, "Length of vector a(%d) doesn't match vector b(%d).\n", a.vlen, b.vlen);
+	REQUIRE(result.vlen == b.vlen, "Length of vector result(%d) doesn't match vector a(%d).\n", result.vlen, a.vlen);
+
+	REQUIRE(!(result.on_gpu()^a.on_gpu()),
```
karlnapf: XOR for victory! :)

tests/unit/mathematics/linalg/operations/Viennacl_operations_unittest.cc
karlnapf: I mean inplace should be avoided where possible (for speed reasons), and only applied in particular cases where the memory cannot be afforded. So in that sense I dont care if the interface is awkward. But still would be nicer to do this somehow nicer (i.e. with a flag or so)

OXPHOS: I would prefer do it like `scale(a, scale, inplace=T) `, which is real "inplace".
I don't see the necessity of allocating the memory in advance like:
```
Vector output(size);
scale(a, scale, output);
```
compared to 
`Vector output = scale(a, scale, output);`
@karlnapf

vigsterkr: @OXPHOS the 'necessity' of allocating the memory advance is because some people want to actually have full control of the mallocs happening within the code

karlnapf: yeah imagine you already have a big matrix where you want to put things for some reason (I sometimes do in my code)

vigsterkr:
@OXPHOS rebase plz

OXPHOS:
@vigsterkr ready

OXPHOS:
@vigsterkr ready for merge as well

PR #3393, opened 2016-07-25
update indentations and whitespace in SGObject and Any

OXPHOS:
No functional change.


OXPHOS:
I'm having multiple errors in meta examples??

PR #3395, opened 2016-07-26
add assertation in CHAID tree

OXPHOS:
The only situation I can think of that will result in `current=NULL` is one applies the machine to data without training.


karlnapf:
Thanks!

PR #3396, opened 2016-07-26
Hide GPLK or CPLEX parameters

vigsterkr:
this way one will not require to supply the compiler flags
for GLPK or CPLEX when using MKL.


vigsterkr:
@lisitsyn ping! any insights as i stole your idea? need this because those header includes of external libraries are giving the user a hard time, when trying to use MKL.

lisitsyn:
:+1:

PR #3399, opened 2016-07-28
LinalgRefactor - Add sum_symmetric and Rm old sum methods

OXPHOS:
There's overlapped SGVector sum method in class `SGVector`. Most classes are calculating vector sum by (for example): `float64_t n=SGVector<float64_t>::sum(m, m_len)/2.0;`

The build became extremely slow when I tried to replace `SGVector<T>.sum(vec)` method with `linalg::sum(SGVector)` method in `LinalgNamespace.` 

I think it's because I included `LinalgNamespace.h` in `Math.h`, and for every class with `Math.h` header, the `LinalgNamespace.h` (together with `LinalgBackendEigen3.h` etc.) are revisited. Though I don't understand why it is doing this.


OXPHOS:
Similiar to [this test](https://travis-ci.org/shogun-toolbox/shogun/jobs/147951294)
Every class that `#include <shogun/mathematics/linalg/linalg.h>` will give endless warning because of the absence of the return.
```
	#define BACKEND_GENERIC_SUM(Type, Container) \
	virtual Type sum(const Container<Type>& a, bool no_diag) const \
	{  \
		SG_SNOTIMPLEMENTED; \
	}
```

`return NULL` also raises warning: cast from null to type etc.
I think this has to be fixed to use the current linalg library

karlnapf:
It has to be fixed you are right. Any ideas?

OXPHOS:
The warnings can be fixed with #3404. But I guess they have to be moved to .cpp to speed up.

OXPHOS:
@karlnapf I don't think macros and .h are "compatible". The build is going to be super slow if I include current linalg library in `Math.h` as it has to build all macros for every single file

karlnapf:
Ok, so what is your suggestion?

OXPHOS:
@karlnapf Move them to `.cpp` - this is the only solution I can think of with macros.

karlnapf:
Ok go for it and let's get this merged!
Can you also pls rebase this against the latest develop?

OXPHOS:
Found something interesting - 
I was replacing old `linalg::sum_symmetric<linalg::Backend::EIGEN3>(Block<>)` with new `linalg::sum(Block<>)` and the changes failed the test `CustomKernelTest, sum_symmetric_block`.
It turned out that, with matrix dimension > (40-45, 40-45) (don't know the precise boundary), the new method:
```
	template <typename T>
	T sum_impl(const linalg::Block<SGMatrix<T>>& mat, bool no_diag=false) const
	{
		typename SGMatrix<T>::EigenMatrixXtMap m = mat.m_matrix;
		Eigen::Block<typename SGMatrix<T>::EigenMatrixXtMap> m_block = m.block(
			mat.m_row_begin, mat.m_col_begin, mat.m_row_size, mat.m_col_size);
		T sum = m_block.sum();
		if (no_diag)
			sum -= m_block.diagonal().sum();

		return sum;
	}
``` 
and the old method:
```
	static T compute(Block<SGMatrix<T> > b, bool no_diag)
	{
		Eigen::Map<MatrixXt> map = b.m_matrix;

		const MatrixXt& m=map.template block(b.m_row_begin, b.m_col_begin,
				b.m_row_size, b.m_col_size);

		// since the matrix is symmetric with main diagonal inside, we can save half
		// the computation with using only the upper triangular part.
		const MatrixXt& m_upper=m.template triangularView<Eigen::StrictlyUpper>();
		T sum=m_upper.sum();

		// the actual sum would be twice of what we computed
		sum*=2;

		// add the diagonal elements if required
		if (!no_diag)
			sum+=m.diagonal().sum();

		return sum;
	}
```
gave results with difference at about 1E-2 ~ 2E-4 ([see here](http://pastebin.com/7nuRYyFq)).

I checked with Python by:
```
In: a = [matrix values]
In: sum(a)
Out: 431.87252859999944
```
And the result is consistent with the **new** sum method (i.e. calculate the symmetric matrix as regular matrix)(Difference: 3.590716744383826e-06).
I am not completely sure which is the correct result - it seems to be some precision issue with data type? @karlnapf

OXPHOS:
Also, there is a `sum_symmetric` method in `kernel.cpp` (The ref used in the above test), and I think it should be removed and replaced with `sum` in linalg.

yorkerlin:
@OXPHOS 
Are you sure that the matrix is `symmetric`?

yorkerlin:
You should check whether the upper triangular part and the lower triangular part are same.

OXPHOS:
@yorkerlin The `sum_symmetric` method checked for symmetry before calculating and I checked the data in Python. It is symmetry.

karlnapf:
So, there is a bug in the old implementation then?

OXPHOS:
@karlnapf Unless there's also bug with Python's internal `sum` method..I would say yes

yorkerlin:
could you test the old function and your function on the following matrix ?
M = K + transpose(K)
where K is the kernel.

karlnapf:
Python bugs are generally possible but not likely on something like a sum method.
I mean do @yorkerlin suggestion sanity check, and if not, replace things with the bug free version :)

OXPHOS:
@yorkerlin @karlnapf I made a mistake - It was the old `sum_symmetric` method that gave the same result as in Python. 
Now here's the new problem:

I tested with `smatrix = kmatrix + kmatrix.tranpose()`. 
The [old regular sum method](https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/mathematics/linalg/internal/implementation/Sum.h#L261): 863.74530029296875 
The [new regular sum method](https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/mathematics/linalg/internal/implementation/Sum.h#L327): 863.74530029296875
The [old sum_symmetric method](https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/mathematics/linalg/internal/implementation/Sum.h#L327): 863.74505615234375
The difference is 2e-4 between [1st, 2nd] and [3rd] result.

And I can get the "correct" result with the new sum method by modifying the sum code from:
```
		T sum = m_block.sum();
		if (no_diag)
			sum -= m_block.diagonal().sum();
```
to:
```
		const Eigen::Matrix<T,Eigen::Dynamic,Eigen::Dynamic>& m_upper=m_block.template triangularView<Eigen::StrictlyUpper>();
		T sum=m_upper.sum();
		sum = sum + sum;

		if (!no_diag)
			sum += m_block.diagonal().sum();
```

OXPHOS:
(Not sure if this is relevant. I am kinda confused here)
The result of `sum_symmetric<Backend::EIGEN3>(Bloack<smatrix, block_begin = 0, block_size = m+n, ...>, no_diag false)`  
  =  2072.17041015625
  = the result of  `sum_symmetric<Backend::EIGEN3>(smatrix, no_diag false) `
  = the result of   `sum<Backend::EIGEN3>(smatrix, no_diag false) `
 != the result of  ` sum<Backend::EIGEN3>(Bloack<smatrix, block_begin = 0, block_size = m+n, ...>, no_diag false)`  = 2072.17333984375
Where `m+n` is the dimension of the matrix and the difference between results is 0.0029296875

karlnapf:
For your suggested change above, I guess these are rounding errors arising from the substraction.
So yes, do the triangular view thing. It will be slightly slower, but I think eigen3 will still vectorize things so all good.

OXPHOS:
@karlnapf I added `sum_symmetric()`(Though I don't like it - it's redundant) and everything is good to go now

PR #3404, opened 2016-07-31
LinalgRefactor - minor fix - warnings

OXPHOS:


src/shogun/mathematics/linalg/LinalgNamespace.h
```diff
@@ -258,7 +258,7 @@ SGVector<T> from_gpu(const SGVector<T>& vec)
 			SG_SERROR("Data memory on GPU but no GPU backend registered. \
 						This can happen if the GPU backend was de-activated \
 						after memory has been transferred to GPU.\n");
-			return NULL;
+			return false;
```
karlnapf: I dont understand why false

OXPHOS: It requires a boolean value to silence the warning..

vigsterkr: ?

vigsterkr: ok so i wonder how a `(SGVector<T>)false` cast is working out. but yeah we have a problem with the API itself here, because how can we return an `SGVector<T>` if the operation is not successful

vigsterkr: @OXPHOS how many other places we do a trick like this?

OXPHOS: I did this because I got warning when I used "return NULL":
```
/Users/zora/Github/shogun/src/shogun/mathematics/linalg/LinalgNamespace.h:338:11: warning: implicit conversion of NULL constant to 'bool' [-Wnull-conversion]
   return __null;
          ^~~~~~
          false
```
Though I do think NULL makes more sense..
Or we can leave the warning there @vigsterkr

OXPHOS: It's only in `to_gpu` and `from_gpu` - but they appear in every unittest file which calls the method

vigsterkr: i know why... the question is do we have APIs that do:
```
SGVector<T> some_api() {
   if (some problem occures, hence we cannot return any sane SGVector<T>)
}
```

OXPHOS: I can check. I guess it won't be easy to dig out...

OXPHOS: okay I didn't find exact same scenario in Shogun  (because it seems not achievable) but I can make some changes:
- `SGVector<T>* from_gpu() { return NULL; }`
or
-   `SGVector<T> from_gpu() { return SGVector<T>(); }`
or
-  Use `boost::optional`

vigsterkr: boost? we dont have boost :)
`std::optional` is just a c++17 feature, that doesn't mean we cannot have it, question what are the compilers atm that support optional?

vigsterkr: what do you mean not achievable? i mean that during your linalg refactor, have you had similar cases? could you maybe go through the changes pretty please? :)

OXPHOS: The similar cases are all in this PR. I did `return 0` for all `SG_SNOTIMPLEMENTED` macros

vigsterkr: @OXPHOS what does `SGVector<T>` null mean? :) but ok if en exception is thrown the return value can be garbage... it's not that nice but acceptable

PR #3406, opened 2016-08-01
Add non-owning Any policy

lisitsyn:
This would enable us to implement SG_ADD so that
parameters are used through map
and member variables at the same time.


vigsterkr:
yeah baby grooooovyyyyy!

PR #3409, opened 2016-08-05
replace setsX("name", obj_X) by sets("name", obj_X)

sanuj:
@vigsterkr Here we go. :beers:


lisitsyn:
I am undecided yet due to one reason. What happens if we do `obj.sets("kernel", GaussianKernel());`? Is it `CSGObject` or `CKernel`? Won't there be some ambiguity?

sanuj:
@lisitsyn This works as expected:
```python
import modshogun as sg
obj = sg.Obj()
gk = sg.GaussianKernel()
obj.sets("gaussian", gk)
obj.getsKernel("gaussian") == gk      # True
```

PR #3410, opened 2016-08-05
Split number literals into int, real (64bit float), and float literals

sorig:
We need to specify precision of floating point literals because C# is very strict about downcasting these.

Fixes problem in #3390

### DIFF
--- a/examples/meta/generator/parse.py
+++ b/examples/meta/generator/parse.py
@@ -43,7 +43,9 @@ class FastParser:
     # Lexer specification
     # ---------------------------------------
     tokens = (
-        "NUMERAL",
+        "INTLITERAL",
+        "REALLITERAL",
+        "FLOATLITERAL",
         "STRINGLITERAL",
         "BOOLLITERAL",
         "BASICTYPE",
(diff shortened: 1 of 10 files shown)


sorig:
With this change we encourage example authors to use correctly typed literals. That is, write 
`float C = 1.0f`
rather than
`float C = 1.0`.

We need to do this because C# doesn't allow implicit downcasting - the latter statement does not compile in C#.
Another reason is that C++ `auto` keyword infers the wrong type which might get us in trouble once we enable integration testing for the other languages (? depending on how we do this @karlnapf)

I have changed the current examples to use correctly typed literals. The integration tests fail because of this. Once shogun-toolbox/shogun-data#121 is merged, the tests should pass.

sorig:
Maybe we should just stop serialising and testing native type scalars? Don't we have to get rid of these anyway once we start integration testing the other languages?

sorig:
I have factored out the breaking changes (the ones that depend on shogun-toolbox/shogun-data#121) into #3413 to make this merge a bit easier.

sorig:
OK, this one can be merged now

PR #3415, opened 2016-08-09
LSH based KNN

Saurabh7:


vigsterkr:
yeeeeey! huge imports FTW.
@Saurabh7 do you have some performance #?

vigsterkr:
@Saurabh7 i've restarted travis, it fails because of in the `eigen_wrapper.h` the direct include of `Eigen/Dense`. plz patch the original code that it uses the shogun eigen wrapper and before you commit run the unit test for knn

Saurabh7:
@vigsterkr I did run the test, it was fine. Dunno what is causing this.

vigsterkr:
@Saurabh7 i've just explained you what is causing this in my previous comment. the root cause is that eigen wrapper there... and plz if you are not so sure use the docker image for testing (the same one as the travis: https://hub.docker.com/r/shogun/shogun-dev/)

Saurabh7:
@vigsterkr I replaced the includes with shogun ones. similar error : `/src/shogun/mathematics/eigen3.h:16:24: fatal error: Eigen/Eigen: No such file or directory
  #include <Eigen/Eigen>
                        ^
`
I will use the docker img thanks.

vigsterkr:
@Saurabh7 squash & go?

Saurabh7:
Squashed, fixed memleaks and added some docs.

Saurabh7:
some numbers with right parameters:

```
Dataset    Shape           SG time    SK time    ACC sg    ACC sk
---------  ------------  ---------  ---------  --------  --------
toy        [5000, 6000]    3.34479    5.05927     0.946     0.984
toy        [5000, 5000]    3.00674    4.17225      0.97     0.994
toy        [5000, 1000]   0.696132   0.896596     0.976     0.978

```

PR #3417, opened 2016-08-10
Fix heuristic for include path search. Warn users if multiple candida…

sorig:
…tes are found.

Fixes include path issue in #3415

### DIFF
--- a/examples/meta/generator/translate.py
+++ b/examples/meta/generator/translate.py
@@ -11,6 +11,7 @@ try:
     set
 except NameError:
     from sets import Set as set
+import warnings
 
 
 def find(key, dictionary):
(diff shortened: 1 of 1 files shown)


sorig:
This is only a minor change of the include path heuristic. It would be a lot cleaner to get ctags to only output paths for classes/enums/typedefs that are available from the modular interfaces. Does anyone have an idea how to make that happen?

sorig:
For each include path candidate we could check if the path exists in an `%include` directive in one of the `.i` files. Maybe that would work?
Can we get SWIG to output us a list of all includes, so we don't have to parse the `.i` files in a hacky way?

PR #3422, opened 2016-08-12
Use proper java enums in SWIG interface

sorig:
WIP

I'm having problems testing on local, so let me just see what travis says


sorig:
Closing as this is now part of #3390

PR #3427, opened 2016-08-15
get distance matrix for euclidean distance

Saurabh7:


Saurabh7:
Should be faster than computing pairwise distance. Came across when looking at some previous benchmarks.
Will use this in KMeans for optional precomputation of all distances.

src/shogun/distance/EuclideanDistance.cpp
```diff
@@ -177,3 +179,38 @@ float64_t CEuclideanDistance::distance_upper_bounded(int32_t idx_a, int32_t idx_
...
+{
+	REQUIRE(lhs->get_feature_class()==C_DENSE,
+		"Left hand side (was %s) has to be CDenseFeatures instance!\n", lhs->get_name());
+	REQUIRE(rhs->get_feature_class()==C_DENSE,
+		"Right hand side (was %s) has to be CDenseFeatures instance!\n", rhs->get_name());
+
+	REQUIRE(lhs->get_feature_type()==F_DREAL,
+		"Left hand side (was %s) has to be of double type!\n", lhs->get_name());
+	REQUIRE(rhs->get_feature_type()==F_DREAL,
+		"Right hand side (was %s) has to be double type!\n", rhs->get_name());
```
lambday: Sorry I am bit late to comment on this but is it a good idea to put a hard REQUIRE in this method, when in the `init(...)` we only need the features to have DOT property active? Maybe have a fail-safe fallback mechanism for other features and add this Eigen3 specialization for dense features (interface should be the same for both). Also, this method should work for 32-bit float features also.

Saurabh7: Yeah you are right, will fallback to the generic method for other feature types. Thanks!

karlnapf: Lets fix and merge :)

Saurabh7: Hmm, actually cannot support float32  as well with this design. As I cannot make the templated `get_distance_matrix` method from base class virtual.

lambday: Hi @Saurabh7. The method `get_distance_matrix` doesn't have to be a template. It should always return a `SGMatrix<float64_t>` (even for sparse features). But internally it should be able to handle both `CDenseFeatures<float64_t>` and `CDenseFeatures<float32_t>`, along with other feature types that supports `FP_DOT`. You can have a template private method for that which handles both the cases - you do the Eigen3 stuffs inside that. Fallback works with `dot` method in the features. Sounds reasonable?

Another thing I missed - the present implementation creates two new feature matrices every time you call this (Please check [`CDenseFeatures::get_feature_matrix()`](https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/features/DenseFeatures.cpp#L273)). So, I think we should have two of these methods

1. One that works without subsets (fail with `SG_ERROR` if subset is present, say that the user should use the other variant if they subsets). It should use [`CDenseFeatures::get_feature_matrix(int32_t&, int32_t&)`](https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/features/DenseFeatures.cpp#L306), then initialize the Eigen3 maps with that. No mem alloc/copy, no-ref-counting business, faster.
2. One that works with subsets. This may give a `SG_INFO` when subsets are **not** present - saying that the users are better off using the other variant. This requires mem alloc/copy, ref-counting for SGMatrix but respects subsets.

We can do this by either having the method as `virtual SGMatrix<float64_t> get_distance_matrix(bool subset_present=false)` or having two different interfaces `get_distance_matrix()` and `get_distance_matrix_with_subsets()`. Thoughts on this?

When we're done with the implementation, we should also have unit tests for all the feature types for which `EuclideanDistance` can be initialized. This way, if something doesn't work, we'll catch it.

Let me know what you think. @karlnapf may have some comments.

Saurabh7: @lambday  
`get_distance_matrix` is already templated in base `CDistance` class.

Saurabh7: Hey @lambday , Doesn't `get_feature_matrix` already do what we need i.e. copy only when subsets are present : https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/features/DenseFeatures.cpp#L268   ?

I have fixed other things.

src/shogun/distance/EuclideanDistance.cpp
```diff
@@ -177,3 +179,69 @@ float64_t CEuclideanDistance::distance_upper_bounded(int32_t idx_a, int32_t idx_
...
+{
+	if (lhs->get_feature_class()==C_DENSE && rhs->get_feature_class()==C_DENSE)
+	{
+		if (lhs->get_feature_type()==F_DREAL && rhs->get_feature_type()==F_DREAL)
+		{
+			CDenseFeatures<float64_t>* casted_lhs=static_cast<CDenseFeatures<float64_t>*>(lhs);
+			CDenseFeatures<float64_t>* casted_rhs=static_cast<CDenseFeatures<float64_t>*>(rhs);
+			return compute_distance_matrix(casted_lhs, casted_rhs);
+		}
+
```
lambday: extra newline

src/shogun/distance/EuclideanDistance.cpp
```diff
@@ -177,3 +179,69 @@ float64_t CEuclideanDistance::distance_upper_bounded(int32_t idx_a, int32_t idx_
...
+		else if (lhs->get_feature_type()==F_SHORTREAL && rhs->get_feature_type()==F_SHORTREAL)
+		{
+
+			CDenseFeatures<float32_t>* casted_lhs=static_cast<CDenseFeatures<float32_t>*>(lhs);
+			CDenseFeatures<float32_t>* casted_rhs=static_cast<CDenseFeatures<float32_t>*>(rhs);
+			return compute_distance_matrix(casted_lhs, casted_rhs);
+		}
+
```
lambday: extra newline

src/shogun/distance/EuclideanDistance.cpp
```diff
@@ -177,3 +179,69 @@ float64_t CEuclideanDistance::distance_upper_bounded(int32_t idx_a, int32_t idx_
...
+		else
+			return CDistance::get_distance_matrix();
+	}
+
```
lambday: extra newline

src/shogun/distance/EuclideanDistance.cpp
```diff
@@ -177,3 +179,69 @@ float64_t CEuclideanDistance::distance_upper_bounded(int32_t idx_a, int32_t idx_
...
+	typename SGMatrix<T>::EigenMatrixXt map_result;
+
+	SGMatrix<T> l_matrix = casted_lhs->get_feature_matrix();
+	SGMatrix<T> r_matrix = casted_rhs->get_feature_matrix();
+	typename SGMatrix<T>::EigenMatrixXtMap map_l(l_matrix.matrix, l_matrix.num_rows, l_matrix.num_cols);
+	typename SGMatrix<T>::EigenMatrixXtMap map_r(r_matrix.matrix, r_matrix.num_rows, r_matrix.num_cols);
+
+	Map<VectorXd> map_lhs_norms(m_lhs_squared_norms.vector, n);
+	typename SGVector<T>::EigenVectorXt casted_lhs_norms = map_lhs_norms.template cast<T>();
```
lambday: Was the casting necessary? Just wondering, does the `+=` operator (in the `map_result.col(row)wise() += casted_l(r)hs_norms` expression) take care of the case when the operands are of different type? If it can, then we don't need that. Can you please double check?

src/shogun/distance/EuclideanDistance.cpp
```diff
@@ -177,3 +179,69 @@ float64_t CEuclideanDistance::distance_upper_bounded(int32_t idx_a, int32_t idx_
...
+
+	Map<VectorXd> map_rhs_norms(m_rhs_squared_norms.vector, m);
+	typename SGVector<T>::EigenVectorXt casted_rhs_norms = map_rhs_norms.template cast<T>();
+
+	map_result = map_l.transpose()*map_r;
```
lambday: I think you can have the `map_result` as `double` valued matrix from the beginning (take the matrix that you'll return in the end in the beginning itself, and then use the map as lvalue). So, if the `lhs` and `rhs` vectors are `float32_t` valued, you'll have just one cast (in this above expression). You won't need to cast the norm vectors. 

In the present code, in addition to casting the map result, we're having to cast the vectors (== mem-alloc/copy for those entire vectors) as well. Also, in the end, we're having to copy all the values back.

What do you think?

src/shogun/distance/EuclideanDistance.h
```diff
@@ -174,12 +175,16 @@ class CEuclideanDistance: public CDistance
 	 */
 	virtual CFeatures* replace_lhs(CFeatures* lhs);
 
+	SGMatrix<float64_t> get_distance_matrix();
```
lambday: Shouldn't this be `virtual`?

tests/unit/distance/EuclideanDistance_unittest.cc
```diff
@@ -106,6 +106,54 @@ TEST(EuclideanDistance,get_distance_matrix)
...
+TEST(EuclideanDistance,get_distance_matrix_float)
+{
+	CDenseFeatures<float64_t>* features_lhs=create_lhs();
+	CDenseFeatures<float64_t>* features_rhs=create_rhs();
+
+	CDenseFeatures<float32_t>* casted_lhs = (CDenseFeatures<float32_t>*) (features_lhs);
+	CDenseFeatures<float32_t>* casted_rhs = (CDenseFeatures<float32_t>*) (features_rhs);
+
+	// put features into distance object to compute squared Euclidean distances
+	CEuclideanDistance* euclidean=new CEuclideanDistance(casted_lhs, casted_rhs);
```
lambday: Can you please use `Some<T>`? Won't need the `SG_UNREF` in the end then.

tests/unit/distance/EuclideanDistance_unittest.cc
```diff
@@ -106,6 +106,54 @@ TEST(EuclideanDistance,get_distance_matrix)
...
+TEST(EuclideanDistance,get_distance_matrix_int)
+{
+	CDenseFeatures<float64_t>* features_lhs=create_lhs();
+	CDenseFeatures<float64_t>* features_rhs=create_rhs();
+
+	CDenseFeatures<int32_t>* casted_lhs = (CDenseFeatures<int32_t>*) (features_lhs);
+	CDenseFeatures<int32_t>* casted_rhs = (CDenseFeatures<int32_t>*) (features_rhs);
+
+	// put features into distance object to compute squared Euclidean distances
+	CEuclideanDistance* euclidean=new CEuclideanDistance(casted_lhs, casted_rhs);
```
lambday: `Some<T>` ?

src/shogun/distance/EuclideanDistance.cpp
lambday: @Saurabh7 hey man! Yeah you're right. I missed that somehow.

tests/unit/distance/EuclideanDistance_unittest.cc
karlnapf: Dont do this yet. We will do this systematically soon

karlnapf:
@lambday can I leave discussing this one for you? And then one of us can merge once youre happy

src/shogun/distance/EuclideanDistance.h
Saurabh7: It is in the base class, but I guess i should add it here too :)

src/shogun/distance/EuclideanDistance.cpp
Saurabh7: Can't operate on different types with eigen, I tried.

src/shogun/distance/EuclideanDistance.cpp
Saurabh7: yes that will reduce one cast, will do that .

lambday: Sounds good :)

src/shogun/distance/EuclideanDistance.h
lambday: You're right.. this is not necessary, but helps with readability and similar practice is followed everywhere in Shogun, so... :)

lambday:
Thanks @Saurabh7 .. let me know when you update the PR.
@karlnapf yeah I'll merge it then :)

src/shogun/distance/EuclideanDistance.cpp
lambday: Aah.. anyway, let's minimize the number of casts needed in the worst case..

src/shogun/distance/EuclideanDistance.cpp
Saurabh7: @lambday  Actually that would mean I have to cast LHS and RHS feature matrix to double. I think just casting the norms is better (even if we have to cast the result matrix) , memory wise ?

src/shogun/distance/EuclideanDistance.cpp
```diff
@@ -177,3 +179,61 @@ float64_t CEuclideanDistance::distance_upper_bounded(int32_t idx_a, int32_t idx_
...
+		return CDistance::get_distance_matrix();
+}
+
+template <typename T>
+SGMatrix<float64_t> CEuclideanDistance::compute_distance_matrix(CDenseFeatures<T>* casted_lhs, CDenseFeatures<T>* casted_rhs)
+{
+	int32_t n = casted_lhs->get_num_vectors();
+	int32_t m = casted_rhs->get_num_vectors();
+
+	MatrixXd map_result;
```
lambday: Hey @Saurabh7 sorry for being annoying again but just a minor comment. Isn't it possible to use the `SGMatrix<float64_t> result` matrix from here itself and then use an Eigen map for it for operations that follow? If operator `=` does a deep copy for Eigen (required in the expression `map_result = (map_l.transpose()*map_r).template cast<double>();`, then we're good, right? Saves an extra alloc and copy. Can you please check once whether it's doable?

Saurabh7: @lambday yep Its possibe, I have updated, can u stop previous travis maybe?

src/shogun/distance/EuclideanDistance.cpp
```diff
@@ -177,3 +179,59 @@ float64_t CEuclideanDistance::distance_upper_bounded(int32_t idx_a, int32_t idx_
...
+	map_result *= -2;
+	map_result.colwise() += map_lhs_norms;
+	map_result.rowwise() += map_rhs_norms.transpose();
+
+	// Set distances with self to zero, to remove rounding errors.
+	if (lhs==rhs)
+		map_result.diagonal().setZero();
+
+	if (!disable_sqrt)
+		map_result = map_result.array().sqrt();
```
lambday: Hey @Saurabh7 ! Sorry again for being a PITA (I should have commented on this before) but doesn't `map_result = map_result.array().sqrt();` create a temporary matrix and then do a deep copy? In that case, would it be possible to have an in-place operation for element-wise sqrt for the `result` SGMatrix? Maybe use our `linalg` methods?

src/shogun/distance/EuclideanDistance.cpp
```diff
@@ -177,3 +179,59 @@ float64_t CEuclideanDistance::distance_upper_bounded(int32_t idx_a, int32_t idx_
...
+
+	SGMatrix<T> l_matrix = casted_lhs->get_feature_matrix();
+	SGMatrix<T> r_matrix = casted_rhs->get_feature_matrix();
+	typename SGMatrix<T>::EigenMatrixXtMap map_l(l_matrix.matrix, l_matrix.num_rows, l_matrix.num_cols);
+	typename SGMatrix<T>::EigenMatrixXtMap map_r(r_matrix.matrix, r_matrix.num_rows, r_matrix.num_cols);
+
+	Map<VectorXd> map_lhs_norms(m_lhs_squared_norms.vector, n);
+	Map<VectorXd> map_rhs_norms(m_rhs_squared_norms.vector, m);
+
+	map_result = (map_l.transpose()*map_r).template cast<double>();
```
lambday: Just curious, what does Eigen `cast<T>` method do when the types are the same?

src/shogun/distance/EuclideanDistance.cpp
Saurabh7: There shouldn't be any runtime implications (claimed in docs), not sure if a temporary object is used.

lambday: Hey @Saurabh7 sorry for delayed reply - was busy with something. Hacked a [quick experiment](https://gist.github.com/lambday/2c67201d6e75c2aa8e10e5eb344df2b4) to see for myself - turns out it does create memory even with `-O3`. We're better off using our own element-wise operator it seems.

karlnapf:
Cool so can we merge this thing?

karlnapf:
How much faster is it? :)

src/shogun/distance/EuclideanDistance.cpp
Saurabh7: @lambday What exactly happens in our linalg with eigen backend ? cant find sqrt .. can you point me?

lambday: @Saurabh7 should use a native implementation of element-wise operator. But I see that's not added in the new feature branch that @OXPHOS is working on, neither it is there in the develop branch but you can actually do it [there](https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/mathematics/linalg/internal/modules/ElementwiseOperations.h#L80).

For now, just use a 
```c++
std::for_each(matrix.data(), matrix.data()+matrix.size(), [](T& value) { value = static_cast<T>(CMath::sqrt(value)); });
```
or something similar, that complies with C++98

lambday: You can make the existing linalg work with your requirements, avoiding the error for `float32_t` by just passing a lambda to the existing element-wise operator
```c++
linalg::elementwise_compute_inplace(your_matrix, [](T value) { return static_cast<T>(CMath::sqrt(value)); });
```

lambday: Just make sure you guard this thing with C++11 macros for now **if** you use C++11 stuff or `linalg`.

lambday: Hey @Saurabh7! Did you happen to have a look at it? If not, I can merge this patch for now and we can change it later. Let me know.

Saurabh7: @lambday  ah sorry couldn't try this out yet, might do so on the weekend.

karlnapf:
@micmn this one could be interesting for you to pick up as well

PR #3433, opened 2016-08-22
Random rotation ensemble

Saurabh7:


src/shogun/multiclass/tree/RRCARTree.cpp
```diff
@@ -0,0 +1,135 @@
...
+SGMatrix<float64_t> CRRCARTree::generate_rotation_matrix(int32_t n)
+{
+	SGMatrix<float64_t> M(n, n);
+	Map<MatrixXd> map_M(M.matrix, n, n);
+	MatrixXd A(n, n);
+	const VectorXd ones (VectorXd::Ones(n));
+	
+	for (int32_t i=0; i<n; ++i)
+		for (int32_t j=0; j<n; ++j)
+			A(i, j) = CMath::random(0.0, 1.0);
```
vigsterkr: you can use `Random.fill_array` to fill a vector fast with random values (it's using SIMD instructions).

src/shogun/multiclass/tree/RRCARTree.cpp
```diff
@@ -0,0 +1,135 @@
...
+	new_mat.zero();
+	Map<MatrixXd> map_new_mat(new_mat.matrix, num_features, num_vectors);
+ 
+	MatrixXd temp = map_feats.transpose()*map_rotation_matrix;
+	map_new_mat = temp.transpose();
+	CFeatures* new_data = new CDenseFeatures<float64_t>(new_mat);
+	return CRandomCARTree::train_machine(new_data);
+}
+
+SGMatrix<float64_t> CRRCARTree::generate_rotation_matrix(int32_t n)
```
vigsterkr: the rotation matrix is independent of the tree being used, right?
in this case it'd be better to move the implementation into a higher class so that any rotation forest/tree implementation would be able to use it instead of re-implementing it.

src/shogun/multiclass/tree/RRCARTree.cpp
vigsterkr: btw: here according to the paper the random should be sampled from a standard normal distribution, i.e. mean = 0, stddev = 1. while CMath::random is just a sample from uniform random distribution

Saurabh7: I see, `CMath::randn_double` gives std normal can use that.  `fill_array` might not be suitable for std normal I guess.

src/shogun/multiclass/tree/RRCARTree.cpp
Saurabh7: Yes, its completely independent, will move it higher.

src/shogun/multiclass/tree/RRCARTree.cpp
vigsterkr: yeah i just realised later that it requires a std normal distr. random. Please use `sg_rand` or `CRandom` directly instead of `CMath` wrappers as they are gonna get deprecated sooner or later

vigsterkr:
lets close this and go with #3435

PR #3435, opened 2016-08-22
Random rotation ensemble 

Saurabh7:


src/shogun/multiclass/tree/RRCARTree.cpp
```diff
@@ -0,0 +1,143 @@
...
+	MatrixXd temp = map_feats.transpose()*map_rotation_matrix;
+	map_new_mat = temp.transpose();
+	CFeatures* new_data = new CDenseFeatures<float64_t>(new_mat);
+	bool result = CRandomCARTree::train_machine(new_data);
+
+	SG_UNREF(new_data);
+	return result;
+}
+
+SGMatrix<float64_t> CRRCARTree::generate_rotation_matrix(int32_t n)
```
vigsterkr: @OXPHOS this is a method that we would like move to linalg backend; would be great if you could join in the discussion where to put it exactly and how

src/shogun/multiclass/tree/RRCARTree.cpp
```diff
@@ -0,0 +1,143 @@
...
+			map_M(i, 0) = -map_M(i, 0);	
+
+	return M;
+}
+
+CMulticlassLabels* CRRCARTree::apply_multiclass(CFeatures* data)
+{
+	REQUIRE(data, "Data required for classification in apply_multiclass\n")
+
+	SGMatrix<float64_t> feats = (dynamic_cast<CDenseFeatures<float64_t>*>(data))->get_feature_matrix();
```
vigsterkr: @Saurabh7 you are writing the exact same lines of code when applying the rotational matrix to the original data matrix in case of `apply_regression` and `apply_multiclass`, right? in this case it would make sense to have a function called `apply_rotation` or something as its unnecessary code duplication.

the other thing is that it would be great to port this part of code (transpose, multiplication of matrix etc) to the linalg backend. i believe everything you need for matrix manipulation is there already. right @OXPHOS ?

OXPHOS:
@vigsterkr Yes I am planning to migrate all computation methods in `SGVector` and `SGMatrix` to `linalg`. I will also move all the linear algebra methods defined in machine learning classes to `linalg`.

src/shogun/multiclass/tree/RRCARTree.cpp
OXPHOS: I would suggest renaming the method to `rotate_matrix` and be placed between `range_fill` and `scale` methods.
If the method only supports CPU backend calculation, the method will need to be registered in `LinalgNameSpace.h`:
```
template <typename T>
SGMatrix<T> rotate_matrix(int32_t n);
{
	REQUIRE(n>0, '');
	return sg_linalg->get_cpu_backend()->rotate_matrix(n);
}
```
and implemented in `LinalgBackendBase.h` and `LinalgBackendEigen.h`.
The Eigen3-`map` method should be called as `typename SGMatrix<T>::EigenMatrixXtMap a_eig = a;`
Other Eigen3 methods will stay the same as below.

And a unit-test in `tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc` is preferred.

To use the method, we can include `shogun/mathematics/linalg/LinalgNamespace.h` in this `.cpp`, and call `linalg::rotate_matrix(n);`

It's gonna be tricky to do GPU implementation as well, because there's no data structure passed to the method and the `linalg` cannot infer the backend to use. We can discuss if we'd like to have the GPU part. 

I can take a closer look once I submit my GSOC final report today : )

vigsterkr: @OXPHOS wouldn't a `rotational_matrix` be a better name? as that api doesn't actually rotate a matrix :) but gives back an operator that'll rotate another matrix. :) other than that it's a good plan! @Saurabh7 can you start working on this?

vigsterkr: yes we would like to have GPU part as well, it's not urgent though but yeah would be great to have a plan how to do it :)

vigsterkr: if we'd to go with a:
```
template <typename T>
void rotate_matrix(int32_t n, SGMatrix<T>& m);
{
    REQUIRE(n>0, '');
....
}
```
we could infer the backend or? :)

vigsterkr:
@Saurabh7 FYI i just rebased the original `feature/linalg_refactor` so you'll have to update your branch as well

src/shogun/multiclass/tree/RRCARTree.cpp
Saurabh7: Ah sry missed these comments, yeah we could change it to `rotate_matrix(int32_t n, SGMatrix<T>& m)`

OXPHOS: Oh I was wondering why there's no matrix ;D
But yeah it can infer the backend with `SGMatrix` in param list. So it would be just like other operations. 
I didn't see matrix rotation methods in `ViennaCL` with a quick search. So we have to start from scratch and maybe use kernels to do the rotation...?

vigsterkr: @oxphos just do the same as in case of eigen:
 1) create A square matrix (n*n)
 2) initialise with random values from a standard normal distrib
 3) QR decompose this matrix
 4) then let M be: 
```
M = Q * (qr.matrixQR().diagonal().array() < 0).select(-ones, ones).asDiagonal();
```
 5) depending on the determinant of the M flip the first column values.

@Saurabh7 btw couldn't we do the the last step (negating) as a result of a scaling (with -1) of the given column? hence this could use some vector multiplication backend, i.e. SIMD?

Saurabh7: yeah we should change it to colwise scaling. will do that.

src/shogun/machine/RRForest.cpp
```diff
@@ -0,0 +1,177 @@
...
+
+CRRForest::CRRForest(int32_t rand_numfeats, int32_t num_bags)
+: CBaggingMachine()
+{
+	init();
+	
+	set_num_bags(num_bags);
+
+	if (rand_numfeats>0)
+		dynamic_cast<CRRCARTree*>(m_machine)->set_feature_subset_size(rand_numfeats);
```
karlnapf: is this always safe?

src/shogun/machine/RRForest.cpp
```diff
@@ -0,0 +1,177 @@
...
+}
+
+CRRForest::CRRForest(CFeatures* features, CLabels* labels, int32_t num_bags, int32_t rand_numfeats)
```
karlnapf: minor: use `rand_num_feats`  -- style consistency

src/shogun/machine/RRForest.cpp
```diff
@@ -0,0 +1,177 @@
...
+	init();
+
+	SG_REF(features);
+	m_features=features;
+	set_labels(labels);
+
+	set_num_bags(num_bags);
+
+	if (rand_numfeats>0)
+		dynamic_cast<CRRCARTree*>(m_machine)->set_feature_subset_size(rand_numfeats);
```
karlnapf: same as above

src/shogun/machine/RRForest.cpp
```diff
@@ -0,0 +1,177 @@
...
+}
+
+CRRForest::CRRForest(CFeatures* features, CLabels* labels, SGVector<float64_t> weights, int32_t num_bags, int32_t rand_numfeats)
```
karlnapf: line wrap

karlnapf: rand_num_feats
Maybe a better name is appropriate as well: num_random_feats, num_rand_feats, or so

src/shogun/machine/RRForest.cpp
```diff
@@ -0,0 +1,177 @@
...
+		dynamic_cast<CRRCARTree*>(m_machine)->set_feature_subset_size(rand_numfeats);
+}
+
+CRRForest::~CRRForest()
+{
+}
+
+void CRRForest::set_machine(CMachine* machine)
+{
+	SG_ERROR("Machine is set as CRRCART and cannot be changed\n")
```
karlnapf: Avoid mentioning class names. Just "Machine cannot be changed.\n"

src/shogun/machine/RRForest.cpp
```diff
@@ -0,0 +1,177 @@
...
+}
+
+SGVector<float64_t> CRRForest::get_weights() const
+{
+	return m_weights;
+}
+
+EProblemType CRRForest::get_machine_problem_type() const
+{
+	REQUIRE(m_machine,"m_machine is NULL. It is expected to be RRCARTree\n")
```
karlnapf: This message will only confuse users, who dont know m_machine, NULL, or similar. Also dont use class names in error messages.
Also, REQUIRE in a getter doesnt make sense

src/shogun/machine/RRForest.cpp
```diff
@@ -0,0 +1,177 @@
...
+	return dynamic_cast<CRRCARTree*>(m_machine)->get_machine_problem_type();
+}
+
+void CRRForest::set_machine_problem_type(EProblemType mode)
+{
+	REQUIRE(m_machine,"m_machine is NULL. It is expected to be RRCARTree\n")
```
karlnapf: Same as above

src/shogun/machine/RRForest.cpp
```diff
@@ -0,0 +1,177 @@
...
+	dynamic_cast<CRRCARTree*>(m_machine)->set_machine_problem_type(mode);
+}
+
+void CRRForest::set_num_random_features(int32_t rand_featsize)
+{
+	REQUIRE(m_machine,"m_machine is NULL. It is expected to be RRCARTree\n")
```
karlnapf: Same as above

src/shogun/machine/RRForest.cpp
```diff
@@ -0,0 +1,177 @@
...
+	REQUIRE(rand_featsize>0,"feature subset size should be greater than 0\n")
```
karlnapf: Please be more careful with error messages.
`REQUIRE(rand_featsize>0, "Feature subset size (%d) should be greater than 0.\n", rand_featsize)`

src/shogun/machine/RRForest.cpp
```diff
@@ -0,0 +1,177 @@
...
+	return dynamic_cast<CRRCARTree*>(m_machine)->get_machine_problem_type();
+}
+
+void CRRForest::set_machine_problem_type(EProblemType mode)
+{
+	REQUIRE(m_machine,"m_machine is NULL. It is expected to be RRCARTree\n")
+	dynamic_cast<CRRCARTree*>(m_machine)->set_machine_problem_type(mode);
+}
+
+void CRRForest::set_num_random_features(int32_t rand_featsize)
```
karlnapf: can you use a better variable name?

src/shogun/machine/RRForest.cpp
```diff
@@ -0,0 +1,177 @@
...
+{
+	REQUIRE(m_machine,"m_machine is NULL. It is expected to be RRCARTree\n")
+	REQUIRE(rand_featsize>0,"feature subset size should be greater than 0\n")
+
+	dynamic_cast<CRRCARTree*>(m_machine)->set_feature_subset_size(rand_featsize);
+}
+
+int32_t CRRForest::get_num_random_features() const
+{
+	REQUIRE(m_machine,"m_machine is NULL. It is expected to be RRCARTree\n")
```
karlnapf: as above

src/shogun/machine/RRForest.cpp
```diff
@@ -0,0 +1,177 @@
...
+	return dynamic_cast<CRRCARTree*>(m_machine)->get_feature_subset_size();
+}
+
+void CRRForest::set_machine_parameters(CMachine* m, SGVector<index_t> idx)
+{
+	REQUIRE(m,"Machine supplied is NULL\n")
```
karlnapf: "No machine provided!\n"

src/shogun/machine/RRForest.cpp
```diff
@@ -0,0 +1,177 @@
...
+	REQUIRE(m_machine,"Reference Machine is NULL\n")
```
karlnapf: Reference machine? Again, please use good error messages that do not confuse users

src/shogun/machine/RRForest.cpp
```diff
@@ -0,0 +1,177 @@
...
+
+	CRRCARTree* tree=dynamic_cast<CRRCARTree*>(m);
```
karlnapf: cast safe?

karlnapf: At least you should put in a comment why

src/shogun/machine/RRForest.cpp
```diff
@@ -0,0 +1,177 @@
...
+
+	SGVector<float64_t> weights(idx.vlen);
+
+	if (m_weights.vlen==0)
+	{
```
karlnapf: no {} for single statements

src/shogun/machine/RRForest.cpp
```diff
@@ -0,0 +1,177 @@
...
+		weights.fill_vector(weights.vector,weights.vlen,1.0);
+	}
+	else
+	{
+		for (int32_t i=0;i<idx.vlen;i++)
```
karlnapf: you could use auto here, or even @lisitsyn 's range

src/shogun/machine/RRForest.h
```diff
@@ -0,0 +1,152 @@
...
+#ifndef _RRFOREST_H__
+#define _RRFOREST_H__
+
+#include <shogun/lib/config.h>
+#include <shogun/machine/BaggingMachine.h>
+
+namespace shogun
+{
+
+/** @brief Random Rotation - Random forest. Feature space is randomly rotated prior to training the base learners.
```
karlnapf: Can you put in two more sentences in here?
Also explain that this only works with certain type of base machines!
All the caveats that might produce error messages

src/shogun/machine/RRForest.h
```diff
@@ -0,0 +1,152 @@
...
+ *
+ *  See: Random Rotation Ensembles Journal of Machine Learning Research, Vol. 17, No. 4. (2016), pp. 1-26 by Rico Blaser, Piotr Fryzlewicz
+ 
+ */
+class CRRForest : public CBaggingMachine
+{
+public:
+	/** constructor */
```
karlnapf: Capital C

src/shogun/machine/RRForest.h
```diff
@@ -0,0 +1,152 @@
...
+	CRRForest();
+
+	/** constructor
```
karlnapf: Capital C, also below

src/shogun/machine/RRForest.h
```diff
@@ -0,0 +1,152 @@
...
+	 * @param weights weights of training feature vectors
+	 * @param num_bags number of trees in forest
+	 * @param num_rand_feats number of attributes chosen randomly during node split in candidate trees
+	 */
+	CRRForest(CFeatures* features, CLabels* labels, SGVector<float64_t> weights, int32_t num_bags=10, int32_t num_rand_feats=0);
+
+	/** destructor */
+	virtual ~CRRForest();
+
+	/** get name
```
karlnapf: remove "get name", just "@return"

src/shogun/machine/RRForest.h
```diff
@@ -0,0 +1,152 @@
...
+	 *
+	 * @return RRForest
+	 */
+	virtual const char* get_name() const { return "RRForest"; }
+
+	/** machine is set to modified CART(RandomCART) and cannot be changed
```
karlnapf: Capital M.
Also please be clearer in your language here

src/shogun/machine/RRForest.h
```diff
@@ -0,0 +1,152 @@
...
+	/** get number of random features to be chosen during node splits
+	 *
+	 * @return number of randomly chosen features during each node split
+	 */
+	int32_t get_num_random_features() const;
+
+protected:
+
+	virtual bool train_machine(CFeatures* data=NULL);
+	/** sets parameters of CARTree - sets machine labels and weights here
```
karlnapf: newline here

src/shogun/multiclass/tree/RRCARTree.cpp
```diff
@@ -0,0 +1,143 @@
...
+}
+
+CRRCARTree::~CRRCARTree()
+{
+}
+
+bool CRRCARTree::train_machine(CFeatures* data)
+{
+	REQUIRE(data,"Data required for training\n")
+	REQUIRE(data->get_feature_class()==C_DENSE,"Dense data required for training\n")
```
karlnapf: proper error message please!

src/shogun/multiclass/tree/RRCARTree.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+	int32_t num_features=(dynamic_cast<CDenseFeatures<float64_t>*>(data))->get_num_features();
```
karlnapf: please dont dynamic cast all the time. Store a typed pointer!
Also is the cast safe?

Saurabh7:
Thanks @karlnapf , I will push some changes in coming days.

vigsterkr:
@Saurabh7 i suppose this is down in the drain as hasn't been touched for 4 months now...

PR #3441, opened 2016-08-30
Fix ROC evaluation to properly access binary labels

lkuchenb:
ROC evaluation accesses uninitialized memory via CLabel::get_value() instead of the actual labels via CBinaryLabel::get_label().

I added the typecasts after checking the label type and replaced the function calls.

### DIFF
--- a/src/shogun/evaluation/ROCEvaluation.cpp
+++ b/src/shogun/evaluation/ROCEvaluation.cpp
@@ -19,15 +19,16 @@ CROCEvaluation::~CROCEvaluation()
 
 float64_t CROCEvaluation::evaluate(CLabels* predicted, CLabels* ground_truth)
 {
-	return evaluate_roc(predicted,ground_truth);
+	REQUIRE(predicted->get_label_type()==LT_BINARY, "ROC evalution requires binary labels.");
+	REQUIRE(ground_truth->get_label_type()==LT_BINARY, "ROC evalution requires binary labels.");
+
+	return evaluate_roc((CBinaryLabels*)predicted,(CBinaryLabels*)ground_truth);
 }
 
-float64_t CROCEvaluation::evaluate_roc(CLabels* predicted, CLabels* ground_truth)
+float64_t CROCEvaluation::evaluate_roc(CBinaryLabels* predicted, CBinaryLabels* ground_truth)
 {
 	ASSERT(predicted && ground_truth)
 	ASSERT(predicted->get_num_labels()==ground_truth->get_num_labels())
-	ASSERT(predicted->get_label_type()==LT_BINARY)
-	ASSERT(ground_truth->get_label_type()==LT_BINARY)
 	ground_truth->ensure_valid();
 
 	// assume threshold as negative infinity
@@ -76,7 +77,7 @@ float64_t CROCEvaluation::evaluate_roc(CLabels* predicted, CLabels* ground_truth
 	// get total numbers of positive and negative labels
 	for(i=0; i<length; i++)
 	{
-		if (ground_truth->get_value(i) >= 0)
+		if (ground_truth->get_label(i) >= 0)
 			pos_count++;
 		else
 			neg_count++;
@@ -106,7 +107,7 @@ float64_t CROCEvaluation::evaluate_roc(CLabels* predicted, CLabels* ground_truth
 
 		m_thresholds[i]=threshold;
 
-		if (ground_truth->get_value(idxs[i]) > 0)
+		if (ground_truth->get_label(idxs[i]) > 0)
 			tp+=1.0;
 		else
 			fp+=1.0;
--- a/src/shogun/transfer/multitask/MultitaskROCEvaluation.cpp
+++ b/src/shogun/transfer/multitask/MultitaskROCEvaluation.cpp
@@ -58,6 +58,12 @@ void CMultitaskROCEvaluation::set_indices(SGVector<index_t> indices)
 
 float64_t CMultitaskROCEvaluation::evaluate(CLabels* predicted, CLabels* ground_truth)
 {
+	REQUIRE(predicted->get_label_type()==LT_BINARY, "ROC evalution requires binary labels.");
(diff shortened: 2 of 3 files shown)


vigsterkr:
@lkuchenb thnx! there was an error actually in the `evaluation_thresholds_modular` script, which i think is actually an error that should be investigated as it just popped up now. just to be sure i'm re-running now the python modular test... let's see what's the outcome.

vigsterkr:
@lkuchenb ok so it's definitely an error caused by this change

lkuchenb:
I reverted parts of the fix. The predicted labels need to be accessed through `get_value()` as these are the confidence values returned from the machine and used for discrimination based on different thresholds for the ROC evaluation.

The ground truth labels however don't necessarily have confidence values. There is also no mechanism that ensures that if there are no confidence values, the labels are used instead.

The constructors

    CBinaryLabels::CBinaryLabels(SGVector<int32_t> src)
    CBinaryLabels::CBinaryLabels(SGVector<int64_t> src)

ensure that values and labels synchronously set. However, I was using

    CBinaryLabels::CBinaryLabels(int32_t num_labels)

src/shogun/evaluation/ROCEvaluation.cpp
```diff
@@ -19,15 +19,16 @@ CROCEvaluation::~CROCEvaluation()
 
 float64_t CROCEvaluation::evaluate(CLabels* predicted, CLabels* ground_truth)
 {
-	return evaluate_roc(predicted,ground_truth);
+	ASSERT(predicted->get_label_type()==LT_BINARY)
```
karlnapf: Can we use REQUIRE here and give the user information of what is wrong?

src/shogun/evaluation/ROCEvaluation.cpp
```diff
@@ -19,15 +19,16 @@ CROCEvaluation::~CROCEvaluation()
...
+	ASSERT(ground_truth->get_label_type()==LT_BINARY)
+
+	return evaluate_roc((CBinaryLabels*)predicted,(CBinaryLabels*)ground_truth);
 }
 
-float64_t CROCEvaluation::evaluate_roc(CLabels* predicted, CLabels* ground_truth)
+float64_t CROCEvaluation::evaluate_roc(CBinaryLabels* predicted, CBinaryLabels* ground_truth)
```
karlnapf: this is a good idea

src/shogun/transfer/multitask/MultitaskROCEvaluation.cpp
```diff
@@ -68,7 +71,7 @@ float64_t CMultitaskROCEvaluation::evaluate(CLabels* predicted, CLabels* ground_
 		//m_tasks_indices[t].display_vector();
 		predicted->add_subset(m_tasks_indices[t]);
 		ground_truth->add_subset(m_tasks_indices[t]);
-		result += evaluate_roc(predicted,ground_truth)/m_tasks_indices[t].vlen;
+		result += evaluate_roc((CBinaryLabels*)predicted,(CBinaryLabels*)ground_truth)/m_tasks_indices[t].vlen;
```
karlnapf: if you do static casts, be more explicit, i.e. put a comment that this is ok, or do the cast just below the assertions above and store the result in a variable

karlnapf:
Commits should be squashed to one

karlnapf:
Fine to merge apart from this.
Is ROC unit test covered?

lkuchenb:
Thanks for your comments, I integrated them and squashed the commits.

I don't know whether there is a ROC test case, but as I said, this is rather related to Labels coming with uninitialized confidence values. Possibly other features in shogun break when ground truth labels are passed with uninitialized confidence values.

src/shogun/evaluation/ROCEvaluation.cpp
```diff
@@ -19,15 +19,16 @@ CROCEvaluation::~CROCEvaluation()
 
 float64_t CROCEvaluation::evaluate(CLabels* predicted, CLabels* ground_truth)
 {
-	return evaluate_roc(predicted,ground_truth);
+	REQUIRE(predicted->get_label_type()==LT_BINARY, "ROC evalution requires binary labels.");
```
karlnapf: Sorry to be picky here:

 * We need a line break (>80chars)
 * There is a newline missing at the end of the string.
 * The message should give information what went wrong. `"Given labels (%d) must be binary (%d).\n", predicted->get_label_type(), LT_BINARY)`

src/shogun/evaluation/ROCEvaluation.cpp
```diff
@@ -19,15 +19,16 @@ CROCEvaluation::~CROCEvaluation()
...
+	REQUIRE(ground_truth->get_label_type()==LT_BINARY, "ROC evalution requires binary labels.");
+
+	return evaluate_roc((CBinaryLabels*)predicted,(CBinaryLabels*)ground_truth);
```
karlnapf: can you demangle this a bit?

src/shogun/transfer/multitask/MultitaskROCEvaluation.cpp
```diff
@@ -58,6 +58,12 @@ void CMultitaskROCEvaluation::set_indices(SGVector<index_t> indices)
 
 float64_t CMultitaskROCEvaluation::evaluate(CLabels* predicted, CLabels* ground_truth)
 {
+	REQUIRE(predicted->get_label_type()==LT_BINARY, "ROC evalution requires binary labels.");
+	REQUIRE(ground_truth->get_label_type()==LT_BINARY, "ROC evalution requires binary labels.");
+
+        CBinaryLabels* predicted_binary = (CBinaryLabels*)predicted;
```
karlnapf: ace

karlnapf:
OK, some minor things then we can merge.
About the unit test. If there was an uninitialised memory error earler, you should add a unit test that causes this error before your patch and that passes after your patch.

karlnapf:
Merging, as travis failures were unrelated

PR #3442, opened 2016-08-30
Feature: Add a cmake switch to disable colpack

lkuchenb:
Linking against colpack currently fails on Mac OS. There was no convenient way to turn it off.


karlnapf:
Thanks!
Whats the fail? Maybe open another issue if you have a minute

PR #3455, opened 2016-10-01
FIX #3218: Remove #ifdef for debugging.

### LINKED ISSUE #3218: Clean up QDA
* use `SGVector` everywhere rather than the old pointer, vlen approach
 * use eigen3/linalg
 * document
 * remove the `#ifdef` for debugging

souravsingh:
The issue is related to #3218


karlnapf:
Thanks!

PR #3458, opened 2016-10-04
Modular support for nystrom class

cfjhallgren:
Had to place SWIG lines after KernelRidgeRegression, since it is a parent class.


karlnapf:
Thanks!

PR #3461, opened 2016-10-12
Meta csharp bool vector

sorig:
Adds interface for translations of element getters/setters on specific vector and matrix types. E.g. a `BoolVector` can now have a separate translation rule from all other vector types.

This change requires the translation code to keep track of the type of initialised objects. I have therefore also added some basic type checking assertions on assignments and method calls. This may help catch typos etc when writing meta examples.

### DIFF
--- a/examples/meta/generator/targets/csharp.json
+++ b/examples/meta/generator/targets/csharp.json
@@ -5,7 +5,6 @@
     "Init": {
         "Construct": "$typeName $name = new $typeName($arguments)",
         "Copy": "$typeName $name = $expr",
-        "BoolVector": "var $name = new bool[$arguments]",
         "CharVector": "var $name = new char[$arguments]",
         "ByteVector": "var $name = new byte[$arguments]",
         "WordVector": "var $name = new ushort[$arguments]",
@@ -36,7 +35,6 @@
         "int": "int",
         "float": "float",
         "real": "double",
-        "BoolVector": "bool[]",
         "CharVector": "char[]",
         "ByteVector": "byte[]",
         "WordVector": "ushort[]",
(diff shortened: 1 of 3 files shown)


sorig:
@karlnapf have a look :)

examples/meta/generator/targets/csharp.json
```diff
@@ -78,11 +78,13 @@
     "Element": {
         "Access": {
             "Vector": "$identifier[$indices]",
-            "Matrix": "$identifier[$indices]"
+            "Matrix": "$identifier[$indices]",
+            "BoolVector": "$identifier.get_element($indices)"
```
karlnapf: Does the order matter here? Or are the specific rules always preferred?

sorig: Order doesn't matter. Specific types have precedence over the generic rules.

sorig:
I'll update the type map now

karlnapf:
Any ideas why we will get [this](https://travis-ci.org/shogun-toolbox/shogun/jobs/167111705#L1916) error?

sorig:
@karlnapf that error is fixed now. Forgot to remove the old `BoolVector` initialiser translation rule.

We get [this](https://travis-ci.org/shogun-toolbox/shogun/jobs/167301553#L2496) now. Seems like set_element and get_element aren't exposed in the swig interface. I checked the python interface and they aren't there either.

karlnapf:
Ah yeah, easy to fix, you need to expose this method to SWIG.
It is currently guarded with '#ifndef HAVE_SWIG' in 'SGVector.h' and 'SGMatrix.h', along with most methods in these classes. Exposing the getter/setter will fix the problem and won't hurt anywhere else

karlnapf:
Ill merge this and fix myself

PR #3472, opened 2016-10-24
enable MKL in GP

yorkerlin:
@karlnapf


yorkerlin:
The Python script can be found at https://gist.github.com/yorkerlin/9a64e00543c10d75db793b843ea00623

karlnapf:
The script should be turned into a cookbook for the GPs......
At least a meta example (super easy)
But I can do that

karlnapf:
Can you please update the news file?

karlnapf:
And as said, an api example would help

PR #3476, opened 2016-10-25
clean up examples

karlnapf:


karlnapf:
* Removes modular interfaces for all languages but python
 * removes all static interface examples and desriptions
 * removes all `GUI*` static interface classes of libshogun

karlnapf:
@vigsterkr please merge if you are fine with it

PR #3477, opened 2016-10-25
give each meta example a build target to avoid re-compilation of meta…

karlnapf:
… examples


karlnapf:
First build passed. Since all are the same I merge

PR #3484, opened 2016-10-29
More cleanup for QDA

souravsingh:
@karlnapf My try in cleaning up QDA. Related to #3218 

I would appreciate it if you can guide me through this.


karlnapf:
This patch is a nop, closing it.
Please only send patches that compile, and please dont send nop patches, they just eat our time

Issue #3485, opened 2016-10-29
add scala interface to docs and tests


karlnapf:
# Which is easy, since it uses the java jar. No swig additions needed, just the build and the docs
- meta examples
- integration testing in the above
- cookbook
- installation page
- INTERFACES.MD how to run

souravsingh:
@karlnapf I would like to know about how it needs to be done.

souravsingh:
@karlnapf I would like to know about how it needs to be done.

souravsingh:
@karlnapf I would like to know About how it needs to be done. I am quite beginner to Shogun contribution but am enthusiastic about contributing to it.

karlnapf:
Here is some start:
 * run cmake with enabling the meta examples being built
 * run `make meta_examples` and figure out what it does (you need to follow the cmake configuration files for that, best to google for a tutorial
 * run `make test` (need cmake switch for that) and figure out how the meta examples are executed

Once you have understood all that, you can check the `examples/meta` folder and the python scripts you need to modify to add scala. Then add a json for scala, then .... Then you need to figure out how to compile the generated listings with scala (rather than java)
The final result: making the meta examples generates a number of scala examples.
All this requires quite a bit of back-engineering how the whole process works.
Let me know if you have *concrete* questions

abhinavrai44:
Hi @karlnapf ,
I would like to solve this issue

karlnapf:
Its a not too hard (also not too easy) cmake / build issue. Only attempt if you know what you are doing, requires backwards-engineering of the existing build

abhinavrai44:
I have been looking into the issue. I am able to generate scala meta examples if you pass -DScalaModular=ON. This is what that need to be done, right ?

karlnapf:
That is step 1. Step2 would be to make them part of the build (i.e. they are executed when running `make test`)
Can you send a pull request so we can discuss there? This will take a few iterations

PR #3488, opened 2016-10-30
Add scala interface 

souravsingh:
Related to #3485. Please check @karlnapf


karlnapf:
This is not nearly what is needed. 
I mean I dont want to discourage you, but maybe consider starting with something simpler.
Ill close this for now as it is a nop.

souravsingh:
@karlnapf I would like to learn about Shogun contribution as quickly as possible. Hence I am picking up complicated issues. I would like to have some guidance in this. I apologize if I made a bad contribution.

PR #3502, opened 2016-10-31
Feature/parallel xval fix

karlnapf:
This reverts @Saurabh7 parallel xvalidation patch, as it produced memory errors which made the xvalidation notebook fail. I will create a feature branch for @Saurabh7 to fix this. Please run valgrind next time

### DIFF
--- a/doc/ipython-notebooks/evaluation/xval_modelselection.ipynb
+++ b/doc/ipython-notebooks/evaluation/xval_modelselection.ipynb
@@ -1,809 +1,826 @@
 {
- "metadata": {
-  "name": "xval_modelselection",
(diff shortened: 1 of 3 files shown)


karlnapf:
Here is your feature branch with your old commit, it would be great if we could put this back in, but working.....

https://github.com/shogun-toolbox/shogun/tree/feature/parallel_xval

Saurabh7:
@karlnapf Notebook runs fine for me locally (on feature branch) after changing `set_conf_int_alpha` method which has since been removed from class. I had run valgrind , tests look clean too.

karlnapf:
Both @vigsterkr and me were getting memory errors, manifesting themselves in the paramter tree at the end of the notebook having a weird name, probably some buffer overwritten or so.

Issue #3522, opened 2016-11-09
fix links in introduction notebook


karlnapf:
http://www.shogun-toolbox.org/notebook/latest/Introduction.html

Also its description should read "Introduction to Machine Learning"

Super easy entrance task

karlnapf:
In fact I just realised all notebooks will have broken links.

The best solve would be to use our shortcuts, like this
[CGaussianKernel](shogun.ml/CGaussianKernel)

Use grep to replace and correct in all notebooks at once. Single commit

piyushgoel997:
Hi, I'm new here and wish to take up this issue. So, kindly guide me through.

Shashi456:
can you tell me what you mean by markdown

Shashi456:
Broken Links : StatLib Library , this notebook (both places), CAccuracyMeasure,  Dense features

PR #3524, opened 2016-11-09
Fix a few links in the notebook

souravsingh:
Related to #3522

### DIFF
--- a/doc/ipython-notebooks/intro/Introduction.ipynb
+++ b/doc/ipython-notebooks/intro/Introduction.ipynb
@@ -124,7 +124,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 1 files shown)


karlnapf:
See my updated description in the issue.
This needs to be applied to all notebooks.....

karlnapf:
Thanks! This is good. But can you fix all links?
The String features ones also appear to be broken.

karlnapf:
We'll fix the other ones differently. Thanks for this

souravsingh:
Thanks for merging the PR @karlnapf. I would like to finish up fixing the rest of the notebooks.

karlnapf:
the opencv one is the only one failing for now I think, this is due to an API change arising from re-structuring the opencv codebase....

Issue #3492, opened 2016-10-30
rename interface cmake switches


karlnapf:
PythonModular -> INTERFACE_PYTHON

etc

PR #3525, opened 2016-11-09
Rename interface switches in cmake

souravsingh:
Related to #3492


vigsterkr:
ok good start but this will totally break backward compatibility... see the failures in travis ;)

karlnapf:
Please `git grep` for the interface names in the shogun root directory. They are hidden at more places.

We can only merge this for the next major release unfortunately. But it is appreciated anyways

PR #3527, opened 2016-11-09
Converted from an older notebook format (v3) to the format (v4)

amarlearning:
Converted from an older notebook format (v3) to the format (v4)


karlnapf:
Thanks!

amarlearning:
@karlnapf should I now modify the notebook according to the new template ?

PR #3532, opened 2016-11-12
Correct the link  doc/readme/INSTALL.md points at

mickyaero:
I was installing shogun recently and found that 
 doc/readme/INSTALL.md  pointed at the wrong location.


karlnapf:
Thanks! :)

PR #3534, opened 2016-11-13
LinalgRefactor - Cholesky - CPU only

OXPHOS:


cmake/external/GoogleTestNMock.cmake
```diff
@@ -8,8 +8,8 @@ SET (GMOCK_REVISION 443)
 include(ExternalProject)
 ExternalProject_Add(
 	GoogleMock
-	URL http://googlemock.googlecode.com/files/gmock-1.7.0.zip
-	URL_MD5 073b984d8798ea1594f5e44d85b20d66
+	URL https://github.com/google/googletest/archive/release-1.8.0.tar.gz
```
karlnapf: Can you please cherry pick this from the develop branch?
Otherwise we will have a merge conflict

src/shogun/mathematics/linalg/LinalgBackendBase.h
```diff
@@ -112,6 +118,20 @@ class LinalgBackendBase
 	#undef BACKEND_GENERIC_IN_PLACE_ADD
 
 	/**
+	 * Wrapper method of cholesky decomposition.
```
karlnapf: remove "method of"

src/shogun/mathematics/linalg/LinalgBackendEigen.h
```diff
@@ -284,6 +300,28 @@ class LinalgBackendEigen : public LinalgBackendBase
...
+
+		Eigen::LLT<Eigen::Matrix<T,Eigen::Dynamic,Eigen::Dynamic> > llt(a_eig);
+		//compute matrix L or U
+		if(lower==false)
+			c_eig = llt.matrixU();
+		else
+			c_eig = llt.matrixL();
+
+		// checking for success
+		REQUIRE(llt.info()!=Eigen::NumericalIssue, "Matrix is not Hermitian positive definite!\n");
```
karlnapf: Can you comment (in source code) on the possible outcomes of the eigen3 info?
And can you properly translate this into a shogun error message?

@vigsterkr what about some proper exception handling where we forward such numerical errors as numerical errors rather than just standard exceptions that make shogun stop?

tests/unit/CMakeLists.txt
```diff
@@ -4,7 +4,7 @@ ExternalProject_Get_Property(GoogleMock source_dir)
 if(SYSTEM_INCLUDES)
 	INCLUDE_DIRECTORIES(SYSTEM ${SYSTEM_INCLUDES})
 endif()
-include_directories(${source_dir}/include ${INCLUDES} ${source_dir}/gtest/include ${CMAKE_CURRENT_SOURCE_DIR})
```
karlnapf: cherry pick

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
```diff
@@ -90,6 +91,44 @@ TEST(LinalgBackendEigen, SGMatrix_add_in_place)
...
+	const index_t size=2;
+	SGMatrix<float64_t> m(size, size);
+
+	m(0,0)=2.0;
+	m(0,1)=1.0;
+	m(1,0)=1.0;
+	m(1,1)=2.5;
+
+	//lower triangular cholesky decomposition
+	SGMatrix<float64_t> L = cholesky(m);
```
karlnapf: please also check for correct dimensions of output

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
```diff
@@ -90,6 +91,44 @@ TEST(LinalgBackendEigen, SGMatrix_add_in_place)
...
+	m(0,1)=1.0;
+	m(1,0)=1.0;
+	m(1,1)=2.5;
+
+	//upper triangular cholesky decomposition
+	SGMatrix<float64_t> U = cholesky(m,false);
+
+	Map<MatrixXd> map_A(m.matrix,m.num_rows,m.num_cols);
+	Map<MatrixXd> map_U(U.matrix,U.num_rows,U.num_cols);
+	EXPECT_NEAR((map_A-map_U.transpose()*map_U).norm(),
```
karlnapf: same here, please check for dimensions

src/shogun/mathematics/linalg/LinalgBackendEigen.h
```diff
@@ -284,6 +300,28 @@ class LinalgBackendEigen : public LinalgBackendBase
...
+	/** Eigen3 Cholesky decomposition */
+	template <typename T>
+	SGMatrix<T> cholesky_impl(const SGMatrix<T>& a, const bool lower) const
+	{
+		SGMatrix<T> c(a.num_rows, a.num_cols);
+		set_const_impl<T>(c, 0);
+		typename SGMatrix<T>::EigenMatrixXtMap a_eig = a;
+		typename SGMatrix<T>::EigenMatrixXtMap c_eig = c;
+
+		Eigen::LLT<Eigen::Matrix<T,Eigen::Dynamic,Eigen::Dynamic> > llt(a_eig);
```
karlnapf: Can we please have a boolean flag (or a separate method with separate tests) for LLT and LDLT.

karlnapf:
Ok so lets do the minor cleanups and then merge this one. 
BTW it is not useful without a triangular solver, maybe this patch should already include it?

OXPHOS:
@karlnapf For the triangular solver, I'm thinking: 
- Add a separate interface such as `SGVector<T> triangular_solver(SGMatrix<T> A, SGVector<T> b)` and call `Eigen3.llt().solve()` inside `triangular_solver()`. This means `triangular_solver()` and `choledsky_llt()` will be two independent methods, and user can still call `cholesky_llt()` for whatever purposes (is this necessary?)

- I assume we still need to implement QR decomposition etc. some time later, so maybe we can implement a generic interface, for eg. `SGVector<T> linear_solver(SGMatrix<T> A, SGVector<T> b, enum decomposition_type ).

karlnapf:
There is upsides and downsides with that.
Maybe we can use "cholesky_solve", like scipy? One methods per factorisation .... I dont like passing enums on the decomposition types around too much

https://docs.scipy.org/doc/scipy-0.14.0/reference/generated/scipy.linalg.cho_solve.html
https://docs.scipy.org/doc/scipy-0.14.0/reference/generated/scipy.linalg.cho_factor.html

karlnapf:
Let me know when this is ready to be merged

karlnapf:
Ok let me know once this is done

OXPHOS:
@karlnapf squashed and ready

src/shogun/mathematics/linalg/LinalgBackendEigen.h
```diff
@@ -284,6 +309,58 @@ class LinalgBackendEigen : public LinalgBackendBase
...
+		 * 1: Eigen::NumericalIssue. The provided data did not satisfy the prerequisites.
+		 */
+		REQUIRE(llt.info()!=Eigen::NumericalIssue, "Matrix is not Hermitian positive definite!\n");
+
+		return c;
+	}
+
+	/** Eigen3 Cholesky solver */
+	template <typename T>
+	SGVector<T> cholesky_solver_impl(const SGMatrix<T>& a, const SGVector<T>& b) const
```
karlnapf: Mmmh, I think this is missing the point a bit.
We want a method that takes a cholesky factor and then solves the system here, it should just be the triangular solve. 
So that I can precompute the Cholesky factor and then solve multiple systems with it without having to re-factor again. 
See what I mean?

OXPHOS: @karlnapf Do you mean that users explicitly call `cholesky_factor` and pass the matrix to `cholesky_solver`? I saw that's how scipy achieved `cholesky_solver`. But they called `lapack::potrs` directly in the function, which can solve `Lu=b`, while in `Eigen3`, the solver is wrapped in `cholesky_factor`. Here I just created a new `llt` instance and it came with the triangular solver.  I think we will have to code back substitution from scratch if we don't call `eigen3.llt.solve()`.

karlnapf: Ah and keep the current way of just computing a cholesky based solve once. In addition, there should be the option of providing a precomputed cholesky factor

```
L = cholesky_factor(X);
alpha1 = chol_solve(L, a, true);
alpha2 = chol_solve(L, b, true);

// and the direct solve (re-factors the matrix)
alpha3 = solve(X, b, ST_CHOLESKY);
```

something like this

src/shogun/mathematics/linalg/LinalgBackendEigen.h
```diff
@@ -33,6 +33,7 @@
 #include <shogun/lib/SGVector.h>
 #include <shogun/mathematics/eigen3.h>
 #include <shogun/mathematics/linalg/LinalgBackendBase.h>
+#include <Eigen/Core>
```
karlnapf: Not sure this is the clean way to do this. Maybe rather in eigen3.h
@vigsterkr ?

OXPHOS: @karlnapf I don't have to include `<Eigen/Core>` to make it work on my machine, and the codes failed on travis even with the header. It seems that  `Eigen::UpLoType` isn't taken as a known enum class on travis and I am not sure why.

karlnapf: This will have to do with different eigen3 versions. The way to solve this is to use guards on the eigen3 version, grep for `#if EIGEN_VERSION_AT_LEAST(3,0,93)`. As we want shogun to work with all the versions ... Check the docs for when this type was defined etc.

karlnapf:
Let me know how things go and whether you need help

OXPHOS:
@karlnapf I bypassed the class name issue but the `(TriangularView.transpose()).solve` is not working on Travis. It is also a version issue and I am trying to make it backward compatible. Or do you think we should just keep it like this but for the users to upgrade `Eigen` by guarding?

OXPHOS:
UPDATE: I downloaded Eigen 3.1.2 on my machine (the minimum required by shogun's cmake) and the all tests can pass. So I am not sure what causes the `inf` return value on Travis

karlnapf:
I'll check this these days...
Is everything else in here clean? Then I'll clone locally and try to come up with a fix...
Ping me if nothing happens within a week or two

yorkerlin:
@OXPHOS  
Nice work!
Please let me know once you implement these solvers using GPU acceleration. I will do test them in GP.

karlnapf:
Merge conflicts

OXPHOS:
@karlnapf updated

src/shogun/mathematics/linalg/LinalgBackendEigen.h
```diff
@@ -294,6 +321,61 @@ class LinalgBackendEigen : public LinalgBackendBase
...
+		set_const_impl<T>(x, 0);
+		typename SGMatrix<T>::EigenMatrixXtMap L_eig = L;
+		typename SGVector<T>::EigenVectorXtMap b_eig = b;
+		typename SGVector<T>::EigenVectorXtMap x_eig = x;
+
+		if (lower == false)
+		{
+			Eigen::TriangularView<Eigen::Matrix<T,Eigen::Dynamic,Eigen::Dynamic>,
+				Eigen::Upper> tlv(L_eig);
+			x_eig = (tlv.transpose()).solve(tlv.solve(b_eig));
```
karlnapf: And this does a triangular solve now? Cool!

src/shogun/mathematics/linalg/LinalgNamespace.h
```diff
@@ -543,6 +580,37 @@ SGVector<T> rowwise_sum(const Block<SGMatrix<T>>& a, bool no_diag=false)
 	return sg_linalg->get_cpu_backend()->rowwise_sum(a, no_diag);
 }
 
+/**
+ * Generic linear solver
+ *
+ * @param A left-hand side matrix
+ * @param b right-hand side array
+ * @param solver_type solver method 1: Cholesky solver
```
karlnapf: Please dont enumerate the methods

src/shogun/mathematics/linalg/LinalgNamespace.h
```diff
@@ -543,6 +580,37 @@ SGVector<T> rowwise_sum(const Block<SGMatrix<T>>& a, bool no_diag=false)
...
+	LINALG_SOLVER_TYPE solver_type)
+{
+	SGMatrix<T> L;
+	SGVector<T> result;
+	switch (solver_type)
+	{
+		/* Solve the linear equations \f$Ax=b\f$,
+		 * where \f$A\f$ is a Hermitian positive definite matrix
+		*/
+		case ST_CHOLESKY:
```
karlnapf: You think it is better to directly call eigen3 here? Then some overhead is avoided.... (storing the lower factor)

src/shogun/mathematics/linalg/LinalgNamespace.h
```diff
@@ -543,6 +580,37 @@ SGVector<T> rowwise_sum(const Block<SGMatrix<T>>& a, bool no_diag=false)
...
+			L = cholesky_factor(A);
+			result = cholesky_solver(L, b);
+			break;
+		default:
+			SG_SERROR("%s:Unknown solver type!\n", solver_type);
```
karlnapf: It would be cool if we had: QR, LU, and SVD.
But next PR I guess :)

tests/unit/regression/lars_unittest.cc
```diff
@@ -396,7 +395,7 @@ TEST(LeastAngleRegression, cholesky_insert)
 	
 	MatrixXd XX=mat_old.transpose()*mat_old;
 	// Compute matrix R which has to be updated
-	SGMatrix<float64_t> R_old=linalg::cholesky(XX, false);
+	SGMatrix<float64_t> R_old=linalg::cholesky_factor(SGMatrix<float64_t>(XX), false);
```
karlnapf: in lars, we also update the cholesky.
This is a method `linalg` should also have, put it on the list :)

src/shogun/mathematics/linalg/LinalgNamespace.h
OXPHOS: I removed the whole generic solver since we're not enumerating the method

src/shogun/mathematics/linalg/LinalgBackendEigen.h
```diff
@@ -294,6 +321,61 @@ class LinalgBackendEigen : public LinalgBackendBase
...
+	{
+		SGVector<T> x(b.size());
+		set_const_impl<T>(x, 0);
+		typename SGMatrix<T>::EigenMatrixXtMap L_eig = L;
+		typename SGVector<T>::EigenVectorXtMap b_eig = b;
+		typename SGVector<T>::EigenVectorXtMap x_eig = x;
+
+		if (lower == false)
+		{
+			Eigen::TriangularView<Eigen::Matrix<T,Eigen::Dynamic,Eigen::Dynamic>,
```
karlnapf: I think this should be selfAdjointView btw

karlnapf: because we know the factor is square

karlnapf:
There is still this inf error on travis. What was it again with that?

OXPHOS:
My machine returns correct results with Eigen 3.1.2 (The minimum required) and the newest version of Eigen so I am actually not sure what happens on travis. (Though I am now thinking maybe there's some problem with the specific version of Eigen Travis is using?)

karlnapf:
Yes it will be a specific version problem.

Would you mind getting the same version that is running on travis (ubuntu 14.04, installed from the repository, see `config/Dockerfile`), and compile with that locally to find out whether you can reproduce the error locally?

Next step: how to fix (I bet it has to do with the upper/lower interpretation changed or something)

Then, if that is possible, we need to write a guard that fixes the problem for the particular eigen3 version

OXPHOS:
Update:
Ubuntu14.04 uses `Eigen3 3.2.0-8` according to [this page](https://launchpad.net/ubuntu/trusty/+source/eigen3). I downloaded `Eigen3 3.2.0` and added `3.2.0-8` patch manually - not sure whether I missed something but the test can pass on my machine. 
I don't understand why Ubuntu seems using "unstable" versions all the time but I did test the "stable" versions that can downloaded from [here](https://bitbucket.org/eigen/eigen/downloads?tab=tags) with a script [here](https://gist.github.com/OXPHOS/68f9d5eba66fed7ce61d024f4523bf80#file-cholesky_solver_test-sh). All `Eigen3 3.2.x`, `Eigen3 3.1.3` and `Eigen3 3.1.4` give correct outputs on `cholesky_solver` test. 
However, `Eigen3 3.3.x` lead to multiple mixed-type [errors](https://gist.github.com/OXPHOS/68f9d5eba66fed7ce61d024f4523bf80#file-eigen3-3-3-x-error-example) and I am looking into them.
@karlnapf

karlnapf:
One way to go ahead here would be to disable the tests, and put a warning about the eigen3 version in cmake, see https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/CMakeLists.txt#L222

OXPHOS:
Update: I reproduced the error in Docker and the trouble came from [this line](https://github.com/shogun-toolbox/shogun/pull/3534/files#diff-a9b6ebe755ace8ac2e3ecd1390090bffR373). I'll see what I can do about it.

OXPHOS:
Note: Compiling error associated with `Eigen3.3.x` was documented in #3644

OXPHOS:
@karlnapf The error is produced on Travis due to the misbehavior of  `Eigen::TriangularView.solve()`. Not sure how this is happening but I temporarily disabled `Cholesky_solver()` test. Otherwise it is ready to merge.

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
```diff
@@ -90,6 +91,75 @@ TEST(LinalgBackendEigen, SGMatrix_add_in_place)
...
+		0.0, 1E-15);
+	EXPECT_EQ(m.num_rows, U.num_rows);
+	EXPECT_EQ(m.num_cols, U.num_cols);
+}
+
+// Temporarily disabled because Travis fails to calculate:
+// Eigen::TriangularView.solve(vector)
+TEST(LinalgBackendEigen, DISABLED_SGMatrix_cholesky_solver)
```
karlnapf: please dont disable, but guard with the eigen3 version that makes it fail.
Put a warning message similar to https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/CMakeLists.txt#L222 in the cmake configuration

PR #3538, opened 2016-11-14
fix for #3460

karlnapf:


karlnapf:
The proper fix:

 * remove the `extern "C"` in the `sg_print_function.cpp`
 * only do when R version is more recent than 3.3.0 since previously it was necessary
 * for that, cmake needs to detect the R version

karlnapf:
@vigsterkr thoughts?

karlnapf:
Actually, I think we should just remove the extern C business.
See the bottom of
http://www.hep.by/gnu/r-patched/r-exts/R-exts_97.html

karlnapf:
Updated the fix to remove extern C -- the R manual says this should have never been done.
This will fix #3460 and should be backwards compatible

PR #3550, opened 2016-11-18
fix bool arrays in meta examples

karlnapf:


karlnapf:
@OXPHOS This makes the cart tree example (aka using boolean vectors as parameters to Shogun methods) work. I am waiting for travis and might have to add some fixes, but generally this should work now

#3282

Issue #3535, opened 2016-11-14
Propper error message when meta-testsuite doesnt exist


karlnapf:
See
http://pastebin.com/yQXbN8N2

It should tell to update the git submodule, so people don't need to ask what this error is about.

This is an extremely minimal cmake related entrance task, one line

sudk1896:
@karlnapf: I obviously know how to solve this :)
Which file should I be editing here ?

karlnapf:
Grep is your friend here

karlnapf:
It should be in install already

sudk1896:
@karlnapf: The instruction for git init submodule is already given  [here](https://github.com/shogun-toolbox/docs/blob/master/INSTALL.md#basics-). It says whenever yo've trouble to do `git submodule update --init`. Should I still add something to the `INSTALL` Readme.

karlnapf:
This is not about changing readmes, but the cmake error message that you get when the data module is not present.

karlnapf:
Fixed by #3551

PR #3551, opened 2016-11-18
echo solution for CMake error  #3535

hongguangguo:
echo solution for CMake error, for there is no meta examples.#3535


karlnapf:
For the future, can you please *update* the pull request rather than opening a new one? You have to add `-f` to your `git push` command for that.

karlnapf:
Thanks!

hongguangguo:
ok, I learn a lot, Thank you very much!

PR #3562, opened 2016-11-23
LinalgRefactor - Matrix elementwise product

OXPHOS:


src/shogun/mathematics/linalg/LinalgBackendEigen.h
```diff
@@ -291,6 +310,30 @@ class LinalgBackendEigen : public LinalgBackendBase
...
+		typename SGMatrix<T>::EigenMatrixXtMap b_eig = b;
+		typename SGMatrix<T>::EigenMatrixXtMap c_eig = c;
+
+		c_eig = a_eig.array() * b_eig.array();
+		return c;
+	}
+
+	/** Eigen3 matrix in-place elementwise product method */
+	template <typename T>
+	void element_prod_impl(SGMatrix<T>& a, SGMatrix<T>& b, SGMatrix<T>& result) const
```
karlnapf: can you remind me of the discussion we had here on preallocated memory for results?

src/shogun/mathematics/linalg/LinalgNamespace.h
```diff
@@ -193,6 +193,48 @@ T dot(const SGVector<T>& a, const SGVector<T>& b)
...
+{
+	REQUIRE(a.num_rows == b.num_rows && a.num_cols == b.num_cols,
+			"Dimension mismatch! A(%d x %d) vs B(%d x %d)\n",
+			a.num_rows, a.num_cols, b.num_rows, b.num_cols);
+	return infer_backend(a, b)->element_prod(a, b);
+}
+
+/** Performs the operation C = A .* B where ".*" denotes elementwise multiplication.
+ *
+ * This version returns the result in-place.
```
karlnapf: And also if/whether it is possible to pass one of the operands arguments (A or B) as a result? (since you do that in unit test)

tests/unit/mathematics/linalg/operations/Viennacl_operations_unittest.cc
```diff
@@ -132,6 +132,54 @@ TEST(LinalgBackendViennaCL, SGVector_dot)
...
+TEST(LinalgBackendViennaCL, SGMatrix_elementwise_product_in_place)
+{
+	sg_linalg->set_gpu_backend(new LinalgBackendViennaCL());
+
+	SGMatrix<float64_t> A(3,3);
+	SGMatrix<float64_t> B(3,3);
+	SGMatrix<float64_t> C(3,3);
+	SGMatrix<float64_t> A_gpu, B_gpu;
+
+	for (index_t i = 0; i < 9; ++i)
```
karlnapf: style, please use A.num_rows * A.num_cols here

karlnapf: Same everywhere else

src/shogun/mathematics/linalg/LinalgBackendEigen.h
```diff
@@ -291,6 +310,30 @@ class LinalgBackendEigen : public LinalgBackendBase
...
+	{
+		typename SGMatrix<T>::EigenMatrixXtMap a_eig = a;
+		typename SGMatrix<T>::EigenMatrixXtMap b_eig = b;
+		typename SGMatrix<T>::EigenMatrixXtMap result_eig = result;
+
+		result_eig = a_eig.array() * b_eig.array();
```
karlnapf: and this works if a==result?
What about eigens aliasing?

src/shogun/mathematics/linalg/LinalgBackendEigen.h
OXPHOS: @karlnapf we had discussions in #3387 and #3391. So it seems that two have two keep two interfaces. But I am now thinking maybe I can call refactor current `Matrix func(Matrix a, Matrixb)` to:
```
alloc(result, num_rows, num_cols);
func_in_place(a, b, result);
return result
```
So the codes won't be duplicated.

Also I am thinking whether it's possible to have a wrapper method for all non-in-place methods, such as
```
Container<T> wrapper(Container<T> a, Container<T> b, lambda x)
{
    alloc(result, size);
    x(a, b, result);
    return result;
}

File: src/shogun/mathematics/linalg/LinalgBackendEigen.h (hunk 4 above)
Comment (OXPHOS) [inline src/shogun/mathematics/linalg/LinalgBackendEigen.h]: I am not sure I understand your question...It works if `a` and `result` point to the same memory

File: src/shogun/mathematics/linalg/LinalgBackendEigen.h (hunk 1 above)
Comment (karlnapf) [inline src/shogun/mathematics/linalg/LinalgBackendEigen.h]: Absolutely do that :)

**Comment by OXPHOS** [comment]
@karlnapf merged in place and not in place methods into one. I don't think passing algebra methods as params can be easily achieved here so I gave up on the generic wrapper idea.

File: src/shogun/mathematics/linalg/LinalgBackendEigen.h (hunk 4 above)
Comment (karlnapf) [inline src/shogun/mathematics/linalg/LinalgBackendEigen.h]: It is dangerous whenever you are overwriting memory that is used for the computation.
This is why eigen3 often allocates temp memory. Make sure you really understand the mechanics here. Array is just a view onto matrices that allows for element wise operations (different operator * overloading for example)

Comment (karlnapf) [inline src/shogun/mathematics/linalg/LinalgBackendEigen.h]: I mean if the test works, then I am fine with it

**Comment by karlnapf** [comment]
Yes, let's not re-invent the wheel there ;)

File: cmake/external/GoogleTestNMock.cmake [hunk 5]
```diff
@@ -8,8 +8,8 @@ SET (GMOCK_REVISION 443)
 include(ExternalProject)
 ExternalProject_Add(
 	GoogleMock
-	URL http://googlemock.googlecode.com/files/gmock-1.7.0.zip
-	URL_MD5 073b984d8798ea1594f5e44d85b20d66
+	URL https://github.com/google/googletest/archive/release-1.8.0.tar.gz
```
karlnapf: rebase for this

src/shogun/mathematics/linalg/LinalgBackendViennacl.h
```diff
@@ -261,6 +272,18 @@ class LinalgBackendViennaCL : public LinalgBackendGPUBase
...
 
+	/** ViennaCL matrix in-place elementwise product method */
+	template <typename T>
+	void element_prod_impl(SGMatrix<T>& a, SGMatrix<T>& b, SGMatrix<T>& result) const
+	{
+		GPUMemoryViennaCL<T>* a_gpu = cast_to_viennacl(a);
+		GPUMemoryViennaCL<T>* b_gpu = cast_to_viennacl(b);
+		GPUMemoryViennaCL<T>* result_gpu = cast_to_viennacl(result);
+
+		result_gpu->data_matrix(a.num_rows, a.num_cols) = viennacl::linalg::element_prod(
```
karlnapf: btw do you have a gpu so that this code is actually tested on your machine?
Remember that our new implementation falls back to eigen3 if it is not available .....

src/shogun/mathematics/linalg/LinalgNamespace.h
```diff
@@ -193,6 +193,56 @@ T dot(const SGVector<T>& a, const SGVector<T>& b)
...
+{
+	REQUIRE(a.num_rows == b.num_rows && a.num_cols == b.num_cols,
+			"Dimension mismatch! A(%d x %d) vs B(%d x %d)\n",
+			a.num_rows, a.num_cols, b.num_rows, b.num_cols);
+	REQUIRE(a.num_rows == result.num_rows && a.num_cols == result.num_cols,
+			"Dimension mismatch! A(%d x %d) vs result(%d x %d)\n",
+			a.num_rows, a.num_cols, result.num_rows, result.num_cols);
+
+	REQUIRE(!(result.on_gpu()^a.on_gpu()),
+			"Cannot operate with matrix result on_gpu (%d) and matrix A on_gpu (%d).\n", result.on_gpu(), a.on_gpu());
```
karlnapf: minor. character line limit

src/shogun/mathematics/linalg/LinalgBackendViennacl.h
OXPHOS: Yes I am positive the codes are tested on GPU on my machine

PR #3563, opened 2016-11-24
LinalgRefactor - Remove old linalg::range_fill and linalg::max

OXPHOS:


karlnapf:
Can you send viktors patch in a separate first PR?

src/shogun/mathematics/linalg/internal/modules/Redux.h
```diff
@@ -196,15 +185,15 @@ typename implementation::int2float<typename Matrix::Scalar>::floatType mean(
...
+ * Wrapper method for internal implementation of matrix rowwise mean of values
  * that works with generic dense matrices
  *
  * @param m the matrix whose rowwise mean has to be computed
  * @param no_diag if true, diagonal entries are excluded from the mean (default - false)
  * @return the rowwise mean computed as \f$\1/N \sum_{j=1}^N m_{i,j}\f$
  */
 template <Backend backend=linalg_traits<Redux>::backend,class Matrix>
-SGVector<typename implementation::rowwise_mean<backend,Matrix>::ReturnDataType> 
+SGVector<typename implementation::rowwise_mean<backend,Matrix>::ReturnDataType>
```
karlnapf: thinking about this. Most ndarray libraries that might support higher order tensors accept an axis for mean computation (rowwise would be 1, column wise would be 0), etc. Might be something to keep in mind for unification ...

karlnapf:
Nevermind about the seperate patch, this just refactors things, so merging

PR #3564, opened 2016-11-25
Temp

sorig:
@karlnapf don't know why git is confused about the entire history. Only the last commit is interesting.


karlnapf:
This needs a rebase with the orginal feature branch. Does that make sense?
Or just drop things then?

PR #3568, opened 2016-11-27
Add cosine-distance meta example

IOcodegeass:
Added **cosine.sg** and **cosine.rst** to create meta examples for distance-cosine and generated cookbook for distance-cosine.


karlnapf:
Can you please not re-create pull requests? You can update the old one instead

PR #3570, opened 2016-11-30
Add missing get_width() and set_width() functions to WeightedDegreeRBFKernel

lkuchenb:


src/shogun/kernel/WeightedDegreeRBFKernel.h
```diff
@@ -77,6 +77,19 @@ class CWeightedDegreeRBFKernel: public CDotKernel
 		virtual EFeatureType get_feature_type() { return F_DREAL; }
 
 
+		/** set width
```
karlnapf: Minor: Capital "S"

src/shogun/kernel/WeightedDegreeRBFKernel.h
```diff
@@ -77,6 +77,19 @@ class CWeightedDegreeRBFKernel: public CDotKernel
...
+		 *
+		 * @param w new width
+		 * @return if setting was successful
+		 */
+		inline bool set_width(float64_t w) { width=w; return true; }
```
karlnapf: no need for inline, and also no need for a bool return

src/shogun/kernel/WeightedDegreeRBFKernel.h
```diff
@@ -77,6 +77,19 @@ class CWeightedDegreeRBFKernel: public CDotKernel
...
+
+		/** get width
```
karlnapf: Same as above

karlnapf:
Thanks! Minor fixes and good to go.

lkuchenb:
I just kept that consistent with the other getter / setter functions in the class. Do you want me to change those, too?

src/shogun/kernel/WeightedDegreeRBFKernel.h
```diff
@@ -77,6 +77,18 @@ class CWeightedDegreeRBFKernel: public CDotKernel
 		virtual EFeatureType get_feature_type() { return F_DREAL; }
 
 
+		/** Set width
+		 *
+		 * @param w new width
+		 */
+		void set_width(float64_t w) { width=w; }
```
karlnapf: we should actually assert positivity here, but since this will change in the future, nevermind :)

karlnapf:
It is not high priority to change the others...but if you want to, go for it, cleanups are always good

PR #3574, opened 2016-12-05
LinalgRefactor - Set rows const

OXPHOS:


src/shogun/mathematics/linalg/LinalgBackendViennaclKernels.h
```diff
@@ -90,6 +90,47 @@ namespace shogun
...
 
+	/** Generates the set_rows_const computation kernel
+	 * The OpenCL kernel that helps to set rows const to SGMatrix from SGVector
+	 */
+	template <class T>
+	static viennacl::ocl::kernel& generate_set_rows_const_kernel()
+	{
+		std::string kernel_name = "set_rows_const_" +
+			linalg::implementation::ocl::get_type_string<T>();
+
```
karlnapf: This needs to be tested. Is it tested on your machine? Travis cannot check it...

src/shogun/mathematics/linalg/LinalgBackendBase.h
```diff
@@ -228,7 +228,20 @@ class LinalgBackendBase
 	DEFINE_FOR_ALL_PTYPE(BACKEND_GENERIC_SET_CONST, SGVector)
 	DEFINE_FOR_ALL_PTYPE(BACKEND_GENERIC_SET_CONST, SGMatrix)
 	#undef BACKEND_GENERIC_SET_CONST
-	
+
+	/**
+	 * Wrapper method that sets const values to vectors or matrices.
```
karlnapf: I am a nagger, i know.
BUT we should not offer such a method. It is extremely cache unfriendly -- matrices are stored in column major format in Shogun. This is why if you change a single element of all rows, this totally screws up the prefetchers hard work.
Devs should be forced to set the columns to constant values, and then potentially transpose afterwards.
@vigsterkr @lambday thoughts?

OXPHOS: @karlnapf Would you suggest to change it to `set_col_const` or just remove it? I didn't see any call on this method in Shogun

karlnapf: If it is really not used, kill it.
Otherwise, having a column version of it is ok, but make sure to use memset in there

PR #3582, opened 2016-12-19
LinalgRefactor - Matrix product

OXPHOS:
- Removed the `bool transpose` parameter and users should transpose the matrix by themselves (currently in `class SGMatrix`).
- Eigen tests passed. 
- I didn't test ViennaCL because my computer doesn't support float/double calculation on GPU but it seems OpenCL has problem dealing with integer sparse matrix ([link](https://github.com/viennacl/viennacl-dev/issues/153))


karlnapf:
I don't agree on the transpose. This flag should be to actually be able to avoid transposing at all, but rather perform the corresponding operation without changing the memory layout of the matrix.

E.g.
 * https://www.tensorflow.org/api_docs/python/math_ops/matrix_math_functions#matmul
 * Numpy's transpose actually doesnt transpose but just creates a view
 * Eigen3's `adjoint()` creates a view as well and doesnt change the memory

src/shogun/lib/SGMatrix.h
```diff
@@ -130,6 +131,15 @@ template<class T> class SGMatrix : public SGReferencedData
...
+		/** Resize matrix
+		 * If the new matrix is larger, fill the new memory with 0.
+		 *
+		 * @param rows new row numbers
+		 * @param cols new column numbers
+		 * @return if resizing was successful
+		 */
+		bool resize_matrix(index_t rows, index_t cols);
+
 		/** Get a column vector
```
OXPHOS: These might need separate patch?

src/shogun/mathematics/linalg/LinalgNamespace.h
```diff
@@ -361,6 +361,124 @@ SGMatrix<T> element_prod(SGMatrix<T>& a, SGMatrix<T>& b)
...
+					B.num_rows, result.num_cols);
+		}
+		else
+		{
+			REQUIRE(A.num_cols == B.num_rows, "Number of columns for A (%d) and "
+					"number of rows for B (%d) should be equal!\n", A.num_cols, B.num_rows);
+			REQUIRE(B.num_cols == result.num_cols, "Number of columns for B (%d) and "
+					"number of columns for result (%d) should be equal!\n",
+					B.num_cols, result.num_cols);
+		}
```
OXPHOS: I don't like this part...but I didn't come up with any better idea.

src/shogun/lib/SGMatrix.h
karlnapf: Yes definitely.
Also I am unsure about such a method... it is potentially slow, likely to allocate and copy anyways, problems with multiple references, etc. Is there a definite reason why we need it?

src/shogun/mathematics/linalg/LinalgNamespace.h
karlnapf: I guess you could simplify via just saying that the dimensions don't match and then printing the given ones plus the transpose information, then you just need one message. But this verbose version has advantages from a user perspective as well..... leaving it to you

src/shogun/lib/SGMatrix.h
OXPHOS: @karlnapf The dimensions of the result matrix are determined by whether the input matrices are transposed. Currently there is no method to set matrices' rows and cols after declaration `SGMatrix<T> mat;`.

Now I am trying to do:
```
SGMatrix<T> result;
if (transposed_A)
    result.resize(rows, cols);
if (transposed_B)
    result.resize(rows, cols);
else
    result.resize(rows, cols);

if (A.on_gpu())
     result = to_gpu(result);
matrix_prod(A, B, result, transpose_A, transpose_B);
```
Without the method, the codes will be like:
```
if (transposed_A)
    result.resize(rows, cols);
    if (A.on_gpu())
         result = to_gpu(result);
    matrix_prod(A, B, result, transpose_A, transpose_B);
if (transposed_B)
    result.resize(rows, cols);
    if (A.on_gpu())
         result = to_gpu(result);
    matrix_prod(A, B, result, transpose_A, transpose_B);
else
    result.resize(rows, cols);
    if (A.on_gpu())
         result = to_gpu(result);
     matrix_prod(A, B, result, transpose_A, transpose_B);
```

karlnapf: We dont you just allocate the matrix in different size rather than resizing it in the first bit of code?

karlnapf:
```
SGMatrix<T> result;
if (transposed_A)
    result = SGMatrix<float64_t>(rows, cols);

....
```

OXPHOS: @karlnapf Oh u r so right...my bad. Just updated

karlnapf:
Cool, so finally this can be merged :)

PR #3584, opened 2016-12-20
LinalgRefactor - Apply/MatrixProd

OXPHOS:


karlnapf:
BTW: check out `git commit -amend`

src/shogun/mathematics/linalg/LinalgBackendEigen.h
```diff
@@ -95,6 +95,15 @@ class LinalgBackendEigen : public LinalgBackendBase
 	DEFINE_FOR_ALL_PTYPE(BACKEND_GENERIC_IN_PLACE_ADD, SGMatrix)
 	#undef BACKEND_GENERIC_IN_PLACE_ADD
 
+	/** Implementation of @see LinalgBackendBase::apply */
+	#define BACKEND_GENERIC_IN_PLACE_APPLY(Type, Container) \
+	virtual void apply(SGMatrix<Type>& a, Container<Type>& b, Container<Type>& result) const \
```
karlnapf: I am unsure about the name "apply" here. Why not matmul (like tensorflow), or just dot?
@lambday you probably have thoughts on this?

src/shogun/mathematics/linalg/LinalgNamespace.h
```diff
@@ -98,6 +98,121 @@ LinalgBackendBase* infer_backend(const Container<T>& a, const Container<T>& b)
 }
 
 /**
+ * Transfers data to GPU memory. Does nothing if no GPU backend registered.
```
karlnapf: why are these changes in here?

OXPHOS: I need to call `to_gpu` in the methods below. Or I can just declare them here but something needs to be in the front.

karlnapf: Ah I get it, so you just moved them up.
MMh, maybe this can be done in a sep patch? It kinda messes what is happening here

src/shogun/mathematics/linalg/LinalgBackendEigen.h
OXPHOS: @karlnapf `matmul` is matrix * matrix but I didn't see why we can't merge `matrix * vector` and `matrix * matrix` into one method. We can throw an error if `type(B) = SGVector && transpose_B = true"

karlnapf: That's true, thanks for pointing it out.

Maybe then the name dot as in numpy? I don't like apply too much...

OXPHOS: @karlnapf sry I missed this. I don't think I can overload `dot` method for both mat*vec, and vec*vec in C++ since they have different return types.

karlnapf:
What is the state here now, I cannot access the old discussion anymore. ...

OXPHOS:
@karlnapf We were saying renaming the `mat * vec` method and you proposed `dot`. However now `dot` is `vec * vec` method, which returns scalar type `T`. I don't think it can be overloaded for `mat * vec` method. I have thought about making `mat * mat` and `mat * vec` one method such as  `matrix_prod`, but I don't think `SGVector` and `SGMatrix` are 100% interchangable.

karlnapf:
I see, then we need different names for each. There is really no way of overloading mat-vec and mat-mat? Calling it `mat_mul` ?

Maybe look out in other libs for some inspiration on consistent naming ...

OXPHOS:
@karlnapf I was being stupid..so I merged mat*vec to mat*mat by overloading `matrix_prod` method.

tests/unit/mathematics/linalg/operations/Viennacl_operations_unittest.cc
```diff
@@ -179,7 +179,121 @@ TEST(LinalgBackendViennaCL, SGMatrix_elementwise_product_in_place)
...
+	x_gpu = to_gpu(x);
+
+	A_gpu = to_gpu(A);
+	b_gpu = to_gpu(b);
+	matrix_prod(A_gpu, b_gpu, x_gpu);
+	x = from_gpu(x_gpu);
+
+	float64_t ref[] = {10, 11.5, 13, 14.5};
+
+	EXPECT_EQ(x.vlen, A.num_rows);
```
karlnapf: this needs to be ASSERT_EQUAL technically otherwise the loop below can segfault, same below. But nevermind

karlnapf:
Its good to go! Nice!

Issue #3508, opened 2016-11-02
Make use of CSignal


karlnapf:
All major outer loops in Shogun should implement CSignal. In particular those ones that implement solver calls that converge at some point.

There needs to be some thought on openmp loops as these cannot be cancelled easily ... maybe c++11 threads can be? pthreads can definitely.
However, for non-parallel loops, this is an easy task. See e.g. KNN

vigsterkr:
a graceful way to break out from openmp loop
http://www.thinkingparallel.com/2007/06/29/breaking-out-of-loops-in-openmp/

yorkerlin:
@karlnapf 
could we use tag `performance improvement` instead of `numerical optimization` ?

yorkerlin:
I will change the tag now. Please let me know if you think tag `numerical optimization` is better.

karlnapf:
This is not about improving performance. This is about outputting the state of optimization algorithms.

Issue #3555, opened 2016-11-18
Port all existing (python, libshogun) examples to meta examples


karlnapf:
#3000 is the full thing, example and cookbook page

This task is simple copy paste: Just take any of the existing python examples `examples/uncodcumented/python_modular` and create a meta example that does the same thing.

Remove the old example afterwards

This is a great task to get started with Shogun's examples, very easy.

There are a few hundred examples, so lots to do :)

souravsingh:
@karlnapf I would like to take a stab at the issue.

karlnapf:
Yeah sure, as said, there is hundreds

karlnapf:
This is still really important

grg121:
I have write a [post in my blog](https://grg121.github.io/2018/01/First-path-to-shogun.html) commenting things I've seen while creating the example I've just submitted. If someone trying to add samples can find it useful I would be delighted :)

karlnapf:
Great! :)

Keep in mind also that `make meta_examples` parses and translates all meta examples found by cmake (if you added a new one you have to rerun cmake), and that `build_cpp_meta_examples` builds the cpp versions of them, and finally, `ctest -R generated_cpp-binary-linear_support_vector_machine` runs your example. If you want more details on how it is executed, append a `-V`

karlnapf:
There are still more examples to port as of early 2019: see `examples/undocumented/python`

taichi-yagami:
@karlnapf I wish to work on making meta examples for remaining py files mentioned in examples/uncodcumented/python directory, I wanted to know that whether there are any documentation / tutorial/ blogs on how to convert those py file into and meta example, and how it works.

karlnapf:
There is a readme

taichi-yagami:
@karlnapf  Ok I will take this task of building meta example for remaining python files.

karlnapf:
We will need to port all examples in `examples/undocumented/python/*.py` to the new meta example system, using factories and base types rather than specialised types (i.e. using `Machine m = machine("LibSVM")` rather than `LibSVM m = LibSVM()`, and using `m.apply().get("labels")` rather than `m.apply_binary().get_labels()`

Hephaestus12:
> There is a readme

Where is this readme available?
Also since we are no longer writing cookbook pages (*.rst) files do we still need to comment sections on the meta example (*.sg) files?

vigsterkr:
[...]

PR #3590, opened 2017-01-10
Modify src/shogun/classifier/ to implement CSignal in outer loops

abhinavrai44:
Issue #3508

### DIFF
--- a/src/shogun/classifier/AveragedPerceptron.cpp
+++ b/src/shogun/classifier/AveragedPerceptron.cpp
@@ -11,6 +11,7 @@
 #include <shogun/labels/Labels.h>
 #include <shogun/mathematics/Math.h>
 #include <shogun/labels/BinaryLabels.h>
+#include <shogun/lib/Signal.h>
 
 using namespace shogun;
 
@@ -59,9 +60,11 @@ bool CAveragedPerceptron::train_machine(CFeatures* data)
 	for (int32_t i=0; i<num_feat; i++)
 		w[i]=1.0/num_feat;
 
+	CSignal::clear_cancel();
+
 	//loop till we either get everything classified right or reach max_iter
 
-	while (!converged && iter<max_iter)
+	while (!(CSignal::cancel_computations()) && (!converged && iter<max_iter))
 	{
 		converged=true;
 		for (int32_t i=0; i<num_vec; i++)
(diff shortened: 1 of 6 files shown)


src/shogun/classifier/AveragedPerceptron.cpp
```diff
@@ -59,9 +60,11 @@ bool CAveragedPerceptron::train_machine(CFeatures* data)
 	for (int32_t i=0; i<num_feat; i++)
 		w[i]=1.0/num_feat;
 
+	CSignal::clear_cancel();
+
 	//loop till we either get everything classified right or reach max_iter
 
-	while (!converged && iter<max_iter)
+	while (!(CSignal::cancel_computations()) && (!converged && iter<max_iter))
 	{
```
karlnapf: I think we could also add an SG_INFO message for the iteration here if you want.....but seperate PR

karlnapf:
Looks good, thanks for that. Ill merge as travis is fine
Did you try it from say python?

abhinavrai44:
I tried the available python examples and they worked fine. One more thing some of the examples in examples/undocumented/libshogun seem to be outdated. For ex. classifier_newtontest.cpp has the include statement **#include <shogun/features/Labels.h>** but now Labels.h is the labels folder and also the functions associated with labels seem to have changed.

abhinavrai44:
Also, the buildbot is giving error reports after this commit. Have I done something wrong ?

karlnapf:
Yes, these examples are partly outdated, there is a blacklist in the `CMakeList.txt` that shows the examples that are not part of the test.

BTW feel free to port any of these examples to the new meta example systen, c.f. #3555

karlnapf:
I dont think the buildbot error is your fault, but thanks for pointing it out!

src/shogun/classifier/AveragedPerceptron.cpp
abhinavrai44: Yeah sure, but what should the message contain ? Which variables need to be highlighted..

karlnapf: Maybe just the iteration number.
If there is some residual that is checked to set the `converged` flag, you can also print that

abhinavrai44: ok

PR #3593, opened 2017-01-13
Port existing BrayCurtis Distance example to meta

abhinavrai44:


karlnapf:
Cool thanks!
Let's not port *all* the distance examples please :)
There are many other methods that would be better exposed.

Also, your data version is not yet merged (I need to do that first)

karlnapf:
Just merged the data commit, restarting travis

karlnapf:
Ping me if green

karlnapf:
Merge conflicts, needs a rebase. Otherwise good

doc/cookbook/source/examples/distance/braycurtis.rst
```diff
@@ -0,0 +1,39 @@
+======================
```
karlnapf: minor: The "===" are longer than the text, which will cause warnings in the pipeline.

karlnapf:
Thanks @abhinavrai44 
Can you please send integration test data and another PR here to update the data version. Testing for the distances are way more important than having more cookbooks....

@vigsterkr I usually make sure the data is included before merging. BTW do you think we should merge more distance cookbooks? I would rather only have the code/test to avoid the cookbook being spammed with 100 distance examples?

vigsterkr:
@karlnapf i've merge the thing without checking the data part but now it's fixed... so no need to do anything now.
but yeah i fully agree we shouldn't do too much of these distance examples as they are all the same....

karlnapf:
What do you mean by "but now its fixed"?

vigsterkr:
i've updated in one fix the data commit hash in the repo :)

karlnapf:
ah cool :)

PR #3598, opened 2017-01-18
Clean Up KNN

MikeLing:


MikeLing:
Hi @lisitsyn, after I do those code refactored, I got a bunch of link error https://pastebin.mozilla.org/8966903, but I don't know which part of symbols been redefined. Could you give me some helps about it? Furthermore, could you or @karlnapf or someone else to give me some feedbacks for this pr? I'm not sure I'm on the right direction, and I will add some more detail comments in this pr.

Thank you very much :)

karlnapf:
This definitely is in the right direction!

Try sending smaller PRs, doing one thing at a time. Here you moved some code to a different file and also changed it. Avoid that, first PR is moving things to different files (without changing anything else). Second PR is then changing the code (SGVector replacement for example).
So this one can be split into multiple smaller patches.

As for your error, try to compile frequently when editing. When you add new files, it might be a good idea to delete the `build` dir and re-create the build due to obsolete cache errors...

src/shogun/metric/LMNNImpl.cpp
```diff
@@ -112,7 +112,7 @@ SGMatrix<index_t> CLMNNImpl::find_target_nn(CDenseFeatures<float64_t>* x,
 		labels_slice->set_labels(labels_vec);
 
 		CKNN* knn = new CKNN(k+1, new CEuclideanDistance(features_slice, features_slice), labels_slice);
-		SGMatrix<int32_t> target_slice = knn->nearest_neighbors();
+		SGMatrix<int32_t> target_slice = knn->get_nearest_neighbors();
```
karlnapf: renaming should be in a single patch that just does that

src/shogun/multiclass/KNN.cpp
```diff
@@ -187,179 +161,15 @@ CMulticlassLabels* CKNN::apply_multiclass(CFeatures* data)
 	CMulticlassLabels* output=new CMulticlassLabels(num_lab);
 
 	//labels of the k nearest neighbors
-	int32_t* train_lab=SG_MALLOC(int32_t, m_k);
+	SGVector<int32_t> train_lab = m_k;
```
karlnapf: These changes should be in their own patch

src/shogun/multiclass/KNNSOLVER.cpp
```diff
@@ -0,0 +1,301 @@
+/*
```
karlnapf: filename would be `KNNSolver`

src/shogun/multiclass/KNNSOLVER.h
```diff
@@ -0,0 +1,195 @@
...
+#include <shogun/io/SGIO.h>
+#include <shogun/features/Features.h>
+#include <shogun/distance/Distance.h>
+#include <shogun/machine/DistanceMachine.h>
+
+namespace shogun
+{
+class CDistanceMachine;
+
+class KNNSolver :  public CDistanceMachine
```
karlnapf: This should have a C prefix, otherwise many things in shogun will stop working

MikeLing: Oh Great! I will add this prefix, thank you :)

PR #3600, opened 2017-01-22
LinalgRefactor - gpu_ptr compatibility in SGMatrix and SGVector

OXPHOS:
- add `assert_on_gpu()` 
- add `on_gpu()` check for elementwise (static) methods

However, the `gpu_ptr` is not serialized. I assume it's users' responsibility to transfer the data back to CPU before serialization. If so, I can add warnings or errors msg to `SGMatrix` serialization methods.

### DIFF
--- a/src/shogun/lib/SGMatrix.h
+++ b/src/shogun/lib/SGMatrix.h
@@ -13,6 +13,7 @@
 #ifndef __SGMATRIX_H__
 #define __SGMATRIX_H__
 
+#include <shogun/io/SGIO.h>
 #include <shogun/lib/config.h>
 #include <shogun/lib/common.h>
 #include <shogun/lib/SGReferencedData.h>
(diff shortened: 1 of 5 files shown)


src/shogun/lib/SGMatrix.h
```diff
@@ -391,6 +392,16 @@ template<class T> class SGMatrix : public SGReferencedData
 		/** overridden to free data */
 		virtual void free_data();
 
+        private:
+		/** Assert whether the data is on GPU
+		 * and raise error if the data is on GPU
+		 */
+		void assert_on_gpu() const
```
karlnapf: I guess
`assert_on_cpu` is the right name here? :)

karlnapf:
Technically, we want everything to be serialized, needs to be sorted!
But for now we can force data to be on the CPU before serialization I guess. Have to do that before merging the branch

karlnapf:
Good to merge after name change

OXPHOS:
@karlnapf updated

PR #3601, opened 2017-01-22
LinalgRefactor - minor fix on linalg::max unittest

OXPHOS:
The old Viennacl SGMatrix max test may cause segmentation fault


karlnapf:
Thanks!
And never use constants in loop counters ;)

PR #3607, opened 2017-01-27
Made changes in Distance.cpp

piyushgoel997:
used OpenMP instead of pthread.
Please verify the changes and suggest improvements or corrections.

### DIFF
--- a/src/shogun/distance/Distance.cpp
+++ b/src/shogun/distance/Distance.cpp
@@ -24,12 +24,6 @@
 #include <string.h>
 #include <unistd.h>
 
-#include <omp.h>
-
-#ifdef HAVE_PTHREAD
-#include <pthread.h>
-#endif
-
 using namespace shogun;
 
 /** distance thread parameters */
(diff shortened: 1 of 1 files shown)


src/shogun/distance/Distance.cpp
```diff
@@ -24,6 +24,8 @@
 #include <string.h>
 #include <unistd.h>
 
+#include <omp.h>
```
vigsterkr: you dont need this....

src/shogun/distance/Distance.cpp
```diff
@@ -24,6 +24,8 @@
...
+
 #ifdef HAVE_PTHREAD
```
vigsterkr: you can remove this if you dont use pthread anymore

vigsterkr:
could you please unify the use-case of num_threads =1 and num_threads=2 or more? it should be quite easy....

src/shogun/distance/Distance.cpp
piyushgoel997: can you tell why isn't it required?

vigsterkr: because you dont use any openmp function in the code below...

src/shogun/distance/Distance.cpp
```diff
@@ -350,51 +344,32 @@ SGMatrix<T> CDistance::get_distance_matrix()
...
-			CDistance::get_distance_matrix_helper<T>(&params[t]);
-		}
 
-		SG_FREE(params);
+	
+	D_THREAD_PARAM<T>* params = SG_MALLOC(D_THREAD_PARAM<T>, num_threads);
+	int64_t step= total_num/num_threads;
+		int32_t t;
 
+	num_threads--;
```
vigsterkr: have you actually tested this? as i'm afraid this is not going to work as it was previously, see the last item of `params` is never gonna be set.... please take more time to actually see through this patch and test it first locally (run the unit tests and integration tests)...

piyushgoel997: Actually, I'm using windows so I'm unable to run the tests. I tried doing what is mentioned in the getting started guide but it doesn't work for windows. I tried the ways mentioned on the internet but couldn't get it working. The only solution I found online to do testing is to use Visual Studio, but since I don't have it installed on my laptop. So can you tell me is there a way to do it using the command line or any better way, or should I install Visual Studio. Thanks.

vigsterkr: @piyushgoel997 dont get me wrong here but you are sending code without even compiling? even long time developers of shogun would have a hard time adding new features like that, let alone a person who has never touched shogun's code.

here's a manual how to develop shogun in general: https://github.com/shogun-toolbox/docs/blob/master/DEVELOPING.md

and yes that doesn't work for windows as till 3 hours ago shogun was not supported on windows. but still... you will need linux do get started with shogun... there are many ways to get a linux running under windwos, please use google to find out how to do it.

you can develop things on windows as well but you'll need Visual Studio for that... and some other things, like cmake

Issue #3553, opened 2016-11-18
Fix meta example when ´-DENABLE_TESTING=OFF`


karlnapf:
Since the `WrappedObjectArray` class is guarded and only available when testing is turned on, the meta examples which explicitly import that class will not compile (java, cpp).

Solution:
If variable storing is turned off in meta example `generator.py` (`storeVars` or so), then the includes for `WrappedObjectArray` also have to be omitted.

Nice non trivial entrance task and crucial for any further release

karlnapf:
For now, the workaround is to compile Shogun with the cmake siwtch `-DENABLE_TESTING`

IOcodegeass:
Yes this workaround worked for cpp interface.
Just by compiling using  **cmake switch -DJavaModular=ON -DENABLE_TESTING=ON ..**

sorig:
@karlnapf close this one

PR #3611, opened 2017-01-29
Enable meta examples when ENABLE_TESTING is OFF.

geektoni:
Resolve issue #3553


examples/meta/cpp/CMakeLists.txt
```diff
@@ -23,7 +23,15 @@ FOREACH(META_EXAMPLE ${META_EXAMPLES})
     ADD_DEPENDENCIES(${GENERATED_CPP_BINARY} meta_examples shogun)
     SET_SOURCE_FILES_PROPERTIES(${GENERATED_CPP} PROPERTIES GENERATED 1)
     TARGET_LINK_LIBRARIES(${GENERATED_CPP_BINARY} shogun ${SANITIZER_LIBRARY})
-    IF(SANITIZER_FLAGS)
+
+	# Suppress unused variable warnings
+	# if we are not testing the meta_examples
+	IF (NOT ENABLE_TESTING)
```
karlnapf: yes I think I like this

examples/meta/cpp/CMakeLists.txt
```diff
@@ -45,3 +53,10 @@ ENDFOREACH()
 add_custom_target(build_cpp_meta_examples ALL
     DEPENDS ${GENERATED_CPP_EXAMPLES}
     COMMENT "Compiled generated cpp examples")
+
+IF (NOT ENABLE_TESTING)
+	MESSAGE(
```
karlnapf: maybe you can put that to the place where you set the flags rather than down here?
Also, the indentation is a bit weird

examples/meta/CMakeLists.txt
```diff
@@ -27,18 +27,25 @@ FOREACH(META_EXAMPLE ${META_EXAMPLES})
     LIST(APPEND EXAMPLE_LISTINGS ${CMAKE_CURRENT_BINARY_DIR}/r/${BASENAME}.R)
     LIST(APPEND EXAMPLE_LISTINGS ${CMAKE_CURRENT_BINARY_DIR}/lua/${BASENAME}.lua)
 
+	# Set generate.py flags
```
karlnapf: indentation not respecting the previous one, makes reviewing much harder for me btw

karlnapf:
This is not the solution that I mentioned in #3553 
Can you explain what you are doing here? Not 100% sure I get what you do ... build still passes though so it seems ok

Also, the warning thing with the flag, could you put that into another patch to separate concerns here?

geektoni:
Hi @karlnapf, 
Basically, since the ```WrappedObjectArray``` class is guarded and cannot be used when we are not testing the meta examples, I decided to prevent the inclusion of ```WrappedObjectArray``` when we are generating the meta examples.

The ```generator.py``` take a ```--store-vars``` flag that, if used, inserts into the meta example the code to include ```WrappedObjectArray```. 

The patch behaviour is:
* If ENABLE_TESTING is OFF, then we don't use the ```--store-vars``` flag and no ```WrappedObjectArray``` code is added;
* If ENABLE_TESTING is ON, then we do use the flag and the meta examples code will have the library;

I thought this was the best approach to solve this problem and I understood this was the way to proceed when I read the issue you made, but probably I didn't catch your idea, my bad :).

Secondly, I wouldn't like to split the unused variable warnings into another patch because we have unused variable warning only when we produce meta example code that has not ```WrappedObjectArray```. If I do that we will have two patches that rely one on the other, and if this is a wrong way to fix the issue the unused variable warning patch will be of little use.

However, if you prefer to have two separate patch I will gladly create another one. :) 

By the way, I moved the unused variable warning and its message close together (since it is more logical as you pointed out) and I corrected the indentation errors.

karlnapf:
Thanks for the good explanation :) I agree with the variable stuff. I should have realised this but I think I looked at this on my phone...

Very nice work! Merging!

Issue #3557, opened 2016-11-18
`make build_csharp_meta_examples` always re-compiles all java examples


karlnapf:
Same as for #3556

vigsterkr:
closing since we got this with the help of @geektoni

PR #3610, opened 2017-01-27
Java meta examples only re-compile when dependencies change.

geektoni:
Addressing issue #3556.

PR #3613, opened 2017-01-31
CSharp meta examples only re-compile when dependencies change.

geektoni:
Resolve issue #3557


karlnapf:
Same question as for java here.

1. typing `build_csharp_meta_examples` twice only compiles them the first time
2. modifying the source code of a meta example followed by the above make only re-compiles the modifed one?

Travis error is unrelated

geektoni:
Same behaviour of #3610 now

PR #3618, opened 2017-02-07
LinalgRefactor - rm Apply and ElementwiseProduct

OXPHOS:


src/shogun/lib/GPUMatrix.h
```diff
@@ -41,6 +41,7 @@
 
 #include <shogun/lib/common.h>
 #include <memory>
+#include <viennacl/linalg/matrix_operations.hpp>
```
OXPHOS: This was in `Apply.h` and it is required for GPU matrix element access.
NOTE: current new GPUMatrix/Vector doesn't support element access. Maybe we can add it back.

karlnapf: Nope, no element access for GPU structures (forces user to transfer back to cpu first)

karlnapf: Nope, no element access for GPU structures (forces user to transfer back to cpu first)

karlnapf: The point is that element-wise product on the GPU is done in vectorised form

PR #3621, opened 2017-02-10
Port CombinedKernel, Distance, DotFeatures and LibSVMKernel to OpenMP

micmn:


src/shogun/distance/Distance.cpp
```diff
@@ -26,37 +26,10 @@
 #include <unistd.h>
 #endif
 
-#ifdef HAVE_PTHREAD
-#include <pthread.h>
-#endif
+#include <omp.h>
```
vigsterkr: guard it with `HAVE_OPENMP`

vigsterkr: but you dont need it because for finding out num available threads you should call `Parallel` singleton.

src/shogun/distance/Distance.cpp
```diff
@@ -341,93 +264,64 @@ SGMatrix<T> CDistance::get_distance_matrix()
...
+	int64_t step;
+	#pragma omp parallel shared(num_threads, step)
 	{
-#ifdef HAVE_PTHREAD
-		pthread_t* threads = SG_MALLOC(pthread_t, num_threads-1);
-		D_THREAD_PARAM<T>* params = SG_MALLOC(D_THREAD_PARAM<T>, num_threads);
-		int64_t step= total_num/num_threads;
+		#pragma omp single
+		{
+			num_threads = omp_get_num_threads();
```
vigsterkr: ```parallel->get_num_threads();``` instead of omp function...

src/shogun/distance/Distance.cpp
```diff
@@ -341,93 +264,64 @@ SGMatrix<T> CDistance::get_distance_matrix()
...
+			step = total_num/num_threads;
+			num_threads--;
+		}
 
-		int32_t t;
+		int32_t thread_num = omp_get_thread_num();
```
vigsterkr: same as above.

vigsterkr: what happens when you dont have openmp backend?
how will this function work at all when `thread_num = 0`, as you only have one thread...

src/shogun/distance/Distance.cpp
```diff
@@ -341,93 +264,64 @@ SGMatrix<T> CDistance::get_distance_matrix()
...
+		bool verbose = (thread_num == 0);
 
-		num_threads--;
-		for (t=0; t<num_threads; t++)
+		int32_t start = compute_row_start(thread_num*step, n, symmetric);
+		int32_t end = (thread_num==num_threads) ? m : compute_row_start((thread_num+1)*step, n, symmetric);
+
+		for (int32_t i=start; i<end; i++)
```
vigsterkr: these nested loops are a bit scary...

src/shogun/features/DotFeatures.cpp
```diff
@@ -77,75 +59,40 @@ void CDotFeatures::dense_dot_range(float64_t* output, int32_t start, int32_t sto
...
-		int32_t step= num_vectors/num_threads;
+		#pragma omp single
+		{
+			num_threads = omp_get_num_threads();
+			step = num_vectors/num_threads;
+			num_threads--;
+		}
 
-		int32_t t;
+		int32_t thread_num = omp_get_thread_num();
```
vigsterkr: same story as above....

src/shogun/features/DotFeatures.cpp
```diff
@@ -158,133 +105,45 @@ void CDotFeatures::dense_dot_range_subset(int32_t* sub_index, int32_t num, float
...
-		dense_dot_range_helper((void*) &params[t]);
-
-		for (t=0; t<num_threads-1; t++)
-			pthread_join(threads[t], NULL);
-
-		SG_FREE(params);
-		SG_FREE(threads);
-	}
-#endif
+		int32_t thread_num = omp_get_thread_num();
```
vigsterkr: same again

src/shogun/distance/Distance.cpp
vigsterkr: mmm ok my bad as then `step = total_num`, right?
if so just guard the code with `HAVE_OPENMP`, something like this:
```
#ifdef HAVE_OPENMP
int32_t thread_num = omp_get_thread_num();
#else
int32_t thread_num = 0;
#endif
```

micmn: so I'll leave (guarded) `#include omp.h`; is it possible that `parallel->get_num_threads()` != 1 and openmp isn't available? in that case if I replace `omp_get_num_threads()` with that the step size would be wrong, do I need to consider that case?

src/shogun/distance/Distance.cpp
micmn: they were like that in `get_distance_matrix_helper` https://github.com/shogun-toolbox/shogun/pull/3621/files/4d1f092534b2d9b1ba730ee19061994265d8ac2c#diff-7542e21aede27fe5307ee5008c99fa45L298

src/shogun/distance/Distance.cpp
vigsterkr: yes it... `parallel->get_num_threads()` just give the detected number of cores or what the user set it to.

src/shogun/distance/Distance.cpp
vigsterkr: yeah it was scary in the first place... i mean if there's no more elegant way to do it, of course leave it.

src/shogun/distance/Distance.cpp
micmn: so isn't better using the actual number of threads spawned and in case of no openmp setting `num_threads=1`?

vigsterkr: yeah you can do that... again that then requires some macro guarding.

vigsterkr: fyi: most of the travis jobs failed because that version of clang has no openmp:
https://travis-ci.org/shogun-toolbox/shogun/jobs/200335452#L1986

vigsterkr:
cool, if all the CI (except the octave modular) passes we can merge it.

src/shogun/distance/Distance.cpp
```diff
@@ -341,93 +266,69 @@ SGMatrix<T> CDistance::get_distance_matrix()
...
-#ifdef HAVE_PTHREAD
-	int32_t num_threads=parallel->get_num_threads();
+	int32_t num_threads;
+	int64_t step;
+	#pragma omp parallel shared(num_threads, step)
+	{
+#ifdef HAVE_OPENMP
+		#pragma opm single
+		{
+			num_threads=omp_get_num_threads();
```
karlnapf: @vigsterkr I am confused here, wouldn't we use `Parallel` to get the number of threads?

vigsterkr: @karlnapf no because if you check Parallel it is active when either openmp or threading backend is available ;) and we want to activate this when openmp is available.

PR #3597, opened 2017-01-17
Add Scala Interface

abhinavrai44:
Issue #3485

PR #3623, opened 2017-02-11
Add Scala to .travis.yml

abhinavrai44:


karlnapf:
Fine from my side
@vigsterkr you go ahead

vigsterkr:
yeah but isn't this part of #3597 already? if so it's fine there... let's just have that rebased that way everything gets tested there in that pr and close this one....

PR #3635, opened 2017-02-17
LinalgRefactor - Memory Transfer Mutex

OXPHOS:


OXPHOS:
Currently the atomic variable is only added to `SGVector` and mutex only realized in `to_gpu` method

OXPHOS:
I confused myself a little bit here - I am glad to move the discussion to elsewhere if necessary.
- Take the `to_gpu` method for example, the way we transfer a CPU vector to GPU is to create a new `SGVector` instance with a pointer to GPU. 
- And most of the time, we expect users to declare a new `SGVector` for the transferred vector, such as `auto a_gpu = to_gpu(a)`. In this scenario, mutex is not required for the memory transfer method
- The only scenario that requires the mutex is when users do `a = to_gpu(a)` (#3636). In this case, `a` will be assigned to the newly allocated GPU `SGVector` instance. Thus, the 2nd thread trying to execute `a = to_gpu(a)` should be held and be notified `a.on_gpu() = true` after the 1st transfer is done.
- Here comes the first problem.  `a = to_gpu(a)`  destroys the original CPU vector. What if another thread is performing operations such as `c = a + b`? Should memory transfer operation suspend all the other operations? Is it users' responsibility to prevent this error?
- Moreover, should we consider the multi-thread safety on operations such as in-place addition, `a = a+b`, or on the different outcomes when operating `c = a + b` and `b = a * c` in the different order? It seems we need to ensure the data structure itself is mutex on reading/writing. Or should users guarantee the conflicts won't happen? If so, why will there be two threads calling `to_gpu(a)` at the same time?
Hope I made myself clear here.. @karlnapf @vigsterkr

karlnapf:
Answer to your third point:
`a=to_gpu(a)` will *only* destroy the CPU vector if there is not other reference to it. If there is another thread working on it, it will have a reference which increased the reference counter and therefore the old SGVector will not be destroyed. In fact,  the memory will be intact and the thread will continue to work on the CPU version. I think .... @vigsterkr ?

src/shogun/mathematics/linalg/LinalgNamespace.h
```diff
@@ -30,15 +30,21 @@
...
 #include <shogun/mathematics/linalg/SGLinalg.h>
+#include <shogun/lib/Lock.h>
 
 namespace shogun
 {
 
 namespace linalg
 {
 
+static CLock m_gpu_transfer;
```
karlnapf: is it a problem that `CLock` only has PTHREAD locking stuff?
We might need to add openmp, or is openmp locking implied by pthread?
http://stackoverflow.com/questions/2396430/how-to-use-lock-in-openmp

While openmp implementations are likely to be based on pthread, the specification does not talk about this --- I think we should better be safe than sorry
http://stackoverflow.com/questions/5661830/using-lock-in-openmp-and-pthreads

vigsterkr: @karlnapf why do you say that `CLock`? i mean yeah in this branch which is not rebased for ages it is still that but latest develop branch has better backend for locking

vigsterkr:
@OXPHOS 
`Here comes the first problem. a = to_gpu(a) destroys the original CPU vector`
is this somewhere explicit or implicit? if the former, then could you please point to the lines of code?

vigsterkr:
@OXPHOS yes it is true that in most of the cases our data structures were not thread safe at all... like sgvector and sgmatrix and it was the user's responsibility to make it like that. but in case of that `on_gpu` i felt that that should be really atomic instead of an `if(pointer != null)`.

OXPHOS:
I figured out the `a = to_gpu(a)` "destroying" the old vector stuff. It was simply a ref number decrease as [here](https://github.com/shogun-toolbox/shogun/pull/3635/files#diff-c02a2e0b28912409b0450601383094e4R113).

src/shogun/mathematics/linalg/LinalgNamespace.h
```diff
@@ -106,17 +112,35 @@ LinalgBackendBase* infer_backend(const Container<T>& a, const Container<T>& b)
  * and a shallow-copy of SGVector with vector on CPU if GPU backend not available
  */
 template <typename T>
-SGVector<T> to_gpu(const SGVector<T>& vector)
+void to_gpu(SGVector<T>& a, SGVector<T>& b)
 {
```
OXPHOS: @karlnapf @vigsterkr Regardless of the technical details, I am reshaping `a = to_gpu(a)` to `to_gpu(a, a)` and/or `to_gpu(a)` as the copy assignment `a = xxx` is not atomic.

src/shogun/mathematics/linalg/LinalgNamespace.h
```diff
@@ -114,20 +109,26 @@ LinalgBackendBase* infer_backend(const Container<T>& a, const Container<T>& b)
...
 	{
-		SG_SWARNING("Trying to access GPU memory without GPU backend registered.\n");
-		b = a;
+		LinalgBackendBase* gpu_backend = sg_linalg->get_gpu_backend();
+
+		if (gpu_backend)
+			b = SGVector<T>(gpu_backend->to_gpu(a), a.vlen);
+		else
+		{
+			SG_SWARNING("Trying to access GPU memory without GPU backend registered.\n");
```
karlnapf: I know this was there before but I just reaslied this will cause tons of warnings for gpu code being executed without a gpu present. I wonder whether that is good?

tests/unit/mathematics/linalg/LinalgBackendViennaCL_unittest.cc
```diff
@@ -2,13 +2,26 @@
 #include <shogun/lib/SGVector.h>
 #include <shogun/mathematics/linalg/LinalgNamespace.h>
 #include <gtest/gtest.h>
+#include <pthread.h>
```
karlnapf: Please use C++11 here.

tests/unit/mathematics/linalg/LinalgBackendViennaCL_unittest.cc
```diff
@@ -130,4 +143,35 @@ TEST(LinalgBackendViennaCL, SGMatrix_clone)
...
+	SGVector<int32_t> a(size);
+	a.range_fill(0);
+
+	const int32_t count = 100;
+	pthread_t * threads = new pthread_t[count];
+
+	for (index_t i = 0; i < count; i++)
+	{
+		if (i % 2)
+			pthread_create(threads+i, NULL, memory_transfer_test_helper_1,
```
karlnapf: std::thread

OXPHOS:
@karlnapf @vigsterkr This is a huge piece..I did the stress test on my machine with up to 100 threads. Too bad it doesn't work on Travis.

tests/unit/mathematics/linalg/LinalgBackendViennaCL_unittest.cc
```diff
@@ -109,25 +122,52 @@ TEST(LinalgBackendViennaCL, SGMatrix_clone)
...
+{
+	sg_linalg->set_gpu_backend(new LinalgBackendViennaCL());
+
+	const index_t size = 10;
+	SGVector<int32_t> a(size);
+	a.range_fill(0);
+
+	const int32_t count = 5;
+	std::thread threads[count];
+
```
OXPHOS: delete[]

src/shogun/mathematics/linalg/LinalgNamespace.h
karlnapf: @vigsterkr I just checked a bit, and most of shogun is not thread safe at all. So I wonder whether we really should care about this too much at this point in time. I mean we should definitely have a go on making operations thread safe (along with proper const-ing), but for now I feel that, yes, it is user responsibility. Even for GPU transfers.
I mean we can still keep this patch here, but I think there should be some more systematic effort to make shogun base lib thread safe.

karlnapf:
Seems good to merge from my side. But let's leave this to @vigsterkr

src/shogun/mathematics/linalg/LinalgNamespace.h
vigsterkr: @karlnapf wtf you mean it's not thread safe at all?

vigsterkr: it's one thing that datastructures are not thread safe... but this is something that has to otherwise things gonna be really fuckedup...

karlnapf: Yes data structures.

vigsterkr: yeah so think about to_gpu being totally not threadsafe... that is something that you dont want to have on user code side... what you will mutex it constantly around the checking for on_gpu?

karlnapf: True.

src/shogun/lib/SGMatrix.cpp
```diff
@@ -39,13 +39,25 @@ SGMatrix<T>::SGMatrix(bool ref_counting) : SGReferencedData(ref_counting)
 template <class T>
 SGMatrix<T>::SGMatrix(T* m, index_t nrows, index_t ncols, bool ref_counting)
 	: SGReferencedData(ref_counting), matrix(m),
-	num_rows(nrows), num_cols(ncols), gpu_ptr(nullptr) { }
+	num_rows(nrows), num_cols(ncols), gpu_ptr(nullptr)
+{
+	m_on_gpu.store(false);
```
vigsterkr:
```
m_on_gpu.store(false, std::memory_order_release);
```

vigsterkr:
```
m_on_gpu.store(false, std::memory_order_release);
```

src/shogun/lib/SGMatrix.cpp
```diff
@@ -39,13 +39,25 @@ SGMatrix<T>::SGMatrix(bool ref_counting) : SGReferencedData(ref_counting)
...
+}
+
+template <class T>
+SGMatrix<T>::SGMatrix(T* m, index_t nrows, index_t ncols, index_t offset)
+	: SGReferencedData(false), matrix(m+offset),
+	num_rows(nrows), num_cols(ncols)
+{
+	m_on_gpu.store(false);
```
vigsterkr: same as above

src/shogun/lib/SGMatrix.cpp
```diff
@@ -39,13 +39,25 @@ SGMatrix<T>::SGMatrix(bool ref_counting) : SGReferencedData(ref_counting)
...
+}
 
 template <class T>
 SGMatrix<T>::SGMatrix(index_t nrows, index_t ncols, bool ref_counting)
 	: SGReferencedData(ref_counting), num_rows(nrows), num_cols(ncols), gpu_ptr(nullptr)
 {
 	matrix=SG_MALLOC(T, ((int64_t) nrows)*ncols);
+	m_on_gpu.store(false);
```
vigsterkr: same as above

src/shogun/lib/SGMatrix.cpp
```diff
@@ -56,6 +68,7 @@ SGMatrix<T>::SGMatrix(SGVector<T> vec) : SGReferencedData(vec)
 	num_rows=vec.vlen;
 	num_cols=1;
 	gpu_ptr = vec.gpu_ptr;
+	m_on_gpu.store(vec.on_gpu());
```
vigsterkr: same as above

src/shogun/lib/SGMatrix.cpp
```diff
@@ -72,13 +85,15 @@ SGMatrix<T>::SGMatrix(SGVector<T> vec, index_t nrows, index_t ncols)
 	num_rows=nrows;
 	num_cols=ncols;
 	gpu_ptr = vec.gpu_ptr;
+	m_on_gpu.store(vec.on_gpu());
```
vigsterkr: same as above

src/shogun/lib/SGMatrix.cpp
```diff
@@ -72,13 +85,15 @@ SGMatrix<T>::SGMatrix(SGVector<T> vec, index_t nrows, index_t ncols)
...
 }
 
 template<class T>
 SGMatrix<T>::SGMatrix(GPUMemoryBase<T>* mat, index_t nrows, index_t ncols)
 	: SGReferencedData(true), matrix(NULL), num_rows(nrows), num_cols(ncols),
 	gpu_ptr(std::shared_ptr<GPUMemoryBase<T>>(mat))
 {
+	m_on_gpu.store(true);
```
vigsterkr: same as above

src/shogun/lib/SGMatrix.cpp
```diff
@@ -92,7 +107,7 @@ SGMatrix<T>::SGMatrix(EigenMatrixXt& mat)
 : SGReferencedData(false), matrix(mat.data()),
 	num_rows(mat.rows()), num_cols(mat.cols()), gpu_ptr(nullptr)
 {
-
+	m_on_gpu.store(false);
```
vigsterkr: again make sure about the order

src/shogun/lib/SGMatrix.cpp
```diff
@@ -1066,6 +1094,7 @@ void SGMatrix<T>::copy_data(const SGReferencedData &orig)
 	matrix=((SGMatrix*)(&orig))->matrix;
 	num_rows=((SGMatrix*)(&orig))->num_rows;
 	num_cols=((SGMatrix*)(&orig))->num_cols;
+	m_on_gpu.store(((SGMatrix*)(&orig))->m_on_gpu.load());
```
vigsterkr:
```
m_on_gpu.store(((SGMatrix*)(&orig))->m_on_gpu.load(std::memory_order_aquire), std::memory_order_release);
```

src/shogun/mathematics/linalg/LinalgNamespace.h
```diff
@@ -102,118 +102,162 @@ LinalgBackendBase* infer_backend(const Container<T>& a, const Container<T>& b)
...
  */
 template <typename T>
-SGVector<T> to_gpu(const SGVector<T>& vector)
+void to_gpu(SGVector<T>& a, SGVector<T>& b)
 {
-	REQUIRE(!vector.on_gpu(), "The vector is already on GPU.\n");
-	LinalgBackendBase* gpu_backend = sg_linalg->get_gpu_backend();
-	if (gpu_backend)
-		return SGVector<T>(gpu_backend->to_gpu(vector), vector.vlen);
+	sg_linalg->m_gpu_transfer.lock();
```
vigsterkr: move the lock part into the else branch...

src/shogun/mathematics/linalg/LinalgNamespace.h
```diff
@@ -102,118 +102,162 @@ LinalgBackendBase* infer_backend(const Container<T>& a, const Container<T>& b)
...
  */
 template <typename T>
-SGMatrix<T> to_gpu(const SGMatrix<T>& mat)
+void to_gpu(SGMatrix<T>& a, SGMatrix<T>& b)
 {
-	REQUIRE(!mat.on_gpu(), "The matrix is already on GPU.\n");
-	LinalgBackendBase* gpu_backend = sg_linalg->get_gpu_backend();
-	if (gpu_backend)
-		return SGMatrix<T>(gpu_backend->to_gpu(mat), mat.num_rows, mat.num_cols);
+	sg_linalg->m_gpu_transfer.lock();
```
vigsterkr: move the lock into the else branch

src/shogun/mathematics/linalg/LinalgNamespace.h
OXPHOS: I don't think the lock should be in the else branch. For example, assuming `a` is a GPU vector, thread1 is transferring gpu vector `a` from GPU back to CPU, and during which time `a.on_gpu() == true`. If thread2 calls `to_gpu()` when thread1 is working, without the lock in the beginning, the `if` clause will be `true` and nothing will be done. Thus, after thread1 and thread2 (`from_gpu()` and `to_gpu()`), `a` will end up in CPU. However with the lock, `a` will be transferred to CPU and moved back to GPU.

vigsterkr: indeed you are right!

vigsterkr:
@OXPHOS great, now you can merge yourself ;)

OXPHOS:
thanks : )

karlnapf:
So do it! :)

PR #3637, opened 2017-02-17
LinalgRefactor - remove linalg::set_rows_const

OXPHOS:


karlnapf:
This is not used anywhere right? :) So good to go

PR #3638, opened 2017-02-17
Fix `make unit-tests` and replace mktemp inside unit tests with a custom method

geektoni:
* ```make unit-tests``` was pointing to a wrong location of ```shogun-unit-test```
* Replaced mktemp with mktemp_cst to avoid many compiler warnings (see https://travis-ci.org/shogun-toolbox/shogun/jobs/202296293), since mktemp is not safe to use (and there is no way to suppress them, apparently).

### DIFF
--- a/tests/unit/utils/Utils.h
+++ b/tests/unit/utils/Utils.h
@@ -1,0 +1,49 @@
+/*
+* BSD 3-Clause License
+*
+* Copyright (c) 2017, Shogun-Toolbox e.V. <shogun-team@shogun-toolbox.org>
+* All rights reserved.
+*
+* Redistribution and use in source and binary forms, with or without
+* modification, are permitted provided that the following conditions are met:
+*
+* * Redistributions of source code must retain the above copyright notice, this
+*   list of conditions and the following disclaimer.
+*
+* * Redistributions in binary form must reproduce the above copyright notice,
+*   this list of conditions and the following disclaimer in the documentation
+*   and/or other materials provided with the distribution.
+*
+* * Neither the name of the copyright holder nor the names of its
+*   contributors may be used to endorse or promote products derived from
+*   this software without specific prior written permission.
+*
+* THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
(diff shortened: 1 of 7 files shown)


tests/unit/utils/Utils.h
```diff
@@ -0,0 +1,15 @@
+#ifndef __UTILS_H__
+#define __UTILS_H__
+
+/*
```
karlnapf: These files need BSD headers (unless they are GPL, in which case we cannot use them)

geektoni: I've added the BSD headers. Not sure if they are the right ones though. Tell me if you prefer a different license version.

karlnapf: This one is good. So you wrote this thing? Just wondering whether we should put credits for someone else?

geektoni: Yes, I wrote it. However, I need to admit that I took inspiration from some C++ guide (like this one http://www.cplusplus.com/reference/cstring/strstr/) and from some StackOverflow answers (for the lambda thing).
Should I insert some links as acknowledgement?

karlnapf: No that is fine, I was just checking on whether it was copied from somewhere.

Ok last question: there is no standard c++ function to do this thing without warning?

geektoni: The most suitable methods that I've found are ```mkstemp```, ```mkdtemp``` and ```tmpnam```. Unfortunately, none of them can be used because:
* ```mkstemp``` creates a new random file and return a file descriptor to it. It is safe, but some unit tests need only to have the filename to open a new file;
* ```mkdtemp``` creates a random directory and it is not what we need;
* ```tempnam``` does create a random filename, but it suffers from the same problems of ```mktemp``` (http://en.cppreference.com/w/cpp/io/c/tmpnam);

Although some of the classes used inside the unit tests (CSerializableAsciiFile, CSerializableXMLFile and CCSVFile) permits to create a file given its FD, the CStreamingAsciiFile class has not this ability. So, I opted to simply write this small function to resolve the warnings, without touching too much the unit tests code.

karlnapf: Cool thanks for the explain

PR #3641, opened 2017-02-18
add tests for KNN and fix an error in KDTree solver

MikeLing:
I found unit test for kdtree actually is using `KNN_BRUTE` as solver type https://github.com/shogun-toolbox/shogun/blob/develop/tests/unit/multiclass/KNN_unittest.cc#L94 and https://github.com/shogun-toolbox/shogun/blob/develop/tests/unit/multiclass/KNN_unittest.cc#L186. And it seems like we missing unit test for covertree solver. 

When I working on the test for kdtree, I found a minor error for the `classify_for_multiple_k` of kdtree. So this pr is for these things I mentioned above. Please take a look at this. Thank you :)

### DIFF
--- a/src/shogun/multiclass/KDTreeKNNsolver.cpp
+++ b/src/shogun/multiclass/KDTreeKNNsolver.cpp
@@ -40,7 +40,7 @@ CMulticlassLabels* CKDTREEKNNSolver::classify_objects(CDistance* distance, const
 		output->set_label(i, out_idx + m_min_label);
 	}
 	SG_UNREF(query);
-
+	SG_UNREF(kd_tree);
 	return output;
 }
 
@@ -49,7 +49,7 @@ int32_t* CKDTREEKNNSolver::classify_objects_k(CDistance* distance, const int32_t
 	int32_t* output=SG_MALLOC(int32_t, m_k*num_lab);
 
 	//allocation for distances to nearest neighbors
-	float64_t* dists=SG_MALLOC(float64_t, m_k);
+	SGVector<float64_t> dists(m_k);
 
 	CFeatures* lhs = distance->get_lhs();
 	CKDTree* kd_tree = new CKDTree(m_leaf_size);
@@ -65,12 +65,15 @@ int32_t* CKDTREEKNNSolver::classify_objects_k(CDistance* distance, const int32_t
 		for (index_t j=0; j<m_k; j++)
 		{
 			train_lab[j] = m_train_labels[ NN(j,i) ];
-			dists[j] = distance->distance(i, NN(j,i));
+			dists[j] = distance->distance(NN(j,i), i);
 		}
-		CMath::qsort_index(dists, train_lab, m_k);
+		CMath::qsort_index(dists.vector, train_lab, m_k);
 
 		choose_class_for_multiple_k(output+i, classes, train_lab, num_lab);
 	}
 
+	SG_UNREF(data);
+	SG_UNREF(kd_tree);
+	
 	return output;
 }
--- a/tests/unit/multiclass/KNN_unittest.cc
+++ b/tests/unit/multiclass/KNN_unittest.cc
@@ -91,7 +91,7 @@ TEST(KNN, kdtree_solver)
 
 	int32_t k=4;
 	CEuclideanDistance* distance = new CEuclideanDistance();	
-	CKNN* knn=new CKNN (k, distance, labels, KNN_BRUTE);
+	CKNN* knn=new CKNN (k, distance, labels, KNN_KDTREE);
 	SG_REF(knn);
 
 	features->add_subset(train);
@@ -159,6 +159,53 @@ TEST(KNN, lsh_solver)
 	SG_UNREF(knn);
 }
 
+TEST(KNN, classify_multiple_brute)
+{
+	int32_t num = 50;
+	int32_t feats = 2;
+	int32_t classes = 3;
+
+	SGVector< float64_t > lab(classes*num);
+	SGMatrix< float64_t > feat(feats, classes*num);
+
+	generate_knn_data(feat, lab, num, classes, feats);
+	SGVector<index_t> train (int32_t(num*classes*0.75));
+	SGVector<index_t> test (int32_t(num*classes*0.25));
+	train.random(0, classes*num-1);
+	test.random(0, classes*num-1);
+
+	CMulticlassLabels* labels = new CMulticlassLabels(lab);
+	CDenseFeatures< float64_t >* features = new CDenseFeatures< float64_t >(feat);
+	CFeatures* features_test = (CFeatures*) features->clone();  
+	CLabels* labels_test = (CLabels*) labels->clone();
+
+	int32_t k=4;
+	CEuclideanDistance* distance = new CEuclideanDistance();                
+	CKNN* knn=new CKNN (k, distance, labels, KNN_BRUTE);
+	SG_REF(knn);
+
+	features->add_subset(train);
+	labels->add_subset(train);
+	knn->train(features);
+
+	// classify for multiple k
+	features_test->add_subset(test);
+	labels_test->add_subset(test);
+
+	CEuclideanDistance* dist = new CEuclideanDistance(features, ((CDotFeatures*)features_test));  
+	knn->set_distance(dist);
+	SGMatrix<int32_t> out_mat =knn->classify_for_multiple_k();
+	features_test->remove_subset();
+
+	for ( index_t i = 0; i < labels_test->get_num_labels(); ++i )
+		for ( index_t j = 0; j < k; ++j )
+			EXPECT_EQ(out_mat(i, j), ((CMulticlassLabels*)labels_test)->get_label(i));
+
+	SG_UNREF(knn);
+	SG_UNREF(features_test);
+	SG_UNREF(labels_test);
+}
+
 
 TEST(KNN, classify_multiple_kdtree)
 {
@@ -169,7 +216,7 @@ TEST(KNN, classify_multiple_kdtree)
 
 	SGVector< float64_t > lab(classes*num);
 	SGMatrix< float64_t > feat(feats, classes*num);
-	
+
 	generate_knn_data(feat, lab, num, classes, feats);
 	SGVector<index_t> train (int32_t(num*classes*0.75));
 	SGVector<index_t> test (int32_t(num*classes*0.25));
@@ -183,7 +230,7 @@ TEST(KNN, classify_multiple_kdtree)
 
 	int32_t k=4;
 	CEuclideanDistance* distance = new CEuclideanDistance();	
-	CKNN* knn=new CKNN (k, distance, labels, KNN_BRUTE);
+	CKNN* knn=new CKNN (k, distance, labels, KNN_KDTREE);
 	SG_REF(knn);
 
 	features->add_subset(train);
@@ -193,7 +240,7 @@ TEST(KNN, classify_multiple_kdtree)
 	// classify for multiple k
 	features_test->add_subset(test);
 	labels_test->add_subset(test);
-	CEuclideanDistance* dist = new CEuclideanDistance(features, ((CDotFeatures*)features_test));	
+	CEuclideanDistance* dist = new CEuclideanDistance(features, ((CDotFeatures*)features_test));
 	knn->set_distance(dist);
 	SGMatrix<int32_t> out_mat =knn->classify_for_multiple_k();
 	features_test->remove_subset();


src/shogun/multiclass/KDTreeKNNsolver.cpp
```diff
@@ -65,7 +65,7 @@ int32_t* CKDTREEKNNSolver::classify_objects_k(CDistance* distance, const int32_t
 		for (index_t j=0; j<m_k; j++)
 		{
 			train_lab[j] = m_train_labels[ NN(j,i) ];
-			dists[j] = distance->distance(i, NN(j,i));
+			dists[j] = distance->distance(NN(j,i), i);
```
karlnapf: why this?
distances are symmetric

tests/unit/multiclass/KNN_unittest.cc
```diff
@@ -91,7 +91,7 @@ TEST(KNN, kdtree_solver)
 
 	int32_t k=4;
 	CEuclideanDistance* distance = new CEuclideanDistance();	
-	CKNN* knn=new CKNN (k, distance, labels, KNN_BRUTE);
+	CKNN* knn=new CKNN (k, distance, labels, KNN_KDTREE);
```
karlnapf: good catch!

tests/unit/multiclass/KNN_unittest.cc
```diff
@@ -113,6 +113,50 @@ TEST(KNN, kdtree_solver)
 	SG_UNREF(labels_test);
 }
 
+TEST(KNN, covertree_solver)
+{
+
+	int32_t num = 50;
+	int32_t feats = 2;
+	int32_t classes = 3;
+
```
karlnapf: I suggest rather than copy pasting all this code, you add a `set_up` method that prepares all the data and instances, and then each unit test just chooses the solver type and checks output

tests/unit/multiclass/KNN_unittest.cc
```diff
@@ -206,5 +250,98 @@ TEST(KNN, classify_multiple_kdtree)
...
+	SGVector< float64_t > lab(classes*num);
+	SGMatrix< float64_t > feat(feats, classes*num);
+	
+	generate_knn_data(feat, lab, num, classes, feats);
```
karlnapf: same here

src/shogun/multiclass/KDTreeKNNsolver.cpp
MikeLing: It actually will get an out boundary error[1] if we don't switch it
 [1]**In file /Users/mikeling/shogun/src/shogun/distance/Distance.cpp line 188: idx_a (0) must be in [0,111] and idx_b (103) must be in [0,36]
unknown file: Failure**

tests/unit/multiclass/KNN_unittest.cc
MikeLing: Sure thing!

MikeLing: How about create a class named `KNNTestData` and put all mock data in here?

MikeLing:
@karlnapf ask review for this pr. Thank you :)

src/shogun/multiclass/KDTreeKNNsolver.cpp
karlnapf: I see, thanks!

tests/unit/multiclass/KNN_unittest.cc
```diff
@@ -9,52 +9,98 @@
...
-
-	//generate random subset for train and test data
-	train.random(0, classes*num-1);
-	test.random(0, classes*num-1);
-
-	CMulticlassLabels* labels = new CMulticlassLabels(lab);
-	
-	CDenseFeatures< float64_t >* features = new CDenseFeatures< float64_t >(feat);
-	CFeatures* features_test = (CFeatures*) features->clone();	
+	KNNTestData* mockdata = new KNNTestData();
```
karlnapf: Either put it on stack, or clear up memory afterwards

tests/unit/multiclass/KNN_unittest.cc
```diff
@@ -9,52 +9,98 @@
...
+			//generate random subset for train and test data
+			train.random(0, classes*num-1);
+			test.random(0, classes*num-1);
+
+			generate_knn_data(feat, lab, num, classes, feats);
+		}
+
+		~KNNTestData() {};
+
+		SGVector<float64_t> return_lab() const
```
karlnapf: get_labels

tests/unit/multiclass/KNN_unittest.cc
```diff
@@ -9,52 +9,98 @@
...
+		{
+			return this->lab;
+		}
+
+		SGMatrix<float64_t> return_feat() const
```
karlnapf: get_feature_matrix

tests/unit/multiclass/KNN_unittest.cc
```diff
@@ -9,52 +9,98 @@
...
+		int32_t feats;
+		int32_t classes;
+		SGVector< float64_t> lab;
+		SGMatrix< float64_t> feat;
+		SGVector<index_t> train;
+		SGVector<index_t> test;
+};
+
+void KNNTestData::generate_knn_data(SGMatrix<float64_t>& feature, SGVector<float64_t>& labels,
+					int32_t number, int32_t clas, int32_t features)
```
karlnapf: class

tests/unit/multiclass/KNN_unittest.cc
```diff
@@ -9,52 +9,98 @@
...
+			feat = SGMatrix< float64_t >(feats, classes*num);
+
+			train = SGVector<index_t> (int32_t(num*classes*0.75));
+			test = SGVector<index_t> (int32_t(num*classes*0.25));
+
+			//generate random subset for train and test data
+			train.random(0, classes*num-1);
+			test.random(0, classes*num-1);
+
+			generate_knn_data(feat, lab, num, classes, feats);
```
karlnapf: since you call this method in only one place, why not just paste it in here?

tests/unit/multiclass/KNN_unittest.cc
```diff
@@ -206,5 +243,70 @@ TEST(KNN, classify_multiple_kdtree)
...
+
+	for ( index_t i = 0; i < labels_test->get_num_labels(); ++i )
+		for ( index_t j = 0; j < k; ++j )
+			EXPECT_EQ(out_mat(i, j), ((CMulticlassLabels*)labels_test)->get_label(i));
+
+	SG_UNREF(knn);
+	SG_UNREF(features_test);
+	SG_UNREF(labels_test);
+}
+#endif /* HAVE_LAPACK */
```
karlnapf: can you please carefully check which KNN solvers depend on LAPACK (i.e. the source code is guarded with this macro), and then only guard the corresponding unit tests?

MikeLing: Hmm, I found actually there is no need LAPACK for unite test. So let's just remove it. :)

karlnapf: Did you check whether any of the KNN classes (or includes) do call lapack?

MikeLing:
Hi @karlnapf , please give me some more time to work on this pr. I just thought it's a easy issue to track, but I found a lot of places have memory leaking issue after we add unit tests for kd-tree and covertree. I have totally no idea why the memory leaking happened for now. 

I guess it's because it we create several pointer instances and failed to free them. So I suppose to use `SGVector` instead of pointer at least in kd tree and cover tree. Maybe we don't need to separate it into different pr?

karlnapf:
The way to go here is to make the PRs minimal. Changing existing things goes into a PR, adding a new test goes into a PR, etc etc. Small steps. So just remove somethings and then we  proceed from there. It is great if you sport memory leaks as we need to fix them

tests/unit/multiclass/KNN_unittest.cc
MikeLing: Hmm, at least no related error and warning comes out after we remove the lapack macro.

MikeLing:
Hi @karlnapf , I had avoiding the memory leaking issue except covertree. You could test it by uncomment the unit test. Is that possible the covertree itself has memory leaking issue? I have no idea why the memory leak happen except that.

By valgrind, you could see no memory leaking happened until here  https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/multiclass/CoverTreeKNNSolver.cpp#L70  [https://pastebin.mozilla.org/8979841] and it happens on https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/multiclass/KNN.cpp#L191 [https://pastebin.mozilla.org/8979842].  So Maybe it's because we haven't free the pointer in https://github.com/shogun-toolbox/shogun/blob/8fdf278dca78f98d4e06058a5ce3defd560d697d/src/shogun/lib/JLCoverTree.h#L58 after we crate a build the cover trees?

tests/unit/multiclass/KNN_unittest.cc
karlnapf: You have to check the code, just removing warnings wont help. Even if it works on YOUR system, it might not for others

MikeLing:
Hi @karlnapf , ask review for this pr and I will create another pr for fixing memory leaking issue and add unittest for it.

src/shogun/multiclass/CoverTreeKNNSolver.cpp
```diff
@@ -11,7 +11,7 @@ using namespace shogun;
 CCoverTreeKNNSolver::CCoverTreeKNNSolver(const int32_t k, const float64_t q, const int32_t num_classes, const int32_t min_label, const SGVector<int32_t> train_labels):
 CKNNSolver(k, q, num_classes, min_label, train_labels) { /* nothing to do */ }
 
-CMulticlassLabels* CCoverTreeKNNSolver::classify_objects(CDistance* distance, const int32_t num_lab, int32_t* train_lab, float64_t* classes) const
+CMulticlassLabels* CCoverTreeKNNSolver::classify_objects(CDistance* knn_distance, const int32_t num_lab, int32_t* train_lab, float64_t* classes) const
```
karlnapf: why the rename?

karlnapf: Such refactoring should go into a separate mini PR please

src/shogun/multiclass/CoverTreeKNNSolver.cpp
```diff
@@ -66,11 +66,12 @@ if (io->get_loglevel()<= MSG_DEBUG)
 		output->set_label(res[i][0].m_index, out_idx+m_min_label);
 	}
 
-
+	SG_UNREF(r);
```
karlnapf: fixing missing UNREFS should go to a separate mini patch that clearly shows whats going on.

src/shogun/multiclass/KDTreeKNNsolver.cpp
```diff
@@ -40,37 +41,42 @@ CMulticlassLabels* CKDTREEKNNSolver::classify_objects(CDistance* distance, const
...
 
-	CFeatures* lhs = distance->get_lhs();
+	CFeatures* lhs = knn_distance->get_lhs();
+	CFeatures* data = knn_distance->get_rhs();
 	CKDTree* kd_tree = new CKDTree(m_leaf_size);
-	kd_tree->build_tree(dynamic_cast<CDenseFeatures<float64_t>*>(lhs));
-	SG_UNREF(lhs);
 
-	CFeatures* data = distance->get_rhs();
+	kd_tree->build_tree(dynamic_cast<CDenseFeatures<float64_t>*>(lhs));
```
karlnapf: missing null pointer check for cast

src/shogun/multiclass/KDTreeKNNsolver.cpp
```diff
@@ -40,37 +41,42 @@ CMulticlassLabels* CKDTREEKNNSolver::classify_objects(CDistance* distance, const
...
 	return output;
 }
 
-int32_t* CKDTREEKNNSolver::classify_objects_k(CDistance* distance, const int32_t num_lab, int32_t* train_lab, int32_t* classes) const
+int32_t* CKDTREEKNNSolver::classify_objects_k(CDistance* knn_distance, const int32_t num_lab, int32_t* train_lab, int32_t* classes) const
 {
 	int32_t* output=SG_MALLOC(int32_t, m_k*num_lab);
 
 	//allocation for distances to nearest neighbors
 	float64_t* dists=SG_MALLOC(float64_t, m_k);
```
karlnapf: can you use SGVector here, then there is no need to free later on

src/shogun/multiclass/KNN.cpp
```diff
@@ -189,7 +185,6 @@ CMulticlassLabels* CKNN::apply_multiclass(CFeatures* data)
 
 	CMulticlassLabels* output = solver->classify_objects(distance, num_lab, train_lab, classes);
 
-	SG_UNREF(solver);
```
karlnapf: seems convoluted. As above, please address issues one by one in smaller PRs

src/shogun/multiclass/KNN.h
```diff
@@ -87,7 +87,11 @@ class CKNN : public CDistanceMachine
 		 */
 		CKNN(int32_t k, CDistance* d, CLabels* trainlab, KNN_SOLVER knn_solver=KNN_BRUTE);
 
-		virtual ~CKNN();
+		virtual ~CKNN()
+		{
+			SG_UNREF(solver);
+			solver = NULL;
```
karlnapf: not necessary

tests/unit/multiclass/KNN_unittest.cc
```diff
@@ -8,98 +8,118 @@
 
 using namespace shogun;
 
-#ifdef HAVE_LAPACK
```
karlnapf: please adress LAPACK guarding in a separate PR

tests/unit/multiclass/KNN_unittest.cc
```diff
@@ -8,98 +8,118 @@
...
-	   	int32_t num, int32_t classes, int32_t feats)
+class KNNTestData
 {
-	CMath::init_random(1);
-	feat = CDataGenerator::generate_gaussians(num,classes,feats);
-	for( int i = 0 ; i < classes ; ++i )
-		for( int j = 0 ; j < num ; ++j )
-			lab[i*num+j] = double(i);
+	public:
+		KNNTestData()
```
karlnapf: adding unit tests should be different PRs than fixing memory leaks
(unless the test you added covers a bug you fixed)

tests/unit/multiclass/KNN_unittest.cc
```diff
@@ -8,98 +8,118 @@
 
 using namespace shogun;
 
-#ifdef HAVE_LAPACK
-void generate_knn_data(SGMatrix<float64_t>& feat, SGVector<float64_t>& lab,
-	   	int32_t num, int32_t classes, int32_t feats)
+class KNNTestData
```
karlnapf: this test data class (and its usage in the unit tests) should be a sep. refactoring PR

tests/unit/multiclass/KNN_unittest.cc
karlnapf: separate patch pls

MikeLing: hmmm, but the purpose of this pr is about add kdtree test. The rest of code is about fix the error I got after add that test like memory leaking issue and out boundary error. So I think keep this change in this pr would be better.

MikeLing:
This PR is about:
1) fix memory leaking issue in kd-tree (by using SGVector in https://github.com/shogun-toolbox/shogun/pull/3641/files#diff-913bcf4c827c58266f307b28249e97d2R52)

2) fix out boundary error in https://github.com/shogun-toolbox/shogun/pull/3641/files#diff-913bcf4c827c58266f307b28249e97d2R68

3) fix the kd tree solver and brute solver test in unit test (https://github.com/shogun-toolbox/shogun/pull/3641/files#diff-913bcf4c827c58266f307b28249e97d2R68 and add brute solver tester in https://github.com/shogun-toolbox/shogun/pull/3641/files#diff-7252be5bda6101eec8452914e67540b8R162) BWT, I know maybe leave https://github.com/shogun-toolbox/shogun/pull/3641/files#diff-7252be5bda6101eec8452914e67540b8R162 in separate pr would be clearer. But That would change the structure of unit test[2 tests for brute_solve, 1 for ash solver and 2 for kdtree solver] I could refactor all these unit test use clearer way like **KNNTestData** usage in another pr. :)

@karlnapf ask review for this pr. Thank you

PR #3650, opened 2017-02-21
Remove pthread header from SVMLightOneClass

abhinavrai44:


src/shogun/classifier/svm/SVMLightOneClass.cpp
```diff
@@ -33,10 +33,6 @@ extern "C" {
 
 #include <shogun/base/Parallel.h>
 
-#ifdef HAVE_PTHREAD
```
karlnapf: this should be a sep. PR

src/shogun/machine/DistanceMachine.cpp
```diff
@@ -51,168 +43,69 @@ void CDistanceMachine::init()
...
 void CDistanceMachine::distances_lhs(float64_t* result,int32_t idx_a1,int32_t idx_a2,int32_t idx_b)
 {
-    // TODO: port to use OpenMP backend instead of pthread
-#ifdef HAVE_PTHREAD
-    int32_t num_threads=parallel->get_num_threads();
-#else
-    int32_t num_threads=1;
-#endif
-    ASSERT(num_threads>0)
+    int32_t num_threads;
```
karlnapf: why do we need a variable for the number of threads?

src/shogun/machine/DistanceMachine.cpp
```diff
@@ -51,168 +43,69 @@ void CDistanceMachine::init()
...
         {
-            params[t].d = distance;
-            params[t].r = result;
-            params[t].idx_r_start=t*step;
-            params[t].idx_start = (t*step)+idx_a1;
-            params[t].idx_stop = ((t+1)*step)+idx_a1;
-            params[t].idx_comp=idx_b;
-
-            pthread_create(&threads[t], &attr, CDistanceMachine::run_distance_thread_lhs, (void*)&params[t]);
+            num_threads=omp_get_num_threads();
```
karlnapf: @vigsterkr can you comment on the use of `Parallel` here?

src/shogun/machine/DistanceMachine.cpp
```diff
@@ -51,168 +43,69 @@ void CDistanceMachine::init()
...
-            pthread_join(threads[t], NULL);
+        int32_t idx_r_start = thread_num * step;
+        int32_t idx_start = (thread_num * step) + idx_a1;
+        int32_t idx_stop = (thread_num==num_threads) ? (idx_a2 + 1) : ((thread_num + 1) * step) + idx_a1;
 
-        pthread_attr_destroy(&attr);
-        SG_FREE(params);
-        SG_FREE(threads);
+        for(int32_t i=idx_r_start; idx_start < idx_stop; i++,idx_start++)
+            result[i] = distance->distance(idx_start,idx_b);
```
karlnapf: I wonder whether the distance matrix computation should be part of `CDistance` and then we just call that method from here?

src/shogun/machine/gp/Inference.cpp
```diff
@@ -39,18 +39,11 @@
 #include <shogun/mathematics/Math.h>
 #include <shogun/lib/Lock.h>
 
-using namespace shogun;
+#ifdef HAVE_OPENMP
```
karlnapf: can you please do every class in a separate PR?

src/shogun/machine/DistanceMachine.cpp
abhinavrai44: @karlnapf The variable involved need to be computed only once as they will remain constant over all the threads. In parallel the same computations will be repeated over all threads.

src/shogun/classifier/svm/SVMLightOneClass.cpp
abhinavrai44: ok.

src/shogun/machine/gp/Inference.cpp
abhinavrai44: Ok :-)

vigsterkr:
dunno what's happening with travis but i think this one is really safe to merge....

Issue #3653, opened 2017-02-23
avoid re-generating meta examples on `make`


karlnapf:
Currently, typing `make` (with meta examples enabled) generated output like
```
[  1%] Generating example gaussian_processes-gaussian_process_classifier
[  2%] Generating example binary_classifier-lda
[  2%] Generating example binary_classifier-linear_svm
[  2%] Generating example binary_classifier-kernel_svm
[  2%] Generating example regression-kernel_ridge_regression
[  2%] Generating example regression-krr_nystrom
[  2%] Generating example regression-linear_ridge_regression
[  2%] Generating example regression-multiple_kernel_learn
```

And then typing `make` again after the first one has finished, all meta examples are generated (and compiled) again. This issue is to stop that happening. This used to be working, so part of the solution will be to investigate why it does not work anymore. Note this will also affect all the other compiled target langauges.

karlnapf:
fixed in #3659

PR #3659, opened 2017-02-27
Avoid re-generating meta examples on `make`

geektoni:
Fix for issue #3653.

I believe this behaviour appeared after the introduction of the Scala interface. As I read from pull #3597, it was decided to run the java .class file with Scala (without generating the .scala files). However, the meta_examples target expects to find the all the .scala files, but since they are not present it triggers the re-generation of the meta examples.

IMHO this is a ugly workaround, but it solves the problem until a more advanced solution arrives.


karlnapf:
There are new errors in the scala build now. Repdoduce locally with `-DENABLE_TESTING=ON`

See https://travis-ci.org/shogun-toolbox/shogun/jobs/205716515#L4975

geektoni:
Correct the Scala errors. The others Travis failures should be unrelated to my changes.

examples/meta/CMakeLists.txt
```diff
@@ -36,13 +36,19 @@ FOREACH(META_EXAMPLE ${META_EXAMPLES})
     SET(EXAMPLE_LISTINGS)
     SET(BASENAME ${EXAMPLE_REL_DIR}/${EXAMPLE_NAME})
     LIST(APPEND EXAMPLE_LISTINGS ${CMAKE_CURRENT_BINARY_DIR}/cpp/${BASENAME}.cpp)
+
+    # FIXME
+    # Hack to avoid generation of scala meta examples dependencies that will
+    # not be generated by generate.py script (since there is no scala.json)
+    LIST(REMOVE_ITEM AVAILABLE_INTERFACES "ScalaModular")
```
karlnapf: fine with me this hack

karlnapf:
Thanks!

examples/meta/CMakeLists.txt
vigsterkr: peeeeeeeeeeeeople `DISABLED_INTERFACES`

geektoni: Adding the Scala module to ```DISABLED_INTERFACES``` will cause to not include examples/meta/scala/CMakeLists.txt. Only the Scala integration tests will be generated (and they will fail) and the flag ```-DScalaModular``` will become a bit useless (it will behave like an alias of ```-DJavaModular=ON```). 

This hack permits to create all the Scala tests anyway (generated and integration), but as I said before, this is just a workaround to prevent the continuous generation of the meta examples code.

PR #3661, opened 2017-02-27
Port Inference.cpp to use OpenMP

abhinavrai44:


vigsterkr:
@abhinavrai44 could you rebase with the latest develop

src/shogun/machine/gp/Inference.cpp
```diff
@@ -214,105 +207,47 @@ get_negative_log_marginal_likelihood_derivatives(CMap<TParameter*, CSGObject*>*
...
 	{
-#endif /* HAVE_PTHREAD */
-		for (index_t i=0; i<num_deriv; i++)
-		{
-			CMapNode<TParameter*, CSGObject*>* node=params->get_node_ptr(i);
-
-			GRADIENT_THREAD_PARAM thread_params;
+        CMapNode<TParameter*, CSGObject*>* node=params->get_node_ptr(i);
+        SGVector<float64_t> gradient;
+        CLock* locked = &lock;
```
vigsterkr: does this really need to be a pointer? you are not using lock as a function arg anymore anywhere afaik

src/shogun/machine/gp/Inference.cpp
```diff
@@ -214,105 +207,47 @@ get_negative_log_marginal_likelihood_derivatives(CMap<TParameter*, CSGObject*>*
...
+			SG_SERROR("Can't compute derivative of negative log marginal "
+					"likelihood wrt %s.%s", node->data->get_name(), node->key->m_name);
 		}
 
-		for (index_t t=0; t<num_deriv; t++)
-			pthread_join(threads[t], NULL);
-
-		SG_FREE(thread_params);
-		SG_FREE(threads);
+		locked->lock();
```
vigsterkr: #pragma omp critical section would be muc more adequate here, but i would rather refactor this in a lock-free way..... and use maybe `#pragma omp parallel for reduction(...)`

src/shogun/machine/gp/Inference.cpp
```diff
@@ -39,18 +39,11 @@
 #include <shogun/mathematics/Math.h>
 #include <shogun/lib/Lock.h>
 
-using namespace shogun;
+#ifdef HAVE_OPENMP
+#include <omp.h>
```
vigsterkr: why do you need this header?
i might missed it but i dont see anywhere in the code omp API being used, or?

abhinavrai44: omp is used in the function : get_negative_log_marginal_likelihood_derivatives

vigsterkr: @abhinavrai44 i dont get it.. where? if you refer to `#pragma omp...` that's not an omp API per se... it's a preprocessor directive and for that you do not need to include `omp.h`

abhinavrai44: @vigsterkr I didn't know that. Will make the chages. Thanks for pointing that out.

abhinavrai44:
@vigsterkr Done

Issue #3520, opened 2016-11-08
`make cookbook` fails when certain cookbooks are guarded


karlnapf:
See http://pastebin.com/sDjUi3bb

karlnapf:
This comes since `FindMetaExamples.cmake` sometimes removes certain meta examples.
The cookbook should still build fine, just ignoring that particular page....

IOcodegeass:
I faced this problem. And installing **nlopt** solved it.
Thanks :)

karlnapf:
Yeah, it is not really a solve but rather a workaround :) We should deal with this properly....

karlnapf:
Fixed in #3665
Thanks!

PR #3665, opened 2017-02-28
Fix `make cookbook` when certain cookbooks are guarded

geektoni:
This should fix issue #3520. These changes fix also ``` make cookbook_sphinx_linkcheck```.

The command ```make cookbook``` still gives me warnings, though, but I don't understand if they are caused by my changes or if they were already there.
See warnings here: http://pastebin.com/chxfGvaq

### DIFF
--- a/cmake/FindMetaExamples.cmake
+++ b/cmake/FindMetaExamples.cmake
@@ -15,40 +15,85 @@ function(is_set CONFIG_FILE FLAG)
 	endif()
 endfunction()
 
-function(find_meta_examples)
-	read_config_header(${shogun_INCLUDE_DIR}/shogun/lib/config.h)
-	is_set(${SHOGUN_CONFIG} HAVE_NLOPT)
-	is_set(${SHOGUN_CONFIG} USE_GPL_SHOGUN)
-	is_set(${SHOGUN_CONFIG} HAVE_LAPACK)
-	is_set(${SHOGUN_CONFIG} USE_SVMLIGHT)
+function(get_excluded_meta_examples)
+    read_config_header(${shogun_INCLUDE_DIR}/shogun/lib/config.h)
+    is_set(${SHOGUN_CONFIG} HAVE_NLOPT)
+    is_set(${SHOGUN_CONFIG} USE_GPL_SHOGUN)
+    is_set(${SHOGUN_CONFIG} HAVE_LAPACK)
+    is_set(${SHOGUN_CONFIG} USE_SVMLIGHT)
 
-	FILE(GLOB_RECURSE META_EXAMPLE_LISTINGS ${CMAKE_SOURCE_DIR}/examples/meta/src/*.sg)
-
-	# temporary hacks to exclude certain meta examples that have dependencies
-	IF(NOT HAVE_NLOPT)
-		LIST(REMOVE_ITEM META_EXAMPLE_LISTINGS
-			${CMAKE_SOURCE_DIR}/examples/meta/src/gaussian_processes/gaussian_process_regression.sg)
-	ENDIF()
+    IF(NOT HAVE_NLOPT)
+        LIST(APPEND EXCLUDED_META_EXAMPLES
+            gaussian_processes/gaussian_process_regression.sg)
+    ENDIF()
 
 	IF(NOT USE_GPL_SHOGUN)
-		LIST(REMOVE_ITEM META_EXAMPLE_LISTINGS
-			${CMAKE_SOURCE_DIR}/examples/meta/src/gaussian_processes/gaussian_process_regression.sg
-			${CMAKE_SOURCE_DIR}/examples/meta/src/multiclass_classifier/multiclass_logisticregression.sg
+		LIST(APPEND EXCLUDED_META_EXAMPLES
+			gaussian_processes/gaussian_process_regression.sg
+			multiclass_classifier/multiclass_logisticregression.sg
 			)
 	ENDIF()
 
 	IF(NOT HAVE_LAPACK)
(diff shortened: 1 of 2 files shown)


geektoni:
Mmmh, I think you are pretty right. It was more like a "quick fix" than a proper solution. I refactored the code a bit to make it more readable and modular. This new approach should be better.

**<off_topic>**
Can I ask why we have to deal with missing dependencies? We could simply add these libraries as required to compile shogun, instead of setting up these "workarounds". Obviously, I'm sure that you have a bunch of good reasons for doing so. ;)
**</off_topic>**

karlnapf:
Yes, an overwhealming number of good reasons for that :)
Come to IRC and we can discuss

cmake/FindMetaExamples.cmake
```diff
@@ -15,40 +15,75 @@ function(is_set CONFIG_FILE FLAG)
...
+    is_set(${SHOGUN_CONFIG} HAVE_LAPACK)
+    is_set(${SHOGUN_CONFIG} USE_SVMLIGHT)
 
-	# temporary hacks to exclude certain meta examples that have dependencies
-	IF(NOT HAVE_NLOPT)
-		LIST(REMOVE_ITEM META_EXAMPLE_LISTINGS
-			${CMAKE_SOURCE_DIR}/examples/meta/src/gaussian_processes/gaussian_process_regression.sg)
+    IF(NOT HAVE_NLOPT)
+		LIST(APPEND EXCLUDED_META_EXAMPLES
+			gaussian_processes/gaussian_process_regression.sg)
```
karlnapf: +1 this is cleaner!

cmake/FindMetaExamples.cmake
```diff
@@ -15,40 +15,75 @@ function(is_set CONFIG_FILE FLAG)
...
+
+endfunction()
+
+# Remove meta example that cannot be built because of missing dependencies
+function(find_meta_examples)
+
+	FILE(GLOB_RECURSE META_EXAMPLE_LISTINGS ${CMAKE_SOURCE_DIR}/examples/meta/src/*.sg)
+    get_excluded_meta_examples()
+
+	# temporary hacks to exclude certain meta examples that have dependencies
```
karlnapf: I guess this comment can be removed now, as it is a bit nicer

cmake/FindMetaExamples.cmake
```diff
@@ -15,40 +15,75 @@ function(is_set CONFIG_FILE FLAG)
...
+    FOREACH(META_EXAMPLE ${EXCLUDED_META_EXAMPLES})
+        LIST(REMOVE_ITEM META_EXAMPLE_LISTINGS ${CMAKE_SOURCE_DIR}/examples/meta/src/${META_EXAMPLE})
+    ENDFOREACH()
+
 	SET(META_EXAMPLES ${META_EXAMPLE_LISTINGS} PARENT_SCOPE)
 endfunction()
+
+# Get the cookbook pages we want to exclude from build
+function(find_excluded_meta_examples)
```
karlnapf: I think this function should be named differently, as it is about cookbook pages, not meta examples themselves

cmake/FindMetaExamples.cmake
```diff
@@ -15,40 +15,75 @@ function(is_set CONFIG_FILE FLAG)
...
+
+    FOREACH(META_EXAMPLE ${EXCLUDED_META_EXAMPLES})
+        LIST(APPEND EXCLUDED_COOKBOOK_PAGES examples/${META_EXAMPLE})
+    ENDFOREACH()
+
+    # Generate a string with all the meta examples separated by commas.
+    # This is made since Sphinx's exclude_patterns option wants
+    # the list's items separated by commas, but cmake's lists use
+    # semicolons instead.
+    # See: https://cmake.org/cmake/help/v3.3/command/list.html
```
karlnapf: good to put that comment

doc/cookbook/CMakeLists.txt
```diff
@@ -1,5 +1,7 @@
 #PROJECT(sphinxdoc)
 #cmake_minimum_required(VERSION 2.8.8)
+include(FindMetaExamples)
+find_excluded_meta_examples()
```
karlnapf: rename as said above

cmake/FindMetaExamples.cmake
```diff
@@ -15,40 +15,75 @@ function(is_set CONFIG_FILE FLAG)
...
 	SET(META_EXAMPLES ${META_EXAMPLE_LISTINGS} PARENT_SCOPE)
 endfunction()
+
+# Get the cookbook pages we want to exclude from build
+function(find_excluded_meta_examples)
+
+    get_excluded_meta_examples()
+
+    FOREACH(META_EXAMPLE ${EXCLUDED_META_EXAMPLES})
+        LIST(APPEND EXCLUDED_COOKBOOK_PAGES examples/${META_EXAMPLE})
```
karlnapf: one thing to consider here: Some meta examples do not have cookbook pages. So they do not need to be excluded technically. Not sure whether that causes problems, but it is definitely cleaner to only work on the existing cookbook pages

geektoni:
I made the suggested changes @karlnapf. 
By the way, I would definitely like to know more about these management choices about shogun dependencies. Are you shogun developers "always online" on IRC or there is a preferred time frame?

karlnapf:
Fine to merge from my side. @vigsterkr ?

karlnapf:
@geektoni IRC hours depend. I am UTC 11am-7pm usually. Others might differ.

vigsterkr:
go

karlnapf:
Weird failures in travis, I just restarted to be safe.

PR #3666, opened 2017-03-01
Remove FeaturesUtil class, use CFeatures methods instead

lambday:
Signed-off-by: lambday <heavensdevil6909@gmail.com>


src/shogun/statistical_testing/internals/Block.cpp
```diff
@@ -48,7 +48,9 @@ Block::Block(CFeatures* feats, index_t index, index_t size) : m_feats(feats)
 
 	// create a shallow copy and subset current block separately
 	CFeatures* block=FeaturesUtil::create_shallow_copy(feats);
+#ifdef USE_REFERENCE_COUNTING
 	ASSERT(block->ref_count()==0);
```
karlnapf: I mean this fixes the error, but I find the line that is guarded incredibly discomforting tbh

lambday: @karlnapf it is internal to just check sanity. Users cannot screw this up, therefore no REQUIRE.

karlnapf: So my thoughts are

 * if this assertion could possibly ever fail, then there is some serious problem in the `create_shallow_copy` above.
 * If this assertion never fails, i.e. the above line guarantees something that is documented in its interface and maybe even unit tested, then why put it?

lambday: Agreed. So, I think 
- this ASSERT should itself be a unit-test (because if someone changes the impl of `FeaturesUtil::create_shallow_copy` later, e.g. `SG_REF` it before returning it, (s)he should be alerted cause it might break things)
- Actually I should move this thing to be a part of the `CFeatures` API maybe. This functionality has no reason to stay within statistical testing hierarchy.

Let me know what you think.

karlnapf: Both yes!
And actually the Shogun policy is to return new objects with ref-count 1

src/shogun/features/FeaturesUtil.h
```diff
@@ -43,37 +43,25 @@ namespace internal
...
-	 * @param other The feature object whose shallow copy has to be created.
-	 * @return A shallow copy of the feature object.
-	 */
-	static CFeatures* create_shallow_copy(CFeatures* other);
-
-	/**
 	 * This creates a merged copy of the two feature objects.
 	 *
+	 * Please note that unlike CFeatures::create_merged_copy, this method
+	 * respects the subsets that are added to **both** these feature instances.
```
karlnapf: Why would this method not be in `CFeatures`
If it should be eventually, could you create an issue or a TODO item for this?

src/shogun/statistical_testing/internals/Block.cpp
```diff
@@ -47,32 +46,33 @@ Block::Block(CFeatures* feats, index_t index, index_t size) : m_feats(feats)
 	SG_REF(m_feats);
 
 	// create a shallow copy and subset current block separately
-	CFeatures* block=FeaturesUtil::create_shallow_copy(feats);
-	ASSERT(block->ref_count()==0);
+	CFeatures* block=feats->shallow_subset_copy();
```
karlnapf: shallow_subset_copy?
Why not shallow_copy and if there is a subset, it is copied?

src/shogun/features/FeaturesUtil.cpp
```diff
@@ -28,43 +28,17 @@
  * either expressed or implied, of the Shogun Development Team.
```
karlnapf: Not sure why we need this new class?

src/shogun/statistical_testing/internals/Block.cpp
lambday: @karlnapf that's what this method does :) I think Saurabh added it last year.

lambday:
Yo @karlnapf. The changes are done. Let's see what the CIs say. Works fine locally.

src/shogun/statistical_testing/internals/Block.cpp
lambday: For our present requirements, this method suffices. But it does not clone the entire subset and attach it to the shallow copy, rather just pushes the last working index into it. So, if someone wants to pop out the subsets in the copy, it can't be done. Should the need arise, will have to add that.

Also, maybe this method could use a better name. Maybe `shallow_copy_with_active_subset()` or so?

karlnapf: No actually I realise now why this was done. Makes sense I think. "shallow_subset"-copy
Lets leave it for now

src/shogun/features/DenseFeatures.cpp
```diff
@@ -1036,24 +1036,34 @@ template<class ST> CFeatures* CDenseFeatures<ST>::create_merged_copy(
 	SGMatrix<ST> data(num_features, num_vectors_merged);
 
 	/* copy data of this instance */
-	SG_DEBUG("copying matrix of this instance\n")
-	memcpy(data.matrix, feature_matrix.matrix,
-			num_features*num_vectors*sizeof(ST));
+	SG_DEBUG("copying matrix of this instance\n");
```
karlnapf: this debug message should go down in the else branch.

karlnapf: Or just remove it actually ....

src/shogun/features/DenseFeatures.cpp
```diff
@@ -1036,24 +1036,34 @@ template<class ST> CFeatures* CDenseFeatures<ST>::create_merged_copy(
...
+	if (m_subset_stack->has_subsets())
+	{
+		SGMatrix<ST> active_feats_mat=get_feature_matrix();
```
karlnapf: Not a fan of this. Double memory usage.
Either create a method that extract the subset matrix into a pre-allocated one, or copy things by hand here.

lambday: Agreed. A separate private method should simply accept a raw ptr (or maybe a SGMatrix) to copy things into. If the ptr (SGMatrix.matrix) is empty or doesn't have enough space to memcopy things into, it reallocates.

Then this method as well as the other get_feature_matrix method can rely on that method.

src/shogun/features/DenseFeatures.cpp
lambday: Alright

lambday:
@karlnapf done! Please have a look.

src/shogun/features/DenseFeatures.cpp
```diff
@@ -265,23 +267,51 @@ template<class ST> void CDenseFeatures<ST>::feature_subset(int32_t* idx, int32_t
...
 
-template<class ST> SGMatrix<ST> CDenseFeatures<ST>::get_feature_matrix()
+template <class ST>
+SGMatrix<ST> CDenseFeatures<ST>::get_feature_matrix()
 {
 	if (!m_subset_stack->has_subsets())
 		return feature_matrix;
 
-	SGMatrix<ST> submatrix(num_features, get_num_vectors());
+	return deep_copy_feature_matrix(SGMatrix<ST>(), 0);
```
karlnapf: So now you changed behaviour: get_feature_matrix now always returns a copy?

karlnapf: Not sure about this. There has to be a getter for the matrix itself (REQUIRING that there is no subset).

Note you can also be careful with const here

src/shogun/features/DenseFeatures.h
```diff
@@ -541,6 +543,23 @@ template<class ST> class CDenseFeatures: public CDotFeatures
 	void init();
 
 protected:
+	/*
+	 * Helper method which copies the working feature matrix into the pre-allocated
```
karlnapf: I would prefer is this method would not get involved with any re-allocation business (sep. of concerns), but rather only REQUIRE that the sizes match.

src/shogun/features/DenseFeatures.h
```diff
@@ -541,6 +543,23 @@ template<class ST> class CDenseFeatures: public CDotFeatures
...
+	 * not sufficient to copy all the feature vectors, it reallocates the memory to fit
+	 * the required size, with the column_offset. It then copies into the target
+	 * matrix, starting from base + (colum_offset * num_features) location.
+	 *
+	 * In case reallocation happens, it initializes the memory till the offset location
+	 * with 0, in order to avoid leaving it uninitialized. However, if reallocation
+	 * does not happen within this method, it is assumed that it is the responsilibity
+	 * of the caller to initialize the offsetted part.
+	 *
+	 * Please note that this method always does a deep copy, irrespective of whether
```
karlnapf: what does deep copy mean here? It is just a copy no?

src/shogun/features/DenseFeatures.cpp
lambday: Haven't changed the behavior. It returns the same underlying matrix when there are no subsets. Please check the above return statement that I haven't touched.

src/shogun/features/DenseFeatures.h
lambday: What I mean is, even if there are no subsets, it still creates a new SGMatrix and copies into it rather than returning the original one. But I agree, only copy would suffice to convey the same sense :)

src/shogun/features/DenseFeatures.cpp
karlnapf: Sorry you are right, of course it still returns ....

OK so let's leave that for now (old behaviour is stil ugly but not your concern here, can do sep patch, maybe create an entrance task?)

src/shogun/features/DenseFeatures.h
karlnapf: rename and merge then :)

karlnapf: Ah no, the resizing/re-allocating, I think that shouldnt be there

lambday:
@karlnapf removed the realloc business, renamed the method.

src/shogun/features/DenseFeatures.h
lambday: agreed. changed it accordingly.

src/shogun/features/DenseFeatures.cpp
```diff
@@ -265,23 +267,51 @@ template<class ST> void CDenseFeatures<ST>::feature_subset(int32_t* idx, int32_t
...
+	REQUIRE(column_offset>=0, "Column offset (%d) cannot be negative!\n", column_offset);
 
-	/* copy a subset vector wise */
-	for (int32_t i=0; i<submatrix.num_cols; ++i)
+	index_t num_vecs=get_num_vectors();
+	index_t num_cols=num_vecs+column_offset;
+
+	REQUIRE(target.matrix!=nullptr, "Provided matrix is not allocated!\n");
+	REQUIRE(target.num_rows==num_features,
+			"Number of rows (%d for given matrix) should be equal to the number of features (%d)!\n",
```
karlnapf: `Number of rows of given matrix (%d)` would be more inline the messages we have. But that is very very minor

src/shogun/features/DenseFeatures.cpp
```diff
@@ -977,101 +1007,66 @@ template<class ST> bool CDenseFeatures<ST>::is_equal(CDenseFeatures* rhs)
...
-		CList* others)
+template <class ST>
+CFeatures* CDenseFeatures<ST>::create_merged_copy(CList* others)
 {
-	SG_DEBUG("entering %s::create_merged_copy()\n", get_name());
+	SG_DEBUG("Entering.\n");
 
-	if (!others)
-		return NULL;
+	REQUIRE(others!=nullptr, "The list of feature instances cannot be null!\n");
```
karlnapf: "null" is not something that should appear in an error message that can be triggered by SWIG.

src/shogun/features/DenseFeatures.cpp
```diff
@@ -977,101 +1007,66 @@ template<class ST> bool CDenseFeatures<ST>::is_equal(CDenseFeatures* rhs)
...
 	{
-		CDenseFeatures<ST>* casted=dynamic_cast<CDenseFeatures<ST>* >(other);
+		auto casted=dynamic_cast<CDenseFeatures<ST>*>(current);
 
-		if (!casted)
-		{
-			SG_ERROR("%s::create_merged_copy(): Could not cast object of %s to "
-					"same type as %s\n",get_name(), other->get_name(), get_name());
-		}
+		REQUIRE(casted!=nullptr, "Could not cast object of %s to same type as %s!\n",
```
karlnapf: Same here, make it user understandable.
"Provided object's type (%s) at index (%d) in list must match own type (%s)

src/shogun/features/DenseFeatures.cpp
```diff
@@ -977,101 +1007,66 @@ template<class ST> bool CDenseFeatures<ST>::is_equal(CDenseFeatures* rhs)
...
+	num_copied+=get_num_vectors();
 
-	/* count number of vectors (not elements) processed so far */
-	index_t num_processed=num_vectors;
+	current=others->get_first_element();
 
-	/* now copy data of other features block wise */
-	other=others->get_first_element();
-	while (other)
+	while (current!=nullptr)
```
karlnapf: entrance task: write iterators for our data structures ? Thoughts?

src/shogun/features/DenseFeatures.h
karlnapf: doxygen @param docs are missing

src/shogun/statistical_testing/internals/StreamingDataFetcher.cpp
```diff
@@ -30,14 +30,15 @@ StreamingDataFetcher::StreamingDataFetcher(CStreamingFeatures* samples)
 : DataFetcher(), parser_running(false)
 {
 	REQUIRE(samples!=nullptr, "Samples cannot be null!\n");
```
karlnapf: If this can be caused by SWIG interface (i.e. user), should not contain the word NULL

tests/unit/features/DenseFeatures_unittest.cc
```diff
@@ -52,13 +69,58 @@ TEST(DenseFeaturesTest,create_merged_copy)
...
+	SGMatrix<float64_t> expected_merged_mat(dim, active_data_1.num_cols+active_data_2.num_cols);
+	std::copy(active_data_1.matrix, active_data_1.matrix+active_data_1.size(),
+			expected_merged_mat.matrix);
+	std::copy(active_data_2.matrix, active_data_2.matrix+active_data_2.size(),
+			expected_merged_mat.matrix+active_data_1.size());
+
+	auto merged=static_cast<CDenseFeatures<float64_t>*>(features_1->create_merged_copy(features_2));
+	SGMatrix<float64_t> merged_mat=merged->steal_feature_matrix();
+
+	EXPECT_EQ(expected_merged_mat.num_rows, merged_mat.num_rows);
```
karlnapf: ASSERT_EQ
otherwise you can get a memory read error in the subsequent loop

tests/unit/features/DenseFeatures_unittest.cc
```diff
@@ -148,9 +205,47 @@ TEST(DenseFeaturesTest, shallow_copy_subset_data)
...
+	auto features=some<CDenseFeaturesMock>(data);
+	features->add_subset(inds);
+
+	index_t offset=3;
+	SGMatrix<float64_t> copy(dim, inds.vlen+offset);
+	auto data_ptr=copy.matrix;
+	std::fill(copy.data(), copy.data()+copy.size(), 0);
+	features->copy_feature_matrix_public(copy, offset);
+
+	EXPECT_EQ(copy.num_rows, dim);
```
karlnapf: ASSERT

src/shogun/features/DenseFeatures.cpp
lambday: Yeah iterators would be really good to have. Will have to discuss with you about the scope.

src/shogun/features/DenseFeatures.h
lambday: This is a protected method :) Doxygen won't see this.

lambday:
Thanks @karlnapf for your comments. Agreed on the fixes and error messages that you suggested. Will fix these after rebase.

This patch has really gone big I agree. But all these changes are related to the fact that we wanted to get rid of FeaturesUtil class and use CFeatures methods instead. So, for this one, I am renaming it like that. In future, one patch for one change. 

And maybe I should do it in feature branches as @vigsterkr suggests instead of PRs, so that if I'm away and merge is stalled because of rebase, someone else can do that and merge to develop. Should be more convenient and effective that way?

src/shogun/features/DenseFeatures.cpp
```diff
@@ -265,23 +267,51 @@ template<class ST> void CDenseFeatures<ST>::feature_subset(int32_t* idx, int32_t
...
+	if (!m_subset_stack->has_subsets())
 	{
-		int32_t real_i = m_subset_stack->subset_idx_conversion(i);
-		memcpy(&submatrix.matrix[i*int64_t(num_features)],
-				&feature_matrix.matrix[real_i * int64_t(num_features)],
-				num_features * sizeof(ST));
+		auto input_begin=feature_matrix.matrix;
+		auto input_end=input_begin+feature_matrix.size();
+		auto output_begin=target.matrix+int64_t(num_features)*column_offset;
+		std::copy(input_begin, input_end, output_begin);
```
lambday: should replace with newly added `shogun::memcpy` here. Commenting just so I don't forget

src/shogun/features/DenseFeatures.cpp
```diff
@@ -265,23 +267,51 @@ template<class ST> void CDenseFeatures<ST>::feature_subset(int32_t* idx, int32_t
...
+	}
+	else
+	{
+		for (int32_t i=0; i<num_vecs; ++i)
+		{
+			auto real_i=m_subset_stack->subset_idx_conversion(i);
+			auto input_begin=feature_matrix.matrix+real_i*int64_t(num_features);
+			auto input_end=input_begin+num_features;
+			auto output_begin=target.matrix+int64_t(num_features)*(column_offset+i);
+			std::copy(input_begin, input_end, output_begin);
```
lambday: here as well.

karlnapf:
Yeah the big things should go in feature branches.
Yeah I see the changes are all related, so merge it.
It is just that it would be easier to review this in an incremental fashion

lambday:
CIs are happy. Merging this one.

PR #3667, opened 2017-03-01
Fix data races in CCrossValidation::evaluate_one_run()

lkuchenb:
* CList::get_first_element() is potentially not thread-safe
 * CEvaluation object was shared

This does not solve all of the problem. The problem persists at a lower frequency. I'll open an issue on this.

### DIFF
--- a/src/shogun/evaluation/CrossValidation.cpp
+++ b/src/shogun/evaluation/CrossValidation.cpp
@@ -270,17 +270,27 @@ float64_t CCrossValidation::evaluate_one_run()
 			CMachine* machine;
 			CFeatures* features;
 			CLabels* labels;
+			CEvaluation* evaluation_criterion;
 
 			if (get_global_parallel()->get_num_threads()==1)
+			{
 				machine=m_machine;
+				features=m_features;
+				evaluation_criterion=m_evaluation_criterion;
+			}
 			else
+			{
 				machine=(CMachine*)m_machine->clone();
+				features=(CFeatures*)m_features->clone();
+				evaluation_criterion=(CEvaluation*)m_evaluation_criterion->clone();
+			}
 
 			/* evtl. update xvalidation output class */
-			CCrossValidationOutput* current=(CCrossValidationOutput*)
-					m_xval_outputs->get_first_element();
+			CCrossValidationOutput* current;
 			#pragma omp critical
 			{
+			current=(CCrossValidationOutput*)
+					m_xval_outputs->get_first_element();
 			while (current)
 			{
 				current->update_fold_index(i);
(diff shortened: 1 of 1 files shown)


src/shogun/evaluation/CrossValidation.cpp
```diff
@@ -270,17 +270,27 @@ float64_t CCrossValidation::evaluate_one_run()
...
+			{
 				machine=m_machine;
+				features=m_features;
+				evaluation_criterion=m_evaluation_criterion;
+			}
 			else
+			{
 				machine=(CMachine*)m_machine->clone();
+				features=(CFeatures*)m_features->clone();
+				evaluation_criterion=(CEvaluation*)m_evaluation_criterion->clone();
```
vigsterkr: this is missing it's `SG_UNREF` pair on the end of the for loop, when `(get_global_parallel()->get_num_threads()==1)`

karlnapf: Does `clone` actually return a zero a one refcount?
If it is zero, it should be `SG_REF`ed here as well, same for features

lkuchenb: clone() SG_REFs the copy

karlnapf: ok thanks!

src/shogun/evaluation/CrossValidation.cpp
```diff
@@ -270,17 +270,27 @@ float64_t CCrossValidation::evaluate_one_run()
...
 
 			if (get_global_parallel()->get_num_threads()==1)
+			{
 				machine=m_machine;
+				features=m_features;
+				evaluation_criterion=m_evaluation_criterion;
+			}
 			else
+			{
 				machine=(CMachine*)m_machine->clone();
```
karlnapf: I wonder whether the features of the machine are cloned here? I guess they will be. Then they will be overwritten with the manual clone below. Maybe it is good to remove any features from the `CMachine` instance before cloning to avoid doing this twice. Alternatively, set the features, clone machine, and then extract them from there.
Requires some checking of the existing code

src/shogun/evaluation/CrossValidation.cpp
```diff
@@ -270,17 +270,27 @@ float64_t CCrossValidation::evaluate_one_run()
...
+			}
 
 			/* evtl. update xvalidation output class */
-			CCrossValidationOutput* current=(CCrossValidationOutput*)
-					m_xval_outputs->get_first_element();
+			CCrossValidationOutput* current;
 			#pragma omp critical
 			{
+			current=(CCrossValidationOutput*)
+					m_xval_outputs->get_first_element();
```
karlnapf: @vigsterkr we really should have some proper iterators for our structures ....

karlnapf: I think `get_first_element` also increases the ref-count?

src/shogun/evaluation/CrossValidation.cpp
```diff
@@ -270,17 +270,27 @@ float64_t CCrossValidation::evaluate_one_run()
...
 			if (get_global_parallel()->get_num_threads()==1)
+			{
 				machine=m_machine;
+				features=m_features;
+				evaluation_criterion=m_evaluation_criterion;
+			}
 			else
+			{
 				machine=(CMachine*)m_machine->clone();
+				features=(CFeatures*)m_features->clone();
```
karlnapf: Next patch: Don't clone the features :)

PR #3670, opened 2017-03-02
LinalgRefactor - enable GPU_WARNINGS switch

OXPHOS:


karlnapf:
Ok from my side.
@vigsterkr you got thoughts on this?

vigsterkr:
go!

vigsterkr:
@karlnapf @OXPHOS although why compile time? why not runtime parameter? like env var?

OXPHOS:
I do feel like runtime makes more sense now since you mentioned

vigsterkr:
ok here's where it should be implemented, and there you can see how it's done for 2 different env vars:
https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/base/init.cpp#L191-L222

OXPHOS:
@vigsterkr Revised. Not sure whether the var name is too long..

vigsterkr:
ahahah yeah that var SHOGUN_LINALG_LIB_WARNINGS is pretty long :)

OXPHOS:
@vigsterkr Hope it's better now. `SHOGUN_GPU_WARNINGS` doesn't sound accurate to me but it is the most concise one I can think of for now.

PR #3673, opened 2017-03-03
Bahsically remove that BAHSIC

lisitsyn:


vigsterkr:
if it's a branch no need to pr as it's gonna then make the CIs double work as you can see :)

karlnapf:
@lambday will we have hsic back at some point?

PR #3674, opened 2017-03-03
Use weight vector *only* through getters/setters

lisitsyn:
Please take a close look this is dangerous indeed.

The PR is a small step towards tags.


karlnapf:
Some failed:
https://travis-ci.org/shogun-toolbox/shogun/jobs/207534368#L6795

lisitsyn:
@karlnapf thanks. I'll take a look.

karlnapf:
Can we benchmark this before and after the change?

PR #3675, opened 2017-03-04
Fix TParameter copy

lkuchenb:
`StringFeatures<T>` contain an `SGString<T>*` parameter that caused `::clone()` on the features to fail.

IMHO this fixes `TParameter::copy()`, however, I'm not familiar with this code, so please check (see inline comments)

### DIFF
--- a/src/shogun/base/Parameter.cpp
+++ b/src/shogun/base/Parameter.cpp
@@ -3844,7 +3844,7 @@ bool TParameter::copy(TParameter* target)
 				*(char**)target->m_parameter=SG_MALLOC(char, num_bytes);
 				/* check whether ptype is SGOBJECT, if yes we need to initialize
 				   the memory with NULL for the way copy_ptype handles it */
-				if (m_datatype.m_ptype==PT_SGOBJECT)
+				if (m_datatype.m_ptype==PT_SGOBJECT || m_datatype.m_stype==ST_STRING)
 					memset(*(void**)target->m_parameter, 0, num_bytes);
 
 				/* use length of source */
@@ -3860,7 +3860,7 @@ bool TParameter::copy(TParameter* target)
 			SG_SDEBUG("length_y: %d\n", *m_datatype.m_length_y)
 			for (index_t i=0; i<*m_datatype.m_length_y; ++i)
 			{
-				SG_SDEBUG("copying element %d which is %d byes from start\n",
+				SG_SDEBUG("copying element %d which is %d bytes from start\n",
 						i, x);
 
 				void* pointer_a=&((*(char**)m_parameter)[x]);
@@ -3874,7 +3874,7 @@ bool TParameter::copy(TParameter* target)
 					return false;
 				}
 
-				x=x+(m_datatype.sizeof_ptype());
+				x=x+(m_datatype.sizeof_stype());
 			}
 
 			break;
--- a/tests/unit/features/StringFeatures_unittest.cc
+++ b/tests/unit/features/StringFeatures_unittest.cc
@@ -14,12 +14,8 @@
 
 using namespace shogun;
 
-TEST(StringFeaturesTest,copy_subset)
+SGStringList<char> generateRandomData(index_t num_strings=10, index_t max_string_length=20, index_t min_string_length=10)
 {
-	index_t num_strings=10;
-	index_t max_string_length=20;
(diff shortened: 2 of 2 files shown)


tests/unit/features/StringFeatures_unittest.cc
```diff
@@ -106,3 +108,30 @@ TEST(StringFeaturesTest,copy_subset)
...
+TEST(StringFeaturesTest,clone)
+{
+	SGStringList<char> strings = generateRandomData();
+
+	CStringFeatures<char>* f=new CStringFeatures<char>(strings, ALPHANUM);
+	CStringFeatures<char>* f_clone = (CStringFeatures<char> *) f->clone();
+
+	for (index_t i=0; i<f->get_num_vectors(); ++i)
+	{
+		int32_t a_len;
```
vigsterkr: index_t

tests/unit/features/StringFeatures_unittest.cc
```diff
@@ -106,3 +108,30 @@ TEST(StringFeaturesTest,copy_subset)
...
+		int32_t b_len;
```
vigsterkr: index_t

src/shogun/base/Parameter.cpp
```diff
@@ -3874,7 +3874,7 @@ bool TParameter::copy(TParameter* target)
 					return false;
 				}
 
-				x=x+(m_datatype.sizeof_ptype());
+				x=x+(m_datatype.sizeof_stype());
```
lkuchenb: If this fix is right then this code never worked for vectors of non-primitive types?

src/shogun/base/Parameter.cpp
```diff
@@ -3844,7 +3844,7 @@ bool TParameter::copy(TParameter* target)
 				*(char**)target->m_parameter=SG_MALLOC(char, num_bytes);
 				/* check whether ptype is SGOBJECT, if yes we need to initialize
 				   the memory with NULL for the way copy_ptype handles it */
-				if (m_datatype.m_ptype==PT_SGOBJECT)
+				if (m_datatype.m_ptype==PT_SGOBJECT || m_datatype.m_stype==ST_STRING)
```
lkuchenb: The downstream function `TParameter::copy_stype()` checks `target_ptr->slen` - so this occasionally has strange effects if not initialized I think...

src/shogun/base/Parameter.cpp
karlnapf: this might well be. want to add a test? :dancing_men:

tests/unit/features/StringFeatures_unittest.cc
```diff
@@ -106,3 +108,30 @@ TEST(StringFeaturesTest,copy_subset)
 	SG_UNREF(f);
 	SG_UNREF(subset_copy);
 }
+
+TEST(StringFeaturesTest,clone)
+{
```
karlnapf: ah great, you are a star!

tests/unit/features/StringFeatures_unittest.cc
```diff
@@ -106,3 +108,30 @@ TEST(StringFeaturesTest,copy_subset)
...
+	CStringFeatures<char>* f_clone = (CStringFeatures<char> *) f->clone();
+
+	for (index_t i=0; i<f->get_num_vectors(); ++i)
+	{
+		index_t a_len;
+		index_t b_len;
+		bool a_free;
+		bool b_free;
+
+		char * a_vec = f->get_feature_vector(i, a_len, a_free);
```
karlnapf: no whitespace before *

tests/unit/features/StringFeatures_unittest.cc
```diff
@@ -106,3 +108,30 @@ TEST(StringFeaturesTest,copy_subset)
...
+		char * b_vec = f_clone->get_feature_vector(i, b_len, b_free);
+
+		EXPECT_EQ(a_len, b_len);
+		EXPECT_EQ(a_free, b_free);
+		for (index_t j = 0; j < a_len; ++j)
+			EXPECT_EQ(a_vec[j], b_vec[j]);
+	}
+
+	SG_UNREF(f);
+	SG_UNREF(f_clone);
```
karlnapf: did you valgrind it?

Issue #3537, opened 2016-11-14
Systematically test serialization of trained models


karlnapf:
We need to make sure that *all* Shogun can dump trained models to disc!

Basically, every learning algorithm in shogun, that is based on training/test (so starting with CMachine), should be trained on a (small) exampe dataset, then serialized, then de-serialized and made sure that it can be applied to test data.

This task is to automate that.

piyushgoel997:
Hi, I'm new here and wish to take up this issue. So, kindly guide me through.

karlnapf:
Well, write something that trains a models, serializes it, deserializes it, and then applying it to test data gives the same results. Then automatically make this happen for all models

piyushgoel997:
Actually I'm new to Open Source and know only the basics Machine Learning(with no experience in any project) so can you please explain to me in detail or assign me to an easier issue. I want to be involved in this project, so please guide me.

karlnapf:
Please read the "getting involved" guide in the wiki, about starting with Shogun. We can't hold your hands unfortunately. I suggest you pick something simpler if you don't know how to start here.

kris-singh:
@karlnapf i want to take this up.
Is serialisation module implemented or not. Or should we only implement the testing part of it.

karlnapf:
We have serialization working at the moment (at least in theory). The details under the hood will soon be replaced, but that doesn't matter for this task.

Try to come up with a simple way to test serialization / de-serialization of all models.

Algorithm for supervised methods (`CMachine` subclasses)

1. Load training data
2. Train model on training data
3. Serialize model to file
4. De-serialize model from file
5. Apply model to test data

Unsupervised methods and other will work similarly, but let's start with the supervised ones.

The best thing was if we were doing this automatically for modes of similar classes (supervised, unsupervised, kernel machine, gaussian process, etc)

You can start via writing a unit test for a single model, as a prototype. However, eventually, we want to use some template engine to generate unit tests for all implementations we have of a particular class. We did a similar thing for serialization itself [here](https://github.com/shogun-toolbox/shogun/tree/develop/tests/unit/io), using jinja2. I would start writing a simple test for say a `CLinearMachine` and then make it automatically be generated for all subclasses.
The cmake file that generates the tests using jinja2 is [here](https://github.com/shogun-toolbox/shogun/blob/develop/tests/unit/CMakeLists.txt)

micmn:
[...]

PR #3679, opened 2017-03-06
Trained models serialization test

micmn:
#3537


tests/unit/base/trained_model_serialization_unittest.cc.jinja2
```diff
@@ -0,0 +1,210 @@
...
+
+	/* predict with deserialized machine */
+	CLabels* deserialized_predictions=deserialized_model->apply(feat_test);
+//	SGVector<float64_t> deserialized_prediction_vector=deserialized_predictions->get_labels();
+
+	/* check whether the predictions are equal */
+//	float64_t accuracy=1e-14;
+//	for (index_t i=0; i<N_TEST; ++i)
+//		EXPECT_NEAR(prediction_vector[i], deserialized_prediction_vector[i], accuracy);
+
```
micmn: I commented out this because `CLabels` doesn't have `get_labels()`

tests/unit/base/trained_model_serialization_unittest.cc.jinja2
```diff
@@ -0,0 +1,210 @@
...
+#include <shogun/base/init.h>
+#include <shogun/features/DenseFeatures.h>
+#include <shogun/labels/Labels.h>
+#include <shogun/labels/BinaryLabels.h>
+#include <shogun/labels/RegressionLabels.h>
+#include <shogun/io/SerializableAsciiFile.h>
+#include <shogun/base/class_list.h>
+#include "utils/Utils.h"
+
+#include <shogun/classifier/AveragedPerceptron.h>
```
karlnapf: can this be automated, i.e. filled in via jinja as well?

tests/unit/base/trained_model_serialization_unittest.cc.jinja2
```diff
@@ -0,0 +1,210 @@
...
+#include <shogun/regression/LinearRidgeRegression.h>
+
+using namespace shogun;
+
+class TrainedModelSerializationTest : public testing::Test
+{
+	protected:
+		static void SetUpTestCase()
+		{
+			/* generate train and test data */
```
karlnapf: can we load datasets instead of generating them?

tests/unit/base/trained_model_serialization_unittest.cc.jinja2
```diff
@@ -0,0 +1,210 @@
...
+
+			SG_UNREF(feat_train_reg);
+			SG_UNREF(feat_test_reg);
+			SG_UNREF(label_train_reg);
+		}
+
+		static const index_t N_TRAIN=30;
+		static const index_t N_TEST=10;
+
+		static CDenseFeatures<float64_t>* feat_train_bin;
```
karlnapf: I think the feature pointers should be the same for regression and classification

tests/unit/base/trained_model_serialization_unittest.cc.jinja2
```diff
@@ -0,0 +1,210 @@
...
+		static CDenseFeatures<float64_t>* feat_test_bin;
+		static CBinaryLabels* label_train_bin;
+
+		static CDenseFeatures<float64_t>* feat_train_reg;
+		static CDenseFeatures<float64_t>* feat_test_reg;
+		static CRegressionLabels* label_train_reg;
+};
+
+CDenseFeatures<float64_t>* TrainedModelSerializationTest::feat_train_bin;
```
karlnapf: Not sure I like these globals....

tests/unit/base/trained_model_serialization_unittest.cc.jinja2
```diff
@@ -0,0 +1,210 @@
...
+
+CDenseFeatures<float64_t>* TrainedModelSerializationTest::feat_train_reg=NULL;
+CDenseFeatures<float64_t>* TrainedModelSerializationTest::feat_test_reg=NULL;
+CRegressionLabels* TrainedModelSerializationTest::label_train_reg=NULL;
+
+{% macro test(class, problem_type) -%}
+TEST_F(TrainedModelSerializationTest, {{class}})
+{
+{% if problem_type == 'binary' %}
+	CDenseFeatures<float64_t>* feat_train=feat_train_bin;
```
karlnapf: what about a factory for labels, given the `CMachine` instance?
Features are the same type, so no need to distinguish I think

tests/unit/base/trained_model_serialization_unittest.cc.jinja2
```diff
@@ -0,0 +1,210 @@
...
+	CDenseFeatures<float64_t>* feat_test=feat_test_bin;
+	CBinaryLabels* label_train=label_train_bin;
+{% elif problem_type == 'regression' %}
+	CDenseFeatures<float64_t>* feat_train=feat_train_reg;
+	CDenseFeatures<float64_t>* feat_test=feat_test_reg;
+	CRegressionLabels* label_train=label_train_reg;
+{% endif %}
+
+	/* instantiate and train the machine */
```
karlnapf: superfluous comment

tests/unit/base/trained_model_serialization_unittest.cc.jinja2
```diff
@@ -0,0 +1,210 @@
...
+	auto machine=new C{{class}}();
+	machine->set_features(feat_train);
+	machine->set_labels(label_train);
+	bool train_success=machine->train();
+	ASSERT_TRUE(train_success);
+
+	/* predict with trained machine */
```
karlnapf: superfluous comment

tests/unit/base/trained_model_serialization_unittest.cc.jinja2
```diff
@@ -0,0 +1,210 @@
...
+	CLabels* predictions=machine->apply(feat_test);
+//	SGVector<float64_t> prediction_vector=predictions->get_labels();
```
karlnapf: I think you can just use `equals`, so no need to actually access the labels. All done via polymorphism

tests/unit/base/trained_model_serialization_unittest.cc.jinja2
```diff
@@ -0,0 +1,210 @@
...
+
+	/* serialize the machine */
```
karlnapf: superfluous comment

tests/unit/base/trained_model_serialization_unittest.cc.jinja2
```diff
@@ -0,0 +1,210 @@
...
+
+	CSerializableAsciiFile *file=new CSerializableAsciiFile(filename, 'w');
+	bool save_success=machine->save_serializable(file);
+	file->close();
+	SG_UNREF(file);
+	ASSERT_TRUE(save_success);
+
+	/* deserialize the machine */
+	file=new CSerializableAsciiFile(filename, 'r');
+	auto deserialized_model=(C{{class}}*)new_sgserializable(class_name.c_str(), PT_NOT_GENERIC);
```
karlnapf: I think you can use jinja to instantiate using the default constructor here
The `new_sgserializable` is quite ugly inside, so I would like to avoid using it. And you know the classname anyways here

tests/unit/base/trained_model_serialization_unittest.cc.jinja2
```diff
@@ -0,0 +1,210 @@
...
+	ASSERT_TRUE(deserialized_model != NULL);
+	bool load_success=deserialized_model->load_serializable(file);
+	file->close();
+	SG_UNREF(file);
+	ASSERT_TRUE(load_success);
+
+	/* predict with deserialized machine */
```
karlnapf: no comments that are self-explained with the two following lines of code pls :)

tests/unit/base/trained_model_serialization_unittest.cc.jinja2
```diff
@@ -0,0 +1,210 @@
...
+	CLabels* deserialized_predictions=deserialized_model->apply(feat_test);
+//	SGVector<float64_t> deserialized_prediction_vector=deserialized_predictions->get_labels();
+
+	/* check whether the predictions are equal */
+//	float64_t accuracy=1e-14;
```
karlnapf: use `CSGObject::equals`

karlnapf:
General point for discussion:

@vigsterkr @lisitsyn 
So the `CMachine` instances currently do serialize the features, and de-serializing loads all of the features back into memory. I am not sure we really want this? For the linear machines in this PR, this is not even necessary: the regression/decision function is simply `f(w.T * x + b)` and `w` is stored explicitly
For kernel machines, we would have `f(x)=\sum_i \alpha_i * k(x, x_i)`, so we do actually need the data (though not all of it e.g. for SVMs). For now, I guess we would call `store_model_features`, and then only serialize those data which are needed in predictions. In general, it would be good to be able to have control over serializing data or not. What are your thoughts?

@micmn I think you can validate my above points on linear machines by removing the feature reference in the trained model before serializing.

vigsterkr:
@karlnapf i'm still not convinced that this should go in to the unit tests... and actually not because of the fact that this is going to take a while to run, but because of linking time of `shogun-unit-test`. try to see how it takes to link that with or without the generated tests (see serialization tests there). i'd actually be happy to move all these slow and long linking time generated stuff into a another executable... any other ideas?

karlnapf:
@vigsterkr I see. I mean it will be much less than the serialization tests (way less classes). But good point. Can't we just turn the auto-generated stuff off via cmake switch (for local developing?)

Question also is whether we want to run this on CI or just buildbot

vigsterkr:
@karlnapf this cmake switch story is getting out of hand... we have so many cmake switches that i rather just build a separate binary and if you dont want it just dont build it. or if you just want really unit tests then you do `make shogun-unit-test` ... and if you want the serialization tests then you generate that one.... and run that one...

karlnapf:
@vigsterkr ok agreed
@micmn you have an idea of what we mean?

micmn:
> I think you can validate my above points on linear machines by removing the feature reference in the trained model before serializing.

You mean something like `machine->set_features(NULL); machine->set_labels(NULL);`?
In this case the tests fail, even `LeastSquaresRegression` fails with `283: [ERROR] In file /home/shogun/src/shogun/machine/LinearMachine.cpp line 81: Specified features are not of type CDotFeatures`, I didn't investigate though.

> you have an idea of what we mean?

Sure, I'll look into that but since I'm just getting started with CMake I may need some suggestions.

So, I tried to automatize the detection of machine classes traversing the headers with regex matching and building a sort of classes hierarchy, don't know if this is a good idea, if there is some cleaner way I'd appreciate suggestions.
In the template there is also a blacklist containing LinearMachines for which the test fails for various reasons.

vigsterkr:
@micmn currently this is how we populate which files to be used for creating the huge `shogun-unit-test` executable:
https://github.com/shogun-toolbox/shogun/blob/develop/tests/unit/CMakeLists.txt#L82

now what we would like to have is that actually you create another executable, just like here:
https://github.com/shogun-toolbox/shogun/blob/develop/tests/unit/CMakeLists.txt#L84

only difference that there's a defined set of files one would like to have compiled for that (let's call it set A). 

and then basically the sources for shogun-unit-test = {all unit test sources - set A}

karlnapf:
@micmn Did you set features and labels to zero *after* training?

micmn:
Yes, see https://github.com/shogun-toolbox/shogun/pull/3679/files#diff-856555dd8e1fe665a77a0707b995f058R73

karlnapf:
So there are a few things about this failure:

 * For linear machines, resetting features and labels after training **has** to work. If it doesn't, this is a bug (and one of the reasons why we want to test this is finding those bugs!)
 * The error your described: `[ERROR] In file /home/shogun/src/shogun/machine/LinearMachine.cpp line 81: Specified features are not of type CDotFeatures` seems weird because the test features you provide in the test are `CDenseFeatures` which are indeed a subclass of `CDotFeatures`. This should not fail on dense features and needs investigation (separate PR!)
 * If you check the code in `CLinearMachine::apply_get_outputs`, you will see that the `features` field is overriden with the object passed. So everything should work once the above bug is resolved

tests/unit/base/trained_model_serialization_unittest.cc.jinja2
```diff
@@ -0,0 +1,136 @@
...
+
+	load_data(machine);
+	machine->set_features(feat_train);
+	machine->set_labels(label_train);
+
+	bool train_success=machine->train();
+	ASSERT_TRUE(train_success);
+
+	CLabels* predictions=machine->apply(feat_test);
+//	machine->set_features(NULL);
```
karlnapf: put a comment here: "to avoid serialization of the data"

tests/unit/base/trained_model_serialization_unittest.cc.jinja2
```diff
@@ -0,0 +1,136 @@
...
+	file=new CSerializableAsciiFile(filename, 'r');
+	auto deserialized_machine=new C{{class}}();
+	bool load_success=deserialized_machine->load_serializable(file);
+	file->close();
+	SG_UNREF(file);
+	ASSERT_TRUE(load_success);
+
+	CLabels* deserialized_predictions=deserialized_machine->apply(feat_test);
+
+	/* check whether the predictions are equal */
```
karlnapf: superfluous comment

tests/unit/base/trained_model_serialization_unittest.cc.py
```diff
@@ -0,0 +1,96 @@
+#!/usr/bin/env python
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 3 of the License, or
+# (at your option) any later version.
+#
+
+def find_guard(content):
+    guards = ['USE_CPLEX', 'USE_GPL_SHOGUN', 'HAVE_LAPACK', 'USE_SWIG_DIRECTORS']
```
karlnapf: these guards should not be hard coded, but rather come from the cmake setup. 
In fact, this whole thing should not be done in a python script I think
@vigsterkr will have comments on how to organise this

tests/unit/base/trained_model_serialization_unittest.cc.py
```diff
@@ -0,0 +1,96 @@
...
+        b = attr['base']
+        while b is not None:
+            attr['ancestors'].append(b)
+            if classes.has_key(b):
+                b = classes[b]['base']
+            else:
+                break
+
+
+def get_classes(basedir="."):
```
karlnapf: check our class_list thing -- it does something very similar

micmn:
> The error your described: [ERROR] In file /home/shogun/src/shogun/machine/LinearMachine.cpp line 81: Specified features are not of type CDotFeatures seems weird because the test features you provide in the test are CDenseFeatures which are indeed a subclass of CDotFeatures. This should not fail on dense features and needs investigation (separate PR!)

I realized this is actually my fault, I didn't notice that `machine->apply(feat_test)` internalizes the features and `set_features(NULL)` was after that call. So moving `apply()` before `set_features()` solves the problem.

karlnapf:
Cool so we dont actually need to serialise the data, which is great!

karlnapf:
Hey hey, any progress here?

micmn:
@karlnapf Hey, I was waiting for some comment on this:
> these guards should not be hard coded, but rather come from the cmake setup.
In fact, this whole thing should not be done in a python script I think
@vigsterkr will have comments on how to organise this

Anyway, should I open a separate PR for the ASCII/XML/HDF5/JSON-serialization test binary?

karlnapf:
yeah a sep PR would be good. Just suggest something sensible and then we can discuss around that

micmn:
I've integrated the hierarchy construction in `class.cpp.py` so it also creates a json containing shogun's classes with some attributes (base class, abstract/template/...). There are a few things than be done, one is refactor `clone_unittest.cc.py` and `trained_model_serialization_unittest.cc.py`, but let me know what you think of this approach.

karlnapf:
BTW we dont need to test this for other serialization types. This is about something else, serialization itself is tested elsewhere

karlnapf:
Ok this is great!
Can we have a seperate PR for the class_list changes and the json stuff?
Way easier to review ...

micmn:
#3750 #3751

PR #3684, opened 2017-03-10
Removed copy_feature_matrix method from DenseFeatures

lambday:
@karlnapf turns out, clone was required (one unit-test failed otherwise). So, put the cloning responsibility in the caller.

### DIFF
--- a/src/shogun/clustering/KMeans.cpp
+++ b/src/shogun/clustering/KMeans.cpp
@@ -43,7 +43,7 @@ void CKMeans::Lloyd_KMeans(SGMatrix<float64_t> centers, int32_t num_centers)
 {
 	CDenseFeatures<float64_t>* lhs=
 		CDenseFeatures<float64_t>::obtain_from_generic(distance->get_lhs());
-	
+
 	int32_t lhs_size=lhs->get_num_vectors();
 	int32_t dim=lhs->get_num_features();
 
@@ -71,7 +71,8 @@ void CKMeans::Lloyd_KMeans(SGMatrix<float64_t> centers, int32_t num_centers)
 				   	Terminating. \n", iter)
 
 		changed=0;
-		rhs_mus->copy_feature_matrix(centers);
+		rhs_mus->set_feature_matrix(centers.clone());
+		rhs_mus->initialize_cache();
 
 		distance->precompute_rhs();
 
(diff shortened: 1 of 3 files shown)


src/shogun/clustering/KMeans.cpp
```diff
@@ -71,7 +71,8 @@ void CKMeans::Lloyd_KMeans(SGMatrix<float64_t> centers, int32_t num_centers)
 				   	Terminating. \n", iter)
 
 		changed=0;
-		rhs_mus->copy_feature_matrix(centers);
+		rhs_mus->set_feature_matrix(centers.clone());
```
karlnapf: so it does need to be cloned in the end?

src/shogun/clustering/KMeans.cpp
```diff
@@ -71,7 +71,8 @@ void CKMeans::Lloyd_KMeans(SGMatrix<float64_t> centers, int32_t num_centers)
...
+		rhs_mus->initialize_cache();
```
karlnapf: this is a new call?

karlnapf: This shouldnt be called in the setter?

src/shogun/clustering/KMeans.cpp
lambday: No clue. Without the clone, one of the unit-test fails. So kept the behavior just the same as before. Maybe Saurabh can fill this in.

src/shogun/clustering/KMeans.cpp
lambday: Well, set_feature_matrix was already an existing method. Didn't change its behavior.

lambday:
@karlnapf merging this one then. 

The white-spaces are hard to keep track of. You see, I don't manually do those changes, but my `.vimrc` setup removes all the trailing ones for me whenever it writes the buffer (it is based on the `shogun/doc/wiki/Code-style.md` BTW).

karlnapf:
In the future, pls put whitespace changes to a separate commit. We ask external contributors to do this, so we do it as well. Your vim should not change existing code in my eyes when you save a file.

lambday:
@karlnapf okay.. will use a different vimrc for shogun. not sure whether sending a separate patch for only the white-space changes will ever happen though :P would be much better if we had a hook that does it automatically at the time of merging, and hence preventing them to exist in the first place.. but I don't (yet) know how to do that..

karlnapf:
It is quite easy to put them in a separate commit. When you look at the diff and you realise there are whitespace changes, just undo your actual changes, save, commit, re-do the changes, commit again.
Just 1 min of work for you and makes things so much cleaner.

PR #3685, opened 2017-03-11
SGMatrix cleanups and speedups

lambday:
Should be much faster.

### DIFF
--- a/src/shogun/lib/SGMatrix.cpp
+++ b/src/shogun/lib/SGMatrix.cpp
@@ -4,6 +4,8 @@
  * the Free Software Foundation; either version 3 of the License, or
  * (at your option) any later version.
  *
+ * Written (W) 2016-2017 Pan Deng
+ * Written (W) 2013-2017 Soumyajit De
  * Written (W) 2011-2013 Heiko Strathmann
  * Written (W) 2012 Fernando Jose Iglesias Garcia
  * Written (W) 2010,2012 Soeren Sonnenburg
@@ -19,10 +21,11 @@
 #include <shogun/mathematics/Math.h>
 #include <shogun/mathematics/lapack.h>
 #include <limits>
-
+#include <algorithm>
 #include <shogun/mathematics/eigen3.h>
 
-namespace shogun {
+namespace shogun
+{
 
 template <class T>
 SGMatrix<T>::SGMatrix() : SGReferencedData()
@@ -56,7 +59,7 @@ template <class T>
 SGMatrix<T>::SGMatrix(index_t nrows, index_t ncols, bool ref_counting)
 	: SGReferencedData(ref_counting), num_rows(nrows), num_cols(ncols), gpu_ptr(nullptr)
 {
-	matrix=SG_MALLOC(T, ((int64_t) nrows)*ncols);
+	matrix=SG_CALLOC(T, ((int64_t) nrows)*ncols);
 	m_on_gpu.store(false, std::memory_order_release);
 }
 
@@ -81,6 +84,7 @@ SGMatrix<T>::SGMatrix(SGVector<T> vec, index_t nrows, index_t ncols)
 	REQUIRE(vec.vlen==nrows*ncols, "Number of elements in the matrix (%d) must "
 			"be the same as the number of elements in the vector (%d)!\n",
 			nrows*ncols, vec.vlen);
+
 	matrix=vec.vector;
 	num_rows=nrows;
 	num_cols=ncols;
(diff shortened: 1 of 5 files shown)


karlnapf:
Travis canceled?

lambday:
@karlnapf some more stuffs coming up with this PR :)

lambday:
@karlnapf Summary of changes:

- `SGMatrix::operator==` is now (potentially) inlined
- `SGMatrix::equals` now works with all floating ptypes as well as complex types
- `SGMatrix::equals` now do not look for element-wise similarity when comparing with itself
- `SGMatrix::clone` does a `std::copy` instead of an element-wise copy
- `SGMatrix::set_const` and `SGMatrix::is_symmetric` now throw user friendly `REQUIRE` errors when trying to read from/write into unallocated/insufficient memory
- `SGMatrix::max_single` uses an off-the-shelf algorithm to find the max-value.
- made several read-only methods const-qualified in `SGMatrix`

src/shogun/multiclass/MCLDA.cpp
```diff
@@ -224,7 +224,6 @@ bool CMCLDA::train_machine(CFeatures* data)
 	int32_t iX = 0;
 
 	m_means.zero();
-	m_cov.zero();
```
lambday: was trying to set zero to an uninitialized memory.

src/shogun/neuralnets/NeuralNetwork.cpp
```diff
@@ -470,7 +470,10 @@ float64_t CNeuralNetwork::compute_gradients(SGMatrix<float64_t> inputs,
 	forward_propagate(inputs);
 
 	for (int32_t i=0; i<m_num_layers; i++)
-		get_layer(i)->get_activation_gradients().zero();
+	{
+		if (!get_layer(i)->is_input())
+			get_layer(i)->get_activation_gradients().zero();
```
lambday: was trying to set zero to uninitialized memory when `actiavtion_gradients` was not-initialized (happens when the layer is not an input).

tests/unit/lib/SGMatrix_unittest.cc
```diff
@@ -519,3 +519,64 @@ TEST(SGMatrixTest, from_eigen3)
...
 		EXPECT_EQ(eigen_mat(i), sg_mat[i]);
 }
+
+TEST(SGMatrixTest, equals)
+{
+	const index_t size=10;
+	SGMatrix<float32_t> mat, copy;
+
+	EXPECT_TRUE(mat.equals(mat));
+	EXPECT_TRUE(mat.equals(copy));
```
lambday: two un-initialized matrices are equal to one another.

tests/unit/lib/SGMatrix_unittest.cc
```diff
@@ -519,3 +519,64 @@ TEST(SGMatrixTest, from_eigen3)
...
+
+	mat=SGMatrix<float32_t>(size, size);
+	CMath::init_random(100);
+	for (uint64_t i=0; i<mat.size(); ++i)
+		mat.matrix[i]=CMath::randn_float();
+
+	EXPECT_TRUE(mat.equals(mat));
+	EXPECT_FALSE(mat.equals(copy));
+
+	copy=SGMatrix<float32_t>(size, size);
```
lambday: just allocating the size doesn't mean we're going to read from the memory. segfault otherwise.

vigsterkr:
@lambday could we use std::memcpy instead of std::copy?

vigsterkr:
@lambday and actually could you wrap std::memcpy with shogun::memcpy in memory.h and use that? just to be able to easily plug in a different memcpy method....

src/shogun/lib/SGMatrix.h
```diff
@@ -215,15 +215,22 @@ template<class T> class SGMatrix : public SGReferencedData
 		}
 
 		/** Check for pointer identity */
-		bool operator==(SGMatrix<T>& other);
+		bool operator==(const SGMatrix<T>& other) const
```
vigsterkr: if you wanna get this inlined use `SG_FORCED_INLINE`, like:
```
SG_FORCED_INLINE bool operator==(const SGMatrix<T>& other) const
```

src/shogun/lib/SGMatrix.cpp
```diff
@@ -229,49 +257,48 @@ bool SGMatrix<complex128_t>::is_symmetric()
...
 template <class T>
 T* SGMatrix<T>::clone_matrix(const T* matrix, int32_t nrows, int32_t ncols)
 {
-	T* result = SG_MALLOC(T, int64_t(nrows)*ncols);
-	for (int64_t i=0; i<int64_t(nrows)*ncols; i++)
-		result[i]=matrix[i];
-
+	auto size=int64_t(nrows)*ncols;
+	T* result=SG_MALLOC(T, size);
+	std::copy(matrix, matrix+size, result);
```
vigsterkr: as previously mentioned could you wrap this with:
```
shogun::memcopy(...)
```
and implement that in `src/lib/memory.h`

lambday: yeah, nice idea. `std::memcpy` and `std::copy` has pretty much same performance in my experience but with this shogun wrapper, we can try different things. I'll do that.

vigsterkr: @lambday i've seen mostly memcpy performing better than copy... (depends of course on the libc and compiler...)

vigsterkr: btw maybe `shogun::memcpy` is better ... just to be consistent

src/shogun/lib/SGMatrix.h
lambday: Thanks. Will add that.

src/shogun/lib/SGMatrix.cpp
lambday: yeah makes sense. Should also make it templated I think. When using `std::memcpy` in the backend, will cast to `void*`, but then the backend can easily be replaced with `std::copy` as long as it is templated.

Should this be in header and inlined BTW or we include `memory.h` in `SGObject` headers and things will blow up?

vigsterkr: yeye go ahead with templating it!
header and inlined definitely... :)

lambday: Cool. Also, ideally, I think we should have 3 such methods
```c++
shogun::memcpy(...)         // no overlapping - fastest of all
shogun::copy_forward(...)   // works with overlapping ..[s_first..d_first..s_last)..d_last
shogun::copy_backward(...)  // works with overlapping ..d_first..[s_first..d_last..s_last)
```
Let me know what you think. well, maybe just one for now would suffice..

vigsterkr: sounds like a plan but i would do this like on-demand... so implement create those wrappers when they are actually required.... or do you use somewhere copy_forward and copy_backward?

lambday: actually, `copy_backward` thing can be useful if we want to do dense feature subset in-place.

vigsterkr: again if you have a use-case for it go ahead... meaning if you already use it somewhere then add it ...

src/shogun/lib/SGMatrix.cpp
```diff
@@ -39,7 +41,9 @@ SGMatrix<T>::SGMatrix(bool ref_counting) : SGReferencedData(ref_counting)
 template <class T>
 SGMatrix<T>::SGMatrix(T* m, index_t nrows, index_t ncols, bool ref_counting)
 	: SGReferencedData(ref_counting), matrix(m),
-	num_rows(nrows), num_cols(ncols) { }
+	num_rows(nrows), num_cols(ncols)
```
karlnapf: whitespace changes to separate commit pls
The patch is very large already and this makes it harder to parse

src/shogun/lib/SGMatrix.cpp
```diff
@@ -99,58 +103,96 @@ SGMatrix<T>::~SGMatrix()
...
+		});
 }
 
 template <class T>
 void SGMatrix<T>::set_const(T const_elem)
 {
-	for (int64_t i=0; i<int64_t(num_rows)*num_cols; i++)
-		matrix[i]=const_elem ;
+	REQUIRE(matrix!=nullptr, "The underlying matrix is null!\n");
+	REQUIRE(num_rows>0, "Number of rows is non-positive!\n");
```
karlnapf: why this? Also should print the number

src/shogun/lib/SGMatrix.cpp
```diff
@@ -159,65 +201,51 @@ bool SGMatrix<T>::is_symmetric()
...
+	\
+	if (num_rows!=num_cols)	\
+		return false;	\
+	\
+	for (int i=0; i<num_rows; ++i)	\
+	{	\
+		for (int j=i+1; j<num_cols; ++j)	\
+		{	\
+			if (!CMath::fequals<real_t>(matrix[j*num_rows+i],	\
+						matrix[i*num_rows+j], std::numeric_limits<real_t>::epsilon()))	\
```
karlnapf: I hope this stuff is portable?

src/shogun/lib/SGMatrix.cpp
```diff
@@ -159,65 +201,51 @@ bool SGMatrix<T>::is_symmetric()
...
+bool SGMatrix<real_t>::is_symmetric() const	\
+{	\
+	REQUIRE(matrix!=nullptr, "The underlying matrix is null!\n");	\
+	REQUIRE(num_rows>0, "Number of rows is non-positive!\n");	\
+	REQUIRE(num_cols>0, "Number of cols is non-positive!\n");	\
+	\
+	if (num_rows!=num_cols)	\
+		return false;	\
+	\
+	for (int i=0; i<num_rows; ++i)	\
```
karlnapf: pls dont use "int"

src/shogun/lib/SGMatrix.cpp
```diff
@@ -229,49 +257,48 @@ bool SGMatrix<complex128_t>::is_symmetric()
...
-	{
-		if (matrix[i]>max)
-			max=matrix[i];
-	}
+	REQUIRE(matrix!=nullptr, "The underlying matrix is null!\n");
+	REQUIRE(num_rows>0, "Number of rows is non-positive!\n");
+	REQUIRE(num_cols>0, "Number of cols is non-positive!\n");
 
-	return max;
+	return *std::max_element(matrix, matrix+size());
```
karlnapf: ah yes good

src/shogun/lib/SGMatrix.cpp
```diff
@@ -229,49 +257,48 @@ bool SGMatrix<complex128_t>::is_symmetric()
...
 }
 
 template <class T>
-SGMatrix<T> SGMatrix<T>::clone()
+SGMatrix<T> SGMatrix<T>::clone() const
 {
-	return SGMatrix<T>(clone_matrix(matrix, num_rows, num_cols),
-			num_rows, num_cols);
+	REQUIRE(matrix!=nullptr, "The underlying matrix is null!\n");
+	REQUIRE(num_rows>0, "Number of rows is non-positive!\n");
```
karlnapf: These copy paste messages all need clean up, wont comment the other ones

src/shogun/lib/SGMatrix.h
```diff
@@ -215,15 +215,22 @@ template<class T> class SGMatrix : public SGReferencedData
...
+		{
+			if (matrix!=other.matrix || num_rows!=other.num_rows || num_cols!=other.num_cols)
+				return false;
+
+			return true;
+		}
 
 		/** Operator overload for element-wise matrix comparison.
-		 * Note that only numerical data is compared
+		 * Note that only numerical data is compared. Works for floating
```
karlnapf: What is this sentence supposed to mean?

src/shogun/lib/SGMatrix.cpp
vigsterkr: @lambday ok so i guess this needs rebase & replacing here the std::copy and then it's good to go!

vigsterkr: @lambday for future reference: if you would be doing this in a feature branch of shogun repo i could do these changes and merge myself the feature branch... but like this unfortunately i cannot... the only option would be to cherry pick your repo's feature branch ....... so yeah please since you have write access on the main repo just do the feature branch within that.

lambday:
CIs are done and green. Merging this one.

PR #3689, opened 2017-03-11
Add CLabels::get_subset_stack()

lkuchenb:
Was available in Features, not in Labels. As far as I can tell it's the only interface to figure out if subsets are currently applied.

### DIFF
--- a/src/shogun/labels/Labels.h
+++ b/src/shogun/labels/Labels.h
@@ -96,6 +96,11 @@ public:
 	 * Calls subset_changed_post() afterwards */
 	virtual void remove_all_subsets();
 
+	/**
+	 * @return subset stack
+	 */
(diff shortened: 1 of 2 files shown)


src/shogun/labels/Labels.h
```diff
@@ -96,6 +96,12 @@ class CLabels : public CSGObject
 	 * Calls subset_changed_post() afterwards */
 	virtual void remove_all_subsets();
 
+	/** returns subset stack
```
karlnapf: No need for a comment, just @return is fine

karlnapf: Question: Does it need to be virtual? I know the other one is, but are there cases where we want this? And ... const?

lkuchenb:
Fixed doxygen comment, leaving the function signature as it is as discussed in chat

Issue #3687, opened 2017-03-11
Remove `apply_locked*` methods from SWIG interfaces


karlnapf:
E.g. `CGaussianNaiveBayes::apply_locked_binary (SGVector< index_t > indices)`

Locked evaluation is an internal speedup hack for cross-validation is never used from SWIG interfaces.

Simply surround by (c.f. `SGVector` for example)

```
#ifndef SWIG // SWIG should skip this part
...
#endif // SWIG // SWIG should skip this part
```

Nice and simple entrance task for GSoC

PR #3690, opened 2017-03-11
Remove apply_locked* methods from SWIG interfaces #3687

tdjogi010:


karlnapf:
All this should be done in `CMachine.h`

Also, `virtual bool train_locked(SGVector<index_t> indices)` should also be guarded

tdjogi010:
@karlnapf what about the apply_lock* methods in other classes(like CKernelMachine, CDirectorKernelMachine,etc)? Should they not be guarded as well? (or is it because they are not included in swig interfaces already since -fvirtual is used with swig?)

karlnapf:
I actually don't know. Would be good to check this out....

karlnapf:
Thanks for the patch!
For the record, I checked the IPython notebooks, and none of them use the methods, so merging

tdjogi010:
Thanks!

Issue #3683, opened 2017-03-10
Memcheck KMeans


lambday:
When ran locally, doing a memcheck on the KMeans unit test suite results the following
```
[lambday@lambday.iitb.ac.in debug]$ ctest -I 23,23 -D ExperimentalMemCheck
   Site: lambday.iitb.ac.in
   Build name: Linux-ccache
Memory check project /home/lambday/Downloads/shogun/build/debug
    Start 23: unit-KMeans
1/1 MemCheck #23: unit-KMeans ......................***Failed   19.39 sec

0% tests passed, 1 tests failed out of 1

Total Test time (real) =  19.87 sec

The following tests FAILED:
	 23 - unit-KMeans (Failed)
-- Processing memory checking output: 
Memory checking results:
Uninitialized Memory Read - 1
        Errors while running CTest
```
Here is some more details
```
==13166== Memcheck, a memory error detector
==13166== Copyright (C) 2002-2013, and GNU GPL'd, by Julian Seward et al.
==13166== Using Valgrind-3.10.1 and LibVEX; rerun with -h for copyright info
==13166== Command: ./bin/shogun-unit-test --gtest_filter=KMeans* --suppressions=../../configs/valgrind.supp
==13166== 
Note: Google Test filter = KMeans*
[==========] Running 6 tests from 1 test case.
[----------] Global test environment set-up.
[----------] 6 tests from KMeans
[ RUN      ] KMeans.manual_center_initialization_test
[       OK ] KMeans.manual_center_initialization_test (879 ms)
[ RUN      ] KMeans.KMeanspp_center_initialization_test
[       OK ] KMeans.KMeanspp_center_initialization_test (446 ms)
[ RUN      ] KMeans.minibatch_training_test
==13166== Invalid read of size 8
==13166==    at 0x7150ECA: shogun::CEuclideanDistance::compute(int, int) (EuclideanDistance.cpp:77)
==13166==    by 0x714F8F0: shogun::CDistance::distance(int, int) (Distance.cpp:216)
==13166==    by 0x70E96BD: shogun::CKMeansMiniBatch::minibatch_KMeans() (KMeansMiniBatch.cpp:99)
==13166==    by 0x70E9BD6: shogun::CKMeansMiniBatch::train_machine(shogun::CFeatures*) (KMeansMiniBatch.cpp:159)
==13166==    by 0x7625C4D: shogun::CMachine::train(shogun::CFeatures*) (Machine.cpp:57)
[...]

PR #3691, opened 2017-03-12
Fix KMeansMiniBatch memcheck error

micmn:
See #3683


lambday:
Shall I merge it then?

lambday:
@karlnapf the error was in fact caught by mem-checking one of the unit-tests. Similar errors are caught in #3685 much more easily when I added a `REQUIRE` in `SGMatrix::set_const` (please check the neural network related file changes).. Some of the underlying methods are not guarded properly.. So until shogun base/lib is properly unit-tested, it's hard to catch them in applications such as this :(

Issue #3509, opened 2016-11-02
Implement a nice progress bar


karlnapf:
We currently have SG_PROGRESS in Shogun.

This task is to:

 * implement progress in all main outer loops
 * think about multi-core cases
 * implement with interactive modular langauges, e.g. https://github.com/tqdm/tqdm

First figure out how SG_PROGRESS works, try to use it, and then go from there....
*All* methods in the docs should have a progress...

geektoni:
@karlnapf 
About this one, would it be better a new implementation that will substitute SG_PROGRESS or an "upgrade" of SG_PROGRESS to make it also show a progress bar?

Btw, the current ```SGIO::progress``` implementation is faulty because in some cases it will display completely wrong timing information (I tried to use it with the cpp KNN meta example and the results were somewhat surprising ;) )

karlnapf:
Steps:

1. Use SG_PROGRESS in cases where it is not used, i.e. all outer loops where it makes sense. This is more a task of identifying methods that dont have any progress at all (x-validation, grid-search, gradient model selection, etc)
2. Fix behaviour of SG_PROGRESS if wrong
3. If there are design short-comings in SG_PROGRESS, discuss/suggest an improvement. Investigate how other libs do this, check if we can adapt ideas. Potentially refactor all of SG_PROGRESS uses in Shogun

All steps are independent but I think this is a good order of approaching them

karlnapf:
LeastAngleRegression needs one

vigsterkr:
this is dooone :)

PR #3693, opened 2017-03-13
Add SG_PROGRESS to some evaluation methods.

geektoni:
See #3509

### DIFF
--- a/src/shogun/evaluation/ClusteringAccuracy.cpp
+++ b/src/shogun/evaluation/ClusteringAccuracy.cpp
@@ -23,6 +23,9 @@ float64_t CClusteringAccuracy::evaluate(CLabels* predicted, CLabels* ground_trut
 	int32_t correct=0;
 	for (int32_t i=0; i < predicted_ilabels.vlen; ++i)
 	{
+		/* print progress */
+		SG_PROGRESS(i, 0, predicted_ilabels.vlen, 1, "EVALUATING ACCURACY: ");
+
 		if (predicted_ilabels[i] == groundtruth_ilabels[i])
 			correct++;
 	}
--- a/src/shogun/evaluation/ClusteringMutualInformation.cpp
+++ b/src/shogun/evaluation/ClusteringMutualInformation.cpp
@@ -54,6 +54,9 @@ float64_t CClusteringMutualInformation::evaluate(CLabels* predicted, CLabels* gr
 	float64_t mutual_info = 0;
 	for (index_t i=0; i < n_class; ++i)
 	{
+		/* print progress */
+		SG_PROGRESS(i, 0, n_class+n_class, 1, "EVALUATING MUTUAL INFORMATION: ");
+
 		for (index_t j=0; j < n_class; ++j)
 		{
 			if (G(i, j) != 0)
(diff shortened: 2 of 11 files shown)


lisitsyn:
One may add a wrapper for https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/base/range.h (something like RangeWithProgress?) that tracks it 'automatically'.

vigsterkr:
and let's make it totally YOLO! what about adding 💃 icon as a progress bar? :>
ok joking... but we could use utf8 chars for actually doing some progress update instead of oldschool ascii art ;)

geektoni:
I didn't see that file! Thanks @lisitsyn :) It should be a better way to proceed instead of tweaking SG_PROGRESS (much more C++11-ish).

@vigsterkr regarding UTF-8, the only limit is imagination I suppose ;)

vigsterkr:
@geektoni yeah let's template the progress character!  👍 @karlnapf 💃 is coming!

vigsterkr:
mmm actually is utf8 in src allowed....? :) or that's a java feature :D

geektoni:
@vigsterkr Representing utf-8 chars in src (something like ```cout << '\u0444'```) should be allowed. It is also fine to store them into a string (e.g ```std::string a = "\u0444"```) and then print that string. These two methods work fine on Linux terminals (I don't know on windows :/ ).

vigsterkr:
@geektoni should be fine then... in worst case we break wind0ze :)

src/shogun/evaluation/ClusteringAccuracy.cpp
```diff
@@ -23,6 +23,9 @@ float64_t CClusteringAccuracy::evaluate(CLabels* predicted, CLabels* ground_trut
 	int32_t correct=0;
 	for (int32_t i=0; i < predicted_ilabels.vlen; ++i)
 	{
+		/* print progress */
+		SG_PROGRESS(i, 0, predicted_ilabels.vlen, 1, "EVALUATING ACCURACY: ");
```
karlnapf: mmmh I am not too sure about progress bars in evaluation methods, this will take less 1s for every case that your memory can handle ;)

And *actually* this should be vectorized, i.e. shouldnt even be a loop ;)

src/shogun/evaluation/ClusteringMutualInformation.cpp
```diff
@@ -54,6 +54,9 @@ float64_t CClusteringMutualInformation::evaluate(CLabels* predicted, CLabels* gr
 	float64_t mutual_info = 0;
 	for (index_t i=0; i < n_class; ++i)
 	{
+		/* print progress */
+		SG_PROGRESS(i, 0, n_class+n_class, 1, "EVALUATING MUTUAL INFORMATION: ");
```
karlnapf: These strings `EVALUATING MUTUAL INFORMATION: `

Can't we use pretty function for that? Not so sure we should write this manually every time. Should be the `CClassName::method_name` or so

@vigsterkr @lisitsyn

src/shogun/evaluation/ClusteringAccuracy.cpp
geektoni: By _vectorized_ you mean "vectorized by" the compiler (using something like `-ftree-vectorize`), right? I understand your point. Maybe the progress bar can be more useful on longer evaluation methods (I suppose the only candidate here is cross-validation).

karlnapf: Yeah I mean like accuracy is just an average of a vector difference so there is no loop needed for doing that but we can just use a linalg or eigen3 call which will SIMD it

geektoni:
Mmmh I think this patch has gone useless (apart from the progress bar discussion). Shall I close it @karlnapf?

karlnapf:
Yep, thanks for pointing out

PR #3694, opened 2017-03-13
Fix wrong behaviour of SGIO::progress

geektoni:
See #3509.

**SG_PROGRESS current behaviour**:

previous_behaviour

**SG_PROGRESS patch behaviour**:

new_behaviour

### DIFF
--- a/src/shogun/io/SGIO.cpp
+++ b/src/shogun/io/SGIO.cpp
@@ -54,7 +54,7 @@ char SGIO::directory_name[FBUFSIZE];
 
 SGIO::SGIO()
 : target(stdout), last_progress_time(0), progress_start_time(0),
-	last_progress(1), show_progress(false), location_info(MSG_NONE),
+	last_progress(-1), show_progress(false), location_info(MSG_NONE),
 	syntax_highlight(true), loglevel(MSG_WARN)
 {
 	m_refcount = new RefCount();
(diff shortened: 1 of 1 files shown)


src/shogun/io/SGIO.cpp
```diff
@@ -181,8 +181,18 @@ void SGIO::progress(
 		v=CMath::clamp(v,1e-5,100.0);
 		last_progress = v-1e-6;
 
-		if ((v!=100.0) && (runtime - last_progress_time<0.5))
+		if ((v!=100.0) && (runtime - last_progress_time<0.5)) {
```
karlnapf: style: we do { in a new line

src/shogun/io/SGIO.cpp
```diff
@@ -181,8 +181,18 @@ void SGIO::progress(
...
+
+			// This is made to display correctly the percentage
+			// if the algorithm execution is too fast
+			if (current_val >= max_val-1){
```
karlnapf: style

geektoni:
Style fixed.

karlnapf:
Do we actually have a unit test for this thing? Just to ensure nothing crashes?

geektoni:
As far as I know, the entire SGIO has no unit tests, which means no, there is no unit test to ensure the correctness of the code. Shall I spend some time creating one just to be sure it works as before?

karlnapf:
I think a test that at least executes the stuff you added would be good. Correctness is tricky as it just prints out. Not sure it is worth the hassle to check this. But making sure it doesnt crash on corner cases would be useful definitely :)

karlnapf:
This one is fine to merge from my side

Issue #3700, opened 2017-03-15
Refactoring for `shogun::memcpy`


lambday:
We recently introduced a wrapper `shogun::memcpy` in `shogun/lib/memory.h` to be used instead of `memcpy` calls in Shogun. This method in turn calls `std::memcpy` as of now. But the goal was to be able to change this to a different implementation later, if possible, that instantly applies Shogun-wise without touching the rest of Shogun.

So, the task is to replace all existing `memcpy` calls with this newly introduced `shogun::memcpy`. However, in order to avoid possible errors due to namespaces, we'd like to rename our wrapper method to `sg_memcpy` first.

Very simple `find`/`sed` task for on boarders :)

lambday:
Done. Closing.

PR #3709, opened 2017-03-15
Sg memcpy

beautifulsoup4:
Fix issue #3700 
https://github.com/shogun-toolbox/shogun/commit/f269f25a5d565bf035e8421eeeeac0b22c4e8dac

### DIFF
--- a/src/shogun/lib/memory.h
+++ b/src/shogun/lib/memory.h
@@ -98,9 +98,9 @@ template <class T> void sg_generic_free(T* ptr)
 }
 
 template <class InputIt, class OutputIt>
-SG_FORCED_INLINE void* memcpy(InputIt dest, OutputIt src, size_t count)
+SG_FORCED_INLINE void* sg_memcpy(InputIt dest, OutputIt src, size_t count)
 {
-	return std::memcpy(static_cast<void*>(dest), static_cast<const void*>(src), count);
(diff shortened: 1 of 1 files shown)


karlnapf:
Pls dont open and close PRs, creates a lot of junk email. Thanks!

src/shogun/lib/memory.h
```diff
@@ -98,9 +98,9 @@ template <class T> void sg_generic_free(T* ptr)
 }
 
 template <class InputIt, class OutputIt>
-SG_FORCED_INLINE void* memcpy(InputIt dest, OutputIt src, size_t count)
+SG_FORCED_INLINE void* sg_memcpy(InputIt dest, OutputIt src, size_t count)
 {
-	return std::memcpy(static_cast<void*>(dest), static_cast<const void*>(src), count);
+	return shogun::memcpy(static_cast<void*>(dest), static_cast<const void*>(src), count);
```
lambday: Hey there. I'm afraid I didn't make the task clear enough :)

* We want to rename this `memcpy` to `sg_memcpy` (which you did) that in turn calls `std::memcpy` just as before (i.e., don't make it call `shogun::memcpy` inside - that won't exist anymore once you rename it)
* We want all existing `memcpy` uses in Shogun to use this newly added `sg_memcpy` instead.

See what I mean?

Issue #3716, opened 2017-03-17
Implement QR solver in linalg library


OXPHOS:
- Implement QR solver interface in [`LinalgNamespace`](https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/mathematics/linalg/LinalgNamespace.h) and `Eigen` and `ViennaCL` QR solver methods in `LinalgBackend*`. 
- Replace QR solver in [`LDA`](https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/classifier/LDA.cpp) and [`FisherLDA`](https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/preprocessor/FisherLDA.cpp) with `linalg::qr_solver`.

ref #2527

OXPHOS:
#3846

PR #3719, opened 2017-03-17
Implement QR solver in linalg library

micmn:
#3716 @OXPHOS I'm opening this PR even if it's just the eigen backend part so you can comment on this and because I have a question: the result of QR decomposition is a matrix containing the triangular matrix and the reflectors plus a vector of coeffs, which is the preferred way to return multiple values (instead of using pair...)?

### DIFF
--- a/src/shogun/mathematics/linalg/LinalgBackendEigen.h
+++ b/src/shogun/mathematics/linalg/LinalgBackendEigen.h
@@ -127,6 +127,16 @@ public:
 	DEFINE_FOR_NON_INTEGER_PTYPE(BACKEND_GENERIC_CHOLESKY_SOLVER, SGMatrix)
 	#undef BACKEND_GENERIC_CHOLESKY_SOLVER
 
+	/** Implementation of @see LinalgBackendBase::qr_solver */
+	#define BACKEND_GENERIC_QR_SOLVER(Type, Container) \
+	virtual Container<Type> qr_solver(const Container<Type>& A, \
+		const Container<Type>& b, linalg::QRDecompositionPivoting pivoting) const \
+	{  \
+		return qr_solver_impl(A, b, pivoting); \
+	}
+	DEFINE_FOR_NON_INTEGER_PTYPE(BACKEND_GENERIC_QR_SOLVER, SGMatrix)
+	#undef BACKEND_GENERIC_QR_SOLVER
+
 	/** Implementation of @see LinalgBackendBase::dot */
 	#define BACKEND_GENERIC_DOT(Type, Container) \
 	virtual Type dot(const Container<Type>& a, const Container<Type>& b) const \
@@ -376,6 +386,40 @@ private:
 			Eigen::TriangularView<Eigen::Map<typename SGMatrix<T>::EigenMatrixXt,
 				0, Eigen::Stride<0,0> >, Eigen::Lower> tlv(L_eig);
 			x_eig = (tlv.transpose()).solve(tlv.solve(b_eig));
+		}
+
+		return x;
+	}
+
+	/** Eigen3 QR solver */
+	template <typename T>
+	SGMatrix<T> qr_solver_impl(const SGMatrix<T>& A,
+			const SGMatrix<T>& b, linalg::QRDecompositionPivoting pivoting) const
+	{
+		SGMatrix<T> x = SGMatrix<T>(b.num_rows, b.num_cols);
+		x.zero();
+
+		typename SGMatrix<T>::EigenMatrixXtMap A_eig = A;
+		typename SGMatrix<T>::EigenMatrixXtMap b_eig = b;
+		typename SGMatrix<T>::EigenMatrixXtMap x_eig = x;
+
+		using linalg::QRDecompositionPivoting;
+		switch (pivoting) {
+			case QRDecompositionPivoting::None: {
+				auto qr = Eigen::HouseholderQR<Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic> >(A_eig);
+				x_eig = qr.solve(b_eig);
+				break;
+			}
+			case QRDecompositionPivoting::Column: {
+				auto qr = Eigen::ColPivHouseholderQR<Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic> >(A_eig);
+				x_eig = qr.solve(b_eig);
+				break;
+			}
+			case QRDecompositionPivoting::Full: {
+				auto qr = Eigen::FullPivHouseholderQR<Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic> >(A_eig);
+				x_eig = qr.solve(b_eig);
+				break;
+			}
 		}
 
 		return x;
--- a/src/shogun/mathematics/linalg/LinalgNamespace.h
+++ b/src/shogun/mathematics/linalg/LinalgNamespace.h
@@ -382,6 +382,28 @@ SGVector<T> cholesky_solver(const SGMatrix<T>& L, const SGVector<T>& b,
 }
 
 /**
+ * Solve the linear equations \f$Ax=b\f$ by means of the
+ * QR factorization of \f$A\f$
+ *
+ * @param A Square matrix
+ * @param b Right-hand side matrix
+ * @param p Pivoting type in QR decomposition
+ * @return \f$\x\f$
+ */
+template <typename T>
+SGMatrix<T> qr_solver(const SGMatrix<T>& A, const SGMatrix<T>& b,
+		QRDecompositionPivoting pivoting = QRDecompositionPivoting::None)
+{
+	REQUIRE(A.num_rows == A.num_cols, "Matrix is not square!\n");
+	REQUIRE((A.num_cols == b.num_rows),
+			"Number of columns of matrix A (%d) must match \
+					number of rows of matrix b (%d).\n",
+			A.num_cols, b.num_rows);
+
+	return infer_backend(A, b)->qr_solver(A, b, pivoting);
+}
+
+/**
  * Vector dot-product that works with generic vectors.
  *
  * @param a First vector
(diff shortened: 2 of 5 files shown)


src/shogun/mathematics/linalg/LinalgBackendEigen.h
```diff
@@ -127,6 +127,25 @@ class LinalgBackendEigen : public LinalgBackendBase
 	DEFINE_FOR_NON_INTEGER_PTYPE(BACKEND_GENERIC_CHOLESKY_SOLVER, SGMatrix)
 	#undef BACKEND_GENERIC_CHOLESKY_SOLVER
 
+	/** Implementation of @see LinalgBackendBase::cholesky_factor */
```
OXPHOS: qr_factor

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
```diff
@@ -158,6 +158,58 @@ TEST(LinalgBackendEigen, SGMatrix_cholesky_solver)
...
+	for (index_t i=0; i<size*size; ++i)
+		m[i]=i*i;
+
+	/* qr decomposition */
+	auto dec = qr_factor(m);
+	auto qr_cal = dec.first;
+	auto h_cal = dec.second;
+
+	Map<MatrixXd> map_qr(qr_cal.matrix,qr_cal.num_rows,qr_cal.num_cols);
+	Map<VectorXd> map_h(h_cal.vector,h_cal.size());
```
OXPHOS: Let's just use SGMatrx and SGVector. The point was to hide Eigen from users

OXPHOS:
@micmn gj! I was thinking about implementing the solver only because as you said we're still not sure how we the return type should be like. Let's ask @karlnapf about his opinion on returning pairs.

It seems pretty good to me. I'll take a closer look later since I'm kinda busy now. But it should just be some details.

micmn:
> I was thinking about implementing the solver only because as you said we're still not sure how we the return type should be like

I've refactored this a bit leaving just the solver so the factor stuff can be tackled separately.

src/shogun/preprocessor/FisherLDA.cpp
```diff
@@ -250,7 +251,10 @@ bool CFisherLDA::fit(CFeatures *features, CLabels *labels, int32_t num_dimension
 		// x=M
 		// MatrixXd M=Sw.householderQr().solve(Sb);
 		// calculate the eigenvalues and eigenvectors of M.
-		EigenSolver<MatrixXd> es(Sw.householderQr().solve(Sb));
+		SGMatrix<float64_t> M=
+				linalg::qr_solver(SGMatrix<float64_t>(Sw), SGMatrix<float64_t>(Sb));
+		SGMatrix<float64_t>::EigenMatrixXtMap M_eig=M;
+		EigenSolver<MatrixXd> es(M_eig);
```
OXPHOS: @karlnapf So the problem for replacing the solvers (including Cholesky solver) is that the matrices here are already mapped to `Eigen` objects. Now to call `linalg::XX_solver()`, we are casting them back to `SGMatrix`, and most likely they'll be cast again to `Eigen` inside linalg library. Do you think it makes sense? Or should we expect a huge refactoring to limit `Eigen` in linalg library only?
It is going to worth is if we use ViennaCL or XLA backends. But Eigen is the primary one now for sure.

src/shogun/mathematics/linalg/LinalgNamespace.h
```diff
@@ -381,6 +381,12 @@ SGVector<T> cholesky_solver(const SGMatrix<T>& L, const SGVector<T>& b,
 	return infer_backend(L, SGMatrix<T>(b))->cholesky_solver(L, b, lower);
 }
 
+template <typename T>
+SGMatrix<T> qr_solver(const SGMatrix<T>& A, const SGMatrix<T>& b)
+{
+	return infer_backend(A, b)->qr_solver(A, b);
```
OXPHOS: I suggest adding a dimension check here

src/shogun/mathematics/linalg/LinalgNamespace.h
```diff
@@ -381,6 +381,12 @@ SGVector<T> cholesky_solver(const SGMatrix<T>& L, const SGVector<T>& b,
 	return infer_backend(L, SGMatrix<T>(b))->cholesky_solver(L, b, lower);
 }
 
+template <typename T>
```
OXPHOS: docs

OXPHOS:
Neat :) 
Shogun doesn't require ublas. We can skip that for now.
You can open another PR for `qr_fact`. We can discuss there and I think your solution should be good.

src/shogun/mathematics/linalg/LinalgNamespace.h
micmn: Which one? `A` square? `A.num_cols == b.num_rows`?

karlnapf:
I think pair is fine. We can use tuple for more than two return parameters.
As linalg wont be expose to SWIG, that is fine (right?)
But please offer convenience methods so that the solver also can take a pair as input

src/shogun/mathematics/linalg/LinalgBackendEigen.h
```diff
@@ -381,6 +391,23 @@ class LinalgBackendEigen : public LinalgBackendBase
...
+	SGMatrix<T> qr_solver_impl(const SGMatrix<T>& A, const SGMatrix<T>& b) const
+	{
+		SGMatrix<T> x = SGMatrix<T>(b.num_rows, b.num_cols);
+		x.zero();
+
+		typename SGMatrix<T>::EigenMatrixXtMap A_eig = A;
+		typename SGMatrix<T>::EigenMatrixXtMap b_eig = b;
+		typename SGMatrix<T>::EigenMatrixXtMap x_eig = x;
+
+		Eigen::HouseholderQR<Eigen::Matrix<T,Eigen::Dynamic,Eigen::Dynamic> > qr(A_eig);
```
karlnapf: Documentation needs the type of solver somewhere. Eigen has multiple ones which are at different speed and stability. Maybe an option?

src/shogun/mathematics/linalg/LinalgNamespace.h
OXPHOS: Yep

src/shogun/preprocessor/FisherLDA.cpp
karlnapf: @lisitsyn suggested that this might work implicitly, i.e. without the explicit cast to SGMatrix, but with an implicit one.
The casting back to eigen should happen in the same line I think

src/shogun/mathematics/linalg/LinalgBackendEigen.h
OXPHOS: I only see HouseholderQR in Shogun so I think we're good for now as long as we doc it.

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
```diff
@@ -158,6 +158,32 @@ TEST(LinalgBackendEigen, SGMatrix_cholesky_solver)
 	EXPECT_EQ(x_ref.size(), x_cal.size());
 }
 
+TEST(LinalgBackendEigen, SGMatrix_qr_solver)
+{
+	const index_t size=3, n_x=2;
+	SGMatrix<float64_t> m(size, size);
```
vigsterkr: data/matrix generators should go into a generator class in unit testing :)

karlnapf: Yes there really should be a patch for this soon

src/shogun/mathematics/linalg/LinalgBackendEigen.h
karlnapf: We want to offer the other variants in eigen3 as well. At least the more stable ones with pivoting.
We can do that via an enum for each solver type.....

See e.g. https://docs.scipy.org/doc/scipy-0.14.0/reference/generated/scipy.linalg.qr.html
They also offer different methods

vigsterkr: whoooooooooo although enum seems like a super simple solution, but i reckon having a separate function for it would be better.... think about how the function that handles the enum will grown and grow and grow :)

karlnapf: I would do a sep. enum for each solver. There are only a handful for QR and they wont grow in numbers

karlnapf: We can also copy the scipy interface and offer a boolean for pivoting....

karlnapf: and a string parameter `kind="stable"` just for @vigsterkr

src/shogun/preprocessor/FisherLDA.cpp
micmn: Cool, the implicit cast to `SGMatrix` works but, at least in this case, the other way around does not: `EigenSolver<T>` doesn't accept a `Map` type and I can't find a way to avoid the explicit cast.

micmn:
> I think pair is fine. We can use tuple for more than two return parameters.
As linalg wont be expose to SWIG, that is fine (right?)
But please offer convenience methods so that the solver also can take a pair as input

The problem with offering convenience method is that each pivoting type would need different parameters, `HouseholderQR` matrix+vector, `ColPivHouseholderQR` also a permutation matrix and so on. Moreover eigen doesn't offer a constructor for `(Full/ColPiv)HouseholderQR` which reuses the already computed decomposition, so one can't take advantage of the `solve()` method and instead the solution has to be computed replicating the eigen's code i.e. (simplest case):
```
x_eig.applyOnTheLeft(Eigen::householderSequence(
    qr_eig.leftCols(rank),
    h_eig.head(rank)).transpose()
);
qr_eig.topLeftCorner(rank, rank)
    .template triangularView<Eigen::Upper()>
    .solveInPlace(x_eig.topRows(rank));
```
What about a wrapper? :)

karlnapf:
if you used tuple, then the number of return arguments could depend on the enum value passed.
And then we can *also* offer a wrapper

PR #3720, opened 2017-03-17
Fix CDynamicObjectArray / DynArray parameter handling + Test

lkuchenb:
CDynamicObjectArray could not be clone()ed properly (see test).

Changes:
 1. **Register `CDynamicObjectArray` parameters correctly**
 The `DynArray` member `array` was allocated + copied with the actual number of elements rather than the reserved number of elements. The cloned `DynArray` would insert new elements into unallocated space.
 1. **Initialize memory in `DynArray` if pointers are stored**
 Relevant when the `DynArray` holds `SGObject` pointers. The copy logic in `Parameter` checks for null and calls `clone()` on the pointers if not null.
 1. **Test**
 The test inserts many new elements to a `clone()`ed `CDynamicObjectArray`, most likely causing a bad access. In any case, the bad writes can be seen in valgrind. Test checks out fine after patch.

### DIFF
--- a/src/shogun/base/DynArray.h
+++ b/src/shogun/base/DynArray.h
@@ -12,6 +12,8 @@
 #ifndef _DYNARRAY_H_
 #define _DYNARRAY_H_
 
+#include <algorithm>
+
 #include <shogun/lib/config.h>
 
 #include <shogun/lib/common.h>
@@ -20,6 +22,19 @@
 namespace shogun
 {
 template <class T> class CDynamicArray;
+
+template <typename T> struct ptr_only
+{
+    template <typename U> static void fill_nullptr(U* ptr, size_t count) {}
+};
+
+template <typename T> struct ptr_only<T*>
+{
+    template <typename U> static void fill_nullptr(U* ptr, size_t count)
+    {
+	std::fill_n(ptr, count, nullptr);
+    }
+};
 
 /** @brief Template Dynamic array class that creates an array that can
  * be used like a list or an array.
@@ -52,6 +67,8 @@ template <class T> class DynArray
 			else
 				array=(T*) malloc(size_t(p_resize_granularity)*sizeof(T));
 
+			// If this is a DynArray of pointers, initialize with nullptr
+			ptr_only<T>::fill_nullptr(array, p_resize_granularity);
 			num_elements=p_resize_granularity;
 			current_num_elements=0;
 		}
@@ -355,6 +372,11 @@ template <class T> class DynArray
 			//in case of shrinking we must adjust last element idx
 			if (n-1<current_num_elements-1)
 				current_num_elements=n;
+
+			//in case of expanding we must initialize the newly
+			//allocated memory in case were are holding pointers
+			if (new_num_elements > num_elements)
+			    ptr_only<T>::fill_nullptr(array+num_elements, new_num_elements - num_elements);
 
 			num_elements=new_num_elements;
 			return true;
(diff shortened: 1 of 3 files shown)


lkuchenb:
The test `SGObject.equals_DynamicObjectArray_equal_after_resize` fails after this patch. This is because now, since

```
m_parameters->add_vector(&m_array.array, &m_array.num_elements, "array",...)
```

the `num_elements` parameter becomes part of the `array`s ptype, thus `equals()` fails for `array`. There is already a special treatment for the parameter `num_elements` itself, I'm not sure how to go from here though...

lkuchenb:
I don't see how this can be fixed without more ugly Dynamic(Object)Array specific hard-coded exceptions similar to [this](https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/base/SGObject.cpp#L714). A more compact way would be to override `clone()` once again, I implemented this in #3721 and suggest to merge that and discard this PR.

src/shogun/base/DynArray.h
```diff
@@ -21,6 +23,19 @@ namespace shogun
 {
 template <class T> class CDynamicArray;
 
+template <typename T> struct ptr_only
```
karlnapf: This would need some docs.
And a SWIG ignore

also @lambday and @vigsterkr should comment

src/shogun/base/DynArray.h
```diff
@@ -52,6 +67,8 @@ template <class T> class DynArray
 			else
 				array=(T*) malloc(size_t(p_resize_granularity)*sizeof(T));
 
+			// If this is a DynArray of pointers, initialize with nullptr
```
karlnapf: no need for comment if the method is documented

src/shogun/base/DynArray.h
```diff
@@ -356,6 +373,11 @@ template <class T> class DynArray
 			if (n-1<current_num_elements-1)
 				current_num_elements=n;
 
+			//in case of expanding we must initialize the newly
```
karlnapf: same here

src/shogun/lib/DynamicObjectArray.h
```diff
@@ -447,11 +447,14 @@ class CDynamicObjectArray : public CSGObject
 		/** register parameters */
 		virtual void init()
 		{
-			m_parameters->add_vector(&m_array.array, &m_array.current_num_elements, "array",
+			m_parameters->add_vector(&m_array.array, &m_array.num_elements, "array",
```
karlnapf: This is a dangerous change that might have a lot of consequences ....

tests/unit/base/DynamicObjectArrayNoTempl_unittest.cc
```diff
@@ -0,0 +1,34 @@
+/* This software is distributed under BSD 3-clause license (see LICENSE file).
```
karlnapf: needs full lincense

karlnapf: filename ....

tests/unit/base/DynamicObjectArrayNoTempl_unittest.cc
```diff
@@ -0,0 +1,34 @@
...
+
+#include <shogun/lib/DynamicObjectArray.h>
+#include <shogun/features/Subset.h>
+#include <gtest/gtest.h>
+
+using namespace shogun;
+
+TEST(DynamicObjectArray,clone)
+{
+	// Base array
```
karlnapf: same comments as in other PR also apply here

lkuchenb:
Closing this one, going with #3721

PR #3721, opened 2017-03-17
Alternative fix for Dynamic(Object)Array::clone()

lkuchenb:
Alternative solution to #3720 - involves overwriting `clone()` once again which might be ugly, but causes less trouble. Contains the same test, I suggest to merge this PR and discard #3720.


tests/unit/base/DynamicObjectArrayNoTempl_unittest.cc
```diff
@@ -0,0 +1,34 @@
+/* This software is distributed under BSD 3-clause license (see LICENSE file).
```
karlnapf: Need the full BSD license in here (I know it is annoying)

karlnapf: Don't get the filename, why not just DynamicObjectArray?

tests/unit/base/DynamicObjectArrayNoTempl_unittest.cc
```diff
@@ -0,0 +1,34 @@
...
+
+#include <shogun/lib/DynamicObjectArray.h>
+#include <shogun/features/Subset.h>
+#include <gtest/gtest.h>
+
+using namespace shogun;
+
+TEST(DynamicObjectArray,clone)
+{
+	// Base array
```
karlnapf: superflous comment, just call it "base_array" (the name do_array doesnt help a lot)

tests/unit/base/DynamicObjectArrayNoTempl_unittest.cc
```diff
@@ -0,0 +1,34 @@
...
+	CDynamicObjectArray * do_array = new CDynamicObjectArray();
+
+	// Something relatively simple to add
+	CSubset * subset = new CSubset();
+	do_array->append_element(subset);
+
+	// Cloned array
```
karlnapf: name it "cloned" and remove comment

tests/unit/base/DynamicObjectArrayNoTempl_unittest.cc
```diff
@@ -0,0 +1,34 @@
...
+	CDynamicObjectArray * do_array_clone = (CDynamicObjectArray*) do_array->clone();
+
+	// Expand the cloned array
```
karlnapf: superflous comment

tests/unit/base/DynamicObjectArrayNoTempl_unittest.cc
```diff
@@ -0,0 +1,34 @@
...
+	for (size_t i=0; i < 100; ++i)
```
karlnapf: use index_t

tests/unit/base/DynamicObjectArrayNoTempl_unittest.cc
karlnapf: I would rather say why you are doing this here. It is not immediately obvious.

karlnapf:
Thanks a lot for this!

src/shogun/lib/DynamicArray.h
```diff
@@ -551,14 +551,15 @@ template <class T> class CDynamicArray :public CSGObject
 		 */
 		inline void set_array_name(const char* p_name)
 		{
-			name=p_name;
+			name.resize_vector(strlen(p_name)+1);
```
karlnapf: is this method even used anywhere?

src/shogun/lib/DynamicArray.h
```diff
@@ -619,13 +620,26 @@ template <class T> class CDynamicArray :public CSGObject
 			m_array.resize_array(m_array.get_num_elements(), true);
 		}
 
+		CSGObject* clone()
+		{
+			CDynamicArray * cloned = (CDynamicArray*) CSGObject::clone();
```
karlnapf: after our discussion I think it would be best to have cloned objects have a different allocated sizes. But equals should still return true. Can be achieved by not registering the parameters and treating them in post loading?

src/shogun/lib/DynamicArray.h
```diff
@@ -619,13 +620,26 @@ template <class T> class CDynamicArray :public CSGObject
...
+			return cloned;
+		}
 
 	private:
 
 		/** register parameters */
 		virtual void init()
 		{
 			set_generic<T>();
+			set_array_name("Array");
```
karlnapf: This is ugly, rather remove ;)

src/shogun/base/SGObject.cpp
```diff
@@ -711,12 +711,12 @@ bool CSGObject::equals(CSGObject* other, float64_t accuracy, bool tolerant)
 		SG_DEBUG("comparing parameter \"%s\" to other's \"%s\"\n",
 				this_param->m_name, other_param->m_name);
 
-		/* hard-wired exception for DynamicObjectArray parameter num_elements */
-		if (!strcmp("DynamicObjectArray", get_name()) &&
+		/* hard-wired exception for [Wrapped|Dynamic]ObjectArray parameter num_elements */
+		if ((strcmp("DynamicObjectArray", get_name()) == 0 || strcmp("WrappedObjectArray", get_name()) == 0) &&
```
lkuchenb: Ha, didn't know `DynamicObjectArray` had a child... And it was forgotten whenever this beauty here was written...

lkuchenb: Ha, didn't know `DynamicObjectArray` had a child... And it was forgotten whenever this beauty here was written...

karlnapf: We cannot possibly accept this. This will blow up at the next oppurtunity...
If this parameter should not be part of equals, it should not be registered (and made transient in the future)

src/shogun/lib/DynamicArray.h
```diff
@@ -619,6 +604,16 @@ template <class T> class CDynamicArray :public CSGObject
...
 		}
 
+		CSGObject* clone()
+		{
+			CDynamicArray * cloned = (CDynamicArray*) CSGObject::clone();
+			// Since the array vector is registered with
+			// current_num_elements as size (see parameter
+			// registration) the cloned version has less memory
+			// allocated than known to dynarray. We fix this here.
+			cloned->m_array.num_elements = cloned->m_array.current_num_elements;
```
karlnapf: So again: 
 * we don't register the size of allocated memory parameter
 * serializing/deserializing will change it. Storing only stores the used part of the array, loading only allocated as much as needed, and re-sizes when something is added next time
 * it is not part of equals as it is not registered

lkuchenb:
OK, actually that's a point. We don't need that `num_elements` param registered anymore.
Threw it out and removed that ugly exception in `equals()`.

src/shogun/base/SGObject.cpp
```diff
@@ -711,24 +711,6 @@ bool CSGObject::equals(CSGObject* other, float64_t accuracy, bool tolerant)
 		SG_DEBUG("comparing parameter \"%s\" to other's \"%s\"\n",
 				this_param->m_name, other_param->m_name);
 
-		/* hard-wired exception for DynamicObjectArray parameter num_elements */
```
karlnapf: aaaah, so much nicer :)

tests/unit/CMakeLists.txt
```diff
@@ -18,16 +18,16 @@ ADD_CUSTOM_COMMAND(OUTPUT clone_unittest.cc
 	COMMENT "Generating clone_unittest.cc")
 LIST(APPEND TEMPLATE_GENERATED_UNITTEST clone_unittest.cc)
 
-ADD_CUSTOM_COMMAND(OUTPUT DynamicObjectArray_unittest.cc
```
karlnapf: think you need to rebase here ....

Issue #3663, opened 2017-02-28
Implement InputParser and ParseBuffer lock-free


vigsterkr:
currently for synchronisation mutex and conditional vars are being used, which is a waste of resources. use atomic types to implement a lock-free InputParser and ParseBuffer

tdjogi010:
@vigsterkr I would like to work on this.

tdjogi010:
@vigsterkr  I would like to work on this.

vigsterkr:
@tdjogi010 sure thing, just send a PR when you have something

tdjogi010:
Even though I am working on it, if anyone wishes to take this one up, please do so. Not sure if I can implement it correctly.

cullengao:
@vigsterkr @tdjogi010 Interesting. Thanks being so generous about the task, would have a stub on it.

tdjogi010:
Hi @vigsterkr , as far as I have understood, InputParser uses the ParseBuffer(which is a ring buffer) to hold the data and does the parsing(with the help of StreamingFile classes) in a separate thread when started. Now the StreamingFeatures fetches example objects one by one through a InputParser object. So there is only one parsing thread(producer) and only one streaming*feature thread(consumer) at all times? Not multiple producer/multiple consumer?

vigsterkr:
yeah pretty much that's the story there.

tdjogi010:
Thanks!

PR #3722, opened 2017-03-18
Lock free InputParser and ParseBuffer

tdjogi010:
See #3663. @vigsterkr  please take a look. It's a incomplete pull request. Only the ParseBuffer has been modified for the time being. Let me know if it's good to go. I only made the changes to the c++11 based implementation. (Not sure what to do with posix based one.)

### DIFF
--- a/src/shogun/io/streaming/ParseBuffer.h
+++ b/src/shogun/io/streaming/ParseBuffer.h
@@ -99,11 +99,14 @@ public:
 	Example<T>* get_free_example()
 	{
 #ifdef HAVE_CXX11
-		std::unique_lock<std::mutex> write_lk(*write_mutex, std::defer_lock);
-		std::unique_lock<std::mutex> current_ex_lock(*ex_in_use_mutex[ex_write_index], std::defer_lock);
-		std::lock(write_lk, current_ex_lock);
-		while (ex_used[ex_write_index] == E_NOT_USED)
-			ex_in_use_cond[ex_write_index]->wait(current_ex_lock);
+//		std::unique_lock<std::mutex> write_lk(*write_mutex, std::defer_lock);
+//		std::unique_lock<std::mutex> current_ex_lock(*ex_in_use_mutex[ex_write_index], std::defer_lock);
+//		std::lock(write_lk, current_ex_lock);
+        int32_t current_write_index = ex_write_index.load(std::memory_order_relaxed);  // only written from parser thread
+        int32_t current_read_index = ex_read_index.load(std::memory_order_acquire); // for synchronisation
+		while (ex_used[current_write_index] == E_NOT_USED)
+//			ex_in_use_cond[ex_write_index]->wait(current_ex_lock);
+            current_read_index = ex_read_index.load(std::memory_order_acquire); // for synchronisation
 		Example<T>* ex=&ex_ring[ex_write_index];
 #elif HAVE_PTHREAD
 		pthread_mutex_lock(write_lock);
@@ -203,7 +206,9 @@ protected:
 	 */
 	virtual void inc_read_index()
 	{
-		ex_read_index=(ex_read_index + 1) % ring_size;
+//		ex_read_index=(ex_read_index + 1) % ring_size;
+        int32_t current_read_index  = ex_read_index.load(std::memory_order_relaxed); // only written from streaming thread
+        ex_read_index.store((current_read_index + 1) % ring_size,std::memory_order_release);
 	}
 
 	/**
@@ -212,7 +217,9 @@ protected:
 	 */
 	virtual void inc_write_index()
 	{
-		ex_write_index=(ex_write_index + 1) % ring_size;
+//		ex_write_index=(ex_write_index + 1) % ring_size;
+        int32_t current_write_index = ex_write_index.load(std::memory_order_relaxed);  // only written from parser thread
+        ex_write_index.store((current_write_index + 1) % ring_size,std::memory_order_release);
 	}
 
 protected:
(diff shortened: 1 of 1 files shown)


src/shogun/io/streaming/ParseBuffer.h
```diff
@@ -245,9 +252,9 @@ template <class T> class CParseBuffer: public CSGObject
 #endif
 
 	/// Write position for next example
-	int32_t ex_write_index;
+	std::atomic<int32_t> ex_write_index;
 	/// Position of next example to be read
-	int32_t ex_read_index;
+    std::atomic<int32_t> ex_read_index;
```
vigsterkr: please check your whitespacing

src/shogun/io/streaming/ParseBuffer.h
```diff
@@ -271,8 +278,8 @@ template <class T> CParseBuffer<T>::CParseBuffer(int32_t size)
 	ex_ring = SG_CALLOC(Example<T>, ring_size);
 	ex_used = SG_MALLOC(E_IS_EXAMPLE_USED, ring_size);
 #ifdef HAVE_CXX11
-	read_mutex = std::make_shared<std::mutex>();
-	write_mutex = std::make_shared<std::mutex>();
+//	read_mutex = std::make_shared<std::mutex>();
```
vigsterkr: these comments can go... we know from the git diff what has changed

src/shogun/io/streaming/ParseBuffer.h
```diff
@@ -365,28 +372,29 @@ template <class T>
...
 
-	if (ex_used[current_index] == E_NOT_USED)
+	if (ex_used[current_read_index] == E_NOT_USED) //indirectly compared with current_write_index for checking available example
 		ex = return_example_to_read();
 	else
 		ex = NULL;
 
 #if defined(HAVE_PTHREAD) && !defined(HAVE_CXX11)
-	pthread_mutex_unlock(&ex_in_use_mutex[current_index]);
+	pthread_mutex_unlock(&ex_in_use_mutex[current_read_index]);
```
vigsterkr: no need for support of pthread

src/shogun/io/streaming/ParseBuffer.h
tdjogi010: I thought reading all of it in one would be better than comparing using git diff for this commit. Will refactor things asap.

vigsterkr: @tdjogi010 i really see no point...

src/shogun/io/streaming/ParseBuffer.h
```diff
@@ -99,11 +99,14 @@ template <class T> class CParseBuffer: public CSGObject
...
-		std::unique_lock<std::mutex> current_ex_lock(*ex_in_use_mutex[ex_write_index], std::defer_lock);
-		std::lock(write_lk, current_ex_lock);
-		while (ex_used[ex_write_index] == E_NOT_USED)
-			ex_in_use_cond[ex_write_index]->wait(current_ex_lock);
+//		std::unique_lock<std::mutex> write_lk(*write_mutex, std::defer_lock);
+//		std::unique_lock<std::mutex> current_ex_lock(*ex_in_use_mutex[ex_write_index], std::defer_lock);
+//		std::lock(write_lk, current_ex_lock);
+        int32_t current_write_index = ex_write_index.load(std::memory_order_relaxed);  // only written from parser thread
+        int32_t current_read_index = ex_read_index.load(std::memory_order_acquire); // for synchronisation
+		while (ex_used[current_write_index] == E_NOT_USED)
```
vigsterkr: @tdjogi010 cool, thnx... this is still an open question though.

tdjogi010:
```C++
while (ex_used[current_write_index] == E_NOT_USED)
         current_read_index = ex_read_index.load(std::memory_order_acquire); // for synchronisation
```
This is what I think:
Even though the `ex_used` itself is not atomic, doing `ex_read_index.load(std::memory_order_acquire);`(soon will replace it by `std::atomic_thread_fence(std::memory_order_acquire);` ) in every iteration will synchronize every store before `ex_read_index.store( ... ,std::memory_order_release);` even the  non atomic ones.(as suggested [here](https://gcc.gnu.org/wiki/Atomic/GCCMM/AtomicSync) ) So changes to the `ex_used` will be reflected after the `std::memory_order_acquire`. 

Let me know what you think.

src/shogun/io/streaming/ParseBuffer.h
tdjogi010: Yeah... don't know what I was thinking back then.

src/shogun/io/streaming/ParseBuffer.h
vigsterkr: i reckon there should be a more elegant way to do this... this is kind of a bruteforce solution. btw have you profiled it, whether this actually results in a speedup (or maybe a slowdown?)
definitely fences should be used since you dont use the values that you load. although i'm thinking whether somehow using exchange could be used here....

tdjogi010: How do I profile it using valgrind (or something else)? 

I am thinking of making the elements in ex_used atomic(like `vector<atomic<E_IS_EXAMPLE_USED> > ex_used`) then we can use something like 
```C++
while(ex_used[current_index].exchange(E_NOT_USED)==E_NOT_USED){}
...//get the example
ex_used[current_index].store(E_USED);
```
If not can you elaborate more on how to use exchange for our case?

vigsterkr: yeah this is starting to be more elegant.... profile = speed benchmarking. you could use any benchmarking tool for that

vigsterkr: i mean just to be clear the reason to switch here from the conditional var/mutex is to speed up the parsing procedure...

tdjogi010: Sorry my bad. I meant to ask what test needs to be run for profiling(or any benchmark in benchmark folder)?

vigsterkr: mmm no:) get a big file (generate one) and use it on the streamingfile
@lambday any chance you have a big input file for streamingfile?

tdjogi010:
I wrote a small program and measured the time taken ([here](https://gist.github.com/tdjogi010/e802a29a217ea93e4d817b1a36da2a19)). It's almost the same. But I think if the input parser is also made lock free it will improve a bit more. 

Also the program crashes(seg fault) if the another code(commented one) is used. @vigsterkr Am I not supposed to that?(it crashes because it tries to free already freed memory in memory.cpp::138)

tdjogi010:
The updated stats are [here](https://gist.github.com/tdjogi010/e802a29a217ea93e4d817b1a36da2a19) (a bit better). @vigsterkr let me know if you want more benchmarking and also if I can proceed to InputParser.

PR #3724, opened 2017-03-19
Refactoring for shogun::memcpy

MikeLing:


MikeLing:
#3700

MikeLing:
@lambday  ask review for this pr, thank you;)

src/shogun/base/Parameter.cpp
```diff
@@ -3354,7 +3354,7 @@ bool TParameter::copy_ptype(EPrimitiveType ptype, void* source, void* target)
 {
 	SG_SDEBUG("entering TParameter::copy_ptype()\n");
 
-	/* rather than using memcpy, use the cumbersome way here and cast all types.
+	/* rather than using sg_memcpy, use the cumbersome way here and cast all types.
```
lambday: haha `sed` all over :D.. love it ;)

lambday:
@MikeLing looks good to me.. merging this one..

But can you check whether we can use it in swig typemaps (Just `grep -Rin "memcpy" src/interfaces`).

MikeLing:
@lambday yep, we are using it in swig. Do we need to address it in swig, too?

Issue #3725, opened 2017-03-19
Resolve object replication for classes with 3rd party dependencies


lkuchenb:
There are several issues with shogun classes that depend on 3rd party libraries, e.g. CMKL or CLibSVM. The named two currently don't have any of their members registered as parameters and for the non-shogun ones (LibSVM, CLPLEX, GLPK, ...) it has to be checked upstream how to implement this first.

 * Check all shogun classes for `Parameter` registration. Test cases for `clone()` and `equals()` exist for CLibSVM etc, but are useless, since they always check out fine with no parameters!
 * Check 3rd party dependencies and how to replicate their objects.

lkuchenb:
The member variables in LibSVM classes could actually be moved to the `train_machine` scope, so maybe there is not much to do there.

lkuchenb:
I prepared a patch for MKL in #3730

lkuchenb:
And a patch for LibSVM classes in #3731

PR #3730, opened 2017-03-20
Register parameters for CMKL and override clone()

lkuchenb:
Registers MKL class parameters and handles 3rd party lib parameters in `::clone()`. See #3725. I couldn't get shogun to compile with CPLEX enabled, it looks to me as if there are plenty of `#ifdef USE_CPLEX...#endif` blocks in shogun that actually don't compile anymore.

### DIFF
--- a/src/shogun/classifier/mkl/MKL.cpp
+++ b/src/shogun/classifier/mkl/MKL.cpp
@@ -240,15 +240,12 @@ public:
 #endif
 };
 
-CMKL::CMKL(CSVM* s) : CSVM(), svm(NULL), C_mkl(0), mkl_norm(1), ent_lambda(0),
-		mkl_block_norm(1),beta_local(NULL), mkl_iterations(0), mkl_epsilon(1e-5),
-		interleaved_optimization(true), w_gap(1.0), rho(0), self()
+CMKL::CMKL(CSVM* s) : CSVM()
 {
+	SG_DEBUG("creating MKL object %p\n", this)
+	register_params();
 	set_constraint_generator(s);
 	self->init();
-
-	SG_DEBUG("creating MKL object %p\n", this)
-	lp_initialized = false ;
 }
 
 CMKL::~CMKL()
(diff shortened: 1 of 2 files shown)


src/shogun/classifier/mkl/MKL.cpp
```diff
@@ -241,14 +241,30 @@ class CMKL::Self
...
 		interleaved_optimization(true), w_gap(1.0), rho(0), self()
 {
 	set_constraint_generator(s);
 	self->init();
 
 	SG_DEBUG("creating MKL object %p\n", this)
 	lp_initialized = false ;
+
+	// Register parameters
+	SG_ADD((CSGObject**) &svm, "svm", "wrapper svm", MS_NOT_AVAILABLE);
```
karlnapf: this needs to be a in a `register_params` method that is called from every constructor

src/shogun/classifier/mkl/MKL.cpp
```diff
@@ -241,14 +241,30 @@ class CMKL::Self
...
+	SG_ADD(&mkl_block_norm, "mkl_block_norm", "mkl sparse trade-off parameter", MS_NOT_AVAILABLE);
+	m_parameters->add_vector(&beta_local, &beta_local_size, "beta_local", "subkernel weights on L1 term of elastic net mkl");
+	SG_ADD(&mkl_iterations, "mkl_iterations", "number of mkl steps", MS_NOT_AVAILABLE);
+	SG_ADD(&mkl_epsilon, "mkl_epsilon", "mkl epsilon", MS_NOT_AVAILABLE);
+	SG_ADD(&interleaved_optimization, "interleaved_optimization", "whether to use mkl wrapper or interleaved opt.", MS_NOT_AVAILABLE);
+	SG_ADD(&w_gap, "w_gap", "gap between interactions", MS_NOT_AVAILABLE);
+	SG_ADD(&rho, "rho", "objective after mkl iterations", MS_NOT_AVAILABLE);
+	SG_ADD(&lp_initialized, "lp_initialized", "if lp is Initialized", MS_NOT_AVAILABLE);
+
+	// Missing: beta_local, W, self, training_time_clock
```
karlnapf: Everything that is not registered will not be part of serialization and clone... is it fine with those here?
What about initialization of the memory of them?

src/shogun/classifier/mkl/MKL.cpp
```diff
@@ -262,6 +278,24 @@ CMKL::~CMKL()
...
+	CMKL* cloned = (CMKL*) CSGObject::clone();
+#ifdef USE_GLPK
+	// GLPK params
+	cloned->self->lp_glpk_parm = SG_MALLOC(glp_smcp, 1);
+	*cloned->self->lp_glpk_parm = *self->lp_glpk_parm;
+	// GLPK problem
+	cloned->self->lp_glpk = glp_create_prob();
+	glp_copy_prob(cloned->self->lp_glpk, self->lp_glpk, GLP_ON);
+#endif
+#ifdef USE_CPLEX
```
karlnapf: this can have its own issue .... 
We dont have CPLEX anywhere so we cannot actually check. Would be good to update this at some point ....

karlnapf:
One question. Do we have some tests of MKL results: unit/integration? Because if not it might be dangerous to merge this....

lkuchenb:
> Everything that is not registered will not be part of serialization and clone... is it fine with those here? What about initialization of the memory of them?

Sorry that was an obsolete comment. These are actually registered, `W*` I removed (was unused).

I'm not registering the clock. If one wanted to serialize a `CMKL` mid-action that might be relevant, but honestly I don't see the practical relevance. Also, `CTime` can't be cloned, it doesn't have it's params registered.

> One question. Do we have some tests of MKL results: unit/integration? Because if not it might be dangerous to merge this....

I'm not sure how well the actual MKL logic is tested. There are tests using MKL though. And even clone it (maybe they are using xval). At least I had a hand full of tests failing when I made mistakes within `clone()` or param registration. Anyways, we know that the before state was wrong, I think this should be merged.

vigsterkr:
@karlnapf i'd merge this... soooo?

src/shogun/classifier/mkl/MKL.cpp
```diff
@@ -262,6 +264,49 @@ CMKL::~CMKL()
...
+	}
+	// GLPK problem
+	if (self->lp_glpk != NULL)
+	{
+		cloned->self->lp_glpk = glp_create_prob();
+		glp_copy_prob(cloned->self->lp_glpk, self->lp_glpk, GLP_ON);
+	}
+#endif
+#ifdef USE_CPLEX
+	// Shogun currently doesn't compile with CPLEX
```
vigsterkr: please add `#error` here.... just in case if one day somebody fixes the rest of the cplex code, but forgets this part :)

src/shogun/classifier/mkl/MKL.cpp
```diff
@@ -262,6 +264,49 @@ CMKL::~CMKL()
...
+
+	// Missing: self (3rd party specific, handled in clone())
+}
+
+CSGObject* CMKL::clone()
+{
+	CMKL* cloned = (CMKL*) CSGObject::clone();
+#ifdef USE_GLPK
+	// GLPK params
+	if (self->lp_glpk_parm != NULL)
```
vigsterkr: why no `nulllptr`?

lkuchenb:
@vigsterkr Added the error and changed `NULL` to `nullptr`

src/shogun/classifier/mkl/MKL.h
```diff
@@ -95,6 +95,8 @@ class CMKL : public CSVM
 		 */
 		virtual ~CMKL();
 
+		CSGObject* clone();
```
vigsterkr: `override`?

lkuchenb: Was thinking about that, too. I don't know how you guys want to go about this in the future. If you want to make use of this feature in future changes, this implies changing more than needed for a fix in many cases, since compilers will give warnings when `override` is used inconsistently within one unit.

vigsterkr: i'm not so sure if i follow... clone is virtual and it needs to stay virtual to override... in c++11 it's override if you like, either way what you do now is not good :) imo

lkuchenb: Writing `virtual` is just eye candy, it's virtual if it's virtual in base. The only specifier that would make an actual difference here is `override`. If you want the `virtual` here for code style, I can add it.

vigsterkr: please do one or the other in order to be consistent with the rest of the codebase...

src/shogun/classifier/mkl/MKL.cpp
```diff
@@ -262,6 +264,49 @@ CMKL::~CMKL()
...
+	SG_ADD(&ent_lambda, "ent_lambda", "elastic net sparsity trade-off parameter", MS_NOT_AVAILABLE);
+	SG_ADD(&mkl_block_norm, "mkl_block_norm", "mkl sparse trade-off parameter", MS_NOT_AVAILABLE);
+	m_parameters->add_vector(&beta_local, &beta_local_size, "beta_local", "subkernel weights on L1 term of elastic net mkl");
+	SG_ADD(&mkl_iterations, "mkl_iterations", "number of mkl steps", MS_NOT_AVAILABLE);
+	SG_ADD(&mkl_epsilon, "mkl_epsilon", "mkl epsilon", MS_NOT_AVAILABLE);
+	SG_ADD(&interleaved_optimization, "interleaved_optimization", "whether to use mkl wrapper or interleaved opt.", MS_NOT_AVAILABLE);
+	SG_ADD(&w_gap, "w_gap", "gap between interactions", MS_NOT_AVAILABLE);
+	SG_ADD(&rho, "rho", "objective after mkl iterations", MS_NOT_AVAILABLE);
+	SG_ADD(&lp_initialized, "lp_initialized", "if lp is Initialized", MS_NOT_AVAILABLE);
+
```
karlnapf: registered parameters need to get default parameters. Lots of problems incoming otherwise.

It is a different story for existing code where parameters were already registered, but as you are adding that here, you need to put default values for all of them in this method. Consequently, need to replace all those init fields in the constructor with explicit assignments

src/shogun/classifier/mkl/MKL.cpp
```diff
@@ -262,6 +264,49 @@ CMKL::~CMKL()
...
+}
+
+CSGObject* CMKL::clone()
+{
+	CMKL* cloned = (CMKL*) CSGObject::clone();
+#ifdef USE_GLPK
+	// GLPK params
+	if (self->lp_glpk_parm != nullptr)
+	{
+		cloned->self->lp_glpk_parm = SG_MALLOC(glp_smcp, 1);
```
karlnapf: this fix is ok for me

karlnapf:
Happy to merge once registered parameter get default values

lkuchenb:
* Moved member init into `register_params()`
* Marked `clone()` as `virtual`

PR #3734, opened 2017-03-20
Use get and set in regression classes

abhinavrai44:


abhinavrai44:
@lisitsyn Plz check .

src/shogun/regression/LeastAngleRegression.cpp
```diff
@@ -185,7 +185,8 @@ bool CLeastAngleRegression::train_machine_templated(CDenseFeatures<ST> * data)
 	// main loop
 	//========================================
 	int32_t nloop=0;
-	while (m_num_active < max_active_allowed && max_corr/n_vec > m_epsilon && !stop_cond)
+	float64_t m_epsilon_value = get_epsilon();
```
karlnapf: no m prefix pls, this is not a member

src/shogun/regression/LeastAngleRegression.cpp
```diff
@@ -298,17 +299,20 @@ bool CLeastAngleRegression::train_machine_templated(CDenseFeatures<ST> * data)
 		// update estimator
 		for (index_t i=0; i < m_num_active; ++i)
 			beta[m_active_set[i]] += gamma * wA(i);
+
+		float64_t m_max_l1_norm_value = get_max_l1_norm();
```
karlnapf: no prefix

karlnapf: @lisitsyn: this is nuts! Storing the variables locally, obscures the code!!!

lisitsyn: hey, sure, we don't need that variable at all. Just call the getter in-place.

lisitsyn:
It looks good now. Travis complained but it was compiler crash so I restarted the job.

PR #3738, opened 2017-03-21
Fix #3605, Port KernelMachine to OpenMP

### LINKED ISSUE #3605: port parallel code using pthread
There are several parts of the code that uses `pthread` for implementing parallel execution of the algorithms. These should be ported to use either `OpenMP` or C++11 `std::thread` (preferably openmp).

some examples:
 * https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/lib/external/shogun_libsvm.cpp#L270
 * https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/distance/Distance.cpp#L368
 * https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/features/DotFeatures.cpp#L105
 * https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/features/hashed/HashedWDFeaturesTransposed.cpp#L242
 * https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/kernel/CombinedKernel.cpp#L465

for finding the full list use: `git grep HAVE_PTHREAD`

iRmantou:


karlnapf:
A unit test for single a multi threads of a kernel machine would help

iRmantou:
@karlnapf  I see this PR has passed Travis. Travis CI can't check this , isn't it ? use `./bin/shogun-unit-test`?

karlnapf:
Yeah travis only checks single threaded.
Can you at least memcheck this locally using multiple threads?
If that is ok, we can merge

iRmantou:
@karlnapf ok, just give me some time!

iRmantou:
@karlnapf  there is no unit test for `KernelMachine` in `tests/unit/machine`, isn't it? So I need to create it by myself ?

iRmantou:
Can I use `LibSVR_unittest` which inherited from `KernelMachine`  to test `KernelMachine` ?

karlnapf:
Yes.
Just run it locally with single thread and multiple ones and make sure your patch didn't change the output/result

iRmantou:
result

@karlnapf  Is this right?

karlnapf:
Almost, run this with a memory checker that checks thread safety

http://valgrind.org/docs/manual/hg-manual.html

iRmantou:
@karlnapf  I cleaned the whole `build` directory and recompiled it with my code, so I am sure the unit test use the changed code, then I used `valgrind` , and there is no error message. The results are as follows
varf
Is there anything I might have missed or need to note?

karlnapf:
What I meant is:
Check the unit test code. Does it run things single AND multithreaded.
Also, tell valgrind to use the `helgrind` tool, see above readme

iRmantou:
I have set `export OMP_NUM_THREADS` before I run the unit test.
What I should check or set in the unit test code like : `LibSVR_unittest.cc` to make sure it run things with  multithreaded and single?

karlnapf:
It doesnt really matter how you do it, the point is that you need to be aware what you are checking there. And that needs things to be explicitly executed single and multithreaded

karlnapf:
Any updates here?

PR #3745, opened 2017-03-23
Add PRange class to replace old SG_PROGRESS (WIP)

geektoni:
This is a prototype for the new progress bar that will (hopefully) replace `SG_PROGRESS`. This is not a full-fledged solution, it is just to show the idea and to get some feedback about it, so be warned ;)

As @lisitsyn suggested back here #3693, I developed a wrapper that takes a `range()` object and an `SGIO` instance. Every time the iterator is incremented, the wrapper should print a message on screen.

More specifically:
* `PRange` is the wrapper class;
* `PRange<T>::PIterator` is the iterator that will increment the given `Range<T>::Iterator`;
* `ProgressPrinter` is the actual class that will implement the progress bar logic;

The syntax should be something like this:
```c++
for (auto i : progress(range(1, 10), io)) 
{
 // Do stuff
}
````

I've thought about more ways to specify a progress object to customize its behaviour (I was inspired by this https://github.com/tqdm/tqdm):
```c++
for (auto i : progress( range(1, 10), io, "Custom_prefix: ", ASCII) )
{
 // Do stuff
}
```
In this example, the `progress` call sets the ASCII chars mode (print the progress bar using only ASCII chars) and a custom prefix that will be printed before the real progress bar.

I am aware that there are many things that can be improved and refactored (for example the execution speed), but, as I said above, this is just an "exploratory" patch.
___
Other changes:
* Remove HAVE_CXX11 guards from range.h since shogun needs C++11 now;
* Correct Range<T>::Iterator::operator++(int) behaviour (it was updating a local variable);

### DIFF
--- a/src/shogun/base/progress.h
+++ b/src/shogun/base/progress.h
@@ -1,5 +1,36 @@
 /*
+* BSD 3-Clause License
+*
+* Copyright (c) 2017, Shogun-Toolbox e.V. <shogun-team@shogun-toolbox.org>
+* All rights reserved.
+*
+* Redistribution and use in source and binary forms, with or without
+* modification, are permitted provided that the following conditions are met:
+*
+* * Redistributions of source code must retain the above copyright notice, this
+*   list of conditions and the following disclaimer.
+*
+* * Redistributions in binary form must reproduce the above copyright notice,
+*   this list of conditions and the following disclaimer in the documentation
+*   and/or other materials provided with the distribution.
+*
+* * Neither the name of the copyright holder nor the names of its
+*   contributors may be used to endorse or promote products derived from
+*   this software without specific prior written permission.
+*
+* THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+* AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+* IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+* DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+* FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+* DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+* SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+* CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+* OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+* OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*
 * Written (W) 2017 Giovanni De Toni
+*
 */
 
 #ifndef __SG_PROGRESS_H__
@@ -100,9 +131,11 @@ namespace shogun
 			// time displayed (decimals and integer).
 			int32_t progress_bar_space =
 			    (m_columns_num - 50 - m_prefix.length()) * 0.9;
-			REQUIRE(
-			    progress_bar_space > 0,
-			    "Not enough terminal space to show the progress bar!\n")
+
+			// TODO: this guy here brokes testing
+			// REQUIRE(
+			//    progress_bar_space > 0,
+			//    "Not enough terminal space to show the progress bar!\n")
 
 			char str[1000];
 			float64_t runtime = CTime::get_curtime();
@@ -112,7 +145,7 @@ namespace shogun
 				    (m_max_value - m_min_value + 1);
 
 			// Set up chunk size
-			size_chunk = difference / (double_t)progress_bar_space;
+			size_chunk = difference / (float64_t)progress_bar_space;
 
 			if (m_last_progress == 0)
 			{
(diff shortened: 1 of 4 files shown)


lisitsyn:
Amazing!

src/shogun/base/progress.h
```diff
@@ -0,0 +1,198 @@
...
+#include <memory>
+#include <string>
+
+#include <shogun/base/range.h>
+#include <shogun/io/SGIO.h>
+
+namespace shogun {
+
+    /** Possible print modes */
+    enum SG_PRG_MODE {
```
karlnapf: minor style comment. Put { to new line

karlnapf:
And any ideas how to integrate this with tmdq :D

lisitsyn:
@karlnapf why integrate it to python lib??

karlnapf:
In other words: How does it look like when shogun is called from python. Can we maybe use it with the tmdq if available, so in ipython notebook we get similar kind of optics....not very important

geektoni:
Possible use case:
```c++
#include <shogun/base/init.h>
#include <shogun/base/progress.h>
#include <shogun/io/SGIO.h>

int main() {
 
 init_shogun_with_defaults();

 SGIO io;
 for (auto i : progress(range(1, 100), io, "PROGRESS: ", ASCII))
 {
  // Do computations
 }
 exit_shogun();

return 0;
}
```
Result: 
optimised

lisitsyn:
@karlnapf I think that's different scope. If we use this new `progress(...)` thing all around we can integrate it to anything.

karlnapf:
Sure differente scope. But I got excited about it anyways.... to cite @oxphos : this is georgeous

src/shogun/base/progress.h
```diff
@@ -0,0 +1,199 @@
...
+    /** Creates @ref PRange given a range.
+     *
+     * @code
+     *  for (auto i : progress(range(0, 100), io)) { ... }
+     * @endcode
+     *
+     * @param   range   range used
+     */
+    template <typename T>
+    inline PRange<T> progress(Range<T> range, const SGIO & io)
```
lisitsyn: Let's add one more `progress` that uses global SGIO to simplify things. It's initialized with `init_shogun_with_defaults`

geektoni: Updated as requested.

lisitsyn:
👍 to merge. @karlnapf what about you?

geektoni:
@lisitsyn merge? This is only a work in progress version, there is no actual progress bar code (I mean, there is only the wrapper for the `range` method). Shouldn't we wait until it's finished?

lisitsyn:
@geektoni hahah I was too excited! Sorry

karlnapf:
Lets tidy it until @geektoni is pleased ;)

geektoni:
I've added the method's implementations. It should behave even better than the old SG_PROGRESS now ;)

src/shogun/base/progress.h
```diff
@@ -0,0 +1,434 @@
...
+			// Check if the progress was enabled
+			if (!m_io.get_show_progress())
+				return;
+
+			if (m_max_value <= m_min_value)
+				return;
+
+			// Check for terminal dimension. This is for provide
+			// a minimal resize functionality.
+			set_screen_size();
```
geektoni: This method is run each time we print the progress bar to account for the terminal size, which can change (for instance, if the user resizes or expands the window). However, I'm not so sure this is worth it (since it doesn't work so well). Should I remove this line? @vigsterkr

src/shogun/base/progress.h
```diff
@@ -0,0 +1,434 @@
...
+			float64_t size_chunk = -1;
+
+			// Check if we have enough space to show the progress bar
+			// Use only a fraction of it to account for the size of the
+			// time displayed (decimals and integer).
+			int32_t progress_bar_space =
+			    (m_columns_num - 50 - m_prefix.length()) * 0.9;
+
+			// TODO: this guy here brokes testing
+			// REQUIRE(
```
geektoni: I was planning to throw an error if the user enables the progress bar, but the terminal size is too small to print it. This is currently disabled because it makes the unit tests crash. Do you think it is something we need? So that I can work on re-enable it without destroying the test suite. @vigsterkr

geektoni:
@vigsterkr Travis fails because of the clang-format script. Probably, it doesn't like the new base branch that I set for this PR (since it works on my machine). All the tests pass, though.

src/shogun/base/progress.h
```diff
@@ -0,0 +1,434 @@
+/*
```
vigsterkr: pls copy here the bsd license

src/shogun/base/progress.h
vigsterkr: could you elaborate the part "doesn't work so well"?

src/shogun/base/progress.h
vigsterkr: mmm i think we can remove this as is.... i mean what is the terminal size when this will be a problem?

src/shogun/base/progress.h
```diff
@@ -0,0 +1,434 @@
...
+
+			char str[1000];
+			float64_t runtime = CTime::get_curtime();
+
+			if (difference > 0.0)
+				v = 100 * (current_value - m_min_value + 1) /
+				    (m_max_value - m_min_value + 1);
+
+			// Set up chunk size
+			size_chunk = difference / (double_t)progress_bar_space;
```
vigsterkr: `float64_t` or?

src/shogun/base/range.h
```diff
@@ -1,128 +1,127 @@
 #ifndef __SG_RANGE_H__
```
vigsterkr: same as above. license is missing :)

src/shogun/base/progress.h
geektoni: The method gets the size of the terminal, which will be used to calculate how much space we will use to print the progress bar. If the user resizes/expands the window, this method will correctly get the new dimensions, but there are cases in which the progress bar will be still printed in a wrong way (even if the program "knows" that the terminal size has changed).

src/shogun/base/progress.h
geektoni: It is calculated dynamically. There is a formula above which will calculate how much space we will use to print the progress bar. Sometimes the result can be 0 (or less), and that is a case in which we cannot print the progress bar.

vigsterkr:
@geektoni have you seen this error:
https://travis-ci.org/shogun-toolbox/shogun/jobs/238246972#L493

and the warnings:
https://travis-ci.org/shogun-toolbox/shogun/jobs/238246972#L1541

PR #3735, opened 2017-03-20
Move generated serialization unit tests into separate executable

micmn:
See discussion in #3679

PR #3747, opened 2017-03-26
Discover generated serialization tests

micmn:
Following #3735 serialization tests were not discovered by cmake, I don't think this was the intended behaviour (travis still builds the tests but doesn't run them, see this build https://travis-ci.org/shogun-toolbox/shogun/jobs/213507345#L5782), this should fix it.


karlnapf:
@vigsterkr pls merge

PR #3750, opened 2017-03-28
Generate json class list for unit tests

micmn:
Continues #3679 , `class_list.cpp.py` also builds a json class list used to generate the trained models serialization tests.


src/shogun/base/class_list.cpp.py
```diff
@@ -67,80 +72,86 @@ def extract_class_name(lines, line_nr, line, blacklist):
...
-    includes = []
-    result.sort()
-    for o in result:
-        includes.append('#include <shogun/%s>' % o.strip().lstrip('./'))
-    return includes
+def get_includes(classes):
+    includes = set()
+    for attrs in classes.values():
+        if not attrs['abstract']:
+            includes.add('#include <shogun/%s>' % attrs['header'])
```
karlnapf: @lisitsyn @vigsterkr I guess we could also use ctags here, just as for the meta examples?

src/shogun/CMakeLists.txt
```diff
@@ -63,15 +64,16 @@ FOREACH(template ${LIBSHOGUN_SRC_TMP})
 
     ADD_CUSTOM_COMMAND(OUTPUT ${generated_cpp}
         COMMAND ${CMAKE_COMMAND} -E copy_if_different ${template} ${generated_cpp}
-        COMMAND ${PYTHON_EXECUTABLE} ${generator_script} ${generated_cpp} -in ${CMAKE_CURRENT_BINARY_DIR}/headers_list.txt
+        COMMAND ${PYTHON_EXECUTABLE} ${generator_script} ${generated_cpp} ${generated_json}
```
karlnapf: can you check the meta example include directory discovery mechanism we do using ctags

karlnapf: https://github.com/shogun-toolbox/shogun/blob/feature/cereal/examples/meta/CMakeLists.txt#L23

karlnapf: https://github.com/shogun-toolbox/shogun/blob/feature/cereal/examples/meta/generator/generate.py#L20

micmn: I'm not sure I'm following here, currently the headers list is given by `{$LIBSHOGUN_HEADERS}`, instead of passing this list as an argument to the script (or `headers_list.txt` in case of MSVC) you want to pass the ctags file and in the script having the function `parseCtags()` from [generate.py](https://github.com/shogun-toolbox/shogun/blob/feature/cereal/examples/meta/generator/generate.py#L20), then extract all the headers from this dictionary, something like:
```
tags=parseCtags(arg)
HEADERS=filter(lambda x: x.endswith('.h'), tags.values())
```
And the remaining part of the script stays the same, correct?

PR #3751, opened 2017-03-28
Trained models serialization test

micmn:
See #3679 #3750

### DIFF
--- a/tests/unit/base/trained_model_serialization_unittest.cc.py
+++ b/tests/unit/base/trained_model_serialization_unittest.cc.py
@@ -1,0 +1,105 @@
+#!/usr/bin/env python
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 3 of the License, or
+# (at your option) any later version.
+#
+
+# Classes that need custom initialization
+IGNORE_LIST = [
+    'CFeatureBlockLogisticRegression', 'CLibLinearMTL',
+    'CMultitaskLinearMachine', 'CMultitaskLogisticRegression',
+    'CMultitaskL12LogisticRegression', 'CMultitaskLeastSquaresRegression',
+    'CMultitaskTraceLogisticRegression', 'CMultitaskClusteredLogisticRegression',
+    'CLatentSVM', 'CLatentSOSVM',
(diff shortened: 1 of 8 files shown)


vigsterkr:
@micmn can we pick this up sometime? :)

vigsterkr:
@micmn could you rebase this one and see what happens on travis? :) would love to get this in ASAP

vigsterkr:
@micmn one thing: would it be possible that we avoid using files from `data` and use the fixtures created by @MikeLing see: https://github.com/shogun-toolbox/shogun/pull/3812

tests/unit/base/trained_model_serialization_unittest.cc.py
```diff
@@ -0,0 +1,72 @@
...
+    import os
+    basename = os.path.basename(output_file)
+    basename = basename.replace('.cc', '')
+    print("Please install jinja2 for trained model serialization unit-tests")
+    outputText = ['''#include <gtest/gtest.h>
+TEST(Dummy, %s_dummy)
+{
+}''' % (basename)]
+
+f = open(output_file, 'w')
```
vigsterkr: could we use `with open.... as f:`

micmn:
I updated this one, so now class hierarchy is based on ctags, but there are still two things that I couldn't handle through ctags: checking if a class is not compiled due to an #ifdef directive (i.e. #have_lapack) and checking if a class is abstract.

tests/unit/base/trained_model_serialization_unittest.cc.py
```diff
@@ -0,0 +1,100 @@
...
+    return any([name + '\timplementation:pure virtual' in tag for tag in tags])
+
+
+def is_shogun_class(c):
+    return c[0] == 'C' and c[1].isupper() and 'class' in c
+
+
+def get_shogun_classes(tags):
+    classes = {}
+    # in ctags format it is TAG\tLOCATION\t..\tinherits:CLASS
```
karlnapf: yeah much nicer to parse ctag ouput rather than source code.
One question, isnt there a python package that does things like you do here?

karlnapf:
Quite excited about having all the subclasses of linear models checked soon! This will make Shogun a much better place
Same will be ace for kernel machines (and reveal problems I am sure)

Issue #3697, opened 2017-03-14
replace relative paths to data dir in notebooks


vigsterkr:
currently in all of our ipython notebooks the data files are referenced by a simple `../../../data/<filename>` relative path, like for example this:
```
shogun_feats_linear = RealFeatures(CSVFile('../../../data/toy/classifier_binary_2d_linear_features_train.dat'))
```
this is causing a lot of trouble if you copy the notebook somewhere else.

in this entrance task, one should replace all these relative paths with an environment variable called `SHOGUN_DATA_DIR`, which then would have a default value if it's not set.

the fix should be applied to almost every ipython notebook here: https://github.com/shogun-toolbox/shogun/tree/develop/doc/ipython-notebooks

amalad:
Hi! Can I take up this issue?

karlnapf:
No need to ask :)

deveshnag1:
so the ../../../data should be replaced by an os.getenv call right?

deveshnag1:
@vigsterkr Is this what you had in mind?
https://github.com/deveshnag1/shogun/blob/notebook_relative/doc/ipython-notebooks/classification/Classification.ipynb

vigsterkr:
@deveshnag1 mmm in a way yes but as you can see in the meanwhile this has been addressed by somebody else in the https://github.com/shogun-toolbox/shogun/pull/3753 pr

vigsterkr:
fixed in #3753

PR #3753, opened 2017-03-28
replace relative paths to data dir in notebooks

tingpan:
issue: #3697 

I created `env.py` to store environment variables such as SHOGUN_DATA_DIR. User could change the value in `env.py` and it would apply to all files

I did not use `os.environ['SHOGUN_DATA_DIR']` as it is dangerous to modify system environment variable here; Besides, I need to put default value in every file if using `%env SHOGUN_DATA_DIR`. 

However I am not sure if this is the best practice so feel free to comment and give suggestions:)

Thanks!


vigsterkr:
@tingpan so you replaced the relative path to the data dir, with a relative path to a script... i mean this is the same thing in different wrapping. the point was that there should be no relative path..

tingpan:
I would find another way to fix this. Thanks for your time:)

doc/ipython-notebooks/classification/Classification.ipynb
```diff
@@ -47,6 +47,7 @@
     "%matplotlib inline\n",
     "import numpy as np\n",
     "import matplotlib.pyplot as plt\n",
+    "SHOGUN_DATA_DIR = os.environ['SHOGUN_DATA_DIR'] if os.environ.has_key('SHOGUN_DATA_DIR') else '../../../data/'\n"
```
vigsterkr: please read the manuals... i.e. os.getenv('SHOGUN_DATA_DIR', '../../../data')

doc/ipython-notebooks/classification/Classification.ipynb
```diff
@@ -87,11 +88,11 @@
    },
    "outputs": [],
    "source": [
-    "shogun_feats_linear = RealFeatures(CSVFile('../../../data/toy/classifier_binary_2d_linear_features_train.dat'))\n",
-    "shogun_labels_linear = BinaryLabels(CSVFile('../../../data/toy/classifier_binary_2d_linear_labels_train.dat'))\n",
+    "shogun_feats_linear = RealFeatures(CSVFile(SHOGUN_DATA_DIR + 'toy/classifier_binary_2d_linear_features_train.dat'))\n",
```
vigsterkr: os.path.join instead of simple concatenation of strings

doc/ipython-notebooks/classification/SupportVectorMachines.ipynb
```diff
@@ -801,7 +802,7 @@
    },
    "outputs": [],
    "source": [
-    "f = open('../../../data/uci/ionosphere/ionosphere.data')\n",
+    "f = open(SHOGUN_DATA_DIR + 'uci/ionosphere/ionosphere.data')\n",
```
vigsterkr: os.path.join

doc/ipython-notebooks/classification/Classification.ipynb
tingpan: Thanks for these comments. Just finished refactored and squashed all fixup commits:)

vigsterkr:
@tingpan great! have you tested 1-2 of these notebooks that it works ?

tingpan:
@vigsterkr yes I've tested lots of pages and data files could be found correctly

vigsterkr:
ok perfect merging

PR #3764, opened 2017-04-03
Implement QR solver in linalg library

SashaMalysheva:
I have implemented SVD and GR decomposition and solver.
The task is incomplete because I've encountered with some problems in replacing QR solver in LDA and FisherLDA.

ref #3716

### DIFF
--- a/src/shogun/mathematics/linalg/LinalgBackendBase.h
+++ b/src/shogun/mathematics/linalg/LinalgBackendBase.h
@@ -132,6 +132,93 @@ public:
 	DEFINE_FOR_NON_INTEGER_PTYPE(BACKEND_GENERIC_CHOLESKY_SOLVER, SGMatrix)
 	#undef BACKEND_GENERIC_CHOLESKY_SOLVER
 
+    /**
+     * Wrapper solver with QR decomposition.
+     *
+     * @see linalg::qr_solver
+     */
+    #define BACKEND_GENERIC_QR_SOLVER(Type, Container) \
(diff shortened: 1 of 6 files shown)


lisitsyn:
@SashaMalysheva what were the problems with LDA?

src/shogun/mathematics/linalg/LinalgBackendBase.h
```diff
@@ -132,6 +132,93 @@ class LinalgBackendBase
 	DEFINE_FOR_NON_INTEGER_PTYPE(BACKEND_GENERIC_CHOLESKY_SOLVER, SGMatrix)
 	#undef BACKEND_GENERIC_CHOLESKY_SOLVER
 
+    /**
+     * Wrapper solver with QR decomposition.
+     *
+     * @see linalg::qr_solver
+     */
+    #define BACKEND_GENERIC_QR_SOLVER(Type, Container) \
+	virtual SGMatrix<Type> qr_solver(const Container<Type>& A, \
```
OXPHOS: identation

OXPHOS: and elsewhere

OXPHOS:
It seems like the `eigen.qr_solver` itself failed from the Travis. No?

karlnapf:
Hi!
This looks quite good. However, you realise that if you want all matrices (U, S, V) of an SVD, it computes the SVD three times?
Instead, the method should return a triple with all matrices ...

PR #3774, opened 2017-04-22
Move sg_memcpy out ifdef TRACE_MEMORY_ALLOCS.

iglesias:
This way sg_memcpy is also defined when TRACE_MEMORY_ALLOCS=OFF.


iglesias:
Should address the current error in nightly_all:
http://buildbot.shogun-toolbox.org/builders/nightly_all/builds/54/steps/compile/logs/stdio

PR #3777, opened 2017-04-23
Refactor unlink to _unlink in tests for compatibility with non-posix.

iglesias:


iglesias:
From Travis it seems that `_unlink` is not available in unix (maybe the right include is missing). It fixed the warnings in Windows build.

A good solution might be to wait until we use C++17 and change to [std::filesystem::remove](http://en.cppreference.com/w/cpp/filesystem/remove).

PR #3783, opened 2017-04-24
Remove HAVE_LARGEFILE conf. option since it is not used.

iglesias:
Am I missing where this is used?


iglesias:
Self-reminder: if we merge this, remove from infra setting of this option.

karlnapf:
a2ab5e48977fd92b2837cbd6cfd0eaedd89ce27b

Issue #3786, opened 2017-04-25
Move all files that are guarded with USE_GPL_SHOGUN


karlnapf:
To the folder `shogun/src/gpl`. The build will take care of the rest (hopefully)

There are example files in `shogun/src/gpl`, so you can see how things work. The feature branch `feature/gpl/` will be merged soon, so then the works on this can start ...

Steps:
 * find file that is guarded with `#ifdef USE_GPL_SHOGUN` (you might find files that are clearly GPL but don't have this guard, then try adding it)
 * move header and implementation into `shogun/src/gpl`
 * you can remove any `#ifdef USE_GPL_SHOGUN` guards once files are in `shogun/src/gpl`

 * if a meta example covers the class, add a guard in `cmake/FindMetaExamples.cmake` (we dont want to run the example if the class is not available)
 * Build locally with both `-DLICENSE_GPL_SHOGUN=ON` and with `-DLICENSE_GPL_SHOGUN=ON -DUSE_SMVLIGHT=OFF` to test both scenarios, run the tests
 * If everything goes fine, send a pull request
 * There is a new build-slave for setting at: http://buildbot.shogun-toolbox.org/builders/nightly_bsd_license

Old discussion/description:  #2958

karlnapf:
Example: #3790

karlnapf:
#3900

vigsterkr:
@karlnapf is this done fully or still there are parts that needs to be cleared up?

karlnapf:
All guarded files are moved, so this can be closed. As far as I can tell, there is no more GPL code in `shogun/src`, but others should check as well

karlnapf:
Did another set of work to push this in #3946

sorig:
I found more files. Updates to follow

karlnapf:
@sorig which ones?

karlnapf:
Files that need guard or need to be moved
```
SVMSGD.cpp
```

PR #3790, opened 2017-04-26
Move multitask clustered LR to gpl directory

lisitsyn:


karlnapf:
#3786

karlnapf:
cah

iglesias:
Should the example python_modular/transfer_multitask_clustered_logistic_regression.py also moved under gpl?

karlnapf:
Thanks for noting that!
Nope, that code (the example itself) is not GPL (I mean it is but we can change that). This example should be guarded in the build instead.

PR #3794, opened 2017-04-27
Add clang-format file and Travis integration to check PRs' code style.

geektoni:
This PR is focused on make Travis able to evaluate PRs' code style. The main idea is to create a clang-format file which will define some Shogun's code style guidelines. These guidelines should be followed when writing patches/code.

This patch's `.clang-format` file is a bit skinny. There are only some basic instructions which were modelled by looking at this page:
* https://github.com/shogun-toolbox/shogun/wiki/Code-style

A simple example on how Travis will (should) behave after this patch can be found at this link: 
* https://travis-ci.org/geektoni/clang-format-test/jobs/226336339

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,6 +54,7 @@ matrix:
       services: docker
       env:
         - CMAKE_OPTIONS="-DOctaveModular=ON -DTRAVIS_DISABLE_UNIT_TESTS=ON -DTRAVIS_DISABLE_LIBSHOGUN_TESTS=ON"
+        - OCTAVE_MODULAR=true
     - compiler: clang
       services: docker
       env:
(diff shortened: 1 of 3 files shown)


vigsterkr:
@geektoni great! let's see the CIs

.travis.yml
```diff
@@ -73,6 +73,7 @@ before_install:
 before_script:
   - docker exec -t devenv /bin/sh -c "cd /opt/shogun/build; cmake -DCMAKE_INSTALL_PREFIX=$HOME/shogun-build -DENABLE_TESTING=ON $CMAKE_OPTIONS .."
 script:
+  - docker exec -t devenv /bin/sh -c "cd /opt/shogun/scripts; bash ./check_format.sh"
```
geektoni: Actually, I'm not really sure this line will work (since the `check_format.sh` script needs bash).

scripts/check_format.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
```
vigsterkr: here change the interpreter to:
```
#!/usr/bin/env bash
```

.travis.yml
vigsterkr: just remove `bash` and make sure that `check_format.sh` is executable and let the `env` find out where bash is ... if available.

karlnapf:
I'm excited about this! :)

lisitsyn:
Prekrasno!

geektoni:
I've realized that the `check_format.sh` script won't work inside the docker container, since it relies on some Travis environmental variables :( I'll export them manually.

karlnapf:
restarted build

geektoni:
@karlnapf the build will fail. There is still a slight issue inside the travis.yml file. I will fix it later.

geektoni:
To clarify, clang-format will check for these things:
* 80 characters line limits;
* Code indentation done by using tabs;
* Line break before braces (in case of class declaration, functions, enums, namespaces etc.);
* Disallow contracting simple braced statements to a single line (e.g is not possible to write `if (a) {return 1;}`);
* Disallow contracting short case labels to a single line (e.g `case 1: x=1; break;`);
* Disallow contracting short function to a single line (e.g `void a(int a) {int b=a;}`)
* Disallow statement like `if (a) return;` or `while(a<3) i++;`;
* Always break after the `template<...>` of a template declaration;
* Namespace indentation;
* Pointer aligned to the left (e.g `int* a`, instead of `int * a` or `int *a`); 
* Sorted #include directives ( more stylish ;)  );

Obviously, all these options can be changed and more things can be done.

Unfortunately, clang-format cannot check for naming issues or other related problems (like using only specific base types).

P.S: CIs should be all green now. Yay!

.travis.yml
```diff
@@ -69,10 +69,12 @@ matrix:
 before_install:
   - docker pull shogun/shogun-dev
   - echo $CC; echo $CXX
```
vigsterkr: can you remove this and the one below :)

.travis.yml
```diff
@@ -69,10 +69,12 @@ matrix:
...
-  - docker run -t -d -P -e "JAVA_HOME=/usr/lib/jvm/java-8-oracle" -e "CC=$CC" -e "CXX=$CXX" --name devenv -v $PWD:/opt/shogun shogun/shogun-dev /bin/sh -c "mkdir /opt/shogun/build;bash"
+  - echo $TRAVIS_PULL_REQUEST; echo $TRAVIS_BRANCH;
```
vigsterkr: there's no need for this anymore right?

vigsterkr:
great stuff! just one minor thing: do you mind adding a mini `if` branch to this:
```
docker exec -t devenv /bin/sh -c "cd /opt/shogun/; ./scripts/check_format.sh"
```
that this docker command is only executed when `CC == gcc` & `-z $OctaveModular` so that the formatting check is only executed in one of the jobs... as it's really unnecessary to have it run in all the jobs.

geektoni:
@vigsterkr updated.

vigsterkr:
awesome, merging!

PR #3801, opened 2017-04-30
Better support for different and unknown architectures

besser82:
This fixes building on all arches; x86 / x86_64 are the only ones, which know about the opcode 'pause'.


besser82:
Well, that would bloat the code in this header a bit too much…

besser82:
As the inlined code already shows up as a function.

besser82:
Greaty!  Will merge this in after TravisCI has passed.  =)

besser82:
Everything should be fine now… Merging!

PR #3802, opened 2017-04-30
Fix test UAIFileTest.preamble failure for some locales

luisffranca:
The unit test _UAIFileTest.preamble_ failed for locales with other decimal delimiters, such as "pt_BR.utf8". Some functions, like strtod(), use the "C" locale and the test failed when they were called with a different one.

This is fixed by using the "C" locale in CUAIFile::get_vector().

### DIFF
--- a/src/shogun/io/UAIFile.cpp
+++ b/src/shogun/io/UAIFile.cpp
@@ -104,30 +104,34 @@ void CUAIFile::init_with_defaults()
     SG_REF(m_line_reader);
 }
 
-#define GET_VECTOR(read_func, sg_type) \
-void CUAIFile::get_vector(sg_type*& vector, int32_t& len) \
-{ \
-    if (!m_line_reader->has_next()) \
-        return; \
-    \
-    SGVector<char> line; \
-    int32_t num_elements = 0; \
-    \
-    line = m_line_reader->read_line(); \
-    m_tokenizer->set_text(line); \
-    while (m_tokenizer->has_next()) \
-    { \
-        int32_t temp_start; \
-        m_tokenizer->next_token_idx(temp_start); \
-        num_elements++; \
-    } \
-    \
-    vector = SG_MALLOC(sg_type, num_elements); \
-    m_parser->set_text(line); \
-    for (int32_t i=0; i<num_elements; i++) \
-        vector[i] = m_parser->read_func(); \
-    len = num_elements; \
-}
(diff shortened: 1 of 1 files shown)


lisitsyn:
@luisffranca hey! Thanks for submitting this patch.

@vigsterkr Travis fails with a style check related to original code. Do we merge it?

iglesias:
Should we update the clang-format style file for this? @geektoni

geektoni:
@iglesias We could force clang-format to align escaped newlines as far left (or far right) as possible.
By doing this, the new code will become like this https://pastebin.com/xi5CH2GP (the output on Travis is messed up because of the terminal which cannot show correctly tabs).

iglesias:
@geektoni I don't have strong feelings about the location of the trailing '\'s; it looks indeed nice all aligned.

Regarding the indentation between the first line with the macro signature and the following lines, I think it would be better to have everything with the same level of indentation since I believe that is more consistent with the current codebase.

geektoni:
@iglesias I don't know if there is an option to prevent indentation inside macro definitions (apart from using `// clang-format off` and `// clang-format on` which disable formatting). I will investigate a bit.

I don't think we will be able to configure clang-format to respect exactly all Shogun code conventions. There will surely be things we cannot force.

geektoni:
@iglesias I've done some research and apparently it is not possible to format `#define` without indentation. There are three options now. We could:
1. Merge without caring about that style issue;
2. Surround the `#define` with `// clang-format off` and `// clang-format on` instruction;
3. Apply changes suggested by clang-format and then merge;

Since the clang-format purpose is to free devs from thinking about code formatting, I would proceed with option 3. I agree that it will change the codebase standard, but in the long run, I think it will be worth it.

iglesias:
I agree with you, @geektoni. I think 3 is the most suitable option as well.

geektoni:
Hi @luisffranca! As you can read, one of the Travis tests failed because of some code style issues. 

To fix that error please give a look to the Travis log below and you'll see the commands you have to run: https://travis-ci.org/shogun-toolbox/shogun/jobs/227279295#L524

Make sure to have `clang-format-3.8` installed on your local machine or the commands won't work.

After doing that, update the PR here and you're done. :tada:

luisffranca:
Hi @geektoni! Thanks for your help.

I've updated the PR and now the tests should work!

PR #3803, opened 2017-04-30
Fix possible IndexError and/or TypeError in undocumented/python_modular

besser82:


besser82:
Yes!  =)  Just packaging v6.0.0 for Fedora  ;)

besser82:
Please check https://github.com/shogun-toolbox/shogun/pull/3801 first…  ;)

lisitsyn:
Does it really fail on some platform? I am curious what makes it fail.

besser82:
For me it those four tests fails on any platform, when building on RHEL or Fedora…  :(

lisitsyn:
Do you happen to know what is the type *before* the cast? The whole world looks fragile to me if range produces not ints but something.

besser82:
They are float() before.

lisitsyn:
Oh I still don't get where floats come from but lets merge it 👍

besser82:
Okie, will do so  =)

PR #3804, opened 2017-04-30
Fix detection of Scala

besser82:


besser82:
This is a sure fix; merging without Travis.

PR #3810, opened 2017-05-09
Port kernelMachine to openmp

MikeLing:


src/shogun/machine/KernelMachine.cpp
```diff
@@ -553,72 +503,57 @@ SGVector<float64_t> CKernelMachine::apply_locked_get_output(
...
+	#pragma omp parallel shared(num_threads, step)
 	{
-		pthread_t* threads = SG_MALLOC(pthread_t, num_threads-1);
-		S_THREAD_PARAM_KERNEL_MACHINE* params=SG_MALLOC(S_THREAD_PARAM_KERNEL_MACHINE, num_threads);
-		int32_t step= num_inds/num_threads;
-
-		int32_t t;
-		for (t=0; t<num_threads-1; t++)
+#ifdef HAVE_OPENMP
+		#pragma opm single
```
lisitsyn: Does that compile? Typo here

MikeLing: mmm, I think so. I compile it under the environment with openmp. I don't know why there's no error comes out. I will fix it right now :)

MikeLing:
Ask review for this pr

src/shogun/machine/KernelMachine.cpp
```diff
@@ -333,64 +337,62 @@ SGVector<float64_t> CKernelMachine::apply_get_outputs(CFeatures* data)
...
-			{
-				pthread_t* threads = SG_MALLOC(pthread_t, num_threads-1);
-				S_THREAD_PARAM_KERNEL_MACHINE* params = SG_MALLOC(S_THREAD_PARAM_KERNEL_MACHINE, num_threads);
-				int32_t step= num_vectors/num_threads;
-
-				int32_t t;
 
-				for (t=0; t<num_threads-1; t++)
+#ifdef HAVE_OPENMP
+#pragma opm single
```
vigsterkr: typo.... due to this as it was mentioned by @karlnapf in https://github.com/shogun-toolbox/shogun/pull/3738#issuecomment-289268657 that be great that our unit test somehow tests both multi and single threaded scenario...

src/shogun/machine/KernelMachine.cpp
```diff
@@ -553,72 +507,60 @@ SGVector<float64_t> CKernelMachine::apply_locked_get_output(
...
-	else
-	{
-		pthread_t* threads = SG_MALLOC(pthread_t, num_threads-1);
-		S_THREAD_PARAM_KERNEL_MACHINE* params=SG_MALLOC(S_THREAD_PARAM_KERNEL_MACHINE, num_threads);
-		int32_t step= num_inds/num_threads;
-
-		int32_t t;
-		for (t=0; t<num_threads-1; t++)
+#ifdef HAVE_OPENMP
+#pragma opm single
```
vigsterkr: same typo here...

src/shogun/machine/KernelMachine.cpp
MikeLing: I found the comment in https://github.com/shogun-toolbox/shogun/pull/3738#issuecomment-289267398 doesn't really helps to test multi threaded situation. `omp_get_num_threads()` always return 1 even I export OMP_NUM_THREADS as 4. How could I validate the openmp really works?

MikeLing: I just use `parallel->get_num_threads()` to get the number of threads, but the result stay the same (single thread) :/

PR #3818, opened 2017-05-18
Update install instructions for homebrew

duhduhdan:
Following the install docs, I encountered an issue with `brew install shogun` in that the formula wasn't found. A bit of digging into the [homebrew-science](https://github.com/Homebrew/homebrew-science) docs led me to solving this by executing `brew install homebrew/science/shogun`.


vigsterkr:
@duhduhdan oh yeah, science tap should be specified! thnx a lot!

PR #3826, opened 2017-06-01
Remove duplicate code in LDA/FisherLDA and port the solvers to use linalg (WIP)

micmn:
Starting point to discuss this task. #3482 

Both classes implement two solvers: canonvar (svd based) whose code is very similar in the two implementations except for the regularization parameter in LDA (this PR drops the LDA's solver), and the classic method using covariance matrix, LDA directly computes a solution for the binary classification problem while FisherLDA computes related eigenvectors, so this code isn't strictly _duplicated_ but of course one could use the latter to solve the former with some modification to FisherLDA (i.e. bias computation, but then it should be aware of treating the classification case differently).

A couple of other points:
- LDA supports a regularization parameter in both methods, so FisherLDA should too if we want to use its canonvar solver.
- LDA's solver is templated (float32/64/max)

### DIFF
--- a/src/shogun/classifier/LDA.cpp
+++ b/src/shogun/classifier/LDA.cpp
@@ -10,14 +10,17 @@
  */
 #include <shogun/lib/config.h>
 
-#include <shogun/lib/common.h>
-#include <shogun/machine/Machine.h>
-#include <shogun/machine/LinearMachine.h>
 #include <shogun/classifier/LDA.h>
-#include <shogun/labels/Labels.h>
 #include <shogun/labels/BinaryLabels.h>
+#include <shogun/labels/Labels.h>
+#include <shogun/lib/common.h>
+#include <shogun/machine/LinearMachine.h>
+#include <shogun/machine/Machine.h>
 #include <shogun/mathematics/Math.h>
 #include <shogun/mathematics/eigen3.h>
+#include <shogun/mathematics/linalg/LinalgNamespace.h>
+#include <shogun/preprocessor/FisherLDA.h>
+#include <vector>
 
 using namespace Eigen;
 using namespace shogun;
@@ -30,9 +33,10 @@ CLDA::CLDA(float64_t gamma, ELDAMethod method)
 	m_gamma=gamma;
 }
 
-CLDA::CLDA(float64_t gamma, CDenseFeatures<float64_t> *traindat,
-			CLabels *trainlab, ELDAMethod method)
-	:CLinearMachine(), m_gamma(gamma)
+CLDA::CLDA(
+    float64_t gamma, CDenseFeatures<float64_t>* traindat, CLabels* trainlab,
+    ELDAMethod method)
+    : CLinearMachine(), m_gamma(gamma)
 {
 	init();
 	set_features(traindat);
@@ -45,10 +49,12 @@ void CLDA::init()
 {
 	m_method=AUTO_LDA;
 	m_gamma=0;
-	SG_ADD((machine_int_t*) &m_method, "m_method",
-		"Method used for LDA calculation", MS_NOT_AVAILABLE);
-	SG_ADD((machine_int_t*) &m_gamma, "m_gamma",
-		"Regularization parameter", MS_NOT_AVAILABLE);
+	SG_ADD(
+	    (machine_int_t*)&m_method, "m_method",
+	    "Method used for LDA calculation", MS_NOT_AVAILABLE);
+	SG_ADD(
+	    (machine_int_t*)&m_gamma, "m_gamma", "Regularization parameter",
+	    MS_NOT_AVAILABLE);
 }
 
 CLDA::~CLDA()
@@ -58,8 +64,11 @@ CLDA::~CLDA()
 bool CLDA::train_machine(CFeatures *data)
 {
 	REQUIRE(m_labels, "Labels for the given features are not specified!\n")
-	REQUIRE(m_labels->get_label_type()==LT_BINARY, "The labels should of type"
-			" CBinaryLabels! you provided %s \n",m_labels->get_name())
+	REQUIRE(
+	    m_labels->get_label_type() == LT_BINARY,
+	    "The labels should of type"
+	    " CBinaryLabels! you provided %s \n",
+	    m_labels->get_name())
 
 	if(data)
 	{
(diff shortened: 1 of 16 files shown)


src/shogun/classifier/LDA.cpp
```diff
@@ -91,59 +93,81 @@ bool CLDA::train_machine(CFeatures *data)
...
+		for (i = 0; i < train_labels.vlen; i++)
+		{
+			if (train_labels.vector[i] == -1)
+				classidx_neg[num_neg++] = i;
+
+			else if (train_labels.vector[i] == +1)
+				classidx_pos[num_pos++] = i;
+		}
+
+		typename SGMatrix<ST>::EigenMatrixXt fmatrix =
```
vigsterkr: the problem with these eigen mapped matricies that in case the linalg backend is on gpu these will all fail, as feature matrix is not on the CPU's mem. :(

vigsterkr: we could address this later in a different commit, but we should at least mark here that it's a `TODO`

micmn: yeah this is something i wanted to ask you, the idea is to port everything to linalg?

vigsterkr: yep. if there's missing functionality it's worth to ask @OXPHOS how to go around implementing those.

vigsterkr:
@micmn after merging #3830 this requires a rebase

micmn:
I'm updating this, a few comments: I'm including some just sketched linalg's operations (svd, triangular solver, transpose), I'll open a pr to properly implement them, also it would be very useful to have a get/set_column method on sgmatrix and some sort of slicing/view so that one can do i.e.
`A.get_column(i)=vector`, `matrix_prod(A.cols(0,3), B)` (maybe in this way we can also get rid of `add_col_vec()`).

This also fixes a wrong mean calculation (https://github.com/shogun-toolbox/shogun/blob/270ff9200daeb273bec6c82543d82d5a275b39d3/src/shogun/preprocessor/FisherLDA.cpp#L154)
and it removes a useless qr factorization (https://github.com/shogun-toolbox/shogun/blob/270ff9200daeb273bec6c82543d82d5a275b39d3/src/shogun/preprocessor/FisherLDA.cpp#L215)

(docs are outdated)

micmn:
Update: regularization in the eigen-based FisherLDA's solver is missing (but it's not used by LDA) and also templated solver, I'll address this in a separate pr.

karlnapf:
Nice to clean this up. I am always in favour of

* OOP to avoid duplicate code, i.e. one class per solver. The alternative is to do it like in the `CPCA` class where an enum is passed. This depends a  bit on the situation (what do you suggest here?)
* I think templating the algorithm class is not a good idea. Algorithms accept CFeatures which knows the type. `CLeastAngleRegression` (LARS) does some hacks inside to infer the type, but that solution is not optimal either. In order to solve this properly, a lot of work is needed, so for now I would replicate the thing in LARS. Thoughts @vigsterkr ?
* When you touch code like this, make sure to use linalg where possible. This will make it much easier in the future to maintain this code.
* Methods should not contain more than 10-30 lines of code (my opinion)

micmn:
Thanks for the feedback.

>  OOP to avoid duplicate code, i.e. one class per solver. The alternative is to do it like in the CPCA class where an enum is passed. This depends a bit on the situation (what do you suggest here?)

I'm for one class per solver since there are three lda classes (binary, multiclass, dim. reduction), so that each class calls the solver and that's it.

>  I think templating the algorithm class is not a good idea. Algorithms accept CFeatures which knows the type. CLeastAngleRegression (LARS) does some hacks inside to infer the type, but that solution is not optimal either.

It seems that LDA is using the same hack already (train_machine_templated).

karlnapf:
whats the state here? neeeds more review?

micmn:
Yeah review please, at the state of the art we're losing the svd solver for float32/max if we don't implement the templated solver also in FisherLDA, but maybe we could address this when we have a first working version of the template-free Matrix and also create a class for each solver (and in the meantime we issue a warning)?

vigsterkr:
@micmn let's rebase :)

src/shogun/classifier/LDA.cpp
```diff
@@ -91,67 +94,81 @@ bool CLDA::train_machine(CFeatures *data)
...
-	int32_t num_neg=0;
-	int32_t num_pos=0;
+{
+	SGMatrix<ST> feature_matrix =
+	    ((CDenseFeatures<ST>*)features)->get_feature_matrix();
+	index_t num_feat = feature_matrix.num_rows;
+	index_t num_vec = feature_matrix.num_cols;
 
-	for(i=0; i<train_labels.vlen; i++)
+	if (m_method == SVD_LDA || (m_method == AUTO_LDA && num_vec <= num_feat))
```
karlnapf: You could make this more readable via pulling the `m_method == AUTO_LDA && num_vec <= num_feat` out into a variable and call it `lda_more_efficient`

src/shogun/classifier/LDA.cpp
```diff
@@ -91,67 +94,81 @@ bool CLDA::train_machine(CFeatures *data)
...
-		mean_neg+=fmatrix.col(classidx_neg[i]);
-	mean_neg/=(ST)num_neg;
+		for (index_t i = 0; i < m_labels->get_num_labels(); ++i)
+			multiclass_labels->set_int_label(
+			    i, (((CBinaryLabels*)m_labels)->get_int_label(i) == 1 ? 1 : 0));
 
-	// get m(-ve) - mean(-ve)
-	for(i=0; i<num_neg; i++)
-		fmatrix.col(classidx_neg[i])-=mean_neg;
+		lda->fit(data, multiclass_labels.get(), 1);
```
karlnapf: whats the `1`?

src/shogun/classifier/LDA.cpp
```diff
@@ -91,67 +94,81 @@ bool CLDA::train_machine(CFeatures *data)
...
+		auto m = lda->get_transformation_matrix();
 
-	//mean pos
-	for(i=0; i<num_pos; i++)
-		mean_pos+=fmatrix.col(classidx_pos[i]);
-	mean_pos/=(ST)num_pos;
+		SGVector<float64_t> w(num_feat);
+		for (index_t i = 0; i < num_feat; ++i)
```
karlnapf: please no loop

src/shogun/classifier/LDA.cpp
```diff
@@ -91,67 +94,81 @@ bool CLDA::train_machine(CFeatures *data)
...
-	SGMatrix<ST>scatter_matrix(num_feat, num_feat);
-	typename SGMatrix<ST>::EigenMatrixXtMap scatter(scatter_matrix.matrix, num_feat, num_feat);
-
-	if (m_method == FLD_LDA || (m_method==AUTO_LDA && num_vec>num_feat))
+		set_w(w);
+		set_bias(-linalg::dot(w, lda->get_mean()));
+	}
+	else
 	{
+		std::vector<index_t> classidx_neg;
```
karlnapf: Can we put the two branches of this if else branch into methods?
Makes the whole thing less bulky

src/shogun/classifier/LDA.cpp
```diff
@@ -91,67 +94,81 @@ bool CLDA::train_machine(CFeatures *data)
...
+		std::vector<index_t> classidx_pos;
```
karlnapf: `idx_pos` is a better name imho

src/shogun/classifier/LDA.cpp
```diff
@@ -160,105 +177,26 @@ bool CLDA::train_machine_templated(SGVector<int32_t> train_labels, CFeatures *da
 		// MatrixXd A=scatter;
 		// VectorXd b=mean_pos-mean_neg;
 		// VectorXd x=w;
-		typename SGVector<ST>::EigenVectorXtMap x(w_st.vector, num_feat);
-		LLT<typename SGMatrix<ST>::EigenMatrixXt> decomposition(scatter);
-		x=decomposition.solve(mean_pos-mean_neg);
+		auto decomposition = linalg::cholesky_factor(scatter_matrix);
```
karlnapf: this might give a numerical error?

src/shogun/classifier/LDA.cpp
```diff
@@ -160,105 +177,26 @@ bool CLDA::train_machine_templated(SGVector<int32_t> train_labels, CFeatures *da
...
+		SGVector<float64_t> w(num_feat);
+		// copy w_st into w
+		for (index_t i = 0; i < w.size(); ++i)
+			w[i] = (float64_t)w_st[i];
+		set_w(w);
 
 		// get the bias.
-		bias=((float64_t)(0.5*(w_neg.dot(mean_neg)-w_pos.dot(mean_pos))));
+		set_bias(
+		    (float64_t)(
```
karlnapf: formatting out of controll.... ;)

src/shogun/classifier/LDA.cpp
```diff
@@ -160,105 +177,26 @@ bool CLDA::train_machine_templated(SGVector<int32_t> train_labels, CFeatures *da
...
+		        0.5 *
+		        (linalg::dot(w_neg, mean_neg) - linalg::dot(w_pos, mean_pos))));
 	}
-	else
-	{
-		//for algorithmic detail, please refer to section 16.3.1. of Bayesian
```
karlnapf: this comment should maybe go to the `@brief` class API docs

src/shogun/preprocessor/FisherLDA.cpp
```diff
@@ -109,175 +113,207 @@ bool CFisherLDA::fit(CFeatures *features, CLabels *labels, int32_t num_dimension
 
 	REQUIRE(C>1, "At least two classes are needed to perform LDA.\n")
 
-	int32_t i=0;
-	int32_t j=0;
-
 	m_num_dim=num_dimensions;
 	// max target dimension allowed.
-	// int32_t max_dim_allowed=C-1;
+	// index_t max_dim_allowed=C-1;
```
karlnapf: remove this line ?

src/shogun/preprocessor/FisherLDA.cpp
```diff
@@ -109,175 +113,207 @@ bool CFisherLDA::fit(CFeatures *features, CLabels *labels, int32_t num_dimension
...
 
 	// clip number if Dimensions to be a valid number
 	if ((m_num_dim<=0) || (m_num_dim>(C-1)))
 		m_num_dim=(C-1);
 
-	MatrixXd fmatrix=Map<MatrixXd>(feature_matrix.matrix, num_features,
-									num_vectors);
-	Map<VectorXd> lvector(labels_vector.vector, num_vectors);
+	SGMatrix<float64_t> feature_centered = feature_matrix.clone();
```
karlnapf: are you carefully thinking about the memory footprint here?

src/shogun/preprocessor/FisherLDA.cpp
```diff
@@ -109,175 +113,207 @@ bool CFisherLDA::fit(CFeatures *features, CLabels *labels, int32_t num_dimension
...
-						=fmatrix.col(j);
-			Sw+=(centered_class_i[i]*centered_class_i[i].transpose())
-				*num_class[i]/(float64_t)(num_class[i]-1);
+			centered_class[i] = SGMatrix<float64_t>(num_features, num_class[i]);
+			linalg::zero(centered_class[i]);
+		}
+		for (index_t i = 0; i < num_vectors; i++)
+		{
+			index_t c = (index_t)labels_vector[i];
+			// TODO: SGMatrix.set_column(i, v)
```
karlnapf: please no todos :)

src/shogun/preprocessor/FisherLDA.cpp
```diff
@@ -109,175 +113,207 @@ bool CFisherLDA::fit(CFeatures *features, CLabels *labels, int32_t num_dimension
...
+		linalg::identity(id);
+		linalg::add(
+		    Sw, id, Sw, (1.0 - m_gamma), trace * (m_gamma) / num_features);
+
+		// within class matrix for canonical variates implementation
+		SGMatrix<float64_t> Sb(num_features, C);
+		linalg::zero(Sb);
+		for (index_t i = 0; i < C; i++)
+		{
+			// TODO: SGMatrix.get_column(i) -> SGVector
```
karlnapf: no todo pls

src/shogun/preprocessor/FisherLDA.cpp
```diff
@@ -109,175 +113,207 @@ bool CFisherLDA::fit(CFeatures *features, CLabels *labels, int32_t num_dimension
...
+		}
 
-		JacobiSVD<MatrixXd> svd(fmatrix1, ComputeThinU | ComputeThinV);
+		index_t r = CMath::min(num_vectors, num_features);
+		SGMatrix<float64_t> U(num_features, r);
+		SGVector<float64_t> singularValues(r);
+		linalg::svd(feature_matrix, singularValues, U);
 		// basis to represent the solution
-		MatrixXd Q;
+		SGMatrix<float64_t> Q;
```
karlnapf: is this design fixed in the way that no matter what the input features are, Q is float64? I am ok, just asking

src/shogun/preprocessor/FisherLDA.cpp
```diff
@@ -109,175 +113,207 @@ bool CFisherLDA::fit(CFeatures *features, CLabels *labels, int32_t num_dimension
...
 
 		if(num_features>num_vectors)
 		{
-			j=0;
-			for (i=0;i<num_vectors;i++)
-				if (svd.singularValues()(i)>m_threshold)
+			index_t j = 0;
+			for (index_t i = 0; i < num_vectors; i++)
```
karlnapf: maybe can have a method `count_threshold_exceed(T thresh, bool upper)`?

src/shogun/preprocessor/FisherLDA.cpp
```diff
@@ -109,175 +113,207 @@ bool CFisherLDA::fit(CFeatures *features, CLabels *labels, int32_t num_dimension
...
+			Q = SGMatrix<float64_t>(U.matrix, num_features, j, false);
 		}
 		else
-			Q=svd.matrixU();
+			Q = U;
 
 		// Sb is the modified between scatter
-		Sb=(Q.transpose())*Sb*(Sb.transpose())*Q;
+		auto aux = linalg::matrix_prod(Q, Sb, true, false);
+		Sb = linalg::matrix_prod(aux, aux, false, true);
```
karlnapf: @vigsterkr @micmn we could have the new `Matrix` class have a boolean flag that denotes transpose, so that these ugly true/false parameters of the matrix_prod call are not there? Thoughts?

src/shogun/preprocessor/FisherLDA.cpp
```diff
@@ -109,175 +113,207 @@ bool CFisherLDA::fit(CFeatures *features, CLabels *labels, int32_t num_dimension
...
+		m_transformation_matrix = SGMatrix<float64_t>(num_features, m_num_dim);
+		linalg::zero(m_transformation_matrix);
 		m_eigenvalues_vector=SGVector<float64_t>(m_num_dim);
-		Map<VectorXd> eigenValues (m_eigenvalues_vector.vector, m_num_dim);
-		eigenValues=svd2.singularValues().topRows(m_num_dim);
+
+		auto Wt = linalg::matrix_prod(Q, W);
+		for (index_t i = 0; i < m_num_dim; ++i)
+		{
+			// TODO: SGMatrix.set_column(i, v)
```
karlnapf: no todo pls

src/shogun/preprocessor/FisherLDA.cpp
```diff
@@ -109,175 +113,207 @@ bool CFisherLDA::fit(CFeatures *features, CLabels *labels, int32_t num_dimension
...
 
-		for (i=0; i<m_num_dim; i++)
+		auto args = CMath::argsort(eigenvalues);
+		for (index_t i = 0; i < m_num_dim; i++)
 		{
-			eigenValues[i]=data[num_features-i-1].first;
-			eigenVectors.col(i)=all_eigenvectors.col(data[num_features-i-1].second);
+			index_t k = args[num_features - i - 1];
+			m_eigenvalues_vector[i] = eigenvalues[k];
+			// TODO: SGMatrix.set_column(i, v)
```
karlnapf: todo

src/shogun/preprocessor/FisherLDA.cpp
micmn: like a method Matrix.transpose() that flips the flag?
then every linalg method should check its status or we should find another way, first thought: when we cast to EigenMatrixXtMap we also check the flag

karlnapf: yes

PR #3827, opened 2017-06-01
Fix check_format.sh script when the destination branch is not develop.

geektoni:


geektoni:
@vigsterkr this should fix #3745 errors.

PR #3828, opened 2017-06-01
[ProgressBar] Add multi threaded progress bar.

geektoni:


src/shogun/base/progress.h
```diff
@@ -260,12 +283,19 @@ namespace shogun
 #endif
 		}
 
+		void increment() const
+		{
+			m_current_value.store(m_current_value.load() + 1);
```
vigsterkr: operator++ works and it's atomic:
http://en.cppreference.com/w/cpp/atomic/atomic/operator_arith

PR #3824, opened 2017-05-29
Replace DynArray with std::vector

MikeLing:

PR #3832, opened 2017-06-04
use std::vector instead of DynArray(on going)

MikeLing:
So this PR is a part of DynArray refactoring(https://github.com/shogun-toolbox/shogun/pull/3824), the reason why we don't use std::vector here is because [1] and [2]. 

[1] https://stackoverflow.com/a/17794965
[2] https://stackoverflow.com/a/8399470

### DIFF
--- a/src/shogun/lib/DynamicArray.h
+++ b/src/shogun/lib/DynamicArray.h
@@ -13,30 +13,35 @@
 
 #include <shogun/lib/config.h>
 
-#include <shogun/base/SGObject.h>
 #include <shogun/base/DynArray.h>
 #include <shogun/base/Parameter.h>
+#include <shogun/base/SGObject.h>
+#include <vector>
 
 namespace shogun
 {
-/** @brief Template Dynamic array class that creates an array that can
- * be used like a list or an array.
- *
- * It grows and shrinks dynamically, while elements can be accessed
- * via index.  It is performance tuned for simple types like float
- * etc. and for hi-level objects only stores pointers, which are not
- * automagically SG_REF'd/deleted.
- */
-template <class T> class CDynamicArray :public CSGObject
-{
+	/** @brief Template Dynamic array class that creates an array that can
+	 * be used like a list or an array.
+	 *
+	 * It grows and shrinks dynamically, while elements can be accessed
+	 * via index.  It is performance tuned for simple types like float
+	 * etc. and for hi-level objects only stores pointers, which are not
+	 * automagically SG_REF'd/deleted.
+	 */
+
+	template <class T> class CDynamicArray : public CSGObject
+	{
 	public:
 		/** default constructor */
-		CDynamicArray()
-		: CSGObject(), m_array()
+		CDynamicArray() : CSGObject()
 		{
 			dim1_size=1;
 			dim2_size=1;
 			dim3_size=1;
+			num_elements = 0;
+			free_array = true;
+			resize_granularity = 128;
+			m_array = std::vector<T>(resize_granularity);
 
 			init();
 		}
@@ -47,13 +52,18 @@ template <class T> class CDynamicArray :public CSGObject
 		 * @param p_dim2_size dimension 2
 		 * @param p_dim3_size dimension 3
 		 */
-		CDynamicArray(int32_t p_dim1_size, int32_t p_dim2_size=1, int32_t p_dim3_size=1)
-		: CSGObject(), m_array(p_dim1_size*p_dim2_size*p_dim3_size)
+		CDynamicArray(
+		    int32_t p_dim1_size, int32_t p_dim2_size = 1,
+		    int32_t p_dim3_size = 1)
+		    : CSGObject()
 		{
 			dim1_size=p_dim1_size;
 			dim2_size=p_dim2_size;
 			dim3_size=p_dim3_size;
-
+			num_elements = 0;
+			free_array = true;
+			resize_granularity = p_dim1_size * p_dim2_size * p_dim3_size;
+			m_array = std::vector<T>(resize_granularity);
 			init();
 		}
 
@@ -64,13 +74,19 @@ template <class T> class CDynamicArray :public CSGObject
 		 * @param p_free_array if array must be freed
 		 * @param p_copy_array if array must be copied
 		 */
-		CDynamicArray(T* p_array, int32_t p_dim1_size, bool p_free_array, bool p_copy_array)
-		: CSGObject(), m_array(p_array, p_dim1_size, p_free_array, p_copy_array)
+		CDynamicArray(
+		    T* p_array, int32_t p_dim1_size, bool p_free_array,
+		    bool p_copy_array)
+		    : CSGObject()
 		{
 			dim1_size=p_dim1_size;
 			dim2_size=1;
 			dim3_size=1;
 
+			free_array = p_free_array;
+			num_elements = p_dim1_size;
+			resize_granularity = 128;
+			m_array.assign(p_array, p_array + (p_dim1_size));
 			init();
 		}
 
@@ -82,14 +98,18 @@ template <class T> class CDynamicArray :public CSGObject
 		 * @param p_free_array if array must be freed
 		 * @param p_copy_array if array must be copied
 		 */
-		CDynamicArray(T* p_array, int32_t p_dim1_size, int32_t p_dim2_size,
-						bool p_free_array, bool p_copy_array)
-		: CSGObject(), m_array(p_array, p_dim1_size*p_dim2_size, p_free_array, p_copy_array)
+		CDynamicArray(
+		    T* p_array, int32_t p_dim1_size, int32_t p_dim2_size,
+		    bool p_free_array, bool p_copy_array)
+		    : CSGObject()
 		{
 			dim1_size=p_dim1_size;
 			dim2_size=p_dim2_size;
 			dim3_size=1;
-
+			free_array = p_free_array;
+			num_elements = p_dim1_size * p_dim2_size;
+			resize_granularity = 128;
+			m_array.assign(p_array, p_array + (p_dim1_size * p_dim2_size));
 			init();
 		}
 
@@ -102,14 +122,19 @@ template <class T> class CDynamicArray :public CSGObject
 		 * @param p_free_array if array must be freed
 		 * @param p_copy_array if array must be copied
 		 */
-		CDynamicArray(T* p_array, int32_t p_dim1_size, int32_t p_dim2_size,
-						int32_t p_dim3_size, bool p_free_array, bool p_copy_array)
-		: CSGObject(), m_array(p_array, p_dim1_size*p_dim2_size*p_dim3_size, p_free_array, p_copy_array)
+		CDynamicArray(
+		    T* p_array, int32_t p_dim1_size, int32_t p_dim2_size,
+		    int32_t p_dim3_size, bool p_free_array, bool p_copy_array)
+		    : CSGObject()
 		{
 			dim1_size=p_dim1_size;
 			dim2_size=p_dim2_size;
 			dim3_size=p_dim3_size;
-
+			free_array = p_free_array;
+			num_elements = p_dim1_size * p_dim2_size * p_dim3_size;
+			resize_granularity = 128;
+			m_array.assign(
+			    p_array, p_array + (p_dim1_size * p_dim2_size * p_dim3_size));
 			init();
 		}
 
@@ -120,13 +145,18 @@ template <class T> class CDynamicArray :public CSGObject
 		 * @param p_dim2_size dimension 2
 		 * @param p_dim3_size dimension 3
 		 */
-		CDynamicArray(const T* p_array, int32_t p_dim1_size=1, int32_t p_dim2_size=1, int32_t p_dim3_size=1)
-		: CSGObject(), m_array(p_array, p_dim1_size*p_dim2_size*p_dim3_size)
+		CDynamicArray(
+		    const T* p_array, int32_t p_dim1_size = 1, int32_t p_dim2_size = 1,
+		    int32_t p_dim3_size = 1)
+		    : CSGObject()
 		{
 			dim1_size=p_dim1_size;
 			dim2_size=p_dim2_size;
 			dim3_size=p_dim3_size;
-
+			num_elements = p_dim1_size * p_dim2_size * p_dim3_size;
+			resize_granularity = 128;
+			m_array.assign(
+			    p_array, p_array + (p_dim1_size * p_dim2_size * p_dim3_size));
 			init();
 		}
 
@@ -139,7 +169,8 @@ template <class T> class CDynamicArray :public CSGObject
 		 */
 		inline int32_t set_granularity(int32_t g)
 		{
-			return m_array.set_granularity(g);
+			resize_granularity = g;
+			return resize_granularity;
 		}
 
 		/** get array size (including granularity buffer)
@@ -148,7 +179,7 @@ template <class T> class CDynamicArray :public CSGObject
 		 */
 		inline int32_t get_array_size()
 		{
-			return m_array.get_array_size();
+			return m_array.size();
 		}
 
 		/** return 2d array size
@@ -199,7 +230,7 @@ template <class T> class CDynamicArray :public CSGObject
 		 */
 		inline int32_t get_num_elements() const
 		{
-			return m_array.get_num_elements();
+			return num_elements;
 		}
 
 		/** get array element at index
@@ -209,9 +240,10 @@ template <class T> class CDynamicArray :public CSGObject
 		 * @param idx3 index 3
 		 * @return array element at index
 		 */
-		inline const T& get_element(int32_t idx1, int32_t idx2=0, int32_t idx3=0) const
+		inline typename std::vector<T>::const_reference
+		get_element(int32_t idx1, int32_t idx2 = 0, int32_t idx3 = 0) const
 		{
-			return m_array.get_array()[idx1+dim1_size*(idx2+dim2_size*idx3)];
+			return m_array[idx1 + dim1_size * (idx2 + dim2_size * idx3)];
 		}
 
 		/** get array element at index
@@ -221,7 +253,8 @@ template <class T> class CDynamicArray :public CSGObject
 		 * @param idx3 index 3
 		 * @return array element at index
 		 */
-		inline const T& element(int32_t idx1, int32_t idx2=0, int32_t idx3=0) const
+		inline typename std::vector<T>::const_reference
+		element(int32_t idx1, int32_t idx2 = 0, int32_t idx3 = 0) const
 		{
 			return get_element(idx1, idx2, idx3);
 		}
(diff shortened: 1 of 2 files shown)


MikeLing:
@lisitsyn could you give me some feedback for this pr? why so many error comes out but not directly relate to DynArray?

MikeLing:
https://github.com/shogun-toolbox/shogun/pull/3833

MikeLing:
sorry @vigsterkr , I failed to solve the error in SplittingStrategy.standard test. I had update the pr and wise you could give me some guide on that issue and pr itself. Maybe it's just a little error and i somehow miss it. But I'm too tired to move on today. I will keep working on it until it's done. :)

BTW, learn a lot of things from this pr and I wise I could use them in the std::vector<T> refactoring. Sorry, I went into an wrong direction and waste a bunch of time. And, I waste a lot of time in building and rebuilding :(

src/shogun/evaluation/StratifiedCrossValidationSplitting.cpp
```diff
@@ -124,7 +124,7 @@ void CStratifiedCrossValidationSplitting::build_subsets()
 				label_indices.get_element(i);
 
 		// external random state important for threads
-		current->shuffle(m_rng);
+		current->shuffle();
```
vigsterkr: in this case you can remove `m_rng` from `CStratifiedCrossValidationSplitting` as this is the only place it is being used in the class.

src/shogun/lib/DynamicArray.h
```diff
@@ -316,17 +330,35 @@ template <class T> class CDynamicArray :public CSGObject
...
+			}
 		}
 
 		/** append array element to the end of array
 		 *
 		 * @param e element to append
 		 * @return if setting was successful
 		 */
-		inline bool append_element(T e)
+		inline void append_element(T e)
```
vigsterkr: i wonder why do you change the return type of the function?
and you kept the doxygen comment which is then now confusing.... please just keep the return value `bool`

src/shogun/lib/DynamicArray.h
```diff
@@ -316,17 +330,35 @@ template <class T> class CDynamicArray :public CSGObject
...
-			return m_array.append_element(e);
+			if (num_elements < m_array.size())
+			{
+				m_array[num_elements] = e;
+			}
+			else
+			{
+				m_array.push_back(e);
+			}
+			num_elements++;
```
vigsterkr: could you use everwhere `++<variable>` instead of `<variable>++`

src/shogun/lib/DynamicArray.h
```diff
@@ -335,14 +367,25 @@ template <class T> class CDynamicArray :public CSGObject
...
+		}
 
 		/** STD VECTOR compatible. Delete array element at the end
 		 *  of array.
 		 */
 		inline void pop_back()
 		{
 			m_array.pop_back();
+			num_elements--;
```
vigsterkr: same as in case of incrementing, could you use `--<variable>`

src/shogun/lib/DynamicArray.h
```diff
@@ -540,10 +645,13 @@ template <class T> class CDynamicArray :public CSGObject
...
 		/** shuffles the array (not thread safe!) */
-		inline void shuffle() { m_array.shuffle(); }
+		inline void shuffle()
+		{
+			std::random_shuffle(m_array.begin(), m_array.end());
+		}
 
 		/** shuffles the array with external random state */
-		inline void shuffle(CRandom * rand) { m_array.shuffle(rand); }
+		// inline void shuffle(CRandom * rand) { m_array.shuffle(rand); }
```
vigsterkr: replace this function with:
```
template<class RandomFunc>
inline void shuffle(RandomFunc&& r)
{
std::random_shuffle(m_array.begin(), m_array.end(), r);
}
```
because we want to be able to fix the seed of the shuffle

src/shogun/lib/DynamicArray.h
```diff
@@ -622,20 +731,13 @@ template <class T> class CDynamicArray :public CSGObject
...
-							  "shrink/grow step size.", MS_NOT_AVAILABLE);
-			SG_ADD(&m_array.use_sg_mallocs,
-							  "use_sg_malloc",
-							  "whether SG_MALLOC or malloc should be used",
-							  MS_NOT_AVAILABLE);
-			SG_ADD(&m_array.free_array,
-							  "free_array",
-							  "whether array must be freed",
-							  MS_NOT_AVAILABLE);
+			SG_ADD(
```
vigsterkr: by removing
```
m_parameters->add_vector(&m_array.array, &m_array.current_num_elements, "array", "Memory for dynamic array.");
```
basically it means that the serialization of this object will not be correct... as the deque is not going to be serialized.
it would be easy to add a vector element to `m_parameters` if we would use `std::vector` as that is a continuous memory block... unfortunately it's not the case of deque. but we still need a way to be able to serialize deque's content :S

@lisitsyn btw how is this gonna be handled with tags? `m_parameters->add_vector`

vigsterkr: this is why some of the python modular tests fails

src/shogun/lib/DynamicArray.h
```diff
@@ -405,9 +468,10 @@ template <class T> class CDynamicArray :public CSGObject
 		 *
 		 * @return the array
 		 */
-		inline T* get_array() const
+		inline T* get_array()
 		{
-			return m_array.get_array();
+			auto it = m_array.begin();
+			return &(*it);
```
vigsterkr: i'm quite sure that this is not correct, as elements in deque are not contiguous in memory.

vigsterkr: namely i dont think that this can be made work with deque, unless you copy the content from it to a vector temporarily and return that vector's memory region, but then how would you release the memory...

vigsterkr:
or of course there's an option to actually not return the array itself but only an iterator... and refactor some parts of shogun to use iterator to copy CDynamicArray... this of course still does not solve our issue with serialization.

src/shogun/lib/DynamicArray.h
```diff
@@ -13,30 +13,872 @@
...
+			// registration) the cloned version has less memory
+			// allocated than known to dynarray. We fix this here.
+			// cloned->num_elements = cloned->array_size;
+			return cloned;
+		}
+
+	private:
+		/** register parameters */
+		virtual void init()
+		{
```
MikeLing: @vigsterkr Hi, could you tell me something about set_generic<>()?
I got bunch of error like
```

/Users/mikeling/shogun/build/src/shogun/base/class_list.cpp:815:110: error: use
      of class template 'CDynamicArray' requires template arguments
  ...g) { return g == PT_NOT_GENERIC? new CDynamicArray(): NULL; }
                                          ^
/Users/mikeling/shogun/src/shogun/base/DynArray.h:34:33: note: template is
      declared here
        template<class U> friend class CDynamicArray;
        ~~~~~~~~~~~~~~~~~              ^
/Users/mikeling/shogun/build/src/shogun/base/class_list.cpp:815:131: error: 
      expected ':'
  ...g) { return g == PT_NOT_GENERIC? new CDynamicArray(): NULL; }
                                                               ^
                                                               : 
/Users/mikeling/shogun/build/src/shogun/base/class_list.cpp:815:104: note: to
      match this '?'
  ...g) { return g == PT_NOT_GENERIC? new CDynamicArray(): NULL; }
                                    ^
/Users/mikeling/shogun/build/src/shogun/base/class_list.cpp:815:131: error: 
      expected expression
  ...g) { return g == PT_NOT_GENERIC? new CDynamicArray(): NULL; }
```

src/shogun/lib/DynamicArray.h
```diff
@@ -13,30 +13,875 @@
...
+{
+	/** @brief Template Dynamic array class that creates an array that can
+	 * be used like a list or an array.
+	 *
+	 * It grows and shrinks dynamically, while elements can be accessed
+	 * via index.  It is performance tuned for simple types like float
+	 * etc. and for hi-level objects only stores pointers, which are not
+	 * automagically SG_REF'd/deleted.
+	 */
+	template <>
```
vigsterkr: i know that this follows the clang-format (or i guess that's why it's formatted like this) but our `class_list.cpp.py` doesn't know how to handle this, i.e. template definition should be in one line:
`template <> class CDynamicArray<bool> : public CSGObject`

vigsterkr: this is true for the real template definition as well, i.e. it should be:
```template <class T> class CDynamicArray : public CSGObject```

instead of
```
       template <class T>
       class CDynamicArray : public CSGObject
```
this will fix the problem of `class_list.cpp`

File: src/shogun/lib/DynamicArray.h [hunk 10]
```diff
@@ -13,30 +13,875 @@
...
+			}
+		}
+
+		/** insert array element at index
+		 *
+		 * @param e element to insert
+		 * @param index index
+		 * @return if setting was successful
+		 */
+		inline bool insert_element(bool element, int32_t index)
```
vigsterkr: `element` shadows `this.element`:
```../src/shogun/lib/DynamicArray.h: In member function ‘bool shogun::CDynamicArray<bool>::insert_element(bool, int32_t)’:
../src/shogun/lib/DynamicArray.h:397:3: warning: declaration of ‘element’ shadows a member of 'this' [-Wshadow]``

please just use a `e` instead of `element`

File: src/shogun/lib/DynamicArray.h [hunk 11]
```diff
@@ -13,30 +13,875 @@
...
+			return get_element(get_num_elements() - 1);
+		}
+
+		/** find first occurence of array element and return its index
+		 * or -1 if not available
+		 *
+		 * @param e element to search for
+		 * @return index of element or -1
+		 */
+		inline int32_t find_element(bool element)
```
vigsterkr: same as in case of `insert_element`:
```
../src/shogun/lib/DynamicArray.h: In member function ‘int32_t shogun::CDynamicArray<bool>::find_element(bool)’:
../src/shogun/lib/DynamicArray.h:466:3: warning: declaration of ‘element’ shadows a member of 'this' [-Wshadow]
```
use just `e` as function arg var.

vigsterkr:
based on gdb of `GaussianProcessClassification.apply_preprocessor_and_binary` the problem is of course is in `shogun::CDynamicArray<bool>::set_element`; it seems it goes into an infinite recursive callstack as there are more than 20000+ set_element calls on the stack.... please try to copy-paste code without **ERRORS**

vigsterkr:
@MikeLing one more thing. please dont even push any commits until you dont have `ctest` running successfully locally...

vigsterkr:
@MikeLing would be great to add unit-tests in the other PR to cover these errors, so that next time if there's a refactor the problem is easily caught

src/shogun/lib/DynamicArray.h
```diff
@@ -13,30 +13,875 @@
...
+		 *
+		 * @param p_array new array
+		 * @param p_num_elements last element index + 1
+		 * @param array_size number of elements in array
+		 */
+		inline void
+		set_array(bool* p_array, int32_t p_num_elements, int32_t p_array_size)
+		{
+			if (m_array != NULL && free_array)
+				SG_FREE(m_array);
```
MikeLing: I had remove all the use_sg_mallocs logic in this push :)

src/shogun/lib/DynamicArray.h
```diff
@@ -13,30 +13,875 @@
...
+		/** shuffles the array (not thread safe!) */
+		inline void shuffle()
+		{
+			for (index_t i = 0; i <= num_elements - 1; ++i)
+				CMath::swap(
+				    m_array[i], m_array[CMath::random(i, num_elements - 1)]);
+		}
+
+		/** shuffles the array with external random state */
+		inline void shuffle(CRandom* rand)
```
MikeLing: mmm, but after look more into it, I found maybe use `shuffle(sg_rand)` in https://github.com/shogun-toolbox/shogun/blob/f7bccee6251d15f5cfee0514c6d22d3bcf5f7a40/src/shogun/evaluation/StratifiedCrossValidationSplitting.cpp#L155 is a better idea. Sorry for remove it before

src/shogun/evaluation/StratifiedCrossValidationSplitting.cpp
```diff
@@ -145,12 +143,11 @@ void CStratifiedCrossValidationSplitting::build_subsets()
 			target_set%=m_subset_indices->get_num_elements();
 			SG_UNREF(next);
 		}
-
 		SG_UNREF(current);
```
MikeLing: Oh, I forget to restore these lines in here. I will do it after the rest of the pr works good.

src/shogun/lib/DynamicArray.h
vigsterkr: ok... anyhow we can refactor this in your next task as that one will heavily touch the random of shogun itself. but we'll discuss that later once DynArray removal is fully done.

src/shogun/lib/DynamicArray.h
```diff
@@ -139,7 +169,7 @@ template <class T> class CDynamicArray :public CSGObject
 		 */
 		inline int32_t set_granularity(int32_t g)
 		{
-			return m_array.set_granularity(g);
+			return resize_granularity = g;
```
MikeLing: Damn...This should been fixed :/

src/shogun/lib/DynamicArray.h
```diff
@@ -280,7 +316,7 @@ template <class T> class CDynamicArray :public CSGObject
 		 */
 		inline T get_last_element() const
 		{
-			return m_array.get_last_element();
+			return m_array.back();
```
MikeLing: Mark, this should be `m_array[num_elements - 1]`

src/shogun/lib/DynamicArray.h
```diff
@@ -335,14 +390,25 @@ template <class T> class CDynamicArray :public CSGObject
...
+			++num_elements;
+		}
 
 		/** STD VECTOR compatible. Delete array element at the end
 		 *  of array.
 		 */
 		inline void pop_back()
 		{
 			m_array.pop_back();
+			--num_elements;
```
MikeLing: I guess this should be m_array[num_element -1] = NULL

vigsterkr: rather `std::vector.erase` no?

src/shogun/lib/DynamicArray.h
```diff
@@ -611,7 +764,7 @@ template <class T> class CDynamicArray :public CSGObject
 			// current_num_elements as size (see parameter
 			// registration) the cloned version has less memory
 			// allocated than known to dynarray. We fix this here.
-			cloned->m_array.num_elements = cloned->m_array.current_num_elements;
+			// cloned->num_elements = cloned->m_array.size();
```
MikeLing: This one the another place may broken serialization. But it will broken SGObject if not comment it. Maybe we need a better way to register parameter.

src/shogun/lib/DynamicArray.h
```diff
@@ -621,21 +774,13 @@ template <class T> class CDynamicArray :public CSGObject
 		virtual void init()
 		{
 			set_generic<T>();
+			T* head = m_array.data();
+			m_parameters->add_vector(
+			    &head, &num_elements, "array", "Memory for dynamic array.");
```
vigsterkr: i would use `m_array.size()` instead of `num_elements` for the size of vector, as you actually wanna have the same vector deserialized, right? meaning if you have `m_array.size() = 15` but `num_elements = 5`, you still wanna have the same sized `m_array` deserialized...so
```m_parameters->add_vector(&head, &(m_array.size()), ....```

Comment (MikeLing) [inline src/shogun/lib/DynamicArray.h]: OK! Thank you! I will try it before next push. I'm working on the DynamicObjectArray refactor, so I can test if all these implementation on DynamicArray can be implemented on DynamicObjectArray also. Also can helps me polish this pr at the same time

File: src/shogun/lib/DynamicArray.h (hunk 17 above)
Comment (MikeLing) [inline src/shogun/lib/DynamicArray.h]: yes, you are right!

Comment (vigsterkr) [inline src/shogun/lib/DynamicArray.h]: i mean i'm not so sure how was this previously done... was the size of the array changed when deleted? if not then of course you need to nullify the element... ```nullptr``` only works if the stored element of DynamicArray is a pointer right? i.e. if it's index_t or other primitive time you have to come up with a different value. but then again `--num_elements` should be enough w/o touching the array itself as you shouldnt be able to read elements past who's index is `index >= num_elements`

File: src/shogun/lib/DynamicArray.h (hunk 19 above)
Comment (MikeLing) [inline src/shogun/lib/DynamicArray.h]: Hi @vigsterkr and @lisitsyn , what's the relationship between these parameters and `SGObject::clone()`? All these parameters will be clone  into generate a new Object? And serialization ?

Comment (karlnapf) [inline src/shogun/lib/DynamicArray.h]: Yes, everything that is registered will be part of serialization and clone (and equals)

Comment (MikeLing) [inline src/shogun/lib/DynamicArray.h]: mmm, `m_parameters->add_vector(&head, &(m_array.size()), ....` doesn't work, and it broken the clone progress of DynamicArray. Here is how I implement it.

```
		virtual CSGObject* clone()
		{
			CDynamicArray * cloned = (CDynamicArray*) CSGObject::clone();
			// Since the array vector is registered with
			// current_num_elements as size (see parameter
			// registration) the cloned version has less memory
			// allocated than known to dynarray. We fix this here.
			// cloned->num_elements = cloned->m_array.size();
			return cloned;
		}

	private:

		/** register parameters */
		virtual void init()
		{
			set_generic<T>();
			T* head = m_array.data();
                        // & doesn't accept a rvalue and it seems like auto will make 'size' unrecognisable for add_vector()
			int32_t size = m_array.size();
			m_parameters->add_vector(
			    &head, &size, "array", "Memory for dynamic array.");

			SG_ADD(
			    &free_array, "free_array", "whether array must be freed",
			    MS_NOT_AVAILABLE);
			SG_ADD(&dim1_size, "dim1_size", "Dimension 1", MS_NOT_AVAILABLE);
			SG_ADD(&dim2_size, "dim2_size", "Dimension 2", MS_NOT_AVAILABLE);
			SG_ADD(&dim3_size, "dim3_size", "Dimension 3", MS_NOT_AVAILABLE);
		}
```

Progress stop in here

> 
> [ RUN      ] DynamicObjectArray.array_SpectrumMismatchRBFKernel_clone_equals
> shogun-unit-test(91199,0x7fffe91903c0) malloc: *** error for object 0x10637f125: pointer being freed was not allocated
> *** set a breakpoint in malloc_error_break to debug
> Abort trap: 6

And here is some lldb debug message https://pastebin.mozilla.org/9024822

File: src/shogun/lib/DynamicArray.h [hunk 20]
```diff
@@ -654,6 +802,876 @@ template <class T> class CDynamicArray :public CSGObject
...
+		int32_t dim1_size;
+
+		/** dimension 2 */
+		int32_t dim2_size;
+
+		/** dimension 3 */
+		int32_t dim3_size;
+
+		/** the number of currently used elements */
+		int32_t array_size;
```
MikeLing: The array_size in here will indicate the size of this array like how many memory been allocated to this array(like `m_array.size() == array_size`). And `num_elements` means how many elements be used in this array. So they are not overlap actually.

Sorry for the wrong conclusion I made last time.

src/shogun/lib/DynamicArray.h
```diff
@@ -654,6 +802,876 @@ template <class T> class CDynamicArray :public CSGObject
...
+		/** dimension 1 */
+		int32_t dim1_size;
+
+		/** dimension 2 */
+		int32_t dim2_size;
+
+		/** dimension 3 */
+		int32_t dim3_size;
+
+		/** the number of currently used elements */
```
MikeLing: this should been changed to the size of array

vigsterkr:
done

PR #3833, opened 2017-06-05
unit test for DynamicArray

MikeLing:


MikeLing:
@vigsterkr , please take a look at this. Thank you  :)

vigsterkr:
@MikeLing just for the record make sure that you create for each set_element interface a separate unit test (as with the deque refactor that's where things went bad)

vigsterkr:
@MikeLing interestingly unit test on travis fails ;)

tests/unit/lib/DynamicArray_unittest.cc
```diff
@@ -0,0 +1,91 @@
+#include <gtest/gtest.h>
+#include <shogun/lib/DynamicArray.h>
+
+using namespace shogun;
+
+TEST(CDynamicArray, set_array)
+{
+	int32_t* array = (int32_t*)malloc(5 * sizeof(5));
```
vigsterkr: use SG_MALLOC

tests/unit/lib/DynamicArray_unittest.cc
```diff
@@ -0,0 +1,91 @@
...
+	for (index_t i = 0; i < 5; i++)
+	{
+		EXPECT_EQ(data.get_element(i), i);
+		EXPECT_EQ(data[i], i);
+	}
+
+	// test resize
+	data.resize_array(10);
+	// why?
+	EXPECT_EQ(data.get_array_size(), 15);
```
MikeLing: @vigsterkr why?

tests/unit/lib/DynamicArray_unittest.cc
```diff
@@ -0,0 +1,91 @@
...
+	data1.set_array(array, 5, false, false);
+	EXPECT_EQ(data.get_num_elements(), 5);
+	for (index_t i = 0; i < 5; i++)
+	{
+		EXPECT_EQ(data[i], i);
+		EXPECT_EQ(data1[i], i);
+	}
+
+	// inited by const array
+	const int32_t* c_array = array;
```
vigsterkr: i would definitely create a different test each of the different `set_array` methods in `CDynamicArray` not having it in one test

tests/unit/lib/DynamicArray_unittest.cc
vigsterkr: because `(10/5 +1)*5 = 15` :)

vigsterkr: see
```
			if (!exact_resize)
			{
				new_num_elements=((n/resize_granularity)+1)*resize_granularity;
			}
```

karlnapf:
Question:
Is serialization of this covered somewhere? I think we should have an automagic test, but that only checks very basic cases, I think there should be at least one non-trivial test case. Not sure whether in here or somewhere else though....thought?

MikeLing:
Hi @karlnapf, sorry for the late reply! Actually I do have serialization unit test for it in https://gist.github.com/MikeLing/665b961fae759a58535ac07b1b93e39a. But the problem is this serialise test **passed** but unit-SerializationAscii and other serialise test failed. So I'm not sure if this  unit test can really works, and that's why it haven't been included in this pr.

PR #3840, opened 2017-06-08
[SmartPointers] Port SGObject and its unit tests to Some<>.

geektoni:
`CSGObject::clone()` is still left and it will be fixed when the method signatures will be changed.


lisitsyn:
Looks nice! No leaks, right?

tests/unit/base/SGObject_unittest.cc
```diff
@@ -42,103 +41,89 @@ TEST(SGObject,equals_NULL_parameter)
...
-	SG_UNREF(array2);
+	EXPECT_TRUE(
+	    TParameter::compare_ptype(PT_SGOBJECT, &array1_tmp, &array2_tmp));
 }
 
 TEST(SGObject,equals_DynamicObjectArray_equal_after_resize)
 {
-	CDynamicObjectArray* array1=new CDynamicObjectArray();
-	CDynamicObjectArray* array2=new CDynamicObjectArray();
+	Some<CDynamicObjectArray> array1 = some<CDynamicObjectArray>();
```
lisitsyn: Can we go `auto`?

src/interfaces/modular/some.i
```diff
@@ -0,0 +1,15 @@
...
+    #include <shogun/base/Parallel.h>
+    #include <shogun/io/SGIO.h>
+%}
+
+%rename (ptr) operator CSGObject*;
+%rename (ptrSGIO) operator SGIO*;
+%rename (ptrParallel) operator Parallel*;
+%include <shogun/base/some.h>
+
+%template (SomeSGObject) shogun::Some<CSGObject>;
```
vigsterkr: this is a good test, but i'm thinking that we dont actually wanna do this (see my email to you @geektoni) but want to actually for example instead of 
```
%rename(GaussianShiftKernel) CGaussianShiftKernel;
```
in `Kernel.i` we would only expose the `Some<>` version:
```
%template (GaussianShiftKernel) shogun::Some<CGaussianShiftKernel>;
```

lisitsyn:
@geektoni want to revive this thing?

lisitsyn:
Discussed in person with @geektoni and @karlnapf. So, we stopped working at it because we had no option of using the ctors with arguments.

We can get back to this PR and hopefully merge it once we get these two things done:
1) Add create-by-name things like in #4022 
2) Make *all* examples use that create-by-name thing. This makes a great gsoc-entrance task

karlnapf:
I will create an entrance task thingi.

PR #3842, opened 2017-06-13
 Port KernelPCA to use linalg and refactor unit test

micmn:
Unit test back on, results are consistent with scikit's kpca but it may need some more testing.

### DIFF
--- a/src/shogun/mathematics/linalg/LinalgNamespace.h
+++ b/src/shogun/mathematics/linalg/LinalgNamespace.h
@@ -444,6 +444,36 @@ namespace shogun
 		}
 
 		/**
+		 * Performs the operation result = alpha * A.col(i) + beta * b,
+		 * for each column of A.
+		 * User should pass an appropriately pre-allocated memory matrix
+		 * or pass the operand argument A as a result.
+		 *
+		 * @param A The matrix
+		 * @param b The vector
+		 * @param result The matrix that saves the result
+		 * @param alpha Constant to be multiplied by the matrix
+		 * @param beta Constant to be multiplied by the vector
+		 */
+		template <typename T>
+		void add_vector(
+		    const SGMatrix<T>& A, const SGVector<T>& b, SGMatrix<T>& result,
+		    T alpha = 1, T beta = 1)
+		{
+			REQUIRE(
+			    A.num_rows == b.vlen, "Number of rows of matrix A (%d) doesn't "
+			                          "match length of vector b (%d).\n",
+			    A.num_rows, b.vlen);
+			REQUIRE(
+			    result.num_rows == A.num_rows && result.num_cols == A.num_cols,
+			    "Dimension mismatch! A (%d x %d) vs result (%d x %d).\n",
+			    A.num_rows, A.num_cols, result.num_rows, result.num_cols);
+
+			infer_backend(A, SGMatrix<T>(b))
+			    ->add_vector(A, b, result, alpha, beta);
+		}
+
+		/**
 		 * Adds a scalar to each element of a vector or a matrix in-place.
 		 *
 		 * @param a Vector or matrix
--- a/src/shogun/preprocessor/KernelPCA.cpp
+++ b/src/shogun/preprocessor/KernelPCA.cpp
@@ -9,19 +9,18 @@
  */
 
 #include <shogun/preprocessor/KernelPCA.h>
-#ifdef HAVE_LAPACK
 #include <shogun/lib/config.h>
 #include <shogun/mathematics/Math.h>
 
 #include <string.h>
 #include <stdlib.h>
 
-#include <shogun/mathematics/lapack.h>
-#include <shogun/lib/common.h>
-#include <shogun/kernel/Kernel.h>
-#include <shogun/preprocessor/DimensionReductionPreprocessor.h>
 #include <shogun/features/Features.h>
 #include <shogun/io/SGIO.h>
+#include <shogun/kernel/Kernel.h>
+#include <shogun/lib/common.h>
+#include <shogun/mathematics/linalg/LinalgNamespace.h>
+#include <shogun/preprocessor/DimensionReductionPreprocessor.h>
 
 using namespace shogun;
 
@@ -80,43 +79,38 @@ bool CKernelPCA::init(CFeatures* features)
 		int32_t m = kernel_matrix.num_rows;
 		ASSERT(n==m)
 
-		float64_t* bias_tmp = SGMatrix<float64_t>::get_column_sum(kernel_matrix.matrix, n,n);
-		SGVector<float64_t>::scale_vector(-1.0/n, bias_tmp, n);
-		float64_t s = SGVector<float64_t>::sum(bias_tmp, n)/n;
-		SGVector<float64_t>::add_scalar(-s, bias_tmp, n);
+		SGVector<float64_t> bias_tmp = linalg::rowwise_sum(kernel_matrix);
+		linalg::scale(bias_tmp, bias_tmp, -1.0 / n);
+		float64_t s = linalg::sum(bias_tmp) / n;
+		linalg::add_scalar(bias_tmp, -s);
 
-		SGMatrix<float64_t>::center_matrix(kernel_matrix.matrix, n, m);
+		linalg::center_matrix(kernel_matrix);
 
-		float64_t* eigenvalues=SGMatrix<float64_t>::compute_eigenvectors(kernel_matrix.matrix, n, n);
+		SGVector<float64_t> eigenvalues(kernel_matrix.num_cols);
+		SGMatrix<float64_t> eigenvectors(
+		    kernel_matrix.num_rows, kernel_matrix.num_cols);
+		linalg::eigen_solver_symmetric(
+		    kernel_matrix, eigenvalues, eigenvectors);
 
+		m_transformation_matrix =
+		    SGMatrix<float64_t>(kernel_matrix.num_rows, kernel_matrix.num_cols);
+		linalg::zero(m_transformation_matrix);
+		auto args = CMath::argsort(eigenvalues);
 		for (int32_t i=0; i<n; i++)
 		{
 			//normalize and trap divide by zero and negative eigenvalues
-			for (int32_t j=0; j<n; j++)
-				kernel_matrix.matrix[i*n+j]/=CMath::sqrt(CMath::max(1e-16,eigenvalues[i]));
+			auto idx = n - args[i] - 1;
+			auto vec = eigenvectors.get_column(idx);
+			auto col = m_transformation_matrix.get_column(i);
+			linalg::add(
+			    col, vec, col, 1.0,
+			    1.0 / CMath::sqrt(CMath::max(1e-16, eigenvalues[idx])));
 		}
 
-		SG_FREE(eigenvalues);
-
-		m_transformation_matrix = SGMatrix<float64_t>(kernel_matrix.matrix,n,n);
-		kernel_matrix.matrix = NULL;
 		m_bias_vector = SGVector<float64_t>(n);
-		SGVector<float64_t>::fill_vector(m_bias_vector.vector, m_bias_vector.vlen, 0.0);
-
-		cblas_dgemv(CblasColMajor, CblasTrans,
-				n, n, 1.0, m_transformation_matrix.matrix, n,
-				bias_tmp, 1, 0.0, m_bias_vector.vector, 1);
-
-		float64_t* rowsum = SGMatrix<float64_t>::get_row_sum(m_transformation_matrix.matrix, n, n);
-		SGVector<float64_t>::scale_vector(1.0/n, rowsum, n);
-
-		for (int32_t i=0; i<n; i++)
-		{
-			for (int32_t j=0; j<n; j++)
-				m_transformation_matrix.matrix[j+n*i] -= rowsum[i];
-		}
-		SG_FREE(rowsum);
-		SG_FREE(bias_tmp);
+		linalg::zero(m_bias_vector);
+		linalg::matrix_prod(
+		    m_transformation_matrix, bias_tmp, m_bias_vector, true);
 
 		m_initialized=true;
 		SG_INFO("Done\n")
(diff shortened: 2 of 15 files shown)


karlnapf:
Quite excited about this. What does it need? Can I help?

tests/unit/preprocessor/KernelPCA_unittest.cc
```diff
@@ -7,37 +7,63 @@
...
+	    test_data, num_features, num_test_vectors, false);
+	CDenseFeatures<float64_t>* feats_test =
+	    new CDenseFeatures<float64_t>(test_matrix);
+	SGMatrix<float64_t> embedding = kpca.apply_to_feature_matrix(feats_test);
+
+	// allow embedding with opposite sign
+	for (index_t i = 0; i < target_dim * num_test_vectors; ++i)
+		EXPECT_LE(
+		    CMath::abs(embedding.matrix[i]) - CMath::abs(resdata[i]), 1E-6);
+}
```
micmn: Self-review: this is wrong.

src/shogun/preprocessor/KernelPCA.cpp
```diff
@@ -80,43 +79,38 @@ bool CKernelPCA::init(CFeatures* features)
...
-		float64_t* rowsum = SGMatrix<float64_t>::get_row_sum(m_transformation_matrix.matrix, n, n);
-		SGVector<float64_t>::scale_vector(1.0/n, rowsum, n);
-
-		for (int32_t i=0; i<n; i++)
-		{
-			for (int32_t j=0; j<n; j++)
-				m_transformation_matrix.matrix[j+n*i] -= rowsum[i];
-		}
-		SG_FREE(rowsum);
-		SG_FREE(bias_tmp);
```
micmn: @karlnapf I can't figure out why he does this (L110-117)

src/shogun/preprocessor/KernelPCA.cpp
```diff
@@ -125,62 +119,49 @@ bool CKernelPCA::init(CFeatures* features)
...
 
-	for (i=0; i<num_vectors; i++)
+	auto rows_sum = linalg::rowwise_sum(kernel_matrix);
+	for (index_t i = 0; i < n; ++i)
 	{
-		for (j=0; j<m_target_dim; j++)
-			new_feature_matrix[i*m_target_dim+j] = m_bias_vector.vector[j];
+		linalg::add_col_vec(
+		    kernel_matrix, i, rows_sum, kernel_matrix, 1.0, -1.0 / n);
+	}
```
micmn: @karlnapf This mean removal step from the kernel is not carried out, is this related with L110-117? Can't see how...

tests/unit/preprocessor/KernelPCA_unittest.cc
karlnapf: LOL :)

src/shogun/preprocessor/KernelPCA.cpp
karlnapf: This is all horrible code
There should be no loops in the kernelPCA algorithm

src/shogun/preprocessor/KernelPCA.cpp
karlnapf: not sure.
Just realise how bad this kernel PCA thing is though.
I mean it should just be a couple of linear in high level land, no?

src/shogun/mathematics/linalg/LinalgNamespace.h
```diff
@@ -1057,6 +1194,34 @@ T trace(const SGMatrix<T>& A)
 }
 
 /**
+ * Method that computes the transpose of a matrix.
+ *
+ * @param A The matrix
+ * @return The transposed matrix
+ */
+template <typename T>
+SGMatrix<T> transpose_matrix(const SGMatrix<T>& A)
```
karlnapf: Rather than transposing matrices, I think it is a nicer idea to set a flag.
Either this can be added to the data structure (to the new linalg::Matrix, not the SGMatrix), or the linalg methods can have flags that take into account the transpose.

Check out eigen3 `adjoint` vs `transpose`

src/shogun/mathematics/linalg/LinalgNamespace.h
```diff
@@ -468,6 +495,63 @@ T dot(const SGVector<T>& a, const SGVector<T>& b)
...
+ * matrix to store the eigenvectors.
+ *
+ * @param A The matrix whose eigenvalues and eigenvectors are to be computed
+ * @param eigenvalues Eigenvalues result vector
+ * @param eigenvectors Eigenvectors result matrix
+ */
+template <typename T>
+void eigen_solver_symmetric(const SGMatrix<T>& A, SGVector<T>& eigenvalues, SGMatrix<T>& eigenvectors)
+{
+	REQUIRE(A.num_rows == A.num_cols, "Matrix A is not square!\n");
```
karlnapf: Always print the information that is there.
`Matrix A (%dx%d) is not square`

src/shogun/mathematics/linalg/LinalgNamespace.h
```diff
@@ -468,6 +495,63 @@ T dot(const SGVector<T>& a, const SGVector<T>& b)
...
+	REQUIRE(A.num_rows == eigenvectors.num_rows,
+			"Number of rows of A (%d) doesn't match eigenvectors' matrix (%d).\n",
+			A.num_rows, eigenvectors.num_rows);
+	REQUIRE(A.num_cols == eigenvectors.num_cols,
```
karlnapf: @vigsterkr (and everyone) I feel like we should put all those checks into functions so that we can unify their formatting.
Any thoughts on that @micmn ?

src/shogun/mathematics/linalg/LinalgNamespace.h
```diff
@@ -1044,6 +1144,43 @@ SGVector<T> rowwise_sum(const Block<SGMatrix<T>>& a, bool no_diag=false)
...
+	{
+		REQUIRE((A.num_rows == U.num_cols),
+				"Number of rows of matrix A (%d) must match number of columns of U (%d).\n",
+				A.num_cols, r);
+	}
+	REQUIRE((s.vlen == r),
+			"Length of vector s (%d) doesn't match A's smaller dimension (%d).\n",
+			s.vlen, r);
+
+	return infer_backend(A)->svd(A, s, U, thin_U);
```
karlnapf: return type is void?

src/shogun/preprocessor/KernelPCA.cpp
```diff
@@ -80,43 +79,38 @@ bool CKernelPCA::init(CFeatures* features)
...
 		}
 
-		SG_FREE(eigenvalues);
-
-		m_transformation_matrix = SGMatrix<float64_t>(kernel_matrix.matrix,n,n);
-		kernel_matrix.matrix = NULL;
 		m_bias_vector = SGVector<float64_t>(n);
-		SGVector<float64_t>::fill_vector(m_bias_vector.vector, m_bias_vector.vlen, 0.0);
-
-		cblas_dgemv(CblasColMajor, CblasTrans,
```
karlnapf: man this code ... :D

tests/unit/preprocessor/KernelPCA_unittest.cc
```diff
@@ -7,37 +7,63 @@
...
+
+TEST(KernelPCA, apply_to_feature_vector_input)
+{
+	float64_t data[] = {1, 1, 1, 1, 2, 3, 5, 6, 1, 2, 2, 2, 1, 1, 1};
```
karlnapf: and all those tests match sklearn you say?

src/shogun/mathematics/linalg/LinalgNamespace.h
micmn: Will address these linalg issues in #3843

tests/unit/preprocessor/KernelPCA_unittest.cc
micmn: I have a gist somewhere...here we go https://gist.github.com/micmn/f93f723b74db2a1eb5875f63d841bdc1

src/shogun/preprocessor/KernelPCA.cpp
```diff
@@ -125,62 +119,49 @@ bool CKernelPCA::init(CFeatures* features)
...
-		float64_t kj = m_kernel->kernel(0,j);
-
-		for (k=0; k<m_target_dim; k++)
-			result.vector[k] += kj*m_transformation_matrix.matrix[(n-k-1)*n+j];
-	}
+	SGMatrix<float64_t> result = apply_to_feature_matrix(features.get());
 
-	m_kernel->cleanup();
-	return result;
+	return SGVector<float64_t>(result.matrix, m_target_dim);
```
micmn: I'm having some trouble with this, in this way data gets deallocated, how do I pass to SGVector without copying?

tests/unit/preprocessor/KernelPCA_unittest.cc
karlnapf: You can actually put a link to this gist as a comment in the c++ unit test

src/shogun/preprocessor/KernelPCA.cpp
karlnapf: You can do `SGVector<float64_t>(result.matrix, m_target_dim, false);`

BUT I am not sure you want to do this. Who is responsible for this memory?

micmn:
I think it's the same, sgmatrix result is responsible so it gets
deallocated at the end of the method, I was wondering if I could transfer
the ownership to sgvector.

Il 16/giu/2017 21:39, "Heiko Strathmann" <notifications@github.com> ha
scritto:

src/shogun/preprocessor/KernelPCA.cpp
karlnapf: Actually, I don't understand what you are after here? Make a SGVector from SGMatrix? I think there is a constructor for this....
What do you mean by "data gets allocated"?

karlnapf:
I see. I think you will need to write a constructor in SGVector that will share the underlying `SGReferencedData` object, so that destroying the SGMatrix leaves the reference counter still at 1 (in the returned SGVector

src/shogun/preprocessor/KernelPCA.cpp
```diff
@@ -213,5 +192,3 @@ CDenseFeatures<float64_t>* CKernelPCA::apply_to_string_features(CFeatures* featu
```
micmn: Is there any reason to handle this differently? It's never used anywhere...

src/shogun/mathematics/linalg/LinalgNamespace.h
micmn: ok (should be done in the already implemented methods too).

src/shogun/mathematics/linalg/LinalgNamespace.h
micmn: Do you mean that we should take the conj transpose when T is complex?

karlnapf: no I mean that transposing a matrix (i.e. changing the order in memory) is often not necessary, but one can just mark them as transposed and the subsequent operation realises that
Nothing to do with complex adjoints

src/shogun/preprocessor/KernelPCA.cpp
karlnapf: Is this an overloaded method? It doesnt make sense indeed....in particular since the code can be shared no? (The kernel matrix is all we need). Maybe this is actually a copy-paste error? :D

vigsterkr:
@micmn let's rebase :)

src/shogun/preprocessor/KernelPCA.cpp
```diff
@@ -80,43 +79,38 @@ bool CKernelPCA::init(CFeatures* features)
...
+		linalg::add_scalar(bias_tmp, -s);
 
-		SGMatrix<float64_t>::center_matrix(kernel_matrix.matrix, n, m);
+		linalg::center_matrix(kernel_matrix);
 
-		float64_t* eigenvalues=SGMatrix<float64_t>::compute_eigenvectors(kernel_matrix.matrix, n, n);
+		SGVector<float64_t> eigenvalues(kernel_matrix.num_cols);
+		SGMatrix<float64_t> eigenvectors(
+		    kernel_matrix.num_rows, kernel_matrix.num_cols);
+		linalg::eigen_solver_symmetric(
```
karlnapf: is there anything that catches numerical errors in here?

src/shogun/preprocessor/KernelPCA.cpp
```diff
@@ -80,43 +79,38 @@ bool CKernelPCA::init(CFeatures* features)
...
+		    kernel_matrix, eigenvalues, eigenvectors);
 
+		m_transformation_matrix =
+		    SGMatrix<float64_t>(kernel_matrix.num_rows, kernel_matrix.num_cols);
+		linalg::zero(m_transformation_matrix);
+		auto args = CMath::argsort(eigenvalues);
```
karlnapf: Shall we move CMath to linalg as well?

src/shogun/preprocessor/KernelPCA.cpp
```diff
@@ -80,43 +79,38 @@ bool CKernelPCA::init(CFeatures* features)
...
 		for (int32_t i=0; i<n; i++)
 		{
 			//normalize and trap divide by zero and negative eigenvalues
-			for (int32_t j=0; j<n; j++)
-				kernel_matrix.matrix[i*n+j]/=CMath::sqrt(CMath::max(1e-16,eigenvalues[i]));
+			auto idx = n - args[i] - 1;
+			SGVector<float64_t> vec(
+			    eigenvectors.get_column_vector(idx), n, false);
+			linalg::add_col_vec(
```
karlnapf: this API is a bit messy maybe? Reminds me of cblas ;)

src/shogun/preprocessor/KernelPCA.cpp
```diff
@@ -125,61 +119,46 @@ bool CKernelPCA::init(CFeatures* features)
...
+	int32_t num_vectors = features->get_num_vectors();
+	int32_t n = m_init_features->get_num_vectors();
 
-	m_kernel->init(features,m_init_features);
+	m_kernel->init(features, m_init_features);
+	auto kernel_matrix = m_kernel->get_kernel_matrix();
 
-	float64_t* new_feature_matrix = SG_MALLOC(float64_t, m_target_dim*num_vectors);
+	auto rows_sum = linalg::rowwise_sum(kernel_matrix);
+	for (index_t i = 0; i < n; ++i)
```
karlnapf: this loop could be in linalg I guess?

src/shogun/preprocessor/KernelPCA.cpp
```diff
@@ -125,61 +119,46 @@ bool CKernelPCA::init(CFeatures* features)
...
-			float64_t kij = m_kernel->kernel(i,j);
+	SGMatrix<float64_t> new_feature_matrix =
+	    linalg::matrix_prod(W, kernel_matrix, true, true);
 
-			for (k=0; k<m_target_dim; k++)
-				new_feature_matrix[k+i*m_target_dim] += kij*m_transformation_matrix.matrix[(n-k-1)*n+j];
-		}
-	}
+	SGVector<float64_t> bias(m_bias_vector, m_target_dim, false);
+	for (index_t i = 0; i < num_vectors; ++i)
```
karlnapf: put into linalg?

tests/unit/preprocessor/KernelPCA_unittest.cc
```diff
@@ -7,37 +7,66 @@
...
-                         -4.032822763552926e-01, -5.151523890814317e-01,
-                         8.444041004961732e-01, -4.318711485273607e-01,
-                         -4.105842439768400e-01, -4.335318603758601e-01,
-                         -1.526879008202015e-02, 6.902776989923268e-01
-	                        };// column-wise
+	float64_t data[] = {1, 1, 1, 1, 2, 3, 5, 6, 1, 2, 2, 2, 1, 1, 1};
+	float64_t test_data[] = {3, 3, 3, 7, 4, 1};
+	// Results compared with sklearn
+	// https://gist.github.com/micmn/f93f723b74db2a1eb5875f63d841bdc1
+	float64_t resdata[] = {-0.17645841, 0.013962, -0.16082441, 0.03640145};
```
karlnapf: I think we should from now on avoid any fix data creation code in unit tests, and always move it to function, fixture, or whatever. What do you think?

tests/unit/preprocessor/KernelPCA_unittest.cc
```diff
@@ -7,37 +7,66 @@
...
+	int32_t num_test_vectors = 2;
+	int32_t target_dim = 2;
+
+	SGMatrix<float64_t> m(data, num_features, num_vectors, false);
+	CDenseFeatures<float64_t>* feats = new CDenseFeatures<float64_t>(m);
+
+	CGaussianKernel* kernel = new CGaussianKernel();
+	kernel->set_width(1);
+
+	CKernelPCA kpca(kernel);
```
karlnapf: on stack?

tests/unit/preprocessor/KernelPCA_unittest.cc
```diff
@@ -7,37 +7,66 @@
...
+	SGMatrix<float64_t> test_matrix(
+	    test_data, num_features, num_test_vectors, false);
+	CDenseFeatures<float64_t>* feats_test =
+	    new CDenseFeatures<float64_t>(test_matrix);
+	SGMatrix<float64_t> embedding = kpca.apply_to_feature_matrix(feats_test);
+
+	// allow embedding with opposite sign
+	auto s = CMath::sign(embedding[0] * resdata[0]);
+	for (index_t i = 0; i < target_dim * num_test_vectors; ++i)
+		EXPECT_NEAR(embedding[i], s * resdata[i], 1E-6);
```
karlnapf: this doesnt leak?

src/shogun/preprocessor/KernelPCA.cpp
micmn: yeah, could we add get_column(i) -> SGVector in SGMatrix that does this?

karlnapf: Yes, add it. Need to tell the SGVector to not free memory at the end though, and add a clear warning in the API doc that it is non-owning

src/shogun/preprocessor/KernelPCA.cpp
micmn: add(matrix, vector) that adds the vector to each column of the matrix?

karlnapf: yep, a broadcasting like add

PR #3846, opened 2017-06-15
Split Eigen3's linalg backend into header and implementation.

micmn:
Waiting for the big linalg refactoring, in this way the backend gets compiled only once and compilation time returns to be acceptable , although travis may still fail without splitting the implementation into more files.

### DIFF
--- a/src/shogun/mathematics/linalg/LinalgBackendEigen.h
+++ b/src/shogun/mathematics/linalg/LinalgBackendEigen.h
@@ -33,10 +33,10 @@
 #ifndef LINALG_BACKEND_EIGEN_H__
 #define LINALG_BACKEND_EIGEN_H__
 
-#include <shogun/lib/SGVector.h>
 #include <shogun/mathematics/eigen3.h>
 #include <shogun/mathematics/linalg/LinalgBackendBase.h>
-#include <numeric>
+#include <shogun/mathematics/linalg/LinalgMacros.h>
+#include <shogun/mathematics/Math.h>
 
 namespace shogun
 {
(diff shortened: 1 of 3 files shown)


src/shogun/mathematics/linalg/LinalgBackendEigen.h
```diff
@@ -99,711 +98,321 @@ class LinalgBackendEigen : public LinalgBackendBase
...
-	}
+	virtual void zero(Container<Type>& a) const;
 	DEFINE_FOR_ALL_PTYPE(BACKEND_GENERIC_ZERO, SGVector)
 	DEFINE_FOR_ALL_PTYPE(BACKEND_GENERIC_ZERO, SGMatrix)
 	#undef BACKEND_GENERIC_ZERO
 
 	#undef DEFINE_FOR_ALL_PTYPE
+	#undef DEFINE_FOR_REAL_PTYPE
+	#undef DEFINE_FOR_NON_INTEGER_PTYPE
+	#undef DEFINE_FOR_NUMERIC_PTYPE
```
micmn: @vigsterkr are you suggesting to remove these undefs?

vigsterkr: argh...... have not seen that one... sooooooooo ugly, soooo redundant :(
we could put these macros into a separate header? and include it

micmn: ok but in this case the header that contains the macros can't be guarded otherwise a second inclusion like in the cpp will not work.

vigsterkr: yep.... indeed. i dont think it's a problem... i mean that the header would only be actually included consciously one would presume :) i.e. this is just to avoid copy-paste errors and actual copy-pasting things around.

vigsterkr:
seems gcc fails due to clang-format errors, and clang job failed because there was a hiccup of github:

```CMake Error at GoogleMock-stamp/download-GoogleMock.cmake:27 (message):

  error: downloading

  'https://github.com/google/googletest/archive/release-1.8.0.tar.gz' failed
```

vigsterkr:
@micmn once you have the factored macro header just push it so then the gcc/clang gets retested. do we wanna fix now the clang-format? or that should be done later as part of another refactor? :)

PR #3852, opened 2017-06-19
add serializable test for DynamicArray

MikeLing:


vigsterkr:
ok this should pass on CIs.. but let's be 100% sure about it and wait for them once it's green then we merge

tests/unit/lib/DynamicArray_unittest.cc
```diff
@@ -251,10 +252,43 @@ TEST_F(CDynamicArrayFixture, append_array_bool)
...
+	file->close();
+	SG_UNREF(file);
+
+	for (int32_t i = 0; i < this->custom_array->get_num_elements(); i++)
+	{
+		EXPECT_EQ(
+		    this->custom_array->get_element(i), new_array->get_element(i));
+	}
+
+	SG_UNREF(new_array);
```
vigsterkr: @MikeLing could you add this on the very end:
```unlink(filename);```

Comment (MikeLing) [inline tests/unit/lib/DynamicArray_unittest.cc]: got it

File: tests/unit/lib/DynamicArray_unittest.cc [hunk 2]
```diff
@@ -251,10 +252,44 @@ TEST_F(CDynamicArrayFixture, append_array_bool)
...
-	wrapper_array_b->save_serializable_pre();
-	wrapper_array_b->load_serializable_pre();
+	for (int32_t i = 0; i < 5; i++)
+	{
+		this->custom_array->push_back((TypeParam)CMath::random(0, 1));
+	}
+
+	/* generate file name */
+	char filename_tmp[] = "serialization-asciiCDynamicArray.XXXXXX";
+#ifdef _WIN32
```
karlnapf: can we put this "generate filename" into a helper function please?

tests/unit/lib/DynamicArray_unittest.cc
```diff
@@ -251,10 +252,44 @@ TEST_F(CDynamicArrayFixture, append_array_bool)
...
+	ASSERT(err == 0);
+#else
+	int fd = mkstemp(filename_tmp);
+	ASSERT(fd != -1);
+	int retval = close(fd);
+	ASSERT(retval != -1);
+#endif
+	char* filename = filename_tmp;
+
+	/* create serialized copy */
```
karlnapf: remove such comments, they are redundant (code speaks for itself)

tests/unit/lib/DynamicArray_unittest.cc
```diff
@@ -251,10 +252,44 @@ TEST_F(CDynamicArrayFixture, append_array_bool)
 	EXPECT_EQ(wrapper_array_b->get_element(2), true);
 }
 
-TEST_F(CDynamicArrayFixture, serializable_array)
+TYPED_TEST(CDynamicArrayTest, serializable_array)
```
karlnapf: please name this test better "serializable_array" is not a good description

tests/unit/lib/DynamicArray_unittest.cc
```diff
@@ -251,10 +252,44 @@ TEST_F(CDynamicArrayFixture, append_array_bool)
...
+	file->close();
+	SG_UNREF(file);
+
+	file = new CSerializableAsciiFile(filename, 'r');
+	CDynamicArray<TypeParam>* new_array = new CDynamicArray<TypeParam>();
+	new_array->load_serializable(file);
+	file->close();
+	SG_UNREF(file);
+
+	for (int32_t i = 0; i < this->custom_array->get_num_elements(); i++)
```
karlnapf: first assert that the number of elements are the same, otherwise this might segfault
Use a variable for that (not just constant 5 in the code above)

karlnapf:
Needs some clean up and then good to be merged

tests/unit/lib/DynamicArray_unittest.cc
MikeLing: just `serialization` maybe better?

tests/unit/lib/DynamicArray_unittest.cc
MikeLing: mmm, these code only use once in here, do we need a helper function to do that?

karlnapf: There is a few places where we use temporary filenames. All of those should use the same helper function

tests/unit/lib/DynamicArray_unittest.cc
karlnapf: In general, I would name the test after the function I test, here `save_serializable`

tests/unit/lib/DynamicArray_unittest.cc
MikeLing: where we should put that helper function? In Utils.h maybe?

tests/unit/lib/DynamicArray_unittest.cc
```diff
@@ -251,10 +254,35 @@ TEST_F(CDynamicArrayFixture, append_array_bool)
...
-	wrapper_array_i->load_serializable_pre();
-	wrapper_array_b->save_serializable_pre();
-	wrapper_array_b->load_serializable_pre();
+	for (int32_t i = 0; i < 5; i++)
+	{
+		this->custom_array->push_back((TypeParam)CMath::random(0, 1));
+	}
+
+	/* generate file name */
+	char filename_tmp[] = "serialization-asciiCDynamicArray.XXXXXX";
```
karlnapf: minor style: why is this a variable? No need imho

tests/unit/utils/Utils.h
```diff
@@ -45,5 +45,6 @@
 * https://www.gnu.org/software/libc/manual/html_node/Temporary-Files.html
 */
 char * mktemp_cst(char * __template);
+char* generate_file_name(char* file_name);
```
karlnapf: API docs please

tests/unit/lib/DynamicArray_unittest.cc
```diff
@@ -1,7 +1,10 @@
 #include <gtest/gtest.h>
+#include <shogun/io/SerializableAsciiFile.h>
 #include <shogun/lib/DynamicArray.h>
 #include <shogun/mathematics/Math.h>
 
+#include "../utils/Utils.h"
```
vigsterkr: i believe that it should be enough to have:
```
#include "utils/Utils.h"
```

or?

tests/unit/utils/Utils.h
MikeLing: oh, I somehow miss it. I will do it right away.

tests/unit/lib/DynamicArray_unittest.cc
vigsterkr: yeah this you can just pass directly to the `generate_file_name`

tests/unit/utils/Utils.cpp
```diff
@@ -68,3 +70,18 @@ char * mktemp_cst(char * __template)
 
     return __template;
 }
+
+char* generate_file_name(char* file_name)
```
vigsterkr: i would call this function rather
```
generate_temp_filename
```

MikeLing: mmm, maybe I misunderstand something in here. If I pass "serialization-asciiCDynamicArray.XXXXXX" directly to `generate_file_name`, it will generate warning like:

>
/Users/mikeling/shogun/tests/unit/lib/DynamicArray_unittest.cc:266:42: warning: ISO C++11 does
      not allow conversion from string literal to 'char *' [-Wwritable-strings]
        char* filename = generate_temp_filename("serialization-asciiCDynamicArray.XXXXXX");
>

vigsterkr: @MikeLing deprecated :) true.... ignore these comments :) the naming of the fucntion though should reflect the temp-ness :)
btw the return type should be void since the passed `char*` will actually contain the right filename :)

vigsterkr: so basically:
```
void generate_temp_filename(char* filename)
```

after you call `generate_temp_filename` then the passed `filename` will contain the right path :)

tests/unit/lib/DynamicArray_unittest.cc
```diff
@@ -251,10 +254,35 @@ TEST_F(CDynamicArrayFixture, append_array_bool)
...
-	wrapper_array_i->load_serializable_pre();
-	wrapper_array_b->save_serializable_pre();
-	wrapper_array_b->load_serializable_pre();
+	for (int32_t i = 0; i < 5; i++)
+	{
+		this->custom_array->push_back((TypeParam)CMath::random(0, 1));
+	}
+
+	/* generate file name */
+	char filename[] = "serialization-asciiCDynamicArray.XXXXXX";
```
MikeLing: mmm, I get Bus error if I change it with  `char* filename = (char*)"serialization-asciiCDynamicArray.XXXXXX";` And it will happened in `mkstemp`. So I guess it's because I haven't assign a length of memory to that pointer. Anyway, I think we don't need to worry about it should be a pointer or not :)

PR #3848, opened 2017-06-15
[PrematureStopping] Refactor CSignal class to use RxCpp utilities.

geektoni:
This PR <del>should</del> contains the new `CSignal` implementation which will be using RxCpp. <del>However, there are still some things that needs to be addressed (`CSignal::cancel_computation`, there is still Windows code etc.).</del>

PR #3853, opened 2017-06-19
[PrematureStopping ] Add signal handler to init methods.

geektoni:
Needs rebase once #3848 gets merged. Only bf2160c is important.


geektoni:
@vigsterkr

PR #3854, opened 2017-06-20
Switch-based untemplated matrix prototype

micmn:
First prototype of untemplated linalg, barebone Matrix class.

A few points:
- do we offer operators `[i]`, `(i,j)` and similar methods? If so, how to handle the return type?
- some linalg functions take scalars of the matrix's type, i.e. `add(SGMatrix<T> a, SGMatrix<T> b, SGMatrix<T> result, T alpha, T beta)`
- Matrix's memory management, just like SGMatrix?

### DIFF
--- a/src/shogun/lib/Matrix.h
+++ b/src/shogun/lib/Matrix.h
@@ -1,0 +1,57 @@
+#ifndef MATRIX_H__
+#define MATRIX_H__
+
+#include <shogun/lib/common.h>
+#include <shogun/lib/DataType.h>
+#include <shogun/lib/SGMatrix.h>
+
+namespace shogun
+{
+
+class Matrix {
+
+	void *m_data;
+	EPrimitiveType m_ptype;
+	index_t m_rows, m_cols;
+
+	template <typename T>
+	static EPrimitiveType ptype();
+
+public:
+	template <typename T>
+	Matrix(T *data, index_t rows, index_t cols)  :
+		m_data(data), m_rows(rows), m_cols(cols)
+	{
+		m_ptype = ptype<T>();
+	}
+
+	template <typename T>
+	operator SGMatrix<T>() const
+	{
+		return SGMatrix<T>(static_cast<T*>(m_data), m_rows, m_cols, false);
+	}
+
+	EPrimitiveType ptype() const
+	{
+		return m_ptype;
+	}
+
+	template <typename T>
+	T* raw_data()
+	{
+		return static_cast<T*>(m_data);
+	}
+};
+
+template <>
+inline EPrimitiveType Matrix::ptype<float32_t>() { return PT_FLOAT32; }
+
+template <>
+inline EPrimitiveType Matrix::ptype<float64_t>() { return PT_FLOAT64; }
+
+template <>
+inline EPrimitiveType Matrix::ptype<floatmax_t>() { return PT_FLOATMAX; }
+
+}
+
+#endif // MATRIX_H_
--- a/src/shogun/mathematics/linalg/LinalgNamespaceMatrix.cpp
+++ b/src/shogun/mathematics/linalg/LinalgNamespaceMatrix.cpp
@@ -1,0 +1,43 @@
+#include <shogun/lib/DataType.h>
+#include <shogun/mathematics/linalg/LinalgNamespaceMatrix.h>
+#include <shogun/mathematics/linalg/LinalgNamespace.h>
+
+#define DEFINE_SWITCH(PTYPE, FUNC, ...) \
+switch (PTYPE) \
+{ \
+	case PT_FLOAT32: \
+		FUNC<float32_t>(__VA_ARGS__); \
+		break; \
+	case PT_FLOAT64: \
+		FUNC<float64_t>(__VA_ARGS__); \
+		break; \
+	case PT_FLOATMAX: \
+		FUNC<floatmax_t>(__VA_ARGS__); \
+		break; \
+	default: \
+		SG_SERROR("Unsupported data type\n"); \
+}
+
+namespace shogun
+{
+
+namespace linalg
+{
+
+template <typename T>
+static inline void add_templated(const Matrix& a, const Matrix& b, Matrix& result)
+{
+	SGMatrix<T> a_sg(a), b_sg(b), result_sg(result);
+	add(a_sg, b_sg, result_sg);
+}
+
+void add(const Matrix& a, const Matrix& b, Matrix& result)
+{
+	DEFINE_SWITCH(a.ptype(), add_templated, a, b, result);
+}
+
+}
+
+}
+
+#undef DEFINE_SWITCH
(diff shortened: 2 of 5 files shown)


src/shogun/mathematics/linalg/LinalgNamespaceMatrix.cpp
```diff
@@ -0,0 +1,43 @@
+#include <shogun/lib/DataType.h>
+#include <shogun/mathematics/linalg/LinalgNamespaceMatrix.h>
+#include <shogun/mathematics/linalg/LinalgNamespace.h>
+
+#define DEFINE_SWITCH(PTYPE, FUNC, ...) \
+switch (PTYPE) \
+{ \
+	case PT_FLOAT32: \
+		FUNC<float32_t>(__VA_ARGS__); \
```
karlnapf: cool! That is indeed the easiest solution for what we have in mind
There is this issue:
 * Imagine you have some rather cheap operations (small dim add)
 * imagine you call this all the time, i.e. in a loop
 * Then the switch might have some run-time drawbacks

There is two competing mindsets here:
 * We can say "we are not in rush to compute" (@lisitsyn) and argue that looped linalg calls should rather be put into a vectorized linalg method that doesnt do the switch
 * We can add a mechanism that kind of "tells" linalg that the next calls are all say `float64_t` (and the switch is ignored)
 * the above could work via registering pointers to the implementations

These have to be traded off. I rather tend to the first solution

src/shogun/mathematics/linalg/LinalgNamespaceMatrix.cpp
```diff
@@ -0,0 +1,43 @@
...
+		break; \
+	case PT_FLOAT64: \
+		FUNC<float64_t>(__VA_ARGS__); \
+		break; \
+	case PT_FLOATMAX: \
+		FUNC<floatmax_t>(__VA_ARGS__); \
+		break; \
+	default: \
+		SG_SERROR("Unsupported data type\n"); \
```
karlnapf: I guess a key headache here is mixed types (we probably wont allow them right?)

src/shogun/mathematics/linalg/LinalgNamespaceMatrix.cpp
```diff
@@ -0,0 +1,43 @@
...
+}
+
+namespace shogun
+{
+
+namespace linalg
+{
+
+template <typename T>
+static inline void add_templated(const Matrix& a, const Matrix& b, Matrix& result)
```
karlnapf: I think this would be the existing `impl` methods we have in linalg rather than new ones?

tests/unit/mathematics/linalg/operations/Eigen3_matrix_operations_unittest.cc
```diff
@@ -0,0 +1,31 @@
...
+
+	TypeParam A_data[] = {0, 1, 2, 3, 4, 5};
+	TypeParam B_data[] = {1, 2, 3, 4, 5, 6};
+	TypeParam result_data[] = {0, 0, 0, 0, 0, 0};
+
+	Matrix A(A_data, nrows, ncols);
+	Matrix B(B_data, nrows, ncols);
+	Matrix result(result_data, nrows, ncols);
+
+	add(A, B, result);
```
karlnapf: ace interface. Type free algo implementations :dancer:

tests/unit/mathematics/linalg/operations/Eigen3_matrix_operations_unittest.cc
```diff
@@ -0,0 +1,31 @@
...
+#include <shogun/mathematics/linalg/LinalgNamespaceMatrix.h>
+#include <gtest/gtest.h>
+
+using namespace shogun;
+using namespace linalg;
+
+template <typename T>
+class LinalgBackendEigenMatrix : public ::testing::Test {};
+
+typedef ::testing::Types<float32_t, float64_t, floatmax_t> PTypes;
```
karlnapf: yeah nice to do a typed test.
(offtopic: this should be done in lots of other places in shogun tests as well)

src/shogun/lib/Matrix.h
```diff
@@ -0,0 +1,57 @@
...
+#include <shogun/lib/SGMatrix.h>
+
+namespace shogun
+{
+
+class Matrix {
+
+	void *m_data;
+	EPrimitiveType m_ptype;
+	index_t m_rows, m_cols;
```
karlnapf: I like the minimal interface.
There should be in and out wrappers for SG* of course (later)

src/shogun/lib/Matrix.h
```diff
@@ -0,0 +1,57 @@
...
+#include <shogun/lib/common.h>
+#include <shogun/lib/DataType.h>
+#include <shogun/lib/SGMatrix.h>
+
+namespace shogun
+{
+
+class Matrix {
+
+	void *m_data;
```
lisitsyn: I believe `void* m_data` is a bit better but don't insist :)

src/shogun/lib/Matrix.h
lisitsyn: I'd suggest one member per line

src/shogun/lib/Matrix.h
```diff
@@ -0,0 +1,57 @@
...
+	Matrix(T *data, index_t rows, index_t cols)  :
+		m_data(data), m_rows(rows), m_cols(cols)
+	{
+		m_ptype = ptype<T>();
+	}
+
+	template <typename T>
+	operator SGMatrix<T>() const
+	{
+		return SGMatrix<T>(static_cast<T*>(m_data), m_rows, m_cols, false);
```
lisitsyn: This cast is unsafe, right? Should be documented or asserted

src/shogun/mathematics/linalg/LinalgNamespaceMatrix.cpp
vigsterkr: ok so here's my thing about the former story: lazy-evaluation (see eigen) or codegen/JIT (see XLA of TF). these are not orthogonal, meaning if we can have lazy evaluation, say:
```
Matrix x, y;
float64_t i;
x.add(y).scale(i);
```
i guess you get what i'm saying... this way you only need once to figure out the type of the matrix (and other elements' type) and then the rest is a straight line. of course this complicates things a lot, but to be honest if we look around ourselves this is the flow of every framework around (a bit of functional approach in the imperative world)

tests/unit/mathematics/linalg/operations/Eigen3_matrix_operations_unittest.cc
vigsterkr: fyi: @karlnapf already done by @MikeLing in couple of tests :P

src/shogun/mathematics/linalg/LinalgNamespaceMatrix.cpp
karlnapf: +1 for the lazy, though quite a bit of API work needed for that. Linalg currently cannot do that.
So whats the best way forward right now in here?

PR #3864, opened 2017-06-24
Create test environments in main()

micmn:
Previous version used singleton and created useless environment instances in `main()`, also I don't think that you can use that type of singleton with gtest environments since `AddGlobalEnvironment()` frees its memory.
In this patch I used global vars as suggested in gtest guide but of course it could be done differently.

Note: envs should be created after `init_shogun_with_defaults()`? But SVMOcas test fails (rand initialization?)

(+ fix typo in `MutiLabelTestEnvironment`)


MikeLing:
oh, I really learn a lot of things from this pr, thank you @micmn :)

PR #3866, opened 2017-06-25
fix the broken GMM

MikeLing:


MikeLing:
Here is the GMM notebook I got, after fix the broken GMM

MikeLing:
https://gist.github.com/MikeLing/4a3c48bfa399d48cedae2c36506611f7

src/shogun/distributions/Gaussian.cpp
```diff
@@ -196,20 +195,29 @@ float64_t CGaussian::update_params_em(float64_t* alpha_k, int32_t len)
...
 	{
-		for (int32_t j = 0; j < num_dim * num_dim; j++)
-			cov_sum[j] /= alpha_k_sum;
+		linalg::scale(cov_sum, cov_sum, 1 / alpha_k_sum);
 
-		SGVector<float64_t> d0;
-		d0 = cov_sum.get_diagonal_vector();
+		SGVector<float64_t> d0(num_dim);
+#ifdef HAVE_LAPACK
+		d0.vector = SGMatrix<float64_t>::compute_eigenvectors(
```
karlnapf: Please  use that method but linalg.
I think a self adjoint eigensilver is in there now
Michele will know or can help

src/shogun/distributions/Gaussian.cpp
```diff
@@ -351,16 +357,24 @@ void CGaussian::decompose_cov(SGMatrix<float64_t> cov)
...
-		    sizeof(float64_t) * cov.num_rows * cov.num_rows);
-
-		m_d.vector = cov.get_diagonal_vector();
-		m_d.vlen = cov.num_rows;
-		m_u.num_rows = cov.num_rows;
-		m_u.num_cols = cov.num_rows;
+		m_u = cov.clone();
+		m_d = SGVector<float64_t>(cov.num_rows);
+#ifdef HAVE_LAPACK
+		m_d.vector = SGMatrix<float64_t>::compute_eigenvectors(
```
karlnapf: Same here

PR #3868, opened 2017-06-25
Linalg Lib clang-format check

OXPHOS:
`clang-format-3.8 -i *.*` in path `src/shogun/mathematics/linalg/` with `./.clang-format` in docker.
No other changes.


OXPHOS:
`src/shogun/mathematics/linalg/.clang-format` with `clang-format-3.8` in docker

PR #3875, opened 2017-06-26
[PrematureStopping] Convert CSignal::cancel_computations() to cancel_computation().

geektoni:
Remove `CSignal::cancel_computations()` and add `TODO` when the class does not inherit directly from CMachine.


karlnapf:
Algorithms to add:
 * LeastAngleRegression
 * Kernel hypothesis testing (@lambday can guide what can be early stopped)
 * Most things in `structure`
 * Fernandos LMNN 
 * many more, make sure to deploy this rather than just replacing SG_PROGRESS, it is very cool!

(those can also get a progress bar btw)

karlnapf:
* ModelSelection
 * Cross-validation

PR #3876, opened 2017-06-26
fix the leaking DynamicArray unittest

MikeLing:


vigsterkr:
@MikeLing seems that broke msvc :)

vigsterkr:
the problem is that in `CDynamicArrayFixture` you create the `wrapper_array*` by:
```
		wrapper_array_b = new CDynamicArray<bool>(this->b_array, 5);
		wrapper_array_i = new CDynamicArray<int32_t>(this->i_array, 5);
```

if you check that'll actually set `CDynamicArray.m_array.free_array = true`, so when you call `SG_UNREF` on `wrapper_array*` it'll free the underlying array itself hence the deconstructor will try to call SG_FREE on an already deallocated memory region....

vigsterkr:
and actually could you fix the fixture to actually use meaningfully the typed tests
```
template<typename T>
class CDynamicArrayFixture : public ::testing::Test
{
protected:
	CDynamicArrayFixture()
	{
		array = SG_MALLOC(T, 5);
		for (int32_t i = 0; i < 5; i++)
		{
			array[i] = (T)i;
		}
	}
	virtual void SetUp()
	{
		wrapper_array = new CDynamicArray<T>(this->array, 5);
	}
	virtual void TearDown()
	{
		SG_UNREF(wrapper_array);
	}
	virtual ~CDynamicArrayFixture()
	{
	}
	
	CDynamicArray<T>* wrapper_array;
	T* array;
};
```

and then of course i'd do this:
```
typedef ::testing::Types<bool, char, int8_t, uint8_t, int16_t, int32_t, int64_t, float32_t, float64_t> DynamicArrayTypes;
TYPED_TEST_CASE(CDynamicArrayFixture, DynamicArrayTypes);
```

and then use `TYPED_TEST` for all the tests you've created for `CDynamicArrayFixture`

vigsterkr:
@MikeLing how is this going along? do you need a hand, the example i've pasted above is pretty easy to apply on your existing code

MikeLing:
still need to verify its memory by valgrind, please don't review it for now :) thank you

tests/unit/lib/DynamicArray_unittest.cc
```diff
@@ -8,265 +8,178 @@
...
-		wrapper_array_b = new CDynamicArray<bool>(this->b_array, 5);
-		wrapper_array_i = new CDynamicArray<int32_t>(this->i_array, 5);
+		wrapper_array = new CDynamicArray<T>(this->array, 5);
 	}
 	virtual void TearDown()
 	{
-		SG_FREE(wrapper_array_b);
-		SG_FREE(wrapper_array_i);
+		SG_UNREF(wrapper_array);
+		SG_FREE(array);
```
vigsterkr: this is double free...!

vigsterkr: it's enough to do `SG_UNREF` in this case...

MikeLing: mmm, I think we still need to free the array. Otherwise we will lost some memory https://pastebin.mozilla.org/9026080. 

And if i change it to 

	virtual void SetUp()
	{
		array = SG_MALLOC(T, 5);
		for (int32_t i = 0; i < 5; i++)
		{
			array[i] = (T)i;
		}
		wrapper_array = new CDynamicArray<T>(m_array, 5);
		SG_FREE(array);
	}

The memory check will be ok https://pastebin.mozilla.org/9026081

vigsterkr:
isn't `wrapper_array = new CDynamicArray<T>(this->array, 5);` ctor actually setting the free_array=bool? but yeah i mean if valgrind says it's ok then it's ok ;)

vigsterkr:
check the formatting error:
https://travis-ci.org/shogun-toolbox/shogun/jobs/248706801#L798

MikeLing:
ok, here is the valgrind's result  https://pastebin.mozilla.org/9026083. Hope it's good to fix the memory leaking issue :)

PR #3880, opened 2017-06-28
Make it possible to obtain type_info from Any

lisitsyn:


tests/unit/lib/Any_unittest.cc
```diff
@@ -112,3 +112,10 @@ TEST(Any, mixing_policies)
 	EXPECT_THROW(owning_any = non_owning_any, std::logic_error);
 	EXPECT_THROW(non_owning_any = owning_any, std::logic_error);
 }
+
+TEST(Any, type_info)
+{
+	int32_t integer = 10;
+	auto any = erase_type(integer);
+	EXPECT_EQ(any.type_info(), typeid(integer));
```
vigsterkr: based on standards this is not necessarily true, see from: http://en.cppreference.com/w/cpp/language/typeid

```
const std::type_info& ti1 = typeid(A);
const std::type_info& ti2 = typeid(A);
 
assert(&ti1 == &ti2); // not guaranteed
assert(ti1.hash_code() == ti2.hash_code()); // guaranteed
assert(std::type_index(ti1) == std::type_index(ti2)); // guaranteed
```

Issue #3884, opened 2017-06-29
ICA algorithms crash if features are empty


karlnapf:
If you do
```
feats = RealFeatures()

ica=FastICA()
ica.apply(feats)
```

you get an eigen3 error (cryptic) due to the NULL matrix in the features. This should rather be a nice error message (coming from a REQUIRE in the ICA base class or so)

@MikeLing this one is for you

MikeLing:
https://github.com/shogun-toolbox/shogun/pull/3887

PR #3887, opened 2017-06-29
add error message when features are empty

MikeLing:
@karlnapf maybe something like this? :)
screen shot 2017-06-30 at 1 24 50 am

### DIFF
--- a/src/shogun/converter/ica/FastICA.cpp
+++ b/src/shogun/converter/ica/FastICA.cpp
@@ -76,6 +76,7 @@ CFeatures* CFastICA::apply(CFeatures* features)
 	SG_REF(features);
 
 	SGMatrix<float64_t> X = ((CDenseFeatures<float64_t>*)features)->get_feature_matrix();
+	REQUIRE(X.data(), "Features have not been provided.\n");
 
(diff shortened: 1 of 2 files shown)


src/shogun/converter/ica/FastICA.cpp
```diff
@@ -76,6 +76,7 @@ CFeatures* CFastICA::apply(CFeatures* features)
 	SG_REF(features);
 
 	SGMatrix<float64_t> X = ((CDenseFeatures<float64_t>*)features)->get_feature_matrix();
+	REQUIRE(X.data(), "you are using an empty matrix");
```
vigsterkr: it's good but the error message should be rather: "Features have not been provided.\n"

vigsterkr:
one more thing: if you mention in your commit message the issue number on github, github will automagically link it to the PR/commit, so for example you could have a commit message:
`Fix #3884 add error message when features are empty`

MikeLing:
Sorry, I get up late today. I will update this and other pr within today, thank you @vigsterkr ,

vigsterkr:
that's ok, thnx.

PR #3888, opened 2017-06-30
get rid of sg_rand

MikeLing:


MikeLing:
The following tests FAILED:
    243 - unit-KernelSelectionMaxCrossValidation (Failed)
    249 - unit-CrossValidationMMD (Failed)
    282 - integration_meta_cpp-distance-cosine (Failed)

Will look into it asap

src/shogun/base/SGObject.h
```diff
@@ -567,6 +573,9 @@ class CSGObject
 	/** Hash of parameter values*/
 	uint32_t m_hash;
 
+	/** random generator */
+	CRandom* m_rng;
```
vigsterkr: i'd rather wrap it with unique_ptr:
```
std::unique_ptr<CRandom> m_rng;
```

src/shogun/clustering/KMeansMiniBatch.cpp
```diff
@@ -131,18 +131,20 @@ SGVector<int32_t> CKMeansMiniBatch::mbchoose_rand(int32_t b, int32_t num)
 {
 	SGVector<int32_t> chosen=SGVector<int32_t>(num);
 	SGVector<int32_t> ret=SGVector<int32_t>(b);
+	CRandom* m_rng = new CRandom();
```
vigsterkr: ```auto m_rng = std::unique_ptr<CRandom>(new CRandom());```

src/shogun/mathematics/Math.h
```diff
@@ -1110,9 +1114,11 @@ class CMath : public CSGObject
 		/// Returns a Gaussian or Normal random number.
 		/// Using the polar form of the Box-Muller transform.
 		/// http://en.wikipedia.org/wiki/Box%E2%80%93Muller_transform#Polar_form
-		static inline float64_t normal_random(float64_t mean, float64_t std_dev)
+		static inline float64_t
+		normal_random(float64_t mean, float64_t std_dev, int32_t seed = 12345)
 		{
-			return sg_rand->normal_distrib(mean, std_dev);
+			float64_t result = m_rng->normal_distrib(mean, std_dev);
```
vigsterkr: what happens with `seed`? is there a place where somebody calls `CMath::normal_random(<mean>, <dev>, <seed>)`?

src/shogun/mathematics/Math.h
```diff
@@ -1124,9 +1130,10 @@ class CMath : public CSGObject
 
 		/// Convenience method for generating Standard Normal random numbers
 		/// Double: Mean = 0 and Standard Deviation = 1
-		static inline float64_t randn_double()
+		static inline float64_t randn_double(int32_t seed = 12345)
 		{
-			return sg_rand->std_normal_distrib();
+			float64_t result = m_rng->std_normal_distrib();
```
vigsterkr: what happens with the seed?

tests/unit/ensemble/MajorityVote_unittest.cc
```diff
@@ -45,9 +45,10 @@ TEST(MajorityVote, binary_combine_vector)
 	expected.zero();
 	v.zero();
 
+	CRandom* m_rng = new CRandom();
```
vigsterkr: let's use more c++11 features:
```
auto m_rng = std::unique_ptr<CRandom>(new CRandom());
```

tests/unit/ensemble/MajorityVote_unittest.cc
```diff
@@ -72,12 +74,12 @@ TEST(MajorityVote, multiclass_combine_vector)
 
 	v.zero();
 	hist.zero();
-
+	CRandom* m_rng = new CRandom();
```
vigsterkr: ```auto m_rng = std::unique_ptr<CRandom>(new CRandom());```

tests/unit/ensemble/WeightedMajorityVote_unittest.cc
```diff
@@ -11,14 +11,15 @@ void generate_random_ensemble_matrix(SGMatrix<float64_t>& em,
 	const SGVector<float64_t>& w)
 {
 	int32_t num_classes = 3;
+	CRandom* m_rng = new CRandom();
```
vigsterkr: ```auto m_rng = std::unique_ptr<CRandom>(new CRandom());```

tests/unit/ensemble/WeightedMajorityVote_unittest.cc
```diff
@@ -69,10 +71,10 @@ TEST(WeightedMajorityVote, binary_combine_vector)
 
 	expected.zero();
 	v.zero();
-
+	CRandom* m_rng = new CRandom();
```
vigsterkr: ```auto m_rng = std::unique_ptr<CRandom>(new CRandom());```

but later we can add a static function to CRandom that wraps this :)

src/shogun/mathematics/Math.h
MikeLing: no, I somehow forget to remove the seed in here.  If we want to init the seed, we should call CMath::init_random(seed)

src/shogun/mathematics/Math.h
MikeLing: same as above, it will be removed in next push

tests/unit/ensemble/MajorityVote_unittest.cc
MikeLing: sure, I will keep that in mind

src/shogun/statistical_testing/kernelselection/internals/MaxCrossValidation.cpp
```diff
@@ -102,7 +102,8 @@ void MaxCrossValidation::compute_measures()
 		compute.m_alpha=alpha;
 		compute.m_num_runs=num_runs;
 		compute.m_rejections=rejections;
-
+		compute.m_kfold_x->set_seed(12345);
+		compute.m_kfold_y->set_seed(12345);
```
MikeLing: Set a fix seed here is about to make CrossValidation in CQuadraticTimeMMD.select_kernel can choose 0.25 kernel width every time[1]. Let's see the CI result(ctest passed locally) and add find a way to set seed for it outside.

[1] https://github.com/shogun-toolbox/shogun/blob/develop/tests/unit/statistical_testing/KernelSelection_unittest.cc#L292

MikeLing:
oh, sorry! I rebase the pr :(  Include a lot of unrelated commit in here

vigsterkr:
@karlnapf @lambday could you help us figure out why these changes breaks these two unit tests
```
KernelSelectionMaxCrossValidation (Failed)
249 - unit-CrossValidationMMD
```

@MikeLing could you put here your specific question fro @karlnapf and @lambday i'm sure they coud help in that case.

src/shogun/statistical_testing/MMD.cpp
```diff
@@ -136,6 +139,16 @@ index_t CMMD::get_num_null_samples() const
...
+{
+	self->seed = p_seed;
+}
+
+index_t CMMD::get_random_seed() const
+{
+	return self->seed;
+}
+
 void CMMD::set_statistic_type(EStatisticType stype)
```
MikeLing: Hi  @karlnapf and @lambday, my question is like this:

In kernlelselect and other places related to CrossValidation, we need to set a specific random seed to make it can generate same result every time. Usually, before we remove sg_rand, it can been done like[ sg_rand->set_seed(12345)](https://github.com/shogun-toolbox/shogun/pull/3888/files#diff-ade8fd54db951771ca40a368b7e792deL53). Then it will pass to the CrossValidations like here[https://github.com/shogun-toolbox/shogun/pull/3888/files#diff-b4b05fef7370d38882f73b554c6230ebL19]. 

But after we remove the global random, it become a little difficult to make that happened, I think. Because in kernel selection test, we can't directly access its cv object like we do in [here](https://github.com/shogun-toolbox/shogun/pull/3888/files#diff-727ce9dc4098d551004a7bf7d6ce8c43R304) to set its seed. 

So I add one more parameter which is the [**seed**](https://github.com/shogun-toolbox/shogun/pull/3888/files#diff-1aeca6bef489a2963141948c747855f5R54), and make it default return an fixed seed(**12345**) and pass it to  MaxCrossValidation like [this]( https://github.com/shogun-toolbox/shogun/pull/3888/files#diff-e76beae1ecab5719393a08a37a5a5ef8R106). 

But I'm not sure if it's appropriate to return a fixed seed a default when we do kernel select, and I'm kind of doesn't understand why we [choose 0.25](https://github.com/shogun-toolbox/shogun/pull/3888/files#diff-ade8fd54db951771ca40a368b7e792deR290) as kernel width rather than others.

src/shogun/clustering/KMeansMiniBatch.cpp
```diff
@@ -131,11 +131,12 @@ SGVector<int32_t> CKMeansMiniBatch::mbchoose_rand(int32_t b, int32_t num)
 {
 	SGVector<int32_t> chosen=SGVector<int32_t>(num);
 	SGVector<int32_t> ret=SGVector<int32_t>(b);
+	auto rng = std::unique_ptr<CRandom>(new CRandom());
```
karlnapf: why is a new object created here? Rather than using `m_rng` ... just asking

src/shogun/mathematics/Math.cpp
```diff
@@ -36,6 +36,8 @@ int32_t CMath::LOGACCURACY         = 0; // 100000 steps per integer
 
 int32_t CMath::LOGRANGE            = 0; // range for logtable: log(1+exp(x))  -25 <= x <= 0
 
+CRandom* CMath::m_rng = new CRandom(12345);
```
karlnapf: why would that 12345 be there?

src/shogun/mathematics/Random.h
```diff
@@ -31,7 +32,7 @@ namespace shogun
 	 * number generator.
 	 *
 	 * */
-	class CRandom : public CSGObject
+	class CRandom
```
karlnapf: what is the thinking here? Not necessary?

src/shogun/statistical_testing/MMD.cpp
```diff
@@ -51,14 +51,17 @@ struct CMMD::Self
...
 	}
 
 	index_t num_null_samples;
 	EStatisticType stype;
 	ENullApproximationMethod null_approximation_method;
 	std::unique_ptr<CKernelSelectionStrategy> strategy;
+	int32_t seed;
 
 	static constexpr index_t DEFAULT_NUM_NULL_SAMPLES = 250;
+	static constexpr int32_t DEFAULT_NUM_OF_SEED = 12345;
```
karlnapf: this seems a bit weird to me, can you explain?

src/shogun/statistical_testing/MMD.cpp
karlnapf: * none of the links work
 * kernel width has to be chosen to something. 1 would be a better choice than 0.25 I agree arbitrary values are bad style
 * I don't like the default seed parameter at all
 * also, why not let the seed be arbitrary when not specified? And then some of the logic here has to be rewritten to work with a local rng. @lambday can you advide here?

src/shogun/statistical_testing/MMD.h
```diff
@@ -223,6 +223,16 @@ class CMMD : public CTwoSampleTest
 	index_t get_num_null_samples() const;
 
 	/**
+	 * Method to sets random seed for CrossValidation.
+	 *
+	 * @param p_seed seed for random generator.
+	 */
+	void set_random_seed(int32_t p_seed);
```
karlnapf: I think this is not needed

tests/unit/mathematics/Random_unittest.cc
```diff
@@ -127,149 +127,158 @@ TEST(Random, random_uint64_1_2)
...
 		double pbin=double(rnds[i])/n_runs*100*array_len;
 		EXPECT_GE(pbin, 99.0);
 	}
+	SG_FREE(prng);
 }
 
 TEST(Random, random_uint32_range)
 {
-	CMath::init_random(17);
+	CRandom* prng = new CRandom(17);
```
karlnapf: smart pointer?

tests/unit/mathematics/Random_unittest.cc
```diff
@@ -312,35 +320,37 @@ TEST(Random, random_float64_range)
...
 	float64_t min=1.0;
 	float64_t max=0.0;
 	for (uint32_t i=0; i<n_runs; i++)
 	{
-		float64_t r=CMath::random((float64_t) 0, (float64_t) 1.0);
+		float64_t r = prng->random((float64_t)0, (float64_t)1.0);
 		min=CMath::min(min, r);
 		max=CMath::max(max, r);
 	}
 	EXPECT_GE(max, 0.99999);
```
karlnapf: I have to say, these are very *nice* test to ensure correctness of randomness :D

tests/unit/mathematics/linalg/ConjugateOrthogonalCGSolver_unittest.cc
```diff
@@ -28,19 +28,19 @@ TEST(ConjugateOrthogonalCGSolver, solve)
 
 	// diagonal non-Hermintian matrix with random complex entries
 	SGVector<complex128_t> diag(size);
-	sg_rand->set_seed(100.0);
+	auto m_rng = std::unique_ptr<CRandom>(new CRandom(100));
```
karlnapf: why not use this above as well?

tests/unit/multiclass/tree/CARTree_unittest.cc
```diff
@@ -546,7 +546,7 @@ TEST(CARTree, form_t1_test)
 
 TEST(CARTree,cv_prune_simple)
 {
-	sg_rand->set_seed(10);
+	CMath::init_random(1);
```
karlnapf: why change the seed?

tests/unit/statistical_testing/internals/CrossValidationMMD_unittest.cc
```diff
@@ -185,7 +194,13 @@ TEST(CrossValidationMMD, unbiased_full)
 	cv.m_alpha=alpha;
 	cv.m_num_runs=num_runs;
 	cv.m_rejections=SGMatrix<float64_t>(num_runs*num_folds, num_kernels);
-	sg_rand->set_seed(12345);
+
+	// set seed like this is about to make CrossValidationSplitting will have a
+	// same seed during this test. Not sure if it's a good thing to do.
```
karlnapf: it is *not* please revert this

tests/unit/statistical_testing/internals/CrossValidationMMD_unittest.cc
```diff
@@ -280,7 +298,14 @@ TEST(CrossValidationMMD, unbiased_incomplete)
...
```
karlnapf: see above

tests/unit/statistical_testing/internals/CrossValidationMMD_unittest.cc
```diff
@@ -294,7 +319,10 @@ TEST(CrossValidationMMD, unbiased_incomplete)
 	permutation_mmd.m_stype=stype;
 	permutation_mmd.m_num_null_samples=num_null_samples;
 
-	sg_rand->set_seed(12345);
+	// set the seed for CrossValidationSplitting
```
karlnapf: superflous comment, and pls remove this

karlnapf:
I am confused why the meta example tests all work as they kind of rely on global random, or did I miss something?

src/shogun/clustering/KMeansMiniBatch.cpp
MikeLing: yes, you are right. It's no necessary in here....

src/shogun/mathematics/Random.h
MikeLing: Because we want to call m_rng->set_set_seed() in CSObject, so it's impossible to do that if CRandom still inherit from CSGObject.

karlnapf: of course, thx!

src/shogun/statistical_testing/MMD.cpp
MikeLing: sure, so basically as I comment in https://github.com/shogun-toolbox/shogun/pull/3888#discussion_r125174002, it's about to set default seed of random as sg_rand->set_seed(12345). And it's going to set a fixed seed in https://github.com/shogun-toolbox/shogun/pull/3888/files#diff-e76beae1ecab5719393a08a37a5a5ef8R101 for CrossValidationMMD. But it's impossible to do that (give a fixed seed to CrossValidationMMD inside), so that's only way I can pass a fixed seed as parameter to it from outside.

tests/unit/statistical_testing/internals/CrossValidationMMD_unittest.cc
MikeLing: ok, I will do this in next push

src/shogun/statistical_testing/MMD.h
MikeLing: yeah, you are right. so that's the problem I have when I'm working on the global random removal issue. Usually when we want to ensure all the module we call will have a fixed seed to ensure they have same output every time, we call **sg_rand->set_seed(seed)** But after we remove that:

- Do we want a fix seed for every module we used in the tests?(I guess yes :)

- How to pass a fix seed to nested module? like when we do kernel selection, we need CrossValidationMMD inside, is that necessary to change some places just to ensure we unit test can passed every time?

- Maybe we have a better way to deal with these unit tests to make sure we get same result every time rather than give a fixed seed?

Maybe I will have more questions about this, but please give me some guides above. Thanks a lot
@karlnapf @lambday

tests/unit/mathematics/Random_unittest.cc
MikeLing: sure, I will do it in next push

src/shogun/mathematics/Math.cpp
MikeLing: It will been removed in next push.

vigsterkr:
@karlnapf good question... we'll see when we through out the actual random implementation that we rely atm and change it to c++11 random engine.

vigsterkr:
@karlnapf note that this is still just a feature branch, so there's plenty of room for improvement, we just wanted to take one step in a time.

PR #3896, opened 2017-07-02
Put swig files into "swig"

lisitsyn:


lisitsyn:
Simple thing and tests are passing. Hence merging

PR #3898, opened 2017-07-03
Rename 'modshogun' to 'shogun'

lisitsyn:


karlnapf:
https://travis-ci.org/shogun-toolbox/shogun/jobs/249789498#L3545

lisitsyn:
I don't get it yet

lisitsyn:
@karlnapf ready to go. It is a major change so please take a look.

lisitsyn:
`shogun.Classifier` etc stops working once merged.

vigsterkr:
@lisitsyn we kind of deprecated that already a longer time ago :D (submodules)

vigsterkr:
@lisitsyn although i dont get why, i mean both in case of python as well as in case of java it'd be great to have a module for `classifier` etc. no?

lisitsyn:
@vigsterkr yeah namespaces are good but IIRC in the end they were too intertwined.

vigsterkr:
@lisitsyn but that's rather because we are not clear on that internally... or you reckon it's unavoidable..?

lisitsyn:
Yeah too many dependencies to have separate modules so we gave up :)

lisitsyn:
YOLO!

PR #3900, opened 2017-07-04
Move GPL code to gpl subdirectory

karlnapf:
WIP

Done:
 * classifier


karlnapf:
#3786

vigsterkr:
there's a problem actually with the cmake file i think; some `___` gets added from somewhere in the build directory path, see:
`build/src/shogun/CMakeFiles/libshogun.dir/__/gpl/shogun/`

vigsterkr:
although this is not related to this PR rather in general of the gpl build

karlnapf:
Fine to merge this despite the formatting errors? I will fix the cmake thing after having moved the files ...

PR #3902, opened 2017-07-05
Import CoreML definitions

lisitsyn:


karlnapf:
So what is next here?

lisitsyn:
Adding code to export models

PR #3908, opened 2017-07-06
Vectorize DotFeatures covariance/mean calculation

micmn:
Would be nice also to parallelize the covariance main loop but is not obvious how to do it properly since  it works on one vector at a time, tried a few variants with openmp but it didn’t improve; the straightforward way to speed-up get_cov would be to compute it by matrix prod (tried, approx. 4X on 4 threads), but this requires the storage of the centered data matrix while the current algorithm doesn’t need that.

### DIFF
--- a/src/shogun/features/DotFeatures.cpp
+++ b/src/shogun/features/DotFeatures.cpp
@@ -16,6 +16,7 @@
 #include <shogun/lib/Signal.h>
 #include <shogun/lib/Time.h>
 #include <shogun/mathematics/Math.h>
+#include <shogun/mathematics/linalg/LinalgNamespace.h>
 
 #ifdef HAVE_OPENMP
 #include <omp.h>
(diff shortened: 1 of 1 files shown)


karlnapf:
Maybe offer a flag for that? And so some clever auto settings of it....

karlnapf:
I'd be keen to have a flag that improves speed substantially here, trading off with storing the dataset twice. Most of the time that is fine anyways.

Thoughts?

micmn:
Yeah, seems reasonable, what about auto setting you mentioned? any
suggestion?

Il 09/lug/2017 17:35, "Heiko Strathmann" <notifications@github.com> ha
scritto:

karlnapf:
For now, I would just offer a flag "copy_data_for_speed" which would be appropriately documented, and enabled by default

PR #3918, opened 2017-07-09
Fix uninitialized memory read in AveragedPerceptron

micmn:
Once merged FreeBSD build on buildbot should be fine.


src/shogun/classifier/AveragedPerceptron.cpp
```diff
@@ -52,6 +52,7 @@ bool CAveragedPerceptron::train_machine(CFeatures* data)
 	ASSERT(num_vec==train_labels.vlen)
 	SGVector<float64_t> w(num_feat);
 	float64_t* tmp_w=SG_MALLOC(float64_t, num_feat);
+	memset(tmp_w, 0, sizeof(float64_t)*num_feat);
```
karlnapf: @micmn thanks for doing that.
BTW we should remove such code from Shogun, allocation of memory, memset, all that. We have more high level constructs that do all this
@MikeLing might be a nice one for you to change such code to using `SGVector`

PR #3921, opened 2017-07-10
Update tapkee_shogun's license to BSD.

iglesias:


src/shogun/lib/tapkee/tapkee_shogun.cpp
```diff
@@ -1,8 +1,4 @@
-/*
- * This program is free software; you can redistribute it and/or modify
- * it under the terms of the GNU General Public License as published by
- * the Free Software Foundation; either version 3 of the License, or
- * (at your option) any later version.
+/* This software is distributed under BSD 3-clause license (see LICENSE file).
```
karlnapf: Can we please have these type of license headers shogun wide? Rather than that long and ugly bulky text in every file?


@iglesias @lisitsyn @vigsterkr

iglesias: Sure, like done here, it is a one-liner referencing to the LICENSE file, @karlnapf.

karlnapf: I think there were some problems with that, but maybe I am wrong?

lisitsyn: It is vague whether it is legally ok but I like it pretty much

karlnapf: I think quite a few other projects do this no?

karlnapf:
I like the minimal LICENSE thing way more

PR #3923, opened 2017-07-11
Disable trained model serialization test for gpl files when building with bsd license

micmn:


karlnapf:
Merging as build that compiles unit test passed

PR #3924, opened 2017-07-11
Change SGMatrix::size() to signed integer to be consistent with index_t

micmn:


vigsterkr:
@micmn this one needs a rebase. could you please do it? 👍

PR #3928, opened 2017-07-13
Make CrossValidation emit fold evaluations (WIP)

geektoni:
@karlnapf this patch will lead to a result like this one below. The XValidation algorithm will emit each fold evaluation for each of the xvalidation run (this version doesn't emit the fold indices, because there is no meaningful way to represent them with Tensorboard). 
 
screenshot from 2017-07-13 16-03-17

The data show in the image were generated by [this](https://github.com/shogun-toolbox/shogun/blob/develop/examples/undocumented/libshogun/evaluation_cross_validation_regression.cpp) Shogun's example (with the addition of ParameterObserverScalar).

### DIFF
--- a/src/shogun/base/SGObject.cpp
+++ b/src/shogun/base/SGObject.cpp
@@ -806,6 +806,26 @@ bool CSGObject::type_erased_has(const BaseTag& _tag) const
 	return self->has(_tag);
 }
 
+class CSGObject::ParameterObserverList
+{
+public:
+	void register_param(
+	    const std::string& name, const std::string& type,
+	    const std::string& description)
+	{
+		m_list_obs_params[name] = std::make_pair(type, description);
+	}
+
+	ObsParamsList get_list() const
+	{
+		return m_list_obs_params;
+	}
+
+private:
+	/** List of observable parameters (name, description) */
+	ObsParamsList m_list_obs_params;
+};
+
 void CSGObject::subscribe_to_parameters(ParameterObserverInterface* obs)
 {
 	auto sub =
@@ -833,26 +853,6 @@ void CSGObject::observe_scalar(
 	m_subscriber_params->on_next(tmp);
 }
 
-class CSGObject::ParameterObserverList
-{
-public:
-	void register_param(
-	    const std::string& name, const std::string& type,
-	    const std::string& description)
-	{
-		m_list_obs_params[name] = std::make_pair(type, description);
-	}
-
-	ObsParamsList get_list() const
-	{
-		return m_list_obs_params;
-	}
-
-private:
-	/** List of observable parameters (name, description) */
-	ObsParamsList m_list_obs_params;
-};
-
 void CSGObject::register_observable_param(
     const std::string& name, const std::string& type,
     const std::string& description)
@@ -862,12 +862,11 @@ void CSGObject::register_observable_param(
 
 void CSGObject::list_observable_parameters()
 {
-	SG_INFO("List of observable parameters of object %s\n", get_name());
-	SG_PRINT("------");
+	SG_PRINT("List of %s's observable parameters\n", get_name());
 	for (auto const& x : param_obs_list->get_list())
 	{
 		SG_PRINT(
-		    "%s [%s]: %s\n", x.first.c_str(), x.second.first.c_str(),
+		    "* %s [%s]: %s\n", x.first.c_str(), x.second.first.c_str(),
 		    x.second.second.c_str());
 	}
 }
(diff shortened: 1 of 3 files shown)


vigsterkr:
BOOOOOOOOOOOOOOOM!

karlnapf:
Cool as f***!

@geektoni I am not sure I understand each plots of the picture. So you are saying that each point in there corresponds to a single run for a fixed fold idx?

If so, this doesnt make sense, as the folds are iid, you can just group all of them.
One might be interested in the inter fold variance, i.e. how much variance is there in a single repetition. You want to have access to

 * the fold results for a single repetition (i.e. a time series of length k, for k folds)
 * the fold results merged for all repetitions (this is k*num_repetitions numbers)

src/shogun/base/SGObject.cpp
```diff
@@ -862,12 +867,12 @@ void CSGObject::register_observable_param(
 
 void CSGObject::list_observable_parameters()
 {
-	SG_INFO("List of observable parameters of object %s\n", get_name());
-	SG_PRINT("------");
+	SG_PRINT("List of %s's observable parameters\n", get_name());
+	SG_PRINT("------\n");
```
karlnapf: I would remove this -----

src/shogun/base/SGObject.cpp
```diff
@@ -862,12 +867,12 @@ void CSGObject::register_observable_param(
...
 	for (auto const& x : param_obs_list->get_list())
 	{
 		SG_PRINT(
-		    "%s [%s]: %s\n", x.first.c_str(), x.second.first.c_str(),
+		    "* %s [%s]: %s\n", x.first.c_str(), x.second.first.c_str(),
```
karlnapf: can you give an example output of this?

src/shogun/base/SGObject.h
```diff
@@ -427,6 +427,12 @@ class CSGObject
 	/** Print to stdout a list of observable parameters */
 	void list_observable_parameters();
 
+	/** Enable/disable parameter emission */
```
karlnapf: needs doxygen `@param`

src/shogun/base/SGObject.h
```diff
@@ -635,6 +641,9 @@ class CSGObject
 
 	/** Subscriber used to call onNext, onComplete etc.*/
 	SGSubscriber* m_subscriber_params;
+
+	/** Enable parameter emission */
```
karlnapf: Wrong grammar (minor) This is a parameter, not a method

src/shogun/evaluation/CrossValidation.cpp
```diff
@@ -59,6 +59,12 @@ void CCrossValidation::init()
 	SG_ADD((CSGObject**)&m_xval_outputs, "m_xval_outputs", "List of output "
 			"classes for intermediade cross-validation results",
 			MS_NOT_AVAILABLE);
+
+	/* Register observable paramaters of CrossValidation */
+	register_observable_param(
+	    "fold_result_0", "scalar",
```
karlnapf: as said, we don't need to distinguish per fold but rather per run

src/shogun/evaluation/CrossValidation.cpp
```diff
@@ -59,6 +59,12 @@ void CCrossValidation::init()
...
+	    "Intermediate fold evaluation results. There will be many parameters "
```
karlnapf: Don't think this "there will be many" is instructive. Rather name them self-exlaining

vigsterkr:
@karlnapf how do you imagine this to work/demonstrate in a notebook? note that TB is a separate JS app

src/shogun/base/SGObject.cpp
```diff
@@ -829,30 +851,13 @@ void CSGObject::subscribe_to_parameters(ParameterObserverInterface* obs)
 void CSGObject::observe_scalar(
     const int64_t step, const std::string& name, const Any& value)
 {
+	if (!m_enable_emit)
```
vigsterkr: rx will take care of this internally, no need to do ourselves:
https://github.com/Reactive-Extensions/RxCpp/blob/master/Rx/v2/src/rxcpp/subjects/rx-subject.hpp#L185

karlnapf:
@vigsterkr as explained in irc, I think we should have some SWIG exposed observer that can access such things ...

vigsterkr:
@karlnapf yeah you dont want TB actually.... but observable structures all around

vigsterkr:
@karlnapf FYI https://github.com/shogun-toolbox/shogun/blob/develop/src/interfaces/swig/ParameterObserver.i

src/shogun/base/SGObject.cpp
geektoni: A possible output is something like:
```
List of CrossValidation's observable parameters
* fold_result_0 [scalar]: Intermediate fold evaluation results. There will be many parameters fold_result_i, where i is between 0 and the number of folds.
```

geektoni:
I close this PR for now because it requires serious reworking of some parts to make it more similar to what you are looking for @karlnapf (I'm talking about being able to access to single runs results, like folds, fold indices etc. without Tensorboard). 

I'll reopen it when things will be more defined (so that I don't pollute Travis with useless builds). 

@vigsterkr

PR #3931, opened 2017-07-14
Fixes to CombinedFeatures CombinedKernel segfaulting on mismatch between numbers of features/kernels

HeyItsDavid:


vigsterkr:
thnx for the fix!
there are some codestyle errors: https://travis-ci.org/shogun-toolbox/shogun/jobs/253777248#L749

vigsterkr:
@HeyItsDavid heheh there's still one more:
```-		SG_SERROR("feature index out of bounds (%d >= %d)\n", idx, get_num_feature_obj());

+		SG_SERROR(

+		    "feature index out of bounds (%d >= %d)\n", idx,

+		    get_num_feature_obj());```

File: src/shogun/features/CombinedFeatures.cpp [hunk 1]
```diff
@@ -48,6 +48,10 @@ CCombinedFeatures::~CCombinedFeatures()
 
 CFeatures* CCombinedFeatures::get_feature_obj(int32_t idx)
 {
+	if (idx >= get_num_feature_obj())
+	{
+		SG_SERROR("feature index out of bounds (%d >= %d)\n", idx, get_num_feature_obj());
```
karlnapf: This should be 
```
REQUIRE(idx < get_num_feature_obj() && idx>=0, "Feature index (%d) must be within [%d, %d]", idx, 0, get_num_feature_obj()-1);
```

src/shogun/kernel/CombinedKernel.h
```diff
@@ -124,7 +124,14 @@ class CCombinedKernel : public CKernel
...
 		inline CKernel* get_kernel(int32_t idx)
 		{
-			return (CKernel*) kernel_array->get_element(idx);
+			if (idx < get_num_kernels())
+			{
+				return (CKernel*)kernel_array->get_element(idx);
+			}
+			else
+			{
+				return 0;
```
karlnapf: I think this is bad, an error should be thrown if the index is bad (either from the array, or by this class). 
Also if you do bound checks, also assert for >=0

src/shogun/kernel/CombinedKernel.cpp
```diff
@@ -121,8 +121,13 @@ bool CCombinedKernel::init(CFeatures* l, CFeatures* r)
 		// skip over features - the custom kernel does not need any
 		if (k->get_kernel_type() != K_CUSTOM)
 		{
-			lf = ((CCombinedFeatures*) l)->get_feature_obj(f_idx);
-			rf = ((CCombinedFeatures*) r)->get_feature_obj(f_idx);
+			if (((CCombinedFeatures*)l)->get_num_feature_obj() > f_idx &&
```
karlnapf: can you explain this?

src/shogun/kernel/CombinedKernel.h
HeyItsDavid: I agree with you -- however the implementation of combined kernel seems to expect 0 being returned if its out of bounds, hence why i went with that.

src/shogun/kernel/CombinedKernel.cpp
HeyItsDavid: Same thing as above, I tried preserving as much of the original code as possible (perhaps not the best choice). The original code seems to expect 0 being returned by convention if index is out of bounds when accessing data from multiple structures. However if you agree i'd be more than happy to rework it so that this isnt the case any more.

src/shogun/kernel/CombinedKernel.h
karlnapf: Ah I see.
The old code does this horrible thing ;)
In that case, I think it is OK to do this here. (Though a patch that cleans this is always welcome)

vigsterkr:
@HeyItsDavid i know it must be pain but could you fix the last remaining indentation error in the patch:
https://travis-ci.org/shogun-toolbox/shogun/jobs/253919606#L750

thnx

karlnapf:
BTW the indentations can be checked locally, or even corrected locally, see the error message in travis for how to do that

PR #3933, opened 2017-07-17
Add KernelMachines to generated serialization tests

micmn:
Machines that are failing: CDomainAdaptationSVM, CMKLRegression, CKRRNystrom, CMKLClassification, CMKLOneClass, CSVM, CLibSVMOneClass.

~Most of them should be failing to due custom initialization needed, I'll check if that's the case.~

KRRNystrom was failing due to wrong inizialization, I made a small modification to the solver.
LibSVMOneClass was failing due to serialization inaccuracy, switching to hdf5 format solves it.
The other ones need custom initialization (MKL, DomainAdaptation) and the generic SVM class doesn't implement a solver.

### DIFF
--- a/src/shogun/regression/KRRNystrom.cpp
+++ b/src/shogun/regression/KRRNystrom.cpp
@@ -58,7 +58,7 @@ less than number of data points (%d)\n", m_num_rkhs_basis, n);
 
 void CKRRNystrom::init()
 {
-	m_num_rkhs_basis=100;
+	m_num_rkhs_basis=0;
 }
 
 SGVector<int32_t> CKRRNystrom::subsample_indices()
@@ -78,6 +78,13 @@ SGVector<int32_t> CKRRNystrom::subsample_indices()
 bool CKRRNystrom::solve_krr_system()
 {
 	int32_t n=kernel->get_num_vec_lhs();
+
+	if (m_num_rkhs_basis == 0)
+	{
+		set_num_rkhs_basis((int32_t)CMath::ceil(n/2.0));
+		SG_SWARNING("Number of sampled rows not set, default is half (%d) "
+					"of the number of data points (%d)\n", m_num_rkhs_basis, n);
+	}
 
 	SGVector<float64_t> y=((CRegressionLabels*)m_labels)->get_labels();
 	if (y==NULL)
--- a/tests/unit/base/trained_model_serialization_unittest.cc.jinja2
+++ b/tests/unit/base/trained_model_serialization_unittest.cc.jinja2
@@ -4,12 +4,14 @@
  */
 
 #include <gtest/gtest.h>
+#include <shogun/kernel/GaussianKernel.h>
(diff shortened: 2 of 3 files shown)


src/shogun/regression/KRRNystrom.cpp
```diff
@@ -79,6 +79,14 @@ bool CKRRNystrom::solve_krr_system()
 {
 	int32_t n=kernel->get_num_vec_lhs();
 
+	if (n < m_num_rkhs_basis)
+	{
+		SG_SWARNING("Number of sampled rows (%d) must be less than or equal to "
```
karlnapf: I would make this a `REQUIRE`

tests/unit/base/trained_model_serialization_unittest.cc.jinja2
```diff
@@ -85,30 +113,50 @@ TEST_F(TrainedModelSerializationTest, {{class}})
...
+
+	CGaussianKernel *kernel = new CGaussianKernel(2.0);
+	kernel->init(train_feats, train_feats);
+	machine->set_kernel(kernel);
+	machine->set_labels(train_labels);
+
+	bool train_success=machine->train();
+	ASSERT_TRUE(train_success);
+
+	CLabels* predictions=machine->apply(test_feats);
```
karlnapf: There is the thing that kernel machines only need to store the data which has non-zero support vector (alpha coefficients). The method `store_model_features` sub-samples the training data and then just keeps the relevant ones.
This then makes the kernel machine, when serialized, store less data.

I think a fixture test that does the serialization with and without that call might be useful here, as it is for example used in cross-validation

What do you think @micmn @vigsterkr

src/shogun/regression/KRRNystrom.cpp
micmn: This is to make it work with generated tests since I can't call `set_num_rkhs_basis()`, otherwise I could set by default `m_num_rkhs_basis = 1` instead of `100`

karlnapf: I see ...
But the test is not very meaningful then, as KRR's point is to use less than all.
Maybe change the default to something like "half of the data" if m=0?

tests/unit/base/trained_model_serialization_unittest.cc.jinja2
micmn: Sure, the only difference is just calling `store_model_features()` before the serialization, right?
I'll address this in the next PR after merging this one since I have already some work done for GP and Multiclass.

karlnapf: Exactly! This also doesnt dump all of the data (for some kernel models at least)
Ace! For GP the API is not that stable, you can test regression classification, but also make sure to check the variance things that the GP can produce....

karlnapf: Yes! Ace!

PR #3935, opened 2017-07-17
Set LAPACK's dsyevr as default symmetric eigensolver + fixes

micmn:
Linag (Eigen):
- set LAPACK's dsyevr as default symmetric eigensolver if available;
- add option to compute only the top-k eigenvalues in symmetric eigensolver;
- split decompositions and eigensolvers implementations.

KernelPCA:
- compute just the needed (top-k) eigenvectors;
- fix sign correctness in unittest.

Fix wrap_dsyevr memory allocation.

### DIFF
--- a/src/shogun/mathematics/lapack.cpp
+++ b/src/shogun/mathematics/lapack.cpp
@@ -311,7 +311,7 @@ void wrap_dsyevr(char jobz, char uplo, int n, double *a, int lda, int il, int iu
 	double vl,vu;
 	double abstol = 0.0;
 	char I = 'I';
-	int* isuppz = SG_MALLOC(int, n);
+	int* isuppz = SG_MALLOC(int, 2 * (iu - il + 1));
 #ifdef HAVE_ACML
 	DSYEVR(jobz,I,uplo,n,a,lda,vl,vu,il,iu,abstol,m,
 	       eigenvalues,eigenvectors,n,isuppz,info);
--- a/src/shogun/mathematics/linalg/LinalgBackendEigen.h
+++ b/src/shogun/mathematics/linalg/LinalgBackendEigen.h
@@ -127,7 +127,7 @@ namespace shogun
 #define BACKEND_GENERIC_EIGEN_SOLVER_SYMMETRIC(Type, Container)                \
 	virtual void eigen_solver_symmetric(                                       \
 	    const Container<Type>& A, SGVector<Type>& eigenvalues,                 \
-	    SGMatrix<Type>& eigenvectors) const;
+	    SGMatrix<Type>& eigenvectors, index_t k) const;
 		DEFINE_FOR_NON_INTEGER_PTYPE(
 		    BACKEND_GENERIC_EIGEN_SOLVER_SYMMETRIC, SGMatrix)
 #undef BACKEND_GENERIC_EIGEN_SOLVER_SYMMETRIC
@@ -439,7 +439,7 @@ namespace shogun
 		template <typename T>
 		void eigen_solver_symmetric_impl(
 		    const SGMatrix<T>& A, SGVector<T>& eigenvalues,
-		    SGMatrix<T>& eigenvectors) const;
+		    SGMatrix<T>& eigenvectors, index_t k) const;
 
 		/** Eigen3 matrix in-place elementwise product method */
 		template <typename T>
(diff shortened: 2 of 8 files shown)


src/shogun/mathematics/lapack.cpp
```diff
@@ -311,7 +311,7 @@ void wrap_dsyevr(char jobz, char uplo, int n, double *a, int lda, int il, int iu
 	double vl,vu;
 	double abstol = 0.0;
 	char I = 'I';
-	int* isuppz = SG_MALLOC(int, n);
+	int* isuppz = SG_MALLOC(int, 2 * (iu - il + 1));
```
karlnapf: can you provide some insight here? :)

src/shogun/mathematics/linalg/LinalgBackendEigen.h
```diff
@@ -622,6 +622,20 @@ namespace shogun
 		template <typename T>
 		void zero_impl(SGMatrix<T>& a) const;
 	};
+
+/*
+ * Eigen's symmetric eigensolver uses a slower algorithm in comparison
+ * to LAPACK's dsyevr, so if LAPACK is available we use it for float64 type.
```
karlnapf: cool!

src/shogun/mathematics/linalg/LinalgNamespace.h
```diff
@@ -623,41 +623,53 @@ namespace shogun
 		}
 
 		/**
-		 * Compute the eigenvalues and eigenvectors of a symmetric matrix.
+		 * Compute the top-k eigenvalues and eigenvectors of a symmetric matrix.
```
karlnapf: Usually, solvers that compute few eigenvalues are slower than those computing all of them.
Are two different solvers used when k<n and k==n?
If so, should be mentioned here

src/shogun/mathematics/linalg/backend/eigen/EigenSolvers.cpp
```diff
@@ -0,0 +1,166 @@
...
+	Eigen::EigenSolver<typename SGMatrix<T>::EigenMatrixXt> solver(A_eig);
+
+	/*
+	 * checking for success
+	 *
+	 * 0: Eigen::Success. Decomposition was successful
+	 * 1: Eigen::NumericalIssue. The input contains INF or NaN values or
+	 * overflow occured
+	 */
+	REQUIRE(
```
karlnapf: This error message could be improved

src/shogun/mathematics/linalg/backend/eigen/EigenSolvers.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+	/*
+	 * checking for success
+	 *
+	 * 0: Eigen::Success. Eigenvalues computation was successful
+	 * 2: Eigen::NoConvergence. Iterative procedure did not converge.
+	 */
+	REQUIRE(
+	    solver.info() != Eigen::NoConvergence,
+	    "Iterative procedure did not converge!\n");
```
karlnapf: This could be improved

src/shogun/mathematics/linalg/backend/eigen/EigenSolvers.cpp
```diff
@@ -0,0 +1,166 @@
...
+	eigenvalues_eig = ev_eig.head(k);
+
+	/*
+	 * checking for success
+	 *
+	 * status == 0: successful exit
+	 * status < 0: the i-th argument had an illegal value
+	 * status > 0: internal error
+	 */
+	REQUIRE(!(status < 0), "The %d-th argument han an illegal value.", -status)
```
karlnapf: Very bad error messages (not useful for user). When printing errors, always assume the user of an algorithm gets them and has no idea what is wrong. So "The %d-th argument han an illegal value." is not optimal here

src/shogun/preprocessor/KernelPCA.cpp
```diff
@@ -78,6 +78,15 @@ bool CKernelPCA::init(CFeatures* features)
 		int32_t n = kernel_matrix.num_cols;
 		int32_t m = kernel_matrix.num_rows;
 		ASSERT(n==m)
+		if (m_target_dim > n)
+		{
+			SG_SWARNING(
+			    "Target dimension (%d) is not a valid value, it must be"
+			    "less or equal than the number of vectors."
```
karlnapf: missing whitespaces when strings are concatenated

src/shogun/mathematics/lapack.cpp
micmn: check http://www.netlib.org/lapack/explore-html/d2/d8a/group__double_s_yeigen_gaeed8a131adf56eaa2a9e5b1e0cce5718.html#gaeed8a131adf56eaa2a9e5b1e0cce5718
> ISUPPZ is INTEGER array, dimension ( 2*max(1,M) )

karlnapf: maybe that link should go into the code? But doesnt have to

src/shogun/mathematics/linalg/LinalgNamespace.h
micmn: Eigen computes all the eigenvalues and dsyevr should too but for the latter I'm not 100% so I'll double check.

micmn: Maybe it's better "compute all the eigenvalues and returns the top-k etc..."

karlnapf: Mmmmmh. I don't think that is a good API to offer.
If you compute all of them anyways, you can just return all, and then the caller picks the ones he wants.
There is also iterative methods that are much faster when only computing a few, and those are the ones that should get this API here, with a warning label ("slower for computing all")

PR #3936, opened 2017-07-18
LinalgRefactor - Drop legacy linalg codes

OXPHOS:


OXPHOS:
Sure! I'll get the new version ready ASAP

PR #3939, opened 2017-07-20
[ParametersObservers] Main refactor of the parameter's observers feature.

geektoni:
* Move all parameter observers into separate directory;
* Add ObservedValue class which will store a measurement;
* Refactor SGObject methods name (observe());
* Add SG_OBS_VALUE_TYPE to indicate what is the observed value for;
* Add macro to check the observed value type;
* Update register_observable_param by using SG_OBS_VALUE_TYPE as type;
* Move filter() implementation to ParameterObserverInterface.cpp;
* Refactor unit tests to use the new implementation;


src/shogun/base/SGObject.cpp
```diff
@@ -36,7 +37,7 @@
 namespace shogun
 {
 	typedef std::map<BaseTag, Any> ParametersMap;
-	typedef std::map<std::string, std::pair<std::string, std::string>>
+	typedef std::map<std::string, std::pair<SG_OBS_VALUE_TYPE, std::string>>
```
vigsterkr: does it need to be a sorted map? :)

vigsterkr: does it need to be a sorted map? :)

src/shogun/base/SGObject.cpp
```diff
@@ -815,33 +816,44 @@ void CSGObject::subscribe_to_parameters(ParameterObserverInterface* obs)
...
 public:
 	void register_param(
-	    const std::string& name, const std::string& type,
+	    const std::string& name, const SG_OBS_VALUE_TYPE type,
 	    const std::string& description)
 	{
 		m_list_obs_params[name] = std::make_pair(type, description);
 	}
 
+	std::string type_name(SG_OBS_VALUE_TYPE type)
```
vigsterkr: this is for debug purposes only?
or is there somewhere we use `type_name` internally?

src/shogun/lib/parameter_observers/ObservedValue.h
```diff
@@ -0,0 +1,209 @@
...
+		{
+		}
+
+		/**
+		 * Constructor for observed value which don't need a name or step
+		 * @param value the value of this observed parameters
+		 * @param type the type of this observed parameters
+		 */
+		ObservedValue(Any value, SG_OBS_VALUE_TYPE type)
+		    : m_step(0), m_name(""), m_value(value), m_type(type)
```
vigsterkr: that empty string as m_name feels a bit 🍶 🐑

src/shogun/lib/parameter_observers/ObservedValue.h
```diff
@@ -0,0 +1,209 @@
...
+
+	/**
+	 * Helper method to generate an ObservedValue (TensorBoard oriented)
+	 * @param step the step
+	 * @param name the param's name we are observing
+	 * @param value the param's value
+	 * @return an ObservedValue object initialized
+	 */
+	SG_FORCED_INLINE ObservedValue
+	make_observation(int64_t step, std::string& name, Any value)
```
vigsterkr: this could be a static function of ObservedValue, just like a factory function and returns a std::shared_ptr/some/*

src/shogun/lib/parameter_observers/ObservedValue.h
```diff
@@ -0,0 +1,209 @@
...
+	 * Observed value with a timestamp
+	 */
+	typedef std::pair<ObservedValue, time_point> TimedObservedValue;
+
+	/**
+	 * Helper method to convert a time_point to std::time_t
+	 * @param value time point we want to convert
+	 * @return the time point converted to std::time_t
+	 */
+	inline double convert_to_millis(const time_point& value)
```
vigsterkr: SG_FORCED_INLINE

src/shogun/base/SGObject.cpp
geektoni: Yes, there is a place where it is used https://github.com/shogun-toolbox/shogun/pull/3939/files#diff-9c3599c0d2090e493be261b079e9b63eR883

PR #3942, opened 2017-07-20
Add Multiclass machines to generated serialization tests

micmn:
MulticlassLabels: algorithms set labels by `set_int_label()` that sets only `m_labels` but this way
`m_current_values` is not initialized and hence equals() fails. [not sure how to handle this, zeroed vectors on allocation in DenseLabels for now].

MCLDA: incomplete params registration [fixed].
QDA: incomplete params registration + SGNDArray not serialized [fixed].

MulticlassLibSVM, ScatterSVM: Error: C <= 0, in fact it's set to zero few lines above [??, ignored].
MulticlassTreeGuidedLogisticRegression, CDomainAdaptationMulticlassLibLinear: need custom initialization [ignored].
ShareBoost: apply takes features subset [ignored].

Test store_model_features true/false in kernel machines
Use Some instead of raw pointers in macros
Add SGMatrix::get_submatrix() method

@karlnapf Don't think this was what you had in mind for get_submatrix so I wait your review, GP is just initial stage, I'd need some guidance on what we want to test.

### DIFF
--- a/src/shogun/lib/SGMatrix.cpp
+++ b/src/shogun/lib/SGMatrix.cpp
@@ -388,6 +388,14 @@ void SGMatrix<T>::create_diagonal_matrix(T* matrix, T* v,int32_t size)
 }
 
 template <class T>
+SGMatrix<T> SGMatrix<T>::submatrix(index_t col_start, index_t col_end) const
+{
+	assert_on_cpu();
+	return SGMatrix<T>(
+		get_column_vector(col_start), num_rows, col_end - col_start, false);
+}
+
+template <class T>
 SGVector<T> SGMatrix<T>::get_column(index_t col) const
 {
 	assert_on_cpu();
--- a/src/shogun/lib/SGMatrix.h
+++ b/src/shogun/lib/SGMatrix.h
@@ -144,6 +144,15 @@ template<class T> class SGMatrix : public SGReferencedData
 			return &matrix[c*num_rows];
 		}
 
+		/** Given a range of columns (start, end), return a view
+		 * of the matrix from column start to end excluded.
+		 * \warning The returned SGMatrix is non-owning!
+		 * @param col_start column index (inclusive)
+		 * @param col_end column index (excluded)
+		 * @return the submatrix
+		 */
+		SGMatrix<T> submatrix(index_t col_start, index_t col_end) const;
+
 		/** Map a column to a SGVector
 		 * \warning The returned SGVector is non-owning!
 		 * @param col column index
--- a/src/shogun/multiclass/QDA.h
+++ b/src/shogun/multiclass/QDA.h
@@ -207,8 +207,11 @@ class CQDA : public CNativeMulticlassMachine
 		/** feature means for each of the classes in the training data */
 		SGMatrix< float64_t > m_means;
 
-		/** matrices computed in training and used in classification */
-		SGNDArray< float64_t > m_M;
+		/** matrices computed in training and used in classification
+		 * the matrices are stacked horizontally into a matrix of size
+		 * (m_dim, m_dim*m_num_classes).
+		 */
+		SGMatrix<float64_t> m_M;
 
 		/** vector computed in training and used in classification */
 		SGVector< float32_t > m_slog;
--- a/tests/unit/base/trained_model_serialization_unittest.cc.jinja2
+++ b/tests/unit/base/trained_model_serialization_unittest.cc.jinja2
@@ -4,6 +4,7 @@
  */
 
 #include <gtest/gtest.h>
+#include <shogun/base/some.h>
 #include <shogun/kernel/GaussianKernel.h>
 #include <shogun/machine/Machine.h>
 #include <shogun/features/DenseFeatures.h>
@@ -14,13 +15,20 @@
 #include <shogun/io/SerializableHdf5File.h>
 #include <shogun/io/CSVFile.h>
 #include <shogun/io/SGIO.h>
+#include <shogun/machine/gp/ExactInferenceMethod.h>
+#include <shogun/machine/gp/GaussianLikelihood.h>
+#include <shogun/machine/gp/ProbitLikelihood.h>
+#include <shogun/machine/gp/SingleLaplaceInferenceMethod.h>
+#include <shogun/machine/gp/ZeroMean.h>
 #include "environments/LinearTestEnvironment.h"
 #include "environments/RegressionTestEnvironment.h"
+#include "environments/MultiLabelTestEnvironment.h"
 #include "utils/Utils.h"
 
 using namespace shogun;
 
 extern LinearTestEnvironment* linear_test_env;
+extern MultiLabelTestEnvironment* multilabel_test_env;
 extern RegressionTestEnvironment* regression_test_env;
 
 class TrainedModelSerializationTest : public ::testing::Test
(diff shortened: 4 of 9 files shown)


src/shogun/lib/SGMatrix.cpp
```diff
@@ -388,6 +388,15 @@ void SGMatrix<T>::create_diagonal_matrix(T* matrix, T* v,int32_t size)
 }
 
 template <class T>
+SGMatrix<T> SGMatrix<T>::get_submatrix(index_t first, index_t last) const
```
karlnapf: first last I would call col_start and col_end

src/shogun/lib/SGMatrix.h
```diff
@@ -144,6 +144,15 @@ template<class T> class SGMatrix : public SGReferencedData
 			return &matrix[c*num_rows];
 		}
 
+		/** Given a range of columns (first, last), return a view
+		 * of the matrix from column first to last excluded.
+		 * \warning The returned SGMatrix is non-owning!
+		 * @param first column index (inclusive)
```
karlnapf: Exactly that I had in mind

src/shogun/lib/SGMatrix.cpp
```diff
@@ -388,6 +388,15 @@ void SGMatrix<T>::create_diagonal_matrix(T* matrix, T* v,int32_t size)
...
+{
+	assert_on_cpu();
+	ASSERT(first >= 0 && last > first && last <= num_cols)
```
karlnapf: REQUIRE pls or no bound checking at all

src/shogun/multiclass/QDA.h
```diff
@@ -207,8 +207,12 @@ class CQDA : public CNativeMulticlassMachine
 		/** feature means for each of the classes in the training data */
 		SGMatrix< float64_t > m_means;
 
-		/** matrices computed in training and used in classification */
-		SGNDArray< float64_t > m_M;
+		/** matrices computed in training and used in classification
```
karlnapf: Actually, I think this can even stay like this.
Therefore, remove the TODO

tests/unit/base/trained_model_serialization_unittest.cc.jinja2
```diff
@@ -111,64 +132,103 @@ TEST_F(TrainedModelSerializationTest, LinearMachine_{{class}})
...
+	auto machine=some<{{class}}>();
+	load_data(machine->get_machine_problem_type());
+
+	CGaussianKernel* kernel=new CGaussianKernel(10, 2);
+	CZeroMean* mean=new CZeroMean();
+
+	CLikelihoodModel* lik;
+	if (machine->get_machine_problem_type() == PT_REGRESSION)
+		lik=new CGaussianLikelihood(1.0);
+	else
```
karlnapf: should be else if

tests/unit/base/trained_model_serialization_unittest.cc.jinja2
```diff
@@ -111,64 +132,103 @@ TEST_F(TrainedModelSerializationTest, LinearMachine_{{class}})
...
+		lik=new CProbitLikelihood();
+
+	CSingleLaplaceInferenceMethod* inf=new CSingleLaplaceInferenceMethod(
```
karlnapf: for classification you can use laplace, for regression maybe use exact
but the other approximations should also be tested, FITC, etc. See the notebooks

tests/unit/base/trained_model_serialization_unittest.cc.py
```diff
@@ -86,9 +102,15 @@ def entry(templateFile, input_file, config_file):
     classes = get_shogun_classes(tags)
     guards = read_defined_guards(config_file)
 
-    # Get all linear/kernel machines
+    # Get all linear/multiclass/kernel/gp machines
```
karlnapf: I think GP probably deserve their own treatment as the are so modular ?

micmn:
@karlnapf I want to add also the preprocessors, I think that the interesting ones are the dimension reduction preprocessors (PCA, KernelPCA, FisherLDA, missed something?), but their initializations differ so they'd need some factory method...thoughts?

karlnapf:
Yeah good idea.
I think a first step here should be to add tests that run preprocessors with all the models (no serialization yet), that might not work ;) so maybe one after the other?

micmn:
ok, so first I'll do GPs in a separate PR since it'll need some iteration to make it modular

micmn:
@karlnapf when you get a chance please review the changes to multiclass, I should have addressed your comments.

src/shogun/lib/SGMatrix.cpp
```diff
@@ -388,6 +388,14 @@ void SGMatrix<T>::create_diagonal_matrix(T* matrix, T* v,int32_t size)
 }
 
 template <class T>
+SGMatrix<T> SGMatrix<T>::get_submatrix(index_t col_start, index_t col_end) const
```
karlnapf: name change suggestion: just "submatrix"

karlnapf: (maybe even rename the existing "get_subvector" for consistency. But nevermind for now

tests/unit/base/trained_model_serialization_unittest.cc.jinja2
```diff
@@ -40,12 +48,23 @@ protected:
...
 		{
 			case PT_BINARY:
+			case PT_CLASS:
 			{
 				std::shared_ptr<GaussianCheckerboard> mockData =
 					linear_test_env->getBinaryLabelData();
 				train_feats = mockData->get_features_train();
 				test_feats = mockData->get_features_test();
-				train_labels = (CBinaryLabels *) mockData->get_labels_train();
+				train_labels = mockData->get_labels_train();
```
karlnapf: minor: "mockData"->"mock_data"

tests/unit/base/trained_model_serialization_unittest.cc.jinja2
```diff
@@ -40,12 +48,23 @@ protected:
...
+				break;
+			}
+
+			case PT_MULTICLASS:
+			{
+				std::shared_ptr<GaussianCheckerboard> mockData =
```
karlnapf: mock_data

tests/unit/base/trained_model_serialization_unittest.cc.jinja2
```diff
@@ -111,70 +133,63 @@ TEST_F(TrainedModelSerializationTest, LinearMachine_{{class}})
 //	machine->set_features(NULL);
 //	machine->set_labels(NULL);
 
-	CLabels* predictions=machine->apply(test_feats);
+	auto predictions=Some<CLabels>(machine->apply(test_feats));
```
karlnapf: this compiles?
"Some"->"some"

tests/unit/base/trained_model_serialization_unittest.cc.py
```diff
@@ -6,20 +6,30 @@
...
-        'CMultitaskL12LogisticRegression', 'CMultitaskLeastSquaresRegression',
-        'CMultitaskTraceLogisticRegression', 'CMultitaskClusteredLogisticRegression',
-        'CLatentSVM', 'CLatentSOSVM', 'CDomainAdaptationSVMLinear'
-    ],
-    'CKernelMachine' : [
-        'CDomainAdaptationSVM', 'CMKLRegression',
-        'CMKLClassification', 'CMKLOneClass', 'CSVM'
-    ]
-}
+# Ignore classes that need custom initialization and:
```
karlnapf: This comment confuses me. I understand that these classes are ignored.
But I wonder whether they are tested elsewhere, what custom initialization means, and the C<=0 error seems to be a bug that should be referenced with a github issue?

tests/unit/base/trained_model_serialization_unittest.cc.py
```diff
@@ -6,20 +6,30 @@
...
+# (Multiclass)SVM: doesn't implement a solver
+# ScatterSVM, MulticlassLibSVM: error C <= 0
```
karlnapf: comments should probably be put begind the lines with the individual classes below rather than all stacked up on top

tests/unit/base/trained_model_serialization_unittest.cc.jinja2
micmn: mmm actually it doesn't compile with "some"

karlnapf: I am confused!

micmn: maybe `wrap()` is the correct one?

micmn:
cleaned up, should be mergeable

Issue #3944, opened 2017-07-21
make a minimal test that everything linked correctly / etc


dougalsutherland:
It would be nice to be able to build some small executable that confirms the package linked correctly and so on, without building the full test suite or requiring ply / etc and building all the meta-examples.

Currently the conda-forge recipe is using `-DBUILD_EXAMPLES=ON` and then [running some of them](https://github.com/conda-forge/shogun-cpp-feedstock/blob/fdc376dac292233eb4c7870d4ff08cc77dc703aa/recipe/run_test.py) for that purpose, but (a) that means the package is distributing all of the examples for no real reason and (b) @karlnapf told me that those examples are going away. :)

karlnapf:
Done in #3945

dougalsutherland:
Cool, thanks!

PR #3945, opened 2017-07-21
Add a libshogun minimal example to build by default (#3944)

karlnapf:
-Introduce new cmake macro to add libshogun examples


karlnapf:
#3944

PR #3946, opened 2017-07-21
move src/gpl folder into submodule

karlnapf:
@vigsterkr @lisitsyn @iglesias let me know if you think something is missing


karlnapf:
Travis passed, I will merge if nobody complaints within today

OXPHOS:
So, 
[this included file](https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/classifier/svm/SVMLight.cpp#L35) is now in `gpl `submodule. 
Should we guard the whole method? Was this the idea?

iglesias:
Is there some link to that included file? Or do this include (and maybe others) need to be updated?

karlnapf:
Oh yes, SVMLight now needs to be guarded to only be available when `USE_GPL_SHOGUN` is set. I did not realize. Want to go ahead? The unit tests, meta examples, libshogun examples also need guarding then

vigsterkr:
Forgive my ignorance here but svmlight != gpl afaik

karlnapf:
Sure, but note it is not in the GPL repo but only in the main repo. And since it is not GPL code (i.e. not viral), it can stay there and just be disabled with a switch.
Furthermore, SVMLight has a GPL dependency, so it needs guarding

OXPHOS:
`pr_loqo.h` is now in `shogun-gpl/shogun/lib/external/` and SVMLight does require `pr_loqo` to work. Now the build will fail if `gpl` module is not linked to the main repo. I'll guard it if it's staying in main repo.

karlnapf:
Yes svmlight needs a gpl guard everywhere it is used in the main repo (code, tests, examples)

PR #3951, opened 2017-07-25
[PrematureStopping] Add premature stopping features to some classes.

geektoni:


src/shogun/evaluation/CrossValidation.h
```diff
@@ -169,6 +166,9 @@ class CCrossValidation: public CMachineEvaluation
 private:
 	void init();
 
+	/** evaluate */
+	virtual CEvaluationResult* evaluate_impl();
```
vigsterkr: private virtual?

src/shogun/regression/LeastAngleRegression.cpp
```diff
@@ -187,6 +187,8 @@ bool CLeastAngleRegression::train_machine_templated(CDenseFeatures<ST> * data)
 	int32_t nloop=0;
 	while (m_num_active < max_active_allowed && max_corr/n_vec > get_epsilon() && !stop_cond)
 	{
+		COMPUTATION_CONTROLLERS
```
vigsterkr: didn't we renamed this to YIELD_something? :)

geektoni: mmh apparently we didn't in the end. :P

src/shogun/evaluation/CrossValidation.cpp
```diff
@@ -122,9 +112,8 @@ CEvaluationResult* CCrossValidation::evaluate()
 
 	/* perform all the x-val runs */
 	SG_DEBUG("starting %d runs of cross-validation\n", m_num_runs)
-	for (index_t i=0; i <m_num_runs; ++i)
+	for (index_t i = 0; i < m_num_runs; i++)
```
vigsterkr: any particular reason for the postfix operator++?

geektoni: I would say it's clang-format fault. Anyway, no, there are no particular reasons for the postifx operator instead of the prefix one.

PR #3953, opened 2017-07-26
[ParametersObservers] Update CrossValidation with parameters observers.

geektoni:
This patch relies on #3939, hence only the last three commits are relevant. 

Changes:
* Refactored `CrossValidation` to use the new obserevers, instead of the previous listeners;
* Added `CrossValidationStorage` class which will store `CrossValidation`'s information;
* Added `ParameterObserverCV` to observer `CrossValidation`'s parameters;
* Added `ParameterObserverCV` tests;
* Ported `CrossValidationMKLStorage` to `ParameterObserverCVMKL`;
* Added a clear() method to `ParameterObserverInterface` to empty the observer from its observations
(if implemented);
* Added the new observers to SWIG;

Edit:
* Removed `CrossValidationPrintOutput` and incorporate it into `ParameterObserverCV`;
* Added `ParameterObsverCVMulticlass`;
* Deleted `CrossValidationOutput` and `CrossValidationMulticlassStorage`;
* Fixed some undocumented meta examples (libshogun and python);


vigsterkr:
needs rebase :)

geektoni:
@vigsterkr so this is almost complete. There are only two minor things which has to be addressed before merging it:
* One meta example (https://github.com/geektoni/shogun/blob/f0bd625b98f5a04ef67afcc934178497299a2c4b/examples/undocumented/libshogun/evaluation_cross_validation_mkl_weight_storage.cpp) leaks memory (here the valgrind report https://pastebin.com/V6NJ4B7z). I don't get why it happens since I should have placed all SG_REF/SG_UNREF correctly;
*  I am not able to expose directly to SWIG the observation vector of the `ParameterObserverCV` class (https://github.com/geektoni/shogun/blob/f0bd625b98f5a04ef67afcc934178497299a2c4b/src/shogun/lib/parameter_observers/ParameterObserverCV.h#L77), therefore interfaces will fail (like the Python one). I added the right `%template` directives (https://github.com/geektoni/shogun/blob/f0bd625b98f5a04ef67afcc934178497299a2c4b/src/interfaces/swig/Evaluation.i#L19), but still the vector is not represented as an iterable object;

src/interfaces/swig/Evaluation.i
```diff
@@ -76,3 +79,6 @@
 %include <shogun/evaluation/CrossValidationSplitting.h>
 %include <shogun/evaluation/StructuredAccuracy.h>
 %include <shogun/evaluation/DirectorContingencyTableEvaluation.h>
+
```
vigsterkr: `namespace shogun` ?

geektoni: It seems it is not needed to make things work.

examples/undocumented/libshogun/evaluation_cross_validation_mkl_weight_storage.cpp
```diff
@@ -50,6 +49,41 @@ void gen_rand_data(SGVector<float64_t> lab, SGMatrix<float64_t> feat,
...
+			/* Allocate memory needed */
+			if (!weights.matrix)
+			{
+				weights = SGMatrix<float64_t>(
+				    w.vlen, o->get_num_folds() * o->get_num_runs());
+			}
+
+			/* Copy the weights inside the matrix */
+			index_t run_shift =
+			    fold->get_current_run_index() * w.vlen * o->get_num_folds();
```
karlnapf: this method is a bit messy. I guess this can be done a bit cleaner?

examples/undocumented/python/evaluation_cross_validation_mkl_weight_storage.py
```diff
@@ -65,18 +64,23 @@ def evaluation_cross_validation_mkl_weight_storage(traindat=traindat, label_trai
...
     result=cross_validation.evaluate()
 
     # print mkl weights
-    weights=mkl_storage.get_mkl_weights()
-    #print "mkl weights during cross--validation"
-    #print weights
+    weights = []
+    for o in mkl_storage.get_observations():
+        for fold in o.get_folds_results():
+            machine = MKLClassification.obtain_from_generic(fold.get_trained_machine())
```
karlnapf: mmmh this cannot be translated into a meta example....
Maybe we can add an overly simplified one at least?
This one should go into the notebook on x-validation, we want to get rid of all language specific examples eventually

examples/undocumented/python/evaluation_cross_validation_multiclass_storage.py
```diff
@@ -71,19 +70,19 @@ def evaluation_cross_validation_multiclass_storage (traindat=traindat, label_tra
...
+    #multiclass_storage=ParameterObserverCV()
+    #multiclass_storage.append_binary_evaluation(F1Measure())
+    #cross_validation.subscribe_to_parameters(multiclass_storage)
+    #cross_validation.set_num_runs(3)
 
     # perform cross-validation
     result=cross_validation.evaluate()
 
-    roc_0_0_0 = multiclass_storage.get_fold_ROC(0,0,0)
+    #roc_0_0_0 = multiclass_storage.get_fold_ROC(0,0,0)
```
karlnapf: ?

src/shogun/classifier/mkl/MKLClassification.cpp
```diff
@@ -38,3 +38,16 @@ void CMKLClassification::init_training()
 	REQUIRE(m_labels->get_num_labels(), "Number of labels is zero.\n");
 	REQUIRE(m_labels->get_label_type() == LT_BINARY, "Labels must be binary.\n");
 }
+
+CMKLClassification* CMKLClassification::obtain_from_generic(CMachine* machine)
```
karlnapf: @micmn wanna do some magic to create a static `cast` method for *all* Shogun classes?
Seems like you would know how to do that fast and then we could delete all the manual ones (and increase coverage) Thoughts?
@vigsterkr

tests/unit/lib/parameter_observers/ParameterObserverCV_unittest.cc
```diff
@@ -0,0 +1,172 @@
...
+#include <shogun/lib/parameter_observers/ParameterObserverCV.h>
+#include <shogun/regression/KernelRidgeRegression.h>
+
+#include <gtest/gtest.h>
+#include <memory>
+
+using namespace shogun;
+
+/**
+ * This test was inspired by the meta example
```
karlnapf: should be inside one of the tests rather than in the outer file

tests/unit/lib/parameter_observers/ParameterObserverCV_unittest.cc
```diff
@@ -0,0 +1,172 @@
...
+CDenseFeatures<float64_t>* features = NULL;
+CRegressionLabels* labels = NULL;
+
+ParameterObserverCV* generate(bool locked = true)
+{
+
+	/* fill data matrix and labels */
+	SGMatrix<float64_t> train_dat(num_features, num_vectors);
+	SGVector<float64_t>::range_fill_vector(train_dat.matrix, num_vectors);
+	for (index_t i = 0; i < num_vectors; ++i)
```
karlnapf: fixture class for test data pls, this is redundant code

tests/unit/lib/parameter_observers/ParameterObserverCV_unittest.cc
```diff
@@ -0,0 +1,172 @@
...
+
+	/* clean up */
+	SG_UNREF(result);
+	SG_UNREF(cross);
+	SG_UNREF(features);
+
+	return par;
+}
+
+TEST(ParameterObserverCV, get_result_locked)
```
karlnapf: I guess "get_observations" would be more appropriate?

tests/unit/lib/parameter_observers/ParameterObserverCV_unittest.cc
```diff
@@ -0,0 +1,172 @@
...
+{
+	std::shared_ptr<ParameterObserverCV> par{generate(true)};
+
+	auto obs = par->get_observations();
+	for (int i = 0; i < 10; i++)
+	{
+		auto run = obs[i];
```
karlnapf: `ASSERT(run);`

examples/undocumented/python/evaluation_cross_validation_mkl_weight_storage.py
geektoni: I'm currently working on a simplified version (https://github.com/geektoni/shogun/blob/add_cross_validation_mkl_meta_example/examples/meta/src/evaluation/cross_validation_mkl_weight_storage.sg), which should show only the weights for one fold and one run (without cycling over all the observations).

tests/unit/lib/parameter_observers/ParameterObserverCV_unittest.cc
geektoni: What do you mean here? :P Should I keep only the test and remove the meta example? Or it's better to port the meta example and keep also the test?

src/shogun/classifier/mkl/MKLClassification.cpp
micmn: one issue I see is that the method declaration cannot be generated since it has to be in the header

Issue #3850, opened 2017-06-16
Random forest probability output 


karlnapf:
The current implementation only outputs binary labels:
https://gist.github.com/olinguyen/ade8e2bd0899b2ec9bee00d03fbb5172

It would be nicer if the forest would average the probabilistic predictions of each tree, as e.g. [done in sklearn](http://scikit-learn.org/stable/modules/ensemble.html#forest)

> In contrast to the original publication [B2001], the scikit-learn implementation combines classifiers by averaging their probabilistic prediction, instead of letting each classifier vote for a single class.

vigsterkr:
@karlnapf for this would you use https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/ensemble/MeanRule.h
once RF can output probabilities for classification?

karlnapf:
Yes in general an average is used. We will have to check the details though.

karlnapf:
Done in #3954

PR #3954, opened 2017-07-28
Added class probabilities for BaggingMachine (and RandomForest)

olinguyen:
Related to issue #3850.

BaggingMachines averages the probabilistic predictions of each tree and returns the Labels accordingly, depending on the Binary or Multiclass case.


src/shogun/machine/BaggingMachine.cpp
```diff
@@ -44,17 +47,68 @@ CBaggingMachine::~CBaggingMachine()
...
-	SGVector<float64_t> combined_vector = apply_get_outputs(data);
+	SGMatrix<float64_t> output = apply_outputs_without_combination(data);
+
+	CMeanRule* mean_rule = new CMeanRule();
+
+	SGVector<float64_t> labels = m_combination_rule->combine(output);
+	SGVector<float64_t> probabilities = mean_rule->combine(output);
+
+	CBinaryLabels* pred = new CBinaryLabels(labels);
+  pred->set_values(probabilities);
```
karlnapf: make sure your text editor doesnt hide formatting issues from you. In fact, one of the checks we run is a style checker, see its output how to convert the files to the right formatting and how to run locally

src/shogun/machine/BaggingMachine.cpp
```diff
@@ -44,17 +47,68 @@ CBaggingMachine::~CBaggingMachine()
...
+	CMulticlassLabels* pred = new CMulticlassLabels(num_samples);
+  pred->allocate_confidences_for(num_classes);
+
+  SGMatrix<float64_t> class_probabilities(num_samples, num_classes);
+  class_probabilities.zero();
+
+  for(int32_t i = 0; i < num_samples; ++i)
+  {
+    for(int32_t j = 0; j < m_num_bags; ++j)
+    {
```
karlnapf: it somehow annoys me that we treat binary and multiclass differently when the former is just a special case of the latter
Not your fault though, just generally a thing that we should think about @MikeLing @geektoni @micmn @vigsterkr

src/shogun/machine/BaggingMachine.cpp
```diff
@@ -44,17 +47,68 @@ CBaggingMachine::~CBaggingMachine()
...
+      int32_t class_idx = bagged_outputs(i, j);
+      class_probabilities(i, class_idx) += 1;
+    }
+  }
+
+  const float64_t alpha = 1.0 / m_num_bags;
+  class_probabilities = linalg::scale(class_probabilities, alpha);
```
karlnapf: ace!

src/shogun/machine/BaggingMachine.cpp
```diff
@@ -44,17 +47,68 @@ CBaggingMachine::~CBaggingMachine()
...
+  for(int32_t i = 0; i < num_samples; ++i)
+  {
+    for(int32_t j = 0; j < m_num_bags; ++j)
+    {
+      int32_t class_idx = bagged_outputs(i, j);
+      class_probabilities(i, class_idx) += 1;
+    }
+  }
+
+  const float64_t alpha = 1.0 / m_num_bags;
```
karlnapf: not sure why this needs to be const, but ok

src/shogun/machine/BaggingMachine.cpp
```diff
@@ -44,17 +47,68 @@ CBaggingMachine::~CBaggingMachine()
...
+  class_probabilities = linalg::scale(class_probabilities, alpha);
+
+  for(int32_t i = 0; i < num_samples; ++i)
+  {
+
+    SGVector<float64_t> confidences = class_probabilities.get_row_vector(i); 
+    int32_t y_pred = CMath::arg_max(confidences.vector, 1, confidences.vlen);
```
karlnapf: you can use the C++ keyword `auto' pretty much everywhere where you store a variable that is returned from a function

tests/unit/multiclass/BaggingMachine_unittest.cc
```diff
@@ -40,6 +188,7 @@ TEST(BaggingMachine, mock_train)
 	using ::testing::Mock;
 	using ::testing::DefaultValue;
 
+
```
karlnapf: try to avoid such whitespace changes in the future, but nevermind for now

tests/unit/multiclass/BaggingMachine_unittest.cc
```diff
@@ -27,11 +29,157 @@
 using namespace shogun;
 using ::testing::Return;
 
+class BaggingMachine : public ::testing::Test
```
karlnapf: very nice improvement!

tests/unit/multiclass/BaggingMachine_unittest.cc
```diff
@@ -233,8 +263,96 @@ TEST(BaggingMachine,classify_CART)
...
+	sg_rand->set_seed(1);
+
+	SGVector<bool> ft=SGVector<bool>(4);
+	ft[0]=true;
+	ft[1]=true;
+	ft[2]=true;
+	ft[3]=true;
+
+	CCARTree* cart=new CCARTree();
+	//CMajorityVote* cv=new CMajorityVote();
```
karlnapf: please no commented out lines in PRs :)

tests/unit/multiclass/BaggingMachine_unittest.cc
```diff
@@ -233,8 +263,96 @@ TEST(BaggingMachine,classify_CART)
...
+	EXPECT_EQ(1.0,res_vector[4]);
+
+	EXPECT_DOUBLE_EQ(1.0,values_vector[0]);
+	EXPECT_DOUBLE_EQ(0.3,values_vector[1]);
+	EXPECT_DOUBLE_EQ(0.3,values_vector[2]);
+	EXPECT_DOUBLE_EQ(1.0,values_vector[3]);
+	EXPECT_DOUBLE_EQ(0.7,values_vector[4]);
+
+	SG_UNREF(result);
+	SG_UNREF(c);
```
karlnapf: if you use `some` in the unit tests, then you don't have to UNREF at the end

tests/unit/multiclass/BaggingMachine_unittest.cc
```diff
@@ -233,8 +263,96 @@ TEST(BaggingMachine,classify_CART)
...
 	SG_UNREF(result);
 	SG_UNREF(c);
 	SG_UNREF(eval);
 }
+
+TEST_F(BaggingMachine, output_binary)
+{
+	sg_rand->set_seed(1);
+
+	SGVector<bool> ft=SGVector<bool>(4);
```
karlnapf: can't this first bit also be put into a common place. e.g. hierarchy of fixtures? so it is not replicated below?

tests/unit/multiclass/BaggingMachine_unittest.cc
```diff
@@ -233,8 +263,96 @@ TEST(BaggingMachine,classify_CART)
...
+	ft[1]=true;
+	ft[2]=true;
+	ft[3]=true;
+
+	CCARTree* cart=new CCARTree();
+	CMeanRule* cv=new CMeanRule();
+
+	cart->set_feature_types(ft);
+	CBaggingMachine* c=new CBaggingMachine(features_train,labels_train);
+	c->parallel->set_num_threads(1);
```
karlnapf: why?

tests/unit/multiclass/tree/RandomForest_unittest.cc
```diff
@@ -279,3 +282,128 @@ TEST(RandomForest,classify_non_nominal_test)
...
+	CMulticlassLabels* labels=new CMulticlassLabels(lab);
+
+	CRandomForest* c=new CRandomForest(feats, labels, 100,2);
+	c->set_feature_types(ft);
+	CMajorityVote* mv = new CMajorityVote();
+	c->set_combination_rule(mv);
+	c->parallel->set_num_threads(1);
+	c->train(feats);
+
+	SGMatrix<float64_t> test(4,5);
```
karlnapf: function (for the people that add tests after you)

karlnapf: or import from other file?

karlnapf:
Check the broken travis build. You can see the logs to see what parts don't compile /tests that fail, etc

tests/unit/multiclass/BaggingMachine_unittest.cc
```diff
@@ -35,6 +36,7 @@ class BaggingMachine : public ::testing::Test
   CDenseFeatures<float64_t>* features_test;
   CDenseFeatures<float64_t>* features_train;
   CMulticlassLabels* labels_train;
+  SGVector<bool>* ft;
```
karlnapf: why pointer?

karlnapf: in other words, please let this guy live on stack

tests/unit/multiclass/tree/RandomForest_unittest.cc
```diff
@@ -359,16 +359,12 @@ TEST(RandomForest, test_output)
 {
   sg_rand->set_seed(1);
 
-  SGMatrix<float64_t> data(2,4);
-  data(0,0) = -1.0;
-  data(0,1) = -1.2;
-  data(0,2) = -3.4;
-  data(0,3) = 1.1;
+	float64_t data_A[] = {-1.0, -1.0,
```
karlnapf: formatting ...

olinguyen:
- The tests failing on Travis don't seem to be from me. Could you confirm? 

> The following tests FAILED:
> 	299 - integration_meta_cpp-binary_classifier-lda (Failed)
> 	453 - generated_cpp-converter-ica_jedi_sep (OTHER_FAULT)
> 	454 - generated_cpp-converter-ica_ff_sep (OTHER_FAULT)
> 	455 - generated_cpp-converter-ica_jade (OTHER_FAULT)
> 	456 - generated_cpp-converter-ica_fast (OTHER_FAULT)
> 	457 - generated_cpp-converter-ica_sobi (OTHER_FAULT)

- Since the tests in RandomForest & BaggingMachine are redundant, may I remove the ones from RandomForest that appear in BaggingMachine?

-

olinguyen:
The BinaryLabels scores are computed as follows
`		labels[i] = src[i] + threshold >= 0 ? +1.0 : -1.0;
It's a little counter-intuitive when the scores are probabilities. Intuitively I would compare the score > threshold. Was this done like this for a reason?

karlnapf:
Yes the failing tests are not your fault probably...ill double check

karlnapf:
Yes remove redundant tests if they make no point

karlnapf:
For the labels, I agree, score>thresh is the way to do them.
Maybe you can find out where this constructor is called and then check why it is done this way?

src/shogun/machine/BaggingMachine.cpp
```diff
@@ -44,17 +47,53 @@ CBaggingMachine::~CBaggingMachine()
 
 CBinaryLabels* CBaggingMachine::apply_binary(CFeatures* data)
 {
-	SGVector<float64_t> combined_vector = apply_get_outputs(data);
+	SGMatrix<float64_t> output = apply_outputs_without_combination(data);
+
+	CMeanRule* mean_rule = new CMeanRule();
```
karlnapf: this will leak, you have to UNREF or make this a some

karlnapf: This reminds me.
Please test your new unit tests using valgrind to check for memory errors
The dev readme has sections on that
`valgrind ctest -R unit-test_name***`

src/shogun/machine/BaggingMachine.cpp
```diff
@@ -44,17 +47,53 @@ CBaggingMachine::~CBaggingMachine()
...
 
 CMulticlassLabels* CBaggingMachine::apply_multiclass(CFeatures* data)
 {
-	SGVector<float64_t> combined_vector = apply_get_outputs(data);
+	SGMatrix<float64_t> bagged_outputs = apply_outputs_without_combination(data);
+
+	int32_t num_samples = bagged_outputs.size() / m_num_bags;
+	int32_t num_classes = dynamic_cast<CMulticlassLabels*>(m_labels)->get_num_classes();
+
+	CMulticlassLabels* pred = new CMulticlassLabels(num_samples);
```
karlnapf: check whether this method is supposed to increase the ref counter for the returned object or if the caller is supposed to do that?

src/shogun/machine/BaggingMachine.cpp
```diff
@@ -44,17 +47,53 @@ CBaggingMachine::~CBaggingMachine()
...
 	return pred;
 }
 
 CMulticlassLabels* CBaggingMachine::apply_multiclass(CFeatures* data)
 {
-	SGVector<float64_t> combined_vector = apply_get_outputs(data);
+	SGMatrix<float64_t> bagged_outputs = apply_outputs_without_combination(data);
+
+	int32_t num_samples = bagged_outputs.size() / m_num_bags;
+	int32_t num_classes = dynamic_cast<CMulticlassLabels*>(m_labels)->get_num_classes();
```
karlnapf: this cast is not safe, so you need to store the casted pointer in a variable and then do a 
`REQUIRE(labels_multiclass, "Labels are not multiclass\n");`
or even better, try to convert them to multiclass (there are constructors to e.g. turn a binary labels into multiclass), and then say "Could not convert labels to multiclass"

tests/unit/multiclass/BaggingMachine_unittest.cc
```diff
@@ -27,11 +30,169 @@
...
+  virtual void TearDown()
+  {
+    SG_UNREF(features_train);
+    SG_UNREF(features_test);
+    SG_UNREF(labels_train);
+  }
+
+	void load_data()
+	{
+    sg_rand->set_seed(1);
```
karlnapf: formatting!
Use the style checker in travis to get the command that converts all your changes to the correct formatting, or do it manually

tests/unit/multiclass/BaggingMachine_unittest.cc
```diff
@@ -27,11 +30,169 @@
...
+  }
+ 
+  virtual void TearDown()
+  {
+    SG_UNREF(features_train);
+    SG_UNREF(features_test);
+    SG_UNREF(labels_train);
+  }
+
+	void load_data()
```
karlnapf: "load" not an appropriate name here. Rather "generate_toy_data"

tests/unit/multiclass/BaggingMachine_unittest.cc
```diff
@@ -27,11 +30,169 @@
...
+    lab[7]=0.0;
+    lab[8]=1.0;
+    lab[9]=1.0;
+    lab[10]=1.0;
+    lab[11]=1.0;
+    lab[12]=1.0;
+    lab[13]=0.0;
+    labels_train=new CMulticlassLabels(lab);
+
+    SGVector<bool> feature_types = SGVector<bool>(4);
```
karlnapf: auto

tests/unit/multiclass/BaggingMachine_unittest.cc
```diff
@@ -230,11 +267,78 @@ TEST(BaggingMachine,classify_CART)
...
+	EXPECT_EQ(0.0,res_vector[2]);
+	EXPECT_EQ(1.0,res_vector[3]);
+	EXPECT_EQ(1.0,res_vector[4]);
+
+  int32_t num_labels = result->get_num_labels();
+
+  for (int32_t i = 0; i < num_labels; ++i)
+  {
+    SGVector<float64_t> confidences = result->get_multiclass_confidences(i);
+    EXPECT_DOUBLE_EQ(1.0, linalg::sum(confidences));
```
karlnapf: such assertions should appear in the test name (if you write the test yourself)
i.e. output_mutliclass_probs_sum_to_one

src/shogun/machine/BaggingMachine.cpp
olinguyen: I'm a little confused when to use UNREF. When should it be used? There are some cases where new objects are created, UNREF is not run and there are no memory leaks.

karlnapf: You should use it if you created an object that is not passed to any other method.
If you create say `CBinaryLabels` and then pass it to a method `set_labels` this will increase the ref-counter by one, and the object that you called `set_labels` on will later on decrease the counter, which means the object will be deleted.
If you never pass it around, or increased the counter yourself, you need to unref later. If you don't see leaks, then nothing to worry

olinguyen:
What are the major things that need to be completed before we can close this PR?
I had a look at sklearn's test for the probability scores and I'm not quite sure how they came up with the expected values for the scores (https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/ensemble/tests/test_voting_classifier.py#L143) since randomness comes into play (they set random_state to a fixed value).

tests/unit/multiclass/BaggingMachine_unittest.cc
```diff
@@ -230,11 +264,78 @@ TEST(BaggingMachine,classify_CART)
...
+	c->set_bag_size(14);
+	c->set_num_bags(10);
+	c->set_combination_rule(cv);
+	c->train(features_train);
+
+	CBinaryLabels* result = c->apply_binary(features_test);
+	SGVector<float64_t> res_vector = result->get_labels();
+	SGVector<float64_t> values_vector = result->get_values();
+
+	EXPECT_EQ(1.0, res_vector[0]);
```
karlnapf: where do these numbers come from?

tests/unit/multiclass/BaggingMachine_unittest.cc
```diff
@@ -230,11 +264,78 @@ TEST(BaggingMachine,classify_CART)
...
+	c->set_num_bags(10);
+	c->set_combination_rule(cv);
+	c->train(features_train);
+
+	CMulticlassLabels* result = c->apply_multiclass(features_test);
+
+	SGVector<float64_t> res_vector = result->get_labels();
+	SGVector<float64_t> values_vector = result->get_values();
+
+	EXPECT_EQ(1.0, res_vector[0]);
```
karlnapf: where do these numbers come from?

tests/unit/multiclass/tree/RandomForest_unittest.cc
```diff
@@ -279,3 +282,126 @@ TEST(RandomForest,classify_non_nominal_test)
...
 }
+
+TEST(RandomForest, test_probabilities)
+{
+	sg_rand->set_seed(1);
+
+	SGMatrix<float64_t> data(4, 14);
+	SGVector<float64_t> lab(14);
+
+	generate_nm_data(data, lab);
```
karlnapf: generate_nm_data is not a good name for a function. It doesnt tell me what "nm" means

tests/unit/multiclass/tree/RandomForest_unittest.cc
```diff
@@ -279,3 +282,126 @@ TEST(RandomForest,classify_non_nominal_test)
...
+	test(3, 3) = weak;
+	test(3, 4) = strong;
+
+	CDenseFeatures<float64_t>* test_feats = new CDenseFeatures<float64_t>(test);
+
+	CBinaryLabels* result = c->apply_binary(test_feats);
+	SGVector<float64_t> res_vector = result->get_labels();
+	SGVector<float64_t> values_vector = result->get_values();
+
+	EXPECT_FLOAT_EQ(0.83, values_vector[0]);
```
karlnapf: where do these numbers come from?

tests/unit/multiclass/tree/RandomForest_unittest.cc
```diff
@@ -279,3 +282,126 @@ TEST(RandomForest,classify_non_nominal_test)
...
+	EXPECT_FLOAT_EQ(0.359999999999, values_vector[1]);
+	EXPECT_FLOAT_EQ(0.27, values_vector[2]);
+	EXPECT_FLOAT_EQ(0.91, values_vector[3]);
+	EXPECT_FLOAT_EQ(0.53, values_vector[4]);
+
+	CMulticlassAccuracy* eval = new CMulticlassAccuracy();
+	EXPECT_NEAR(0.642857, c->get_oob_error(eval), 1e-6);
```
karlnapf: where does this number come from?

tests/unit/multiclass/tree/RandomForest_unittest.cc
```diff
@@ -279,3 +282,126 @@ TEST(RandomForest,classify_non_nominal_test)
...
+
+	CRandomForest* c = new CRandomForest(features_train, labels_train, 10, 2);
+	SGVector<bool> ft = SGVector<bool>(2);
+	ft[0] = true;
+	ft[1] = true;
+	c->set_feature_types(ft);
+
+	CMeanRule* mr = new CMeanRule();
+	c->set_combination_rule(mr);
+	c->parallel->set_num_threads(1);
```
karlnapf: why single threaded?

tests/unit/multiclass/tree/RandomForest_unittest.cc
```diff
@@ -279,3 +282,126 @@ TEST(RandomForest,classify_non_nominal_test)
...
+	c->train(features_train);
+
+	CBinaryLabels* result = c->apply_binary(features_train);
+	SGVector<float64_t> res_vector = result->get_labels();
+	SGVector<float64_t> values_vector = result->get_values();
+
+	EXPECT_GT(0.5, values_vector[0]);
+	EXPECT_EQ(-1.0, res_vector[0]);
```
karlnapf: again: where do these come from?

tests/unit/multiclass/tree/RandomForest_unittest.cc
olinguyen: I didn't write these tests, but by changing the number of threads set to 1, the assertion for the oob error fails (exceed 1e-6 tolerance).

tests/unit/multiclass/tree/RandomForest_unittest.cc
olinguyen: Here I assumed that the previous test data values were correct, and just asserted that the probability had to be > 0.5 for a +1 output label

olinguyen:
The changes made in the other PR aren't reflected in this branch but will affect the apply_binary of BaggingMachine since it's dealing with probabilities with a 0.5 threshold. What'st the best way to go about this?

tests/unit/multiclass/tree/RandomForest_unittest.cc
karlnapf: Ok that is fine!

karlnapf: And also serves as an illustration why one should document the source of numbers in unit tests ;)

src/shogun/machine/BaggingMachine.cpp
```diff
@@ -44,17 +47,59 @@ CBaggingMachine::~CBaggingMachine()
...
 }
 
 CMulticlassLabels* CBaggingMachine::apply_multiclass(CFeatures* data)
 {
-	SGVector<float64_t> combined_vector = apply_get_outputs(data);
+	SGMatrix<float64_t> bagged_outputs =
+	    apply_outputs_without_combination(data);
+
+	auto labels_multiclass = dynamic_cast<CMulticlassLabels*>(m_labels);
+	REQUIRE(labels_multiclass, "Labels are not multiclass\n");
```
karlnapf: minor: missing period at end of sentence.
Better error message
`REQUIRE(labels_multiclass, "Labels (%s) are not compatible with multiclass.\n", m_labels->get_name());`
(This needs a `REQUIRE(m_labels, "Labels not set.\n");` to avoid segfaults for NULL labels

src/shogun/machine/BaggingMachine.cpp
```diff
@@ -44,17 +47,59 @@ CBaggingMachine::~CBaggingMachine()
...
+
+	int32_t num_samples = bagged_outputs.size() / m_num_bags;
```
karlnapf: this should be index_t (or auto)

src/shogun/machine/BaggingMachine.cpp
```diff
@@ -44,17 +47,59 @@ CBaggingMachine::~CBaggingMachine()
...
+		{
+			int32_t class_idx = bagged_outputs(i, j);
+			class_probabilities(i, class_idx) += 1;
+		}
+	}
+
+	float64_t alpha = 1.0 / m_num_bags;
+	class_probabilities = linalg::scale(class_probabilities, alpha);
+
+	for (int32_t i = 0; i < num_samples; ++i)
```
karlnapf: auto i

karlnapf: you can also use `range` here, see range.h

tests/unit/multiclass/BaggingMachine_unittest.cc
```diff
@@ -27,11 +30,168 @@
...
+	virtual void TearDown()
+	{
+		SG_UNREF(features_train);
+		SG_UNREF(features_test);
+		SG_UNREF(labels_train);
+	}
+
+	void generate_toy_data_weather()
+	{
+		sg_rand->set_seed(1);
```
karlnapf: pls dont fix the seed in here, this method is not random at all

tests/unit/multiclass/tree/RandomForest_unittest.cc
karlnapf: But you actually DID write this particular test here, no?

tests/unit/multiclass/tree/RandomForest_unittest.cc
```diff
@@ -279,3 +283,69 @@ TEST(RandomForest,classify_non_nominal_test)
...
+	CMeanRule* mr = new CMeanRule();
+	c->set_combination_rule(mr);
+	c->parallel->set_num_threads(1);
+	c->train(features_train);
+
+	auto result = c->apply_binary(features_test);
+	SGVector<float64_t> res_vector = result->get_labels();
+	SGVector<float64_t> values_vector = result->get_values();
+
+	EXPECT_EQ(-1.0, res_vector[0]);
```
karlnapf: so I guess these come from your manually created function and this just asserts that the RF is working correctly, right?

tests/unit/multiclass/tree/RandomForest_unittest.cc
```diff
@@ -279,3 +283,69 @@ TEST(RandomForest,classify_non_nominal_test)
 	SG_UNREF(c);
 	SG_UNREF(eval);
 }
+
+TEST(RandomForest, test_scores)
+{
+	// Test data generated using script
+	// https://gist.github.com/olinguyen/d93cd4222b404d26dc859f100d700be6
+	// Generates data for y = (x1 v x2) < 5 as decision boundary
+	sg_rand->set_seed(1);
```
karlnapf: dont think you need a seed for this test

tests/unit/multiclass/tree/RandomForest_unittest.cc
```diff
@@ -279,3 +283,69 @@ TEST(RandomForest,classify_non_nominal_test)
 	SG_UNREF(c);
 	SG_UNREF(eval);
 }
+
+TEST(RandomForest, test_scores)
```
karlnapf: not sure I get what this test does.
Also it needs a better (self-explaining) name

tests/unit/multiclass/BaggingMachine_unittest.cc
```diff
@@ -27,11 +30,168 @@
...
+	}
+
+	virtual void TearDown()
+	{
+		SG_UNREF(features_train);
+		SG_UNREF(features_test);
+		SG_UNREF(labels_train);
+	}
+
+	void generate_toy_data_weather()
```
karlnapf: wait a sec, is this function now defined multiple times?

tests/unit/multiclass/tree/RandomForest_unittest.cc
```diff
@@ -279,3 +283,165 @@ TEST(RandomForest,classify_non_nominal_test)
...
+	EXPECT_EQ(1.0, values_vector[8]);
+	EXPECT_EQ(1.0, values_vector[9]);
+
+	SG_UNREF(result);
+}
+
+TEST(RandomForest, score_compare_sklearn_toydata)
+{
+	// Comparison with sklearn's RandomForest probability outputs
+	// https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/ensemble/tests/test_voting_classifier.py#L143
```
karlnapf: better to reference an explicit commit hash in the link (master will change)

karlnapf: Also can you put a gist link?

tests/unit/multiclass/tree/RandomForest_unittest.cc
```diff
@@ -279,3 +283,165 @@ TEST(RandomForest,classify_non_nominal_test)
...
+	SGMatrix<float64_t> data(data_A, 2, 4, false);
+
+	CDenseFeatures<float64_t>* features_train =
+	    new CDenseFeatures<float64_t>(data);
+
+	float64_t labels[] = {0.0, 0.0, 1.0, 1.0};
+	SGVector<float64_t> lab(labels, 4);
+	CMulticlassLabels* labels_train = new CMulticlassLabels(lab);
+
+	float64_t expected_probabilities[] = {0.2, 0.2, 0.8, 0.7};
```
karlnapf: defined this next to the data so it is clear that it is coming from the gist

tests/unit/multiclass/tree/RandomForest_unittest.cc
```diff
@@ -279,3 +283,165 @@ TEST(RandomForest,classify_non_nominal_test)
...
+	SGVector<float64_t> values_vector = result->get_values();
+
+	EXPECT_EQ(-1.0, res_vector[0]);
+	EXPECT_EQ(-1.0, res_vector[1]);
+	EXPECT_EQ(+1.0, res_vector[2]);
+	EXPECT_EQ(+1.0, res_vector[3]);
+
+	for (auto i = 0; i < 4; ++i)
+	{
+		EXPECT_NEAR(expected_probabilities[i], values_vector[i], 11e-1);
```
karlnapf: Think twice about this check `11e-1==1.1 ` which is basically passing for *any* returned probability.
You will need to explore the output a bit here to make it consistent. It might be that the number of trees will have to be larger for this to stabilize ....

tests/unit/multiclass/BaggingMachine_unittest.cc
olinguyen: the RandomForest unit test also uses the same data. What's the best way to share data across unit tests?

tests/unit/multiclass/tree/RandomForest_unittest.cc
```diff
@@ -279,3 +283,165 @@ TEST(RandomForest,classify_non_nominal_test)
...
+
+	SG_UNREF(result);
+}
+
+TEST(RandomForest, score_random_labels)
+{
+	// Test data generated using script
+	// https://gist.github.com/olinguyen/d93cd4222b404d26dc859f100d700be6
+	// Generates data with random labels where trees don't all agree
+	sg_rand->set_seed(42);
```
karlnapf: dont think we should set a seed here

olinguyen: The probabilities change though unless I increase the # of trees as you stated earlier. think it can stabilize the results a bit

tests/unit/multiclass/tree/RandomForest_unittest.cc
```diff
@@ -279,3 +283,165 @@ TEST(RandomForest,classify_non_nominal_test)
...
+	c->set_feature_types(ft);
+
+	CMeanRule* mr = new CMeanRule();
+	c->set_combination_rule(mr);
+	c->train(features_train);
+	auto result = c->apply_binary(features_test);
+	SGVector<float64_t> res_vector = result->get_labels();
+	SGVector<float64_t> values_vector = result->get_values();
+
+	EXPECT_NEAR(0.7, values_vector[0], 1e-1);
```
karlnapf: thats better!
But I suggest in fact to

 * generate the data randomly (CMath::randn_double)
 * generate the labels randomly, sign(randn())
 * increase the dataset size
 * assert for 0.5

tests/unit/multiclass/BaggingMachine_unittest.cc
karlnapf: Put the function in a shared file and then include it

olinguyen:
Shall I remove the new tests for now, or can you merge the previous commits?

tests/unit/multiclass/tree/RandomForest_unittest.cc
olinguyen: woops, I meant to set it to 0.11 since it exceeded by 0.10000000000000003

tests/unit/multiclass/tree/RandomForest_unittest.cc
olinguyen: Sorry, since I'm using sklearn's toy data, what should be in this gist?

tests/unit/multiclass/BaggingMachine_unittest.cc
olinguyen: Is it appropriate to create a generate_test_data header and cpp file just for data generation? Should I place this in the tests/unit/utils folder? Sorry for asking a lot for this point, I just want to avoid re-changing it later.

vigsterkr: yeah you can use the utils part to add these type of generators/utility functions

olinguyen:
despite randomly generating the labels and the training set, asserting for 0.5 will fail in some test cases.

karlnapf:
Ok leave the tests in for now (as they are almost working and indeed making sense)

tests/unit/multiclass/tree/RandomForest_unittest.cc
karlnapf: Ah the data is fixed, sorry, then just leave it

tests/unit/multiclass/tree/RandomForest_unittest.cc
karlnapf: Ok 0.1 is acceptable.
If you run this 100 times, how many times does it fail?
You can do this quickly via some bash magic, something like
`for i in {1..100}; do ./shogun-unit-test --gtest_filter=TestName; done`

karlnapf:
If 0.5 sometimes fails (acceptable) then you can decrease the tolerance.
Run 100 times or so and make the tolerance so that it passes all of them (and still makes sense)

tests/unit/multiclass/BaggingMachine_unittest.cc
```diff
@@ -1,37 +1,78 @@
-#include "machine/MockMachine.h"
+#include "../utils/Utils.h"
```
vigsterkr: please dont do relative includes :(
i know that there are places where it is done the same way but 
```
#include "utils/Utils.h"
```
should do it.

tests/unit/multiclass/tree/RandomForest_unittest.cc
```diff
@@ -274,8 +148,187 @@ TEST(RandomForest,classify_non_nominal_test)
...
+
+	CMeanRule* mr = new CMeanRule();
+	c->set_combination_rule(mr);
+	c->train(features_train);
+
+	auto result = c->apply_binary(features_test);
+	SGVector<float64_t> res_vector = result->get_labels();
+	SGVector<float64_t> values_vector = result->get_values();
+
+	EXPECT_NEAR(0.5, values_vector[0], 5e-1);
```
vigsterkr: 0.5 is the error? :) so basically 0.1 or 1.0 is fine as well?

vigsterkr: or this is something that it's going to be addressed in a next pr?

olinguyen: it will be addressed in a future PR

karlnapf: No wait, I am confused here. What do you mean by addressed in a next pr?

olinguyen: we said we'd leave the tests as is and and iterate over them in another PR. Please correct my if there was a misunderstanding.

PR #3956, opened 2017-08-02
Initialising BinaryLabels & MulticlassLabels with probabilities scores

olinguyen:
- BinaryLabels constructor with confidence scores is more intuitive (comparing >= threshold)
- Added MulticlassLabels constructor with probability matrix

### DIFF
--- a/src/shogun/labels/BinaryLabels.cpp
+++ b/src/shogun/labels/BinaryLabels.cpp
@@ -54,7 +54,7 @@ CBinaryLabels::CBinaryLabels(SGVector<float64_t> src, float64_t threshold) : CDe
 	SGVector<float64_t> labels(src.vlen);
 	for (int32_t i = 0; i < labels.vlen; i++)
 	{
-		labels[i] = src[i] + threshold >= 0 ? +1.0 : -1.0;
+		labels[i] = src[i] >= threshold ? +1.0 : -1.0;
 	}
 	set_labels(labels);
 	set_values(src);
--- a/src/shogun/labels/MulticlassLabels.cpp
+++ b/src/shogun/labels/MulticlassLabels.cpp
@@ -20,25 +20,26 @@ CMulticlassLabels::CMulticlassLabels(const SGVector<float64_t> src) : CDenseLabe
 	set_labels(src);
 }
 
-CMulticlassLabels::CMulticlassLabels(const SGMatrix<float64_t> src)
+CMulticlassLabels::CMulticlassLabels(const SGMatrix<float64_t> confidences)
     : CDenseLabels()
 {
 	init();
-	int32_t n_classes = src.num_cols;
+	int32_t n_classes = confidences.num_cols;
 
-	auto labels = SGVector<float64_t>(src.num_rows);
+	SGVector<float64_t> labels(confidences.num_rows);
 	labels.zero();
 	set_labels(labels);
 
 	allocate_confidences_for(n_classes);
 
-	for (int32_t i = 0; i < src.num_rows; ++i)
+	for (int32_t i = 0; i < confidences.num_rows; ++i)
 	{
-		auto confidences = src.get_row_vector(i);
-		auto y_pred = CMath::arg_max(confidences.vector, 1, confidences.vlen);
+		auto confidences_i = confidences.get_row_vector(i);
+		auto y_pred =
+		    CMath::arg_max(confidences_i.vector, 1, confidences_i.vlen);
 
 		set_label(i, y_pred);
-		set_multiclass_confidences(i, confidences);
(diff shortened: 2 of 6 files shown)


src/shogun/labels/MulticlassLabels.cpp
```diff
@@ -20,6 +20,28 @@ CMulticlassLabels::CMulticlassLabels(const SGVector<float64_t> src) : CDenseLabe
 	set_labels(src);
 }
 
+CMulticlassLabels::CMulticlassLabels(const SGMatrix<float64_t> src)
```
karlnapf: `src` is not a good name, a better one would be `scores`

src/shogun/labels/MulticlassLabels.cpp
```diff
@@ -20,6 +20,28 @@ CMulticlassLabels::CMulticlassLabels(const SGVector<float64_t> src) : CDenseLabe
...
+    : CDenseLabels()
+{
+	init();
+	int32_t n_classes = src.num_cols;
+
+	auto labels = SGVector<float64_t>(src.num_rows);
```
karlnapf: somehow cleaner is: `SGVector<float64_t> labels(src.num_rows)`

src/shogun/labels/MulticlassLabels.cpp
karlnapf: or confidences

src/shogun/labels/MulticlassLabels.h
```diff
@@ -53,6 +53,12 @@ class CMulticlassLabels : public CDenseLabels
 
 		/** constructor
 		 *
+		 * @param src labels to set from a probability matrix
```
karlnapf: this is not necessarily a probability matrix right? Can be arbritary scores

karlnapf: You could say from a matrix that contains scores for each class for each point. Those could be probabilities but do not have to be normalized

tests/unit/labels/BinaryLabels_unittest.cc
```diff
@@ -66,4 +86,27 @@ TEST(BinaryLabels, serialization)
 		EXPECT_NEAR(labels->get_label(i), new_labels->get_label(i), 1E-15);
 	}
 	unlink(filename);
+
+	SG_UNREF(labels);
+	SG_UNREF(new_labels);
+}
+
+TEST_F(BinaryLabels, set_values_labels_from_constructor)
```
karlnapf: good!

tests/unit/labels/BinaryLabels_unittest.cc
```diff
@@ -66,4 +86,27 @@ TEST(BinaryLabels, serialization)
...
+{
+	CBinaryLabels* labels = new CBinaryLabels(probabilities);
+
+	SGVector<float64_t> labels_vector = labels->get_labels();
+	SGVector<float64_t> values_vector = labels->get_values();
+
```
karlnapf: you could put `ASSERT` here for the pointer not being NULL, and for the size of the vectors to be both equal to N

tests/unit/labels/MulticlassLabels_unittest.cc
```diff
@@ -43,3 +75,24 @@ TEST(MulticlassLabelsTest,confidences)
 	}
 	SG_UNREF(labels);
 }
+
+TEST_F(MulticlassLabelsTest, prob_matrix_label_initialization)
```
karlnapf: as said, this dont have to be probabilities

src/shogun/labels/BinaryLabels.cpp
```diff
@@ -54,7 +54,7 @@ CBinaryLabels::CBinaryLabels(SGVector<float64_t> src, float64_t threshold) : CDe
 	SGVector<float64_t> labels(src.vlen);
 	for (int32_t i = 0; i < labels.vlen; i++)
 	{
-		labels[i] = src[i] + threshold >= 0 ? +1.0 : -1.0;
+		labels[i] = src[i] >= threshold ? +1.0 : -1.0;
```
karlnapf: can you check whether and where this constructor is called. You are changing behaviour of existing code here, so need to account for that

olinguyen: It is called in apply_binary, but I set the default threshold value to 0, as it initially was, so it shouldn't be affecting the existing code.

karlnapf: ok!

karlnapf:
btw you committed changes to the `data` submodule and this is why the tests fail.
Needs to be removed before a merge

karlnapf:
Ok lets wait for travis and then merge

PR #3960, opened 2017-08-04
WIP for int32_t replacement

MikeLing:


src/shogun/features/StringFeatures.cpp
```diff
@@ -1973,8 +1973,8 @@ template<> void CStringFeatures<sg_type>::load(CFile* loader)		\
 																			\
 	SG_SET_LOCALE_C;													\
 	SGString<sg_type>* strs;												\
-	int32_t num_str;														\
-	int32_t max_len;														\
+	index_t num_str;														\
+	index_t max_len;														\
 	loader->f_load(strs, num_str, max_len);									\
```
MikeLing: Hi @vigsterkr , I got a bunch of warnings in this LOAD macro like https://pastebin.mozilla.org/9029036. All of them are about` 'num_str' is uninitialized when used here`. Do I need to give it a value like '0' or ?

vigsterkr: you can initialize them with 0 to make the compiler silent...

vigsterkr: you can initialize them with 0 to make the compiler silent...

src/shogun/kernel/Kernel.h
```diff
@@ -1064,9 +1064,9 @@ class CKernel : public CSGObject
 		bool lhs_equals_rhs;
 
 		/// number of feature vectors on left hand side
-		int32_t num_lhs;
+		index_t num_lhs;
 		/// number of feature vectors on right hand side
-		int32_t num_rhs;
+		index_t num_rhs;
```
MikeLing: @vigsterkr Hi, Does all these changes for Kernel.h make sense to u? I just want to make sure these are necessary changes so I can change other places related to kernel. Thank you :)

vigsterkr:
@MikeLing i did a rebase and small changes in `[wip] 2nd round` so you maybe wanna remove those two commits from your branch in order to resolve this merge conflict now :)

src/shogun/mathematics/linalg/linop/SparseMatrixOperator.cpp
```diff
@@ -129,8 +129,8 @@ SparsityStructure* CSparseMatrixOperator<T>::get_sparsity_structure(
 		}
 
 		// create the sparsity structure using the final power
-		int32_t* outerIndexPtr=const_cast<int32_t*>(matrix_power.outerIndexPtr());
-		int32_t* innerIndexPtr=const_cast<int32_t*>(matrix_power.innerIndexPtr());
+		index_t* outerIndexPtr=const_cast<index_t*>(matrix_power.outerIndexPtr());
+		index_t* innerIndexPtr=const_cast<index_t*>(matrix_power.innerIndexPtr());
```
MikeLing: Hi @vigsterkr , so the change in here is actually wrong, since `const_cast from 'StorageIndex *' (aka 'int *') to 'index_t *' (aka 'long long *') is not allowed`. But as you know the StorageIndex* is actually more complex than int32_t which can't change directly. So what should I do next? Thank you!

tests/unit/multiclass/tree/RandomForest_unittest.cc
```diff
@@ -272,7 +272,7 @@ TEST(RandomForest,classify_non_nominal_test)
 	EXPECT_EQ(1.0,res_vector[4]);
 
 	CMulticlassAccuracy* eval=new CMulticlassAccuracy();
-	EXPECT_NEAR(0.714285,c->get_oob_error(eval),1e-6);
+	EXPECT_NEAR(0.642857,c->get_oob_error(eval),1e-6);
```
MikeLing: Hi @vigsterkr and @karlnapf , so the int32_t replacement finally comes to the unit test refactoring. First of all, 
I'm not sure if these changes for `RandomForest_unittes`, `KernelSelection_unittest` and  `RandomCARTree_unittest.cc` make sense to you. But I remember we had discuss these test for random refactor and we need to change them somehow. So I just want to point it out in here :)

Secondly, for CMath permute in [here](https://github.com/shogun-toolbox/shogun/pull/3960/commits/9a6c83f25bf09897cc4b13529c1388a9f74b13a7#diff-fad5669b3799897497e99f4822939b91R387), I really don't know why the result changed. Does replace int32_t with index_t can make that change happened? I think I need to look into it later. And verify its result on Linux platform(I'm working on the OSX right now)

there still have some unit tests failed locally:

```
The following tests FAILED:
	179 - unit-ProbingSampler (Failed)
	285 - integration_meta_cpp-clustering-gmm (Failed)
	294 - integration_meta_cpp-distance-cosine (Failed)
```

And for ProbingSampler, it depend on colpack and looks like doesn't aceept `index_t`.(it can only use int, I think). I'm not sure if that's the reason why make the result different, will look into it later :)

MikeLing:
Hi @vigsterkr , for this pr we have:

1) it can build by [clang++ ](https://travis-ci.org/shogun-toolbox/shogun/jobs/267826695#L3263) but failed on [gcc](https://travis-ci.org/shogun-toolbox/shogun/jobs/267826696#L1659) with an error I don't know what's that mean. I also google it but haven't found something useful answer for it.

2) ProbingSampler.probing_samples_big_diag_matrix unit test failed

3) cosin_distance and gmm meta test failed but I'm not sure if it's because we use index_t or I mess up something inside. We could look more into it after we fix  issue (1) :)

tests/unit/mathematics/linalg/ProbingSampler_unittest.cc
```diff
@@ -106,8 +106,9 @@ TEST(ProbingSampler, probing_samples_big_diag_matrix)
 
 	// test coloring stuffs
 	SGVector<int32_t> coloring_vector=trace_sampler->get_coloring_vector();
+
 	EXPECT_EQ(trace_sampler->get_num_samples(), 1);
-	for (index_t i=0; i<coloring_vector.vlen; ++i)
+	for (int32_t i = 0; i < coloring_vector.vlen; ++i)
```
vigsterkr: the `i` here should be the type of `index_t` as you are comparing it with `coloring_vector.vlen` which is `index_t`

tests/unit/mathematics/linalg/ProbingSampler_unittest.cc
```diff
@@ -86,14 +86,14 @@ TEST(ProbingSampler, probing_samples_big_diag_matrix)
...
-	const index_t size=10000;
+	const int32_t size = 10000;
 	SGSparseMatrix<float64_t> sm(size, size);
 	CSparseMatrixOperator<float64_t>* op=new CSparseMatrixOperator<float64_t>(sm);
 	SG_REF(op);
 
 	// set its diagonal
 	SGVector<float64_t> diag(size);
-	for (index_t i=0; i<size; ++i)
+	for (int32_t i = 0; i < size; ++i)
```
vigsterkr: this should be `index_t` as well instead of `int32_t`

tests/unit/mathematics/linalg/ProbingSampler_unittest.cc
```diff
@@ -107,7 +107,7 @@ TEST(ProbingSampler, probing_samples_big_diag_matrix)
 	// test coloring stuffs
 	SGVector<int32_t> coloring_vector=trace_sampler->get_coloring_vector();
 
-	EXPECT_EQ(trace_sampler->get_num_samples(), 1);
+	EXPECT_EQ(1, trace_sampler->get_num_samples());
 	for (index_t i = 0; i < coloring_vector.vlen; ++i)
 		EXPECT_EQ(coloring_vector[i], 0);
```
MikeLing: @vigsterkr mmm, so I keep this part due to we have this error https://pastebin.mozilla.org/9030480, it seems like AllOf require two parameters more than one?

src/shogun/mathematics/linalg/ratapprox/tracesampler/ProbingSampler.cpp
```diff
@@ -171,8 +171,8 @@ void CProbingSampler::precompute()
 	for (std::vector<int32_t>::iterator it=vi_VertexColors.begin();
 		it!=vi_VertexColors.end(); it++)
 	{
-		int32_t i =
-			static_cast<int32_t>(std::distance(vi_VertexColors.begin(), it));
+		auto i =
+			static_cast<index_t>(std::distance(vi_VertexColors.begin(), it));
```
MikeLing: so the unit test for ProbSampler still failed :(

tests/unit/mathematics/linalg/ProbingSampler_unittest.cc
vigsterkr: ok so i guess then it would be:
```
EXPECT_THAT(coloring_vector, Each(0));
```

PR #3966, opened 2017-08-17
Recurrent Neural Networks

olinguyen:
Draft of the RNN module I'd like to add to Shogun. @karlnapf, I wanted to know if this makes sense on a high-level and have some feedback from you.

- How should I go about for using different activation functions (ReLU, tanh, sigmoid) without having to implement new classes e.g. NeuralRecurrentRectifiedLinearLayer, etc. Or is that the way we should go about it?


olinguyen:
Took some ideas from these places:

- http://www.wildml.com/2015/09/recurrent-neural-networks-tutorial-part-1-introduction-to-rnns/
- https://gist.github.com/karpathy/d4dee566867f8291f086
- https://github.com/llSourcell/recurrent_neural_network/blob/master/RNN.ipynb

grig-guz:
@olinguyen  Do you mind if I continue working on this one?

olinguyen:
@grig-guz Sure, please take over :)!

grig-guz:
@karlnapf in progress, maybe 1 more day

grig-guz:
@karlnapf Is there build-in Shogun implementation of 3D matrices, analogous to SGMatrix for 2D matrices? I need to be able to store multiple hidden state weight matrices at once during training to backprop through them.

vigsterkr:
@grig-guz you mean like ndarray?

grig-guz:
@vigsterkr yes! I've updated my question above, does it sound reasonable? Something like this:
https://github.com/shogun-toolbox/shogun/blob/8d578f454422dd1e390757271208a1107b332078/src/shogun/neuralnets/NeuralConvolutionalLayer.cpp#L72

vigsterkr:
@grig-guz there is actually `SGNDArray` but if the matrices are of the same dimensions you are probably better off storing it in an `std::vector<SGMatrix<Type>>` or something

grig-guz:
@vigsterkr sure, that looks better. Thank you!

gauravcr7rm:
Can I start from scratch for RNN as here one of the checks was not successful.
I also want to add LSTM too with it.
Please give me green signal if that will prove fruitful for the organisation.
Thanks in advance

karlnapf:
@gauravcr7rm feel free to start with this. But I would actually encourage you to take one of the other implementations, and modernize them a bit (use `lingalg`, SIMD, parallel loops, and better APIs). This is a good warm up to get a feeling for how those things are built within shogun. Then you can add a recurrent network. So green lights go ahead :)

karlnapf:
Closing this one as inactive

PR #3968, opened 2017-08-18
[Features-Refactor] Implement DotIterator

micmn:
Iterators:
- implement DotIterator
- Perceptron training with iterators + unit test

Linalg (separate commit since I'll need it in another PR):
- add element-wise product between vectors (there was only matrix)
- fix add()/element_prod() parameters constness so that rvalues can be used too


micmn:
Nice, swig doesn't support nested classes.

lisitsyn:
I missed this PR but this looks great. What is missed to merge?

PR #3989, opened 2017-10-15
Use the correct type.

zoq:
Just looked into the sparse matrix implementation and found a minor issue.


karlnapf:
Thanks Marcus!

Ouch, https://travis-ci.org/shogun-toolbox/shogun/jobs/288298933#L5120
I will look into this!

PR #4003, opened 2017-12-04
Copy toy data with examples during make install

### LINKED ISSUE #3997: make install should install toy data with examples
The (meta) examples need `data/toy`, i.e. a subfolder of the `data` submodule.
`make install` should ensure this is installed with the examples under `prefix/share/shogun/examples` in a way that the examples work out of the box after installing.

Currently, one has to do something like this (I here link the folder)
```
ln -s /opt/shogun/data/toy/ /usr/local/share/shogun/examples/data
cd /usr/local/share/shogun/examples/python/binary_classifier
python kernel_svm.py
```

durovo:
Fixes #3997


karlnapf:
This should be done via the data repository installation. Sorry about the issue, was my mistake ..

Issue #4004, opened 2017-12-05
UNABLE TO UPDATE SUBMODULES FROM BEHIND A PROXY SERVER


sudz123:
The command "git submodule update --init" and "git clone https://github.com/shogun-toolbox/shogun.git" does not work from behind a proxy server even though proxy has been configured in the terminal.

karlnapf:
This is not Shogun related. Google for general solution

PR #4005, opened 2017-12-05
SOLUTION TO ISSUE #4004

sudz123:
Adding the given lines of code before cloning allows a user running from behind a proxy server to update the to the latest submodules without encountering any errors.


karlnapf:
Thanks anyways! Generally, such patches are really appreciated. Maybe ping us in irc before sending the PR
Let me know if you need any pointers.

vigsterkr:
as said this is great but this is a general git related thing so it shoulnd't be part of our manual that should be as simple as possible and concentrate on things that are related to the library itself

sudz123:
Sure,

My sincere apologies. Yes I too agree that the solution is not shogun
specific but it's of great help to a novice.

Best regards

Issue #3999, opened 2017-12-03
Parallel building with meta examples fails


orivej:
I'm building shogun 6.1.2 on Linux, and the parallel building fails because [here](https://github.com/shogun-toolbox/shogun/blob/bc89ca00981a6967310c90c8843a6d19f43c0f11/examples/meta/generator/parse.py#L22) the [generator](https://github.com/shogun-toolbox/shogun/blob/bc89ca00981a6967310c90c8843a6d19f43c0f11/examples/meta/CMakeLists.txt#L68) invokes yacc without explicit `tabmodule` argument, and the default is overwritten by parallel generators.

The error may look like this:
```
[12/872] Generating example meta_api-enums
FAILED: examples/meta/cpp/meta_api/enums.cpp examples/meta/python/meta_api/enums.py examples/meta/octave/meta_api/enums.m examples/meta/java/meta_api/enums.java examples/
meta/ruby/meta_api/enums.rb examples/meta/csharp/meta_api/enums.cs examples/meta/r/meta_api/enums.R examples/meta/lua/meta_api/enums.lua
cd /tmp/nds-build-shogun/source/build/examples/meta && /nix/store/9n2npaz7xy8b4q0a8k28nr3sjzfy625i-python-2.7.14/bin/python /tmp/nds-build-shogun/source/examples/meta/gen
erator/generate.py -i /tmp/nds-build-shogun/source/examples/meta/src/meta_api/enums.sg -o /tmp/nds-build-shogun/source/build/examples/meta -t /tmp/nds-build-shogun/source
/examples/meta/generator/targets -g /tmp/nds-build-shogun/source/build/src/shogun/tags --parser_files_dir=/tmp/nds-build-shogun/source/build/examples/meta/parser_files
Traceback (most recent call last):
  File "/tmp/nds-build-shogun/source/examples/meta/generator/generate.py", line 158, in <module>
    generatedFilesOutputDir=args.parser_files_dir)
  File "/tmp/nds-build-shogun/source/examples/meta/generator/generate.py", line 74, in translateExamples
    ast = parse(file.read(), filePath, generatedFilesOutputDir)
  File "/tmp/nds-build-shogun/source/examples/meta/generator/parse.py", line 320, in parse
    parser = FastParser(generatedFilesOutputDir)
  File "/tmp/nds-build-shogun/source/examples/meta/generator/parse.py", line 25, in __init__
    debug=generateFiles)
[...]

PR #4008, opened 2017-12-06
Fix meta example parser bug in parallel builds

sorig:
Solves #3999.

@vigsterkr Is the CMake change OK? I'm not sure if it's clean or hacky (not very experienced with CMake)

### DIFF
--- a/examples/meta/CMakeLists.txt
+++ b/examples/meta/CMakeLists.txt
@@ -23,8 +23,18 @@ LIST(APPEND GENERATOR_DEPENDENCIES ${CMAKE_SOURCE_DIR}/examples/meta/generator/t
 FILE(GLOB TARGET_LANGUAGES ${CMAKE_SOURCE_DIR}/examples/meta/generator/targets/*.json)
 LIST(APPEND GENERATOR_DEPENDENCIES ${TARGET_LANGUAGES})
 
-# parser output
-file(MAKE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/parser_files)
+# parser generation
+set(GENERATED_PARSER_FILES)
+LIST(APPEND GENERATED_PARSER_FILES ${CMAKE_CURRENT_BINARY_DIR}/parser_files/parsetab.py)
+LIST(APPEND GENERATED_PARSER_FILES ${CMAKE_CURRENT_BINARY_DIR}/parser_files/lextab.py)
+add_custom_command(
+    OUTPUT ${GENERATED_PARSER_FILES}
+    DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/generator/parse.py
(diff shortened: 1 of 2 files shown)


examples/meta/CMakeLists.txt
```diff
@@ -25,6 +25,10 @@ LIST(APPEND GENERATOR_DEPENDENCIES ${TARGET_LANGUAGES})
 
 # parser output
 file(MAKE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/parser_files)
+execute_process(COMMAND ${PYTHON_EXECUTABLE}
```
vigsterkr: note from cmake manual: "(execute_process) run while CMake is processing the project prior to build system generation."
i'm not so sure we wanna have this.... because you wanna run this when you build the target, i.e. the meta examples :)
for that use `add_custom_target` or `add_custom_command` and make sure that whoever depends on the output of `parse.py` depends on this new target/command. dont worry cmake will make sure that this target will be only called once and not as many times somebody depends on it.

vigsterkr: there are couple of examples within shogun cmake files how to use add_custom_target or add_custom_command, but if you are in doubt ask :)

sorig:
@vigsterkr Makes sense. I thought the other solution was a bit fishy. Have a look again :)

vigsterkr:
great! lets see what CIs say :)

examples/meta/CMakeLists.txt
```diff
@@ -23,8 +23,18 @@ LIST(APPEND GENERATOR_DEPENDENCIES ${CMAKE_SOURCE_DIR}/examples/meta/generator/t
...
 
-# parser output
-file(MAKE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/parser_files)
+# parser generation
+set(GENERATED_PARSER_FILES)
+LIST(APPEND GENERATED_PARSER_FILES ${CMAKE_CURRENT_BINARY_DIR}/parser_files/parsetab.py)
+LIST(APPEND GENERATED_PARSER_FILES ${CMAKE_CURRENT_BINARY_DIR}/parser_files/lextab.py)
+add_custom_command(
+    OUTPUT ${GENERATED_PARSER_FILES}
+    DEPENDS ${GENERATOR_DEPENDENCIES}
```
orivej: `parse.py` does not depend on any generator dependencies except `${CMAKE_CURRENT_SOURCE_DIR}/generator/parse.py`.

sorig: You're right. Didn't read what was in the generator dependencies list properly. Thanks for feedback!

sorig:
The CI errors look unrelated

vigsterkr:
@sorig indeed... just restarted the ruby job, should run quickly, if it's all good i'll merge! :)
thnx both @sorig and @orivej to get this fixed!

vigsterkr:
great, all's good! merging!

Issue #3851, opened 2017-06-16
Implement a probability calibration scheme


karlnapf:
Shogun needs a way to turn calibrated probability outputs of its models (SVM, RF, etc), into calibrated ones. This means that the predicted class probability actually matches the actual class probability (as estimated over some validation dataset)

Inspiration:
http://scikit-learn.org/stable/modules/calibration.html

This requires some form of x-validation, so can be tied into Shogun's framework for that

Somehow related to #3850

karlnapf:
@durovo That would be a great addition to Shogun, such a welcome contribution!

It is a bit more tricky, make sure to share early drafts on this soon. Don't get into writing code before we have discussed a design draft. I suggest to read our corss-validation code before you start.

durovo:
@karlnapf Understood. I am currently going through the cross-validation code. I will share a rough draft as soon as I have figured it out. Should I use this thread to discuss the design?

karlnapf:
Yep this thread is totally fine. I would also be interested in hearing about the sklearn approach to see whether we can learn from it. I guess it would be best to re-use the x-validation code.

durovo:
@karlnapf I have gone through the sklearn approach and the shogun code related to this problem. Sklearn is using two methods for probability calibration:
1. Pratts scaling 
2. Isotonic regression
The `BinaryLabels` class in shogun already does probability calibration in the function `scores_to_probabilities`,  which uses `CStatistics::fit_sigmoid` for the same. 

The sklearn approach:

1. CalibratorCV has the same interface as that of estimators.
2. An estimator and a cross-validation strategy must be passed to the CalibratorCV.
3. The x-validation strategy is later used to split the dataset and labels when `fit` is called on them
4. For each fold, the estimator is trained on the training data. One of the two approaches given above is used to fit the output of the estimator on test data to the test labels. 
5. The calibrator object stores a copy of the trained estimator, this calibrator object is appended to a global list.
6. Later, when predict_probability is called on some data: 
  For each calibrator in the list:
    1. The estimator is used to predict the probabilities.
    2. These are then fed to the fitted calibration method to generate scaled probabilities.
7. The mean over all estimator-calibrator pairs is then returned.

karlnapf:
Any news on this?

durovo:
[...]

Issue #2841, opened 2015-05-06
Make: *** [all] Error 2


iarroyof:
Hi, I currently trying to install Shogun in Lubuntu. I followed this:

$ git clone git://github.com/shogun-toolbox/shogun.git
:~/shogun$ cd shogun
:~/shogun$ mkdir build
:~/shogun/build$ cmake -DPythonModular=ON -DCMAKE_INSTALL_PREFIX="$HOME/shogun-install" ..

However, when making I have an Error:

:~/shogun/build$ make -j2
[  1%] Generating version header
[  1%] Generating /home/iarroyof/shogun/src/shogun/base/class_list.cpp
[  1%] Built target version
[  1%] Built target class_list
Scanning dependencies of target libshogun
[  1%] Building CXX object src/shogun/CMakeFiles/libshogun.dir/base/class_list.cpp.o
In file included from /home/iarroyof/shogun/src/shogun/kernel/normalizer/KernelNormalizer.h:16:0,
                 from /home/iarroyof/shogun/src/shogun/transfer/multitask/MultitaskKernelMaskPairNormalizer.h:16,
                 from /home/iarroyof/shogun/src/shogun/base/class_list.cpp:44:
/home/iarroyof/shogun/src/shogun/kernel/Kernel.h: In member function ‘float64_t shogun::CKernel::kernel(int32_t, int32_t)’:
/home/iarroyof/shogun/src/shogun/kernel/Kernel.h:212:21: error: invalid use of incomplete type ‘class shogun::CKernelNormalizer’
    return normalizer->normalize(compute(idx_a, idx_b), idx_a, idx_b);
                     ^
In file included from /home/iarroyof/shogun/src/shogun/kernel/normalizer/KernelNormalizer.h:16:0,
                 from /home/iarroyof/shogun/src/shogun/transfer/multitask/MultitaskKernelMaskPairNormalizer.h:16,
                 from /home/iarroyof/shogun/src/shogun/base/class_list.cpp:44:
/home/iarroyof/shogun/src/shogun/kernel/Kernel.h:35:8: error: forward declaration of ‘class shogun::CKernelNormalizer’
  class CKernelNormalizer;
        ^
src/shogun/CMakeFiles/libshogun.dir/build.make:2525: recipe for target 'src/shogun/CMakeFiles/libshogun.dir/base/class_list.cpp.o' failed
[...]

PR #4009, opened 2017-12-09
Implement cross-validated calibration #3851

durovo:
Implement a cross-validated calibration scheme (see #3851 ). This is still a work in progress. I will be completing the following tasks after discussing the API:
1. Add more tests
2. Fix memory leaks
3. Documentation

EDIT:
Fixed the code to conform to the style guidelines.

### DIFF
--- a/src/shogun/evaluation/Calibration.cpp
+++ b/src/shogun/evaluation/Calibration.cpp
@@ -1,0 +1,282 @@
+/*
+ * Copyright (c) The Shogun Machine Learning Toolbox
+ * Written (w) 2012 - 2013 Heiko Strathmann
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ * this
+ *    list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE
+ * FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * The views and conclusions contained in the software and documentation are
+ * those
+ * of the authors and should not be interpreted as representing official
+ * policies,
+ * either expressed or implied, of the Shogun Development Team.
+ */
+
+#include <shogun/lib/config.h>
+
+#include <shogun/evaluation/Calibration.h>
+#include <shogun/evaluation/CalibrationMethod.h>
+#include <shogun/labels/LabelsFactory.h>
+#include <shogun/machine/Machine.h>
+#include <shogun/mathematics/Math.h>
+#include <shogun/mathematics/Statistics.h>
+#include <shogun/mathematics/linalg/LinalgNamespace.h>
+
+using namespace shogun;
+
+CCalibration::CCalibration() : CMachine()
+{
+	init();
+}
+
+CCalibration::~CCalibration()
+{
+	SG_UNREF(m_calibration_machines)
+	SG_UNREF(m_machine)
+	SG_UNREF(m_labels)
+	SG_UNREF(m_method)
+}
+
+void CCalibration::init()
+{
+	m_machine = new CMachine();
+	m_labels = new CBinaryLabels();
+	m_method = new CCalibrationMethod();
+	m_calibration_machines = new CDynamicObjectArray();
+
+	SG_ADD(
+	    (CSGObject**)&m_machine, "m_machine", "learning machine to use",
+	    MS_NOT_AVAILABLE);
+	SG_ADD(
+	    (CSGObject**)&m_labels, "m_labels", "target_labels", MS_NOT_AVAILABLE);
+	SG_ADD(
+	    (CSGObject**)&m_method, "m_method", "calibration method",
+	    MS_NOT_AVAILABLE);
+	SG_ADD(
+	    (CSGObject**)&m_calibration_machines, "m_calibration_machines",
+	    "array of calibration method machines", MS_NOT_AVAILABLE);
+}
+
+template <typename T>
+CBinaryLabels* CCalibration::get_binary_result(T data)
+{
+	CLabels* result = apply_once(data);
+	CBinaryLabels* result_labels = CLabelsFactory::to_binary(result);
+	CCalibrationMethod* method =
+	    (CCalibrationMethod*)m_calibration_machines->get_element(0);
+	SGVector<float64_t> confidence_values =
+	    method->apply_binary(result_labels->get_values());
+	result_labels->set_values(confidence_values);
+
+	return result_labels;
+}
+
+CBinaryLabels* CCalibration::apply_binary(CFeatures* features)
+{
+	return get_binary_result(features);
+}
+
+CBinaryLabels*
+CCalibration::apply_locked_binary(SGVector<index_t> subset_indices)
+{
+	return get_binary_result(subset_indices);
+}
+
+CMulticlassLabels* CCalibration::get_multiclass_result(
+    CMulticlassLabels* result_labels, index_t num_calibration_machines)
+{
+	for (index_t i = 0; i < num_calibration_machines; ++i)
+	{
+		CCalibrationMethod* method =
+		    (CCalibrationMethod*)m_calibration_machines->get_element(i);
+		SGVector<float64_t> confidence_values =
+		    method->apply_binary(result_labels->get_multiclass_confidences(i));
+		result_labels->set_multiclass_confidences(i, confidence_values);
+		SG_UNREF(method)
+	}
+
+	SGVector<float64_t> temp_confidences =
+	    result_labels->get_multiclass_confidences(0);
+	temp_confidences.zero();
+
+	index_t num_classes = result_labels->get_num_classes();
+
+	index_t num_samples = temp_confidences.vlen;
+
+// normalize the probabilities
+#pragma omp parallel for
+	for (index_t i = 0; i < num_classes; ++i)
+	{
+		SGVector<float64_t> confidence_values =
+		    result_labels->get_multiclass_confidences(i);
+		linalg::add(temp_confidences, confidence_values, temp_confidences, 1.0);
+	}
+#pragma omp parallel for
+	for (index_t i = 0; i < num_classes; ++i)
+	{
+		SGVector<float64_t> confidence_values =
+		    result_labels->get_multiclass_confidences(i);
+		for (index_t j = 0; j < num_samples; ++j)
+		{
+			confidence_values[j] /= temp_confidences[j];
+		}
+		result_labels->set_multiclass_confidences(i, confidence_values);
+	}
+
+	return result_labels;
+}
+
+CMulticlassLabels* CCalibration::apply_multiclass(CFeatures* features)
+{
+	index_t num_calibration_machines =
+	    m_calibration_machines->get_num_elements();
+	CLabels* result = m_machine->apply(features);
+	CMulticlassLabels* result_labels = CLabelsFactory::to_multiclass(result);
+	return get_multiclass_result(result_labels, num_calibration_machines);
+}
+
+CMulticlassLabels*
+CCalibration::apply_locked_multiclass(SGVector<index_t> subset_indices)
+{
+	index_t num_calibration_machines =
+	    m_calibration_machines->get_num_elements();
+	CLabels* result = m_machine->apply_locked(subset_indices);
+	CMulticlassLabels* result_labels = CLabelsFactory::to_multiclass(result);
+	return get_multiclass_result(result_labels, num_calibration_machines);
+}
+
+EProblemType CCalibration::get_machine_problem_type() const
+{
+	return m_machine->get_machine_problem_type();
+}
+
+bool CCalibration::train_one_machine(CFeatures* features)
+{
+	return m_machine->train(features);
+}
+
+bool CCalibration::train_one_machine(SGVector<index_t> subset_indices)
+{
+	return m_machine->train_locked(subset_indices);
+}
+
+CLabels* CCalibration::apply_once(CFeatures* features)
+{
+	return m_machine->apply(features);
+}
+
+CLabels* CCalibration::apply_once(SGVector<index_t> subset_indices)
+{
+	return m_machine->apply_locked(subset_indices);
+}
+
+template <typename T>
+bool CCalibration::train_calibration_machine(T training_data)
+{
+	CCalibrationMethod* calibration_machine = NULL;
+	if (get_machine_problem_type() == PT_MULTICLASS)
+	{
+		SGVector<float64_t> confidences;
+		index_t num_calibration_machines =
+		    (CLabelsFactory::to_multiclass(get_labels()))->get_num_classes();
+		m_calibration_machines =
+		    new CDynamicObjectArray(num_calibration_machines);
+		train_one_machine(training_data);
+		CLabels* result = apply_once(training_data);
+		CMulticlassLabels* result_labels =
+		    CLabelsFactory::to_multiclass(result);
+
+		for (index_t i = 0; i < num_calibration_machines; ++i)
+		{
+			confidences = result_labels->get_multiclass_confidences(i);
+
+			calibration_machine = (CCalibrationMethod*)m_method->clone();
+			if (!calibration_machine->train(confidences))
+			{
+				return false;
+			}
+			m_calibration_machines->set_element(calibration_machine, i);
+			SG_UNREF(calibration_machine)
+		}
+		SG_UNREF(result_labels)
+	}
+	else
+	{
+		SGVector<float64_t> confidences;
+		m_calibration_machines = new CDynamicObjectArray(1);
+		train_one_machine(training_data);
+		CLabels* result = apply_once(training_data);
+		CBinaryLabels* result_labels = CLabelsFactory::to_binary(result);
+
+		confidences = result_labels->get_values();
+
+		SG_UNREF(result_labels)
+
+		calibration_machine = (CCalibrationMethod*)m_method->clone();
+		if (!calibration_machine->train(confidences))
+		{
+			return false;
+		}
+		m_calibration_machines->set_element(calibration_machine, 0);
+	}
+
+	return true;
+}
+
+bool CCalibration::train(CFeatures* features)
+{
+	return train_calibration_machine(features);
+}
+
+bool CCalibration::train_locked(SGVector<index_t> subset_indices)
+{
+	return train_calibration_machine(subset_indices);
+}
+
+void CCalibration::set_calibration_method(CCalibrationMethod* method)
+{
+	SG_UNREF(m_method);
+	m_method = method;
+	SG_REF(m_method);
+}
+
+void CCalibration::set_machine(CMachine* machine)
+{
+	SG_UNREF(m_machine);
+	m_machine = machine;
+	SG_REF(m_machine);
+}
+
+CMachine* CCalibration::get_machine()
+{
+	SG_REF(m_machine);
+	return m_machine;
+}
+
+CCalibrationMethod* CCalibration::get_calibration_method()
+{
+	SG_REF(m_method);
+	return m_method;
+}
--- a/src/shogun/evaluation/Calibration.h
+++ b/src/shogun/evaluation/Calibration.h
@@ -1,0 +1,113 @@
+/*
+ * Copyright (c) The Shogun Machine Learning Toolbox
+ * Written (w) 2012 - 2013 Heiko Strathmann
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ * this
+ *    list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE
+ * FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * The views and conclusions contained in the software and documentation are
+ * those
+ * of the authors and should not be interpreted as representing official
+ * policies,
+ * either expressed or implied, of the Shogun Development Team.
+ */
+
+#ifndef _CALIBRATION_H__
+#define _CALIBRATION_H__
+
+#include <shogun/lib/config.h>
+
+#include <shogun/evaluation/CalibrationMethod.h>
+#include <shogun/machine/Machine.h>
+
+namespace shogun
+{
+
+	class CCalibration : public CMachine
+	{
+	public:
+		/** constructor
+		 */
+		CCalibration();
+
+		virtual ~CCalibration();
+
+		virtual const char* get_name() const
+		{
+			return "Calibration";
+		}
+
+		virtual EProblemType get_machine_problem_type() const;
+
+		virtual bool train(CFeatures* data = NULL);
+
+		virtual bool train_locked(SGVector<index_t> subset_indices);
+
+		virtual CBinaryLabels* apply_binary(CFeatures* features);
+
+		virtual CMulticlassLabels* get_multiclass_result(
+		    CMulticlassLabels* result_labels, index_t num_calibration_machines);
+
+		virtual CMulticlassLabels* apply_multiclass(CFeatures* features);
+
+		virtual CMulticlassLabels*
+		apply_locked_multiclass(SGVector<index_t> subset_indices);
+
+		virtual CBinaryLabels*
+		apply_locked_binary(SGVector<index_t> subset_indices);
+
+		virtual CMachine* get_machine();
+
+		virtual void set_machine(CMachine* machine);
+
+		virtual void
+		set_calibration_method(CCalibrationMethod* calibration_method);
+
+		virtual CCalibrationMethod* get_calibration_method();
+
+	private:
+		CLabels* apply_once(CFeatures* features);
+
+		CLabels* apply_once(SGVector<index_t> subset_indices);
+
+		template <typename T>
+		bool train_calibration_machine(T training_data);
+
+		template <typename T>
+		CBinaryLabels* get_binary_result(T data);
+
+		void init();
+
+		bool train_one_machine(SGVector<index_t> subset_indices);
+
+		bool train_one_machine(CFeatures* features);
+
+	private:
+		CMachine* m_machine;
+		CDynamicObjectArray* m_calibration_machines;
+		CCalibrationMethod* m_method;
+	};
+}
+#endif
--- a/src/shogun/evaluation/CalibrationMethod.h
+++ b/src/shogun/evaluation/CalibrationMethod.h
@@ -1,0 +1,82 @@
+/*
+ * Copyright (c) The Shogun Machine Learning Toolbox
+ * Written (w) 2012 - 2013 Heiko Strathmann
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ * this
+ *    list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE
+ * FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * The views and conclusions contained in the software and documentation are
+ * those
+ * of the authors and should not be interpreted as representing official
+ * policies,
+ * either expressed or implied, of the Shogun Development Team.
+ */
+
+#ifndef _CALIBRATION_METHOD_H__
+#define _CALIBRATION_METHOD_H__
+
+#include <shogun/lib/config.h>
+
+#include <shogun/machine/Machine.h>
+
+namespace shogun
+{
+
+	class CCalibrationMethod : public CMachine
+	{
+	public:
+		CCalibrationMethod();
+
+		/** constructor
+		 * @param target_values, true value of each label
+		 */
+		CCalibrationMethod(SGVector<float64_t> target_values);
+
+		virtual ~CCalibrationMethod();
+
+		virtual const char* get_name() const
+		{
+			return "CalibrationMethod";
+		}
+
+		virtual EProblemType get_machine_problem_type() const
+		{
+			return PT_BINARY;
+		}
+
+		virtual bool train(SGVector<float64_t> values);
+
+		virtual SGVector<float64_t> apply_binary(SGVector<float64_t> values);
+
+		virtual void set_target_values(SGVector<float64_t> target_values);
+
+	private:
+		void init();
+
+	private:
+		SGVector<float64_t> m_target_values;
+	};
+}
+#endif
--- a/src/shogun/evaluation/CrossValidatedCalibration.cpp
+++ b/src/shogun/evaluation/CrossValidatedCalibration.cpp
@@ -1,0 +1,399 @@
+/*
+ * Copyright (c) The Shogun Machine Learning Toolbox
+ * Written (w) 2012 - 2013 Heiko Strathmann
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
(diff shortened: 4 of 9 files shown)


karlnapf:
This is great! Thanks!
I will review this in detail tomorrow

src/shogun/evaluation/Calibration.h
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * Written (W) 2011-2012 Heiko Strathmann
+ * Copyright (C) 2011 Berlin Institute of Technology and Max-Planck-Society
+ */
+
+#include <shogun/evaluation/CalibrationMethod.h>
+#include <shogun/lib/config.h>
+#include <shogun/machine/Machine.h>
+
+#ifndef _CALIBRATION_H__
```
vigsterkr: move this to the beginning of the file with the define as well.... this way the above includes will be included only once even if `shogun/evaluation/Calibration.h` included more than once

src/shogun/evaluation/Calibration.cpp
```diff
@@ -0,0 +1,244 @@
...
+#include <shogun/machine/Machine.h>
+#include <shogun/mathematics/Math.h>
+#include <shogun/mathematics/Statistics.h>
+
+using namespace shogun;
+
+CBinaryLabels* CCalibration::apply_binary(CFeatures* features)
+{
+
+	CBinaryLabels* result_labels = (CBinaryLabels*)m_machine->apply(features);
```
vigsterkr: this explicit casting will cause a lot of trouble if the returned label is not binary of the machine :) you should either do a dynamic_cast to check for this or simply `get_label_type() == LT_BINARY`

src/shogun/evaluation/Calibration.cpp
```diff
@@ -0,0 +1,244 @@
...
+	result_labels->set_values(confidence_values);
+
+	return result_labels;
+}
+
+CMulticlassLabels* CCalibration::apply_multiclass(CFeatures* features)
+{
+	index_t num_calibration_machines =
+	    m_calibration_machines->get_num_elements();
+	CMulticlassLabels* result_labels =
```
vigsterkr: same as above... do a safe typecheck

src/shogun/evaluation/Calibration.cpp
```diff
@@ -0,0 +1,244 @@
...
+			}
+		}
+	}
+	for (index_t i = 0; i < num_classes; ++i)
+	{
+		SGVector<float64_t> confidence_values =
+		    result_labels->get_multiclass_confidences(i);
+		for (index_t j = 0; j < num_samples; ++j)
+		{
+			confidence_values[j] = confidence_values[j] / temp_confidences[j];
```
vigsterkr: this could be simply:
```
confidence_values[j] /= temp_confidences[j];
```

src/shogun/evaluation/Calibration.cpp
```diff
@@ -0,0 +1,244 @@
...
+
+	return result_labels;
+}
+
+CMulticlassLabels*
+CCalibration::apply_locked_multiclass(SGVector<index_t> subset_indices)
+{
+	index_t num_calibration_machines =
+	    m_calibration_machines->get_num_elements();
+	CMulticlassLabels* result_labels =
```
vigsterkr: ensure type

src/shogun/evaluation/Calibration.cpp
```diff
@@ -0,0 +1,244 @@
...
+			confidence_values[j] = confidence_values[j] / temp_confidences[j];
+		}
+		result_labels->set_multiclass_confidences(i, confidence_values);
+	}
+
+	return result_labels;
+}
+
+CMulticlassLabels*
+CCalibration::apply_locked_multiclass(SGVector<index_t> subset_indices)
```
vigsterkr: this method is almost the copy-paste of `CCalibration::apply_multiclass`. you can basically add the same parts to a common function and use that

src/shogun/evaluation/Calibration.cpp
```diff
@@ -0,0 +1,244 @@
...
+}
+
+void CCalibration::set_machine(CMachine* machine)
+{
+	m_machine = machine;
+}
+
+void CCalibration::init()
+{
+	m_machine = NULL;
```
vigsterkr: nullptr

src/shogun/evaluation/Calibration.cpp
```diff
@@ -0,0 +1,244 @@
...
+	m_labels = NULL;
```
vigsterkr: nullptr

src/shogun/evaluation/Calibration.cpp
```diff
@@ -0,0 +1,244 @@
...
+	m_method = method;
+}
+
+void CCalibration::set_machine(CMachine* machine)
+{
+	m_machine = machine;
+}
+
+void CCalibration::init()
+{
```
vigsterkr: you are missing the `SG_ADD` macros for the class parameters. check other objects for examples

src/shogun/evaluation/Calibration.h
```diff
@@ -0,0 +1,66 @@
...
+
+		virtual CMulticlassLabels* apply_multiclass(CFeatures* features);
+
+		virtual CMulticlassLabels*
+		apply_locked_multiclass(SGVector<index_t> subset_indices);
+
+		virtual CBinaryLabels*
+		apply_locked_binary(SGVector<index_t> subset_indices);
+
+		/** constructor
```
vigsterkr: move to the top of public

src/shogun/evaluation/Calibration.h
```diff
@@ -0,0 +1,66 @@
...
+		 */
+		CCalibration();
+
+		~CCalibration();
```
vigsterkr: its a virtual right?

src/shogun/evaluation/CalibrationMethod.h
```diff
@@ -0,0 +1,52 @@
...
+ * (at your option) any later version.
+ *
+ * Written (W) 2011-2012 Heiko Strathmann
+ * Copyright (C) 2011 Berlin Institute of Technology and Max-Planck-Society
+ */
+
+#include <shogun/lib/config.h>
+#include <shogun/machine/Machine.h>
+
+#ifndef _CALIBRATION_METHOD_H__
```
vigsterkr: move this to the beginning of the file with the define as well....

src/shogun/evaluation/CrossValidatedCalibration.h
```diff
@@ -0,0 +1,77 @@
...
+ * Written (W) 2011-2012 Heiko Strathmann
+ * Copyright (C) 2011 Berlin Institute of Technology and Max-Planck-Society
+ */
+
+#include <shogun/evaluation/Calibration.h>
+#include <shogun/evaluation/SplittingStrategy.h>
+#include <shogun/lib/config.h>
+#include <shogun/machine/Machine.h>
+
+#ifndef _CROSS_VALIDATED_CALIBRATION_H__
```
vigsterkr: move this to the beginning of the file with the define as well....

src/shogun/evaluation/CrossValidatedCalibration.cpp
```diff
@@ -0,0 +1,406 @@
...
+			result_labels = temp_result;
+		}
+		else
+		{
+			result_values += temp_result->get_values();
+		}
+	}
+
+#pragma omp parallel for
+	for (index_t i = 0; i < result_values.vlen; ++i)
```
vigsterkr: this could be substituted by a simple `linalg::scale`
has this code been borrowed from somewhere?

src/shogun/evaluation/CrossValidatedCalibration.cpp
```diff
@@ -0,0 +1,406 @@
...
+CBinaryLabels* CCrossValidatedCalibration::apply_binary(CFeatures* features)
+{
+	index_t num_machines = m_calibration_machines->get_num_elements();
+
+	CBinaryLabels* temp_result;
+	CMachine* temp_machine;
+	SGVector<float64_t> result_values;
+	CBinaryLabels* result_labels;
+
+	for (index_t i = 0; i < num_machines; ++i)
```
vigsterkr: this could be done in parallel....

src/shogun/evaluation/CrossValidatedCalibration.cpp
```diff
@@ -0,0 +1,406 @@
...
+			result_labels = temp_result;
+		}
+		else
+		{
+			result_values += temp_result->get_values();
+		}
+	}
+
+#pragma omp parallel for
+	for (index_t i = 0; i < result_values.vlen; ++i)
```
vigsterkr: linalg::scale

src/shogun/evaluation/CrossValidatedCalibration.cpp
```diff
@@ -0,0 +1,406 @@
...
+	{
+		result_values[i] = result_values[i] / num_machines;
+	}
+
+	result_labels->set_values(result_values);
+
+	return result_labels;
+}
+
+CBinaryLabels* CCrossValidatedCalibration::apply_locked_binary(
```
vigsterkr: almost copy-paste of `CCrossValidatedCalibration::apply_binary`. plz simplify

src/shogun/evaluation/CrossValidatedCalibration.cpp
```diff
@@ -0,0 +1,406 @@
...
+			}
+		}
+	}
+
+	for (index_t i = 0; i < num_classes; ++i)
+	{
+		temp_values = result_labels->get_multiclass_confidences(i);
+
+#pragma omp parallel for
+		for (index_t j = 0; j < temp_values.vlen; j++)
```
vigsterkr: use `linalg::scale`

src/shogun/evaluation/CrossValidatedCalibration.cpp
```diff
@@ -0,0 +1,406 @@
...
+			{
+				result_values = temp_result->get_multiclass_confidences(j);
+				temp_values = result_labels->get_multiclass_confidences(j);
+				temp_values += result_values;
+				result_labels->set_multiclass_confidences(j, temp_values);
+			}
+		}
+	}
+
+	for (index_t i = 0; i < num_classes; ++i)
```
vigsterkr: you can omp this one

src/shogun/evaluation/CrossValidatedCalibration.cpp
```diff
@@ -0,0 +1,406 @@
...
+	index_t num_machines = m_calibration_machines->get_num_elements();
+
+	index_t num_classes = ((CMulticlassLabels*)m_labels)->get_num_classes();
+
+	CMulticlassLabels* temp_result;
+	CMachine* temp_machine;
+	SGVector<float64_t> result_values, temp_values;
+	CMulticlassLabels* result_labels;
+
+	for (index_t i = 0; i < num_machines; ++i)
```
vigsterkr: omp?

src/shogun/evaluation/CrossValidatedCalibration.cpp
```diff
@@ -0,0 +1,406 @@
...
+		{
+			temp_values[j] = temp_values[j] / num_machines;
+		}
+		result_labels->set_multiclass_confidences(i, temp_values);
+	}
+
+	return result_labels;
+}
+
+CMulticlassLabels* CCrossValidatedCalibration::apply_locked_multiclass(
```
vigsterkr: almost the same as `CCrossValidatedCalibration::apply_multiclass` so plz define a template or a function where you pas the apply method

src/shogun/evaluation/CrossValidatedCalibration.cpp
```diff
@@ -0,0 +1,406 @@
...
+			{
+				result_values = temp_result->get_multiclass_confidences(j);
+				temp_values = result_labels->get_multiclass_confidences(j);
+				temp_values += result_values;
+				result_labels->set_multiclass_confidences(j, temp_values);
+			}
+		}
+	}
+
+	for (index_t i = 0; i < num_classes; ++i)
```
vigsterkr: same as above.. omp?

src/shogun/evaluation/CrossValidatedCalibration.cpp
```diff
@@ -0,0 +1,406 @@
...
+	{
+		temp_values = result_labels->get_multiclass_confidences(i);
+
+#pragma omp parallel for
+		for (index_t j = 0; j < temp_values.vlen; j++)
```
vigsterkr: linalg::scale

src/shogun/evaluation/CrossValidatedCalibration.cpp
```diff
@@ -0,0 +1,406 @@
...
+{
+	init();
+
+	m_machine = machine;
+	m_labels = labels;
+	m_splitting_strategy = splitting_strategy;
+	m_calibration_method = calibration_method;
+}
+
+void CCrossValidatedCalibration::init()
```
vigsterkr: missing SG_ADD macros

src/shogun/evaluation/SigmoidCalibrationMethod.h
```diff
@@ -0,0 +1,58 @@
...
+		 * @param autolock autolock
+		 */
+		CSigmoidCalibrationMethod();
+
+		CSigmoidCalibrationMethod(SGVector<float64_t> target_values);
+
+		~CSigmoidCalibrationMethod();
+
+	private:
+		float64_t a, b;
```
vigsterkr: m_a and m_b

src/shogun/evaluation/SigmoidCalibrationMethod.cpp
```diff
@@ -0,0 +1,49 @@
...
+{
+	CStatistics::SigmoidParamters params = CStatistics::fit_sigmoid(values);
+	a = params.a;
+	b = params.b;
+
+	return true;
+}
+
+CSigmoidCalibrationMethod::CSigmoidCalibrationMethod() : CCalibrationMethod()
+{
```
vigsterkr: need to init a and b variables and SG_ADD them

durovo:
Thanks a lot for the detailed review. I'll make all the required changes and make a pr. From my understanding, in general, I have to make the following rectifications:
1. The ifdef should contain the whole header file.
2. I can parallelize using omp wherever possible.
3. Safe typecasting.
4. Push repeated code to a common function.
5. SG_ADD everywhere

src/shogun/evaluation/CrossValidatedCalibration.cpp
durovo: I don't think that I borrowed this from somewhere. I'm not sure though..

src/shogun/evaluation/SigmoidCalibrationMethod.cpp
durovo: Should they be initialized to 0?

src/shogun/evaluation/Calibration.cpp
durovo: Should I use SG_ERROR to indicate that the machine should return binary labels when the labels are not binary?

vigsterkr: you could use REQUIRE

src/shogun/evaluation/CrossValidatedCalibration.cpp
vigsterkr: ok so then just change it to linalg::scale
the reason i've asked is because if it borrowed then we should change it there as well :)

src/shogun/evaluation/SigmoidCalibrationMethod.cpp
vigsterkr: yep that's a good value as initial value

vigsterkr:
@durovo yes more or less that's the idea. as soon as you fix those the travis builds should be green. the reason they fail now is because you miss SG_ADD

karlnapf:
@vigsterkr was faster with reviewing :)

One comment: Let's maybe first make it work without parallel omp, as that might complicate things quite a bit. We can work on that once the other comments around duplicate code and style are resolved.

We could then solve parallel xvalidation and parallel calibration in the same go.

Also, what I would love to see is *using* the x-validation code in shogun to calibrate, rather than *copying* it (@vigsterkr mentioned this in his comments).

Great first start, thanks for the effort. Let's get this going, make sure to ask in IRC if you have questions or get stuck. Looking forward to the update

src/shogun/evaluation/Calibration.cpp
```diff
@@ -0,0 +1,244 @@
+/*
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 3 of the License, or
+ * (at your option) any later version.
```
karlnapf: pls use a BSD header, see e.g. https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/statistical_testing/BTestMMD.h

karlnapf: (in all the files)

src/shogun/evaluation/Calibration.cpp
```diff
@@ -0,0 +1,244 @@
...
+		    method->apply_binary(result_labels->get_multiclass_confidences(i));
+		result_labels->set_multiclass_confidences(i, confidence_values);
+	}
+
+	SGVector<float64_t> temp_confidences;
+	index_t num_classes = result_labels->get_num_classes();
+	index_t num_samples;
+
+	// normalize the probabilities
+	for (index_t i = 0; i < num_classes; ++i)
```
karlnapf: all of this can be done in a vectorised fashion using linalg, see e.g. https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/preprocessor/KernelPCA.cpp#L92

src/shogun/evaluation/Calibration.cpp
karlnapf: I am also confused, why is this necessary?

tests/unit/evaluation/CrossValidatedCalibration_unittest.cc
```diff
@@ -0,0 +1,65 @@
...
+#include <shogun/features/DataGenerator.h>
+#include <shogun/features/DenseFeatures.h>
+#include <shogun/labels/MulticlassLabels.h>
+#include <shogun/multiclass/MulticlassOCAS.h>
+
+using namespace shogun;
+
+extern MultiLabelTestEnvironment* multilabel_test_env;
+
+#ifdef HAVE_LAPACK
```
karlnapf: what is LAPACK needed for here?

tests/unit/evaluation/CrossValidatedCalibration_unittest.cc
```diff
@@ -0,0 +1,65 @@
...
+	CMulticlassLabels* pred = (CMulticlassLabels*)cross->apply(test_feats);
+
+	index_t num_classes = pred->get_num_classes();
+
+	SGVector<float64_t> confidence_sums;
+	for (index_t i = 0; i < num_classes; ++i)
+	{
+		SGVector<float64_t> scores = pred->get_multiclass_confidences(i);
+		if (i == 0)
+			confidence_sums = scores;
```
karlnapf: rather than writing unit tests here, a good sanity check would be to reproduce the plots in http://scikit-learn.org/stable/_images/sphx_glr_plot_calibration_curve_002.png
You could do that via say Python bindings. Once those work, then we can start adding more unit tests ...
Let me know what you think

durovo: That looks like a great idea. I will try doing that.

src/shogun/evaluation/Calibration.cpp
durovo: Yes, I'm currently trying to vectorize wherever possible.

src/shogun/evaluation/Calibration.cpp
durovo: Understood

src/shogun/evaluation/Calibration.cpp
durovo: I have now used a template function to hold the common code. I'll make a pr shortly

src/shogun/evaluation/Calibration.cpp
durovo: Should they not be initialized to NULL?
I was following what was done here:
https://github.com/shogun-toolbox/shogun/blob/15fdbcd7520191d27ee9ad69f196da0023c841e2/src/shogun/evaluation/MachineEvaluation.cpp#L81

vigsterkr: http://en.cppreference.com/w/cpp/language/nullptr

vigsterkr: yes we do have the oldschool `NULL`, but since this is new code, it'd be good to follow higher std standard features

tests/unit/evaluation/CrossValidatedCalibration_unittest.cc
durovo: The python interfaces for these files are generated automatically right? Or do I have to add these files to some list? I am asking because I am not able to access these (CrossValidatedCalibration, Calibration, etc) from the shogun module in Python.

vigsterkr: @durovo yeah as of this state the python interface will not contain these methods.. you'll need to add it to some definition files
but that should be another PR. for the time being you should write unit tests for this and test it from the C++ API

durovo: Okay

src/shogun/evaluation/Calibration.cpp
durovo: Hello @vigsterkr, I have used the approach that you have mentioned (using `get_label_type` and `REQUIRE`). However, shogun seems to have a LabelsFactory class which has methods for doing the same thing. Is that the recommended way of doing this?

vigsterkr: mmm oh yeah labelfactory, indeed its the best option

tests/unit/evaluation/CrossValidatedCalibration_unittest.cc
durovo: I borrowed this from another test. I will write proper test cases after we discuss the api. Or should I write them now?

durovo:
I have tried to make all the suggested changes. However, I have not added omp everywhere which was suggested by @vigsterkr. As @karlnapf suggested, I will add it after I know that the code works correctly without it. I could add it now if required..

vigsterkr:
@durovo its fine to do the openmp part later... the point was rather that you should do it in other part of the code as well as use linalg:: when possible.

src/shogun/evaluation/SigmoidCalibrationMethod.h
```diff
@@ -0,0 +1,79 @@
...
+#include <shogun/lib/config.h>
+
+#include <shogun/evaluation/CalibrationMethod.h>
+#include <shogun/machine/Machine.h>
+#include <shogun/mathematics/Math.h>
+
+namespace shogun
+{
+
+	class CSigmoidCalibrationMethod : public CCalibrationMethod
```
vigsterkr: doxygen. please make sure at least that there's a brief description of the method itself

src/shogun/evaluation/CrossValidatedCalibration.cpp
```diff
@@ -0,0 +1,399 @@
...
+	CMachine* temp_machine;
+	SGVector<float64_t> result_values, temp_values;
+	CMulticlassLabels* result_labels;
+	CLabels* result;
+
+	temp_machine = (CMachine*)m_calibration_machines->get_element(0);
+	result = apply_once(temp_machine, data);
+	result_labels = CLabelsFactory::to_multiclass(result);
+
+	for (index_t i = 1; i < num_machines; ++i)
```
vigsterkr: why from 1 and not 0?

tests/unit/evaluation/CrossValidatedCalibration_unittest.cc
```diff
@@ -0,0 +1,65 @@
...
+	cross->train(train_feats);
+
+	CMulticlassLabels* pred = (CMulticlassLabels*)cross->apply(test_feats);
+
+	index_t num_classes = pred->get_num_classes();
+
+	SGVector<float64_t> confidence_sums;
+	confidence_sums = pred->get_multiclass_confidences(0);
+
+	for (index_t i = 1; i < num_classes; ++i)
```
vigsterkr: same as above... why index from 1?

vigsterkr:
@durovo make sure that the next time you submit that the unit tests are passing, see:
```	 53 - unit-CrossValidatedCalibrationTest (SEGFAULT)```

on travis CI

**Comment by durovo** [comment]
@vigsterkr  strange.. I'm sure that the unit tests are passing on my computer.

File: src/shogun/evaluation/SigmoidCalibrationMethod.h (hunk 31 above)
Comment (durovo) [reply src/shogun/evaluation/SigmoidCalibrationMethod.h]: Understood, I'll add the descriptions for all methods (and params) and classes.

File: src/shogun/evaluation/CrossValidatedCalibration.cpp (hunk 32 above)
Comment (durovo) [reply src/shogun/evaluation/CrossValidatedCalibration.cpp]: The result from the first machine has already retrieved just before the loop. This is a dirty hack that I used because I couldn't find any method to get the number of samples in the features.

File: tests/unit/evaluation/CrossValidatedCalibration_unittest.cc [hunk 34]
```diff
@@ -0,0 +1,65 @@
...
+	{
+		SGVector<float64_t> scores = pred->get_multiclass_confidences(i);
```
karlnapf: I suggest you use `setUp` to set up your cross estimator, and then have multiple tests that test different characteristics.

* just calling pred (without asserting anything)
* the sum probability thing you did
* other things that are meaningful

tests/unit/evaluation/CrossValidatedCalibration_unittest.cc
karlnapf: also we can `auto` that

src/shogun/evaluation/CrossValidatedCalibration.cpp
karlnapf: I would still start the loop at 0 for readability

durovo: Understood. However, what is the standard way of getting the number of samples from a given set of feature? (i.e. how do I know how many labels will be there in the predictions before actually calling apply?)

tests/unit/evaluation/CrossValidatedCalibration_unittest.cc
durovo: Okay.

src/shogun/evaluation/CrossValidatedCalibration.cpp
karlnapf: You can still get the first element to do that.
And then you should ensure that all of the features have the same number of elements

karlnapf: Wait, isnt that just `get_num_vectors' ?

karlnapf:
Any updates here?
I am keen to help if something is unclear ...

durovo:
@karlnapf I'm done with the descriptions. I have almost figured out the tests too. However, I had to shift to a new pc and now I'm facing this issue: https://github.com/shogun-toolbox/shogun/issues/2841
I will get back here with the updates on this soon .

karlnapf:
Mmmmh weird. I don't understand this error and never managed to re-produce it locally. Let us know how you get on with it, and if you need help. You can ping me in irc anytime

durovo:
@karlnapf I think the problem is related to some package installed using conda. For now, I'm using a vm for development. I will try experimenting a bit later to figure out the exact reason behind the problem. 
I was writing the tests for calibration and I wanted to know that how do we calculate the log loss in shogun? The log loss for the calibrated predictions should be less than the uncalibrated predictions. There is a LogLoss implementation in the loss folder, however I could not find it being used anywhere.

karlnapf:
If you have a conda version of shogun installed you need to make sure it is not used in the compilation process ....

durovo:
I don't have the conda version of shogun installed. The reason behind the problem seems to be something else.
I was talking about this log loss:
http://www.exegetic.biz/blog/2015/12/making-sense-logarithmic-loss/

From what I understand, logarithmic loss should decrease after calibration since in a way we are optimizing this metric.

durovo:
Maybe I misunderstood you.. are you asking me to define the log loss function myself? If that is what you were asking, then I yes certainly. Should I make a new class for it or define it as a throwaway function within the test?

karlnapf:
I see, we will need to implement that under `CEvaluation` don't think Shogun has it. Separate PR :)

karlnapf:
I think for testing this, some simpler tests would be a better start. A really simple case with a Mock model that outputs a fixed class probability, and then see whether the calibration works on that one.

I guess before writing tests, you could even first try to reproduce the plots for the sklearn calibration example (or something similar).

karlnapf:
As usual, let me know if you need more input on that

durovo:
Understood. Log loss could come later I guess :). I will try some sanity checks just as you suggested and then write the tests.

durovo:
@karlnapf `fit_sigmoid` method in Statistics.cpp (https://github.com/shogun-toolbox/shogun/blob/c87d5844628dae56ed8b47ed287d0f84e1d3037f/src/shogun/mathematics/Statistics.cpp#L824) does not use the target labels to set the values for the vector `t` (line 863). In the following paper, the target labels are used to set the values of `t` (see page 10).
 https://www.csie.ntu.edu.tw/~cjlin/papers/plattprob.pdf
Is this intentional? Wouldn't this produce somewhat different results?
Also, does the `splitting strategy` randomize the subsets?
I managed to set up the python bindings and was trying to compare this implementation's results with sklearn's.

durovo:
image
This is the calibration plot that I am getting. The score in the legend is the brier score loss.

durovo:
@karlnapf the curve improves when I modify the fit_sigmoid function to use the true labels. Should I change that too?

karlnapf:
Hi!
Sorry for the long delay. This is great that you made the plots!
I have to run now, but will review tomorrow at latest

durovo:
Understood :). Thanks a lot.

karlnapf:
Re the sigmoid fit, are you talking about the line `if (label[i] > 0)`?
And in shogun, this is `scores[i]>0`?
If so, think about the decision function for SVMs, which is taking the sign of the scores.
However, it *actually* is `f(x)=\sign ( \sum_i \alpha_i k(x,x_i)    +b ). So our implementation misses the bias term. BAD. I will open an issue to investigate this, let's address this is another PR.

karlnapf:
Question: Why does the calibrated one in Shogun appears to not be calibrated in the plot ? :)

src/shogun/evaluation/Calibration.cpp
```diff
@@ -0,0 +1,282 @@
+/*
+ * Copyright (c) The Shogun Machine Learning Toolbox
+ * Written (w) 2012 - 2013 Heiko Strathmann
```
karlnapf: you might want to add you own name here :)

src/shogun/evaluation/Calibration.h
```diff
@@ -0,0 +1,113 @@
+/*
+ * Copyright (c) The Shogun Machine Learning Toolbox
+ * Written (w) 2012 - 2013 Heiko Strathmann
```
karlnapf: name

src/shogun/evaluation/Calibration.h
```diff
@@ -0,0 +1,113 @@
...
+
+#include <shogun/lib/config.h>
+
+#include <shogun/evaluation/CalibrationMethod.h>
+#include <shogun/machine/Machine.h>
+
+namespace shogun
+{
+
+	class CCalibration : public CMachine
```
karlnapf: Is Calibration the best name? Could be ambiguous... CalibratedClassifier?

karlnapf: This class needs a `@brief` description to tell users what it is doing
 * it is a meta class that wraps a classifier
 * it performs x-validation to...
 * ... calibrate the probability outputs
 * this is expensive
 * what method is used to calibrate
 * etc

src/shogun/evaluation/CalibrationMethod.h
```diff
@@ -0,0 +1,82 @@
...
+#define _CALIBRATION_METHOD_H__
+
+#include <shogun/lib/config.h>
+
+#include <shogun/machine/Machine.h>
+
+namespace shogun
+{
+
+	class CCalibrationMethod : public CMachine
```
karlnapf: Description!

src/shogun/evaluation/CalibrationMethod.h
```diff
@@ -0,0 +1,82 @@
...
+	{
+	public:
+		CCalibrationMethod();
+
+		/** constructor
```
karlnapf: For all comments: start English sentences with an upper case, and end them with a period i.e. "Constructor."

src/shogun/evaluation/CalibrationMethod.h
karlnapf: I find it a bit weird that both `CCalibration` and `CCalibrationMethod` both inherit from `CMachine`

src/shogun/evaluation/CalibrationMethod.h
```diff
@@ -0,0 +1,82 @@
...
+		virtual ~CCalibrationMethod();
+
+		virtual const char* get_name() const
+		{
+			return "CalibrationMethod";
+		}
+
+		virtual EProblemType get_machine_problem_type() const
+		{
+			return PT_BINARY;
```
karlnapf: this should be documented in the class docs, i.e. this is for binary problems
If that is the case, you should probably also change the name `BinaryCalibration` or something

src/shogun/evaluation/CrossValidatedCalibration.h
```diff
@@ -0,0 +1,115 @@
...
+#define _CROSS_VALIDATED_CALIBRATION_H__
+
+namespace shogun
+{
+
+	class CCrossValidatedCalibration : public CMachine
```
karlnapf: another `CMachine` inherited class.... not a fan of that

karlnapf: In my opinon, this is the only machine class you need. This one is the wrapper for CMachine and this one takes a machine and hides it behind another CMachine interface, wraps the calls, does x.validation etc

src/shogun/evaluation/SigmoidCalibrationMethod.h
karlnapf: this is still needed

src/shogun/evaluation/SigmoidCalibrationMethod.h
```diff
@@ -0,0 +1,79 @@
...
+
+		virtual bool train(SGVector<float64_t> values);
+
+		virtual SGVector<float64_t> apply_binary(SGVector<float64_t> values);
+
+	private:
+		void init();
+
+	private:
+		float64_t m_a, m_b;
```
karlnapf: I think we have something like `SigmoidParameters`

tests/unit/evaluation/CrossValidatedCalibration_unittest.cc
karlnapf: After API is frozen

tests/unit/evaluation/CrossValidatedCalibration_unittest.cc
karlnapf: This still needs to be done

src/shogun/evaluation/CrossValidatedCalibration.cpp
```diff
@@ -0,0 +1,399 @@
...
+	m_splitting_strategy->build_subsets();
+
+	if (!m_machine->is_data_locked())
+	{
+		SG_ERROR("please lock the data before running train_locked")
+		return false;
+	}
+
+	SG_DEBUG("starting unlocked calibration\n", get_name())
+	/* tell machine to store model internally
```
karlnapf: pls dont copy code comments around

src/shogun/evaluation/CrossValidatedCalibration.cpp
```diff
@@ -0,0 +1,399 @@
...
+		/* feature subset for calibration */
+		SGVector<index_t> subset_indices =
+		    m_splitting_strategy->generate_subset_indices(i);
+
+		/* set subset for testing labels */
+		m_labels->add_subset(subset_indices);
+
+		/* produce output for desired indices */
+		CCalibration* calibrator = new CCalibration();
+		calibrator->set_machine((CMachine*)m_machine->clone());
```
karlnapf: this replicates the data as well .... slow

src/shogun/evaluation/CrossValidatedCalibration.cpp
```diff
@@ -0,0 +1,399 @@
...
+	 * (otherwise changing subset of features will kaboom the classifier) */
+	m_machine->set_store_model_features(true);
+
+	for (index_t i = 0; i < num_subsets; ++i)
+	{
+		CMachine* machine;
+		CFeatures* features;
+		CLabels* labels;
+
+		if (get_global_parallel()->get_num_threads() == 1)
```
karlnapf: Just realised we really don't want to copy/replicate this buggy x-validation codes ...

durovo:
@karlnapf Thanks a lot for the valuable feedback! I don't mind the delays at all, I understand that you must be busy with other work as well. I have already added the descriptions to my local repo but I have not pushed it yet. Maybe I wasn't clear about the fact that I didn't push any code this time. I was just discussing the calibration curves and checking if the `fit_sigmoid` was working correctly since the calibrated curve that I got does not seem calibrated at all.
I will try and remove the replicated x-validation code. The code would indeed be much cleaner if could reuse the existing x-validation API. I will try to make a more minimal PR just as you suggested.

src/shogun/evaluation/CrossValidatedCalibration.cpp
durovo: Understood.

src/shogun/evaluation/CalibrationMethod.h
durovo: Understood.

src/shogun/evaluation/Calibration.h
durovo: CalibratedClassifier does seem like a better name

src/shogun/evaluation/Calibration.cpp
durovo: Okay :)

durovo:
@karlnapf I still have some doubts regarding the `fit_sigmoid` implementation. 
I understand that we can check the scores for +1/-1. However, shouldn't we be fitting the sigmoid parameters with respect to the true labels? Currently, the implementation only takes the decision function of the machine to fit the parameters. Shouldn't it also take the labels as an input too? Doesn't the current approach introduce some kind of a bias?
I have gone through the implementation [paper](https://www.csie.ntu.edu.tw/%7Ecjlin/papers/plattprob.pdf) on Platt scaling which I believe was followed in the shogun implementation and also seen the scikit-learn implementation. Both take the decision function and the true labels (y) as input when fitting. To be specific, I believe that https://github.com/shogun-toolbox/shogun/blob/c87d5844628dae56ed8b47ed287d0f84e1d3037f/src/shogun/mathematics/Statistics.cpp#L865 should be `if (labels[i] > 0)` and we should take labels as an input to the function. Please correct me if I'm wrong.

durovo:
@karlnapf I still have some doubts regarding the `fit_sigmoid` implementation. 
I understand that we can check the scores for +1/-1. However, shouldn't we be fitting the sigmoid parameters with respect to the true labels? Currently, the implementation only takes the decision function of the machine to fit the parameters. Shouldn't it also take the labels as an input too? Doesn't the current approach introduce some kind of a bias?
I have gone through the implementation [paper](https://www.csie.ntu.edu.tw/%7Ecjlin/papers/plattprob.pdf) on Platt scaling which I believe was followed in the shogun implementation and also seen the scikit-learn implementation. Both take the decision function and the true labels (y) as input when fitting. To be specific, I believe that https://github.com/shogun-toolbox/shogun/blob/c87d5844628dae56ed8b47ed287d0f84e1d3037f/src/shogun/mathematics/Statistics.cpp#L865 should be `if (labels[i] > 0)` and we should take labels as an input to the function. Please correct me if I'm wrong.

durovo:
The libsvm implementation also uses the target labels https://github.com/cjlin1/libsvm/blob/88a1881f03ca139beff93170d7e6f36477fabe54/svm.cpp#L1733

karlnapf:
Think when is `labels[i] > 0` true. It is exactly then true when `scores[i]>0` is true.
This is since `labels[i] = sign(scores[i])`

durovo:
@karlnapf Sorry, for pestering you about this, however, I still don't think it is correct. From what I read, with platt scaling we are fitting our predictions (i.e. `sign(scores[i])`) to the true labels. The method in shogun gives the correct results when the machine is 100% accurate (i.e. when `true_labels[i] == sign(scores[i])`). This will not happen when the predictions are not accurate. The sklearn implementation of Sigmoid Calibration uses the true labels, in fact they have a test that checks for changes in calibrated results when the true labels are changed (after training the machine).

karlnapf:
Hey!
I was in fact wrong! Thanks for pestering me :)
Yes it should use the provided training labels, not the predicted ones. I was on the wrong train there.
Wanna fix this?

durovo:
Yes, I will fix this. :)

durovo:
@karlnapf After #4166 , the calibration is working correctly
image

karlnapf:
Hey that is great! I will review the other one now and then we can get this here in

Issue #4011, opened 2017-12-10
Update requirement.txt for examples/meta to include Ctags along with ply


shubham808:
Currently the requrement.txt has 
`ply`
changing it to include Ctags as the `examples/meta/CmakeLists.txt` requires it

PR #4012, opened 2017-12-10
Fixes issue #4011

shubham808:
Updates examples/meta/requrements.txt


vigsterkr:
thnx for the patch but, ctags is not a python package, hence this requirements will always fail

Issue #3996, opened 2017-11-15
cmake does not check for openblas headers


karlnapf:
I.e. when compiling shogun against openblas using anaconda, compilation fails due to a missing `cblas.h`, which can be fixed via `apt-get install libopenblas-dev`. It would be good if cmake checked this and errors

vermashresth:
@karlnapf Can I work on this issue?

vigsterkr:
@vermashresth yes go ahead. take a look how it is done in case of MKL

vermashresth:
@vigsterkr Currently, the [ShogunFindLAPACK.cmake](https://github.com/shogun-toolbox/shogun/blob/develop/cmake/ShogunFindLAPACK.cmake#L70-L72) checks for the existence of `cblas.h` by searching for `cblas_dgemv` function in the found Lapack library. But it is a weak check as it might be possible that `cblas_dgemv` exists but still there is no `cblas.h` header file (for example having libopenblas-base installed and not having libopenblas-dev causes this error).
Would it be better then to remove this check altogether and only look for `cblas.h` header file like it is done in case of MKL?

vigsterkr:
@vermashresth noup that's not a good idea as you need to consider the case when it's an atlas version of LAPACK/BLAS.... basically you need to figure it out that it's openblas based LAPACK/BLAS and if that's the case then you need to do a search for cblas.h

vigsterkr:
@vermashresth mmm it's not necessary to create a `FindOpenBLAS` file to detect the BLAS/LAPACK type.. of course you can do that as well... it's up to you. only thing is that you should be able to figure out whether the detected LAPACK API is openblas... and in case it is then it should be able to find the cblas.h or error if there's no such header available....

vermashresth:
@vigsterkr I'm stuck on figuring out whether the detected LAPACK API is openblas. The path names don't always contain openblas keywords so can't use regex and detecting openblas otherwise doesn't ensure whether it actually is the detected version. Can you suggest some other way? I'm still a beginner so taking time to understand

vigsterkr:
how do you mean that the path name in case of openblas is not always containing openblas keyword? can you show me a distro/installation where this is the case?

vermashresth:
@vigsterkr In my system, I have `libopenblas-dev` installed and at installation time it sets `/usr/lib/libblas.so` to point to `/usr/lib/openblas-base/libblas.so.3` and same with liblapack.so
[...]

PR #4013, opened 2017-12-13
detect cblas.h during cmake for openblas

vermashresth:
resolves https://github.com/shogun-toolbox/shogun/issues/3996
When Atlas is not detected, a further check is implemented to see if detected LAPACK API is OpenBLAS using search for an openBLAS specific function. If it is then cblas.h is checked otherwise all LAPACK flags are unset and error is shown.

### DIFF
--- a/cmake/ShogunFindLAPACK.cmake
+++ b/cmake/ShogunFindLAPACK.cmake
@@ -91,6 +91,22 @@ IF (LAPACK_FOUND)
       ENDIF()
     ENDIF()
 
+    IF(FOUND_CBLAS_DGEMV AND NOT HAVE_ATLAS )
+      #check is detected BLAS/LAPACK is OpenBLAS by looking for an OpenBLAS specific function 
+      check_library_exists("${LAPACK_LIBRARIES}" openblas_set_num_threads "" FOUND_OPENBLAS)
+      IF(FOUND_OPENBLAS)
+        #check if cblas.h exists
+        FIND_PATH(CBLAS_INCLUDE_DIR cblas.h)
+          IF(NOT CBLAS_INCLUDE_DIR)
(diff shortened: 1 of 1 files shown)


cmake/ShogunFindLAPACK.cmake
```diff
@@ -91,6 +91,22 @@ IF (LAPACK_FOUND)
       ENDIF()
     ENDIF()
 
+    IF(FOUND_CBLAS_DGEMV AND NOT HAVE_ATLAS )
```
vigsterkr: remove the whitespace after HAVE_ATLAS

cmake/ShogunFindLAPACK.cmake
```diff
@@ -91,6 +91,22 @@ IF (LAPACK_FOUND)
...
+      #check is detected BLAS/LAPACK is OpenBLAS by looking for an OpenBLAS specific function 
+      check_library_exists("${LAPACK_LIBRARIES}" openblas_set_num_threads "" FOUND_OPENBLAS)
```
vigsterkr: to follow the logic <PKG>_FOUND please change this to `OpenBLAS_FOUND`

cmake/ShogunFindLAPACK.cmake
```diff
@@ -91,6 +91,22 @@ IF (LAPACK_FOUND)
...
+      IF(FOUND_OPENBLAS)
+        #check if cblas.h exists
+        FIND_PATH(CBLAS_INCLUDE_DIR cblas.h)
+          IF(NOT CBLAS_INCLUDE_DIR)
+	    MESSAGE(STATUS "Found OPENBLAS, but not cblas.h. Make sure that cblas headers are available in order to use OPENBLAS as BLAS/Lapack backend")
```
vigsterkr: `Make sure that cblas.h header is available within the header search path in order to use OpenBLAS as BLAS/Lapack backend`

PR #4026, opened 2017-12-17
GMM: possible memory leaks.

zoq:
Fix possible memory leaks in the GMM code, took me some time to get the idea of the ``train_smem`` method, but really liked the structure.


karlnapf:
Cool! thx! Let's see what travis says. I don't remember our memory tests complaining about this code ....


We want to make this code a  bit cleaner at some point, using linalg / iterators, and proper base classes for the distributions ...
@micmn You could linalg'ise this at some point  ;)

PR #4027, opened 2017-12-17
Make SGVector(len) 'ctor zero-initialize memory

micmn:
Same behavior as SGMatrix(rows, cols) [I don't know if it makes sense or this difference is ok and I should call SGVector::zero()].

Should fix DotIterator::add() failing unittest due to non initialized memory.
http://buildbot.shogun-toolbox.org:8080/#/builders/25/builds/158


karlnapf:
I think we at some point agreed that we should not initialise to zero...
But if matrix already does it, whatever :)

PR #4031, opened 2017-12-17
Set classifier_mkl_svmlight_modelselection_bug loglevel to info

micmn:
Valgrind analysis build fails with: `parser error : xmlSAX2Characters: huge text node` (
http://buildbot.shogun-toolbox.org:8080/#/builders/2/builds/56/steps/9/logs/stdio).
I think that the cause is the script `classifier_mkl_svmlight_modelselection_bug.py`: the log level is set to debug and its outputs ~115MB of text.

I tried running xsltproc on the output file removing that test case (the file goes down to ~5MB) and it generates the html.


karlnapf:
Lol, I mean this example was just added to document this bug that there is in MKL. It is funny that we managed to maintain that things for such a long time ....

PR #4033, opened 2017-12-18
Get/put instead of get/set

lisitsyn:
Motivation is basic
- No need to replace set -> sets so no need to rename get
- Consistent with Java Map API


lisitsyn:
GREEN!

karlnapf:
SHOEGUEN!!!!

lisitsyn:
SHØGÜN

PR #4034, opened 2017-12-19
Test python examples against build

lisitsyn:


karlnapf:
Now we have python using the build and R using the system install?

PR #4035, opened 2017-12-19
Do not use anything but headers for tags

lisitsyn:
We use tags to add dependencies such as `#include`. Only headers are to be used here.

@sorig that was causing that segfault of C++ meta_api/calls example! 🤣

### DIFF
--- a/src/shogun/CMakeLists.txt
+++ b/src/shogun/CMakeLists.txt
@@ -461,6 +461,7 @@ IF (CTAGS_FOUND)
         # functions, classes, macroses, enumerations, enumerators, typedefs
         --c++-kinds=fcdgetp
         --fields=+im
+        -h "h.hpp"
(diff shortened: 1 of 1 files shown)


src/shogun/CMakeLists.txt
```diff
@@ -461,6 +461,7 @@ IF (CTAGS_FOUND)
         # functions, classes, macroses, enumerations, enumerators, typedefs
         --c++-kinds=fcdgetp
         --fields=+im
+        -h "h.hpp"
```
karlnapf: ah of course!!

lisitsyn: No that didn't help. But in the next patch I added something that helped ;)

Now we don't have nothing from .cpp in tags file.

karlnapf: lol

well ... anyways :dancer:

PR #4036, opened 2017-12-20
Memory leaks fixes for unit tests

geektoni:
This patch fixes some memory leaks discovered by valgrind inside unit tests.


src/shogun/base/class_list.h
```diff
@@ -43,6 +43,7 @@ namespace shogun {
 		auto* cast = dynamic_cast<T*>(object);
 		if (!cast)
 		{
+			delete object;
```
lisitsyn: Catch!

lisitsyn:
looks good!

PR #4039, opened 2017-12-22
Feature/fix bsd buidlot

karlnapf:
Don't run recently merged mkl example if SVMLight is not available (as on the bsd build)


karlnapf:
gnaaaaa

karlnapf:
Ok, let's see what the buildbot says

PR #4045, opened 2017-12-23
Better non-owning any implementation

lisitsyn:


src/shogun/lib/any.h
```diff
@@ -207,7 +197,7 @@ namespace shogun
 		 */
 		virtual void set(void** storage, const void* v) const
 		{
-			*(storage) = const_cast<void*>(v);
+			*static_cast<T*>(*(storage)) = T(*reinterpret_cast<T const*>(v));
```
karlnapf: party

src/shogun/lib/any.h
```diff
@@ -324,11 +312,20 @@ namespace shogun
...
+				return *(this);
+			}
 			policy->clear(&storage);
-			policy = other.policy;
-			policy->set(&storage, other.storage);
-			return *(this);
+			if (other.policy->policy_type() == PolicyType::NON_OWNING) {
+				policy = other.policy;
+				storage = other.storage;
+				return *(this);
```
karlnapf: could be moved outside the `if/then/else`

tests/unit/lib/Any_unittest.cc
```diff
@@ -105,13 +105,94 @@ TEST(Any, erase_type_non_owning)
...
+	int32_t integer = 10;
+	int32_t other = 42;
+	auto any = erase_type_non_owning(&integer);
+	EXPECT_EQ(any.as<int32_t>(), integer);
+	any = erase_type(other);
+	EXPECT_EQ(any.as<int32_t>(), other);
+	EXPECT_EQ(integer, other);
+}
+
+TEST(Any, assign_non_owning_into_non_owning_then_owning)
```
karlnapf: :D

src/shogun/lib/any.h
```diff
@@ -396,28 +393,19 @@ namespace shogun
 			return policy->type_info();
 		}
 
-	private:
-		void assert_same_policy_type(BaseAnyPolicy* other_policy)
```
karlnapf: I understand we don't need this anymore? Is that why it is removed?

karlnapf:
:heart:

src/shogun/lib/any.h
lisitsyn: Yes, but we need another check now

lisitsyn:
I believe it should cover all of them. But I missed the type-checking part so need to add a few more tests and checks.

src/shogun/lib/any.h
lisitsyn: Funny is that I don't think it actually was correct before.

lisitsyn:
Some SGObject tests are failing yet.

lisitsyn:
@karlnapf all green, could you please take a look again and merge if all good

PR #4048, opened 2017-12-24
Make SG_ADD register parameters into the new parameter map

lisitsyn:


src/shogun/modelselection/GradientModelSelection.cpp
```diff
@@ -88,12 +88,14 @@ class GradientModelSelectionCostFunction: public FirstOrderCostFunction
 		SG_ADD((CSGObject **)&m_obj, "GradientModelSelectionCostFunction__m_obj",
 			"obj in GradientModelSelectionCostFunction", MS_NOT_AVAILABLE);
 		m_func_data = NULL;
+        /*
```
lisitsyn: @karlnapf need your comments here. They were registered by value which is not good.

src/shogun/lib/SGSparseVector.h
```diff
@@ -216,6 +216,13 @@ template <class T> class SGSparseVector : public SGReferencedData
 
 };
 
+template <class T>
```
lisitsyn: Anything that goes into Any should have operator==, that's why I added it here

lisitsyn:
Apparently, this probably slows down the compilation as running tests on travis times out most of the times.

vigsterkr:
@lisitsyn the reason for the compilation slowdown is because you've changed a .h that basically being used everywhere hence the ccache-ed objects cannot be used... hence travis needs to compile everything from scratch... see in case of appveyor the compilation time is always 40+ mins ;)

karlnapf:
ho ho ho

src/shogun/base/SGObject.h
```diff
@@ -76,19 +76,23 @@ template <class T> class SGStringList;
...
 
-#define SG_ADD4(param, name, description, ms_available) {\
-		m_parameters->add(param, name, description);\
-		if (ms_available)\
-			m_model_selection_parameters->add(param, name, description);\
-}
+#define SG_ADD4(param, name, description, ms_available)                        \
+	{                                                                          \
+		m_parameters->add(param, name, description);                           \
+		watch_param(name, param);                                              \
```
karlnapf: this reminds me. We currently have two maps: one for registering all paramters, one for modelselection available only.
I suggest we don't keep this design with tags, but rather have one and somehow mark them otherwise ?

lisitsyn: I think the parameter we store should be 'rich'. This means it is not just 'Any' but 'Any' + extras. Such as this model_selection_enabled.

karlnapf: Yep I agree. I think this is a place to include the domain bounds, information about gradients, sampling callbacks (?), priors, etc

PR #4056, opened 2017-12-28
Support equals in operator== of Any

lisitsyn:


src/shogun/lib/any.h
```diff
@@ -71,6 +71,54 @@ namespace shogun
...
+		struct more_important : general
+		{
+		};
+
+		struct maybe_most_important : more_important
+		{
+		};
+
+		template <class T>
+		auto compare_impl(by_default, T& lhs, T& rhs) = delete;
```
vigsterkr: `auto` return is only c++14. but no worries just up the required version here:
https://github.com/shogun-toolbox/shogun/blob/develop/CMakeLists.txt#L35

lisitsyn:
@vigsterkr sir I did address your comments!

src/shogun/lib/any.h
karlnapf: Oh wow so this happened? :dancing_men:

Issue #2747, opened 2015-02-28
make use of linalg within Shogun


karlnapf:
This task is to make Shogun's internal make use of the new [linalg framework](https://github.com/shogun-toolbox/shogun/wiki/README_linalg)

It is pretty simple: Take a method that you want to replace, say some dot product. Then go through Shogun's code and search for appearances of dot products. Those could take a lot of different forms

 * manual loops
 * lapack calls (look up the dot product call in lapack)
 * eigen3 calls (easier to find, search for eigen3 in the code)

For example, linear solves are much easier to find as you can search for eigen3/lapack's method for that directly.

karlnapf:
Useful for the linalg GSoC project. https://github.com/shogun-toolbox/shogun/wiki/GSoC_2015_project_linalg

sanuj:
Are we using c++11 everywhere in shogun then?

SunilMahendrakar:
@karlnapf I have already started on this with scale https://github.com/shogun-toolbox/shogun/issues/2717 pull https://github.com/shogun-toolbox/shogun/pull/2744. I will add more soon.

karlnapf:
@sanuj it is not yet done. @lambday will guide this and merge the requirement patch too ;)

ajaybhat:
@karlnapf & @SunilMahendrakar could i know the status of this? Maybe I could pick up where you left off.

ckousik:
https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/distributions/Gaussian.cpp#L318
@karlnapf Shouldn't the LAPACK call be scrapped in favor of the linalg call instead of the conditional compilation?

karlnapf:
absolutely! I think most of linalg that is needed is there, so that should be easy. Make sure to send a few minimal unit tests before modifying the code ...

karlnapf:
There are still many loops who wait to be replaced with linalg calls

PR #4058, opened 2017-12-29
Issue #2747 - Using linalg everywhere

ckousik:
This PR is a work in progress which will transition shogun to using linalg for linear algebra operations. Minimal unit tests will be written for required classes.
#2747

### DIFF
--- a/tests/unit/distribution/Gaussian_unittest.cc
+++ b/tests/unit/distribution/Gaussian_unittest.cc
@@ -1,0 +1,149 @@
+/*
+ * Copyright (c) The Shogun Machine Learning Toolbox
+ * Written (w) 2014 Parijat Mazumdar
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * The views and conclusions contained in the software and documentation are those
+ * of the authors and should not be interpreted as representing official policies,
+ * either expressed or implied, of the Shogun Development Team.
+ */
+
+#include <gtest/gtest.h>
+#include <shogun/lib/SGVector.h>
+#include <shogun/features/DenseFeatures.h>
+#include <shogun/lib/SGMatrix.h>
+#include <shogun/distributions/Gaussian.h>
+/* #include <shogun/linalg/LinalgNamespace.h> */
+#include <shogun/mathematics/Math.h>
+
+using namespace shogun;
+
+TEST(Gaussian, log_pdf_single_1d)
+{
+	SGVector<float64_t> mean(1), point(1);
+	SGMatrix<float64_t> cov(1,1);
+	
+	mean[0] = 0;
+	point[0] = 0;
+	cov(0,0) = 1.0;
+
+	CGaussian *gauss = new CGaussian(mean, cov, FULL);
+	float64_t log_pdf = gauss->compute_log_PDF(point);
+	EXPECT_NEAR(log_pdf, -0.918938533205, 1e-8);
+	SG_UNREF(gauss);
+}
+
+TEST(Gaussian, log_pdf_multiple_independent_2d)
+{
+	SGVector<float64_t> mean(2);
+	SGMatrix<float64_t> cov(2,2);
+
+	mean[0] = mean[1] = 0;
+	// Random variables are independent
+	cov(0,0) = 1.0;
+	cov(0,1) = 0.0;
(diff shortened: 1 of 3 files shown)


lisitsyn:
Thanks for the patch!

Please take a look at the Travis CI status, the tests are apparently failing.

@karlnapf can you take a look?

ckousik:
@lisitsyn I have been through the travis build. The failure is caused by the style checker. I'll push another commit shortly to fix it. Sorry :(

tests/unit/distribution/Gaussian_unittest.cc
```diff
@@ -0,0 +1,149 @@
...
+TEST(Gaussian, log_pdf_single_1d)
+{
+	SGVector<float64_t> mean(1), point(1);
+	SGMatrix<float64_t> cov(1,1);
+	
+	mean[0] = 0;
+	point[0] = 0;
+	cov(0,0) = 1.0;
+
+	CGaussian *gauss = new CGaussian(mean, cov, FULL);
```
karlnapf: you could do `auto gaussian = some<CGaussian>(mean, cov, FULL)`

tests/unit/distribution/Gaussian_unittest.cc
```diff
@@ -0,0 +1,149 @@
...
+	SG_UNREF(gauss);
+}
+
+TEST(Gaussian, log_pdf_multiple_independent_2d)
+{
+	SGVector<float64_t> mean(2);
+	SGMatrix<float64_t> cov(2,2);
+
+	mean[0] = mean[1] = 0;
+	// Random variables are independent
```
karlnapf: no need for this comment

tests/unit/distribution/Gaussian_unittest.cc
```diff
@@ -0,0 +1,149 @@
...
+	cov(0,0) = 1.0;
+	cov(0,1) = 0.0;
+	cov(1,0) = 0.0;
+	cov(1,1) = 1.0;
+
+	auto gauss = new CGaussian(mean, cov, FULL);
+	// Find the log of the distribution function at the mean
```
karlnapf: no need for comment

tests/unit/distribution/Gaussian_unittest.cc
```diff
@@ -0,0 +1,149 @@
...
+	// position.
+	float64_t log_pdf = gauss->compute_log_PDF(mean);
+	EXPECT_NEAR(log_pdf, -1.83787706641, 1e-8);
```
karlnapf: float64 is 16 digits, not just 8

tests/unit/distribution/Gaussian_unittest.cc
```diff
@@ -0,0 +1,149 @@
...
+	SGMatrix<float64_t> data(1, 500);	
+	for(int32_t i = 0; i < 500; i++)
+		data(0, i) = CMath::randn_double();
+	auto train_features = new CDenseFeatures<float64_t>(data);
+	CGaussian *gauss = new CGaussian();
+	gauss->train(train_features);
+
+	auto mean = gauss->get_mean();
+	auto cov = gauss->get_cov();
+	EXPECT_NEAR(mean[0], 0.0, 1e-1);
```
karlnapf: i would rather compute the mean by hand and check whether it is the same here.

tests/unit/distribution/Gaussian_unittest.cc
```diff
@@ -0,0 +1,149 @@
...
+			auto rd = CMath::randn_double();
+			auto sigma = CMath::sqrt(sample_cov(i,i));
+			data(i, j) = sigma*rd + sample_mean[i];
+		}
+	}
+
+	CGaussian *gauss = new CGaussian();
+	auto train_features = new CDenseFeatures<float64_t>(data);
+	gauss->train(train_features);
+
```
karlnapf: see comments above, also hold here

tests/unit/distribution/Gaussian_unittest.cc
```diff
@@ -0,0 +1,149 @@
+/*
+ * Copyright (c) The Shogun Machine Learning Toolbox
+ * Written (w) 2014 Parijat Mazumdar
```
karlnapf: you can put your name here :)

ckousik:
@karlnapf I am getting precision errors when testing with eps set to 1e-16. I checked the DotFeatures unit tests and they are only checked for 1e-8.

ckousik:
@karlnapf I'm also concerned about the usefulness of these tests. The Gaussian mean and covairance matrices are generated from DotFeatures, and so this feels like a more elaborate test for DotFeatures than for Gaussian.

tests/unit/distribution/Gaussian_unittest.cc
```diff
@@ -0,0 +1,187 @@
...
+	cov(1,1) = 2.0;
+	auto gauss = some<CGaussian>(mean, cov, FULL);
+	float64_t log_pdf = gauss->compute_log_PDF(mean);
+	EXPECT_NEAR(log_pdf, -2.8108321409370021, 1e-16);
+}
+
+TEST(Gaussian, train_univariate)
+{
+	float64_t eps = 1e-8;
+	sg_rand->set_seed(1);
```
karlnapf: why do you need a fixed seed when you test for the estimation? Should be ok everytime no?

tests/unit/distribution/Gaussian_unittest.cc
```diff
@@ -0,0 +1,187 @@
...
+		data(0, i) = CMath::randn_double();
+
+	auto gauss = some<CGaussian>();
+	auto train_features = some<CDenseFeatures<float64_t>>(data);
+	gauss->train(train_features);
+
+	auto mean = gauss->get_mean();
+	auto cov = gauss->get_cov();
+
+	for (int32_t i = 0; i < sample_size; i++)
```
karlnapf: for `auto i : range(sample_size)` ... minor

tests/unit/distribution/Gaussian_unittest.cc
```diff
@@ -0,0 +1,187 @@
...
+
+TEST(Gaussian, train_univariate)
+{
+	float64_t eps = 1e-8;
+	sg_rand->set_seed(1);
+
+	SGMatrix<float64_t> data(1, 500);
+
+	int64_t sample_size = 500;
+	float64_t mn = 0.0, cv = 0.0;
```
karlnapf: I dont like this variable names. What about something meaningful? mean and covariance?

karlnapf: mu and Sigma?

karlnapf: or "mean_est" or so

tests/unit/distribution/Gaussian_unittest.cc
```diff
@@ -0,0 +1,187 @@
...
+	sample_mean[0] = 0.0;
+	sample_mean[1] = 2.0;
+
+	sample_cov(0,0) = 2.0;
+	sample_cov(1,1) = 4.0;
+	sample_cov(0,1) = sample_cov(1,0) = 0.0;
+
+	SGMatrix<float64_t> data(2, sample_size);
+
+	for(int32_t i = 0; i < 2; i++){
```
karlnapf: @vigsterkr don't we have ways to fill arrays with random gaussian numbers?

tests/unit/distribution/Gaussian_unittest.cc
```diff
@@ -0,0 +1,187 @@
...
+
+	auto gauss = some<CGaussian>();
+	auto train_features = some<CDenseFeatures<float64_t>>(data);
+	gauss->train(train_features);
+
+	linalg::zero(sample_mean);
+	linalg::zero(sample_cov);
+
+	for (int32_t i = 0; i < sample_size; i++)
+		train_features->add_to_dense_vec(1.0, i, sample_mean.vector, 2);
```
karlnapf: you are right, this should not be in here
I think there are linalg methods

karlnapf:
But this is already improving, let's iterate a few more times, and this will be good to merge.
THANKS! :)

PR #4069, opened 2018-01-05
Feature/equals with tags

karlnapf:
@lisitsyn This **almost** works, and side-steps **a lot** of older code (will delete later).

One thing that is quite annoying is "old" parameter registrations are not covered by tags. See e.g. 
```
m_parameters->add_vector(&m_array.array, &m_array.current_num_elements, "array",
			"Memory for dynamic array.");
```
in `CDynamicObjectArray`. This causes some tests to fail when using `equals` via tags. I am on it, but let me know if you have a quick fix for things like the above (reason `SG_ADD` is not used is since only a smaller part of the array is registered...)


src/shogun/lib/any.h
```diff
@@ -153,7 +153,12 @@ namespace shogun
 		auto compare_impl(maybe_most_important, T* lhs, T* rhs)
 		    -> decltype(lhs->equals(rhs))
 		{
-			return lhs->equals(rhs);
+			if (lhs && rhs)
```
karlnapf: @lisitsyn let me know if this is the right place to add those checks

src/shogun/base/AnyParameter.h
```diff
@@ -88,6 +88,16 @@ namespace shogun
 			return m_properties;
 		}
 
+		inline bool operator==(const AnyParameter& other)
+		{
+			return m_value == other.get_value();
```
karlnapf: good? not good?

src/shogun/base/SGObject.cpp
```diff
@@ -28,6 +28,7 @@
 
 #include <stdlib.h>
 #include <stdio.h>
+#include <typeinfo>
```
karlnapf: good? not good?

src/shogun/base/SGObject.cpp
```diff
@@ -820,7 +743,7 @@ AnyParameter CSGObject::get_parameter(const BaseTag& _tag) const
 	const auto& parameter = self->get(_tag);
 	if (parameter.get_value().empty())
 	{
-		SG_ERROR("There is no parameter called \"%s\" in %s",
+		SG_ERROR("There is no parameter called \"%s\" in %s\n",
```
karlnapf: gotcha!

src/shogun/base/SGObject.cpp
```diff
@@ -1016,3 +939,57 @@ std::string CSGObject::to_string() const
...
+	if (other==this)
+		return true;
+
+	if (other==nullptr)
+	{
+		SG_DEBUG("No object to compare to provided.\n");
+		return false;
+	}
+
+	/* Assumption: can use SGObject::get_name to distinguish types */
```
karlnapf: good? not good?

src/shogun/base/SGObject.cpp
```diff
@@ -1016,3 +939,57 @@ std::string CSGObject::to_string() const
...
+	/* Assumption: objects of same type have same set of tags. */
+	for (auto it=self->map.cbegin(); it!=self->map.cend(); it++)
+	{
+		auto tag = it->first;
+		auto own = it->second;
+		auto given = other->get_parameter(tag);
+
+		SG_SDEBUG("Comparing parameter %s::%s.\n", this->get_name(),
+				tag.name().c_str());
+		if (own != given)
```
karlnapf: all the magic happens in this single line, awesome!

src/shogun/base/SGObject.cpp
```diff
@@ -1016,3 +939,57 @@ std::string CSGObject::to_string() const
...
+		{
+			if (io->get_loglevel()<=MSG_DEBUG)
+			{
+				std::stringstream ss;
+				std::unique_ptr<AnyVisitor> visitor(new ToStringVisitor(&ss));
+
+				ss << "Own parameter " << this->get_name() << "::" << tag.name() << "=";
```
karlnapf: I would actually like to print the differences here (also for vectors which is currently shown as `...`), as that was useful in unit-testing in the past

src/shogun/base/AnyParameter.h
lisitsyn: Probably ok. Maybe needs a comment that it doesn't consider properties.

src/shogun/base/SGObject.cpp
lisitsyn: Was it missed?

src/shogun/base/SGObject.cpp
```diff
@@ -1016,3 +939,57 @@ std::string CSGObject::to_string() const
...
+	/* Assumption: can use SGObject::get_name to distinguish types */
+	if (strcmp(this->get_name(), other->get_name()))
+	{
+		SG_DEBUG("Own type %s differs from provided %s.\n",
+				get_name(), other->get_name());
+		return false;
+	}
+
+	/* Assumption: objects of same type have same set of tags. */
+	for (auto it=self->map.cbegin(); it!=self->map.cend(); it++)
```
lisitsyn: Why not `auto it : self->map`?

src/shogun/base/SGObject.cpp
lisitsyn: It could be a custom visitor. Or we can add another interface 'Reducer'.

src/shogun/lib/any.h
lisitsyn: Yeah that's a good catch.

lisitsyn:
You might want to fix the codestyle ;)
```
git clang-format-3.8 --commit 6f1fd162892988e212b5ddfc6de15db3a8bf1026 --binary /usr/bin/clang-format-3.8
```

tests/unit/base/SGObject_unittest.cc
```diff
@@ -80,38 +127,19 @@ TEST(SGObject,ref_unref_simple)
...
-	CRegressionLabels* labels2=new CRegressionLabels(10);
-
-	EXPECT_FALSE(labels->equals(labels2));
-
-	SG_UNREF(labels);
-	SG_UNREF(labels2);
-}
-
-TEST(SGObject,equals_DynamicObjectArray_equal)
+TEST(SGObject,DISABLED_equals_DynamicObjectArray_equal)
```
karlnapf: @lisitsyn I had to disable this until we fix the registration of that partly used array parameter

karlnapf:
@lisitsyn ok it was green now apart form style, which I fixed. Let's wait for travis to be sure, but you can review now. And potentially merge

src/shogun/base/init.cpp
```diff
@@ -147,6 +154,26 @@ namespace shogun
 		return sg_io;
 	}
 
+	float64_t get_global_fequals_epsilon()
+	{
+		return sg_fequals_epsilon;
```
karlnapf: @vigsterkr this is fucked up, but the way it was before was even more so. We should just get rid of all of this accuracy business and rather fix the serialization, which should not be lossy (in my eyes)

src/shogun/mathematics/Math.h
```diff
@@ -301,27 +301,31 @@ class CMath : public CSGObject
...
+			    const T diff = CMath::abs<T>((a-b));
 
 				// Handle this separately since NAN is unordered
 				if (CMath::is_nan((float64_t)a) && CMath::is_nan((float64_t)b))
 					return true;
 
 				// Required for JSON Serialization Tests
-				if (tolerant)
-					return CMath::fequals_abs<T>(a, b, eps);
+			    if (get_global_fequals_tolerant())
```
karlnapf: @vigsterkr here the beauty is

src/shogun/mathematics/Math.h
```diff
@@ -301,27 +301,31 @@ class CMath : public CSGObject
...
-				const T absB = CMath::abs<T>(b);
-				const T diff = CMath::abs<T>((a-b));
+		    template <class T, class = typename std::enable_if<
+		                           std::is_floating_point<T>::value>::type>
+		    static inline bool
+		    fequals(const T& a, const T& b, const float64_t eps_)
+		    {
+			    // global fequals epsilon might override passed one
+			    // hack for lossy serialization formats
+			    float64_t eps = std::max(eps_, get_global_fequals_epsilon());
```
karlnapf: here it comes ...

tests/unit/base/SGObject_unittest.cc
```diff
@@ -80,38 +127,19 @@ TEST(SGObject,ref_unref_simple)
...
-	EXPECT_TRUE(TParameter::compare_ptype(PT_SGOBJECT, &array1, &array2));
+	EXPECT_TRUE(array1->equals(array2));
+	EXPECT_TRUE(array2->equals(array1));
 
 	SG_UNREF(array1);
 	SG_UNREF(array2);
 }
 
-TEST(SGObject,equals_DynamicObjectArray_equal_after_resize)
+TEST(SGObject, DISABLED_equals_DynamicObjectArray_equal_after_resize)
```
karlnapf: @lisitsyn @vigsterkr I disabled this for now. With this patch, DynamicObjectArray is not serializable anymore. We need to fix the registration of the underlying array (not all of it, only used parts)

lisitsyn:
Ok this is deffs improvement even though accuracy business is meh :)

karlnapf:
Ok baby, lets merge it, just fixed the style issues....

karlnapf:
I take that back, windows is not happy, https://ci.appveyor.com/project/vigsterkr/shogun/build/1668#L16427

karlnapf:
One unit test was actually broken (and under linux only passed accidentally with an invalid read memory error). Should be fixed now.

karlnapf:
windows times out and travis is angry about style (even though I ran the checker, I think it is confused).
I think that means merge ... Objections?

karlnapf:
I'll just do it and fix tomorrow if there are problems

PR #4072, opened 2018-01-07
Support tag-based vector/matrix parameters in SWIG (float64 by default)

lisitsyn:


src/interfaces/swig/SGBase.i
```diff
@@ -528,3 +530,5 @@ SUPPORT_TAG(String, string, std::string)
 SUPPORT_TAG(Float64, float, float64_t)
 SUPPORT_TAG(Int64, int, int64_t)
 SUPPORT_TAG(Object, object, CSGObject*)
+SUPPORT_TAG(Vector, vector, SGVector<float64_t>)
```
vigsterkr: we have the ancient macros like USE_FLOAT64 etc. )

vigsterkr:
this is solved already

Issue #4068, opened 2018-01-04
OSX 10.13.2, Python 3.5. Segfault code: 11 on importing shogun


dhelekal:
OSX Version: OSX 10.13.2
This issue seems to be OSX specific, failed to replicate on linux.

After creating an anaconda environment via
`conda create -n envname python=3.5`
and installing shogun via
`conda install -c conda-forge shogun`

Importing shogun as such:
```
$ python
Python 3.5.4 |Anaconda, Inc.| (default, Nov  8 2017, 18:11:28) 
[GCC 4.2.1 Compatible Clang 4.0.1 (tags/RELEASE_401/final)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from shogun import *
Segmentation fault: 11
```

Crashes with segmentation fault: 11.

lldb output:

```
Process 2842 launched: '/usr/local/anaconda3/envs/zeppelin_env/bin/python' (x86_64)
Python 3.5.4 |Anaconda, Inc.| (default, Nov  8 2017, 18:11:28) 
[GCC 4.2.1 Compatible Clang 4.0.1 (tags/RELEASE_401/final)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from shogun import *
Process 2842 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x25)
    frame #0: 0x0000000105b398ef libpython3.5m.dylib`PyUnicode_InternInPlace + 111
libpython3.5m.dylib`PyUnicode_InternInPlace:
->  0x105b398ef <+111>: movb   0x25(%rax), %r15b
    0x105b398f3 <+115>: movq   0x1c559e(%rip), %rax      ; _PyThreadState_Current
    0x105b398fa <+122>: movb   $0x1, 0x25(%rax)
    0x105b398fe <+126>: movq   %rbx, %rsi
Target 0: (python) stopped.
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x25)
  * frame #0: 0x0000000105b398ef libpython3.5m.dylib`PyUnicode_InternInPlace + 111
    frame #1: 0x0000000105b4966a libpython3.5m.dylib`PyUnicode_InternFromString + 74
    frame #2: 0x0000000105b04515 libpython3.5m.dylib`PyType_Ready + 389
[...]

PR #4073, opened 2018-01-07
Do not link to python in conda (#4068)

lisitsyn:


vigsterkr:
@lisitsyn great that you've tracked down this one... only thing: dont' we wanna switch to static linking of the modular interfaces with libshogun eventually? would that actually fix this buck per se?

lisitsyn:
Yeah, I agree we want to link statically. But that's shogun linking. In this case it is python linking.

lisitsyn:
Ok apparently I broke octave somehow :)

PR #4077, opened 2018-01-13
Use linalg in GMM

vinx13:
This PR add unit test of GMM, and replace LAPACK usage with linalg . (#2747)

### DIFF
--- a/src/shogun/clustering/GMM.cpp
+++ b/src/shogun/clustering/GMM.cpp
@@ -9,15 +9,12 @@
  */
 #include <shogun/lib/config.h>
 
-#ifdef HAVE_LAPACK
-
 #include <shogun/base/Parameter.h>
 #include <shogun/clustering/GMM.h>
 #include <shogun/clustering/KMeans.h>
 #include <shogun/distance/EuclideanDistance.h>
 #include <shogun/labels/MulticlassLabels.h>
 #include <shogun/mathematics/Math.h>
-#include <shogun/mathematics/lapack.h>
 #include <shogun/mathematics/linalg/LinalgNamespace.h>
 #include <shogun/multiclass/KNN.h>
 #include <vector>
@@ -423,7 +420,11 @@ void CGMM::partial_em(int32_t comp1, int32_t comp2, int32_t comp3, float64_t min
 		SGMatrix<float64_t> c2=components[2]->get_cov();
 		linalg::add(c1, c2, c1, alpha1, alpha2);
 
-		components[1]->set_d(SGVector<float64_t>(SGMatrix<float64_t>::compute_eigenvectors(c1.matrix, dim_n, dim_n), dim_n));
+		SGVector<float64_t> eigenvalues(dim_n);
+		SGMatrix<float64_t> eigenvectors(dim_n, dim_n);
+		linalg::eigen_solver(c1, eigenvalues, eigenvectors);
+
+		components[1]->set_d(eigenvalues);
 		components[1]->set_u(c1);
 
 		float64_t new_d=0;
@@ -585,11 +586,8 @@ void CGMM::max_likelihood(SGMatrix<float64_t> alpha, float64_t min_cov)
 			switch (cov_type)
 			{
 				case FULL:
-				    cblas_dger(
-				        CblasRowMajor, num_dim, num_dim,
-				        alpha.matrix[j * alpha.num_cols + i], v.vector, 1,
-				        v.vector, 1, (double*)cov_sum.matrix, num_dim);
-
+				    linalg::dger(
+				        alpha.matrix[j * alpha.num_cols + i], v, v, cov_sum);
 				    break;
 			    case DIAG:
 			    {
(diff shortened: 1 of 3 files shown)


tests/unit/clustering/gmm_unittest.cc
```diff
@@ -0,0 +1,90 @@
...
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * The views and conclusions contained in the software and documentation are
+ * those
+ * of the authors and should not be interpreted as representing official
+ * policies,
+ * either expressed or implied, of the Shogun Development Team.
+ *
+ * Authors: 2018 Wuwei Lin
```
karlnapf: pls stick to the style we use

tests/unit/clustering/gmm_unittest.cc
```diff
@@ -0,0 +1,90 @@
...
+	rect(0, 1) = 1;
+	rect(0, 2) = 10;
+	rect(0, 3) = 4;
+	rect(1, 0) = 0;
+	rect(1, 1) = 7;
+	rect(1, 2) = 6;
+	rect(1, 3) = 4;
+
+	const int num_components = 2;
+	CGMM* clustering = new CGMM(num_components, FULL);
```
karlnapf: `auto clustering = some<CGMM>(num_components, FULL)` and then you dont need to `SG_UNREF` anything later

tests/unit/clustering/gmm_unittest.cc
```diff
@@ -0,0 +1,90 @@
...
+	CDenseFeatures<float64_t>* features = new CDenseFeatures<float64_t>(rect);
+	SG_REF(clustering);
+	SG_REF(features);
+
+	clustering->train(features);
+	clustering->train_em();
+
+	SGVector<float64_t> coef = clustering->get_coef();
+
+	EXPECT_NEAR(coef[0], 0.75, 1e-10);
```
karlnapf: where do these numbers come from?

src/shogun/clustering/GMM.cpp
```diff
@@ -585,11 +583,8 @@ void CGMM::max_likelihood(SGMatrix<float64_t> alpha, float64_t min_cov)
 			switch (cov_type)
 			{
 				case FULL:
-				    cblas_dger(
-				        CblasRowMajor, num_dim, num_dim,
-				        alpha.matrix[j * alpha.num_cols + i], v.vector, 1,
-				        v.vector, 1, (double*)cov_sum.matrix, num_dim);
-
+				    linalg::dger(
```
karlnapf: didn't know we had that, cool!

karlnapf:
https://travis-ci.org/shogun-toolbox/shogun/jobs/328667417#L718

karlnapf:
This broke the windows build, due to removing the lapack guard
```[00:28:32]   C:\projects\shogun\src\shogun\clustering\GMM.cpp(424): error C2039: 'compute_eigenvectors': is not a member of 'shogun::SGMatrix<float64_t>' [C:\projects\shogun\build\src\shogun\libshogun.vcxproj]```

If you want to remove that guard, you will need to use linalg eigenvectors

File: tests/unit/clustering/gmm_unittest.cc (hunk 3 above)
Comment (vinx13) [reply tests/unit/clustering/gmm_unittest.cc]: They are computed with shogun, before my modification.

Comment (karlnapf) [reply tests/unit/clustering/gmm_unittest.cc]: I see.
While this is ok-ish, even better would be to compute them somehow otherwise. How do we know it was correct before. And actually for self consistency of results, we have the integration tests. So I would suggest to remove this unit test, and add a meta example +test data on GMM first (see developing and examples readme), and then do the change you did

Alternatively, we could test against sklearn, pen&paper, or something that ensures absolute correctness in some way

Comment (vinx13) [reply tests/unit/clustering/gmm_unittest.cc]: There's already a GMM meta example and test data. So I removed this unit test.

**Comment by karlnapf** [comment]
FYI: Your change did not change any of the outputs of the gmm example, see https://travis-ci.org/shogun-toolbox/shogun/jobs/329120177#L5035

File: src/shogun/clustering/GMM.cpp [hunk 5]
```diff
@@ -423,7 +420,11 @@ void CGMM::partial_em(int32_t comp1, int32_t comp2, int32_t comp3, float64_t min
 		SGMatrix<float64_t> c2=components[2]->get_cov();
 		linalg::add(c1, c2, c1, alpha1, alpha2);
 
-		components[1]->set_d(SGVector<float64_t>(SGMatrix<float64_t>::compute_eigenvectors(c1.matrix, dim_n, dim_n), dim_n));
+		SGVector<float64_t> eigenvalues(dim_n);
```
karlnapf: good!
Maybe we should also offer an API for just eigenvalues?
But that could be done later

karlnapf:
Great, thanks for the patch.
Want to have a look at `LinearRidgeRegression` next? I saw it is also guarded, but it in fact uses Eigen3 inside

PR #4078, opened 2018-01-13
Implement clone for any

lisitsyn:
@karlnapf I removed some SG_SDEBUG from any.h as it is kind of circular dependency. Gets too hard to test separately.

### DIFF
--- a/src/shogun/lib/SGVector.cpp
+++ b/src/shogun/lib/SGVector.cpp
@@ -393,7 +393,7 @@ void SGVector<T>::free_data()
 }
 
 template <class T>
-bool SGVector<T>::equals(SGVector<T>& other) const
+bool SGVector<T>::equals(const SGVector<T>& other) const
 {
 	assert_on_cpu();
 	if (other.vlen!=vlen)
@@ -411,7 +411,7 @@ bool SGVector<T>::equals(SGVector<T>& other) const
 #ifndef REAL_EQUALS
 #define REAL_EQUALS(real_t)                                                    \
 	template <>                                                                \
-	bool SGVector<real_t>::equals(SGVector<real_t>& other) const               \
+	bool SGVector<real_t>::equals(const SGVector<real_t>& other) const         \
 	{                                                                          \
 		assert_on_cpu();                                                       \
 		if (other.vlen != vlen)                                                \
--- a/src/shogun/lib/any.cpp
+++ b/src/shogun/lib/any.cpp
@@ -43,7 +43,7 @@ namespace shogun
 #ifndef REAL_COMPARE_IMPL
 #define REAL_COMPARE_IMPL(real_t)                                              \
 	template <>                                                                \
-	bool compare_impl(general, const real_t& lhs, const real_t& rhs)           \
+	bool compare_impl_eq(const real_t& lhs, const real_t& rhs)                 \
 	{                                                                          \
 		SG_SDEBUG("Comparing using fequals<" #real_t ">(lhs, rhs).\n");        \
 		return CMath::fequals(                                                 \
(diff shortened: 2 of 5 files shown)


src/shogun/lib/SGVector.cpp
```diff
@@ -393,7 +393,7 @@ void SGVector<T>::free_data()
 }
 
 template <class T>
-bool SGVector<T>::equals(SGVector<T>& other) const
+bool SGVector<T>::equals(const SGVector<T>& other) const
```
karlnapf: thx for spotting

src/shogun/lib/any.cpp
```diff
@@ -42,7 +42,8 @@ namespace shogun
 
 #ifndef REAL_COMPARE_IMPL
 #define REAL_COMPARE_IMPL(real_t)                                              \
-	auto compare_impl(more_important, real_t& lhs, real_t& rhs)->bool          \
+	template <>                                                                \
```
karlnapf: lol ... thx :)

src/shogun/lib/any.h
```diff
@@ -54,7 +53,7 @@ namespace shogun
 	 * @return human readable name of class
 	 */
 	template <typename T>
-	std::string demangledType()
+	std::string demangled_type()
```
karlnapf: good!

src/shogun/lib/any.h
```diff
@@ -138,37 +137,40 @@ namespace shogun
...
+		{
+			return compare_impl_eq(lhs, rhs);
+		}
 
 		template <class T>
-		auto compare_impl(more_important, T& lhs, T& rhs)
+		auto compare_impl(more_important, const T& lhs, const T& rhs)
 		    -> decltype(lhs.equals(rhs))
 		{
-			SG_SDEBUG("Comparing using lhs.equals(rhs).\n");
```
karlnapf: I think these are incredibly useful.
Can't you move the implementation to the .cpp and keep them?

src/shogun/lib/any.h
```diff
@@ -178,13 +180,58 @@ namespace shogun
...
 		}
+
+		template <class T>
+		inline T clone_impl(general, T& value)
+		{
+			return value;
+		}
+
+		template <class T>
+		inline auto clone_impl(maybe_most_important, T* value)
```
karlnapf: cool thx for that. Can use it to write the big clone method

src/shogun/lib/any.h
```diff
@@ -178,13 +180,58 @@ namespace shogun
...
+		template <class T>
+		inline T& mutable_value_of(void** ptr)
+		{
+			return *static_cast<T*>(*ptr);
+		}
 	}
 
+	using any_detail::typed_pointer;
+	using any_detail::value_of;
+	using any_detail::mutable_value_of;
```
karlnapf: we are slowly getting quite complex with this stuff .... :D

src/shogun/lib/any.h
```diff
@@ -307,11 +371,12 @@ namespace shogun
 		 * @param other_storage pointer to a pointer to another storage
 		 * @return true if both storages have same value
 		 */
-		bool equals(void** storage, void** other_storage) const
+		bool equals(const void* storage, const void* other_storage) const
```
karlnapf: ++

tests/unit/lib/Any_unittest.cc
```diff
@@ -305,3 +313,45 @@ TEST(Any, equals_value)
...
+}
+
+TEST(Any, clone_into_owning_via_copy)
+{
+	int a = 3;
+	int other = 5;
+	auto a_any = erase_type(a);
+	a_any.clone_from(erase_type(other));
+	EXPECT_EQ(a_any.as<int>(), other);
+}
```
karlnapf: I think we are missing SG*.clone usage, since that cannot be assigned by value.

src/shogun/lib/any.h
lisitsyn: No that's actually to improve readability. I was having hard time checking all the * and &

src/shogun/lib/any.h
lisitsyn: Yes we can. Let me do that in other PR.

src/shogun/lib/SGVector.cpp
lisitsyn: Well it does not compile otherwise.

tests/unit/lib/Any_unittest.cc
lisitsyn: I don't get it. We have a test for a simple class that has 'clone'.

PR #4079, opened 2018-01-14
Use github link to download googletest

lisitsyn:
Resolves #40766


lisitsyn:
Ok that was a mistake lol :)

PR #4081, opened 2018-01-14
add tag-based clone

karlnapf:
will remove old code in a second PR

### DIFF
--- a/src/shogun/lib/SGMatrix.cpp
+++ b/src/shogun/lib/SGMatrix.cpp
@@ -354,9 +354,12 @@ SGMatrix<T> SGMatrix<T>::clone() const
 template <class T>
 T* SGMatrix<T>::clone_matrix(const T* matrix, int32_t nrows, int32_t ncols)
 {
-	REQUIRE(matrix!=nullptr, "The underlying matrix is not allocated!\n");
-	REQUIRE(nrows>0, "Number of rows (%d) has to be positive!\n", nrows);
-	REQUIRE(ncols>0, "Number of cols (%d) has to be positive!\n", ncols);
+	if (!matrix || !nrows || !ncols)
+		return nullptr;
+
+	if (matrix)
+		REQUIRE(nrows > 0, "Number of rows (%d) has to be positive!\n", nrows);
+	REQUIRE(ncols > 0, "Number of cols (%d) has to be positive!\n", ncols);
 
 	auto size=int64_t(nrows)*ncols;
 	T* result=SG_MALLOC(T, size);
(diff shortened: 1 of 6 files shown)


tests/unit/base/SGObject_unittest.cc
```diff
@@ -99,6 +99,74 @@ TEST(SGObject, equals_other_has_NULL_parameter)
...
+{
+	SGMatrix<float64_t> data(10, 10);
+	for (auto i : range(data.num_rows * data.num_cols))
+		data.matrix[i] = CMath::randn_double();
+
+	auto object = some<CDenseFeatures<float64_t>>(data);
+	ASSERT_TRUE(data.equals(object->get_feature_matrix()));
+
+	auto clone = object->clone();
+	EXPECT_NE(
```
karlnapf: @lisitsyn this is what I meant in my other comment. This assertion here fails, because the pointer is the same

tests/unit/base/SGObject_unittest.cc
```diff
@@ -99,6 +99,74 @@ TEST(SGObject, equals_other_has_NULL_parameter)
...
+{
+	SGVector<float64_t> data(10);
+	for (auto i : range(data.vlen))
+		data.vector[i] = CMath::randn_double();
+
+	auto object = some<CRegressionLabels>(data);
+	ASSERT_TRUE(data.equals(object->get_labels()));
+
+	auto clone = object->clone();
+	EXPECT_NE(data.data(), clone->get<SGVector<float64_t>>("labels").data());
```
karlnapf: This fails as well

tests/unit/base/SGObject_unittest.cc
```diff
@@ -99,6 +99,74 @@ TEST(SGObject, equals_other_has_NULL_parameter)
...
+		data.matrix[i] = CMath::randn_double();
+
+	auto object_parameter = some<CDenseFeatures<float64_t>>(data);
+	auto object = some<CGaussianKernel>();
+	object->init(object_parameter, object_parameter);
+	ASSERT_TRUE(object_parameter->equals(object->get<CSGObject*>("lhs")));
+	ASSERT_TRUE(object_parameter->equals(object->get<CSGObject*>("rhs")));
+
+	auto clone = object->clone();
+	EXPECT_NE(object_parameter, clone->get<CSGObject*>("lhs"));
```
karlnapf: this does not fail since `->clone` was called

tests/unit/base/SGObject_unittest.cc
lisitsyn: ok got it now. Seems trivial to add? :)

karlnapf: I'll do it :)

karlnapf:
will have to work all sorts of stuff out, don't look at this for now

tests/unit/base/SGObject_unittest.cc
karlnapf: #4083

PR #4083, opened 2018-01-15
add clone via ".clone" method for SGVector etc

karlnapf:


tests/unit/lib/Any_unittest.cc
```diff
@@ -355,3 +357,35 @@ TEST(Any, clone_into_owning_via_copy)
 	a_any.clone_from(erase_type(other));
 	EXPECT_EQ(a_any.as<int>(), other);
 }
+
+TEST(Any, clone_sgvector)
```
lisitsyn: Well actually you don't have to test SGVector here, it might be some simple class (like Simple we have) to speed up compilation and decouple. It is a minor issue tho.

karlnapf: good point

PR #4085, opened 2018-01-16
Replace lapack with linalg in MahalanobisDistance

vinx13:
This PR replaces lapack usage with linalg in `MahalanobisDistance`.
A unit test of `MahalanobisDistance` is added (reference computed with Matlab)

### DIFF
--- a/cmake/FindMetaExamples.cmake
+++ b/cmake/FindMetaExamples.cmake
@@ -35,7 +35,6 @@ function(get_excluded_meta_examples)
 
 	IF(NOT HAVE_LAPACK)
 		LIST(APPEND EXCLUDED_META_EXAMPLES
-			distance/mahalanobis.sg
 			)
 	ENDIF()
 
--- a/src/shogun/distance/MahalanobisDistance.cpp
+++ b/src/shogun/distance/MahalanobisDistance.cpp
@@ -5,7 +5,6 @@
  *          Evan Shelhamer, Sergey Lisitsyn
  */
 
-#ifdef HAVE_LAPACK
 #include <shogun/lib/config.h>
 
 #include <shogun/distance/MahalanobisDistance.h>
@@ -37,21 +36,25 @@ CMahalanobisDistance::~CMahalanobisDistance()
 bool CMahalanobisDistance::init(CFeatures* l, CFeatures* r)
 {
 	CRealDistance::init(l, r);
-
+	SGMatrix<float64_t> cov;
 
 	if ( l == r)
 	{
 		mean = ((CDenseFeatures<float64_t>*) l)->get_mean();
-		icov  = ((CDenseFeatures<float64_t>*) l)->get_cov();
+		cov = ((CDenseFeatures<float64_t>*)l)->get_cov();
 	}
 	else
 	{
 		mean = ((CDenseFeatures<float64_t>*)l)
 		           ->compute_mean((CDotFeatures*)lhs, (CDotFeatures*)rhs);
-		icov = CDotFeatures::compute_cov((CDotFeatures*) lhs, (CDotFeatures*) rhs);
+		cov = CDotFeatures::compute_cov((CDotFeatures*)lhs, (CDotFeatures*)rhs);
 	}
 
-	SGMatrix<float64_t>::inverse(icov);
+	auto num_features = cov.num_rows;
+	chol_cov_L = SGMatrix<float64_t>(num_features, num_features);
+	chol_cov_d = SGVector<float64_t>(num_features);
+	chol_cov_p = SGVector<index_t>(num_features);
+	linalg::ldlt_factor(cov, chol_cov_L, chol_cov_d, chol_cov_p);
 
 	return true;
 }
@@ -82,7 +85,7 @@ float64_t CMahalanobisDistance::compute(int32_t idx_a, int32_t idx_b)
 	for (int32_t i=0; i < diff.vlen; i++)
 		diff[i] = bvec.vector[i] - diff[i];
 
-	auto v = linalg::matrix_prod(icov, diff);
+	auto v = linalg::ldlt_solver(chol_cov_L, chol_cov_d, chol_cov_p, diff);
 	float64_t result = linalg::dot(v, diff);
 
 	if (!use_mean)
@@ -105,4 +108,3 @@ void CMahalanobisDistance::init()
 	m_parameters->add(&use_mean, "use_mean", "If distance shall be computed between mean vector and vector from rhs or between lhs and rhs.");
 }
 
-#endif // HAVE_LAPACK
--- a/src/shogun/distance/MahalanobisDistance.cpp
+++ b/src/shogun/distance/MahalanobisDistance.cpp
@@ -36,18 +36,37 @@ CMahalanobisDistance::~CMahalanobisDistance()
 bool CMahalanobisDistance::init(CFeatures* l, CFeatures* r)
 {
 	CRealDistance::init(l, r);
+
+	REQUIRE(
+	    lhs->get_feature_class() == C_DENSE,
+	    "Left hand side (was %s) has to be CDenseFeatures instance.\n",
+	    lhs->get_name());
+	REQUIRE(
+	    rhs->get_feature_class() == C_DENSE,
+	    "Right hand side (was %s) has to be CDenseFeatures instance.\n",
+	    rhs->get_name());
+
+	REQUIRE(
+	    lhs->get_feature_type() == F_DREAL,
+	    "Left hand side (was %s) has to be of double type.\n", lhs->get_name());
+	REQUIRE(
+	    rhs->get_feature_type() == F_DREAL,
+	    "Right hand side (was %s) has to be double type.\n", rhs->get_name());
+
 	SGMatrix<float64_t> cov;
+
+	auto feat_l = static_cast<CDenseFeatures<float64_t>*>(l);
+	auto feat_r = static_cast<CDenseFeatures<float64_t>*>(r);
 
 	if ( l == r)
 	{
-		mean = ((CDenseFeatures<float64_t>*) l)->get_mean();
-		cov = ((CDenseFeatures<float64_t>*)l)->get_cov();
+		mean = feat_l->get_mean();
+		cov = feat_r->get_cov();
 	}
 	else
 	{
-		mean = ((CDenseFeatures<float64_t>*)l)
-		           ->compute_mean((CDotFeatures*)lhs, (CDotFeatures*)rhs);
-		cov = CDotFeatures::compute_cov((CDotFeatures*)lhs, (CDotFeatures*)rhs);
+		mean = feat_l->compute_mean(feat_l, feat_r);
+		cov = CDotFeatures::compute_cov(feat_l, feat_r);
 	}
 
 	auto num_features = cov.num_rows;
@@ -65,9 +84,10 @@ void CMahalanobisDistance::cleanup()
 
 float64_t CMahalanobisDistance::compute(int32_t idx_a, int32_t idx_b)
 {
+	auto feat_l = static_cast<CDenseFeatures<float64_t>*>(lhs);
+	auto feat_r = static_cast<CDenseFeatures<float64_t>*>(rhs);
 
-	SGVector<float64_t> bvec = ((CDenseFeatures<float64_t>*) rhs)->
-		get_feature_vector(idx_b);
+	SGVector<float64_t> bvec = feat_r->get_feature_vector(idx_b);
 
 	SGVector<float64_t> diff;
 	SGVector<float64_t> avec;
@@ -76,7 +96,7 @@ float64_t CMahalanobisDistance::compute(int32_t idx_a, int32_t idx_b)
 		diff = mean.clone();
 	else
 	{
-		avec = ((CDenseFeatures<float64_t>*) lhs)->get_feature_vector(idx_a);
+		avec = feat_l->get_feature_vector(idx_a);
 		diff=avec.clone();
 	}
 
@@ -86,12 +106,12 @@ float64_t CMahalanobisDistance::compute(int32_t idx_a, int32_t idx_b)
 		diff[i] = bvec.vector[i] - diff[i];
 
 	auto v = linalg::ldlt_solver(chol_cov_L, chol_cov_d, chol_cov_p, diff);
-	float64_t result = linalg::dot(v, diff);
+	auto result = linalg::dot(v, diff);
 
 	if (!use_mean)
-		((CDenseFeatures<float64_t>*) lhs)->free_feature_vector(avec, idx_a);
+		feat_l->free_feature_vector(avec, idx_a);
 
-	((CDenseFeatures<float64_t>*) rhs)->free_feature_vector(bvec, idx_b);
+	feat_r->free_feature_vector(bvec, idx_b);
 
 	if (disable_sqrt)
 		return result;
--- a/src/shogun/distance/MahalanobisDistance.cpp
+++ b/src/shogun/distance/MahalanobisDistance.cpp
@@ -7,14 +7,12 @@
 
 #include <shogun/lib/config.h>
 
-#ifdef HAVE_LAPACK
-
-#include <shogun/lib/common.h>
-#include <shogun/io/SGIO.h>
 #include <shogun/distance/MahalanobisDistance.h>
 #include <shogun/features/Features.h>
+#include <shogun/io/SGIO.h>
+#include <shogun/lib/common.h>
 #include <shogun/mathematics/Math.h>
-#include <shogun/mathematics/lapack.h>
+#include <shogun/mathematics/linalg/LinalgNamespace.h>
 
 using namespace shogun;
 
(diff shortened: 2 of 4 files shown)


karlnapf:
Great!

cmake/FindMetaExamples.cmake
```diff
@@ -19,7 +19,6 @@ function(get_excluded_meta_examples)
     read_config_header(${shogun_INCLUDE_DIR}/shogun/lib/config.h)
     is_set(${SHOGUN_CONFIG} HAVE_NLOPT)
     is_set(${SHOGUN_CONFIG} USE_GPL_SHOGUN)
-    is_set(${SHOGUN_CONFIG} HAVE_LAPACK)
```
karlnapf: we want to keep that, there might be other examples

cmake/FindMetaExamples.cmake
```diff
@@ -37,13 +36,6 @@ function(get_excluded_meta_examples)
 			)
 	ENDIF()
 
-	IF(NOT HAVE_LAPACK)
-		LIST(APPEND EXCLUDED_META_EXAMPLES
```
karlnapf: please keep the block, so we can add things there that pop up in the future, just don't append anything

src/shogun/regression/LinearRidgeRegression.cpp
```diff
@@ -8,7 +8,6 @@
  */
 #include <shogun/lib/config.h>
 
-#ifdef HAVE_LAPACK
```
karlnapf: ah so this guard was just outdated?

I think we should use linalg rather than eigen3 directly inside though

vinx13: Yes, just outdated. I will look into that to find the possibility of using pure linalg.

vinx13:
both `MahalanobisDistance` and `LinearRidgeRegression` are covered by integration tests.

src/shogun/regression/LinearRidgeRegression.cpp
karlnapf: leave it as it is for now (you can already change the rest)
Then you could create this method in linalg, and then send another PR

karlnapf:
I think the LAPACK guard was still necessary.
BTW if you remove such guards, make sure that Shogun compiles if the library that you guarded is not available. Really important.

The windows build doesnt have lapack installed, and here you go:
```
[00:29:53]   C:\projects\shogun\src\shogun\distance\MahalanobisDistance.cpp(56): error C2039: 'inverse': is not a member of 'shogun::SGMatrix<float64_t>' [C:\projects\shogun\build\src\shogun\libshogun.vcxproj]
[00:29:53]   C:\projects\shogun\src\shogun\distance\MahalanobisDistance.cpp(56): error C3861: 'inverse': identifier not found [C:\projects\shogun\build\src\shogun
```

karlnapf:
It is this line:
https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/distance/MahalanobisDistance.cpp#L58

I am sure that can be done with linalg.
But actually this is bad generally. One should never ever store matrix inverse. Instead, it should store a cholesky factorization of the covariance matrix. Then, rather than multiplying the inverse covariance matrix to a vector in the `compute` method, you should do a cholesky (triangular) solve. This is implemented in linalg

src/shogun/regression/LinearRidgeRegression.cpp
karlnapf: Or, you could just add a loop for now.

vinx13:
@karlnapf Since cov matrix is PSD, we should use [LDLT](https://eigen.tuxfamily.org/dox/classEigen_1_1MatrixBase.html#a0ecf058a0727a4cab8b42d79e95072e1) (linalg is using LLT) for Cholesky decomposition, right? Shall I add it to linalg?

karlnapf:
The matrix is PSD indeed, so LDLT would be more stable.
https://eigen.tuxfamily.org/dox/group__TutorialLinearAlgebra.html

I think adding an argument to cholesky would be good, so that the factorization can operate in both modes, something like or `use_ldlt` which by default is false. Also make sure to update the docs appropriately so it is clear what is used.

Good catch!

vinx13:
Seems simply adding a `use_ldlt` option will not work. In LDLT, we want both L and D. One solution is to overload `cholesky_factor` as `SGMatrix<T> cholesky_factor(const SGMatrix<T>& A, SGMatrix<T> &D)`. Since D is diagonal, alternatively we could use a SGVector. Besides, we may need to add a LDLT solver.

karlnapf:
You are absolutely right. In fact I now remember a discussion we previously had about this some years back. I think a clean solution would be to expose a new method and a few solver method as well. Everything else is too messy.

OXPHOS:
We did have some discussion years ago. The best way seems to be having separate methods for `LLT` and `LDLT`. And since now the `linalg` methods are split into several files, having more than one cholesky decomposition or solver would be less messy...

vinx13:
I see. I'll try to implement that.

src/shogun/distance/MahalanobisDistance.h
```diff
@@ -10,44 +10,45 @@
...
+	 * \f]
+	 *
+	 * i.e., instead of the mean as reference two vector \f$x_i\f$ and
+	 * \f$x_i'\f$
+	 * are compared.
+	 *
+	 * @see <a href="http://en.wikipedia.org/wiki/Mahalanobis_distance">
+	 * Wikipedia: Mahalanobis Distance</a>
+	 */
+	class CMahalanobisDistance : public CRealDistance
```
karlnapf: BTW why this reformatting?
Pls don't put this into a PR that changes small bits as this one. Makes everything much harder to review

src/shogun/regression/LinearRidgeRegression.cpp
```diff
@@ -72,28 +68,16 @@ bool CLinearRidgeRegression::train_machine(CFeatures* data)
...
-    Map<MatrixXd> eigen_kernel_matrix(kernel_matrix.matrix, num_feat,num_feat);
-    Map<MatrixXd> eigen_feats_matrix(feats_matrix.matrix, num_feat,num_vec);
-    Map<VectorXd> eigen_y(y.vector, num_feat);
-    Map<VectorXd> eigen_labels(((CRegressionLabels*)m_labels)->get_labels(),num_vec);
-    Map<VectorXd> eigen_tau(tau_vector.vector, num_feat);
-
-    eigen_kernel_matrix = eigen_feats_matrix*eigen_feats_matrix.transpose();
-
-    eigen_kernel_matrix.diagonal() += eigen_tau;
+	linalg::matrix_prod(feats_matrix, feats_matrix, kernel_matrix, false, true);
```
karlnapf: nice much cleaner

src/shogun/regression/LinearRidgeRegression.cpp
```diff
@@ -72,28 +68,16 @@ bool CLinearRidgeRegression::train_machine(CFeatures* data)
...
 
-    LLT<MatrixXd> llt;
-    llt.compute(eigen_kernel_matrix);
-    if(llt.info() != Eigen::Success)
-    {
-    	SG_WARNING("Features covariance matrix was not positive definite\n");
-    	return false;
-    }
-    eigen_y = llt.solve(eigen_y);
+	auto decomposition = linalg::cholesky_factor(kernel_matrix);
```
karlnapf: why did you change the indentation? Is this the style cheker?

karlnapf:
Could you pls split the PR into two: one for regression and one for the distance.
You still haven't checked whether shogun compiles with the guard remove and no lapack installed.
Pls don't remove guards without checking whether things still work: https://ci.appveyor.com/project/vigsterkr/shogun/build/1873#L6478

src/shogun/distance/MahalanobisDistance.h
vinx13: Seems the style checker changed other codes than mine. I'm using command like `git clang-format-3.8 --commit xxx` , is it right ?

src/shogun/regression/LinearRidgeRegression.cpp
vinx13: Just noticed that the original code is indented with spaces.

vinx13:
@karlnapf Created a new PR for `LinearRidgeRegression` (#4149 ). I will wait until #4130 be merged before I can continue with the refactor of `MahalanobisDistance`

karlnapf:
yep, you can just force update this one now

karlnapf:
But you will need to fix the "inverse" problem (see above) first.

src/shogun/distance/MahalanobisDistance.cpp
```diff
@@ -38,21 +36,25 @@ CMahalanobisDistance::~CMahalanobisDistance()
...
 {
 	CRealDistance::init(l, r);
-
+	SGMatrix<float64_t> cov;
 
 	if ( l == r)
 	{
 		mean = ((CDenseFeatures<float64_t>*) l)->get_mean();
-		icov  = ((CDenseFeatures<float64_t>*) l)->get_cov();
+		cov = ((CDenseFeatures<float64_t>*)l)->get_cov();
```
vigsterkr: while you are changing these lines, could you please replace these explicit casts to `dynamic_cast` at least plz? :)
or even better do a `REQUIRED` assertation for the type

src/shogun/distance/MahalanobisDistance.cpp
```diff
@@ -38,21 +36,25 @@ CMahalanobisDistance::~CMahalanobisDistance()
...
 	}
 	else
 	{
 		mean = ((CDenseFeatures<float64_t>*)l)
 		           ->compute_mean((CDotFeatures*)lhs, (CDotFeatures*)rhs);
-		icov = CDotFeatures::compute_cov((CDotFeatures*) lhs, (CDotFeatures*) rhs);
+		cov = CDotFeatures::compute_cov((CDotFeatures*)lhs, (CDotFeatures*)rhs);
```
vigsterkr: another casting

src/shogun/distance/MahalanobisDistance.cpp
```diff
@@ -83,12 +85,8 @@ float64_t CMahalanobisDistance::compute(int32_t idx_a, int32_t idx_b)
...
 		diff[i] = bvec.vector[i] - diff[i];
 
-	SGVector<float64_t> v = diff.clone();
-	cblas_dgemv(CblasColMajor, CblasNoTrans,
-		icov.num_rows, icov.num_cols, 1.0, icov.matrix,
-		diff.vlen, diff.vector, 1, 0.0, v.vector, 1);
-
-	float64_t result = cblas_ddot(v.vlen, v.vector, 1, diff.vector, 1);
+	auto v = linalg::ldlt_solver(chol_cov_L, chol_cov_d, chol_cov_p, diff);
+	float64_t result = linalg::dot(v, diff);
```
vigsterkr: this could be auto or? :)

src/shogun/distance/MahalanobisDistance.cpp
```diff
@@ -39,20 +37,43 @@ bool CMahalanobisDistance::init(CFeatures* l, CFeatures* r)
 {
 	CRealDistance::init(l, r);
 
+	REQUIRE(
+	    lhs->get_feature_class() == C_DENSE,
+	    "Left hand side (was %s) has to be CDenseFeatures instance.\n",
+	    lhs->get_name());
```
karlnapf: this segfaults if one of the features is not set (or does the above call check that?)

karlnapf: Also: "Left hand side feature type (%s) has be be (%s)\n"
One should not use type names in the messages (think refactor), would be better to have a (static) method CDenseFeatures::get_name(). I'll send a patch and you can use that

vigsterkr: https://github.com/vinx13/shogun/blob/39e7dd2f390d13e665678be23b0b0221f1212033/src/shogun/distance/Distance.cpp#L80

vigsterkr: but the dense checks can be dropped as they are implicitly tested in `CRealDistance::init` and belove

src/shogun/distance/MahalanobisDistance.cpp
```diff
@@ -39,20 +37,43 @@ bool CMahalanobisDistance::init(CFeatures* l, CFeatures* r)
...
+	REQUIRE(
+	    rhs->get_feature_class() == C_DENSE,
+	    "Right hand side (was %s) has to be CDenseFeatures instance.\n",
+	    rhs->get_name());
+
+	REQUIRE(
+	    lhs->get_feature_type() == F_DREAL,
+	    "Left hand side (was %s) has to be of double type.\n", lhs->get_name());
```
karlnapf: "Left hand side feature type (%d) has to be REAL"

src/shogun/distance/MahalanobisDistance.cpp
```diff
@@ -39,20 +37,43 @@ bool CMahalanobisDistance::init(CFeatures* l, CFeatures* r)
 {
 	CRealDistance::init(l, r);
```
karlnapf: this call covers all the checks, so we don't need them again in this call (just convolutes the code)
@vigsterkr both REAL and DENSE is checked. That was the whole point of having the base class here iirc

karlnapf: I would do a cast after the call and add a comment

src/shogun/distance/MahalanobisDistance.cpp
```diff
@@ -39,20 +37,43 @@ bool CMahalanobisDistance::init(CFeatures* l, CFeatures* r)
...
+	REQUIRE(
+	    lhs->get_feature_class() == C_DENSE,
+	    "Left hand side (was %s) has to be CDenseFeatures instance.\n",
+	    lhs->get_name());
+	REQUIRE(
+	    rhs->get_feature_class() == C_DENSE,
+	    "Right hand side (was %s) has to be CDenseFeatures instance.\n",
+	    rhs->get_name());
+
+	REQUIRE(
```
vigsterkr: and actually this is implicitly tested in `CRealDistance::init`
https://github.com/vinx13/shogun/blob/39e7dd2f390d13e665678be23b0b0221f1212033/src/shogun/distance/RealDistance.h#L34

src/shogun/distance/MahalanobisDistance.cpp
```diff
@@ -63,9 +84,10 @@ void CMahalanobisDistance::cleanup()
 
 float64_t CMahalanobisDistance::compute(int32_t idx_a, int32_t idx_b)
 {
+	auto feat_l = static_cast<CDenseFeatures<float64_t>*>(lhs);
```
karlnapf: this is reliable due to old code design principles by sören

src/shogun/distance/MahalanobisDistance.cpp
vigsterkr: instead of the REQUIRED sections it should be just check the return type of `CRealDistance::init(l, r)` as it checks for everything you need want....

src/shogun/distance/MahalanobisDistance.cpp
```diff
@@ -83,17 +105,13 @@ float64_t CMahalanobisDistance::compute(int32_t idx_a, int32_t idx_b)
 	for (int32_t i=0; i < diff.vlen; i++)
 		diff[i] = bvec.vector[i] - diff[i];
 
-	SGVector<float64_t> v = diff.clone();
-	cblas_dgemv(CblasColMajor, CblasNoTrans,
-		icov.num_rows, icov.num_cols, 1.0, icov.matrix,
-		diff.vlen, diff.vector, 1, 0.0, v.vector, 1);
-
-	float64_t result = cblas_ddot(v.vlen, v.vector, 1, diff.vector, 1);
+	auto v = linalg::ldlt_solver(chol_cov_L, chol_cov_d, chol_cov_p, diff);
```
karlnapf: yeah this is much much from a numerical perspective!

src/shogun/distance/MahalanobisDistance.cpp
karlnapf: haha I wrote the same thing above ;) sorry didnt see

src/shogun/distance/MahalanobisDistance.cpp
vigsterkr: i would actually change the design of this and have instead of init returning a boolean the actual typecasted values available... as i dont see really the point of an assertation failure that will just throw an exception and in case of success have a true :)
this function call should actually tell us that yes the types are cool and here you go this is what you wanted anyways

vigsterkr: @vinx13 you can ignore most of this... just make sure that you do an ASSERT(CRealDistance::init(l, r)); and add a //FIXME: above with reference to the above comment

karlnapf: @vigsterkr I mean this whole type system in the features is messed up. I would just rely on `dynamic_cast` everywhere. The bool is also pointless. And the enums are even worse (too many types, I would make those strings or even better drop them)

vinx13:
@karlnapf could you plz take a look

src/shogun/distance/MahalanobisDistance.cpp
```diff
@@ -37,22 +35,31 @@ CMahalanobisDistance::~CMahalanobisDistance()
 
 bool CMahalanobisDistance::init(CFeatures* l, CFeatures* r)
 {
-	CRealDistance::init(l, r);
+	// FIXME: See comments in
```
karlnapf: fixme?

PR #4087, opened 2018-01-18
Add linalg API to allow adding a vector to diagonal

vinx13:
As mentioned in #4085 by @karlnapf , we need a API in linalg to allow adding a vector to a matrix's diagonal. This PR adds `add_diag_vec` that performs operation `A.diagonal = alpha * A.diagonal + beta * b`. Could you provide some feedback on the API design ?

### DIFF
--- a/src/shogun/mathematics/linalg/LinalgBackendBase.h
+++ b/src/shogun/mathematics/linalg/LinalgBackendBase.h
@@ -124,6 +124,23 @@ namespace shogun
 #undef BACKEND_GENERIC_ADD_COL_VEC
 
 /**
+ * Wrapper method of add diagonal vector A.diagonal = alpha * A.diagonal +
+ * beta * b.
+ *
+ * @see linalg::add_diag
+ */
+#define BACKEND_GENERIC_ADD_DIAG(Type, Container)                              \
+	virtual void add_diag(                                                     \
+	    SGMatrix<Type>& A, const SGVector<Type>& b, Type alpha, Type beta)     \
+	    const                                                                  \
+	{                                                                          \
+		SG_SNOTIMPLEMENTED;                                                    \
+		return;                                                                \
+	}
+		DEFINE_FOR_ALL_PTYPE(BACKEND_GENERIC_ADD_DIAG, SGMatrix);
+#undef BACKEND_GENERIC_ADD_DIAG
+
+/**
  * Wrapper method of add vector to each column of matrix.
  *
  * @see linalg::add_vector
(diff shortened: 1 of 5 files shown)


tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
```diff
@@ -189,6 +190,35 @@ TEST(LinalgBackendEigen, SGMatrix_add_col_vec_in_place)
 		}
 }
 
+TEST(LinalgBackendEigen, add_diag_vec)
+{
+	const float64_t alpha = 0.7;
```
karlnapf: could you do this test more explicit, i.e. just write down the matrices by hand (no loops)
Otherwise good.

Also make sure to use `EXPECT_THROW` for the case of invalid input arugments (the `REQUIRE`)

Finally, test some corner cases (empty matrix doesnt give memory error, etc)

src/shogun/mathematics/linalg/LinalgBackendBase.h
```diff
@@ -123,6 +123,23 @@ namespace shogun
 		DEFINE_FOR_ALL_PTYPE(BACKEND_GENERIC_ADD_COL_VEC, SGMatrix)
 #undef BACKEND_GENERIC_ADD_COL_VEC
 
+/**
+ * Wrapper method of add diagonal vector A.diagonal = alpha * A.diagonal +
+ * beta * b.
+ *
+ * @see linalg::add_diag_vec
+ */
+#define BACKEND_GENERIC_ADD_DIAG_VEC(Type, Container)                          \
```
karlnapf: I guess I would call this `add_diag` rather than `add_diag_vec`

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
```diff
@@ -189,6 +190,43 @@ TEST(LinalgBackendEigen, SGMatrix_add_col_vec_in_place)
...
+{
+	SGMatrix<float64_t> A1(2, 2);
+	SGVector<float64_t> b1(3);
+	SGMatrix<float64_t> A2;
+	SGVector<float64_t> b2;
+	EXPECT_THROW(add_diag(A1, b1), ShogunException);
+	EXPECT_THROW(add_diag(A1, b2), ShogunException);
+	EXPECT_THROW(add_diag(A2, b1), ShogunException);
+	EXPECT_THROW(add_diag(A2, b2), ShogunException);
+}
```
OXPHOS: IMHO, I'd shove all `add_diag`-related tests into one test unit for the neatness.
Otherwise good to me :)

karlnapf:
Good work, thanks!

PR #4096, opened 2018-01-21
Implement array reference with any

lisitsyn:


src/shogun/lib/any.h
```diff
@@ -80,6 +81,37 @@ namespace shogun
...
+		{
+			if (*(m_length) != *(other.m_length))
+			{
+				return false;
+			}
+			if (*(m_ptr) == *(other.m_ptr))
+			{
+				return true;
+			}
+			return std::equal(*(m_ptr), *(m_ptr) + *(m_length), *(other.m_ptr));
```
karlnapf: needs to be specialized for floats I guess?

karlnapf: In fact need to check whether the elements support ".equals" or "->equals"

src/shogun/lib/any.h
```diff
@@ -239,6 +262,22 @@ namespace shogun
...
+		{
+			return false;
+		}
+		if (*(m_ptr) == *(other.m_ptr))
+		{
+			return true;
+		}
+		return std::equal(
+		    *(m_ptr), *(m_ptr) + *(m_length), *(other.m_ptr),
+		    [](T lhs, T rhs) -> bool { return any_detail::compare(lhs, rhs); });
```
karlnapf: this is pretty cool!

PR #4099, opened 2018-01-22
systematically test CSGObject::equals

karlnapf:


lisitsyn:
Looks very good to me :+1:

PR #4100, opened 2018-01-22
tag-register raw array

karlnapf:
to make `::CSGObject::equals` work


src/shogun/features/StringFeatures.cpp
```diff
@@ -1730,12 +1730,17 @@ template<class ST> void CStringFeatures<ST>::init()
 	original_num_symbols=0;
 
 	m_parameters->add((CSGObject**) &alphabet, "alphabet");
+
 	m_parameters->add_vector(&features, &num_vectors, "features",
```
karlnapf: every raw array, which was previously registered with `add_vector` should also be registered with `watch_param`

PR #4112, opened 2018-01-23
Add helper function to retrieve available objects

lisitsyn:


src/shogun/base/class_list.cpp.templ
```diff
@@ -74,3 +74,14 @@ void shogun::delete_object(CSGObject* object)
 {
 	delete object;
 }
+
+std::set<std::string> available_objects()
```
karlnapf: is this exposed to swig?

lisitsyn: I am not sure if it works. But would work if we include std_set.i

Issue #4110, opened 2018-01-23
tag-register all raw matrices


karlnapf:
This is just like #4101 but for `add_matrix`

See 79d91811fda816d116d150840fad647eda50c37e for an example.

nggih:
Hey, I would like to work on this, but I am new here, so I might ask a lot of questions

nggih:
Is this right? 
For example: shogun/src/shogun/distance/AttenuatedEuclideanDistance.cpp
this code below needs to be changed:
`void CAttenuatedEuclideanDistance::init()
{
	disable_sqrt=false;

	m_parameters->add(&disable_sqrt, "disable_sqrt", "If sqrt shall not be applied.");
}`

Into this :
`void CAttenuatedEuclideanDistance::init()
{
	disable_sqrt=false;

	watch_param("disable_sqrt", &disable_sqrt);
}`

PR #4119, opened 2018-01-25
Tag-register all raw matrices

guruhegde:
Made changes as suggested in #4110 

- [x] pass style check.
- [x] pass all tests.


karlnapf:
Great, thanks!

Issue #4117, opened 2018-01-24
Make sure tag parameters cover exactly all old registered parameters


karlnapf:

syashakash:
I am taking this up.

guruhegde:
@syashakash Hi, I started addressing this issue in #4123. sorry for the confusion:)

PR #4116, opened 2018-01-24
tag-register all raw arrays

guruhegde:
Made changes as suggested in #4101 

All tests pass.

PR #4123, opened 2018-01-26
Register parameters in tags (#4117)

guruhegde:
According to SGObjectAll test case, registered parameters in tags to match old parameters for all classes. 

Exceptions:
For class LatentSOSVM & SVMOcas, old parameters and tag parameters don't match yet.(I am not able to find those classes, please help me find those). 

@karlnapf There are many classes in this PR where static_cast on parameters preset. you asked in PR #4116 not add watch_param call & to comment add tags call in separate PR.
Would you like me to do this all together in separate PR? (I am curious what is the exact reason for that suggestion)


src/shogun/classifier/svm/SVMSGD.cpp
```diff
@@ -214,13 +214,32 @@ void CSVMSGD::init()
...
+	watch_param("epochs", &epochs);
+
+	m_parameters->add(&skip, "skip", "skip");
+	watch_param("skip", &skip);
+
+	m_parameters->add(&count, "count", "count");
+	watch_param("count", &count);
+
+	m_parameters->add(&use_bias, "use_bias", "Indicates if bias is used.");
+	watch_param("use_bias", &use_bias);
```
karlnapf: Actually, we just realised that we probably want to keep the comment as well.
But that requires a minor change to the registration method .... Let me cook something up and ping you

karlnapf:
@lisitsyn Would you mind adding the possibility to add a comment in the parameters?
I would like to have that in before this one is merged so we don't have to touch everything again.
I am travelling now for a week, so cannot do it myself ... back Thursday and will continue to look into it then.

lisitsyn:
@karlnapf yes, will add that.

lisitsyn:
In #4132 I am adding parameter description. Lets rebase once it is merged and add these descriptions here as well. Otherwise the descriptions would get lost one day.

guruhegde:
Sure. I will add descriptions in this PR.

lisitsyn:
@guruhegde I just merged #4132 so you can rebase and add descriptions.

guruhegde:
I added the description of scalar parameters.  I will send another PR to add the description for vector and matrix type.

guruhegde:
Remark: I ran tag parameter coverage test. Equality check with old parameter and tag parameter fails even though both lists have all & same parameters, this is because of the different order of parameter names in two lists. The solution I tried, sort old parameter names, which works.

karlnapf:
@guruhegde yes we need to sort. It used to be a set, but was recently changed to a vector....thats the danger of using the `auto` keyword to heavily :) Will fix. Thanks!

src/shogun/classifier/PluginEstimate.cpp
```diff
@@ -22,16 +22,43 @@ CPluginEstimate::CPluginEstimate(float64_t pos_pseudo, float64_t neg_pseudo)
 {
 	m_parameters->add(&m_pos_pseudo,
 			"pos_pseudo","pseudo count for positive class");
+	watch_param(
+	    "pos_pseudo", &m_pos_pseudo,
+	    AnyParameterProperties(
+	        "pseudo count for positive class", MS_NOT_AVAILABLE,
```
karlnapf: @guruhegde  I have one more request to keep this cleaner:
Can you make the second two parameter of the `AnyParameterProperties` constructor have default values set to false.
And then don't put `MS_NOT_AVAILABLE` every time, but only the descriptions ...

karlnapf: This way, things will look way more neat

guruhegde: @karlnapf ok. That looks lot cleaner. Will modify and update.

would you like me to send a PR to set default argument in existing constructor of AnyParameterProperties? or Make new constructor(like you commented in #4132)

I guess this PR can wait till above-mentioned changes get merged.

karlnapf: Do a separate PR since that is very small and minimal. Then rebase this one here

guruhegde:
@karlnapf Hi this is ready for review:)

karlnapf:
Thanks!
We unfortunately still need to hold this for a while as we need to figure out whether we want to do the downcasting to `CSGObject*` or not. We are in an ongoing discussion about this in IRC. The PR is really appreciated and we will merge it, but give us some time :)

gf712:
@karlnapf isn't this what we already do?

karlnapf:
I think this is probably still useful? Rebase and merge?

PR #4126, opened 2018-01-26
Updated INSTALL.md

yashdusing:
The 'git submodule update --init' command works only after changing directory to shogun before actually running the command


karlnapf:
Thanks! appreciated! :)

Issue #3612, opened 2017-01-31
Stratified Cross Validation with Combined Kernels using Custom Kernels


github-ssi:
Hello,

I use libshogun 5.1.0 with gcc 4.9.2 on Debian 8, and have the following issue: I am training a CMKLMulticlass machine using a CCombinedKernel. This combined kernel contains several linear/Gaussian kernels and one CCustomKernel which I am constructing from an SGMatrix.

I have 100 feature vectors. Hence the Kernel Matrix of my custom kernel is 100x100, and an initial training on the whole dataset (mkl->train()) works out fine.

Problem is, when I use a 2-fold CStratifiedCrossValidationSplitting, the cross->evaluate() call fails with the following error:

In file /home/berti/src/shogun/shogun/src/shogun/kernel/CombinedKernel.cpp line 169:
  Number of lhs-feature vectors (50) not match with number of rows (100) of custom kernel

I understand that my cross validation uses a set of 50 features per run. The other sub-kernels are implicitly initialized using init() to the size of 50x50, but my custom kernel doesn't have an init() method, or at least such a method is not called from CombinedKernel::init(). Hence the error.

Do you see any chance to use a combined kernel containing a custom kernel with CCrossValidation?

Thank you very much!

Best regards
Bert

vigsterkr:
this should definitely work, we'll look into it!

karlnapf:
Hi Bert,

can you post some code that lets us reproduce your error?
Thanks!

github-ssi:
Hi karlnapf,

the following code just generates a combined kernel from one custom kernel (from SGMatrix) and one Gaussian kernel. MKL is used on 10 input vectors, 10 dimensions each. Training on the full dataset works, but cross validation with 2 folds produces the following error:

    [DEBUG] entering CKernel::remove_lhs_and_rhs
    [DEBUG] leaving CKernel::remove_lhs_and_rhs
    [DEBUG] entering CKernel::remove_lhs_and_rhs
    [DEBUG] leaving CKernel::remove_lhs_and_rhs
    [DEBUG] leaving CKernel::init(0x6b7f00, 0x6b7f00)
    [DEBUG] Starting for loop for kernels
    [DEBUG] Initializing 0x0x6ba250 - "CustomKernel"
    [DEBUG] entering CKernel::remove_lhs_and_rhs
    [DEBUG] leaving CKernel::remove_lhs_and_rhs
[...]

PR #4127, opened 2018-01-27
Fix issue #3612

dhelekal:
This PR addresses issue #3612, by allowing CombinedKernel init to properly subset CustomKernel subkernels. Hence CustomKernel should now work with mkl / x-validation

Test code used to verify the fix:
```
from shogun import *
from numpy import *

init_shogun_with_defaults()

feats_train=RealFeatures(random.rand(10,10))  
labels=BinaryLabels(array([1,1,1,1,1,-1,-1,-1,-1,-1]))   

kerg = GaussianKernel(2.0)

kerg.init(feats_train,feats_train)

ker1 = CustomKernel(kerg.get_kernel_matrix())
ker2 = GaussianKernel(2.0)

cker = CombinedKernel()

cker.append_kernel(ker1)
cker.append_kernel(ker2)

cft = CombinedFeatures()

cft.append_feature_obj(feats_train)
cft.append_feature_obj(feats_train)

cker.init(cft,cft)

libsvm = LibSVM()
svm = MKLClassification(libsvm)
svm.set_interleaved_optimization_enabled(False)
svm.set_kernel(cker)
svm.set_C_mkl(1000)
svm.set_mkl_norm(1)

splitting_strategy = StratifiedCrossValidationSplitting(labels, 3)
evaluation_criterium = AccuracyMeasure()
cross = CrossValidation(svm, cft, labels, splitting_strategy, evaluation_criterium)
cross.set_autolock(False)
cross.set_num_runs(2)

mkl_obs = ParameterObserverCV(True)
cross.subscribe_to_parameters(mkl_obs)

result = CrossValidationResult()
result = CrossValidationResult.obtain_from_generic(cross.evaluate())

obs = mkl_obs.get_observation(0)
fold = obs.get_fold(0)
machine = MKLClassification.obtain_from_generic(fold.get_trained_machine())

k = CombinedKernel.obtain_from_generic(machine.get_kernel())
k1 = k.get_kernel(0)
k2 = k.get_kernel(1)

print(k1.get_kernel_matrix())
print("\n")
print(k2.get_kernel_matrix())
print("\n")
print(k1.get_kernel_matrix()-k2.get_kernel_matrix())
```

Sample output for the test  to verify subsetting works properly:
`print(k1.get_kernel_matrix())`
```
[[ 1.          0.50979882  0.49804497  0.28813449  0.27958548  0.30324122]
 [ 0.50979882  1.          0.38971996  0.27236992  0.29889661  0.40002599]
 [ 0.49804497  0.38971996  1.          0.25557584  0.49059519  0.41748011]
 [ 0.28813449  0.27236992  0.25557584  1.          0.48876277  0.16951311]
 [ 0.27958548  0.29889661  0.49059519  0.48876277  1.          0.53417933]
 [ 0.30324122  0.40002599  0.41748011  0.16951311  0.53417933  1.        ]]
```
`print(k2.get_kernel_matrix())`
```
[[ 1.          0.50979884  0.49804496  0.28813448  0.27958548  0.30324122]
 [ 0.50979884  1.          0.38971996  0.27236991  0.29889661  0.400026  ]
 [ 0.49804496  0.38971996  1.          0.25557583  0.49059518  0.41748012]
 [ 0.28813448  0.27236991  0.25557583  1.          0.48876277  0.1695131 ]
 [ 0.27958548  0.29889661  0.49059518  0.48876277  1.          0.53417934]
 [ 0.30324122  0.400026    0.41748012  0.1695131   0.53417934  1.        ]]
```
`print(k1.get_kernel_matrix()-k2.get_kernel_matrix())`
```
[[  0.00000000e+00  -1.68008413e-08   6.80469564e-09   9.58425667e-09
   -2.85334684e-09   6.21391272e-11]
 [ -1.68008413e-08   0.00000000e+00   5.66272729e-09   1.40209140e-08
    3.09241743e-09  -3.52935375e-09]
 [  6.80469564e-09   5.66272729e-09   0.00000000e+00   3.50433216e-09
    1.25876629e-08  -1.27479330e-08]
 [  9.58425667e-09   1.40209140e-08   3.50433216e-09   0.00000000e+00
   -6.48492471e-09   6.75771872e-09]
 [ -2.85334684e-09   3.09241743e-09   1.25876629e-08  -6.48492471e-09
    0.00000000e+00  -9.37992795e-09]
 [  6.21391272e-11  -3.52935375e-09  -1.27479330e-08   6.75771872e-09
   -9.37992795e-09   0.00000000e+00]]
```

### DIFF
--- a/src/shogun/kernel/CombinedKernel.cpp
+++ b/src/shogun/kernel/CombinedKernel.cpp
@@ -148,7 +148,7 @@ bool CCombinedKernel::init(CFeatures* l, CFeatures* r)
 			if (!k->has_features())
 				SG_ERROR("No kernel matrix was assigned to this Custom kernel\n")
 
-			//extract row & column subset indices from combined features
+			// extract row & column subset indices from combined features
 			auto lhs_subset_stack = ((CCombinedFeatures*)l)->get_subset_stack();
 			auto rhs_subset_stack = ((CCombinedFeatures*)r)->get_subset_stack();
 
--- a/src/shogun/kernel/CombinedKernel.cpp
+++ b/src/shogun/kernel/CombinedKernel.cpp
@@ -111,8 +111,10 @@ bool CCombinedKernel::init_with_extracted_subsets(
 				SG_UNREF(lf);
 				SG_UNREF(rf);
 				SG_UNREF(k);
-				SG_ERROR("CombinedKernel: Number of features/kernels does not "
-				         "match - bailing out\n")
+				SG_ERROR(
+				    "CombinedKernel: Number of features/kernels does not "
+				    "match - bailing out\n")
+
 			}
 
 			SG_DEBUG("Initializing 0x%p - \"%s\"\n", this, k->get_name())
@@ -173,8 +175,9 @@ bool CCombinedKernel::init_with_extracted_subsets(
 	if (((CCombinedFeatures*)l)->get_num_feature_obj() <= 0 ||
 	    ((CCombinedFeatures*)l)->get_num_feature_obj() !=
 	        ((CCombinedFeatures*)r)->get_num_feature_obj())
-		SG_ERROR("CombinedKernel: Number of features/kernels does not match - "
-		         "bailing out\n")
+		SG_ERROR(
+		    "CombinedKernel: Number of features/kernels does not match - "
+		    "bailing out\n")
 
 	init_normalizer();
 	initialized = true;
@@ -234,8 +237,9 @@ bool CCombinedKernel::init(CFeatures* l, CFeatures* r)
 	    l->get_feature_type() == r->get_feature_type() &&
 	    l->get_feature_class() != C_COMBINED)
 	{
-		SG_DEBUG("Initialising combined kernel's combined features with the "
-		         "same instance from parameters\n");
+		SG_DEBUG(
+		    "Initialising combined kernel's combined features with the "
+		    "same instance from parameters\n");
 		/* construct combined features with each element being the parameter
 		 * The we must make sure that we make any custom kernels aware of any
 		 * subsets present!
--- a/src/shogun/kernel/CombinedKernel.cpp
+++ b/src/shogun/kernel/CombinedKernel.cpp
@@ -61,108 +61,124 @@ void CCombinedKernel::init_subkernel_weights()
 	eigen_wt = eigen_wt.array().exp();
 	set_subkernel_weights(wt);
 }
-bool CCombinedKernel::init_with_extracted_subsets(CFeatures *l, CFeatures *r, SGVector<index_t> lhs_subset,
-                                                  SGVector<index_t> rhs_subset) {
+bool CCombinedKernel::init_with_extracted_subsets(
+    CFeatures* l, CFeatures* r, SGVector<index_t> lhs_subset,
+    SGVector<index_t> rhs_subset)
+{
 
+	CKernel::init(l, r);
+	REQUIRE(
+	    l->get_feature_class() == C_COMBINED,
+	    "%s::init(): LHS features are"
+	    " of class %s but need to be combined features!\n",
+	    get_name(), l->get_name());
+	REQUIRE(
+	    r->get_feature_class() == C_COMBINED,
+	    "%s::init(): RHS features are"
+	    " of class %s but need to be combined features!\n",
+	    get_name(), r->get_name());
+	ASSERT(l->get_feature_type() == F_UNKNOWN)
+	ASSERT(r->get_feature_type() == F_UNKNOWN)
 
-    CKernel::init(l,r);
-    REQUIRE(l->get_feature_class()==C_COMBINED, "%s::init(): LHS features are"
-            " of class %s but need to be combined features!\n",
-            get_name(), l->get_name());
-    REQUIRE(r->get_feature_class()==C_COMBINED, "%s::init(): RHS features are"
-            " of class %s but need to be combined features!\n",
-            get_name(), r->get_name());
-    ASSERT(l->get_feature_type()==F_UNKNOWN)
-    ASSERT(r->get_feature_type()==F_UNKNOWN)
+	CFeatures* lf = NULL;
+	CFeatures* rf = NULL;
+	CKernel* k = NULL;
 
-    CFeatures* lf=NULL;
-    CFeatures* rf=NULL;
-    CKernel* k=NULL;
+	bool result = true;
+	index_t f_idx = 0;
 
-    bool result=true;
-    index_t f_idx = 0;
+	SG_DEBUG("Starting for loop for kernels\n")
+	for (index_t k_idx = 0; k_idx < get_num_kernels() && result; k_idx++)
+	{
+		k = get_kernel(k_idx);
 
-    SG_DEBUG("Starting for loop for kernels\n")
-    for (index_t k_idx=0; k_idx<get_num_kernels() && result; k_idx++)
-    {
-        k = get_kernel(k_idx);
+		if (!k)
+			SG_ERROR("Kernel at position %d is NULL\n", k_idx);
 
-        if (!k)
-        SG_ERROR("Kernel at position %d is NULL\n", k_idx);
+		// skip over features - the custom kernel does not need any
+		if (k->get_kernel_type() != K_CUSTOM)
+		{
+			if (((CCombinedFeatures*)l)->get_num_feature_obj() > f_idx &&
+			    ((CCombinedFeatures*)r)->get_num_feature_obj() > f_idx)
+			{
+				lf = ((CCombinedFeatures*)l)->get_feature_obj(f_idx);
+				rf = ((CCombinedFeatures*)r)->get_feature_obj(f_idx);
+			}
 
-        // skip over features - the custom kernel does not need any
-        if (k->get_kernel_type() != K_CUSTOM)
-        {
-            if (((CCombinedFeatures*)l)->get_num_feature_obj() > f_idx &&
-                ((CCombinedFeatures*)r)->get_num_feature_obj() > f_idx)
-            {
-                lf = ((CCombinedFeatures*)l)->get_feature_obj(f_idx);
-                rf = ((CCombinedFeatures*)r)->get_feature_obj(f_idx);
-            }
+			f_idx++;
+			if (!lf || !rf)
+			{
+				SG_UNREF(lf);
+				SG_UNREF(rf);
+				SG_UNREF(k);
+				SG_ERROR("CombinedKernel: Number of features/kernels does not "
+				         "match - bailing out\n")
+			}
 
-            f_idx++;
-            if (!lf || !rf)
-            {
-                SG_UNREF(lf);
-                SG_UNREF(rf);
-                SG_UNREF(k);
-                SG_ERROR("CombinedKernel: Number of features/kernels does not match - bailing out\n")
-            }
+			SG_DEBUG("Initializing 0x%p - \"%s\"\n", this, k->get_name())
+			result = k->init(lf, rf);
+			SG_UNREF(lf);
+			SG_UNREF(rf);
 
-            SG_DEBUG("Initializing 0x%p - \"%s\"\n", this, k->get_name())
-            result=k->init(lf,rf);
-            SG_UNREF(lf);
-            SG_UNREF(rf);
+			if (!result)
+				break;
+		}
+		else
+		{
+			SG_DEBUG(
+			    "Initializing 0x%p - \"%s\" (skipping init, this is a CUSTOM "
+			    "kernel)\n",
+			    this, k->get_name())
+			if (!k->has_features())
+				SG_ERROR(
+				    "No kernel matrix was assigned to this Custom kernel\n")
 
-            if (!result)
-                break;
-        }
-        else
-        {
-            SG_DEBUG("Initializing 0x%p - \"%s\" (skipping init, this is a CUSTOM kernel)\n", this, k->get_name())
-            if (!k->has_features())
-            SG_ERROR("No kernel matrix was assigned to this Custom kernel\n")
+			// clear all previous subsets
+			((CCustomKernel*)k)->remove_all_row_subsets();
+			// apply new subset
+			((CCustomKernel*)k)->add_row_subset(lhs_subset);
 
+			((CCustomKernel*)k)->remove_all_col_subsets();
+			// apply new subset
+			((CCustomKernel*)k)->add_col_subset(rhs_subset);
 
-                // clear all previous subsets
-                ((CCustomKernel*)k)->remove_all_row_subsets();
-                // apply new subset
-                ((CCustomKernel*)k)->add_row_subset(lhs_subset);
+			if (k->get_num_vec_lhs() != num_lhs)
+				SG_ERROR(
+				    "Number of lhs-feature vectors (%d) not match with number "
+				    "of rows (%d) of custom kernel\n",
+				    num_lhs, k->get_num_vec_lhs())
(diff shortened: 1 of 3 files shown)


src/shogun/kernel/CombinedKernel.cpp
```diff
@@ -147,6 +147,31 @@ bool CCombinedKernel::init(CFeatures* l, CFeatures* r)
...
+				// apply new subset
+				((CCustomKernel*)k)->add_row_subset(lhs_subset);
+			}
+
+			if (rhs_subset_stack->has_subsets())
+			{
+				auto rhs_subset =
+				    rhs_subset_stack->get_last_subset()->get_subset_idx();
+				// clear all previous subsets
+				((CCustomKernel*)k)->remove_all_col_subsets();
```
karlnapf: I think you shouldn’t clear the subsets here.
This should happen from outside (and only the last one) as otherwise you might remove stuff a user has set previously

karlnapf:
A minimal unit test that checks the matrices as in your output would be super helpful

src/shogun/kernel/CombinedKernel.cpp
dhelekal: I operated under the assumption that the subset should be maintained in sync with the combinedfeatures used to init the combinedkernel. Normally, as per my understanding, the subset for the subkernels is obtained from the features used to init them. (The init(features l, features r) method of custom kernel in fact clears the subsets).
As the custom kernel has no features in the 1st place, I attempted to replicate what an init() would do if it had features.
Alternative approach: I could instantiate CIndexFeatures using lhs_subset/rhs_subset, then init the custom kernel using these. This might actually be a better approach, on a second though

dhelekal: Update: There seems to be an inconsistency within `CCustomKernel::init(CFeatures* l, CFeatures* r)`.
Namely, if l & r are of type CIndexFeatures the normalizer doesnt get called, see following block
```
	if (l->get_feature_class()==C_INDEX && r->get_feature_class()==C_INDEX)
	{
		CIndexFeatures* l_idx = (CIndexFeatures*)l;
		CIndexFeatures* r_idx = (CIndexFeatures*)r;

		remove_all_col_subsets();
		remove_all_row_subsets();

		add_row_subset(l_idx->get_feature_index());
		add_col_subset(r_idx->get_feature_index());

		lhs_equals_rhs=m_is_symmetric;

		return true;
	}

	/* For other types of CFeatures do the default actions below */
	CKernel::init(l, r);

	lhs_equals_rhs=m_is_symmetric;

	SG_DEBUG("num_vec_lhs: %d vs num_rows %d\n", l->get_num_vectors(), kmatrix.num_rows)
	SG_DEBUG("num_vec_rhs: %d vs num_cols %d\n", r->get_num_vectors(), kmatrix.num_cols)
	ASSERT(l->get_num_vectors()==kmatrix.num_rows)
	ASSERT(r->get_num_vectors()==kmatrix.num_cols)
	return init_normalizer();
```
Is this intentional or not?

karlnapf: Most things in there were written in demand if one specific use case, so probably no.

I think everything should be consistent between normal kernels and custom kernels.

So maybe we keep it like this.

Memory check is important before merging though

dhelekal: I'm not 100% certain what do you mean re: consistency; could you please clarify it to me, perhaps on irc?

karlnapf: Sorry. I meant do it the same way for custom kernels as for normal kernels (clearing the stack)

dhelekal: OK, so keep the current method, rather than providing dummy indexfeatures, and calling the customkernels init on those? (+ and fix the memory issues)

dhelekal:
Update; the fix turned out to be a bit larger than expected; mainly because the added unit test revealed subsetting combined kernel wouldnt behave as expected if custom kernels were contained, and the features passed during init were of other typed than combined features. In order to fix this it was necessary to make the method aware of any subsets present. Hence instead of the past recursive call, there's a call to a helper method which now contains the bulk of the old init, and in addition taking the subsets as arguments.

src/shogun/kernel/CombinedKernel.cpp
```diff
@@ -144,13 +126,33 @@ bool CCombinedKernel::init(CFeatures* l, CFeatures* r)
...
+			    "Initializing 0x%p - \"%s\" (skipping init, this is a CUSTOM "
+			    "kernel)\n",
+			    this, k->get_name())
 			if (!k->has_features())
-				SG_ERROR("No kernel matrix was assigned to this Custom kernel\n")
+				SG_ERROR(
+				    "No kernel matrix was assigned to this Custom kernel\n")
+
+			// clear all previous subsets
+			((CCustomKernel*)k)->remove_all_row_subsets();
```
vigsterkr: instead of the explicit cast could we do a dynamic_cast here so that way we can clearly get a bad_cast exception in case of some weird `k`?

src/shogun/kernel/CombinedKernel.cpp
```diff
@@ -169,15 +171,100 @@ bool CCombinedKernel::init(CFeatures* l, CFeatures* r)
 		return false;
 	}
 
-	if ( ((CCombinedFeatures*) l)->get_num_feature_obj()<=0 ||
-		((CCombinedFeatures*) l)->get_num_feature_obj() != ((CCombinedFeatures*) r)->get_num_feature_obj() )
-		SG_ERROR("CombinedKernel: Number of features/kernels does not match - bailing out\n")
+	if (((CCombinedFeatures*)l)->get_num_feature_obj() <= 0 ||
```
vigsterkr: you can dynamic cast these into a var and then use them... one would of course suppose that the compiler would do this as well under the hood... but it might be easier to read the code...?

vigsterkr:
thnx for the fix!

src/shogun/kernel/CombinedKernel.cpp
```diff
@@ -84,6 +84,9 @@ bool CCombinedKernel::init_with_extracted_subsets(
 	CFeatures* rf = NULL;
 	CKernel* k = NULL;
 
+    auto l_combined = dynamic_cast<CCombinedFeatures*>(l);
```
vigsterkr: imo this will be a formatting error :(

dhelekal: oh that's odd, no clue why my editor suddenly switched to spaces instead of tabs. (

vigsterkr:
https://travis-ci.org/shogun-toolbox/shogun/jobs/334467279#L745
:(

src/shogun/kernel/CombinedKernel.cpp
```diff
@@ -61,54 +61,37 @@ void CCombinedKernel::init_subkernel_weights()
...
-	ASSERT(l->get_feature_type()==F_UNKNOWN)
-	ASSERT(r->get_feature_type()==F_UNKNOWN)
-
-	CFeatures* lf=NULL;
-	CFeatures* rf=NULL;
-	CKernel* k=NULL;
-
-	bool result=true;
+	CKernel::init(l, r);
+	REQUIRE(
```
karlnapf: if you require this, you could have asked for those types in the signature of the method, no?

src/shogun/kernel/CombinedKernel.cpp
```diff
@@ -61,54 +61,37 @@ void CCombinedKernel::init_subkernel_weights()
...
+	    l->get_feature_class() == C_COMBINED,
+	    "%s::init(): LHS features are"
```
karlnapf: we actually don't do "CLASSNAME::METHODNAME():" anymore

src/shogun/kernel/CombinedKernel.cpp
```diff
@@ -61,54 +61,37 @@ void CCombinedKernel::init_subkernel_weights()
...
-	ASSERT(r->get_feature_type()==F_UNKNOWN)
-
-	CFeatures* lf=NULL;
-	CFeatures* rf=NULL;
-	CKernel* k=NULL;
-
-	bool result=true;
+	CKernel::init(l, r);
+	REQUIRE(
+	    l->get_feature_class() == C_COMBINED,
```
karlnapf: this segfaults if the caller passes `nullptr`

src/shogun/kernel/CombinedKernel.cpp
```diff
@@ -61,54 +61,37 @@ void CCombinedKernel::init_subkernel_weights()
...
+	    " of class %s but need to be combined features!\n",
+	    get_name(), r->get_name());
+	ASSERT(l->get_feature_type() == F_UNKNOWN)
+	ASSERT(r->get_feature_type() == F_UNKNOWN)
+
+	CFeatures* lf = NULL;
+	CFeatures* rf = NULL;
+	CKernel* k = NULL;
+
+	auto l_combined = dynamic_cast<CCombinedFeatures*>(l);
```
karlnapf: why dynamic cast if you don't check whether it worked?

src/shogun/kernel/CombinedKernel.cpp
```diff
@@ -169,15 +176,99 @@ bool CCombinedKernel::init(CFeatures* l, CFeatures* r)
...
 }
 
+bool CCombinedKernel::init(CFeatures* l, CFeatures* r)
+{
+	if (enable_subkernel_weight_opt && !weight_update)
+	{
+		init_subkernel_weights();
+	}
+	/*
+	 * The two subsets, we will be passing those to init_with_extracted_subsets
```
karlnapf: usually, we try to not have comments like this, a bit superflous, the code tells the story :)

src/shogun/kernel/CombinedKernel.cpp
```diff
@@ -169,15 +176,99 @@ bool CCombinedKernel::init(CFeatures* l, CFeatures* r)
...
+	 */
+	SGVector<index_t> lhs_subset;
+	SGVector<index_t> rhs_subset;
+
+	/*
+	 * We will be passing these features to init_with_extracted_subsets
```
karlnapf: same as above

src/shogun/kernel/CombinedKernel.cpp
```diff
@@ -169,15 +176,99 @@ bool CCombinedKernel::init(CFeatures* l, CFeatures* r)
...
+	 */
+	CCombinedFeatures* combined_l;
+	CCombinedFeatures* combined_r;
+
+	/*
+	 * Extract the subsets so that we can pass them on
```
karlnapf: and same again

src/shogun/kernel/CombinedKernel.cpp
```diff
@@ -169,15 +176,99 @@ bool CCombinedKernel::init(CFeatures* l, CFeatures* r)
...
+	else
+	{
+		rhs_subset = SGVector<index_t>(r->get_num_vectors());
+		rhs_subset.range_fill();
+	}
+
+	SG_UNREF(l_subset_stack);
+	SG_UNREF(r_subset_stack);
+
+	/* if the specified features are not combined features, but a single other
```
karlnapf: this is a good comment

src/shogun/kernel/CombinedKernel.cpp
```diff
@@ -169,15 +176,99 @@ bool CCombinedKernel::init(CFeatures* l, CFeatures* r)
...
+		for (index_t i = 0; i < get_num_subkernels(); ++i)
+		{
+			combined_l->append_feature_obj(l);
+			combined_r->append_feature_obj(r);
+		}
+	}
+	else
+	{
+		/*
+		 * Otherwise, we just pass l & r straight on
```
karlnapf: this can go

tests/unit/kernel/CombinedKernel_unittest.cc
```diff
@@ -37,6 +42,135 @@ TEST(CombinedKernelTest,test_array_operations)
 	SG_UNREF(combined);
 }
 
+TEST(CombinedKernelTest, test_subset_mixed)
+{
+
+	CMeanShiftDataGenerator* gen = new CMeanShiftDataGenerator(0, 2);
```
karlnapf: if you use `auto gen = some<CMeanShiftDataGenerator>(0,2);`, then you dont need to `SG_UNREF` later on

tests/unit/kernel/CombinedKernel_unittest.cc
```diff
@@ -37,6 +42,135 @@ TEST(CombinedKernelTest,test_array_operations)
...
+
+		SGMatrix<float64_t> gauss_matrix = k_g->get_kernel_matrix();
+		SGMatrix<float64_t> custom_matrix_1 = k_0->get_kernel_matrix();
+		SGMatrix<float64_t> custom_matrix_2 = k_3->get_kernel_matrix();
+
+		for (index_t j = 0; j < 10; ++j)
+		{
+			for (index_t k = 0; k < 10; ++k)
+			{
+				EXPECT_LE(
```
karlnapf: EXPECT_NEAR would be much cleaner

tests/unit/kernel/CombinedKernel_unittest.cc
```diff
@@ -37,6 +42,135 @@ TEST(CombinedKernelTest,test_array_operations)
...
+	cf->append_feature_obj(feats);
+
+	combined->append_kernel(custom_2);
+	combined->append_kernel(gaus_2);
+	cf->append_feature_obj(feats);
+
+	SGVector<index_t> inds(10);
+	inds.range_fill();
+
+	for (index_t i = 0; i < 10; ++i)
```
karlnapf: why not put the `10` into a variable with a meaninful name?

tests/unit/kernel/CombinedKernel_unittest.cc
```diff
@@ -37,6 +42,135 @@ TEST(CombinedKernelTest,test_array_operations)
 	SG_UNREF(combined);
 }
 
+TEST(CombinedKernelTest, test_subset_mixed)
+{
+
+	CMeanShiftDataGenerator* gen = new CMeanShiftDataGenerator(0, 2);
+	CFeatures* feats = gen->get_streamed_features(10);
+
+	CCombinedFeatures* cf = new CCombinedFeatures();
```
karlnapf: better name: `feats_combined`  (good names avoid the need for code comments)

tests/unit/kernel/CombinedKernel_unittest.cc
```diff
@@ -37,6 +42,135 @@ TEST(CombinedKernelTest,test_array_operations)
...
+
+	for (index_t i = 0; i < 10; ++i)
+	{
+		CMath::permute(inds);
+
+		cf->add_subset(inds);
+		combined->init(cf, cf);
+
+		CKernel* k_g = combined->get_kernel(1);
+		CKernel* k_0 = combined->get_kernel(0);
```
karlnapf: these names are confusing me

tests/unit/kernel/CombinedKernel_unittest.cc
```diff
@@ -37,6 +42,135 @@ TEST(CombinedKernelTest,test_array_operations)
...
+		CKernel* k_3 = combined->get_kernel(2);
+
+		SGMatrix<float64_t> gauss_matrix = k_g->get_kernel_matrix();
+		SGMatrix<float64_t> custom_matrix_1 = k_0->get_kernel_matrix();
+		SGMatrix<float64_t> custom_matrix_2 = k_3->get_kernel_matrix();
+
+		for (index_t j = 0; j < 10; ++j)
```
karlnapf: I think `SGMatrix` has an `equals` method

PR #4128, opened 2018-01-27
Extend the meta-language to support keyword arguments

sorig:
Allows keyword arguments in class constructors and factory functions (when the factory function is called to set the value of a newly initialised variable).

The update in meta-language allows translation of statements like these (meta-language):
```
Kernel k = kernel_machine("SomeKernel", some_parameter=2)

Kernel k2(parameter=500)
```
**A)** To something like this (python)
```
k = kernel_machine("SomeKernel")
k.put("some_parameter", 2)

k2 = Kernel()
k2.put("parameter", 500)
```
**B)** Or something like this (python):
```
k = kernel_machine("SomeKernel", some_parameter=2)

k2 = Kernel(parameter=500)
```
# Changing from translation type A) to B)
To get A), the `targets/python.json` file should contain:
```
{
    ...
    "Init": {
        "Construct": "$name = $typeName($arguments)\n$kwargs",
        "Copy": "$name = $expr\n$kwargs",
        "KeywordArguments": {
            "List": "$elements",
            "Element": "$name.put(\"$keyword\",$expr)",
            "Separator": "\n",
            "InitialSeperatorWhenArgs>0": false
        },
        ...
    },
    ...
    "Expr": {
        ...
        ...
    },
    ...
}
```

And to get B), the `targets/python.json` should contain:
```
{
    ...
    "Init": {
        "Construct": "$name = $typeName($arguments$kwargs)",
        "Copy": "$name = $expr",
        "KeywordArguments": {
            "List": "$elements",
            "Element": "$keyword=$expr",
            "Separator": ", ",
            "InitialSeperatorWhenArgs>0": true
        },
        ...
    },
    ...
    "Expr": {
        ...
        "GlobalCall": "$method($arguments$kwargs)",
        ...
    },
    ...
}
```

# Where do keyword arguments not work?

Keywords arguments cannot be used anywhere else than in class constructors to initiliase variables and global functions to initialise variables. See the following examples:
```
# Keyword arguments are supported when constructing objects

# Either when using factory functions:
Kernel k = kernel_machine(a=2)
Kernel k2 = kernel_machine("GaussianKernel", c=1.0)

# Or when constructing objects from class constructors
GaussianKernel k3(1, a=2, b=3)

# This fails (keyword arguments appear before normal arguments)
#Kernel k4(asd=123,"SomeParam",kw=val)

# This fails (keywords not allowed outside initialisations of variables. This is a re-assignment of a previously initialised variable)
#k4 = kernel_machine("GaussianKernel", b=2)

# This fails (keywords not allowed outside initialisation of variables. This is a nested kwarg, but it doesn't appear in the call that initialises `k6` directlry)
#Kernel k6 = kernel_machine("GaussianKernel", a=glob_fun(kw=123))

# This is fine:
Kernel k6 = kernel_machine("GaussianKernel", a=glob_fun(ordinary_argument))

# This also fails (both kwargs appear outside of variable initialisation)
#kernel_machine("GaussianKernel", a=glob_fun(kw=123))
```

### DIFF
--- a/examples/meta/generator/parse.py
+++ b/examples/meta/generator/parse.py
@@ -2,6 +2,7 @@ import sys
 import json
 import argparse
 import ply
+import os
 
 
 # The FastParser parses input using PLY
@@ -24,7 +25,7 @@ class FastParser:
                                 outputdir=generatedFilesOutputDir,
                                 debug=generateFiles)
 
-    def parse(self, programString, filePath="Unknown"):
+    def parse(self, programString, filePath=None):
         """
         Parse a program string.
         The path of the program file is added as metadata to returned object
@@ -36,7 +37,10 @@ class FastParser:
 
         # Parse program and add FilePath key to object
         program = self.parser.parse(programString)
-        program["FilePath"] = filePath
+        if filePath is not None:
+            program["FilePath"] = os.path.abspath(filePath)
+        else:
+            program["FilePath"] = "Unknown"
 
         return program
 
@@ -181,15 +185,6 @@ class FastParser:
         "objecttype : IDENTIFIER"
         p[0] = {"ObjectType": p[1]}
 
-    def p_argumentList_nonEmpty(self, p):
-        """
-        argumentListNonEmpty : expr
-                             | expr COMMA argumentListNonEmpty
-        """
-        p[0] = [p[1]]
-        if len(p) > 2:
-            p[0].extend(p[3])
-
     def p_argumentList(self, p):
         """
         argumentList : argumentListNonEmpty
@@ -200,6 +195,25 @@ class FastParser:
             arguments = p[1]
 
         p[0] = {"ArgumentList": arguments}
+
+    def p_argumentListNonEmpty(self, p):
+        """
+        argumentListNonEmpty : argumentListElement
+                             | argumentListElement COMMA argumentListNonEmpty
+        """
+        p[0] = [p[1]]
+        if len(p) > 2:
+            p[0].extend(p[3])
+
+    def p_argumentListElement(self, p):
+        """
+        argumentListElement : expr
+                            | identifier EQUALS expr
+        """
+        if "Expr" in p[1]:
+            p[0] = p[1]
+        else:
+            p[0] = {"KeywordArgument": [p[1], p[3]]}
 
     def p_identifier(self, p):
         "identifier : IDENTIFIER"
--- a/examples/meta/src/meta_api/kwargs.sg
+++ b/examples/meta/src/meta_api/kwargs.sg
@@ -1,0 +1,38 @@
+# Keyword arguments are supported when constructing objects
+
+# Either when using factory functions:
+#Kernel k1 = kernel_factory(param=2)
+#Kernel k2 = kernel_factory("GaussianKernel", c=1.0)
+
+# Or when constructing objects from class constructors
+#SomeKernelClass k3(1, a=2, b=3)
+
+# This fails (keyword arguments appear before normal arguments)
+#Kernel k4(asd=123, "SomeParam", kw=val)
+
+# This fails (keywords not allowed outside initialisations of variables)
+#k4 = kernel_factory("GaussianKernel", b=2)
+
+# This fails (keywords not allowed outside initialisation of variables)
+#Kernel k5 = kernel_factory("GaussianKernel", a=glob_fun(kw=123))
+
+# This is fine:
+#Kernel k6 = kernel_factory("GaussianKernel", a=glob_fun(ordinary_argument))
+
+# This also fails (keywords not allowed outside initialisation variables)
+#kernel_factory("GaussianKernel", a=glob_fun(ordinary_argument))
+
+
+# Real example
+# ------------
+
+# The following lines fail in Octave because of a float literal issue
+#KernelMachine svm = kernel_machine("LibSVM", C1=2.0)
+#GaussianKernel k(log_width=2.0)
+#k.put("log_width", 4.0)
+
+# The following line fails in Python because of a bool literal issue
+#GaussianKernel k2(log_width=14.0, lhs_equals_rhs=True)
+
+# The following line currently gives a type error
+#svm.put("kernel", k)
--- a/examples/meta/generator/generate.py
+++ b/examples/meta/generator/generate.py
@@ -69,8 +69,8 @@ def translateExamples(inputDir, outputDir, targetsDir, ctagsFile,
         # print("Translating {}".format(os.sep.join([dirRelative, filename])))
 
         # Parse the example file
-        filePath = os.path.join(dirRelative, filename)
-        with open(os.path.join(inputDir, dirRelative, filename), 'r') as file:
+        filePath = os.path.join(inputDir, dirRelative, filename)
+        with open(filePath, 'r') as file:
             ast = parse(file.read(), filePath, generatedFilesOutputDir)
 
         # Translate ast to each target language
--- a/examples/meta/generator/targets/cpp.json
+++ b/examples/meta/generator/targets/cpp.json
@@ -10,8 +10,14 @@
     "Statement": "$statement;\n",
     "Comment": "//$comment\n",
     "Init": {
-        "Construct": "auto $name = some<C$typeName>($arguments)",
-        "Copy": "auto $name = $expr",
+        "Construct": "auto $name = some<C$typeName>($arguments)$kwargs",
+        "Copy": "auto $name = $expr$kwargs",
+        "KeywordArguments": {
+            "List": ";\n$elements",
+            "Element": "$name->put(\"$keyword\", $expr)",
+            "Separator": ";\n",
+            "InitialSeperatorWhenArgs>0": false
+        },
         "BoolVector": "auto $name = SGVector<bool>($arguments)",
         "CharVector": "auto $name = SGVector<char>($arguments)",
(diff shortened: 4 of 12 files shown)


sorig:
@karlnapf Can we get a real example of this instead of my `examples/meta/src/meta_api/kwargs.sg` file? That would make it easier to change all the other target `.json` files (and see what edge cases I've missed)

karlnapf:
Ah this is great, thanks for that!

I will review a bit in my phone as I am travelling

karlnapf:
As for real examples, keep in mind that the kwargs.sg file is translated and executed.
I think it should fail as you’re setting parameters that don’t exist....

I think creating a few objects, mimicking their constructors, in your file should re real enough?

karlnapf:
Can’t see the travis logs but all jobs failed which is due to the putting Parameter c of a Gaussian kernel

karlnapf:
You can both cover the current constructors and a factory method

sorig:
Yeah I know, but they would fail anyway since I've only added Python support for now.

karlnapf:
Sweet! I would start with making an existing ctor work and then do the factory method next

examples/meta/generator/parse.py
```diff
@@ -201,6 +192,25 @@ def p_argumentList(self, p):
 
         p[0] = {"ArgumentList": arguments}
 
+    def p_argumentList_nonEmpty(self, p):
```
karlnapf: Not sure about this capitalisation underscore mix

sorig: haha yeah you're right. Will make this consistent

sorig: Initial underscores `t_` and `p_` are enforced by the parsing library, that's why there is some confusion in this file

examples/meta/src/meta_api/kwargs.sg
```diff
@@ -0,0 +1,38 @@
...
+#kernel_machine("GaussianKernel", a=glob_fun(kw=123))
+
+
+# REAL EXAMPLE
+# ------------
+# The cpp translation of the following line fails with exception:
+# > libc++abi.dylib: terminating with uncaught exception of type shogun::ShogunException
+# > [ERROR] In file /Users/esben/Code/shogun/src/shogun/base/class_list.h line 51: Type mismatchAbort trap: 6
+
+# GaussianKernel k0 = kernel_machine("GaussianKernel", log_width=2.0)
```
sorig: Executing the C++ translation of this line, i.e.
```
auto k0 = kernel_machine("GaussianKernel");
k0->put("log_width", 2.0);
```
Throws the exception
```
libc++abi.dylib: terminating with uncaught exception of type shogun::ShogunException
[ERROR] In file /Users/esben/Code/shogun/src/shogun/base/class_list.h line 51: Type mismatchAbort trap: 6
```

@karlnapf can you guide me here? Don't know what I'm doing wrong.

I got the C++ interface working with the statements below this line (using class constructors), and I'm currently trying to get it working for the other interfaces.

sorig:
While `examples/meta/src/meta_api/kwargs.sg` works with C++, Python fails with the error:
```
386/392 Test #364: generated_python-meta_api-kwargs ........................................................***Failed    0.20 sec
  File "/opt/shogun/build/examples/meta/python/meta_api/kwargs.py", line 44, in <module>
    k2.put("lhs_equals_rhs", True)
RuntimeError: [ERROR] In file /opt/shogun/src/shogun/base/SGObject.h line 349: Type for parameter with name "lhs_equals_rhs" is not correct.
```

`lhs_equals_rhs` is declared as a C++ `bool` in `CKernel`. Is this a problem with a type map somewhere?

karlnapf:
@lisitsyn here is someone using tags, so maybe you can help?
I’ll look back into from friday

sorig:
Similarly, as the only language octave is complaining about:
```
k0 = kernel_machine("GaussianKernel")
k0.put("log_width", 2.0)
```
throwing this error:
```
[ERROR] In file /opt/shogun/src/shogun/base/SGObject.h line 349: Type for parameter with name "log_width" is not correct.
error: C++ side threw an exception of type shogun::ShogunException
error: called from
    /opt/shogun/build/examples/meta/octave/meta_api/kwargs.m at line 37 column 1
```

while the equivalent code in all other languages works fine

sorig:
@lisitsyn Here's an issue related to the SWIG interfaces and the parameter framework. In python, I can't see what the parameters of an object is:
```
In [1]: from shogun import GaussianKernel

In [2]: k = GaussianKernel()

In [3]: k.parameter_names()
Out[3]: <Swig Object of type 'std::set< std::string > *' at 0x11408e458>
```

Which means I have to dig around in C++ code to find the parameter names. Can this be fixed?

vigsterkr:
I reckon a .i include missing in our swig descriptors for typemapping std::set

vigsterkr:
@sorig i've just tried to do that what i said above.... it's not working... i'll let u know if i have found a solution

vigsterkr:
@sorig ok i've managed to do this:
```>>> import shogun
>>> k = shogun.GaussianKernel()
>>> k.parameter_names()
('cache_size', 'combined_kernel_weight', 'lhs', 'lhs_equals_rhs', 'log_width', 'm_distance', 'm_precomputed_distance', 'normalizer', 'num_lhs', 'num_rhs', 'opt_type', 'optimization_initialized', 'properties', 'rhs')```

i guess this is what you would like to have right? :)

**Comment by sorig** [comment]
@vigsterkr Amazing, thanks! Crucial functionality for people working from the swig interfaces

**Comment by vigsterkr** [comment]
@sorig lemme push it to develop

**Comment by vigsterkr** [comment]
3cc0e3740c5de1e0f31aebbb427615d8ef0e4d07 is the patch that gets u the above mentioned functionality

**Comment by vigsterkr** [comment]
we have actually a problem: std::set isn't supported in the following interfaces: lua, java, set, r, octave, c#.
in other words wrapper for std::set is only available in case of python, ruby, in the rest of the case the api will not be available... which is far from optimal so we should actually return our own `CSet` :D yaaay for custom things :(

**Comment by sorig** [comment]
Yeah, it looks like those interfaces just broke with the change. Is it really necessary to have `std::set` exposed on the interfaces? It seems to me that `parameter_names` would be absolutely fine as a `std::list`. Are there examples in the code base where `std::set` is absolutely necessary?

**Comment by vigsterkr** [comment]
@sorig if we wanna switch we should then switch the interface to use `std::vector<std::string>` as that's the only typemap that is available in all of the languages that we support (std::list is not available in c# and lua)

**Comment by vigsterkr** [comment]
@lisitsyn what do you think? custom `CSet<std::string>` or `std::vector<std::string>` as return value for `parameter_names()`?

**Comment by lisitsyn** [comment]
@vigsterkr std::vector for sure!

**Comment by vigsterkr** [comment]
@lisitsyn do you have any clue why
```
%include "std_vector.i"
```

is surrounded by `#if !defined(SWIGJAVA)` ?
i.e. why no java typemaps for 
```
    %template(IntStdVector) vector<int32_t>;
    %template(DoubleStdVector) vector<float64_t>;
``` 
?

**Comment by vigsterkr** [comment]
commit 5423ead4de9a0304087112c027b7795fb0c0fa14
Author: Christian Widmer <cwidmer@tuebingen.mpg.de>
Date:   Fri Jun 15 17:29:53 2012 +0200

    included std_vector typemap while avoiding java compile error


one would hope that there's no more compiler error, or? :D

**Comment by lisitsyn** [comment]
Who knows.. :)

**Comment by vigsterkr** [comment]
@sorig 532a8d45770e4e82ba8291c35f3e0985beec1930 should be the thing for you :)

**Comment by sorig** [comment]
Great, thanks a lot!

**Comment by sorig** [comment]
All green except for Octave (see error description above) which seems to be a bug with the octave interface and float types. I'm not currently able to install octave on my macine to figure out what the issue is. Is there a server somewhere where I can test all the interfaces?

It looks like there are issues with basic types in several of the interface languages. Would be good to write a thorough test case once this PR is merged.

File: examples/meta/src/meta_api/kwargs.sg [hunk 3]
```diff
@@ -0,0 +1,37 @@
...
+#kernel_machine("GaussianKernel", a=glob_fun(kw=123))
+
+
+# REAL EXAMPLE
+# ------------
+KernelMachine svm = kernel_machine("LibSVM", C1=2.0)
+
+GaussianKernel k(log_width=2.0)
+
+GaussianKernel k2(log_width=14.0)
```
karlnapf: why two kernels?

examples/meta/src/meta_api/kwargs.sg
```diff
@@ -0,0 +1,37 @@
...
+#Kernel k6 = kernel_machine("GaussianKernel", a=glob_fun(kw=123))
+
+# This is fine:
+#Kernel k6 = kernel_machine("GaussianKernel", a=glob_fun(ordinary_argument))
+
+# This also fails (keywords not allowed outside initialisation variables)
+#kernel_machine("GaussianKernel", a=glob_fun(kw=123))
+
+
+# REAL EXAMPLE
```
karlnapf: Real example

examples/meta/src/meta_api/kwargs.sg
sorig: One of them had a boolean keyword argument earlier, but that fails in python. I’ll revert it and comment it out

sorig:
OK, included last review changes. This is ready to be merged

karlnapf:
Excellent, thanks!

PR #4129, opened 2018-01-28
Correct syntax for compatibilty with python3

syashakash:
The argument in line 127 for `re.compile()` doesn't need `u` flag as the argument has no Unicode letters used. Moreover, this shows as syntax error when run with python3 interpretor.
Removing `u` flag has no adverse consequences on the remaining setup process.


lisitsyn:
Thanks for the patch!

PR #4131, opened 2018-01-28
Add ammended correct path to versionstring_h

syashakash:
On running setup.py, the following error is thrown:

`Traceback (most recent call last):
  File "setup.py", line 300, in <module>
    version = get_shogun_version(),
  File "setup.py", line 207, in get_shogun_version
    shogun_version = parse_shogun_version(shogun_versionstring_h)
  File "setup.py", line 130, in parse_shogun_version
    with open(version_header, 'r') as f:
IOError: [Errno 2] No such file or directory: 'src/shogun/lib/versionstring.h'`

Currently setup.py:iline 48, `shogun_versionstring_h` stores location to `src/shogun/lib/versionstring.h` that doesn't exists, either CMake fails to convert versionstring.h.in  to versionstring.h or it does convert, but it is stored in /build/install/include/shogun/lib/. I think the latter should be the case.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -45,7 +45,7 @@ setup_py_location = os.path.abspath(os.path.dirname(__file__))
 
 shogun_build_directory = os.path.join(setup_py_location, 'build')
 shogun_generated_install = os.path.join(shogun_build_directory, 'install')
-shogun_versionstring_h = os.path.abspath('src/shogun/lib/versionstring.h')
+shogun_versionstring_h = os.path.join(shogun_generated_install, 'include/shogun/lib/versionstring.h')
 shogun_python_packages_location = None
 
 shogun_completed_cmake = False


vigsterkr:
@syashakash thnx for the patch but there needs to be way more improvement to get the setup.py working....

syashakash:
@vigsterkr yes I understand. That means the path was still incorrect! I thought it kinda worked and it did pass, now I am getting another error something like this:

`Shogun Python package data files:
|->[lib]
    |--> libshogun.so.18.0
    |--> libshogun.so.18
    |--> libshogun.so
|->[lib/python2.7/dist-packages]
    |--> _shogun.so
usage: setup.py [global_opts] cmd1 [cmd1_opts] [cmd2 [cmd2_opts] ...]
   or: setup.py --help [cmd1 cmd2 ...]
   or: setup.py --help-commands
   or: setup.py cmd --help

error: no commands supplied`

To this, I am struggling to get a solution!
Could you suggest how I could help make it working ?

vigsterkr:
@syashakash this needs to be rewritten from scratch, this was just an attempt that was never working.

syashakash:
@vigsterkr I get it. I would someday get back to it, help make it from scratch so that installation becomes more easier. Thanks by the way.

karlnapf:
@vigsterkr Shall we just drop `setup.py` for now?

vigsterkr:
totally

PR #4132, opened 2018-01-28
Store parameter descriptions

lisitsyn:


lisitsyn:
Clang is ok but gcc constantly gets out of time limit. I would not expect it to fail hence merging.

src/shogun/base/AnyParameter.h
```diff
@@ -24,22 +26,30 @@ namespace shogun
 	{
 	public:
 		AnyParameterProperties()
-		    : m_model_selection(MS_NOT_AVAILABLE),
+		    : m_description(), m_model_selection(MS_NOT_AVAILABLE),
 		      m_gradient(GRADIENT_NOT_AVAILABLE)
 		{
 		}
 		AnyParameterProperties(
+		    std::string description,
```
karlnapf: I think we should have a constructor that only takes the description and sets the other two to false (they are not used most of the time)

PR #4137, opened 2018-01-30
Add meta example for canberra distance, remove python example from undocumented. 

syashakash:
#3555


karlnapf:
Please don't close PRs if you are sending another one. Rather update it

PR #4139, opened 2018-02-01
vector & matrix parameter descriptions in tags

guruhegde:
Made changes as suggested.

* Tests pass.
* style check pass.

Comments: Some parameters in couple of classes doesn't have a description. Kept them as it is.

### DIFF
--- a/src/shogun/classifier/mkl/MKL.cpp
+++ b/src/shogun/classifier/mkl/MKL.cpp
@@ -278,7 +278,10 @@ void CMKL::register_params()
 	SG_ADD(&mkl_block_norm, "mkl_block_norm", "mkl sparse trade-off parameter", MS_NOT_AVAILABLE);
 
 	m_parameters->add_vector(&beta_local, &beta_local_size, "beta_local", "subkernel weights on L1 term of elastic net mkl");
-	watch_param("beta_local", &beta_local, &beta_local_size);
+	watch_param(
+	    "beta_local", &beta_local, &beta_local_size,
+	    AnyParameterProperties(
+	        "subkernel weights on L1 term of elastic net mkl"));
 
 	SG_ADD(&mkl_iterations, "mkl_iterations", "number of mkl steps", MS_NOT_AVAILABLE);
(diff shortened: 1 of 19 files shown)


src/shogun/classifier/mkl/MKL.cpp
```diff
@@ -278,7 +278,11 @@ void CMKL::register_params()
 	SG_ADD(&mkl_block_norm, "mkl_block_norm", "mkl sparse trade-off parameter", MS_NOT_AVAILABLE);
 
 	m_parameters->add_vector(&beta_local, &beta_local_size, "beta_local", "subkernel weights on L1 term of elastic net mkl");
-	watch_param("beta_local", &beta_local, &beta_local_size);
+	watch_param(
+	    "beta_local", &beta_local, &beta_local_size,
+	    AnyParameterProperties(
+	        "subkernel weights on L1 term of elastic net mkl", MS_NOT_AVAILABLE,
```
karlnapf: same here, pls let's not have this `MS_NOT_AVAILABLE, GRADIENT_NOT_AVAILABLE` everytime, but instead hide it, and only write it out if things are available

guruhegde:
@karlnapf Hi this is ready for review:)

src/shogun/classifier/mkl/MKL.cpp
```diff
@@ -278,7 +278,10 @@ void CMKL::register_params()
 	SG_ADD(&mkl_block_norm, "mkl_block_norm", "mkl sparse trade-off parameter", MS_NOT_AVAILABLE);
 
 	m_parameters->add_vector(&beta_local, &beta_local_size, "beta_local", "subkernel weights on L1 term of elastic net mkl");
-	watch_param("beta_local", &beta_local, &beta_local_size);
+	watch_param(
+	    "beta_local", &beta_local, &beta_local_size,
+	    AnyParameterProperties(
```
karlnapf: Couldn't we have this without the `AnyParameterProperties`? but just string?
@lisitsyn

PR #4140, opened 2018-02-01
#3555 add meta_example for classifier_perceptron undocumented-sample

grg121:
Hi! My name is Paco and I'm interested in this project and I would like to get involved with it.

I've created the sample and tested c and python result. They are both the same. 
I'm sending another PR to shogun-data with the integration testing data output of the c++ executable and I'll make a cookbook page then.

### DIFF
--- a/examples/meta/src/neural_nets/classification_perceptron.sg
+++ b/examples/meta/src/neural_nets/classification_perceptron.sg
@@ -1,0 +1,35 @@
+CSVFile f_feats_train("../../data/classifier_binary_2d_linear_features_train.dat")
+CSVFile f_feats_test("../../data/classifier_binary_2d_linear_features_test.dat")
+CSVFile f_labels_train("../../data/classifier_binary_2d_linear_labels_train.dat")
+CSVFile f_labels_test("../../data/classifier_binary_2d_linear_labels_test.dat")
+
+#![create_features]
+RealFeatures features_train(f_feats_train)
+RealFeatures features_test(f_feats_test)
+BinaryLabels labels_train(f_labels_train)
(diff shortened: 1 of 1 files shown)


karlnapf:
Welcome Paco, thanks for the patch! :)

karlnapf:
The example looks good. I have in the meantime merged you `data` PR, so you can now add the update to this PR as well

grg121:
Done! :) 

Thank you! Now I'm gonna write a Cookbook page.

karlnapf:
Thanks for the contribution! :)

karlnapf:
btw is there still an undocumented python example that can be deleted?

examples/meta/src/neural_nets/classification_perceptron.sg
```diff
@@ -0,0 +1,35 @@
+CSVFile f_feats_train("../../data/classifier_binary_2d_linear_features_train.dat")
```
karlnapf: ah, I had overseen this.
Can you pls send another PR where you  move the perceptron to the "binary" folder, it does not really follow the NN structure in Shogun. Same for the data file.
Also could you remove the python example in the same PR?

grg121: Ahh ok, I thought it had to be with NN examples, I'll move it and remove the python example...

Also... I'm seeing that one check didn't pass which is weird because I just send the data file... don't know :/ 

Well, I'lll do that and send another commit!

grg121: another PR* xp

PR #4148, opened 2018-02-04
first systematic unit test for SGObject::clone

karlnapf:


karlnapf:
@lisitsyn I need that clone of arrays

PR #4149, opened 2018-02-05
Use linalg in LinearRidgeRegression

vinx13:
This PR replaces lapack guard and Eigen usage with linalg in `LinearRidgeRegression`


karlnapf:
Thanks! :)

PR #4150, opened 2018-02-05
Replace old clone tests with systematic new one

karlnapf:
CSGObject::clone should be fully covered now i.e.

* all parameter combinations that are used in Shogun are cloned in the test
* after clone, object.equals(...) is true
* non-empty objects to make meaningful test


karlnapf:
no windows-specific changes here, merging

PR #4153, opened 2018-02-05
Cleanup PR 4127

dhelekal:


scripts/check_format.sh
```diff
@@ -41,7 +41,7 @@ function check_shogun_style {
             fi
         done
 
-        RESULT_OUTPUT="$(git clang-format-3.8 --commit $BASE_COMMIT --diff --binary `which clang-format-3.8` $LIST)"
```
vigsterkr: need to get the Docker image fixed for this on travis

dhelekal: Yeah, sorry, i didnt want to commit that change

src/shogun/kernel/CombinedKernel.cpp
```diff
@@ -66,27 +66,20 @@ bool CCombinedKernel::init_with_extracted_subsets(
     SGVector<index_t> rhs_subset)
 {
 
+	auto l_combined = dynamic_cast<CCombinedFeatures*>(l);
+	auto r_combined = dynamic_cast<CCombinedFeatures*>(r);
+
+	if (!l_combined || !r_combined)
+		SG_ERROR("Cast failed - unsupported features passed")
```
vigsterkr: imo this will never happen (unless l or r is null) since if the typecasting fails it'll throw a bad_casting exception.

src/shogun/kernel/CombinedKernel.cpp
```diff
@@ -138,6 +131,8 @@ bool CCombinedKernel::init_with_extracted_subsets(
 				    "No kernel matrix was assigned to this Custom kernel\n")
 
 			auto k_custom = dynamic_cast<CCustomKernel*>(k);
+			if (!k_custom)
```
vigsterkr: same as above, see http://en.cppreference.com/w/cpp/language/dynamic_cast

vigsterkr: /me should really RTFM before saying RTFM

PR #4154, opened 2018-02-06
kwargs ~works~ for objects ~already~

karlnapf:


examples/meta/src/meta_api/kwargs.sg
```diff
@@ -22,17 +22,14 @@
...
 # ------------
 
-# The following lines fail in Octave because of a float literal issue
-#KernelMachine svm = kernel_machine("LibSVM", C1=2.0)
-#GaussianKernel k(log_width=2.0)
-#k.put("log_width", 4.0)
-
-# The following line fails in Python because of a bool literal issue
-#GaussianKernel k2(log_width=14.0, lhs_equals_rhs=True)
+Kernel k = kernel("GaussianKernel")
```
karlnapf: floats still don't work in octave...

karlnapf:
@lisitsyn works with cpp, python, java, octave on my machine

examples/meta/src/meta_api/kwargs.sg
```diff
@@ -22,17 +22,14 @@
 # This also fails (keywords not allowed outside initialisation variables)
 #kernel_factory("GaussianKernel", a=glob_fun(ordinary_argument))
 
+# doesnt work in ruby (TODO, don't allow in meta grammar)
```
karlnapf: @sorig we will need to block this, ruby doesn't like it

src/shogun/base/SGObject.cpp
```diff
@@ -1012,3 +1012,40 @@ CSGObject* CSGObject::create_empty() const
...
+	put(Tag<T>(name), (T)value);
+
+#define SGOBJECT_PUT_DEFINE_NUMBER(numeric_t) \
+void CSGObject::put(const std::string& name, numeric_t const & value)  throw(ShogunException)\
+{ \
+	/* use correct type of possible, otherwise cast-convert */ \
+	if (has(Tag<numeric_t>(name))) \
+		put(Tag<numeric_t>(name), value); \
+	PUT_DEFINE_CHECK_AND_CAST(int32_t) \
+	PUT_DEFINE_CHECK_AND_CAST(float32_t) \
```
karlnapf: @lisitsyn since it is only octave that cannot decide between float and int, maybe some of this could be done in swig....though I don't know how that would work....

karlnapf: https://www.gnu.org/software/octave/doc/v4.2.0/Integer-Data-Types.html

examples/meta/src/meta_api/kwargs.sg
sorig: What does ruby say? I can disable nested global calls, but that may disable some nested calls that are actually useful.

sorig: I think we're just using the wrong syntax. [See here](https://en.wikibooks.org/wiki/Ruby_Programming/Syntax/Method_Calls#Method_Calls)

PR #4158, opened 2018-02-07
Switch translation of ruby function calls to allow nested calls

sorig:
Fixes issue with ruby in #4154

### DIFF
--- a/examples/meta/src/meta_api/kwargs.sg
+++ b/examples/meta/src/meta_api/kwargs.sg
@@ -22,10 +22,6 @@
 # This also fails (keywords not allowed outside initialisation variables)
 #kernel_factory("GaussianKernel", a=glob_fun(ordinary_argument))
 
-# doesnt work in ruby (TODO, don't allow in meta grammar)
-# KernelMachine svm2 = kernel_machine("LibSVM", kernel=kernel("GaussianKernel"))
-
-
 
 # Real example
 # ------------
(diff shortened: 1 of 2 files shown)


sorig:
@karlnapf just seeing if this passes. May solve the issue you were seeing

karlnapf:
Just merged #4154 if you wanna rebase

karlnapf:
You will need to test that, I removed the nested call in the `kwargs.sg`

sorig:
Just rebased and added the test

examples/meta/src/meta_api/kwargs.sg
```diff
@@ -22,14 +22,11 @@
...
-
-
 
 # Real example
 # ------------
 
 Kernel k = kernel("GaussianKernel", log_width=2.0)
 KernelMachine svm = kernel_machine("LibSVM", C1=1.1, kernel=k)
 
+KernelMachine svm2 = kernel_machine("LibSVM", kernel=kernel("GaussianKernel"))
```
karlnapf: can also put the scalars in there :) we have some hackery that covers it now

karlnapf: but of course not necessary

karlnapf:
Thanks for the quick fix! :)

karlnapf:
Seems like ruby likes you :)

PR #4159, opened 2018-02-08
Add const qualifier to input of linalg api

vinx13:
There are some methods in linalg using non-const reference for input. This PR add const qualifiers to them.

PR #4160, opened 2018-02-08
Add const qualifiers to input of linalg APIs and fix #4159

vinx13:
Continue work in #4159 and fix missing const qualifiers in ViennaCL backend


karlnapf:
Cool! Thx. Travis is green so let's try

vinx13:
Sorry, seems still one missing 'const'. I created another PR.

PR #4165, opened 2018-02-09
WIP: swig and put

karlnapf:
What works

 * `CSGObject` (even using the old `SG_ADD((CSGObject**) &m_obj, ...)` thing)
* `SGMatrix` (modulo a missing jblas import in `SGObject.java` ... on it)
* scalars (including `4.0` in octave)

What doesn't work

* `SGVector` (on it right now)


karlnapf:
Pushed this into the typedo feature branch

Issue #4164, opened 2018-02-09
Create method to learn softmax parameters


durovo:
The OneVsAll softmax probability calibration method in shogun currently uses parameters learnt on a sigmoid. A method is needed that can be used to learn parameters for a softmax function instead.

The current `fit_sigmoid` call https://github.com/shogun-toolbox/shogun/blob/d1763b83fbc029bae1487201ef56977efdfe43a0/src/shogun/machine/MulticlassMachine.cpp#L132
should be replaced by some method using which we can learn params for softmax.
See description of equation 7 in paper https://hal.inria.fr/inria-00103955/document.

PR #4166, opened 2018-02-09
Fix platt scaling

durovo:
This PR fixes platt scaling and adds new classes SigmoidCalibration and Calibration. Previously, `fit_sigmoid` was only using the predictions to fit parameters which is incorrect. SigmoidCalibration has been added so that `score_to_probabilites` from `CBinaryLabels` can be removed. This will make the x-validated calibration API cleaner (#4009 ). The old `fit_sigmoid` and `scores_to_probabilities` have still not be removed since they are still needed in multiclass calibration. It would probably be safe to remove them after fixing #4164

### DIFF
--- a/examples/undocumented/libshogun/classifier_libsvm_probabilities.cpp
+++ b/examples/undocumented/libshogun/classifier_libsvm_probabilities.cpp
@@ -1,9 +1,10 @@
+#include <iostream>
 #include <shogun/base/init.h>
+#include <shogun/classifier/svm/LibSVM.h>
+#include <shogun/evaluation/SigmoidCalibration.h>
 #include <shogun/features/DenseFeatures.h>
 #include <shogun/kernel/LinearKernel.h>
 #include <shogun/labels/BinaryLabels.h>
-#include <shogun/classifier/svm/LibSVM.h>
-#include <iostream>
 
 using namespace shogun;
 
@@ -50,45 +51,53 @@ int main(int argc, char** argv)
 
     //create train labels
     CLabels* labels=new CBinaryLabels(labelVector);
+	SG_REF(labels);
 
-    //create train features
-    CDenseFeatures<float64_t>* features=new CDenseFeatures<float64_t>();
-    SG_REF(features);
-    features->set_feature_matrix(featureMatrix);
+	// create train features
+	CDenseFeatures<float64_t>* features = new CDenseFeatures<float64_t>();
+	SG_REF(features);
+	features->set_feature_matrix(featureMatrix);
 
-    //create linear kernel
-    CLinearKernel* kernel=new CLinearKernel();
-    SG_REF(kernel);
-    kernel->init(features, features);
+	// create linear kernel
+	CLinearKernel* kernel = new CLinearKernel();
+	SG_REF(kernel);
+	kernel->init(features, features);
 
-    //create svm classifier by LibSVM
-    CLibSVM* svm=new CLibSVM(svm_C,kernel, labels);
-    SG_REF(svm);
-    svm->train();
+	// create svm classifier by LibSVM
+	CLibSVM* svm = new CLibSVM(svm_C, kernel, labels);
+	SG_REF(svm);
+	svm->train();
 
-    //classify data points
-    CBinaryLabels* out_labels=CLabelsFactory::to_binary(svm->apply());
+	// classify data points
+	CBinaryLabels* out_labels = CLabelsFactory::to_binary(svm->apply());
 
-    /*convert scores to calibrated probabilities  by fitting a sigmoid function
-    using the method described in Lin, H., Lin, C., and Weng,  R. (2007). A note
-    on Platt's probabilistic outputs for support vector machines.
-    See BinaryLabels documentation for details*/
-    out_labels->scores_to_probabilities();
+	/*convert scores to calibrated probabilities  by fitting a sigmoid function
+	using the method described in Lin, H., Lin, C., and Weng,  R. (2007). A note
+	on Platt's probabilistic outputs for support vector machines.
+	See BinaryLabels documentation for details*/
+	CSigmoidCalibration* sigmoid_calibration = new CSigmoidCalibration();
+	sigmoid_calibration->fit_binary(
+	    out_labels, CLabelsFactory::to_binary(labels));
+	CBinaryLabels* calibrated_labels =
+	    sigmoid_calibration->calibrate_binary(out_labels);
 
-    //display output labels and probabilities
-    for (int32_t i=0; i<num_samples; i++)
-    {
-        SG_SPRINT("out[%d]=%f (%f)\n", i, out_labels->get_label(i),
-            out_labels->get_value(i));
-    }
+	// display output labels and probabilities
+	for (int32_t i = 0; i < num_samples; i++)
+	{
+		SG_SPRINT(
+		    "out[%d]=%f (%f)\n", i, out_labels->get_label(i),
+		    calibrated_labels->get_value(i));
+	}
 
-    //clean up
-    SG_UNREF(out_labels);
-    SG_UNREF(kernel);
-    SG_UNREF(features);
-    SG_UNREF(svm);
+	// clean up
+	SG_UNREF(sigmoid_calibration);
+	SG_UNREF(calibrated_labels);
+	SG_UNREF(out_labels);
+	SG_UNREF(kernel);
+	SG_UNREF(features);
+	SG_UNREF(svm);
 
-    exit_shogun();
+	exit_shogun();
 
-    return 0;
+	return 0;
 }
--- a/examples/undocumented/libshogun/labels_binary_fit_sigmoid.cpp
+++ b/examples/undocumented/libshogun/labels_binary_fit_sigmoid.cpp
@@ -5,6 +5,7 @@
  */
 
 #include <shogun/base/init.h>
+#include <shogun/evaluation/SigmoidCalibration.h>
 #include <shogun/labels/BinaryLabels.h>
 
 using namespace shogun;
@@ -12,16 +13,24 @@ using namespace shogun;
 void test_sigmoid_fitting()
 {
 	CBinaryLabels* labels=new CBinaryLabels(10);
+	CBinaryLabels* predictions = new CBinaryLabels(10);
 	labels->set_values(SGVector<float64_t>(labels->get_num_labels()));
 
 	for (index_t i=0; i<labels->get_num_labels(); ++i)
-		labels->set_value(i%2==0 ? 1 : -1, i);
+	{
+		predictions->set_value(i % 2 == 0 ? 1 : -1, i);
+		labels->set_value(i % 4 == 0 ? 1 : -1, i);
+	}
+	predictions->get_values().display_vector("scores");
+	CSigmoidCalibration* sigmoid_calibration = new CSigmoidCalibration();
+	sigmoid_calibration->fit_binary(predictions, labels);
+	CBinaryLabels* calibrated_labels =
+	    sigmoid_calibration->calibrate_binary(predictions);
+	calibrated_labels->get_values().display_vector("probabilities");
 
-	labels->get_values().display_vector("scores");
-	labels->scores_to_probabilities();
-	labels->get_values().display_vector("probabilities");
-
-
+	SG_UNREF(predictions);
+	SG_UNREF(sigmoid_calibration);
+	SG_UNREF(calibrated_labels);
 	SG_UNREF(labels);
 }
 
--- a/src/shogun/evaluation/Calibration.h
+++ b/src/shogun/evaluation/Calibration.h
@@ -1,0 +1,69 @@
+/*
+ * This software is distributed under BSD 3-clause license (see LICENSE file).
+ *
+ * Authors: Heiko Strathmann, Dhruv Arya
+ */
+
+#ifndef _CALIBRATION_H__
+#define _CALIBRATION_H__
+
+#include <shogun/lib/config.h>
+
+#include <shogun/machine/Machine.h>
+
+namespace shogun
+{
+	/** @brief Base class for all calibration methods. Call fit_binary() or
+	fit_multiclass() to
+	fit the parameters on the predictions and the true labels. Call calibrate to
+	calibrate predictions. **/
+	class CCalibration : public CSGObject
+	{
+	public:
+		/** Constructor. */
+		CCalibration()
+		{
+		}
+		/** Destructor. */
+		virtual ~CCalibration()
+		{
+		}
+
+		virtual const char* get_name() const
+		{
+			return "Calibration";
+		}
+
+		/** Fit calibration parameters for binary labels.
+		* @param predictions The predictions outputted by the machine
+		* @param targets The true labels corresponding to the predictions
+		* @return Indicates whether the calibration was succesful
+		**/
+		virtual bool
+		fit_binary(CBinaryLabels* predictions, CBinaryLabels* targets) = 0;
+
+		/** Calibrate binary predictions based on parameters learned by calling
+		*fit.
+		* @param predictions The predictions outputted by the machine
+		* @return Calibrated binary labels
+		**/
+		virtual CBinaryLabels* calibrate_binary(CBinaryLabels* predictions) = 0;
+
+		/** Fit calibration parameters for multiclass labels.
+		* @param predictions The predictions outputted by the machine
+		* @param targets The true labels corresponding to the predictions
+		* @return Indicates whether the calibration was succesful
+		**/
+		virtual bool fit_multiclass(
+		    CMulticlassLabels* predictions, CMulticlassLabels* targets) = 0;
+
+		/** Calibrate multiclass predictions based on parameters learned by
+		*calling fit.
+		* @param predictions The predictions outputted by the machine
+		* @return Calibrated binary labels
+		**/
+		virtual CMulticlassLabels*
+		calibrate_multiclass(CMulticlassLabels* predictions) = 0;
+	};
+}
+#endif
--- a/src/shogun/evaluation/SigmoidCalibration.cpp
+++ b/src/shogun/evaluation/SigmoidCalibration.cpp
@@ -1,0 +1,202 @@
+/*
+ * This software is distributed under BSD 3-clause license (see LICENSE file).
+ *
+ * Authors: Heiko Strathmann, Dhruv Arya
+ */
+
+#include <shogun/evaluation/SigmoidCalibration.h>
+#include <shogun/mathematics/Math.h>
+#include <shogun/mathematics/Statistics.h>
+#include <shogun/mathematics/linalg/LinalgNamespace.h>
+
+using namespace shogun;
+
+CSigmoidCalibration::CSigmoidCalibration() : CCalibration()
+{
+	init();
+}
+
+CSigmoidCalibration::~CSigmoidCalibration()
+{
+}
+
+void CSigmoidCalibration::init()
+{
+	m_maxiter = 100;
+	m_minstep = 1E-10;
+	m_sigma = 1E-12;
+	m_epsilon = 1E-5;
+
+	SG_ADD(
+	    &m_sigmoid_as, "m_sigmoid_as",
+	    "Vector of paramter A of sigmoid for each class.", MS_NOT_AVAILABLE);
+	SG_ADD(
+	    &m_sigmoid_bs, "m_sigmoid_bs",
+	    "Vector of paramter B of sigmoid for each class.", MS_NOT_AVAILABLE);
+	SG_ADD(
+	    &m_maxiter, "m_maxiter", "Maximum number of iteration for search.",
+	    MS_NOT_AVAILABLE);
+	SG_ADD(
+	    &m_minstep, "m_minstep", "Minimum step taken in line search.",
+	    MS_NOT_AVAILABLE);
+	SG_ADD(
+	    &m_sigma, "m_sigma",
+	    "Positive parameter to ensure positive semi-definite Hessian.",
+	    MS_NOT_AVAILABLE);
+	SG_ADD(&m_epsilon, "m_epsilon", "Stopping criteria.", MS_NOT_AVAILABLE);
+}
+
+void CSigmoidCalibration::set_maxiter(index_t maxiter)
+{
+	m_maxiter = maxiter;
+}
+
+index_t CSigmoidCalibration::get_maxiter()
+{
+	return m_maxiter;
+}
+
+void CSigmoidCalibration::set_minstep(float64_t minstep)
+{
+	m_minstep = minstep;
+}
+
+float64_t CSigmoidCalibration::get_minstep()
+{
+	return m_minstep;
+}
+
+void CSigmoidCalibration::set_sigma(float64_t sigma)
+{
+	m_sigma = sigma;
+}
+
+float64_t CSigmoidCalibration::get_sigma()
+{
+	return m_sigma;
+}
+
+void CSigmoidCalibration::set_epsilon(float64_t epsilon)
+{
+	m_epsilon = epsilon;
+}
+
+float64_t CSigmoidCalibration::get_epsilon()
+{
+	return m_epsilon;
+}
+
+bool CSigmoidCalibration::fit_binary(
+    CBinaryLabels* predictions, CBinaryLabels* targets)
+{
+	m_sigmoid_as.resize_vector(1);
+	m_sigmoid_bs.resize_vector(1);
+	auto sigmoid_params = CStatistics::fit_sigmoid(
+	    predictions->get_values(), targets->get_labels(), m_maxiter, m_minstep,
+	    m_sigma, m_epsilon);
+
+	m_sigmoid_as[0] = sigmoid_params.a;
+	m_sigmoid_bs[0] = sigmoid_params.b;
+
+	return true;
+}
+
+CBinaryLabels* CSigmoidCalibration::calibrate_binary(CBinaryLabels* predictions)
+{
+	REQUIRE(
+	    m_sigmoid_as.size() == 1,
+	    "Parameters not fitted, which need to be fitted before calibrating.\n");
+	CStatistics::SigmoidParamters params;
+	params.a = m_sigmoid_as[0];
+	params.b = m_sigmoid_bs[0];
+
+	auto probabilities = calibrate_values(predictions->get_values(), params);
+
+	CBinaryLabels* calibrated_predictions = new CBinaryLabels(probabilities);
+
+	return calibrated_predictions;
+}
+
+bool CSigmoidCalibration::fit_multiclass(
+    CMulticlassLabels* predictions, CMulticlassLabels* targets)
+{
+	index_t num_classes = predictions->get_num_classes();
+
+	m_sigmoid_as.resize_vector(num_classes);
+	m_sigmoid_bs.resize_vector(num_classes);
+
+	for (index_t i = 0; i < num_classes; ++i)
+	{
+		auto class_targets = targets->get_binary_for_class(i);
+		auto pred_values = predictions->get_confidences_for_class(i);
+		auto target_labels = class_targets->get_labels();
+
+		auto sigmoid_params = CStatistics::fit_sigmoid(
+		    pred_values, target_labels, m_maxiter, m_minstep, m_sigma,
+		    m_epsilon);
+		m_sigmoid_as[i] = sigmoid_params.a;
+		m_sigmoid_bs[i] = sigmoid_params.b;
+
+		SG_UNREF(class_targets);
+	}
+
+	return true;
+}
+
+CMulticlassLabels*
+CSigmoidCalibration::calibrate_multiclass(CMulticlassLabels* predictions)
+{
+	index_t num_classes = predictions->get_num_classes();
+	index_t num_samples = predictions->get_num_labels();
+	REQUIRE(
+	    m_sigmoid_as.size() == num_classes,
+	    "Parameters not fitted, which need to be fitted before calibrating.\n");
+
+	/** Temporarily stores the probabilities. */
+	SGMatrix<float64_t> confidence_values(num_samples, num_classes);
+
+	for (index_t i = 0; i < num_classes; ++i)
+	{
+		auto class_values = predictions->get_confidences_for_class(i);
+
+		CStatistics::SigmoidParamters sigmoid_params;
+		sigmoid_params.a = m_sigmoid_as[i];
+		sigmoid_params.b = m_sigmoid_bs[i];
+
+		SGVector<float64_t> calibrated_values =
+		    calibrate_values(class_values, sigmoid_params);
+
+		confidence_values.set_column(i, calibrated_values);
+	}
+
+	auto result_labels = new CMulticlassLabels(num_samples);
+	result_labels->allocate_confidences_for(num_classes);
+
+	/** Normalize the probabilities. */
+	for (index_t i = 0; i < num_samples; ++i)
+	{
+		SGVector<float64_t> values = confidence_values.get_row_vector(i);
+		float64_t sum = SGVector<float64_t>::sum(values);
+
+		/** All classes have equal probability when sum is zero */
+		if (sum == 0)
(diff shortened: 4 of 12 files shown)


durovo:
The SIGSEV on SGObject seems to be due to some code in upstream develop.

src/shogun/evaluation/Calibration.cpp
```diff
@@ -0,0 +1,76 @@
...
+ * policies,
+ * either expressed or implied, of the Shogun Development Team.
+ */
+
+#include <shogun/evaluation/Calibration.h>
+#include <shogun/machine/Machine.h>
+
+using namespace shogun;
+
+CCalibration::CCalibration()
```
vigsterkr: this seems to be an abstract class, as nothing is implemented.
why dont you just drop the implementation and specify all the virtual methods pure virtual?

src/shogun/evaluation/SigmoidCalibration.h
```diff
@@ -0,0 +1,112 @@
...
+
+		/** Helper function that calibrates values of given vector using the given sigmoid parameters
+		* @param values The values to be calibrated
+		* @param params The sigmoid paramters to be used for calibration
+		*/
+		SGVector<float64_t> calibrate_values(SGVector<float64_t> values, CStatistics::SigmoidParamters params);
+
+	private:
+		/** Array to store sigmoid parameters for each class. In case of binary labels, only one pair of parameters are stored. */
+		CStatistics::SigmoidParamters* m_sigmoid_parameters;
```
vigsterkr: wouldn't storing this in an std::vector would be better?

vigsterkr: or in worst case DynamicObjectArray, for assuring ref counting

src/shogun/evaluation/SigmoidCalibration.cpp
```diff
@@ -0,0 +1,158 @@
...
+
+	return calibrated_predictions;
+}
+
+bool CSigmoidCalibration::fit_multiclass(CMulticlassLabels* predictions, CMulticlassLabels* targets)
+{
+	index_t num_classes =
+	    predictions->get_num_classes();
+
+	SG_FREE(m_sigmoid_parameters);
```
vigsterkr: as said i would rather use something more high level than this

src/shogun/mathematics/Statistics.cpp
```diff
@@ -814,6 +814,171 @@ SGMatrix<float64_t> CStatistics::sample_from_gaussian(SGVector<float64_t> mean,
...
+
+	REQUIRE(scores.vector, "CStatistics::fit_sigmoid() requires "
+			"scores vector!\n");
+
+	/* count prior0 and prior1 if needed */
+	int32_t prior0=0;
+	int32_t prior1=0;
+	SG_SDEBUG("counting number of positive and negative labels\n")
+	{
+		for (index_t i=0; i<scores.vlen; ++i)
```
vigsterkr: isn't it enough to count the one of them as the other can be simply deducted?
i would use something like `std::count_if` or `std::transform`

src/shogun/mathematics/Statistics.cpp
```diff
@@ -814,6 +814,171 @@ SGMatrix<float64_t> CStatistics::sample_from_gaussian(SGVector<float64_t> mean,
...
+				prior1++;
+			else
+				prior0++;
+		}
+	}
+	SG_SDEBUG("%d pos; %d neg\n", prior1, prior0)
+
+	/* parameter setting */
+	/* maximum number of iterations */
+	index_t maxiter=100;
```
vigsterkr: all these parameters should be set somewhere else... as you want to be able to control them without actually changing the code... i.e. there should be a getter/setter for them

src/shogun/mathematics/Statistics.cpp
```diff
@@ -814,6 +814,171 @@ SGMatrix<float64_t> CStatistics::sample_from_gaussian(SGVector<float64_t> mean,
...
+	float64_t sigma=1E-12;
+	float64_t eps=1E-5;
+
+	/* construct target support */
+	float64_t hiTarget=(prior1+1.0)/(prior1+2.0);
+	float64_t loTarget=1/(prior0+2.0);
+	index_t length=prior1+prior0;
+
+	SGVector<float64_t> t(length);
+	for (index_t i=0; i<length; ++i)
```
vigsterkr: same as above... too vanilla implementation

src/shogun/mathematics/Statistics.cpp
durovo: Yes, counting one of them would be enough. I using `std::count_if` would indeed be neat.

src/shogun/mathematics/Statistics.cpp
durovo: I am not the one who originally implemented, I merely made a few corrections in `fit_sigmoid`. However, the code can certainly be improved. I'm on it.

src/shogun/evaluation/SigmoidCalibration.h
durovo: I will replace this with `std::vector`... DynamicObjectArray seems to support only CSGObjects, it can't hold structs like `CStatistics::SigmoidParamters`. Please correct me if I'm wrong about that :)

src/shogun/evaluation/Calibration.cpp
durovo: I was thinking of adding common `fit` and `calibrate` methods that would call `fit_binary`, `fit_multiclass` etc as required

durovo:
Thanks for the review! I will try and make all the requested changes.

src/shogun/evaluation/SigmoidCalibration.cpp
durovo: Understood

durovo:
@vigsterkr While making all the methods pure virtual would be nice, wouldn't we lose the informative messages generated by `SG_NOTIMPLEMENTED`? Suppose, the implementation class only want to use `fit_binary` wouldn't it be useful to have the message generated by `SG_NOTIMPLEMENTED` when `fit_multiclass` is called?

src/shogun/evaluation/SigmoidCalibration.h
```diff
@@ -94,6 +94,46 @@ namespace shogun
 		**/
 		virtual CMulticlassLabels* calibrate_multiclass(CMulticlassLabels* predictions);
 
+	 	/** Set maximum number of iterations
+	 	* @param maxiter maximum number of iterations
+	 	*/
+	 	virtual void set_maxiter(index_t maxiter);
```
vigsterkr: why are these all virtual? you will have a calibration method that is based on Sigmoid?

src/shogun/evaluation/SigmoidCalibration.cpp
```diff
@@ -42,28 +42,75 @@
 
 using namespace shogun;
 
-CSigmoidCalibration::CSigmoidCalibration() : CCalibration()
+CSigmoidCalibration::CSigmoidCalibration()
```
vigsterkr: it should call the ctor of `CCalibration`

src/shogun/evaluation/Calibration.h
```diff
@@ -64,26 +70,44 @@ namespace shogun
 		* @param targets The true labels corresponding to the predictions
 		* @return boolean indicating whether the calibration was succesful
 		**/
-		virtual bool fit_binary(CBinaryLabels* predictions, CBinaryLabels* targets);
+		virtual bool fit_binary(CBinaryLabels* predictions, CBinaryLabels* targets)
```
vigsterkr: i still dont get why do we need default implementation with error

src/shogun/evaluation/SigmoidCalibration.cpp
```diff
@@ -42,28 +42,75 @@
...
 
 void CSigmoidCalibration::init()
 {
-	m_sigmoid_parameters = SG_MALLOC(CStatistics::SigmoidParamters, 1);
+	m_sigmoid_parameters.resize(1);
+	m_maxiter = 100;
+	m_minstep = 1E-10;
+	m_sigma = 1E-12;
+	m_epsilon = 1E-5;
+
```
vigsterkr: `m_sigmoid_parameters` is not registered... that'll make it impossible to clone this class. since currently registering std::vector is not possible you would need to use DynArray :(

src/shogun/evaluation/SigmoidCalibration.h
durovo: Probably not. Indeed there is no need to make them virtual.

src/shogun/evaluation/SigmoidCalibration.cpp
durovo: Thanks for telling me about `DynArray`, I was looking for the standard shogun way of storing arrays of non-CSGObjects.

src/shogun/evaluation/SigmoidCalibration.cpp
durovo: Yes, my bad

src/shogun/evaluation/Calibration.h
durovo: I'm not sure about making all methods pure virtual... suppose a calibration method does not support multiclass calibration, then the implementation will have to derive `fit_multiclass` and give error message manually. Would that be fine?

src/shogun/evaluation/Calibration.h
```diff
@@ -0,0 +1,113 @@
...
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE
+ * FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
```
karlnapf: could you use the new (shorter) bsd header? see some other files

durovo: This one? /*
 *
* Authors: 
*/

src/shogun/evaluation/SigmoidCalibration.cpp
durovo: I have opted to use two SGVectors for now

src/shogun/evaluation/Calibration.h
karlnapf: yes

tests/unit/evaluation/SigmoidCalibration_unittest.cc
```diff
@@ -0,0 +1,99 @@
+/*
+ * Copyright (c) The Shogun Machine Learning Toolbox
+ * Written (w) 2012 - 2013 Heiko Strathmann
+ * Written (w) 2018 Dhruv Arya
+ * All rights reserved.
+ *
```
karlnapf: we can also use the small license here

src/shogun/mathematics/Statistics.h
```diff
@@ -296,6 +296,29 @@ class CStatistics: public CSGObject
...
+	 * @param minstep minimum step taken in line search
+	 * @param sigma set to a value greater than zero to ensure that Hessian matrix is positive semi-definite
+	 * @param eps stopping criteria
+	 * @return struct containing the sigmoid's shape parameters a and b
+	 */
+	static SigmoidParamters fit_sigmoid(SGVector<float64_t> scores, SGVector<float64_t> labels,
+		index_t maxiter = 100, float64_t minstep = 1E-10,
+		float64_t sigma = 1E-12, float64_t eps = 1E-5);
+
+	/** TODO: Remove this method after fixing MuticlassCalibration
```
karlnapf: pls no todo in patches.

src/shogun/mathematics/Statistics.cpp
```diff
@@ -814,6 +815,157 @@ SGMatrix<float64_t> CStatistics::sample_from_gaussian(SGVector<float64_t> mean,
...
+		{
+			SG_SWARNING("CStatistics::fit_sigmoid(): line search fails, A=%f, "
+					"B=%f, g1=%f, g2=%f, dA=%f, dB=%f, gd=%f\n",
+					a, b, g1, g2, dA, dB, gd);
+		}
+	}
+
+	if (it>=maxiter-1)
+	{
+		SG_SWARNING("CStatistics::fit_sigmoid(): reaching maximal iterations,"
```
karlnapf: we dont do this ClassName::MethodName(): anymore. Just the message

src/shogun/mathematics/Statistics.cpp
```diff
@@ -814,6 +815,157 @@ SGMatrix<float64_t> CStatistics::sample_from_gaussian(SGVector<float64_t> mean,
...
+		if (labels[i]>0)
+			t[i]=hiTarget;
+		else
+			t[i]=loTarget;
+	}
+
+	/* initial Point and Initial Fun Value */
+	/* result parameters of sigmoid */
+	float64_t a=0;
+	float64_t b=CMath::log((prior0+1.0)/(prior1+1.0));
```
karlnapf: where is this code coming from?

src/shogun/mathematics/Statistics.cpp
```diff
@@ -814,6 +815,157 @@ SGMatrix<float64_t> CStatistics::sample_from_gaussian(SGVector<float64_t> mean,
...
+	float64_t fval=0.0;
+
+	for (index_t i=0; i<length; ++i)
+	{
+		float64_t fApB=scores[i]*a+b;
```
karlnapf: if this code is copy pasted, fine. Otherwise, can we rename this variable? :)

src/shogun/mathematics/Statistics.h
durovo: Understood

tests/unit/evaluation/SigmoidCalibration_unittest.cc
durovo: Okay

src/shogun/mathematics/Statistics.cpp
durovo: It's all from the old 'fit_sigmoid'

src/shogun/mathematics/Statistics.cpp
durovo: It is mostly copy-pasted. Should I rename it?

src/shogun/mathematics/Statistics.cpp
durovo: Okay

durovo:
@karlnapf I have added a test for multiclass calibration.

src/shogun/mathematics/Statistics.cpp
durovo: This code is mostly taken from the previous `fit_sigmoid`. Do you recommend changing all the messages?

karlnapf: yeah let's modernize a bit :)

src/shogun/mathematics/Statistics.cpp
karlnapf: maybe leave it so it is clear where the code is coming from

src/shogun/evaluation/Calibration.h
```diff
@@ -0,0 +1,64 @@
...
+#ifndef _CALIBRATION_H__
+#define _CALIBRATION_H__
+
+#include <shogun/lib/config.h>
+
+#include <shogun/machine/Machine.h>
+
+namespace shogun
+{
+	/** @brief Base class for all calibration methods. Call fit to
```
karlnapf: Can you tell doxygen that "fit" is a method here?

src/shogun/evaluation/Calibration.h
```diff
@@ -0,0 +1,64 @@
...
+
+		virtual const char* get_name() const
+		{
+			return "Calibration";
+		}
+
+		/** Fit calibration parameters for binary labels.
+		* @param predictions The predictions outputted by the machine
+		* @param targets The true labels corresponding to the predictions
+		* @return boolean indicating whether the calibration was succesful
```
karlnapf: minor: no need to state the type in return, c++ tell you that already

src/shogun/labels/MulticlassLabels.cpp
```diff
@@ -71,6 +71,19 @@ void CMulticlassLabels::allocate_confidences_for(int32_t n_classes)
 	m_multiclass_confidences = SGMatrix<float64_t>(n_classes,n_labels);
 }
 
+SGVector<float64_t> CMulticlassLabels::get_confidences_for_class(int32_t i)
+{
+	REQUIRE((m_multiclass_confidences.num_rows != 0) && 
+		(m_multiclass_confidences.num_cols != 0),
+		"Call allocate_confidences_for() before fetching confidences");
```
karlnapf: missing newline

karlnapf: I woudl also suggest to change the error msg
"Empty confidences, which need to be allocated before fetching.\n")

tests/unit/evaluation/SigmoidCalibration_unittest.cc
```diff
@@ -0,0 +1,149 @@
...
+	auto calib_result = calibration_method->calibrate_multiclass(predictions);
+
+ // 	 [[0.75674645, 0.20144443, 0.04180912],
+ //       [0.82689269, 0.05965391, 0.11345341],
+ //       [0.81409222, 0.12715657, 0.05875121],
+ //       [0.48765218, 0.38500051, 0.12734731],
+ //       [0.31003506, 0.60319518, 0.08676976],
+ //       [0.26630134, 0.01476002, 0.71893864],
+ //       [0.72470815, 0.20387599, 0.07141586],
+ //       [0.074445  , 0.80897488, 0.11658012],
```
karlnapf: could you remove those?

tests/unit/evaluation/SigmoidCalibration_unittest.cc
```diff
@@ -0,0 +1,149 @@
...
+ //       [0.20620103, 0.71452701, 0.07927195],
+ //       [0.16983872, 0.72947987, 0.10068142]]
+
+	float64_t expected_probs[] = {
+	   0.75674645, 0.20144443, 0.04180912,
+       0.82689269, 0.05965391, 0.11345341,
+       0.81409222, 0.12715657, 0.05875121,
+       0.48765218, 0.38500051, 0.12734731,
```
karlnapf: where do the numbers in the tests come from?

src/shogun/labels/MulticlassLabels.cpp
durovo: Okay, that looks much better :)

tests/unit/evaluation/SigmoidCalibration_unittest.cc
durovo: Sure

tests/unit/evaluation/SigmoidCalibration_unittest.cc
durovo: I used trained a linear SVC in sklearn, this is the decision function of that SVC.

src/shogun/evaluation/Calibration.h
durovo: Understood.

durovo:
@karlnapf I have made all the requested changes. Code style fix should be done after this has been approved right?

tests/unit/evaluation/SigmoidCalibration_unittest.cc
karlnapf: thats good! Could maybe be added as a comment in there. But nevermind for now

examples/undocumented/libshogun/classifier_libsvm_probabilities.cpp
```diff
@@ -73,16 +75,19 @@ int main(int argc, char** argv)
     using the method described in Lin, H., Lin, C., and Weng,  R. (2007). A note
     on Platt's probabilistic outputs for support vector machines.
     See BinaryLabels documentation for details*/
-    out_labels->scores_to_probabilities();
+    CSigmoidCalibration* sigmoid_calibration = new CSigmoidCalibration();
```
karlnapf: is this deleted anywhere? SG_REF?

examples/undocumented/libshogun/labels_binary_fit_sigmoid.cpp
```diff
@@ -6,22 +6,28 @@
...
 #include <shogun/base/init.h>
 #include <shogun/labels/BinaryLabels.h>
+#include <shogun/evaluation/SigmoidCalibration.h>
 
 using namespace shogun;
 
 void test_sigmoid_fitting()
 {
 	CBinaryLabels* labels=new CBinaryLabels(10);
+	CBinaryLabels* predictions=new CBinaryLabels(10);
```
karlnapf: deleted somewhere?

examples/undocumented/libshogun/labels_binary_fit_sigmoid.cpp
```diff
@@ -6,22 +6,28 @@
...
-
+	{
+		predictions->set_value(i%2==0 ? 1 : -1, i);
+		labels->set_value(i%4==0 ? 1: -1, i);
+	}
+	predictions->get_values().display_vector("scores");
+	CSigmoidCalibration* sigmoid_calibration = new CSigmoidCalibration();
+	sigmoid_calibration->fit_binary(predictions, labels);
+	CBinaryLabels* calibrated_labels = sigmoid_calibration->calibrate_binary(predictions);
+	calibrated_labels->get_values().display_vector("probabilities");
```
karlnapf: why display?

src/shogun/evaluation/SigmoidCalibration.cpp
karlnapf: Think that should be OK as a workaround for now, especially since the number of parameter is fixed

karlnapf: actually, even better would be to put the parameters explicitly. The vector length should not be variable

src/shogun/evaluation/SigmoidCalibration.cpp
```diff
@@ -0,0 +1,188 @@
...
+CBinaryLabels*
+CSigmoidCalibration::calibrate_binary(CBinaryLabels* predictions)
+{
+	CStatistics::SigmoidParamters params; 
+	params.a = m_sigmoid_as[0];
+	params.b = m_sigmoid_bs[0];
+
+	/** Convert predictions to probabilties. */
+	auto values = calibrate_values(predictions->get_values(), params);
```
karlnapf: rather than the comment, I would name the variable `probabilities` and then you dont need it. Minor of course

src/shogun/evaluation/SigmoidCalibration.cpp
```diff
@@ -0,0 +1,188 @@
...
+
+bool CSigmoidCalibration::fit_multiclass(CMulticlassLabels* predictions, CMulticlassLabels* targets)
+{
+	index_t num_classes =
+	    predictions->get_num_classes();
+
+	m_sigmoid_as.resize_vector(num_classes);
+	m_sigmoid_bs.resize_vector(num_classes);
+
+	 /** Fit and store parameters for for each class seperately. */
```
karlnapf: no need for this comment

src/shogun/evaluation/SigmoidCalibration.cpp
```diff
@@ -0,0 +1,188 @@
...
+	return true;
+}
+
+CMulticlassLabels*
+CSigmoidCalibration::calibrate_multiclass(CMulticlassLabels* predictions)
+{
+	index_t num_classes = predictions->get_num_classes();
+	index_t num_samples = predictions->get_num_labels();
+
+	/** Matrix to temporarily store the probabilities. */
```
karlnapf: I would never write the type of a variable in a comment, but minor

src/shogun/evaluation/SigmoidCalibration.cpp
```diff
@@ -0,0 +1,188 @@
...
+
+/** Normalize the probabilities. */
+	for (index_t i = 0; i < num_samples; ++i)
+	{
+		SGVector<float64_t> values =
+		    confidence_values.get_row_vector(i);
+		float64_t sum = SGVector<float64_t>::sum(values);
+
+		/** All classes have equal probability when sum is zero */
+		if (sum == 0)
```
karlnapf: good catch!

src/shogun/evaluation/SigmoidCalibration.h
```diff
@@ -0,0 +1,133 @@
...
+#include <shogun/lib/config.h>
+
+#include <shogun/evaluation/Calibration.h>
+#include <shogun/mathematics/Statistics.h>
+
+
+namespace shogun
+{
+	/** @brief Calibrates labels based on Platt Scaling [1]. Note that first calibration parameters need to be fitted by 
+	* calling fit_binary() or fit_multiclass(). Then call calibrate on calibrate_binary() or calibrate_multiclass() on
```
karlnapf: what will happen if a user forgets to call fit first?

src/shogun/evaluation/SigmoidCalibration.h
```diff
@@ -0,0 +1,133 @@
...
+		void init();
+
+		/** Helper function that calibrates values of given vector using the given sigmoid parameters
+		* @param values The values to be calibrated
+		* @param params The sigmoid paramters to be used for calibration
+		*/
+		SGVector<float64_t> calibrate_values(SGVector<float64_t> values, CStatistics::SigmoidParamters params);
+
+	private:
+		/** Vector to store parameter A of sigmoid for each class. In case of binary labels, only one pair of parameters are stored. */
```
karlnapf: no type in the comments or doxygen, it is clear from the code

src/shogun/mathematics/Statistics.cpp
```diff
@@ -814,6 +815,156 @@ SGMatrix<float64_t> CStatistics::sample_from_gaussian(SGVector<float64_t> mean,
 	return S;
 }
 
+CStatistics::SigmoidParamters CStatistics::fit_sigmoid(SGVector<float64_t> scores, SGVector<float64_t> labels,
+	index_t maxiter, float64_t minstep,
+	float64_t sigma, float64_t epsilon)
+{
+	SG_SDEBUG("entering CStatistics::fit_sigmoid()\n")
+ 
+	REQUIRE(scores.vector, "Specify scores vector!\n");
```
karlnapf: "Provided scores are empty.\n"

src/shogun/mathematics/Statistics.cpp
```diff
@@ -814,6 +815,156 @@ SGMatrix<float64_t> CStatistics::sample_from_gaussian(SGVector<float64_t> mean,
 	return S;
 }
 
+CStatistics::SigmoidParamters CStatistics::fit_sigmoid(SGVector<float64_t> scores, SGVector<float64_t> labels,
+	index_t maxiter, float64_t minstep,
+	float64_t sigma, float64_t epsilon)
+{
+	SG_SDEBUG("entering CStatistics::fit_sigmoid()\n")
```
karlnapf: remove

src/shogun/evaluation/SigmoidCalibration.cpp
durovo: Okay

src/shogun/mathematics/Statistics.cpp
karlnapf: yes, can't this be done a bit more modern?

src/shogun/evaluation/SigmoidCalibration.cpp
durovo: That would look much better indeed

tests/unit/evaluation/SigmoidCalibration_unittest.cc
```diff
@@ -0,0 +1,138 @@
...
+	SG_REF(predictions)
+
+	CSigmoidCalibration* sigmoid_calibration = new CSigmoidCalibration();
+
+	auto calibrated = sigmoid_calibration->fit_binary(predictions, labels);
+
+	EXPECT_EQ(calibrated, true);
+
+	auto calibrated_labels = sigmoid_calibration->calibrate_binary(predictions);
+
```
karlnapf: why all the empty lines?

examples/undocumented/libshogun/labels_binary_fit_sigmoid.cpp
durovo: I'm not sure myself. It was in the original code https://github.com/shogun-toolbox/shogun/blob/297f9a0c5d4771154515fef2fac865fdc1a1af5e/examples/undocumented/libshogun/labels_binary_fit_sigmoid.cpp#L22

src/shogun/evaluation/SigmoidCalibration.h
durovo: Should it the first line be something like: "Stores parameter A of sigmoid for each class"?

src/shogun/mathematics/Statistics.cpp
vigsterkr: you can either use `std::transform` or at least use the fact that SGVector has iterator so it can actually do a range-loop

src/shogun/mathematics/Statistics.cpp
durovo: Understood.

examples/undocumented/libshogun/labels_binary_fit_sigmoid.cpp
karlnapf: I see. Not your fault of course. Would still be better removed :)
In fact, what would be best was to remove the example altogether (another PR) and have a meta example for that

src/shogun/evaluation/SigmoidCalibration.h
karlnapf: Yes, but that is really me nitpicking. Just for the future

src/shogun/evaluation/SigmoidCalibration.cpp
durovo: Yeah, it seems does seem very redundant. Will change this.

src/shogun/mathematics/Statistics.cpp
durovo: Okay

examples/undocumented/libshogun/classifier_libsvm_probabilities.cpp
durovo: I missed this, sorry for that.

tests/unit/evaluation/SigmoidCalibration_unittest.cc
durovo: Not required? Will remove them.

src/shogun/mathematics/Statistics.cpp
durovo: Will do, thanks for the suggestion

durovo: Will do, thanks for the suggestion

src/shogun/evaluation/SigmoidCalibration.h
durovo: The length of the parameter vector is checked in calibrate now and an appropriate message is displayed.

src/shogun/mathematics/Statistics.cpp
durovo: @vigsterkr I opted to use `std::transform`

durovo:
@vigsterkr @karlnapf I have made the requested changes. Should I fix the style, rebase and push?

karlnapf:
Yes! style fix and let's merge. Nice work!

karlnapf:
:tada:

PR #4167, opened 2018-02-10
Use linalg in place of Eigen in LMNN

vinx13:
See #2747

### DIFF
--- a/src/shogun/metric/LMNN.cpp
+++ b/src/shogun/metric/LMNN.cpp
@@ -8,14 +8,10 @@
 
 #include <shogun/base/progress.h>
 #include <shogun/mathematics/Math.h>
+#include <shogun/mathematics/linalg/LinalgNamespace.h>
 #include <shogun/metric/LMNNImpl.h>
 
-// useful shorthand to perform operations with Eigen matrices
-// trace of the product of two matrices computed fast using trace(A*B)=sum(A.*B')
-#define	TRACE(A,B)		(((A).array()*(B).transpose().array()).sum())
-
 using namespace shogun;
-using namespace Eigen;
 
 CLMNN::CLMNN()
 {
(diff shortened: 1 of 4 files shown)


vigsterkr:
great stuff! although LMNN some parts could get some openmp stuff but that could be the next pr

vigsterkr:
i've just restarted the travis jobs to see whether that was a hiccup or there's an actual bug now :(

vinx13:
there might be an actual bug, previous commit (d1763b8) has the same failure on [travis](https://travis-ci.org/shogun-toolbox/shogun/jobs/339385159#L5397)

vigsterkr:
doh.... that's definitely that :((( see the output of our buildbot
http://buildbot.shogun-toolbox.org:8080/#/console

@karlnapf @lisitsyn d1763b83fbc029bae1487201ef56977efdfe43a0 broke builds

src/shogun/metric/LMNN.cpp
```diff
@@ -109,8 +104,13 @@ void CLMNN::train(SGMatrix<float64_t> init_transform)
 
 		// Compute the objective, trace of Mahalanobis distance matrix (L squared) times the gradient
 		// plus the number of current impostors to account for the margin
+		// trace of the product of two matrices computed fast using
+		// trace(A*B)=sum(A.*B')
 		SG_DEBUG("Computing objective.\n")
-		obj[iter] = TRACE(L.transpose()*L,gradient) + m_regularization*cur_impostors.size();
+		obj[iter] = m_regularization * cur_impostors.size();
```
karlnapf: but this is slower than the old version :(
Maybe introduce this macro as a function in linalg? `trace_dot` or so?

src/shogun/metric/LMNNImpl.cpp
```diff
@@ -131,49 +124,46 @@ SGMatrix<index_t> CLMNNImpl::find_target_nn(CDenseFeatures<float64_t>* x,
 	return target_neighbors;
 }
 
-MatrixXd CLMNNImpl::sum_outer_products(CDenseFeatures<float64_t>* x, const SGMatrix<index_t> target_nn)
+SGMatrix<float64_t> CLMNNImpl::sum_outer_products(
```
karlnapf: I think this method could be pulled into linalg as it is useful for other situations as well.

src/shogun/metric/LMNNImpl.cpp
```diff
@@ -131,49 +124,46 @@ SGMatrix<index_t> CLMNNImpl::find_target_nn(CDenseFeatures<float64_t>* x,
...
 
 	// sum the outer products stored in C using the indices specified in target_nn
 	for (index_t i = 0; i < target_nn.num_cols; ++i)
 	{
 		for (index_t j = 0; j < target_nn.num_rows; ++j)
 		{
-			VectorXd dx = X.col(i) - X.col(target_nn(j,i));
-			sop += dx*dx.transpose();
+			auto dx = linalg::add(
+			    X.get_column(i), X.get_column(target_nn(j, i)), 1.0, -1.0);
```
karlnapf: actually yes definitely pull it out to hide this code here, messy

PR #4168, opened 2018-02-11
Add meta example for ChiSquare Distance

syashakash:
Refer #3555 
All tests passed.

### DIFF
--- a/examples/meta/src/distance/chi_square.sg
+++ b/examples/meta/src/distance/chi_square.sg
@@ -1,0 +1,20 @@
+CSVFile f_feats_a("../../data/fm_train_real.dat")
+CSVFile f_feats_b("../../data/fm_test_real.dat")
+
+#![create_features]
+RealFeatures features_a(f_feats_a)
+RealFeatures features_b(f_feats_b)
+#![create_features]
+
+#![create_instance]
+ChiSquareDistance distance(features_a, features_a)
+#![create_instance]
+
+#![extract_distance]
+RealMatrix distance_matrix_aa = distance.get_distance_matrix()
+#![extract_distance]
+
+#![refresh_distance]
+distance.init(features_a, features_b)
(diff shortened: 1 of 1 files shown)


syashakash:
unit-SGObject encounters Segmentation Fault, but the test works fine locally. I would get back to this later this week. I can't figure out what could possibly have gone wrong!!

syashakash:
Added data file for this example.

syashakash:
Added data revision to this PR.

karlnapf:
Thanks!
We will need to wait with merging this, as there currently is a problem with the internal tests that needs fixing first (and likely update of the data file).

syashakash:
Agreed :)

vigsterkr:
@syashakash could you rebase this with the latest develop as the integration test is back.

syashakash:
Ok, I see that the integration tests have failed! I would see how to solve this.

examples/meta/src/distance/chisquare.sg
```diff
@@ -0,0 +1,20 @@
+CSVFile f_feats_a("../../data/fm_train_real.dat")
```
karlnapf: pls rename the file to chi_square

syashakash:
@karlnapf @vigsterkr this example fails the new binary kernel SVM integration tests for ChiSquareDistance. The example's code is similar to that of other `distance` examples. I can't understand where the code fails as it does what it is supposed to do.

karlnapf:
The integration test fails if the listing produces different data than the one in the reference data. See the examples and the development readme

karlnapf:
Shall we get this merged?

syashakash:
@karlnapf the integration tests fail locally. Moreover, the data file needs another update, and this PR has wrong file.

karlnapf:
Sure, wanna go ahead with that?

karlnapf:
There now also is a merge conflict, so you will need to update this PR

PR #4176, opened 2018-02-15
Add trace_dot in linalg

vinx13:
* Add method that computes trace of the product of two matrixes in linalg
* Use `trace_dot` in LMNN

### DIFF
--- a/src/shogun/mathematics/linalg/LinalgNamespace.h
+++ b/src/shogun/mathematics/linalg/LinalgNamespace.h
@@ -1691,6 +1691,19 @@ namespace shogun
 		}
 
 		/**
+		 * Method that computes the trace of \f$AB\f$ as \f$sum(A.*B')$\f
+		 *
+		 * @param A The matrix A
+		 * @param B The matrix B
+		 * @return The trace of the product of A and B
+		 */
+		template <typename T>
+		T trace_dot(const SGMatrix<T>& A, const SGMatrix<T>& B)
+		{
+			return sum(element_prod(A, transpose_matrix(B)));
+		}
+
+		/**
 		 * Solve the linear equations \f$Lx=b\f$,
 		 * where \f$L\f$ is a triangular matrix.
 		 *
--- a/src/shogun/mathematics/linalg/LinalgNamespace.h
+++ b/src/shogun/mathematics/linalg/LinalgNamespace.h
@@ -1700,7 +1700,7 @@ namespace shogun
 		template <typename T>
 		T trace_dot(const SGMatrix<T>& A, const SGMatrix<T>& B)
 		{
-			return sum(element_prod(A, transpose_matrix(B)));
+			return sum(element_prod(A, B, false, true));
 		}
 
 		/**
--- a/src/shogun/metric/LMNN.cpp
+++ b/src/shogun/metric/LMNN.cpp
@@ -104,13 +104,10 @@ void CLMNN::train(SGMatrix<float64_t> init_transform)
 
 		// Compute the objective, trace of Mahalanobis distance matrix (L squared) times the gradient
 		// plus the number of current impostors to account for the margin
-		// trace of the product of two matrices computed fast using
-		// trace(A*B)=sum(A.*B')
 		SG_DEBUG("Computing objective.\n")
 		obj[iter] = m_regularization * cur_impostors.size();
-		obj[iter] += linalg::trace(
-		    linalg::element_prod(
-		        linalg::matrix_prod(L, L, true, false), gradient));
+		obj[iter] +=
+		    linalg::trace_dot(linalg::matrix_prod(L, L, true, false), gradient);
 
 		// Correct step size
 		CLMNNImpl::correct_stepsize(stepsize, obj, iter);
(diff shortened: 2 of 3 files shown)


src/shogun/metric/LMNN.cpp
```diff
@@ -104,13 +104,10 @@ void CLMNN::train(SGMatrix<float64_t> init_transform)
...
 		// Compute the objective, trace of Mahalanobis distance matrix (L squared) times the gradient
 		// plus the number of current impostors to account for the margin
-		// trace of the product of two matrices computed fast using
-		// trace(A*B)=sum(A.*B')
 		SG_DEBUG("Computing objective.\n")
 		obj[iter] = m_regularization * cur_impostors.size();
-		obj[iter] += linalg::trace(
-		    linalg::element_prod(
-		        linalg::matrix_prod(L, L, true, false), gradient));
+		obj[iter] +=
```
karlnapf: Is this the same as before? Doesnt seem to me but maybe I am blind :)

src/shogun/mathematics/linalg/LinalgNamespace.h
```diff
@@ -1681,6 +1681,19 @@ namespace shogun
...
+		 * Method that computes the trace of \f$AB\f$ as \f$sum(A.*B')$\f
+		 *
+		 * @param A The matrix A
+		 * @param B The matrix B
+		 * @return The trace of the product of A and B
+		 */
+		template <typename T>
+		T trace_dot(const SGMatrix<T>& A, const SGMatrix<T>& B)
+		{
+			return sum(element_prod(A, transpose_matrix(B)));
```
karlnapf: is transpose matrix changing the memory (it shouldnt)

Also, the API seems inconsistent, for some methods we pass flags for whether matrices should be transposed. And for others, this flag doesnt exist and we have to transpose matrices from the outside ?

src/shogun/metric/LMNN.cpp
vinx13: They are different. Actually I was wrong in #4167. Now it is the same as the original one.  https://github.com/shogun-toolbox/shogun/blob/431e56b67344e0662444c0efbd4aefde47990fa4/src/shogun/metric/LMNN.cpp#L113

src/shogun/mathematics/linalg/LinalgNamespace.h
vinx13: transpose matrix creates a new matrix.  
The API is indeed inconsistent, maybe we should add transpose flags to element_prod ?

karlnapf: yep!

src/shogun/metric/LMNN.cpp
karlnapf: EHM!!!
Why did the tests pass? Or are there none?

vinx13: Well, here the trace is used as the termination condition. Seems that it doesn't stop until reaching maximum number of iterations when trace is wrong in tests.

karlnapf: I see now! makes sense
Ok, this definitely should be tested somehow (at least in a unit test)
Wanna do send another PR for that?

karlnapf:
What is the status of this one then? :)

src/shogun/mathematics/linalg/LinalgNamespace.h
vinx13: @karlnapf I made another pr #4183 , with which I could make it clear here

vinx13:
This one would be ready for review after cleaning up calls to transpose_matrix.

karlnapf:
ok ping me here once that is the case

vinx13:
@karlnapf Ready for review now :)

src/shogun/metric/LMNN.cpp
karlnapf: integration tests are back, so now let's see :)

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
```diff
@@ -1734,6 +1734,23 @@ TEST(LinalgBackendEigen, SGMatrix_trace)
 	EXPECT_NEAR(trace(A), tr, 1e-15);
 }
 
+TEST(LinalgBackendEigen, SGMatrix_tract_dot)
```
karlnapf: typo

karlnapf:
typo fix and travis green and we can merge this :)

karlnapf:
restarted the hickup builds...should be fine soon

src/shogun/metric/LMNN.cpp
iglesias: Hey guys. Why isn't the same as before? I don't see it :-/ Or maybe this discussion is about a previous diff?

karlnapf: Check the history of that line, it changed back and forth IIRC

iglesias: I checked and only found the three mentioned here: the original with TRACE plust the two in this diff.

vinx13: The original with TRACE is correct, the former one in this diff is wrong.

iglesias: Where is the difference between them, @vinx13?

vinx13: We are calculating trace of L' * L * gradient, which is sum( (L'*L) .* gradient' ), instread of trace( (L'*L) .* gradient) that appears in the previous commit

Issue #4049, opened 2017-12-25
Clean up xvalidation API


karlnapf:
* Let's make the splitting strategy an iterator
 * let's rename `generate_subset_indices` and `generate_subset_inverse` to something like `get_train` and `get_validation` or even better `train` and `validation` 
 * Let's get rid of the `build_subsets` method (and do it in the constructor or lazily when the first index set is requested)

Probably more to come. Have a look at the sklearn splittings at e.g. [here](http://scikit-learn.org/stable/modules/generated/sklearn.model_selection.KFold.html)
See also #3965

karlnapf:
Good example for time-series splitting based on simple iterator approach
https://github.com/scikit-learn/scikit-learn/blob/ea6ea815e1071bfb9de18953fea98b136a7fa8ed/sklearn/model_selection/_split.py#L770

braceletboy:
Hey!! Just to confirm, we are supposed to re-implement the CSplittingStrategy base class with changes mentioned in second, third and fourth points of the first comment right? I am actually a little confused by the first point of that comment.

braceletboy:
I wanna take this up.

karlnapf:
Go for it

luisffranca:
Hi @karlnapf !

By "make the splitting strategy an iterator" you mean creating a method (like _split_ on sklearn's example) that would return iterators for the training (and test) set indices for that subset index? Such a method would internally call generate_subset_indices and generate_subset_inverse (with their new names as you suggested). Is my understanding correct?

karlnapf:
An iterator would be a next step, much cleaner, yes

PR #4180, opened 2018-02-19
Changing splitting strategy to clean up Cross Validation API

luisffranca:
Changing splitting strategy to clean up Cross Validation API (#4049)

- generate_subset_indices() and generate_subset_inverse() are renamed,
respectively, to validation() and train();
- The methods are now const and return const vectors;
- build_subsets() won't be explicitly called, but rather internally
in the constructor of the derived classes;
- Examples, unit tests and notebooks were changed accordingly.


src/shogun/evaluation/CrossValidationSplitting.cpp
```diff
@@ -23,7 +23,7 @@ CCrossValidationSplitting::CCrossValidationSplitting(
 	m_rng = sg_rand;
 }
 
-void CCrossValidationSplitting::build_subsets()
+void CCrossValidationSplitting::build_subsets() const
```
karlnapf: this method cannot be const, it changes members

src/shogun/evaluation/SplittingStrategy.cpp
```diff
@@ -33,7 +33,7 @@ CSplittingStrategy::CSplittingStrategy(CLabels* labels, int32_t num_subsets)
 	reset_subsets();
 }
 
-void CSplittingStrategy::reset_subsets()
+void CSplittingStrategy::reset_subsets() const
```
karlnapf: cannot be const ...

src/shogun/evaluation/SplittingStrategy.cpp
```diff
@@ -69,13 +69,11 @@ CSplittingStrategy::~CSplittingStrategy()
 	SG_UNREF(m_subset_indices);
 }
 
-SGVector<index_t> CSplittingStrategy::generate_subset_indices(index_t subset_idx)
+const SGVector<index_t> CSplittingStrategy::validation(index_t subset_idx) const
```
karlnapf: I think it is a good idea to make this method const (if things are precomputed)
However, there is a "build_subsets" call below, which cannot be const, so the current const doenst make sense

src/shogun/evaluation/SplittingStrategy.h
```diff
@@ -58,36 +59,32 @@ class CSplittingStrategy: public CSGObject
 	 * @param subset_idx subset index of the to be generated vector indices
 	 * @return newly created vector of subset indices of the specified
 	 * subset is written here.
-	 *
-	 * Error if there are no index sets
 	 */
-	SGVector<index_t> generate_subset_indices(index_t subset_idx);
```
karlnapf: I like this renaming

src/shogun/evaluation/SplittingStrategy.h
```diff
@@ -58,36 +59,32 @@ class CSplittingStrategy: public CSGObject
...
-	 *
-	 * Error if there are no index sets
 	 */
-	SGVector<index_t> generate_subset_inverse(index_t subset_idx);
+	const SGVector<index_t> train(index_t subset_idx) const;
 
 	/** @return number of subsets. */
 	index_t get_num_subsets() const;
 
+protected:
```
karlnapf: so you make this method private. Why not call it in the constructor then? Then we can make the train/validation method const..

src/shogun/evaluation/SplittingStrategy.h
```diff
@@ -99,14 +96,14 @@ class CSplittingStrategy: public CSGObject
 	CLabels* m_labels;
 
 	/** subset indices */
-	CDynamicObjectArray* m_subset_indices;
+	mutable CDynamicObjectArray* m_subset_indices;
```
karlnapf: I see now what your thinking was.
I don't think this is the best style, making it mutable and the other things const.
There are simpler ways of reaching this goal imo

src/shogun/evaluation/SplittingStrategy.cpp
vigsterkr: the function arg can be const :)

src/shogun/evaluation/SplittingStrategy.h
luisffranca: I've changed that, thanks.

luisffranca:
Hi @karlnapf @vigsterkr !

Thanks for the review. I've changed the constness strategy and now build_subsets() is called in the constructors of the derived classes.

After these changes, `CrossValidationMMD` unit tests started to fail. That was due to the use of randomness in the creation of the subsets. Given that now build_subsets() is in the constructor, I've adapted the test in order to fix this.

src/shogun/evaluation/SplittingStrategy.h
```diff
@@ -58,33 +58,29 @@ class CSplittingStrategy: public CSGObject
 	 * @param subset_idx subset index of the to be generated vector indices
 	 * @return newly created vector of subset indices of the specified
 	 * subset is written here.
-	 *
-	 * Error if there are no index sets
 	 */
-	SGVector<index_t> generate_subset_indices(index_t subset_idx);
+	const SGVector<index_t> validation(const index_t subset_idx) const;
```
karlnapf: not sure why we want the return type to be const?

src/shogun/evaluation/TimeSeriesSplitting.cpp
```diff
@@ -45,6 +45,7 @@ CTimeSeriesSplitting::CTimeSeriesSplitting(CLabels* labels, index_t num_subsets)
     : CSplittingStrategy(labels, num_subsets)
 {
 	init();
+	build_subsets();
```
karlnapf: the method seem to be called in `init`, why call it again?

src/shogun/evaluation/TimeSeriesSplitting.cpp
```diff
@@ -99,6 +100,9 @@ void CTimeSeriesSplitting::set_min_subset_size(index_t min_size)
 	    "subsets and labels.",
 	    num_labels - (num_subsets - 1) * (num_labels / num_subsets) - 1);
 	m_min_subset_size = min_size;
+
+	// Rebuild subsets considering the new minimum subset size
```
karlnapf: so it is built twice?

src/shogun/evaluation/SplittingStrategy.cpp
```diff
@@ -69,7 +69,8 @@ CSplittingStrategy::~CSplittingStrategy()
 	SG_UNREF(m_subset_indices);
 }
 
-SGVector<index_t> CSplittingStrategy::generate_subset_indices(index_t subset_idx)
+const SGVector<index_t>
+CSplittingStrategy::validation(const index_t subset_idx) const
```
karlnapf: no need for const on index_t

karlnapf:
Added a few comments, the build seems to be fine. So tests passing.
Once we have cleaned up I think we can merge this.
Did you check **all** ipython notebooks and meta examples for the new API?

src/shogun/evaluation/TimeSeriesSplitting.cpp
luisffranca: In init() `m_min_subset_size` is set to 1, which is then used by build_subsets() (called by the constructor). If set_min_subset_size() is called, we should rebuild the subsets. I've found this due to a failure of `timeseries_subset_linear_splits` unit test.

src/shogun/evaluation/TimeSeriesSplitting.cpp
luisffranca: build_subsets() you mean? It isn't called on init() in this class.

src/shogun/evaluation/SplittingStrategy.cpp
luisffranca: Done.

src/shogun/evaluation/SplittingStrategy.h
luisffranca: I've changed this according to the Issue description. I think this shouldn't be a problem, given the possibility to return it to const or non-const variables. Should I remove it?

src/shogun/evaluation/TimeSeriesSplitting.cpp
karlnapf: Let's try to avoid this. Maybe you can shift things around that it is only built ones .. ?

src/shogun/evaluation/SplittingStrategy.h
karlnapf: I think it should be ok

src/shogun/evaluation/TimeSeriesSplitting.cpp
luisffranca: One way to achieve this would be to allow min_subset_size to be passed as an argument in the constructor (default value 1) and remove the method set_min_subset_size(). Do you think that would be a problem?

karlnapf:
Any updates here?

luisffranca:
Hi @karlnapf ! Sorry for my late reply.

I'm sending a new patch set changing build_subsets() in `TimeSeriesSplitting` according to your last comments. I've adapted the unit tests and notebook accordingly.

As far as I've checked, all python notebooks and meta examples for the new API were changed.

examples/undocumented/libshogun/splitting_LOO_crossvalidation.cpp
```diff
@@ -38,14 +38,12 @@ int main(int argc, char **argv)
...
-		splitting->build_subsets();
-
 		for (index_t i=0; i<num_labels; ++i)
 		{
 			//SG_SPRINT("subset %d\n", i);
 
-			SGVector<index_t> subset=splitting->generate_subset_indices(i);
-			SGVector<index_t> inverse=splitting->generate_subset_inverse(i);
+			SGVector<index_t> subset = splitting->validation(i);
+			SGVector<index_t> inverse = splitting->train(i);
```
karlnapf: just realised a minor inconsistency. Should be called "training" if the counterpart is called "validation"
But I would even prefer "train" and "test" ..

src/shogun/evaluation/CrossValidationSplitting.h
```diff
@@ -39,11 +39,12 @@ class CCrossValidationSplitting: public CSplittingStrategy
...
 	}
 
-	/** implementation of the standard cross-validation splitting strategy */
-	virtual void build_subsets();
-
 	/** custom rng if using cross validation across different threads */
 	CRandom * m_rng;
+
+protected:
+	/** implementation of the standard cross-validation splitting strategy */
```
karlnapf: this comment seems weird to me

src/shogun/evaluation/SplittingStrategy.h
karlnapf: see comment above

karlnapf:
This is almost ready.
Just my comment on the naming. What are your thoughts here?

src/shogun/evaluation/CrossValidationSplitting.h
luisffranca: I agree and will remove it.

luisffranca:
I like "train" and "test"! I'll make the changes and submit another patch set.

Thanks!

luisffranca:
Hi @karlnapf !

Any other comments in this PR? Can I squash the commits to clean it up?

karlnapf:
Thanks for the udpate and ping! :)

Looks good! No need to squash as github can do that these days

One thing I want to make sure.Did you run all unit and integration tests?

src/shogun/evaluation/CrossValidationSplitting.cpp
```diff
@@ -21,6 +21,7 @@ CCrossValidationSplitting::CCrossValidationSplitting(
 	CSplittingStrategy(labels, num_subsets)
 {
 	m_rng = sg_rand;
+	build_subsets();
```
vinx13: a problem here is that we can only call the default ctor from swig, then we may still need to call build_subset from the outside? @karlnapf

PR #4181, opened 2018-02-19
Meta examples: add functionality for special translations of certain method calls

sorig:
This allows typed getters like `get_real` to be translated to `get<float64_t>` in C++ and `get_real` in the SWIG interfaces.

Note that typed setters (`put`) would be more tricky as we'd have to type the keyword arguments

### DIFF
--- a/examples/meta/generator/targets/cpp.json
+++ b/examples/meta/generator/targets/cpp.json
@@ -90,7 +90,10 @@
         "IntLiteral": "$number",
         "RealLiteral": "$number",
         "FloatLiteral": "${number}f",
-        "MethodCall": "$object->$method($arguments)",
+        "MethodCall": {
+            "Default": "$object->$method($arguments)",
+            "get_real": "$object->get<float64_t>($arguments)"
+        },
         "StaticCall": "C$typeName::$method($arguments)",
         "GlobalCall": "$method($arguments)",
         "Identifier": "$identifier",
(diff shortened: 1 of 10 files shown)


examples/meta/generator/targets/cpp.json
```diff
@@ -90,7 +90,10 @@
         "IntLiteral": "$number",
         "RealLiteral": "$number",
         "FloatLiteral": "${number}f",
-        "MethodCall": "$object->$method($arguments)",
+        "MethodCall": {
+            "Default": "$object->$method($arguments)",
+            "get_real": "$object->get<float64_t>($arguments)"
+        },
         "StaticCall": "C$typeName::$method($arguments)",
         "GlobalCall": "$method($arguments)",
```
sorig: Do you think you would need something similar for `StaticCall` and `GlobalCall`? Can easily implement haven't done it for now

karlnapf: I think this should suffice for now

examples/meta/src/meta_api/getters.sg
```diff
@@ -0,0 +1,2 @@
+Kernel k = kernel("GaussianKernel")
```
karlnapf: could you remove this file? it makes the build fail as the method is not implemented.
I will use things in the feature branch

karlnapf:
Thanks for the quick response :)

karlnapf:
Travis is happy now

PR #4183, opened 2018-02-20
Add transpose flags to element_prod

vinx13:
Add transpose flags to arguments of `linalg::element_prod`
```
void element_prod(const SGMatrix<T>& A, const SGMatrix<T>& B, SGMatrix<T>& result,
                                bool transpose_A = false, bool transpose_B = false)
```
This makes `element_prod` have a consistent interface as `matrix_prod`

### DIFF
--- a/src/shogun/mathematics/linalg/LinalgBackendViennaCL.h
+++ b/src/shogun/mathematics/linalg/LinalgBackendViennaCL.h
@@ -107,14 +107,14 @@ namespace shogun
 	#undef BACKEND_GENERIC_DOT
 
 	/** Implementation of @see LinalgBackendBase::element_prod */
-	#define BACKEND_GENERIC_IN_PLACE_ELEMENT_PROD(Type, Container) \
+	#define BACKEND_GENERIC_IN_PLACE_MATRIX_ELEMENT_PROD(Type, Container) \
 	virtual void element_prod(const Container<Type>& a, const Container<Type>& b,\
-		Container<Type>& result) const \
+		Container<Type>& result, bool transpose_A, bool transpose_B) const \
 	{  \
-		element_prod_impl(a, b, result); \
+		element_prod_impl(a, b, result, transpose_A, transpose_B); \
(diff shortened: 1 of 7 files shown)


src/shogun/mathematics/linalg/backend/eigen/BasicOps.cpp
```diff
@@ -269,7 +286,14 @@ void LinalgBackendEigen::element_prod_impl(
 	Eigen::Block<typename SGMatrix<T>::EigenMatrixXtMap> b_block =
 	    b_eig.block(b.m_row_begin, b.m_col_begin, b.m_row_size, b.m_col_size);
 
-	result_eig = a_block.array() * b_block.array();
+	if (transpose_A && transpose_B)
+		result_eig = a_block.transpose().array() * b_block.transpose().array();
+	else if (transpose_A)
```
karlnapf: why the copy paste? why doesnt one method call the other?

src/shogun/mathematics/linalg/LinalgBackendViennaCL.h
```diff
@@ -364,16 +364,40 @@ namespace shogun
...
 			GPUMemoryViennaCL<T>* result_gpu = cast_to_viennacl(result);
 
-			result_gpu->data_matrix(a.num_rows, a.num_cols) =
-			    viennacl::linalg::element_prod(
-			        a_gpu->data_matrix(a.num_rows, a.num_cols),
-			        b_gpu->data_matrix(a.num_rows, a.num_cols));
+			if (transpose_A && transpose_B)
+				result_gpu->data_matrix(result.num_rows, result.num_cols) =
+				    viennacl::linalg::element_prod(
+				        viennacl::trans(
```
karlnapf: cant we store the "trans" in a varialbe and then use that in the if-then-else. Then we don't copy the whole code block 3 times ...

karlnapf:
Thanks!

PR #4188, opened 2018-02-27
Fix LMNN termination condition

vinx13:


karlnapf:
Don't we have some unit tests to cover this?
Something like. Asserts that residuals match requires tolerance on termination?

karlnapf:
Otherwise fine from my side.
Integration tests should soon be back so then we can compare using that

vinx13:
@karlnapf This is not covered by unit tests, i think we can check the number of iterations.

karlnapf:
Yeah would be good to add something in these lines to this PR

vinx13:
Added a tests. We need to fix trace first (#4176 ) to pass the test.

karlnapf:
Great!

tests/unit/metric/LMNN_unittest.cc
```diff
@@ -62,6 +62,57 @@ TEST(LMNN,train_identity_init)
 	SG_UNREF(lmnn)
 }
 
+TEST(LMNN, train_termination)
+{
+	// create features, each column is a feature vector
```
karlnapf: could you remove all these superfluous comments?

karlnapf: also, does this test share code with other tests? If so it should be pulled out and shared

tests/unit/metric/LMNN_unittest.cc
```diff
@@ -62,6 +62,57 @@ TEST(LMNN,train_identity_init)
...
+	// check linear transform solution
+	SGMatrix<float64_t> L=lmnn->get_linear_transform();
+	EXPECT_NEAR(L(0,0),0.000041647483219,1e-10);
+	EXPECT_NEAR(L(0,1),0,1e-10);
+	EXPECT_NEAR(L(1,0),0,1e-10);
+	EXPECT_NEAR(L(1,1),0.988162395685451,1e-10);
+
+	// check number of iterations
+	auto stat = lmnn->get_statistics();
+	EXPECT_EQ(stat->obj.vlen, 1234);
```
karlnapf: 1234?

tests/unit/metric/LMNN_unittest.cc
```diff
@@ -62,6 +62,57 @@ TEST(LMNN,train_identity_init)
...
+	SG_UNREF(lmnn)
```
karlnapf: use `some`, then you don't need to unref

tests/unit/metric/LMNN_unittest.cc
vinx13: Yes. I'm testing against Matlab prototype here https://github.com/iglesias/lmnn-gsoc2013-review

tests/unit/metric/LMNN_unittest.cc
vinx13: code that creates features and labels could be shared with other tests, but this could make it difficult to modify test input. shall we do this refactor?

tests/unit/metric/LMNN_unittest.cc
karlnapf: cool!

karlnapf:
Once the tests pass this can be merged

karlnapf:
:tada:

src/shogun/metric/LMNNImpl.cpp
```diff
@@ -292,6 +292,7 @@ bool CLMNNImpl::check_termination(float64_t stepsize, const SGVector<float64_t>
 
 	if (iter >= 10)
 	{
+		obj_threshold *= obj[iter - 1];
```
iglesias: Nice catch @vinx13, thanks a lot for fixing.

Issue #4186, opened 2018-02-23
drop CMath class


karlnapf:
Large entrance task. The goal is to stop using `CMath` in the shogun codebase and use the C++ standard library instead, e.g. `std::log` rather than `CMath::log` or `std::sort` rather than `CMath::qsort`.

Once a method is not used anymore, it should be dropped from the `CMath` class entirely.

PRs should be sent method by method pls.

Good entrance task if you want to work on modernizing shogun

Yashs744:
Hello, 
I tried doing this and updated DynArray.h but when I sent PR, it didn't pass the first test and I'm not getting what went wrong. Can you help?

shubham808:
@karlnapf I am ready with a PR to get rid of ```CMath::log``` but the ```gpl``` submodule needs to be updated as well..... should i make a PR against shogun-gpl ? (Its causing build to break on Travis)

karlnapf:
Yes, exactly

PR #4193, opened 2018-03-02
Replace CMath::exp with std::exp

syashakash:
Refers #4186 
- Replaced all `CMath::exp` in all files including `Math.h` and `Math.cpp`.

- Made sure `cmake ..` and `make` do not fail.


karlnapf:
Thx! :)

PR #4194, opened 2018-03-03
dropped CMath::max #4186

im029:
replaced `CMath::max` with `std::max` in __DynArray.h__.


karlnapf:
Hi! Thanks for the patch!
Could you pls send a PR that only contains changes in the lines that you acutally worked on? Otherwise I cannot review it. The stylechecker only checks lines that were changed by you so it will not fail if the rest of the file doesnt obbey the style

im029:
Please check in the earlier commit in this pull request it contains the lines on which I actually worked. The travis build failed on earlier commit due to style isues.

PR #4195, opened 2018-03-03
Drop CMath::max in DynArray.h

im029:
Dropped `CMath::max` and added `std::max` in _DynArray.h_

Created this new PR with just the lines in which actual changes were made.


im029:
I am getting a "The job exceeded the maximum time limit for jobs, and has been terminated." error. Any ideas what could be the cause?

karlnapf:
Pls don't close and re-open PRs. You can update your existing ones.
Build time is not your fault, if you ping us, we can restart the build

Issue #4177, opened 2018-02-15
ruby typemaps bugs for SGMatrix


karlnapf:
Problems:

Passing matrices to methods that expect vectors makes typemap match and leads to type error
```
require 'shogun'
svm = Shogun::LibLinear.new()
matrix = NMatrix.float(2,2)
svm.set_w(matrix)
```
results in `TypeError: can't convert Array into Float`

Passing matrices to methods that expect vectors makes typemap match
```
require 'shogun'
svm = Shogun::LibSVM.new()
matrix = NMatrix.float(2,2)
svm.set_kernel(matrix)
```
resulting in `TypeError: no implicit conversion of nil into String`

In both cases, the error arises since SWIG will call a conversion method after the typemap matched.

karlnapf:
blocks new `put`

vigsterkr:
i've just created a possible fix for this bug in swig, see swig/swig#1717

PR #4198, opened 2018-03-05
Feature/factory

karlnapf:


examples/meta/generator/targets/cpp.json
```diff
@@ -11,7 +11,7 @@
     "Comment": "//$comment\n",
     "Init": {
         "Construct": "auto $name = some<C$typeName>($arguments)$kwargs",
-        "Copy": "auto $name = $expr$kwargs",
+        "Copy": "auto $name = wrap($expr)$kwargs",
```
karlnapf: @lisitsyn re-introduced wrap since we will get memory errors otherwise (things are not ref'ed and then deleted by objects)

examples/meta/ruby/CMakeLists.txt
```diff
@@ -2,6 +2,7 @@ STRING(REGEX REPLACE "(.*)/narray.*$" "\\1" NARRAY_PATH ${NARRAY_LIB})
 
 LIST(APPEND EXCLUDED_RUBY_META_EXAMPLES
 			base_api-put_get
+      base_api-factory
```
karlnapf: @vigsterkr same problem as for put: the check for SGMatrix/object bails

karlnapf: (so I blacklisted it for now)

examples/meta/src/base_api/factory.sg
```diff
@@ -0,0 +1,13 @@
...
+Kernel kernel_gaussian = kernel("GaussianKernel")
+Kernel kernel_linear = kernel("LinearKernel")
+
+RealMatrix real_matrix(2,2)
+Features features_from_matrix = features(real_matrix)
+
+File features_file = csv_file("../../data/classifier_binary_2d_nonlinear_features_train.dat")
+Features features_from_file = features(features_file, enum EPrimitiveType.PT_FLOAT64)
+
+File labels_file = csv_file("../../data/classifier_binary_2d_nonlinear_labels_train.dat")
```
karlnapf: labels from file will be `CDenseLabels` which are then converted in the classes (needs some refactor in algorithms, see below)

src/shogun/base/class_list.h
```diff
@@ -37,19 +39,28 @@ namespace shogun {
...
+	    const char* name,
+	    EPrimitiveType pt = PT_NOT_GENERIC) throw(ShogunException)
 	{
-		auto* object = create(name, PT_NOT_GENERIC);
+		auto* object = create(name, pt);
 		if (!object)
 		{
-			SG_SERROR("No such class %s", name);
+			SG_SERROR(
+			    "Class %s with primitive type %s does not exist.\n", name,
```
karlnapf: IMO we need std::iostream for our logs, then we could just overload the << operator and dont have to introduce all these to_string methods ...

src/shogun/classifier/svm/LibSVM.cpp
```diff
@@ -48,7 +48,7 @@ bool CLibSVM::train_machine(CFeatures* data)
 	struct svm_node* x_space;
 
 	ASSERT(m_labels && m_labels->get_num_labels())
-	ASSERT(m_labels->get_label_type() == LT_BINARY)
+	auto binary_labels = m_labels->as_binary();
```
karlnapf: @lisitsyn @vigsterkr this is how labels could be converted in algorithms, if they need an interface to CBinaryLabels for example

src/shogun/features/SparseFeatures.cpp
```diff
@@ -45,7 +45,9 @@ template<class ST> CSparseFeatures<ST>::CSparseFeatures(const CSparseFeatures &
 	SG_REF(m_subset_stack);
 }
 
-template<class ST> CSparseFeatures<ST>::CSparseFeatures(CDenseFeatures<ST>* dense)
+template <class ST>
+CSparseFeatures<ST>::CSparseFeatures(CDenseFeatures<ST>* dense)
+	: CDotFeatures(0), feature_cache(NULL)
```
karlnapf: this was causing a segfault in python legacy examples every now and then

src/shogun/labels/BinaryLabels.cpp
```diff
@@ -6,17 +6,22 @@
...
 #include <shogun/lib/SGVector.h>
+#include <shogun/mathematics/Statistics.h>
 
 using namespace shogun;
 
 CBinaryLabels::CBinaryLabels() : CDenseLabels()
 {
 }
 
+CBinaryLabels::CBinaryLabels(const CBinaryLabels& orig) : CDenseLabels(orig)
```
karlnapf: I gave labels copy constructors to the sake of converting them, see `as_binary` below

src/shogun/labels/DenseLabels.cpp
```diff
@@ -77,15 +83,15 @@ void CDenseLabels::set_labels(SGVector<float64_t> v)
 	m_labels = v;
 }
 
-SGVector<float64_t> CDenseLabels::get_labels()
+SGVector<float64_t> CDenseLabels::get_labels() const
```
karlnapf: I made lots of getters const

examples/meta/ruby/CMakeLists.txt
vigsterkr: but that should be fixed in develop, so rebase over develop, otherwise i dont understand what's the problem

examples/meta/src/base_api/factory.sg
vigsterkr: if you have just added the factory.sg could you change the relative path (`../../`) to an env var? i'm happy to add the support for that for every language 💃

vigsterkr: as if we install every these meta examples, the location of data should be somewhere very else :)

src/shogun/base/class_list.h
vigsterkr: although it looks shitty but `operator <<` will not solve your problem to convert some objects to string.. that only works with primitive types

vigsterkr: on the other hand having these macros is nice in a long run as we could switch finally our logging backend. maybe spdlog? that allows us to redirect the logs into any `sink` not only stdout/err

src/shogun/labels/DenseLabels.h
```diff
@@ -256,7 +262,15 @@ class CDenseLabels : public CLabels
...
 		 */
-		virtual ELabelType get_label_type() const=0;
+		virtual ELabelType get_label_type() const override
+		{
+			return LT_DENSE_GENERIC;
+		}
+
+		virtual const char* get_name() const override
+		{
+			return "DenseLabels";
```
karlnapf: @lisitsyn made the class non-abstract, for the sake of loading from files (one doenst know what type of labels it is at that point, only when the algorithm tries to convert to say binary labels we know)

src/shogun/labels/BinaryLabels.cpp
vigsterkr: why do we need a copy? i mean it's over the same datastruct no? do we really want the `as_binary` be a copy? dont we wanna use c++11's feature of std::move and rvalue?

src/shogun/labels/Labels.cpp
```diff
@@ -90,7 +100,33 @@ void CLabels::set_values(SGVector<float64_t> values)
 	m_current_values = values;
 }
 
-SGVector<float64_t> CLabels::get_values()
+SGVector<float64_t> CLabels::get_values() const
 {
 	return m_current_values;
 }
+
+Some<CBinaryLabels> CLabels::as_binary() const
```
karlnapf: mentioned conversion method to convert label types into each other

vigsterkr: see above regarding move/rvalue

src/shogun/lib/SGVector.cpp
```diff
@@ -927,7 +927,7 @@ void SGVector<T>::scale(T alpha)
 
 template<class T> void SGVector<T>::load(CFile* loader)
 {
-	REQUIRE(loader, "Require a valid 'c FILE pointer'\n");
```
karlnapf: great error message for users

src/shogun/statistical_testing/QuadraticTimeMMD.cpp
```diff
@@ -66,7 +67,7 @@ struct CQuadraticTimeMMD::Self
 	SGVector<float64_t> gamma_fit_null();
 
 	CQuadraticTimeMMD& owner;
-	unique_ptr<CMultiKernelQuadraticTimeMMD> multi_kernel;
+	shared_ptr<CMultiKernelQuadraticTimeMMD> multi_kernel;
```
karlnapf: meta examples now use `some` so cannot have a unique ptr inside shogun and expose the raw pointer via SWIG

src/shogun/util/factory.h
```diff
@@ -0,0 +1,94 @@
...
+			result->load(file);
+		}
+		else
+			SG_SERROR("Cannot load features from %s.\n", file->get_name());
+
+		SG_REF(result);
+		return result;
+	}
+
+	CLabels* labels(CFile* file)
```
karlnapf: @vigsterkr @lisitsyn this is the stuff behind the factory API

src/shogun/statistical_testing/QuadraticTimeMMD.cpp
vigsterkr: but this way you loose:
 1) std::move the thing around - maybe not so important
 2) enforce sole ownership of the resource.... @lambday was this intentional to have it as unique? i would presume yes....

examples/meta/ruby/CMakeLists.txt
karlnapf: fixed in develop? I don't remember us fixing that SWIG bug? Did I miss it?
https://github.com/shogun-toolbox/shogun/issues/4177

examples/meta/src/base_api/factory.sg
karlnapf: Yep absoltely agree. But I think maybe let's do that for all meta examples in another PR?
I can do that next

src/shogun/labels/BinaryLabels.cpp
karlnapf: Yeah I thought about this. The reason why I made it a new instance was the fact that in the `as_binary` below, all other cases return a new instance, so to be consistent.
And also then the `as_binary` method can be const. On the other hand, `as_binary` could return const pointer ...
If you think it is better to just cast, I can do that. Not sure I get what you mean with the move and rvalue

src/shogun/base/class_list.h
karlnapf: I know but I can overload the << thing, no? Then I dont need the function calls when I write error messages, but can collect them somewhere else?
Totally up for doing a new logging backend. GSoC project part?

src/shogun/statistical_testing/QuadraticTimeMMD.cpp
karlnapf: discussed with him, he was fine with this change. Alternative would be to always return a new instance in the `::multiclass()` method (which currently returns the raw pointer)

src/shogun/labels/BinaryLabels.cpp
karlnapf: yep!

src/shogun/base/class_list.h
vigsterkr: new logging backend = just change the macro definitions.
yeah you can offload the operator << but that means that you force to use `std::iostream`-like interface to be used, which is not the most liked IO back-end (for good reasons) in c++ community

vigsterkr: on the other hand you can do the same thing, i.e. implement to_string() for that object where you wanna have operator <<, right?

src/shogun/labels/BinaryLabels.cpp
vigsterkr: so then use rvalue/move instead of this pattern.

karlnapf: how do you want to deal with converting say multiclass labels (that only contain two classes) to binary? For this, the actual vector needs to be changed, so a new instance is needed.

karlnapf: For converting the same datastruct, I see the point

vigsterkr:
If u need to have new mem of course u need a new instance but thats not always the case. Hence have both of them and the compiler will figure it out for u

vigsterkr:
Btw now thinking of this i’m not so sure we need this things explicit... as it makes the whiole thing less smooth.... afaik you could have the whole thing implicit with the right ctors

examples/meta/generator/targets/cpp.json
lisitsyn: This is a bit strange. Copy ctor doesn't ref?

src/shogun/base/class_list.h
lisitsyn: I think we can introduce an std::cout-like object which supports <<. Internally it could use any logger we want.

src/shogun/classifier/svm/LibSVM.cpp
lisitsyn: Looks good.

src/shogun/labels/BinaryLabels.cpp
lisitsyn: I think it is fine to copy BinaryLabels most of the time. What should be rather copy-on-write is the data beneath.

src/shogun/lib/DataType.h
```diff
@@ -57,6 +57,12 @@ enum EPrimitiveType
 	PT_COMPLEX128=14,
 	PT_UNDEFINED=15
 };
+
+/** Returns string representation of primitive type
+ * @param pt primitive type
+ * @return string for primitive type
+ */
+std::string ptype(EPrimitiveType pt);
```
lisitsyn: I like `primitive_name` a bit better.

src/shogun/util/factory.h
lisitsyn: I like the approach. It would become the main interface.

examples/meta/generator/targets/cpp.json
karlnapf: The point here is if you do
'labels = svm->precict'
then Shogun returns a raw pointer. Now if you pass this to another object that increases the ref-counter, and then decreases it, the object is destroyed. So calling a method on it subsequently crashes. This is why I re-added it, just like when you introduced it in the first place.
Or you mean something else?

src/shogun/labels/BinaryLabels.cpp
karlnapf: I would actually prefer if this thing gave you a read-only labels object (we wanna have immutable labels anyways)
Note that memory is not copied here.
Also, after reading on it, I think the move constructor stuff doesnt really help here

src/shogun/lib/DataType.h
karlnapf: will update

karlnapf:
We could also have a constructor in CBinaryLabels that takes a CLabels pointer and then does the same thing as this introduced `as_binary` method. Not sure that conflicts with a copy constructor in some way. I actually like the explicitness

karlnapf:
@lisitsyn see updated version

src/shogun/base/SGObject.h
```diff
@@ -492,36 +492,54 @@ class CSGObject
...
 		return nullptr;
 	}
 
+	/** Specializes the object to the specified type.
+	 * Throws exception if the object cannot be specialized.
+	 *
+	 * @return The requested type
+	 */
+	template <class T>
+	const T* as() const
```
vigsterkr: i wouldn't copy-paste this... some parts of it at least can be merged

src/shogun/base/some.h
```diff
@@ -153,6 +153,11 @@ namespace shogun
 		return value;
 	}
 
+	inline const char* wrap(const char* ptr)
```
vigsterkr: for inline-ness plz use the macro: SG_FORCED_INLINE
which will make it sure that it's really inlined

src/shogun/labels/BinaryLabels.cpp
```diff
@@ -148,3 +149,53 @@ CLabels* CBinaryLabels::shallow_subset_copy()
 
 	return shallow_copy_labels;
 }
+
+namespace shogun
+{
+	Some<CBinaryLabels> binary_from_binary(CBinaryLabels* orig)
```
vigsterkr: que?

karlnapf: these are helper methods, could also move the contents directly into the switch below

src/shogun/labels/BinaryLabels.cpp
```diff
@@ -148,3 +149,53 @@ CLabels* CBinaryLabels::shallow_subset_copy()
...
+		{
+			if (!CMath::fequals(labels[i], 1.0, 0.0) &&
+			    !CMath::fequals(labels[i], -1.0, 0.0))
+			{
+				SG_SERROR(
+				    "Label[%d] is %f, but must be either +1.0 or -1.0.\n", i,
+				    labels[i])
+			}
+		}
+		auto result = new CBinaryLabels(labels);
```
vigsterkr: afaik this copies the whole vvector itself.... do we really need to copy the data where it doesn't change at all....

src/shogun/labels/BinaryLabels.cpp
```diff
@@ -148,3 +149,53 @@ CLabels* CBinaryLabels::shallow_subset_copy()
...
 }
+
+namespace shogun
+{
+	Some<CBinaryLabels> binary_from_binary(CBinaryLabels* orig)
+	{
+		return Some<CBinaryLabels>::from_raw(orig);
+	}
+
+	Some<CBinaryLabels> binary_from_dense(CDenseLabels* orig)
```
vigsterkr: que?
do we need to spell this one out? does this need to be expose to anybody apart from ourself internally?

vigsterkr: ah this is in .cpp. nevermind

src/shogun/labels/BinaryLabels.cpp
vigsterkr: cpp only, nevermind

src/shogun/labels/MulticlassLabels.cpp
```diff
@@ -184,4 +185,53 @@ CMulticlassLabels* CMulticlassLabels::obtain_from_generic(CLabels* labels)
...
+	}
+
+	Some<CMulticlassLabels> multiclass_from_dense(CDenseLabels* orig)
+	{
+		auto int_labels = orig->get_int_labels();
+		std::set<int32_t> unique(int_labels.begin(), int_labels.end());
+		REQUIRE(
+				(*std::min_element(unique.begin(), unique.end())) == 0 &&
+				(*std::max_element(unique.begin(), unique.end())) == (index_t)unique.size()-1,
+				"Multiclass labels must be contiguous integers in [0, ..., num_classes -1].\n"
```
vigsterkr: can we fixme/todo this... as this should not be a hard requirement, right?

tests/meta/tester.cpp
```diff
@@ -62,6 +83,6 @@ int main(int argc, const char *argv[])
 	SG_UNREF(a_ref);
 
 	exit_shogun();
-    return equal != true;
+	return (loaded_equals_ref && ref_equals_loaded) != true;
```
vigsterkr: we could drop the `!= true` part and just have a negation, right?

tests/unit/utils/factory_unittest.cc
```diff
@@ -0,0 +1,79 @@
...
+	file_save->close();
+
+	auto file_load = some<CCSVFile>(filename.c_str(), 'r');
+	auto* obj = features(file_load, PT_FLOAT64);
+	EXPECT_TRUE(obj != nullptr);
+	auto* cast = dynamic_cast<CDenseFeatures<float64_t>*>(obj);
+	ASSERT_TRUE(cast != nullptr);
+	auto loaded_mat = cast->get_feature_matrix();
+	EXPECT_TRUE(loaded_mat.equals(mat));
+	delete obj;
```
vigsterkr: delete the file as well plz.

tests/unit/utils/factory_unittest.cc
```diff
@@ -0,0 +1,79 @@
...
+	file_save->close();
+
+	auto file_load = some<CCSVFile>(filename.c_str(), 'r');
+	auto* obj = labels(file_load);
+	EXPECT_TRUE(obj != nullptr);
+	auto* cast = dynamic_cast<CDenseLabels*>(obj);
+	ASSERT_TRUE(cast != nullptr);
+	auto loaded_vec = cast->get_labels();
+	EXPECT_TRUE(loaded_vec.equals(vec));
+	delete obj;
```
vigsterkr: plz remove the temporary file on the end

src/shogun/labels/BinaryLabels.cpp
karlnapf: it copies indeed, which I actually didnt realise. Why does it copy? Doesnt seem to make sense to me. And agree, it should not

src/shogun/labels/MulticlassLabels.cpp
karlnapf: yep!

src/shogun/labels/BinaryLabels.cpp
karlnapf: no this is only internally, didnt put it in the header.
You think it should be just inside the `switch`?

vigsterkr: see the 2nd comment... it's foobar, just ignore the first comment

karlnapf: browser didnt reload ;)

examples/meta/generator/targets/cpp.json
lisitsyn: Yes I've got it now. I guess we change it once we get rid of raw pointers.

karlnapf:
I would like to merge this.
Both travis and windows timeout unfortunately.
Locally, the tests pass.

vigsterkr:
lets try to make the CIs pass as this anyways is going to be a huge problem with GSoC. i've tried dropping some heavy compilation tasks (serialization) but seems it is still not enough

src/shogun/base/some.h
vigsterkr: ?

src/shogun/evaluation/MeanSquaredError.cpp
```diff
@@ -16,12 +16,15 @@ float64_t CMeanSquaredError::evaluate(CLabels* predicted, CLabels* ground_truth)
...
-	REQUIRE(predicted->get_label_type() == LT_REGRESSION, "Predicted label type (%d) must be regression (%d).\n", predicted->get_label_type(), LT_REGRESSION)
-	REQUIRE(ground_truth->get_label_type() == LT_REGRESSION, "Ground truth label type (%d) must be regression (%d).\n", ground_truth->get_label_type(), LT_REGRESSION)
 	int32_t length = predicted->get_num_labels();
 	float64_t mse = 0.0;
+	auto predicted_regression = regression_labels(predicted);
+	auto ground_truth_regression = regression_labels(ground_truth);
+
 	for (int32_t i=0; i<length; i++)
-		mse += CMath::sq(((CRegressionLabels*) predicted)->get_label(i) - ((CRegressionLabels*) ground_truth)->get_label(i));
+		mse += CMath::sq(
```
vigsterkr: `std::`?

karlnapf: this is autoformater, I didnt change that line

karlnapf: ah no actually I did
sure can change that

src/shogun/labels/BinaryLabels.cpp
```diff
@@ -148,3 +149,55 @@ CLabels* CBinaryLabels::shallow_subset_copy()
...
+		{
+			if (!CMath::fequals(labels[i], 1.0, 0.0) &&
+			    !CMath::fequals(labels[i], -1.0, 0.0))
+			{
+				SG_SERROR(
+				    "Label[%d] is %f, but must be either +1.0 or -1.0.\n", i,
+				    labels[i])
+			}
+		}
+		auto result = new CBinaryLabels();
```
vigsterkr: wouldn't be better to actually drop those copy ctors that actually copy and fix it instead of this hack?

karlnapf: Maybe, but I was afraid that this will lead to all sorts of other required changes in places that depend on the fact that labels copy the data.
I can look into it, though I think the patch is already big enough

vigsterkr: imo this is just nurturing the bad design so would be just better to drop that faulty copy-ctor.

src/shogun/evaluation/MeanSquaredError.cpp
vigsterkr: doesn't matter based on your changes already going to create a problem from the other PR....

src/shogun/base/some.h
vigsterkr: just replace it with the given macro above.

vigsterkr: in case you insist on using inline... otherwise just drop the keyword `inline`

karlnapf:
Shall I contact travis to get more time on the builds?

karlnapf:
Alternatively, we could disable the libshogun examples by default

vigsterkr:
imo that's not going to work as that's a global counter and they wont change that unless you actually start paying for their services.... and since we anyways still would have problems with appveyor i'd rather try to lower our compilation time

karlnapf:
removing old examples then?

vigsterkr:
that's any easy fix just remove from the `all` target... that's a good question whether that actually would solve the compilation time problem. but it's worth a shot

vigsterkr:
but that means that you should remove the libshogun examples from ctest as well... as they won't be built by default

karlnapf:
Looking at the logs, this will not help unfortunately :( it is killen when compiling shogun itself

src/shogun/base/some.h
karlnapf: I don't insist, I am just following the rest of the file, also it doesnt have any shogun includes

karlnapf: but let me remove the keyword

src/shogun/labels/BinaryLabels.cpp
karlnapf: I agree, and will do it

src/shogun/base/some.h
vigsterkr: dont get it why is it better to remove instead of using a macro that ensures the inline-ess... hence make the whole code faster... ¯\_(ツ)_/¯

karlnapf:
Alright, let's see what the buildbot says to this...

Issue #4201, opened 2018-03-08
cannot kill a running task


vigsterkr:
try any long running python example and try to ctrl-c.... only `kill -9 <PID>` works.

geektoni:
I think this may be caused by the premature stopping feature. I believe this line https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/base/init.cpp#L101 is causing the problem you are running into. I'll have a closer look later today.

vigsterkr:
@geektoni so i'm guessing the reason it is currently not doing anything is because of this:
https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/lib/Signal.cpp#L17

since in the handler:
https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/lib/Signal.cpp#L36

do you know why it is by default turned off? :)

geektoni:
I think we wanted to let the user decide what behaviour Ctrl+C to use. In fact, here https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/lib/Signal.h#L76 there is the method to enable the custom signal handler.

Anyway, this line https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/lib/Signal.cpp#L37 is clearly wrong, it should have an exit call instead of return. I'll provide a patch to fix that. Shall I also change the behaviour of the signal handler to be always active?

vigsterkr:
@geektoni i would rather have it by default active as currently if you run a model and wanna cancel it the only way to do is to kill the process :) which is not the most user friendly behaviour :P

PR #4205, opened 2018-03-13
Fix issue #4201.

geektoni:
#4201

### DIFF
--- a/src/shogun/lib/Signal.cpp
+++ b/src/shogun/lib/Signal.cpp
@@ -14,7 +14,7 @@
 using namespace shogun;
 using namespace rxcpp;
 
-bool CSignal::m_active = false;
+bool CSignal::m_active = true;
 CSignal::SGSubjectS* CSignal::m_subject = new rxcpp::subjects::subject<int>();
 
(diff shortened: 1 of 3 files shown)


src/shogun/lib/Signal.cpp
```diff
@@ -32,9 +32,9 @@ CSignal::~CSignal()
 
 void CSignal::handler(int signal)
 {
-	/* If the handler is not enabled, then return */
+	/* If the handler is not enabled exit */
 	if (!m_active)
-		return;
+		exit(-1);
```
vigsterkr: are we sure we want exit(-1) ? i mean in this case of a REPL we would exit the REPL itself, or?

geektoni: The user won't enter into REPL loop if m_active is set to false. In such a situation, the handler won't ask the user any inputs, it will just terminate the execution and it won't print any message to the screen either. So I decided to maintain a consisten behaviour between the Shogun's handler and the standard system handler when dealing with a SIGINT: exit and set the exit code != 0.

Issue #4207, opened 2018-03-16
Use fixtures / templates in LibLinear unit test


karlnapf:
The unit tests for `CLibLinear` are highly redundant copy paste style code -- hard to maintain/read/change.
https://github.com/shogun-toolbox/shogun/blob/develop/tests/unit/classifier/svm/LibLinear_unittest.cc

This easy entrance task is to use googletest's fixture concept to reduce the number of lines of code in this test. It should be possible to reduce things down to 1-2 explicitly coded out tests.

Useful ressources
* Grep for `TEST_F` in the shogun tests 
* https://github.com/google/googletest/blob/master/googletest/docs/Primer.md (search `TEST_F`)
* https://github.com/google/googletest/blob/master/googletest/docs/AdvancedGuide.md (search `TYPED_TEST`)

vigsterkr:
closing as this has been fixed

PR #4208, opened 2018-03-17
Refactored the CLibLinear test codes and reduced them from 1300 lines to 470.

FaroukY:
Regarding issue #4207.

### DIFF
--- a/tests/unit/classifier/svm/LibLinear_unittest.cc
+++ b/tests/unit/classifier/svm/LibLinear_unittest.cc
@@ -15,7 +15,6 @@
 #include <shogun/evaluation/ContingencyTableEvaluation.h>
 #include <shogun/mathematics/Math.h>
 #include <gtest/gtest.h>
-#include <map>
 #include <string>
 #include <vector>
 
@@ -51,11 +50,12 @@ public:
 		else
 				generate_data_l2();
 
-		CLibLinear* ll = new CLibLinear();
+		auto ll = some<CLibLinear>();
 
-		CBinaryLabels* pred = NULL;
 		float64_t liblin_accuracy;
-		CContingencyTableEvaluation* eval = new CContingencyTableEvaluation();
+		auto eval = some<CContingencyTableEvaluation>();
+
+
 
 		ll->set_bias_enabled(biasEnable);
 		ll->set_features(train_feats);
@@ -63,7 +63,7 @@ public:
 
 		ll->set_liblinear_solver_type(liblinear_solver_type);
 		ll->train();
-		pred = ll->apply_binary(test_feats);
+		auto pred = ll->apply_binary(test_feats);
 		SG_REF(pred);
 
 		liblin_accuracy = eval->evaluate(pred, ground_truth);
@@ -84,11 +84,13 @@ public:
 		else
 				generate_data_l2_simple();
 
-		CLibLinear* ll = new CLibLinear();
+		auto ll = some<CLibLinear>();
 
-		CBinaryLabels* pred = NULL;
+
 		float64_t liblin_accuracy;
-		CContingencyTableEvaluation* eval = new CContingencyTableEvaluation();
+		auto eval = some<CContingencyTableEvaluation>();
+
+
 
 		ll->set_bias_enabled(biasEnable);
 		ll->set_features(train_feats);
@@ -98,7 +100,7 @@ public:
 		ll->set_liblinear_solver_type(liblinear_solver_type);
 		ll->train();
 
-		pred = ll->apply_binary(test_feats);
+		auto pred = ll->apply_binary(test_feats);
 		SG_REF(pred);
 		liblin_accuracy = eval->evaluate(pred, ground_truth);
 
@@ -144,6 +146,7 @@ protected:
 		*/
 
 		index_t num_samples = 50;
+		sg_rand->set_seed(5);
 		CMath::init_random(5);
 		SGMatrix<float64_t> data =
 			CDataGenerator::generate_gaussians(num_samples, 2, 2);
--- a/tests/unit/classifier/svm/LibLinear_unittest.cc
+++ b/tests/unit/classifier/svm/LibLinear_unittest.cc
@@ -151,7 +151,6 @@ protected:
 	void generate_data_l1()
 	{
 		generate_data_l2();
-		//Transpose the train_feats matrix
 		auto old_train_feats = train_feats;
 		train_feats = train_feats->get_transposed();
 		SG_UNREF(old_train_feats);
@@ -190,7 +189,6 @@ protected:
 	void generate_data_l1_simple()
 	{
 		generate_data_l2_simple();
-		//transpose train_feats matrix
 		auto old_train_feats = train_feats;
 		train_feats = train_feats->get_transposed();
 		SG_UNREF(old_train_feats);
--- a/tests/unit/classifier/svm/LibLinear_unittest.cc
+++ b/tests/unit/classifier/svm/LibLinear_unittest.cc
@@ -151,7 +151,9 @@ protected:
 	void generate_data_l1()
 	{
 		generate_data_l2();
+		auto old_train_feats = train_feats;
 		train_feats = train_feats->get_transposed();
+		SG_UNREF(old_train_feats);
 		SG_REF(train_feats);
 	}
 	void generate_data_l2_simple()
@@ -187,7 +189,9 @@ protected:
 	void generate_data_l1_simple()
 	{
 		generate_data_l2_simple();
+		auto old_train_feats = train_feats;
 		train_feats = train_feats->get_transposed();
+		SG_UNREF(old_train_feats);
 		SG_REF(train_feats);
 	}
 };
--- a/tests/unit/classifier/svm/LibLinear_unittest.cc
+++ b/tests/unit/classifier/svm/LibLinear_unittest.cc
@@ -1,11 +1,11 @@
 /*
- * This program is free software; you can redistribute it and/or modify
- * it under the terms of the GNU General Public License as published by
- * the Free Software Foundation; either version 3 of the License, or
- * (at your option) any later version.
+ * This software is distributed under BSD 3-clause license (see LICENSE file).
  *
- * Written (W) 2014 pl8787
+ * Authors:
+ *	Written (W) 2014 pl8787
+ *	Refactored (R) 2018 Elfarouk Yasser
  */
+
 #include <shogun/classifier/svm/LibLinear.h>
 #include <shogun/features/DataGenerator.h>
 #include <shogun/features/DenseFeatures.h>
@@ -86,7 +86,7 @@ public:
 		ll->set_bias_enabled(biasEnable);
 		ll->set_features(train_feats);
 		if (C_value)
-				ll->set_C(0.1,0.1); //Only in the case of L2R_L1LOSS_SVC_DUAL
+			ll->set_C(0.1,0.1); //Only in the case of L2R_L1LOSS_SVC_DUAL
 		ll->set_labels(ground_truth);
 		ll->set_liblinear_solver_type(liblinear_solver_type);
 		ll->train();
@@ -98,12 +98,12 @@ public:
 
 		if (!biasEnable)
 		{
-			for(int i=0;i<t_w.vlen;i++)
+			for (auto i : range(t_w.vlen))
 				EXPECT_NEAR(ll->get_w()[i], t_w[i], 1e-5);
 		}
 		else
 		{
-			for(int i=0;i<t_w.vlen;i++)
+			for (auto i : range(t_w.vlen))
 				EXPECT_NEAR(ll->get_w()[i], t_w[i], 4e-5);
 			EXPECT_NEAR(ll->get_bias(), t_w[2], 4e-5);
 		}
@@ -115,19 +115,7 @@ public:
 	}
 
 protected:
-	void transpose_feats_matrix()
-	{
-		//Transpose train_feats matrix
-		SGMatrix<float64_t> train_matrix = train_feats->get_feature_matrix();
-		SGMatrix<float64_t>::transpose_matrix(train_matrix.matrix,
-				train_matrix.num_rows, train_matrix.num_cols);
-
-		SG_UNREF(train_feats);
-		train_feats = new CDenseFeatures<float64_t>(train_matrix);
-		SG_REF(train_feats);
-
-	}
-	void initialize_train_and_test_and_ground()
+	void generate_data_l2()
 	{
 		index_t num_samples = 50;
 		sg_rand->set_seed(5);
@@ -164,20 +152,17 @@ protected:
 	}
 	void generate_data_l1()
 	{
-		initialize_train_and_test_and_ground();
-
-
-		transpose_feats_matrix();
-
+		generate_data_l2();
+		//Transpose the train_feats matrix
+		auto old_train_feats = train_feats;
+		train_feats = train_feats->get_transposed();
+		SG_UNREF(old_train_feats);
+		SG_REF(train_feats);
 	}
-	void generate_data_l2()
-	{
-		initialize_train_and_test_and_ground();
-	}
-	void initialize_train_and_test_and_ground_simple()
+	void generate_data_l2_simple()
 	{
 		SGMatrix<float64_t> train_data(2, 10);
-		SGMatrix<float64_t> test_data(2, 10); //Always 2x10, doesn't matter l1 or l2
+		SGMatrix<float64_t> test_data(2, 10);
 
 		index_t num_samples = 10;
 		SGVector<int32_t> x{0,0,1,1,2,2,3,3,4,4,5,5,6,6,7,7,8,8,9,9};
@@ -206,13 +191,12 @@ protected:
 	}
 	void generate_data_l1_simple()
 	{
-			initialize_train_and_test_and_ground_simple();
-
-			transpose_feats_matrix();
-	}
-	void generate_data_l2_simple()
-	{
-			initialize_train_and_test_and_ground_simple();
+		generate_data_l2_simple();
+		//transpose train_feats matrix
+		auto old_train_feats = train_feats;
+		train_feats = train_feats->get_transposed();
+		SG_UNREF(old_train_feats);
+		SG_REF(train_feats);
 	}
 };
 
@@ -315,12 +299,6 @@ TEST_F(LibLinear, train_L2R_LR_DUAL_BIAS)
 	// bias, and not l1
 	train_with_solver(liblinear_solver_type, true, false);
 }
-
-/*
-* --------------------------------
-*	Simple set tests start from here
-* --------------------------------
-*/
 
 TEST_F(LibLinear, simple_set_train_L2R_LR)
 {
--- a/tests/unit/classifier/svm/LibLinear_unittest.cc
+++ b/tests/unit/classifier/svm/LibLinear_unittest.cc
@@ -6,21 +6,15 @@
  *
  * Written (W) 2014 pl8787
  */
-#include <gtest/gtest.h>
-#include "utils/Utils.h"
-
 #include <shogun/classifier/svm/LibLinear.h>
 #include <shogun/features/DataGenerator.h>
 #include <shogun/features/DenseFeatures.h>
 #include <shogun/evaluation/ContingencyTableEvaluation.h>
 #include <shogun/mathematics/Math.h>
 #include <gtest/gtest.h>
-#include <string>
-#include <vector>
 
 using namespace shogun;
 
-#ifdef HAVE_LAPACK
 
 class LibLinear : public ::testing::Test
 {
@@ -39,7 +33,7 @@ public:
 		SG_UNREF(test_feats);
 		SG_UNREF(ground_truth);
 	}
-	//Helper that tests can call to drastically reduce code
+
 	void train_with_solver
 	(LIBLINEAR_SOLVER_TYPE llst, bool biasEnable, bool l1)
 	{
@@ -53,11 +47,8 @@ public:
 		auto ll = new CLibLinear();
 		SG_REF(ll);
 
-		float64_t liblin_accuracy;
 		auto eval = new CContingencyTableEvaluation();
 		SG_REF(eval);
-
-
 
 		ll->set_bias_enabled(biasEnable);
 		ll->set_features(train_feats);
@@ -68,12 +59,12 @@ public:
 		auto pred = ll->apply_binary(test_feats);
 		SG_REF(pred);
 
-		liblin_accuracy = eval->evaluate(pred, ground_truth);
+		auto liblin_accuracy = eval->evaluate(pred, ground_truth);
 
 		EXPECT_NEAR(liblin_accuracy, 1.0, 1e-6);
 
-		SG_UNREF(ll);
 		SG_UNREF(eval);
+		SG_UNREF(ll);
 		SG_UNREF(pred);
 	}
 	void train_with_solver_simple(
@@ -89,12 +80,8 @@ public:
 		auto ll = new CLibLinear();
 		SG_REF(ll);
 
-
-		float64_t liblin_accuracy;
-		auto eval = new CContingencyTableEvaluation();
+		auto eval = new  CContingencyTableEvaluation();
 		SG_REF(eval);
-
-
 
 		ll->set_bias_enabled(biasEnable);
 		ll->set_features(train_feats);
@@ -106,7 +93,8 @@ public:
 
 		auto pred = ll->apply_binary(test_feats);
 		SG_REF(pred);
-		liblin_accuracy = eval->evaluate(pred, ground_truth);
+
+		auto liblin_accuracy = eval->evaluate(pred, ground_truth);
 
 		if (!biasEnable)
 		{
@@ -127,33 +115,27 @@ public:
 	}
 
 protected:
-	void generate_data_l1()
+	void transpose_feats_matrix()
 	{
-		generate_data("L1");
-	}
-	void generate_data_l2()
-	{
-		generate_data("L2");
-	}
-	void generate_data_l1_simple()
-	{
-		generate_data_simple("L1_SIMPLE");
-	}
-	void generate_data_l2_simple()
-	{
-			generate_data_simple("L2_SIMPLE");
-	}
-	void generate_data(std::string type) //Type either "L1" or "L2"
-	{
-		/*
-			First part is shared by l1 and l2
-		*/
+		//Transpose train_feats matrix
+		SGMatrix<float64_t> train_matrix = train_feats->get_feature_matrix();
+		SGMatrix<float64_t>::transpose_matrix(train_matrix.matrix,
+				train_matrix.num_rows, train_matrix.num_cols);
 
+		SG_UNREF(train_feats);
+		train_feats = new CDenseFeatures<float64_t>(train_matrix);
+		SG_REF(train_feats);
+
+	}
+	void initialize_train_and_test_and_ground()
+	{
 		index_t num_samples = 50;
 		sg_rand->set_seed(5);
 
+
 		SGMatrix<float64_t> data =
 			CDataGenerator::generate_gaussians(num_samples, 2, 2);
+
 		CDenseFeatures<float64_t> features(data);
 
 		/*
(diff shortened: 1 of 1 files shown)


FaroukY:
The Pull request seems to have duplicated the data from the previous Pull request. The only files that should be changed here is the last one (LibLinear_unittest.cc). All the other commits are from other pull requests.

vigsterkr:
@FaroukY yeah i think you forked your branch `feature_fix_lib_linear_doc` not from `develop` but from the other fix' branch. just checkout develop pull the latest changes and then simply cherry pick 505f6abd9c17a66a0cbebfdffc8a0b93f1e13003 into that and forcepush it to your remote `feature_fix_lib_linear_doc`, this way you dont even need to create a new PR ;)

tests/unit/classifier/svm/LibLinear_unittest.cc
```diff
@@ -6,1292 +6,469 @@
...
-	test_feats =  (CDenseFeatures<float64_t>*)features.copy_subset(test_idx);
+		if (l1)
+				generate_data_l1();
+		else
+				generate_data_l2();
 
-	SGMatrix<float64_t> train_matrix = train_feats->get_feature_matrix();
-	SGMatrix<float64_t>::transpose_matrix(train_matrix.matrix,
-			train_matrix.num_rows, train_matrix.num_cols);
+		CLibLinear* ll = new CLibLinear();
```
vigsterkr: you can use some in many of these cases, so you dont need to call SG_UNREF :)
for example:
```auto ll = some<CLibLinear>();```

File: tests/unit/classifier/svm/LibLinear_unittest.cc [hunk 2]
```diff
@@ -6,1292 +6,469 @@
...
 
-	SG_UNREF(train_feats);
+		CBinaryLabels* pred = NULL;
+		float64_t liblin_accuracy;
+		CContingencyTableEvaluation* eval = new CContingencyTableEvaluation();
```
vigsterkr: same here as above:
```auto eval = some<CContingencyTableEvaluation>();```

File: tests/unit/classifier/svm/LibLinear_unittest.cc [hunk 3]
```diff
@@ -6,1292 +6,469 @@
...
 
-	train_feats = new CDenseFeatures<float64_t>(train_matrix);
+		ll->set_bias_enabled(biasEnable);
+		ll->set_features(train_feats);
+		ll->set_labels(ground_truth);
 
-	ground_truth = new CBinaryLabels(labels);
+		ll->set_liblinear_solver_type(liblinear_solver_type);
+		ll->train();
+		pred = ll->apply_binary(test_feats);
```
vigsterkr: this could be simply
```
auto pred = ll->apply_binary(test_feats);
```

vigsterkr: and then you can remove the declaration of `CBinaryLabels* pred = NULL;`

tests/unit/classifier/svm/LibLinear_unittest.cc
```diff
@@ -6,1292 +6,469 @@
...
-
-	train_feats = new CDenseFeatures<float64_t>(train_data);
-	test_feats = new CDenseFeatures<float64_t>(test_data);
-
-	SGVector<float64_t> labels(num_samples);
-	for (index_t i = 0; i < num_samples; ++i)
-	{
-		labels[i] = (i < num_samples/2) ? 1.0 : -1.0;
-	}
+		CLibLinear* ll = new CLibLinear();
```
vigsterkr: same as above... this code shares quite a lot of lines with `train_with_solver`... we could combine them :)

tests/unit/classifier/svm/LibLinear_unittest.cc
```diff
@@ -6,1292 +6,469 @@
...
+			generate_data_simple("L2_SIMPLE");
+	}
+	void generate_data(std::string type) //Type either "L1" or "L2"
+	{
+		/*
+			First part is shared by l1 and l2
+		*/
+
+		index_t num_samples = 50;
+		CMath::init_random(5);
```
vigsterkr: use `sg_rand->set_seed` plz

tests/unit/classifier/svm/LibLinear_unittest.cc
```diff
@@ -6,1292 +6,469 @@
...
 #include <shogun/evaluation/ContingencyTableEvaluation.h>
 #include <shogun/mathematics/Math.h>
 #include <gtest/gtest.h>
+#include <map>
+#include <string>
+#include <vector>
 
 using namespace shogun;
 
 #ifdef HAVE_LAPACK
```
vigsterkr: i think this can go as this was required for the gaussian blob generator, but that's not LAPACK only anymore...

karlnapf:
This needs a rebase as I fixed some missing SG_REF recently.
Should be pretty minimal.

FaroukY:
@karlnapf Going through the diff, the only difference I see in this file is the SG_REF(pred). Is that safe to assume, or am I missing another SG_REF?

tests/unit/classifier/svm/LibLinear_unittest.cc
FaroukY: I used the string and vector in the tests, so I will only delete <map>, is that okay?

tests/unit/classifier/svm/LibLinear_unittest.cc
FaroukY: Done

tests/unit/classifier/svm/LibLinear_unittest.cc
FaroukY: I did think about this, but when I did it, it had a lot of if statements and it wasn't very readable (for a test case). Are we sure we wanna do this?

tests/unit/classifier/svm/LibLinear_unittest.cc
FaroukY: Done

tests/unit/classifier/svm/LibLinear_unittest.cc
FaroukY: Done

tests/unit/classifier/svm/LibLinear_unittest.cc
FaroukY: Done

tests/unit/classifier/svm/LibLinear_unittest.cc
FaroukY: Done

vigsterkr:
@FaroukY ok in order to be able to merge this you need to remove the unrelated commits in this PR

FaroukY:
@vigsterkr Removed all the unrelated commits. This should be okay to merge now.

tests/unit/classifier/svm/LibLinear_unittest.cc
karlnapf: lapack can still go imo, no?

tests/unit/classifier/svm/LibLinear_unittest.cc
```diff
@@ -6,1320 +6,474 @@
...
+	{
+		sg_rand->set_seed(1);
+	}
+	virtual void TearDown()
+	{
+		SG_UNREF(train_feats);
+		SG_UNREF(test_feats);
+		SG_UNREF(ground_truth);
+	}
+	//Helper that tests can call to drastically reduce code
```
karlnapf: I would remove the comment and let the code speak for itself :) minor

karlnapf:
The liblinear test segfaults under clang.
Could you pls run it with valgrind and make sure it neither leaks nor does uninitialised memory reads?
https://travis-ci.org/shogun-toolbox/shogun/jobs/359675411#L4561

Pls check those things yourself next time, it is all there....

karlnapf:
Windows build passed, but that is since it doesnt have LAPACK installed and you have the test still guarded so it is not executed

FaroukY:
@karlnapf @vigsterkr  Got rid of the memory leak. It seems like Travis timed out on two jobs, but the remaining ones passed. Any ideas?

karlnapf:
clang passed so the tests executed in there were fine. good.
the timeout is not your fault.

I think this is OK to be merged. Thanks for the simplification.

tests/unit/classifier/svm/LibLinear_unittest.cc
```diff
@@ -6,1320 +6,480 @@
...
+	{
+		generate_data("L1");
+	}
+	void generate_data_l2()
+	{
+		generate_data("L2");
+	}
+	void generate_data_l1_simple()
+	{
+		generate_data_simple("L1_SIMPLE");
```
karlnapf: let's not do this string argument to determine the behaviour of the function. Just split it a bit further into helpers.

tests/unit/classifier/svm/LibLinear_unittest.cc
```diff
@@ -6,1320 +6,480 @@
...
+	}
+	void generate_data_l2_simple()
+	{
+			generate_data_simple("L2_SIMPLE");
+	}
+	void generate_data(std::string type) //Type either "L1" or "L2"
```
karlnapf: as said this big function should be split into multiple helpers and pls not string arguments to determine behaviour

tests/unit/classifier/svm/LibLinear_unittest.cc
```diff
@@ -6,1320 +6,480 @@
...
 
-	SG_UNREF(ll);
-	SG_UNREF(train_feats);
-	SG_UNREF(test_feats);
-	SG_UNREF(ground_truth);
-	SG_UNREF(eval);
-	SG_UNREF(pred);
-}
+		SGMatrix<float64_t> data =
+			CDataGenerator::generate_gaussians(num_samples, 2, 2);
```
karlnapf: could you pls check whether this class is guarded with lapack, and if not, remove the lapack guard around the test?

tests/unit/classifier/svm/LibLinear_unittest.cc
```diff
@@ -6,1320 +6,480 @@
...
+		CDenseFeatures<float64_t> features(data);
 
-TEST(LibLinear,train_L2R_L2LOSS_SVC)
-{
-	LIBLINEAR_SOLVER_TYPE liblinear_solver_type = L2R_L2LOSS_SVC;
+		/*
```
karlnapf: the formatting of this comment is weird. Also the comment is superflous: if you name your variables nicely (they are) then the code is self explaining and we dont need this comment as it doesnt add any information

tests/unit/classifier/svm/LibLinear_unittest.cc
```diff
@@ -6,1320 +6,480 @@
...
+				cols = 10;
+			}
+			SGMatrix<float64_t> train_data(rows, cols);
+			SGMatrix<float64_t> test_data(2, 10); //Always 2x10, doesn't matter l1 or l2
+
+			/*
+				We have to transpose the data if its l2. If it is l1, then leave it as it is (Since this is the data of l1 originally)
+			*/
+			index_t num_samples = 10;
+			std::vector<int32_t> x{0,0,1,1,2,2,3,3,4,4,5,5,6,6,7,7,8,8,9,9};
```
karlnapf: you can initialise SGVector in the same way btw

tests/unit/classifier/svm/LibLinear_unittest.cc
```diff
@@ -6,1320 +6,480 @@
...
+			std::vector<int32_t> y{0,1,0,1,0,1,0,1,0,1,0,1,0,1,0,1,0,1,0,1};
+			std::vector<int32_t> z{-1,1,-1,0,-1,-1,-2,0,0,-1,2,2,3,1,3,2,3,3,4,2};
+
+			for(index_t i=0; i<num_samples*2; ++i){
+				if (type=="L1_SIMPLE")
+						train_data(x[i],y[i])=z[i];
+				else
+						train_data(y[i], x[i])=z[i]; //transpose
```
karlnapf: this is quite cryptic, why not do a transpose_matrix call?

tests/unit/classifier/svm/LibLinear_unittest.cc
```diff
@@ -6,1320 +6,480 @@
...
+
+			SGVector<float64_t> labels(num_samples);
+			for (index_t i = 0; i < num_samples; ++i)
+			{
+				labels[i] = (i < num_samples/2) ? 1.0 : -1.0;
+			}
+
+			ground_truth = new CBinaryLabels(labels);
+
+			SG_REF(ground_truth);
```
karlnapf: all instances you are creating yourself could be created with `some` and then you dont need to REF and UNREF them.

tests/unit/classifier/svm/LibLinear_unittest.cc
```diff
@@ -6,1320 +6,480 @@
...
-	SG_UNREF(ll);
-	SG_UNREF(train_feats);
-	SG_UNREF(test_feats);
-	SG_UNREF(ground_truth);
-	SG_UNREF(eval);
-	SG_UNREF(pred);
+TEST_F(LibLinear, train_L2R_LR)
+{
+	LIBLINEAR_SOLVER_TYPE liblinear_solver_type = L2R_LR;
+	train_with_solver(liblinear_solver_type, false, false); //No bias, and not l1
```
karlnapf: minor: pls put comments not inline but in a separate line above

tests/unit/classifier/svm/LibLinear_unittest.cc
FaroukY: Removed it and simplified the functions with extra helpers

tests/unit/classifier/svm/LibLinear_unittest.cc
FaroukY: Done

tests/unit/classifier/svm/LibLinear_unittest.cc
FaroukY: It doesn't use lapack, so I removed the guard.

tests/unit/classifier/svm/LibLinear_unittest.cc
FaroukY: Done, I replaced vector with SGVector and removed both includes for string and vector

tests/unit/classifier/svm/LibLinear_unittest.cc
FaroukY: Done, I replaced this with transpose_matrix after the training data is filled in.

tests/unit/classifier/svm/LibLinear_unittest.cc
FaroukY: I did use some in the beginning but it leads to several problems since some function calls that instantiate elements do a SG_REF with new, so in the end, when I UNREF some of them, sometimes they don't need SG_UNREF (because they used some) and some need SG_UNREF because they used new. This lead to a lot of segmentation errors so I just replaced all with new so that I keep track of what memory i need to free.

tests/unit/classifier/svm/LibLinear_unittest.cc
FaroukY: done, moved all inline comments to  separate lines.

tests/unit/classifier/svm/LibLinear_unittest.cc
karlnapf: ah yes I know this problem. leave it for now then

FaroukY:
@karlnapf Addressed the major request. 

I also did all the minor ones apart from changing new+SG_REF to some. 

The reason is a lot of the function calls that generate objects use new+SG_REF inside them, so when I SG_UNREF, I had to trace back every object to make sure whether some or SG_UNREF was used and it was just a pain (+ tons of segfaults) to keep track of what memory I allocated, so I switched everything to new+SG_REF to keep track of all the memory used. 

Let's wait for Travis, and if everything is good, this pr should be good to go.

tests/unit/classifier/svm/LibLinear_unittest.cc
```diff
@@ -16,1310 +15,456 @@
...
-		SG_REF(ground_truth);
+protected:
+	void transpose_feats_matrix()
+	{
+		//Transpose train_feats matrix
+		SGMatrix<float64_t> train_matrix = train_feats->get_feature_matrix();
+		SGMatrix<float64_t>::transpose_matrix(train_matrix.matrix,
+				train_matrix.num_rows, train_matrix.num_cols);
+
+		SG_UNREF(train_feats);
```
karlnapf: why removing the old one?

karlnapf: also `CDenseFeatures<ST>::get_transposed()`

tests/unit/classifier/svm/LibLinear_unittest.cc
```diff
@@ -16,1310 +15,456 @@
...
-	ll->set_features(train_feats);
-	ll->set_labels(ground_truth);
+			labels[i/2] = (i < data.num_cols/2) ? 1.0 : -1.0;
+		}
 
-	ll->set_liblinear_solver_type(liblinear_solver_type);
-	ll->train();
-	pred = ll->apply_binary(test_feats);
-	SG_REF(pred);
+		/*
```
karlnapf: minor: could you make those 3 lines one line?

tests/unit/classifier/svm/LibLinear_unittest.cc
karlnapf: I think you dont need an own method for this, just use the above call

tests/unit/classifier/svm/LibLinear_unittest.cc
```diff
@@ -16,1310 +15,456 @@
...
-	LIBLINEAR_SOLVER_TYPE liblinear_solver_type = L2R_L2LOSS_SVC_DUAL;
+	}
+	void generate_data_l2()
+	{
+		initialize_train_and_test_and_ground();
+	}
+	void initialize_train_and_test_and_ground_simple()
+	{
+		SGMatrix<float64_t> train_data(2, 10);
+		SGMatrix<float64_t> test_data(2, 10); //Always 2x10, doesn't matter l1 or l2
```
karlnapf: pls remove this comment, it doesnt help

tests/unit/classifier/svm/LibLinear_unittest.cc
```diff
@@ -16,1310 +15,456 @@
...
 
-	ll->set_liblinear_solver_type(liblinear_solver_type);
-	ll->train();
-	pred = ll->apply_binary(test_feats);
-	SG_REF(pred);
-	liblin_accuracy = eval->evaluate(pred, ground_truth);
+	}
+	void generate_data_l1_simple()
+	{
+			initialize_train_and_test_and_ground_simple();
```
karlnapf: what about removing the _simple suffix?

tests/unit/classifier/svm/LibLinear_unittest.cc
```diff
@@ -6,7 +6,6 @@
  *
  * Written (W) 2014 pl8787
```
karlnapf: while we are at it, could you make the license BSD? just copy it from anothoer file and maintain the old authors (add yourself)

tests/unit/classifier/svm/LibLinear_unittest.cc
```diff
@@ -16,1310 +15,456 @@
...
+		SG_REF(eval);
 
-	SG_REF(ground_truth);
-	SG_REF(train_feats);
-	SG_REF(test_feats);
-}
+		ll->set_bias_enabled(biasEnable);
+		ll->set_features(train_feats);
+		if (C_value)
+				ll->set_C(0.1,0.1); //Only in the case of L2R_L1LOSS_SVC_DUAL
```
karlnapf: the comment doesnt help.
Also formatting issues

tests/unit/classifier/svm/LibLinear_unittest.cc
```diff
@@ -16,1310 +15,456 @@
...
-		for (index_t i = 0; i < num_samples; ++i)
+		if (!biasEnable)
 		{
-			labels[i] = (i < num_samples/2) ? 1.0 : -1.0;
+			for(int i=0;i<t_w.vlen;i++)
+				EXPECT_NEAR(ll->get_w()[i], t_w[i], 1e-5);
 		}
+		else
+		{
+			for(int i=0;i<t_w.vlen;i++)
```
karlnapf: `for (auto i : range(t_w.vlen))` ... but this is minor

tests/unit/classifier/svm/LibLinear_unittest.cc
```diff
@@ -16,1310 +15,456 @@
...
+			Now fill in train_feats and test_feats, and ground_truth
+		*/
+		train_feats = (CDenseFeatures<float64_t>*)features.copy_subset(train_idx);
+		test_feats =  (CDenseFeatures<float64_t>*)features.copy_subset(test_idx);
+		ground_truth = new CBinaryLabels(labels);
 
-	liblin_accuracy = eval->evaluate(pred, ground_truth);
+		SG_REF(ground_truth);
+	}
+	void generate_data_l1()
```
karlnapf: why this wrapper, you could just call the thing inside :)

tests/unit/classifier/svm/LibLinear_unittest.cc
```diff
@@ -16,1310 +15,456 @@
...
-}
+		transpose_feats_matrix();
 
-TEST(LibLinear,train_L2R_L2LOSS_SVC_DUAL)
-{
-	LIBLINEAR_SOLVER_TYPE liblinear_solver_type = L2R_L2LOSS_SVC_DUAL;
+	}
+	void generate_data_l2()
+	{
+		initialize_train_and_test_and_ground();
```
karlnapf: same here, the wrapper doesnt help

tests/unit/classifier/svm/LibLinear_unittest.cc
```diff
@@ -16,1310 +15,456 @@
...
+	void generate_data_l1_simple()
+	{
+			initialize_train_and_test_and_ground_simple();
 
-	EXPECT_NEAR(liblin_accuracy, 1.0, 1e-6);
+			transpose_feats_matrix();
+	}
+	void generate_data_l2_simple()
+	{
+			initialize_train_and_test_and_ground_simple();
```
karlnapf: formatting is all over the place here

tests/unit/classifier/svm/LibLinear_unittest.cc
```diff
@@ -16,1310 +15,456 @@
...
-	SG_UNREF(ground_truth);
-	SG_UNREF(eval);
-	SG_UNREF(pred);
+	// bias, and not l1
+	train_with_solver(liblinear_solver_type, true, false);
 }
 
-TEST(LibLinear,simple_set_train_L2R_LR)
+/*
+* --------------------------------
```
karlnapf: remove this

tests/unit/classifier/svm/LibLinear_unittest.cc
FaroukY: Done

tests/unit/classifier/svm/LibLinear_unittest.cc
FaroukY: there is already initialize_train_and_test_and_ground() for the normal data, so I used the _simple suffix to indicate this is the simple data initializer (which is entirely different from the non simple one)

tests/unit/classifier/svm/LibLinear_unittest.cc
FaroukY: Good point, did some renaming to get rid of initialize_train_and_test_and_ground() and initialize_train_and_test_and_ground_simple()

tests/unit/classifier/svm/LibLinear_unittest.cc
FaroukY: Actually I agree, I got rid of both initialize_train_and_test_and_ground and initialize_train_and_test_and_ground_simple. Thanks!

tests/unit/classifier/svm/LibLinear_unittest.cc
FaroukY: Done

tests/unit/classifier/svm/LibLinear_unittest.cc
FaroukY: Fixed

tests/unit/classifier/svm/LibLinear_unittest.cc
FaroukY: Fixed

tests/unit/classifier/svm/LibLinear_unittest.cc
FaroukY: Changed it.

tests/unit/classifier/svm/LibLinear_unittest.cc
FaroukY: I removed the transpose_feats_matrix() method and replaced it with the calls with get_transposed().

tests/unit/classifier/svm/LibLinear_unittest.cc
FaroukY: Done, changed it to BSD license.

tests/unit/classifier/svm/LibLinear_unittest.cc
FaroukY: Not sure which 3 lines. Do you mean the if statement?

tests/unit/classifier/svm/LibLinear_unittest.cc
FaroukY: Done

FaroukY:
@karlnapf  Done.

tests/unit/classifier/svm/LibLinear_unittest.cc
karlnapf: the comment ... but it is minor

tests/unit/classifier/svm/LibLinear_unittest.cc
```diff
@@ -1,11 +1,11 @@
...
- * This program is free software; you can redistribute it and/or modify
- * it under the terms of the GNU General Public License as published by
- * the Free Software Foundation; either version 3 of the License, or
- * (at your option) any later version.
+ * This software is distributed under BSD 3-clause license (see LICENSE file).
  *
- * Written (W) 2014 pl8787
+ * Authors:
+ *	Written (W) 2014 pl8787
+ *	Refactored (R) 2018 Elfarouk Yasser
```
karlnapf: just add yourself as author in the standard way...makes later processing of headers easier

tests/unit/classifier/svm/LibLinear_unittest.cc
```diff
@@ -164,20 +152,17 @@ class LibLinear : public ::testing::Test
...
-
-	}
-	void generate_data_l2()
-	{
-		initialize_train_and_test_and_ground();
+		generate_data_l2();
+		//Transpose the train_feats matrix
+		auto old_train_feats = train_feats;
+		train_feats = train_feats->get_transposed();
+		SG_UNREF(old_train_feats);
```
karlnapf: why the unref here?

tests/unit/classifier/svm/LibLinear_unittest.cc
```diff
@@ -206,13 +191,12 @@ class LibLinear : public ::testing::Test
...
 	{
-			initialize_train_and_test_and_ground_simple();
-
-			transpose_feats_matrix();
-	}
-	void generate_data_l2_simple()
-	{
-			initialize_train_and_test_and_ground_simple();
+		generate_data_l2_simple();
+		//transpose train_feats matrix
```
karlnapf: the comment is useless ... super minor :)

tests/unit/classifier/svm/LibLinear_unittest.cc
FaroukY: @karlnapf  Doesn't get_transposed() create a new transposed matrix? So I'm just deallocating the old matrix (that i transposed)? Or does get_transposed() act in place?

tests/unit/classifier/svm/LibLinear_unittest.cc
FaroukY: Fixed

tests/unit/classifier/svm/LibLinear_unittest.cc
FaroukY: Removed it :)

tests/unit/classifier/svm/LibLinear_unittest.cc
karlnapf: I guess I would just delete the old one from the same scope that you created it in (tearDown or whatever), but this is also minor.

Make sure to run the test with valgrind to see whether the memory is fine btw

karlnapf:
Ok cool! This is much better. Lets wait for the CI and then merge it!

tests/unit/classifier/svm/LibLinear_unittest.cc
FaroukY: Ohh I see, get_transposed() handles the memory deallocation inside of it. So no need for SG_UNREF :)

karlnapf: I dont think it does ...

FaroukY:
@karlnapf I readded the SG_UNREFS before the get_transposed() since after rereading it, it doesn't handle the memory deallocation. The SG_UNREF for the transposed matrix will be called in TearDown so it should be okay now.

FaroukY:
Last CI passed, let's wait for CI for now after the last commit and merge it :)

tests/unit/classifier/svm/LibLinear_unittest.cc
FaroukY: Yea i got a bit confused. Since I store the transposed matrix always in train_feats, I need to deallocate the memory before I leave the function or else we leak. The TearDown will release the transposed matrix (which is not stored in train_feats)

karlnapf:
Did run `valgrind` to check the memory usage of this test? No leaks, no uninitialized reads?

FaroukY:
@karlnapf 
Valgrind says no leaks: 

https://ibb.co/iLR5ux

karlnapf:
Great!
Thanks so much for this clean-up. Very useful

PR #4209, opened 2018-03-18
Changed sqrt function from CMATH to std

grig-guz:
Hints on how to fix the failed builds will be appreciated!


geektoni:
Hi @grig-guz! Thanks for the contribution!

The failing builds are Travis related (we're hitting the maximum time for the build jobs) and there is nothing we can do to avoid it (maybe @vigsterkr will prove me wrong). Your PR should still be mergeable since it's not your fault.

vigsterkr:
yeah unfortunately currently our clang/gcc jobs are timeouting... i'll try to check on it but as @geektoni said it shouldn't be related to your patch.

vigsterkr:
@grig-guz could you rebase your branch so we can resolve the conflict? thnx!

grig-guz:
The problem should be solved.

vigsterkr:
@grig-guz mmm seems there's still some problem... in general you should not merge develop into your branch but rather rebase it.... please read about git in general here https://git-scm.com/doc

grig-guz:
Is it ok now?

vigsterkr:
mmm unfortunately not... it contains commits from the develop that is done by me. you need to create a pull request that only contains commits that are your patches only

grig-guz:
Ok, sure, I will close this one now then.

Issue #4191, opened 2018-02-28
remove deprecated warnings


karlnapf:
When compiling shogun at the moment (e.g. meta examples), there are a lot of deprecation warnings.

```
In file included from /home/heiko/git/shogun/build/examples/meta/cpp/evaluation/cross_validation_multiple_kernel_learning_weights_storage.cpp:3:0:
/home/heiko/git/shogun/src/shogun/classifier/mkl/MKLClassification.h:49:30: note: declared here
   static CMKLClassification* obtain_from_generic(CMachine* machine);
```

This task is to remove those via stopping to use `obtain_from_generic` and replace it by `as<T>`

Yashs744:
Sir, something like this ```static CMKLClassification* as<T>(CMachine* machine);``` ??

PR #4224, opened 2018-04-02
Replace obtain_from_generic with .as

syashakash:
Refer #4191 
This is partial wok. There are more cases that need to be re-factored.
 
PS The build is expected to fail terribly when interface options are set. This is because `build/src/interfaces/{language}/shogun{language}wrap.cxx` needs refactoring wherever `obtain_from_generic` is used.

### DIFF
--- a/src/shogun/classifier/GaussianProcessClassification.cpp
+++ b/src/shogun/classifier/GaussianProcessClassification.cpp
@@ -120,10 +120,8 @@ CBinaryLabels* CGaussianProcessClassification::apply_binary(
 		if (m_method->get_inference_type()== INF_FITC_LAPLACE_SINGLE)
 		{
 #ifdef USE_GPL_SHOGUN
-			CSingleFITCLaplaceInferenceMethod* fitc_method=
-				CSingleFITCLaplaceInferenceMethod::obtain_from_generic(m_method);
+			CSingleFITCLaplaceInferenceMethod* fitc_method = m_method->as<CSingleFITCLaplaceInferenceMethod>();
 			data=fitc_method->get_inducing_features();
-			SG_UNREF(fitc_method);
 #else
 			SG_GPL_ONLY
 #endif //USE_GPL_SHOGUN
@@ -156,10 +154,8 @@ bool CGaussianProcessClassification::train_machine(CFeatures* data)
 		if (m_method->get_inference_type()==INF_FITC_LAPLACE_SINGLE)
 		{
 #ifdef USE_GPL_SHOGUN
-			CSingleFITCLaplaceInferenceMethod* fitc_method=
-				CSingleFITCLaplaceInferenceMethod::obtain_from_generic(m_method);
+			CSingleFITCLaplaceInferenceMethod* fitc_method = m_method->as<CSingleFITCLaplaceInferenceMethod>();
 			fitc_method->set_inducing_features(data);
-			SG_UNREF(fitc_method);
 #else
 			SG_ERROR("Single FITC Laplace inference only supported under GPL.\n")
 #endif //USE_GPL_SHOGUN
--- a/src/shogun/machine/gp/FITCInferenceMethod.cpp
+++ b/src/shogun/machine/gp/FITCInferenceMethod.cpp
@@ -112,9 +112,8 @@ SGVector<float64_t> CFITCInferenceMethod::get_diagonal_vector()
 		update();
 
 	// get the sigma variable from the Gaussian likelihood model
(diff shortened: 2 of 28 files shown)


karlnapf:
Overlooking the diff, this looks good.
I dont understand why the build would break.
You are right that `as` is not available through the interfaces, but if you dont change the example code, then there should not be any problems (we will solve `obtain_from_generic` in the examples slightly differently)

karlnapf:
Are these ALL instances of `obtain_from_generic` in `src/shogun` ?

src/shogun/evaluation/GradientResult.h
```diff
@@ -43,23 +43,6 @@ class CGradientResult : public CEvaluationResult
 	 */
 	virtual const char* get_name() const { return "GradientResult"; }
 
-	/** helper method used to specialize a base class instance
```
karlnapf: dont remove these methods for now (the examples might stop working)

src/shogun/features/LatentFeatures.cpp
```diff
@@ -78,16 +78,4 @@ void CLatentFeatures::init()
 {
 	SG_ADD((CSGObject**) &m_samples, "samples", "Array of examples",
 			MS_NOT_AVAILABLE);
-}
-
-CLatentFeatures* CLatentFeatures::obtain_from_generic(CFeatures* base_feats)
```
karlnapf: as you can see, the current `obtain_from_generic` methods are inconsistent: some of the increase the reference counter, and some of them dont. This one doesnt.

`as` does NOT increase the reference counter. This explains all the segfaults in the CI.
Unfortunately, if you replace a `obtain_from_generic` that increased the reference counter, you will have to add a `SG_REF` after the call.
However, I actually think that neither `as` nor `obtain_from_generic` should increase the reference counter at all, so the best would be just to modifz the code downstream to not decrease the reference counter downstream

src/shogun/kernel/Chi2Kernel.cpp
```diff
@@ -81,17 +81,4 @@ float64_t CChi2Kernel::compute(int32_t idx_a, int32_t idx_b)
...
-CChi2Kernel* CChi2Kernel::obtain_from_generic(CKernel* kernel)
-{
-	if (kernel->get_kernel_type()!=K_CHI2)
-	{
-		SG_SERROR("Provided kernel is "
-				"not of type CChi2Kernel!\n");
-	}
-
-	/* since an additional reference is returned */
-	SG_REF(kernel);
```
karlnapf: here is an example of an increased reference counter

src/shogun/machine/gp/FITCInferenceMethod.cpp
```diff
@@ -83,19 +83,6 @@ void CFITCInferenceMethod::update()
...
-CFITCInferenceMethod* CFITCInferenceMethod::obtain_from_generic(
-		CInference* inference)
-{
-	if (inference==NULL)
-		return NULL;
-
-	if (inference->get_inference_type()!=INF_FITC_REGRESSION)
-		SG_SERROR("Provided inference is not of type CFITCInferenceMethod!\n")
-
-	SG_REF(inference);
```
karlnapf: another increase. I think the ones that do the increase are the minority

syashakash:
I have not deleted any declarations and definitions of `obtain_from_generic` so that examples work fine. Apart from that all the `obtain_from_generic` function calls in `src/shogun` have been re-factored.
I have removed the statements that decremented the reference count wherever there were calls to `obtain_from_generic` that would increase the reference count.

src/shogun/classifier/GaussianProcessClassification.cpp
```diff
@@ -120,10 +120,8 @@ CBinaryLabels* CGaussianProcessClassification::apply_binary(
 		if (m_method->get_inference_type()== INF_FITC_LAPLACE_SINGLE)
 		{
 #ifdef USE_GPL_SHOGUN
-			CSingleFITCLaplaceInferenceMethod* fitc_method=
-				CSingleFITCLaplaceInferenceMethod::obtain_from_generic(m_method);
```
karlnapf: `CSingleFITCLaplaceInferenceMethod::obtain_from_generic` method increases the reference counter, but your replacement doesnt.

Do you see what I mean here? You have written that you have dealt with this, but that is not true here

karlnapf: ah **sorry**, you in fact remove the UNREF below ...
all good!

karlnapf:
Ok cool! thx for this!

PR #4227, opened 2018-04-03
Recurrent neural nets

grig-guz:
Ongoing work, comments are greatly appreciated. Created another pull request since merging to different fork every time would probably be a pain.
#3966


grig-guz:
The overall structure seems to be very different from other NN classes that are already implemented here. Since we need to backprop through entire sequence to get gradients, we have to store to store hidden states and outputs, so let me know if storing it as vector of SGMatrix'es is the best idea.
Also, we are forced to apply tanh inside  compute_activations of NeuralRecurrentLayer, since it is required to transition to the next state. I don't really see a way to use NeuralTanhLayer here.

grig-guz:
Oops, no tests ran here:
https://travis-ci.org/shogun-toolbox/shogun/jobs/361568597
@karlnapf  @vigsterkr  could you please check what's wrong with the build?

karlnapf:
The CI errors might come from your test that fails to compile https://travis-ci.org/shogun-toolbox/shogun/jobs/362010308#L1948

It might be from the order of the includes in your test. Try including the googletest headers first (or last) dont remember)

karlnapf:
Next thing, could you chunk the PR into multiple smaller ones? I.e. make a copy of your changes locally and update this PR with a smaller number of changes. It is better to send 5 small PRs than one big one (if possible). Try to separate individual additions. It makes things so much easier to review ( the reason why this wasnt reviewed yet is because it is so big)

karlnapf:
Next thing, could you chunk the PR into multiple smaller ones? I.e. make a copy of your changes locally and update this PR with a smaller number of changes. It is better to send 5 small PRs than one big one (if possible). Try to separate individual additions. It makes things so much easier to review ( the reason why this wasnt reviewed yet is because it is so big)

src/shogun/neuralnets/NeuralLinearLayer.cpp
```diff
@@ -188,7 +188,7 @@ float64_t CNeuralLinearLayer::compute_error(SGMatrix<float64_t> targets)
 	float64_t sum = 0;
 	int32_t length = m_num_neurons*m_batch_size;
 	for (int32_t i=0; i<length; i++)
-		sum += (targets[i]-m_activations[i])*(targets[i]-m_activations[i]);
```
karlnapf: better!
Even better would be to use a single linalg call to compute this, rather than the loop.
The advantage also is that it would be SIMD and therefore faster

src/shogun/neuralnets/NeuralRecurrentLayer.h
```diff
@@ -0,0 +1,220 @@
...
+#ifndef __NEURALRECURRENTLAYER_H__
+#define __NEURALRECURRENTLAYER_H__
+
+#include <shogun/lib/common.h>
+#include <shogun/neuralnets/NeuralLinearLayer.h>
+
+namespace shogun
+{
+/** @brief Recurrent Neural layer with linear neurons, with an identity activation
+ * function. can be used as a hidden layer or an output layer
```
karlnapf: Typo, Can with capital C

src/shogun/neuralnets/NeuralRecurrentLayer.h
```diff
@@ -0,0 +1,220 @@
...
+ *
+ * Each neuron in the layer is connected to all the neurons in all the input
+ * and hidden layers that connect into this layer.
+ *
+ * The hidden state at time step t is computed according to the previous hidden state
```
karlnapf: name the hidden state, i.e. The hidden state $h_t$ at time step $t$ is computed ...
(also pls use proper math mode for all math variables)

src/shogun/neuralnets/NeuralRecurrentLayer.h
```diff
@@ -0,0 +1,220 @@
...
+ * and the input at the current state as:
+ *
+ * \f$ h_t = f(W x_t + U s_t_-_1 + b_h) \f$ where \f$ b_h \f$ is the bias vector of
```
karlnapf: the subscript s_t_-1 is wrong
$s_{t-1}$
Pls take extra care when writing math in the doxygen, you can render it locally via `make doc` and then check `index.html` it in `doc`

src/shogun/neuralnets/NeuralRecurrentLayer.h
```diff
@@ -0,0 +1,220 @@
...
+ * and \f$ U \f$ is the weights matrix between this hidden layer and itself
+ *
+ * The output layer is computed as:
+ *
+ * \f$ y_t = f(V h_t + b_y) \f$ where \f$ b_y \f$ is the bias vector of the output layer,
+ * \f$ h_t \f$ is the value of the hidden states at the current time step and \f$ V \f$ is
+ * the weights matrix between this hidden layer and the output layer
+ *
+ */
+class CNeuralRecurrentLayer : public CNeuralLinearLayer
```
karlnapf: Could you explain to me why the reccurent layer inherits from the linear layer?
I am not too familiar with the NN architechure we have but this puzzles me a bit?

src/shogun/neuralnets/NeuralRecurrentLayer.h
```diff
@@ -0,0 +1,220 @@
...
+{
+public:
+	/** default constructor */
```
karlnapf: "Default"

grig-guz:
@karlnapf sure, thanks, I'll submit PRs for every big function in these classes. I'm finishing tests for activation function of RecurrentLayer, it's quite buggy on the remote now.

src/shogun/neuralnets/NeuralRecurrentLayer.h
```diff
@@ -0,0 +1,220 @@
...
+	 * @param layers Array of layers that form the network that this layer is
+	 * being used with
+	 *
+	 * @param input_indices  Indices of the layers that are connected to this
+	 * layer as input
+	 */
+	virtual void initialize_neural_layer(CDynamicObjectArray* layers,
+			SGVector<int32_t> input_indices);
+
+	/** Sets the batch_size and allocates memory for m_activations and
```
karlnapf: The new Shogun has generic `set` and `get` methods that can change any parameter that was registered using `SG_ADD`. Therefore we dont need to add any getters or setters ...

Pls give those variables default values, and avoid "must be called before ..."

src/shogun/neuralnets/NeuralRecurrentLayer.h
```diff
@@ -0,0 +1,220 @@
...
+	 * m_input_gradients accordingly. Must be called before forward or backward
+	 * propagation is performed
+	 *
+	 * @param batch_size number of training/test cases the network is
+	 * currently working with
+	 */
+	virtual void set_batch_size(int32_t batch_size);
+
+	/** Initializes the layer's parameters. The layer should fill the given
```
karlnapf: What do you mean with "The layer should fill the given arrays .... "? This is a function call and the comment should describe what it does

src/shogun/neuralnets/NeuralRecurrentLayer.h
```diff
@@ -0,0 +1,220 @@
...
+	 * arrays with the initial value for its parameters
+	 *
+	 * @param parameters Vector of size get_num_parameters()
+	 *
+	 * @param parameter_regularizable Vector of size get_num_parameters().
```
karlnapf: I would call this "Boolean mask"

src/shogun/neuralnets/NeuralRecurrentLayer.h
```diff
@@ -0,0 +1,220 @@
...
+	 * This controls which of the layer's parameter are
+	 * subject to regularization, i.e to turn off regularization for parameter
+	 * i, set parameter_regularizable[i] = false. This is usally used to turn
+	 * off regularization for bias parameters.
+	 *
+	 * @param sigma standard deviation of the gaussian used to random the
```
karlnapf: Capital S

karlnapf: A few comments

* "random" is not a verb. Should be be explicit about what actually is done
* What if I want non-Gaussian noise to be added? Shouldnt this be more modular?

src/shogun/neuralnets/NeuralRecurrentLayer.h
grig-guz: Ooops, you're right. That one should inherit from NeuralLayer.

src/shogun/neuralnets/NeuralRecurrentLayer.h
```diff
@@ -0,0 +1,220 @@
...
+	 * parameters
+	 */
+	virtual void initialize_parameters(SGVector<float64_t> parameters,
+			SGVector<bool> parameter_regularizable,
+			float64_t sigma);
+
+	/** Computes the activations of the neurons in this layer, results should
```
karlnapf: "results should be stored" should be "results are stored"
However, this is public user API? Then there is not need to talk about internal states (this are hidden from the user)
If the method is an internal helper, it should be protected/private and then the docs can talk about internal states ...

src/shogun/neuralnets/NeuralRecurrentLayer.cpp
```diff
@@ -0,0 +1,190 @@
...
+		CDynamicObjectArray* layers)
+{
+	float64_t* hidden_biases = parameters.vector;
+	float64_t* output_biases = parameters.vector + m_num_neurons;
+
+	typedef Eigen::Map<Eigen::MatrixXd> EMappedMatrix;
+	typedef Eigen::Map<Eigen::VectorXd> EMappedVector;
+
+	EMappedMatrix  A(m_activations.matrix, m_output_dim, m_batch_size);
+	EMappedMatrix  H(m_hidden_activation.matrix, m_num_neurons, m_batch_size);
```
karlnapf: pls use linalg and not eigen

src/shogun/neuralnets/NeuralRecurrentLayer.cpp
```diff
@@ -0,0 +1,190 @@
...
+		// parameter_regularizable[i] = (i >= m_num_neurons);
+	}
+}
+// VERY BIG TODO: Figure out how to manage input from multiple layers
+// Now we assume that we only have 1 input layer
+
+void CNeuralRecurrentLayer::compute_activations(SGVector<float64_t> parameters,
+		CDynamicObjectArray* layers)
+{
+	float64_t* hidden_biases = parameters.vector;
```
karlnapf: pls no raw pointers

src/shogun/neuralnets/NeuralRecurrentLayer.cpp
```diff
@@ -0,0 +1,190 @@
...
+	typedef Eigen::Map<Eigen::VectorXd> EMappedVector;
+
+	EMappedMatrix  A(m_activations.matrix, m_output_dim, m_batch_size);
+	EMappedMatrix  H(m_hidden_activation.matrix, m_num_neurons, m_batch_size);
+	EMappedVector  Bh(hidden_biases, m_num_neurons);
+	EMappedVector  By(output_biases, m_num_neurons);
+
+	int32_t weights_index_offset = m_num_neurons * 2;
+
+	for (int i = 0; i < m_time_series_length; i++) {
```
karlnapf: auto

src/shogun/neuralnets/NeuralRecurrentLayer.cpp
```diff
@@ -0,0 +1,190 @@
...
+			(CNeuralLayer*)layers->element(m_input_indices[0]);
+
+		EMappedMatrix X(layer->get_activations().matrix,
+				layer->get_num_neurons(), m_batch_size);
+
+		// TODO Maybe do something to be able to choose ReLU instead,
+		// but ReLU kill too much gradient when unrolling the recursion
+		// in backprop
+		H += (Wxh * X + Whh * H).colwise() + Bh;
+		H = H.unaryExpr<float64_t(*)(float64_t)>(&std::tanh);
```
karlnapf: this should definitely be modular!

src/shogun/neuralnets/NeuralRecurrentLayer.cpp
```diff
@@ -0,0 +1,190 @@
...
+		SG_UNREF(layer);
+	}
+}
+
+void CNeuralRecurrentLayer::compute_gradients(
+		SGVector<float64_t> parameters,
+		SGMatrix<float64_t> targets,
+		CDynamicObjectArray* layers,
+		SGVector<float64_t> parameter_gradients)
+{
```
karlnapf: ?

grig-guz:
@karlnapf Thank you very much for the reviews! 
P.S. I tried to follow the style/structure of already implemented layers (linear, ReLU, etc.), and this TanhLayer test is a slightly tweaked copy of ReLU test. Seems like, given your comments, old implementations require A LOT of updates. I think we could create an issue for somebody to pretty much follow your comments in this PR and update LinearLayer, NeuralLayer and tests.

karlnapf:
Yes, definitely a lot of work. But hey -- the code is old, and old code requires love :)
I realise you followed the structure and that is totally fine. However, let`s not make the mess bigger. If we are going to extend this, lets at least get some improvements into place as well, to keep this maintainable and extendable. A lot of Shogun is like that: you wanna do something new, you first have to fix the mess (at least some parts of it)
I agree it is a good idea to isolate some of the comments in issues. Do you want to go ahead with some of that?

karlnapf:
We can discuss more details in IRC.

PR #4228, opened 2018-04-03
add progress bar to RBM fix #3889

### LINKED ISSUE #3889: Systematically test all iterative methods
All models that are trained iteratively should

 * have a progress bar (add to missing ones)
 * have early stopping implemented (with working predictions afterwards) (unit test)
 * should give a *warning* when not converged (unit test), not fail.
 * Have the option to "continue" training more, if that makes sense (if not converged, or early stopped) (unit test)

This is a mix of automatic unit testing, API updates, and bug-fixes in the methods

Example: ICA methods, @geektoni knows all the other models

@geektoni @micmn maybe you can address a few of those

sunalbert:
add progress bar to RBM algorithm

### DIFF
--- a/src/shogun/neuralnets/RBM.cpp
+++ b/src/shogun/neuralnets/RBM.cpp
@@ -35,6 +35,7 @@
 
 
 #include <shogun/base/Parameter.h>
+#include <shogun/base/progress.h>
 #include <shogun/mathematics/Math.h>
 #include <shogun/mathematics/eigen3.h>
 
@@ -137,7 +138,7 @@ void CRBM::train(CDenseFeatures<float64_t>* features)
 
 	int32_t counter = 0;
 
-	auto pb = progress(range(0， max_num_epochs));
+	auto pb = progress(range(0, max_num_epochs));
 
 	for (int32_t i=0; i<max_num_epochs; i++)
 	{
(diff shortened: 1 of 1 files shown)


karlnapf:
looks good to me, lets wait for the CI.
thanks for the patch! :)

src/shogun/neuralnets/RBM.cpp
```diff
@@ -136,6 +137,9 @@ void CRBM::train(CDenseFeatures<float64_t>* features)
 		buffer = SGMatrix<float64_t>(m_num_hidden, m_batch_size);
 
 	int32_t counter = 0;
+
+	auto pb = progress(range(0, max_num_epochs));
+
 	for (int32_t i=0; i<max_num_epochs; i++)
```
geektoni: You could have used the compact version here, instead of the `pb.print_progress/complete` methods :) namely:
`for (auto i : progress(range(0, max_num_epochs)))`

sunalbert: Ok, thanks. I will modify the code.

karlnapf:
Ill merge once CI is green

sunalbert:
The CI fails again. How can I fix the error "The job exceeded the maximum time limit for jobs, and has been terminated"

geektoni:
Unfortunately, that error is not something you can fix from your side because it is caused by Travis itself. However, this PR should be mergeable since all the other builds are green.

sunalbert:
Ok, thank you

Issue #4074, opened 2018-01-08
Refactor log-determinant estimation code to work without computation engines


karlnapf:
We want to delete the files in `lib/computation/*`

However, we don't want to delete @lambday 's nice code to estimate log-determinants of large sparse matrices.

This task is to re-factor things such that we can delete the computation classes. This is a nice entrance task that does only require to read/refactor existing code and tests. This will give you a nice first exposure to the internals of shogun without being overwhelming. Good non-trivial entrance task therefore.

Here is a rough list of steps
0. Read through the code, coming from the log-determinant examples
1. Compile Shogun withe the log-det code enabled (see guards in code).
2, Isolate and run all log-det tests
3. Delete all files in `shogun/lib/computation`. Re-run `cmake` and `make`. The build is now broken, i.e. shogun won't compile
4. Fix the build (will require slight API changes)
5. Make sure all tests work
6. Send PR

shubham808:
@karlnapf i want to take this up.

shubham808:
After looking at the code I got rid of ```computation engines``` and directly computed jobs treating them as ```independent jobs```.
So I deleted ```engines``` directory and then debugged to get everything to work. 
Is this the right approach ? And should I keep going similarly to eventually get rid of all classes in ```computation``` directory ?

karlnapf:
yes, thats exactly it. make sure to send a pr soon to we can give feedback.
Keep in mind we also want to get rid of independent jobs.
Best would be if you used openmp instead for now

shubham808:
@karlnapf using ```omp``` is a good idea. I got rid of ```IndependentJob``` and also the ```aggregators``` are not needed too. However, I am stuck at getting rid of ```JobResult``` classes also i think keeping them around would simplify things.

salonirk11:
Also, is it necessary to remove ```lib/computation/aggregator``` ? Can we reuse these classes from ```mathematics/linalg/rattaprox/logdet/computation/aggregator```?

karlnapf:
Everything in the `computation` folder should go

karlnapf:
Done via #4103

PR #4229, opened 2018-04-03
deletes everything we deprecated for #4074

shubham808:


tests/unit/mathematics/linalg/RationalApproximationIndividualJob_unittest.cc
```diff
@@ -1,123 +0,0 @@
...
-	float64_t result=r->get_result();
-	SG_UNREF(agg);
-	SG_UNREF(op);
-	SG_UNREF(identity);
-	SG_UNREF(solver);
-
-	EXPECT_NEAR(result, -0.73170731707317049, 1E-15);
-}
-
-TEST(RationalApproximationIndividualJob, compute_cocg)
```
karlnapf: are these things covered by other tests? I mean the numerical results ... If not, we need to do that!

shubham808:
Yeah I am working on that....  I will make a separate pr :)

PR #4230, opened 2018-04-04
add some cmake errors to prevent linking against GPL libraries

karlnapf:


src/shogun/CMakeLists.txt
```diff
@@ -360,7 +361,8 @@ SHOGUN_DEPENDENCIES(
 SHOGUN_DEPENDENCIES(
   LIBRARY LibLZMA
   SCOPE PRIVATE
-  CONFIG_FLAG USE_LZMA)
+  CONFIG_FLAG USE_LZMA
+  GPL_ONLY)
```
vigsterkr: because liblzma is gpl?

src/shogun/CMakeLists.txt
```diff
@@ -370,7 +372,8 @@ SHOGUN_DEPENDENCIES(
 SHOGUN_DEPENDENCIES(
   LIBRARY LZO
   SCOPE PRIVATE
-  CONFIG_FLAG USE_LZO)
+  CONFIG_FLAG USE_LZO
+  GPL_ONLY)
```
vigsterkr: because liblzo is gpl?

src/shogun/CMakeLists.txt
```diff
@@ -400,18 +403,22 @@ ELSE()
   SHOGUN_DEPENDENCIES(
     LIBRARY NLopt
     SCOPE PRIVATE
-    CONFIG_FLAG HAVE_NLOPT)
+    CONFIG_FLAG HAVE_NLOPT
+    GPL_ONLY)
```
vigsterkr: because nlopt is LGPL? note it's lgpl and not gpl

src/shogun/CMakeLists.txt
```diff
@@ -400,18 +403,22 @@ ELSE()
...
 ENDIF()
 
 SHOGUN_DEPENDENCIES(
   LIBRARY LpSolve
   SCOPE PRIVATE
-  CONFIG_FLAG USE_LPSOLVE)
+  CONFIG_FLAG USE_LPSOLVE
+  GPL_ONLY)
```
vigsterkr: because lpsolve ig LGPL?

src/shogun/CMakeLists.txt
```diff
@@ -400,18 +403,22 @@ ELSE()
...
 
 SHOGUN_DEPENDENCIES(
   LIBRARY ColPack
   SCOPE PRIVATE
-  CONFIG_FLAG HAVE_COLPACK)
+  CONFIG_FLAG HAVE_COLPACK
```
vigsterkr: because colpack is LGPL?

src/shogun/CMakeLists.txt
karlnapf: it a actually is PD, I got this one wrong

src/shogun/CMakeLists.txt
karlnapf: yep you are right this we can link against

src/shogun/CMakeLists.txt
karlnapf: will remove

karlnapf:
My motivation was: linking (even dynamically) against GPL code makes the code GPL. Therefore, I added some guards to prevent this. Correct me if I am wrong.
Definitely the LGPL things need to be removed.

src/shogun/CMakeLists.txt
karlnapf: I just realised we can also move back the `CNLOPTMinimizer` class to the main repo, which is very good as then all the GP codes are still in the BSD release!

karlnapf:
Apparently the FSF sees dynamic linking as derivative work:
https://en.wikipedia.org/wiki/GNU_General_Public_License#Linking_and_derived_works

vigsterkr:
yeah but in case you use it as a plugin, then it never violates the gpl.

vigsterkr:
so, in this particular case if `src/shogun/lib/Compressor.cpp` would be written in a modular fashion and the lzo decoder would be a plugin then all this can be ignored... for the time being i'd rather suggest to drop LZO as a whole.... as anyways this should not effect anything in the base library  - once modular.

vigsterkr:
and i dont really see anybody complaining of not being able to decompress string features that are lzo compressed... and because of this doxygen comment:
```Note that besides lzo compression, this library is thread safe.```

vigsterkr:
so on the end only GLPK would need this if we call dynamic linking a derivative work, but as soon as there's plugin support this would need to be reverted....

karlnapf:
Sure then let`s leave it

PR #4234, opened 2018-04-06
add progress bar to FastICA

sunalbert:
add progress bar to FastICA algorithm.

### DIFF
--- a/src/shogun/converter/ica/FastICA.cpp
+++ b/src/shogun/converter/ica/FastICA.cpp
@@ -129,12 +129,7 @@ CFeatures* CFastICA::apply(CFeatures* features)
 	W = sym_decorrelation(W);
 
 	float64_t lim = tol+1;
-	for (auto i : progress(
-	    range(0, max_iter),
-	    "PROGRESS: ",
-	    UTF8,
-	    [&] { return lim > tol; })
-	)
+	for (auto i : progress(range(0, max_iter),[&] { return lim > tol; }))
 	{
 		MatrixXd wtx = W * WX;
 
--- a/src/shogun/converter/ica/FastICA.cpp
+++ b/src/shogun/converter/ica/FastICA.cpp
@@ -8,7 +8,7 @@
 
 #include <shogun/features/DenseFeatures.h>
 
-
+#include <shogun/base/progress.h>
 #include <shogun/mathematics/Math.h>
 #include <shogun/mathematics/eigen3.h>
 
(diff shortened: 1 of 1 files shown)


src/shogun/converter/ica/FastICA.cpp
```diff
@@ -128,9 +128,9 @@ CFeatures* CFastICA::apply(CFeatures* features)
 
 	W = sym_decorrelation(W);
 
-	int iter = 0;
 	float64_t lim = tol+1;
-	while (lim > tol && iter < max_iter)
+	for (auto i : progress(
+	         range(0, max_iter), "PROGRESS: ", UTF8, [&] { return lim > tol; }))
```
iglesias: Minor, for consistency, indent and break lines like in the doc reference you've cited.

sunalbert: Thanks @iglesias . The style has been modified.

iglesias:
Sorry that I did not mention it earlier, I have noticed only now the number of commits. Can you squash the 4 into 1? Thanks!

src/shogun/converter/ica/FastICA.cpp
```diff
@@ -130,7 +130,11 @@ CFeatures* CFastICA::apply(CFeatures* features)
 
 	float64_t lim = tol+1;
 	for (auto i : progress(
-	         range(0, max_iter), "PROGRESS: ", UTF8, [&] { return lim > tol; }))
+	    range(0, max_iter),
+		"PROGRESS: ",
```
iglesias: Put this line and the two below at the same indentation than range, like in the doc we were referring to.

sunalbert:
Thanks @iglesias I will change the code style and squash all the commits into one.

karlnapf:
@iglesias we can now squash with github, so people can just keep on adding commits to PRs (better to see trajectory, history, and preserving discussions (at least for small changes that make sense to group into a single commit)

src/shogun/converter/ica/FastICA.cpp
```diff
@@ -128,9 +128,13 @@ CFeatures* CFastICA::apply(CFeatures* features)
 
 	W = sym_decorrelation(W);
 
-	int iter = 0;
 	float64_t lim = tol+1;
-	while (lim > tol && iter < max_iter)
+	for (auto i : progress(
+	    range(0, max_iter),
+	    "PROGRESS: ",
```
karlnapf: @geektoni this seems a bit verbose. Isnt there a shorter version?

karlnapf: Sorry! I just only realise this is the advanced way of using it

karlnapf: @geektoni Could we have a way to pass lambda functions that does not require to pass the string and the encoding?
I.e. something as compact as the `for (auto i: progress(range(0, 10)))` but with a custom stopping condition?

geektoni: #4240

karlnapf: @sunalbert could you rebase and use the new function?

sunalbert: Ok, Thanks @iglesias @karlnapf

iglesias: Of course, you are welcome!

karlnapf:
All good the error is not caused by you

PR #4236, opened 2018-04-07
factory methods in lda meta example #4247

shubham808:
```lda.get_real("m_gamma")``` did not work due to this casting of gamma(float) as int

### DIFF
--- a/examples/meta/src/binary/linear_discriminant_analysis.sg
+++ b/examples/meta/src/binary/linear_discriminant_analysis.sg
@@ -1,13 +1,13 @@
-CSVFile f_feats_train("../../data/classifier_binary_2d_linear_features_train.dat")
-CSVFile f_feats_test("../../data/classifier_binary_2d_linear_features_test.dat")
-CSVFile f_labels_train("../../data/classifier_binary_2d_linear_labels_train.dat")
-CSVFile f_labels_test("../../data/classifier_binary_2d_linear_labels_test.dat")
+File f_feats_train = csv_file("../../data/classifier_binary_2d_linear_features_train.dat")
+File f_feats_test = csv_file("../../data/classifier_binary_2d_linear_features_test.dat")
+File f_labels_train = csv_file("../../data/classifier_binary_2d_linear_labels_train.dat")
+File f_labels_test = csv_file("../../data/classifier_binary_2d_linear_labels_test.dat")
 
 #![create_features]
 Features features_train = features(f_feats_train)
 Features features_test = features(f_feats_test)
-BinaryLabels labels_train(f_labels_train)
-BinaryLabels labels_test(f_labels_test)
+Labels labels_train = labels(f_labels_train)
+Labels labels_test = labels(f_labels_test)
 #![create_features]
 
 #![create_instance]
@@ -16,17 +16,15 @@ Machine lda = machine("LDA", labels=labels_train)
 
 #![train_and_apply]
 lda.train(features_train)
-BinaryLabels labels_predict = lda.apply_binary(features_test)
+Labels labels_predict = lda.apply(features_test)
+RealVector labels = labels_predict.get_real_vector("labels")
 #![train_and_apply]
 
 #![extract_weights]
-RealVector w = lda.get_real_vector("w")
+RealVector weights = lda.get_real_vector("w")
 #![extract_weights]
 
 #![evaluate_accuracy]
-AccuracyMeasure eval()
+Evaluation eval = evaluation("AccuracyMeasure")
 real accuracy = eval.evaluate(labels_predict, labels_test)
 #![evaluate_accuracy]
-
-#additional integration testing variables
-RealVector output = labels_predict.get_labels()
--- a/src/shogun/classifier/LDA.cpp
+++ b/src/shogun/classifier/LDA.cpp
@@ -49,7 +49,7 @@ void CLDA::init()
 	SG_ADD(
 	    (machine_int_t*)&m_method, "m_method",
 	    "Method used for LDA calculation", MS_NOT_AVAILABLE);
-	SG_ADD(&m_gamma, "m_gamma", "Regularization parameter", MS_NOT_AVAILABLE);
+	SG_ADD(&m_gamma, "m_gamma", "Regularization parameter", MS_AVAILABLE);
 	SG_ADD(&m_bdc_svd, "m_bdc_svd", "Use BDC-SVD algorithm", MS_NOT_AVAILABLE);
 }
 
@@ -105,16 +105,16 @@ template <typename ST>
 bool CLDA::solver_svd()
 {
 	auto dense_feat = static_cast<CDenseFeatures<ST>*>(features);
-  auto labels = multiclass_labels(m_labels);
-  REQUIRE(labels->get_num_classes()==2, "Number of classes exceeds 2 for binary classifier: %d provided\n", labels->get_num_classes())
+	auto labels = multiclass_labels(m_labels);
+	REQUIRE(
+	    labels->get_num_classes() == 2, "Number of classes (%d) must be 2\n",
+	    labels->get_num_classes())
 
 	// keep just one dimension to do binary classification
 	const index_t projection_dim = 1;
 	auto solver = std::unique_ptr<LDACanVarSolver<ST>>(
 	    new LDACanVarSolver<ST>(
-	        dense_feat,
-	        labels,
-	        projection_dim, m_gamma, m_bdc_svd));
+	        dense_feat, labels, projection_dim, m_gamma, m_bdc_svd));
 
 	SGVector<ST> w_st(solver->get_eigenvectors());
 
@@ -140,15 +140,14 @@ template <typename ST>
 bool CLDA::solver_classic()
 {
 	auto dense_feat = static_cast<CDenseFeatures<ST>*>(features);
-  auto labels = multiclass_labels(m_labels);
-  REQUIRE(labels->get_num_classes()==2, "Number of classes exceeds 2 for binary classifier: %d provided\n", labels->get_num_classes())
-  index_t num_feat = dense_feat->get_num_features();
+	auto labels = multiclass_labels(m_labels);
+	REQUIRE(
+	    labels->get_num_classes() == 2, "Number of classes (%d) must be 2\n",
+	    labels->get_num_classes())
+	index_t num_feat = dense_feat->get_num_features();
 
 	auto solver = std::unique_ptr<LDASolver<ST>>(
-	    new LDASolver<ST>(
-	        dense_feat,
-          labels,
-	        m_gamma));
+	    new LDASolver<ST>(dense_feat, labels, m_gamma));
 
 	auto class_mean = solver->get_class_mean();
 	auto class_count = solver->get_class_count();
--- a/src/shogun/classifier/LDA.cpp
+++ b/src/shogun/classifier/LDA.cpp
@@ -49,9 +49,7 @@ void CLDA::init()
 	SG_ADD(
 	    (machine_int_t*)&m_method, "m_method",
 	    "Method used for LDA calculation", MS_NOT_AVAILABLE);
-	SG_ADD(
-	    (machine_int_t*)&m_gamma, "m_gamma", "Regularization parameter",
-	    MS_NOT_AVAILABLE);
+	SG_ADD(&m_gamma, "m_gamma", "Regularization parameter", MS_NOT_AVAILABLE);
 	SG_ADD(&m_bdc_svd, "m_bdc_svd", "Use BDC-SVD algorithm", MS_NOT_AVAILABLE);
 }
 
@@ -62,10 +60,6 @@ CLDA::~CLDA()
 bool CLDA::train_machine(CFeatures *data)
 {
 	REQUIRE(m_labels, "Labels for the given features are not specified!\n")
-	REQUIRE(
-	    m_labels->get_label_type() == LT_BINARY,
-	    "The labels should of type CBinaryLabels! Provided type is %s \n",
-	    m_labels->get_name())
 
 	if(data)
 	{
@@ -77,12 +71,6 @@ bool CLDA::train_machine(CFeatures *data)
 	{
 		REQUIRE(data, "Features have not been provided.\n")
 	}
-
-	REQUIRE(
-	    features->get_num_vectors() == m_labels->get_num_labels(),
-	    "Number of training examples(%d) should be equal to number of labels "
-	    "(%d)!\n",
-	    features->get_num_vectors(), m_labels->get_num_labels());
 
 	REQUIRE(
 	    features->get_feature_class() == C_DENSE,
@@ -117,13 +105,15 @@ template <typename ST>
 bool CLDA::solver_svd()
 {
 	auto dense_feat = static_cast<CDenseFeatures<ST>*>(features);
+  auto labels = multiclass_labels(m_labels);
+  REQUIRE(labels->get_num_classes()==2, "Number of classes exceeds 2 for binary classifier: %d provided\n", labels->get_num_classes())
 
 	// keep just one dimension to do binary classification
 	const index_t projection_dim = 1;
 	auto solver = std::unique_ptr<LDACanVarSolver<ST>>(
 	    new LDACanVarSolver<ST>(
 	        dense_feat,
-	        new CMulticlassLabels(static_cast<CBinaryLabels*>(m_labels)),
+	        labels,
 	        projection_dim, m_gamma, m_bdc_svd));
 
 	SGVector<ST> w_st(solver->get_eigenvectors());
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: examples/meta/src/binary/linear_discriminant_analysis.sg [hunk 1]
```diff
@@ -11,18 +11,17 @@ BinaryLabels labels_test(f_labels_test)
...
-lda.set_features(features_train)
-lda.set_labels(labels_train)
+Machine lda = machine("LDA")
+lda.put("labels", labels_train)
 #![create_instance]
 
 #![train_and_apply]
-lda.train()
+lda.train(features_train)
 BinaryLabels labels_predict = lda.apply_binary(features_test)
```
karlnapf: this can just be "apply"

src/shogun/classifier/LDA.cpp
```diff
@@ -49,9 +49,7 @@ void CLDA::init()
 	SG_ADD(
 	    (machine_int_t*)&m_method, "m_method",
 	    "Method used for LDA calculation", MS_NOT_AVAILABLE);
-	SG_ADD(
-	    (machine_int_t*)&m_gamma, "m_gamma", "Regularization parameter",
-	    MS_NOT_AVAILABLE);
+	SG_ADD(&m_gamma, "m_gamma", "Regularization parameter", MS_NOT_AVAILABLE);
```
karlnapf: I think this might break the integration tests (did you check?)
If so, pls generate a new test data file

karlnapf:
Nice, thanks of for the patch.
This is an incomplete refactoring though, see `binary/kernel_support_vector_machine.sg` for a fully refactored example (including loading the files etc)

karlnapf:
I think fully changing the examples will break the ruby tests though.
 (due to a potential bug in swig ruby)
That is fine, I have discussed this with @lisitsyn and we decided to disable the tests for ruby for now in order to move on with the refactoring ...

shubham808:
I cant create a BinaryLabels object by using factory methods at the moment. Due to [this](https://github.com/shogun-toolbox/shogun/blob/32a6746251e4864269feb08702786ed1214be0b6/src/shogun/util/factory.h#L70) the type check in LDA.cpp [here](https://github.com/shogun-toolbox/shogun/blob/32a6746251e4864269feb08702786ed1214be0b6/src/shogun/classifier/LDA.cpp#L67) fails. I think we will have to tweak factory.h a bit more...

src/shogun/classifier/LDA.cpp
shubham808: no test broken :)

karlnapf:
Check this commit: https://github.com/shogun-toolbox/shogun/commit/c85feddc80340c61261f1a57328521adf889bc0e#diff-197df1e4436c2e76a5d4e179a2ee6ac6

The way to get around this problem is to replace the runtime type assertion with a conversion, i.e. replace something like
```
ASSERT(m_labels->get_label_type() == LT_BINARY)
```
with
```
auto binary_labels = m_labels->as_binary();
```

(and then use `binary_labels` downstream). Function defined here: https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/labels/BinaryLabels.cpp#L161

examples/meta/src/binary/linear_discriminant_analysis.sg
```diff
@@ -1,34 +1,31 @@
...
+Features features_test = features(f_feats_test)
+Labels labels_train = labels(f_labels_train)
+Labels labels_test = labels(f_labels_test)
 #![create_features]
 
 #![create_instance]
-LDA lda()
-lda.set_features(features_train)
-lda.set_labels(labels_train)
+Machine lda = machine("LDA")
```
karlnapf: pls use the kwargs construct. Check `meta_api/kwargs.sg` for how to, or the kernel svm example I posted

src/shogun/classifier/LDA.cpp
```diff
@@ -62,10 +60,6 @@ CLDA::~CLDA()
 bool CLDA::train_machine(CFeatures *data)
 {
 	REQUIRE(m_labels, "Labels for the given features are not specified!\n")
-	REQUIRE(
```
karlnapf: You cannot just remove such a check, this will cause problems.
The reason why we have this check is that we cast the labels to binary later on.

E.g. there is a line
```
new CMulticlassLabels(static_cast<CBinaryLabels*>(m_labels))
```

Keep in mind `m_labels` here is of type `CLabels*`.

If you remove the check doing a static cast might cause memory errors. You will need to replace any static cast with a conversion, as in the commit I posted earlier.
Furthermore here, you could directly convert to multiclass labels, rather than going via the binary ones (though it is not clear whether that works yet, you will need to experiment)

Pls add a unit test where you put in labels of the wrong type and make sure that an exception is thrown (rather than the program crashing). This will also make sure that you understand what is going on

shubham808: I understand.. i am replacing static casts with conversions like in kernel svm. That line looks like ```new CMulticlassLabels(binary_labels(m_labels))``` the check can now be removed because we are already making sure of correct label type downstream (in binary_labels).

About directly converting to dense labels to multiclass that is not yet complete [here](https://github.com/shogun-toolbox/shogun/blob/32a6746251e4864269feb08702786ed1214be0b6/src/shogun/labels/MulticlassLabels.cpp#L196).

tests/unit/classifier/LDA_unittest.cc
```diff
@@ -233,6 +233,31 @@ TEST(LDA, DISABLED_CheckProjection_SVD)
...
+
+	labels = new CRegressionLabels(lab);
+	features = new CDenseFeatures<float64_t>(feat);
+
+	SG_REF(labels);
+	SG_REF(features);
+
+	CLDA lda_SVD(0, features, labels, SVD_LDA);
+
+	EXPECT_THROW(lda_SVD.train(), ShogunException);
```
karlnapf: great!

few comments

* can we avoid all this redundant code via re-use the code from other tests?
* could you test other label types as well (like to test the multiclass conversion)?

src/shogun/classifier/LDA.cpp
```diff
@@ -154,8 +141,7 @@ bool CLDA::solver_classic()
 
 	auto solver = std::unique_ptr<LDASolver<ST>>(
 	    new LDASolver<ST>(
-	        dense_feat,
-	        new CMulticlassLabels(static_cast<CBinaryLabels*>(m_labels)),
+	        dense_feat, new CMulticlassLabels(binary_labels(m_labels)),
```
karlnapf: pls lets not do this, but directly convert to multiclass

karlnapf: You, all you need to do is to add a case in `multiclass_labels`, which deals with the case when `LT_BINARY` is provided. The conversion will be the same as in the constructor you mentioned (just using `std::transform`, and some checks). What is good about this is that **if** the user passed multiclass labels in the first place, there would no conversion. If the user passes binary labels, everything is as before (nonsense conversion, but we can leave that for now as we are working on a different topic here)

I don't understand what you mean re the contiguous labels.

shubham808: yes we can take care of binary and multiclass easily however if we pass dense labels we will need the std::transform conversion (like in binary) too. That is do the same thing we do with binary labels (converting ```[-1, 1]``` to ```[0, 1]```)except now the labels are dense. This is not already present in multiclass_labels but instead we have a require statement that checks the dense labels to be of type ```[0, 1...]``` (this is why we needed to convert first to binary and then  to multiclass earlier) what we could now is replace the require statement with a loop that does exactly that with discrete valued dense labels.

shubham808:
I am not sure if this is how it should be done... please guide me

karlnapf:
What being done? :)

src/shogun/labels/MulticlassLabels.cpp
```diff
@@ -193,17 +193,13 @@ namespace shogun
...
-		// For now, if we don't enforce this, there will be crashes as the
-		// class is used to index vectors
-		REQUIRE(
-		    (*std::min_element(unique.begin(), unique.end())) == 0 &&
-		        (*std::max_element(unique.begin(), unique.end())) ==
-		            (index_t)unique.size() - 1,
-		    "Multiclass labels must be contiguous integers in [0, ..., "
-		    "num_classes -1].\n");
-
+		// anything discrete should be a possible multiclasslabels.
```
karlnapf: few comments here:

* if there is no conversio needed, then no conversion should be done (i.e. leave the check in and only convert if that is false)
* pls do only copy the label vector **once**, currently I think you are copying things multiple times (your loop, `set_int_labels`, ...)
* why do you call `set_int_labels`? I would rather just convert the label vector in here and then pass that vector itself to `some<CMulticlassLabels>`, rather than modifying the given label
* we might in fact want to make the passed `orig` pointer const to avoid changing the original structure ....

src/shogun/labels/MulticlassLabels.cpp
```diff
@@ -219,6 +215,8 @@ namespace shogun
 				    orig->as<CMulticlassLabels>());
 			case LT_DENSE_GENERIC:
 				return to_multiclass(orig->as<CDenseLabels>());
+			case LT_BINARY:
+				return some<CMulticlassLabels>(orig->as<CBinaryLabels>());
```
karlnapf: This constructor is duplicate code with the thing you wrote above, could you unify that?

src/shogun/labels/MulticlassLabels.cpp
karlnapf: actually, I am just working on this anyways, so let me do the fix and then you can use my code....stay tuned

karlnapf: #4251

shubham808:
#4251 multiclass_labels looks really good now :)

tests/unit/classifier/LDA_unittest.cc
shubham808: will it not be better to add the regressionlabels exception test in multiclasslabels unit test ?

src/shogun/classifier/LDA.cpp
```diff
@@ -150,12 +140,14 @@ template <typename ST>
 bool CLDA::solver_classic()
 {
 	auto dense_feat = static_cast<CDenseFeatures<ST>*>(features);
-	index_t num_feat = dense_feat->get_num_features();
+  auto labels = multiclass_labels(m_labels);
```
iglesias: Have a look at the output of the style checker ;-)

src/shogun/classifier/LDA.cpp
```diff
@@ -117,13 +105,15 @@ template <typename ST>
 bool CLDA::solver_svd()
 {
 	auto dense_feat = static_cast<CDenseFeatures<ST>*>(features);
+  auto labels = multiclass_labels(m_labels);
```
iglesias: Minor style fixing needed here as well.

tests/unit/classifier/LDA_unittest.cc
```diff
@@ -233,6 +203,32 @@ TEST(LDA, DISABLED_CheckProjection_SVD)
...
 }
 
+// label type exception test
+TEST(LDA, num_classes_in_labels_exception)
+{
+	SGVector<float64_t> lab;
+	SGMatrix<float64_t> feat;
+
+	CDenseLabels* labels;
+	CDenseFeatures<float64_t>* features;
```
iglesias: For conciseness and better readiblity, write the variable declarations together with their initialization, for the ones  where it's possible.

src/shogun/classifier/LDA.cpp
shubham808: yeah i am aware of that

src/shogun/classifier/LDA.cpp
shubham808: actually this pr needs some more work so i am holding style check off till the end :)

shubham808:
@iglesias locally all tests passed... MixtureModel has been acting weird lately i think its not related to our code here, take a look at #4241

src/shogun/classifier/LDA.cpp
iglesias: For sure, then I won't be shouting about them 🙊

examples/meta/src/binary/linear_discriminant_analysis.sg
```diff
@@ -1,34 +1,30 @@
...
+RealVector labels = labels_predict.get_real_vector("labels")
 #![train_and_apply]
 
 #![extract_weights]
-RealVector w = lda.get_w()
+RealVector weights = lda.get_real_vector("w")
 #![extract_weights]
 
 #![evaluate_accuracy]
 AccuracyMeasure eval()
```
karlnapf: since you are touching this, could you use a factory here:
`Evaluation eval = evaluation("AccuracyMeasure")`

karlnapf: since you are touching this, could you use a factory here:
`Evaluation eval = evaluation("AccuracyMeasure")`

shubham808:
@karlnapf can we go over this once again ?

karlnapf:
Totally, let me go through it

examples/meta/src/binary/linear_discriminant_analysis.sg
```diff
@@ -16,17 +16,15 @@ Machine lda = machine("LDA", labels=labels_train)
...
 #![extract_weights]
-RealVector w = lda.get_real_vector("w")
+RealVector weights = lda.get_real_vector("w")
 #![extract_weights]
 
 #![evaluate_accuracy]
-AccuracyMeasure eval()
+Evaluation eval = evaluation("AccuracyMeasure")
 real accuracy = eval.evaluate(labels_predict, labels_test)
 #![evaluate_accuracy]
```
karlnapf: example refactoring is OK!

karlnapf: Why did the integration test data change?

karlnapf: ah the order!
OK!

src/shogun/classifier/LDA.cpp
karlnapf: i think gamma should be model selectable

src/shogun/classifier/LDA.cpp
karlnapf: formatting pls

src/shogun/classifier/LDA.cpp
```diff
@@ -117,13 +105,15 @@ template <typename ST>
...
+  REQUIRE(labels->get_num_classes()==2, "Number of classes exceeds 2 for binary classifier: %d provided\n", labels->get_num_classes())
```
karlnapf: formatting.
Also please make this
"Number of classes (%d) must be 2"

src/shogun/classifier/LDA.cpp
```diff
@@ -150,12 +140,14 @@ template <typename ST>
 bool CLDA::solver_classic()
 {
 	auto dense_feat = static_cast<CDenseFeatures<ST>*>(features);
-	index_t num_feat = dense_feat->get_num_features();
+  auto labels = multiclass_labels(m_labels);
+  REQUIRE(labels->get_num_classes()==2, "Number of classes exceeds 2 for binary classifier: %d provided\n", labels->get_num_classes())
```
karlnapf: cant this check be pulled out and only done once?

tests/unit/classifier/LDA_unittest.cc
karlnapf: +1

tests/unit/classifier/LDA_unittest.cc
```diff
@@ -233,6 +203,32 @@ TEST(LDA, DISABLED_CheckProjection_SVD)
...
+
+	generate_test_data<float64_t>(lab, feat);
+	labels = new CDenseLabels();
+	lab[0]=2;
```
karlnapf: this is very indirect and makes the code hard to read. Why not create some really really simple test data to make the point with the exception?

tests/unit/classifier/LDA_unittest.cc
```diff
@@ -233,6 +203,32 @@ TEST(LDA, DISABLED_CheckProjection_SVD)
...
+	labels->set_labels(lab);
+	features = new CDenseFeatures<float64_t>(feat);
+
+	SG_REF(labels);
+	SG_REF(features);
+
+	CLDA lda(0, features, labels, SVD_LDA);
+
+	EXPECT_THROW(lda.train(), ShogunException);
```
karlnapf: maybe put a comment on what is expected to fail here

tests/unit/classifier/LDA_unittest.cc
```diff
@@ -233,6 +203,32 @@ TEST(LDA, DISABLED_CheckProjection_SVD)
...
+
+	SG_UNREF(features);
```
karlnapf: cant you use `auto` everywhere?

src/shogun/classifier/LDA.cpp
```diff
@@ -117,13 +105,15 @@ template <typename ST>
...
 
 	// keep just one dimension to do binary classification
 	const index_t projection_dim = 1;
 	auto solver = std::unique_ptr<LDACanVarSolver<ST>>(
 	    new LDACanVarSolver<ST>(
 	        dense_feat,
-	        new CMulticlassLabels(static_cast<CBinaryLabels*>(m_labels)),
+	        labels,
```
karlnapf: yep that is better, only one conversion don here

src/shogun/classifier/LDA.cpp
shubham808: i am not sure how.... I tried doing it at LDASolver but that messed up FisherLDA so i guess its better to do it here.

tests/unit/classifier/LDA_unittest.cc
shubham808: I have simplified it a bit more now :+1:

shubham808:
#4274

shubham808:
sorry stupid typo (meant to go with 4247)

karlnapf:
Travis error is unrelated (and fixed in HEAD)

karlnapf:
wanna move on by porting a legacy python example to meta example?

PR #4239, opened 2018-04-11
Support clone of SGObjects' arrays

lisitsyn:


src/shogun/lib/any.h
```diff
@@ -327,18 +327,40 @@ namespace shogun
...
+		{
+			std::copy(begin, end, dst);
+		}
+
+		template <class T>
+		inline auto copy_array(T** begin, T** end, T** dst)
+		    -> decltype((void)begin[0]->clone())
+		{
+			std::transform(
+			    begin, end, dst, [](auto each) { return each->clone(); });
```
karlnapf: What if some elements of the array are null?

lisitsyn: Good catch

src/shogun/lib/any.cpp
```diff
@@ -40,6 +40,20 @@ namespace shogun
 	namespace any_detail
 	{
 
+		inline CSGObject* copy_object(CSGObject* object)
```
karlnapf: maybe `SG_FORCED_INLINE`?

karlnapf:
Fine to merge from my side occe CI is happy

karlnapf:
Seems CI cannot be happy as this invalidates the cache ...

src/shogun/lib/any.h
```diff
@@ -327,18 +327,34 @@ namespace shogun
...
 			SG_FREE(ptr);
 		}
+
+		template <class T>
+		inline void copy_array(T* begin, T* end, T* dst)
+		{
+			std::copy(begin, end, dst);
+		}
+
+		void copy_array(CSGObject** begin, CSGObject** end, CSGObject** dst);
```
karlnapf: There is a problem here:
This will only match if the pointer type is `CSGObject` but not for any subclass (discovered this in my own tests when writing clone). See below how to reproduce
The thing is: we decided to register the base classes directly, e.g. `CKernel`, without casting them to `CSGObject`
I think we will have to use the `decltype` trick you did for cloning single objects ....

tests/unit/lib/Any_unittest.cc
```diff
@@ -507,6 +512,60 @@ TEST(Any, clone_array2d)
 	delete[] src;
 }
 
+TEST(Any, clone_array_sgobject)
+{
+	int src_len = 3;
+	CSGObject** src = new CSGObject*[src_len];
```
karlnapf: if you make a test where the pointer type is `Object**`, this will go into the `std::copy` branch above

PR #4252, opened 2018-04-14
Minor fix of knn's parameter naming for consistency.

iglesias:
Quite minor. Also opening to test the change in Travis.

### DIFF
--- a/examples/meta/src/base_api/put_get.sg
+++ b/examples/meta/src/base_api/put_get.sg
@@ -1,5 +1,5 @@
 KNN knn()
-knn.put("m_k", 2)
+knn.put("k", 2)
 
 Kernel k = kernel("GaussianKernel")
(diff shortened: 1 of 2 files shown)


examples/meta/src/base_api/put_get.sg
```diff
@@ -1,5 +1,5 @@
 KNN knn()
-knn.put("m_k", 2)
+knn.put("k", 2)
```
karlnapf: this will break integration tests, so you will need to update the test data file.
See the developing.md readme for details

karlnapf: If it doesnt break the integration tests, then no change is needed ...

karlnapf: actually nevermind. as long as you change all the example code listings, this is not a problem

PR #4254, opened 2018-04-15
Be more explicit in Some

lisitsyn:


src/shogun/kernel/CombinedKernel.cpp
```diff
@@ -907,6 +907,11 @@ CCombinedKernel* CCombinedKernel::obtain_from_generic(CKernel* kernel)
 	return (CCombinedKernel*)kernel;
 }
 
+CCombinedKernel* CCombinedKernel::obtain_from_generic(Some<CSGObject> object)
```
lisitsyn: @karlnapf I hacked it this way

PR #4264, opened 2018-04-22
Some-ize Gaussian Mixture model UT. Closes #4241 (luckily).

### LINKED ISSUE #4241: unit-MixtureModel is flaky
fails every now and then....

iglesias:
The fix was a bit unexpected. The test failed often with *segfault* locally. Valgrind showed it was at unref.

After applying the fix, tested locally that valgrind is happy and the test passed always after running a bunch of times.

The test is a bit bad nonetheless since it relies on magic numbers. But hopefully this will turn it green.

### DIFF
--- a/tests/unit/distribution/MixtureModel_unittest.cc
+++ b/tests/unit/distribution/MixtureModel_unittest.cc
@@ -28,10 +28,11 @@
  * either expressed or implied, of the Shogun Development Team.
  */
 
-#include <shogun/lib/SGVector.h>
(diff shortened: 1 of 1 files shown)


vigsterkr:
formatting issues :)
https://travis-ci.org/shogun-toolbox/shogun/jobs/369770930#L736

iglesias:
Argh the formatting again, didn't even think about this time since it was just editing. Anyways, I will set up a commit hook locally.

PR #4270, opened 2018-05-01
implement CSGObject::add method to append to object parameter arrays

karlnapf:


lisitsyn:
@karlnapf yeah that's exactly what I meant. A bit hacky but at least we can move on.

src/shogun/base/SGObject.cpp
```diff
@@ -1060,6 +1060,49 @@ void CSGObject::put(const std::string& name, CSGObject* value)
...
+			    "is not instantiated.\n",
+			    value->get_name(), get_name(), name.c_str(),
+			    self->map[BaseTag(name)].get_value().type().c_str());
+		}
+
+		array->push_back(value);
+		return;
+	}
+
+	Tag<std::vector<CKernel*>> tag_array_kernel(name);
```
vigsterkr: and what happens if it's std::vector<FOO>? i mean the idea here is that you dispatch into whatever we support?

karlnapf: Yes exactly, this is inherited from the design of put.

vigsterkr: for the record: i dont like it... and its even worse than the put :)

vigsterkr: these type of dynamic dispatchers, along with the ones that are actually being generated by swig should go to the swig interface itself. i mean in some way or other we are just re-implementing here things *manually* that are actually being automagically generated by SWIG - see dynamic dispatching depending on function args.

PR #4271, opened 2018-05-02
Feature/port mkl examples

karlnapf:
@lambday @lisitsyn here is `add` in action.

### DIFF
--- a/examples/meta/src/binary/multiple_kernel_learning.sg
+++ b/examples/meta/src/binary/multiple_kernel_learning.sg
@@ -1,40 +1,38 @@
-CSVFile f_feats_train("../../data/classifier_binary_2d_nonlinear_features_train.dat")
-CSVFile f_feats_test("../../data/classifier_binary_2d_nonlinear_features_test.dat")
-CSVFile f_labels_train("../../data/classifier_binary_2d_nonlinear_labels_train.dat")
-CSVFile f_labels_test("../../data/classifier_binary_2d_nonlinear_labels_test.dat")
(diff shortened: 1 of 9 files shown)


examples/meta/src/binary/multiple_kernel_learning.sg
```diff
@@ -1,49 +1,47 @@
...
+Labels labels_train = labels(f_labels_train)
+Labels labels_test = labels(f_labels_test)
 #![create_features]
 
 #![create_kernel]
-PolyKernel poly_kernel(10,2)
-GaussianKernel gauss_kernel_1(2.0)
-GaussianKernel gauss_kernel_2(3.0)
+Kernel poly_kernel = kernel("PolyKernel", degree=2)
+Kernel gauss_kernel_1 = kernel("GaussianKernel", log_width=0.0)
```
karlnapf: due to the changed gaussian kernel parameters, i needed to re-generate the integration testing data.

src/shogun/classifier/mkl/MKL.cpp
```diff
@@ -271,7 +271,7 @@ void CMKL::register_params()
 	rho = 0;
 	lp_initialized = false;
 
-	SG_ADD((CSGObject**) &svm, "svm", "wrapper svm", MS_NOT_AVAILABLE);
+	SG_ADD((CMachine**)&svm, "svm", "wrapper svm", MS_NOT_AVAILABLE);
```
karlnapf: This is done as we want to register members by base type.
Here is an issue with this: users can now set **any** machine type to mkl.
This will lead to a runtime error, when the machine is used as an svm

PR #4276, opened 2018-05-04
No more dynamic dispatching in CSGObject::put

karlnapf:


src/shogun/base/SGObject.h
```diff
@@ -39,6 +39,30 @@ class Parameter;
 class CSerializableFile;
 class ParameterObserverInterface;
 
+// all shogun base classes for put/add templates
```
vigsterkr: great but i recon since here is everything a fwd declaration as SGObject.h is already quite cramped with stuff maybe it'd be better to factor this out somewhere else and just include that header in this header

src/shogun/base/SGObject.h
```diff
@@ -370,13 +394,19 @@ class CSGObject
...
+	/** Typed setter for an object class parameter of a Shogun base class type,
+	 * identified by a name.
 	 *
 	 * @param name name of the parameter
 	 * @param value value of the parameter
 	 */
-	void put(const std::string& name, CSGObject* value);
+	template <class T,
+		      class X = typename std::enable_if<is_sg_base<T>::value>::type,
+		      class Z = void>
```
vigsterkr: why the `Z`?

src/shogun/base/SGObject.h
```diff
@@ -39,6 +39,30 @@ class Parameter;
...
+class CKernel;
+class CDistance;
+class CFeatures;
+class CLabels;
+class CECOCEncoder;
+class CECOCDecoder;
+class CMulticlassStrategy;
+
+template <class T>
+struct is_sg_base
```
vigsterkr: i wonder whether actually we could generate from a constexpr or something as for a while i imagine that this template will be changed....

src/shogun/base/SGObject.h
karlnapf: because SWIG confuses it with the double templated put (for non-objects) and fails. 
`
%template(put) CSGObject::put<CMachine, CMachine, void>;
`

src/shogun/base/SGObject.h
karlnapf: this will be changed definitely.
I dont know exactly what you have in mind there. Have an example?

lisitsyn: You might specialize is_sg_base for specific `T`.

src/interfaces/swig/shogun.i
```diff
@@ -239,4 +239,18 @@ namespace shogun
 %template(get_int_vector) CSGObject::get_vector_as_matrix_dispatcher<SGMatrix<int32_t>, int32_t>;
 #endif // SWIGJAVA
 
+%define PUT_ADD(sg_class)
+%template(put) CSGObject::put<sg_class, sg_class, void>;
```
karlnapf: need to have this as put/add are now templates.
Also explains why I need 3 template parameters for put (there already is one with 2)

src/shogun/base/SGObject.cpp
```diff
@@ -1030,79 +1031,6 @@ CSGObject* CSGObject::create_empty() const
...
-			    "is not instantiated.\n",
-			    value->get_name(), get_name(), name.c_str(),
-			    self->map[BaseTag(name)].get_value().type().c_str());
-		}
-
-		array->push_back(value);
-		return;
-	}
-
-	Tag<std::vector<CKernel*>> tag_array_kernel(name);
```
karlnapf: this doesnt happen anymore now as we have the type information

src/shogun/base/SGObject.cpp
```diff
@@ -1121,3 +1049,9 @@ CSGObject* CSGObject::get(const std::string& name)
 	    self->map[BaseTag(name)].get_value().type().c_str());
 	return nullptr;
 }
+
+void CSGObject::push_back(CDynamicObjectArray* array, CSGObject* value)
```
karlnapf: need this as cannot include the dynamic object array header in the SGObject.h

PR #4277, opened 2018-05-08
revome methods from labels factory

shubham808:
[WIP] its a start towards eventually deleting CLabelsFactory.

### DIFF
--- a/src/shogun/labels/LabelsFactory.h
+++ b/src/shogun/labels/LabelsFactory.h
@@ -29,29 +29,6 @@ namespace shogun
 class CLabelsFactory : public CSGObject
 {
 public:
-	/** specialize a base class instance to CBinaryLabels
-	 *
-	 * @param base_labels its dynamic type must be CBinaryLabels
-	 */
-	static CBinaryLabels* to_binary(CLabels* base_labels);
-
-	/** specialize a base class instance to CLatentLabels
-	 *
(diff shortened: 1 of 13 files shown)


src/shogun/labels/LabelsFactory.h
```diff
@@ -29,29 +29,6 @@ namespace shogun
 class CLabelsFactory : public CSGObject
 {
 public:
-	/** specialize a base class instance to CBinaryLabels
```
karlnapf: i guess you can just delete the file?

shubham808:
@karlnapf see [here](https://gist.github.com/shubham808/864261eed299b5b1a8fbcb89d6d260e9) there is something wrong.
However the KNN unit test works fine.. thats a good sign :)

src/shogun/labels/LabelsFactory.h
shubham808: yes i am working on it

karlnapf:
You will need to run this before and after the changes, see if there is a difference.
If there is, pls identify the change that causes the difference and point me to it. Then I can help

shubham808:
I dont think it is related to my changes (StanPerceptron error)

shubham808:
@karlnapf same thing happened with valgrind on develop too so it is probably not related to this pr

karlnapf:
Cool then!

PR #4278, opened 2018-05-09
distance meta examples

shubham808:
Some very easy meta examples... I have also deleted canaberra.py because canaberra.sg already exists.


karlnapf:
Thanks!

PR #4280, opened 2018-05-10
CStoppableSGObject class

shubham808:
[WIP] This simply takes all premature stopping code and places it in a new class. Most of the python legacy examples broke locally. I am struggling with the swig part.

### DIFF
--- a/src/shogun/machine/Machine.cpp
+++ b/src/shogun/machine/Machine.cpp
@@ -14,9 +14,10 @@
 using namespace shogun;
 
 CMachine::CMachine()
-    : CSGObject(), m_max_train_time(0), m_labels(NULL), m_solver_type(ST_AUTO),
-      m_cancel_computation(false), m_pause_computation_flag(false)
+    : CStoppableSGObject(), m_max_train_time(0), m_labels(NULL),
+      m_solver_type(ST_AUTO)
 {
+	init_stoppable();
 	m_data_locked=false;
 	m_store_model_features=false;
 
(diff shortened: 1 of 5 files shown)


src/shogun/machine/Machine.cpp
```diff
@@ -14,9 +14,9 @@
 using namespace shogun;
 
 CMachine::CMachine()
-    : CSGObject(), m_max_train_time(0), m_labels(NULL), m_solver_type(ST_AUTO),
-      m_cancel_computation(false), m_pause_computation_flag(false)
+    : CStoppableSGObject(), m_max_train_time(0), m_labels(NULL), m_solver_type(ST_AUTO)
 {
+  init_stoppable();
```
vigsterkr: tab?

shubham808:
@vigsterkr see CMachineEvaluation and CMachine. There is some duplicate code. (also CMachineEvaluation can be a non machine use case too)

geektoni:
Great! :+1: I've added some comments, but it seems okay already. I guess that for fix the errors on Travis you should add this new file "StoppableSGObject.h" to SWIG (since here [https://travis-ci.org/shogun-toolbox/shogun/jobs/377448506#L1715](https://travis-ci.org/shogun-toolbox/shogun/jobs/377448506#L1715) CStoppableSGObject is ignored). Please have a look at [SGBase.i](https://github.com/shogun-toolbox/shogun/blob/develop/src/interfaces/swig/SGBase.i) to see how it's done for other base classes

src/shogun/machine/Machine.h
```diff
@@ -147,7 +142,7 @@ enum EProblemType
  * locking.
  *
  */
-class CMachine : public CSGObject
+class CMachine : public CStoppableSGObject
```
karlnapf: @vigsterkr @lisitsyn can we start using multiple inheritance in the c++ framework? Or is there still a blocker?

karlnapf:
Nice one!

PR #4290, opened 2018-05-18
Add mpark variant to dependency as alternative of cxx17 variant

vinx13:


cmake/external/variant.cmake
```diff
@@ -0,0 +1,17 @@
...
+SET(VARIANT_SOURCE_DIR ${THIRD_PARTY_DIR}/variant)
+SET(VARIANT_INCLUDE_DIR ${THIRD_PARTY_INCLUDE_DIR}/variant)
+include(ExternalProject)
+ExternalProject_Add(
+		variant
+		PREFIX ${CMAKE_BINARY_DIR}/variant
+		SOURCE_DIR ${VARIANT_SOURCE_DIR}
+		GIT_REPOSITORY https://github.com/mpark/variant.git
+		GIT_SHALLOW 1
+		GIT_TAG single-header
```
vigsterkr: this should be ${VARIANT_RELEASE_VERSION} no?

vinx13: this is a branch for release, each version is a single directory

vigsterkr: yeah but they have tags as well, see https://github.com/mpark/variant/tree/v1.3.0

vigsterkr: because unfortunately the branch HEAD will always contain the latest release, meaning you will always checkout the latest release which probably not going to be in future `VARIANT_RELEASE_VERSION` and it's a good idea in general to depend on a *specific* version of an external lib.

vinx13: install with provided cmakelist requires a higher version cmake than ours
https://github.com/mpark/variant/blob/29319715a1f0eb0980d380db8a2fda5af8d58feb/CMakeLists.txt#L8

vigsterkr:
@vinx13 i'll ping you when you should rebase this branch over the latest develop so we can get an actual good picture of the CI status... sorry it is my fault :(

vigsterkr:
@vinx13 could you please rebase your branch over latest develop and push it so that this PR gets properly tested. thnx!

PR #4269, opened 2018-05-01
use features factory in notebooks 

vinx13:
[WIP]

PR #4292, opened 2018-05-20
Update notebooks with factory and getter / setter

vinx13:
continue from #4269 , update more Jupyter notebooks with feature factory, tag getter / setter


karlnapf:
If nobody objects, Ill merge this

PR #4293, opened 2018-05-21
Improve cancel_computation to enable testing.

geektoni:
@karlnapf @shubham808 

This PR contains the first implementation of the new `cancel_computation` method. It is still a WIP so any comments on it are appreciated. Please see the unit test to see the new behaviour.

### DIFF
--- a/src/shogun/lib/StoppableSGObject.cpp
+++ b/src/shogun/lib/StoppableSGObject.cpp
@@ -16,15 +16,15 @@ using namespace shogun;
 		continue;                                                              \
 	pause_computation();
 
-CStoppableSGObject::CStoppableSGObject() : CSGObject(){};
-
-CStoppableSGObject::~CStoppableSGObject(){};
-
-void CStoppableSGObject::init_stoppable()
+CStoppableSGObject::CStoppableSGObject() : CSGObject()
 {
 	m_cancel_computation = false;
 	m_pause_computation_flag = false;
-}
+
+	m_callback = nullptr;
+};
+
+CStoppableSGObject::~CStoppableSGObject(){};
 
 rxcpp::subscription CStoppableSGObject::connect_to_signal_handler()
 {
@@ -39,6 +39,12 @@ rxcpp::subscription CStoppableSGObject::connect_to_signal_handler()
 	    [this]() { this->on_complete(); });
 	return get_global_signal()->get_observable()->subscribe(subscriber);
 }
+
+void CStoppableSGObject::add_callback(std::function<bool()> callback=nullptr)
+{
+	m_callback = callback;
+}
+
 
 void CStoppableSGObject::reset_computation_variables()
 {
--- a/src/shogun/lib/StoppableSGObject.cpp
+++ b/src/shogun/lib/StoppableSGObject.cpp
@@ -40,7 +40,7 @@ rxcpp::subscription CStoppableSGObject::connect_to_signal_handler()
 	return get_global_signal()->get_observable()->subscribe(subscriber);
 }
 
-void CStoppableSGObject::add_callback(std::function<bool()> callback=nullptr)
+void CStoppableSGObject::set_callback(std::function<bool()> callback=nullptr)
 {
 	m_callback = callback;
 }
--- a/src/shogun/lib/StoppableSGObject.cpp
+++ b/src/shogun/lib/StoppableSGObject.cpp
@@ -40,11 +40,10 @@ rxcpp::subscription CStoppableSGObject::connect_to_signal_handler()
 	return get_global_signal()->get_observable()->subscribe(subscriber);
 }
 
-void CStoppableSGObject::set_callback(std::function<bool()> callback=nullptr)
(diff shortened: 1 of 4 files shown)


tests/unit/lib/StoppableSGObject_unittest.cc
```diff
@@ -0,0 +1,96 @@
...
+
+
+TEST(StoppableSGObject, custom_callback)
+{
+	Mock_model a;
+	a.train();
+	EXPECT_TRUE(a.get_check() == 5);
+}
+
+TEST(StoppableSGObject, custom_callback_by_user)
```
geektoni: This test shows more clearly what we are going to do in the future when testing the premature stopping feature.

src/shogun/lib/StoppableSGObject.cpp
```diff
@@ -40,6 +40,12 @@ rxcpp::subscription CStoppableSGObject::connect_to_signal_handler()
 	return get_global_signal()->get_observable()->subscribe(subscriber);
 }
 
+void CStoppableSGObject::add_callback(std::function<bool()> callback=nullptr)
```
karlnapf: add suggests to add them to a stack of callbacks

tests/unit/lib/StoppableSGObject_unittest.cc
```diff
@@ -0,0 +1,96 @@
...
+		if (i>=3) {
+			get_global_signal()->get_subscriber()->on_next(SG_BLOCK_COMP);
+			return true;
+		}
+		i++;
+		return false;
+	};
+
+
+	Mock_model a;
```
karlnapf: So how would this be tested with existing machines?

tests/unit/lib/StoppableSGObject_unittest.cc
```diff
@@ -0,0 +1,96 @@
...
+	virtual bool train_machine(CFeatures* data=NULL)
+	{
+		for (int k=0; k<10; k++)
+		{
+			COMPUTATION_CONTROLLERS
+			m_check++;
+		}
+		return true;
+	}
+	int m_check;
```
karlnapf: maybe those two can be renamed to something more descriptive

tests/unit/lib/StoppableSGObject_unittest.cc
```diff
@@ -0,0 +1,96 @@
...
+			if (m_i >= 5)
+			{
+				get_global_signal()->get_subscriber()->on_next(SG_BLOCK_COMP);
+				return true;
+			}
+			m_i++;
+			return false;
+		};
+
+		// We then add the callback
```
karlnapf: superflous comment :)

tests/unit/lib/StoppableSGObject_unittest.cc
```diff
@@ -0,0 +1,96 @@
...
+#include <functional>
+#include <gtest/gtest.h>
+#include <shogun/machine/Machine.h>
+#include <shogun/lib/Signal.h>
+#include <rxcpp/rx-lite.hpp>
+
+using namespace shogun;
+using namespace std;
+
+class Mock_model : public CMachine
```
karlnapf: `MockModel`

tests/unit/lib/StoppableSGObject_unittest.cc
karlnapf: Ok I guess your idea here is that we only test things for the Mock_model once and not again for all the subsequent machines?

src/shogun/lib/StoppableSGObject.h
```diff
@@ -32,14 +32,16 @@ namespace shogun
...
 		/** @return whether the algorithm needs to be stopped */
 		SG_FORCED_INLINE bool cancel_computation() const
 		{
-			return m_cancel_computation.load();
+			/* Execute the callback, if present*/
+			bool result_callback = false;
+			if (m_callback)
+				result_callback = m_callback();
+
+			return m_cancel_computation.load() || result_callback;
```
vigsterkr: since it's `||` maybe it'd be better to have this as:
```
return (m_callback) ? (m_cancel_computation.load() || result_callback()) : m_cancel_computation.load();
```

just for the sake of lazy evaluation of or :)

tests/unit/lib/StoppableSGObject_unittest.cc
```diff
@@ -0,0 +1,109 @@
...
+
+	/* Addition control variable that is incremented each time by the callback.*/
+	int m_last_iteration;
+};
+
+
+TEST(StoppableSGObject, empty_callback)
+{
+	MockModel a;
+	a.set_callback(nullptr);
```
vigsterkr: you either use the default (nullptr) valued `set_callback`, i.e. use

```
a.set_callback();
```

or remove the default value of the function arg. i would opt for the latter

geektoni:
@karlnapf regarding the testing of the machine, I was thinking about using the same approach used here  #3537 or typed tests of the Google suite. This way we could test their consistency after being stopped.

PR #4294, opened 2018-05-22
Cost functions now support Stan Math, Kept the previous classes for backward compatability. 

FaroukY:
We can now write any arbitrary cost functions in Stan Math, and get the gradients with respect to all variables, regardless how complex the cost function is. This PR has the class FirstOrderSAGCostFunctionInterface which acts as an interface for building Stochastic Average Cost functions. I've also added a unit test to show how to use the class. All reference values where validated on Tensorflow.


src/shogun/optimization/FirstOrderSAGCostFunctionInterface.cpp
```diff
@@ -0,0 +1,178 @@
+/*
```
vigsterkr: you should use the shorter version of the license, see for example SGObject.h

src/shogun/optimization/FirstOrderSAGCostFunctionInterface.h
```diff
@@ -0,0 +1,173 @@
+/*
```
vigsterkr: should use the shorter version of the license

src/shogun/optimization/FirstOrderSAGCostFunctionInterface.h
```diff
@@ -0,0 +1,173 @@
...
+
+#include <stan/math.hpp>
+#include <Eigen/Dense>
+#include <shogun/lib/config.h>
+#include <shogun/optimization/FirstOrderSAGCostFunction.h>
+#include <functional>
+#include <vector>
+
+
+namespace shogun{
```
vigsterkr: whitespace

src/shogun/optimization/FirstOrderSAGCostFunctionInterface.h
```diff
@@ -0,0 +1,173 @@
...
+ * of the authors and should not be interpreted as representing official policies,
+ * either expressed or implied, of the Shogun Development Team.
+ *
+ */
+
+#ifndef FIRSTORDERSAGCOSTFUNCTIONINTERFACE_H
+#define FIRSTORDERSAGCOSTFUNCTIONINTERFACE_H
+
+#include <stan/math.hpp>
+#include <Eigen/Dense>
```
vigsterkr: we never include eigen directly in shogun headers, only in linalg but even then it's always via shogun/mathematics/eigen3.h

src/shogun/optimization/FirstOrderSAGCostFunctionInterface.h
```diff
@@ -0,0 +1,173 @@
...
+   *
+   * The cost function must be Written as a finite sample-specific sum of cost.
+   * For example, least squares cost function,
+   * \f[
+   * f(w)=\frac{ \sum_i{ (y_i-w^T x_i)^2 } }{2}
+   * \f]
+   * where \f$(y_i,x_i)\f$ is the i-th sample,
+   * \f$y_i\f$ is the label and \f$x_i\f$ is the features
+   */
+  class FirstOrderSAGCostFunctionInterface : public FirstOrderSAGCostFunction{
```
vigsterkr: i guess this class is never gonna be part of an SGObject, hence we could make a pass over all the anti-shogun-patterns that you've used below

src/shogun/optimization/FirstOrderSAGCostFunctionInterface.h
```diff
@@ -0,0 +1,173 @@
...
+  public:
+    /** Consturctor that initializes X, y, trainable_parameters, and the cost_function */
+  	FirstOrderSAGCostFunctionInterface(
+  		Eigen::Matrix<float64_t, Eigen::Dynamic, Eigen::Dynamic>* X,
```
vigsterkr: why not SGVector<float64_t>?

src/shogun/optimization/FirstOrderSAGCostFunctionInterface.h
```diff
@@ -0,0 +1,173 @@
...
+  		Eigen::Matrix<float64_t, 1, Eigen::Dynamic>* y,
```
vigsterkr: why not SGVector<float64_t>?

src/shogun/optimization/FirstOrderSAGCostFunctionInterface.h
```diff
@@ -0,0 +1,173 @@
...
+      std::function < stan::math::var(std::vector<stan::math::var>*) >* total_cost
+  	);
+
+  	/** Default constructor, use setter helpers to set X, y, trainable_parameters, and
+  	*		cost_function
+  	*/
+  	FirstOrderSAGCostFunctionInterface() {};
+
+  	/** Setter for the training data X */
+  	virtual void set_training_data(
```
vigsterkr: these args coudl be SGVectors

src/shogun/optimization/FirstOrderSAGCostFunctionInterface.h
```diff
@@ -0,0 +1,173 @@
...
+  	);
+
+  	/** Setter for the cost function definition using stan */
+  	virtual void set_ith_cost_function(std::vector< std::function < stan::math::var(int32_t) > >* new_cost_f);
+
+    /** Setter for the overall cost function */
+    virtual void set_cost_function(std::function < stan::math::var(std::vector<stan::math::var>*) >* total_cost);
+
+
+  	~FirstOrderSAGCostFunctionInterface();
```
vigsterkr: override/virtual?

src/shogun/optimization/FirstOrderSAGCostFunctionInterface.h
```diff
@@ -0,0 +1,173 @@
...
+
+  	/** Get the cost given current target variables
+  	 *
+  	 * For least squares, that is the value of \f$f(w)\f$.
+  	 *
+  	 * @return cost
+  	 */
+  	virtual float64_t get_cost();
+
+    /** Get the sample size
```
vigsterkr: indenting.... plz use clang-format

src/shogun/optimization/FirstOrderSAGCostFunctionInterface.h
```diff
@@ -0,0 +1,173 @@
...
+     *
+     * @return the sample size
+     */
+    virtual int32_t get_sample_size();
```
vigsterkr: index_t

src/shogun/optimization/FirstOrderSAGCostFunctionInterface.h
```diff
@@ -0,0 +1,173 @@
...
+  	Eigen::Matrix<float64_t, Eigen::Dynamic, Eigen::Dynamic>* m_X;
+
+  	/** y is the ground truth, or the correct prediction */
+  	Eigen::Matrix<float64_t, 1, Eigen::Dynamic>* m_y;
+
+  	/** trainable_parameters are the variables that are optimized for */
+  	Eigen::Matrix<stan::math::var, Eigen::Dynamic, 1>* m_trainable_parameters;
+
+    /** cost_for_ith_point is the cost contributed by each point in the training data */
+    std::vector< std::function < stan::math::var(int32_t) > >* m_cost_for_ith_point;
```
vigsterkr: indent

tests/unit/optimization/FirstOrderSAGCostFunctionInterface_unittest.cc
```diff
@@ -0,0 +1,184 @@
+/*
+* Copyright (c) The Shogun Machine Learning Toolbox
```
vigsterkr: use the shorter license

src/shogun/optimization/FirstOrderSAGCostFunctionInterface.h
```diff
@@ -0,0 +1,173 @@
...
+     * where \f$f(w)=\frac{ \sum_i^n{ (y_i-w^t x_i)^2 } }{2}\f$
+     *
+     * @return average gradient of target variables
+     */
+    virtual SGVector<float64_t> get_average_gradient();
+
+
+  protected:
+  	/** X is the training data in column major matrix format */
+  	Eigen::Matrix<float64_t, Eigen::Dynamic, Eigen::Dynamic>* m_X;
```
vigsterkr: you have not added these to the param framework, nor this will be ever be able to be added... you need to use SGVector for that

src/shogun/optimization/FirstOrderSAGCostFunctionInterface.h
```diff
@@ -0,0 +1,173 @@
...
+
+  	/** y is the ground truth, or the correct prediction */
+  	Eigen::Matrix<float64_t, 1, Eigen::Dynamic>* m_y;
```
vigsterkr: same as above... SGVector

src/shogun/optimization/FirstOrderSAGCostFunctionInterface.h
```diff
@@ -0,0 +1,173 @@
...
+
+  	/** trainable_parameters are the variables that are optimized for */
+  	Eigen::Matrix<stan::math::var, Eigen::Dynamic, 1>* m_trainable_parameters;
```
vigsterkr: this currently is foobar... you wont be able to serialize reproducible this class

src/shogun/optimization/FirstOrderSAGCostFunctionInterface.h
```diff
@@ -0,0 +1,173 @@
...
+
+    /** cost_for_ith_point is the cost contributed by each point in the training data */
+    std::vector< std::function < stan::math::var(int32_t) > >* m_cost_for_ith_point;
+
+    /** total_cost is the total cost to be minimized, that in this case is a form of sum of cost_for_ith_point*/
+    std::function < stan::math::var(std::vector<stan::math::var>*) >* m_total_cost;
```
vigsterkr: indent + problems with serialization

src/shogun/optimization/FirstOrderSAGCostFunctionInterface.h
vigsterkr: SG_ADD is not used anywhere in this class...

vigsterkr:
plz run clang-format on your code prior pushing into a PR

src/shogun/optimization/FirstOrderSAGCostFunction.h
```diff
@@ -96,23 +96,23 @@ class FirstOrderSAGCostFunction
...
 	 * \f$\frac{\partial f_i(w) }{\partial w}\f$ given \f$w\f$ is known
-	 * where the index \f$i\f$ is obtained by next_sample() 
+	 * where the index \f$i\f$ is obtained by next_sample()
 	 *
 	 * @return sample gradient of target variables
 	 */
 	virtual SGVector<float64_t> get_gradient()=0;
 
-	/** Get the cost given current target variables 
+	/** Get the cost given current target variables
```
karlnapf: could you adjust your editor so it doesnt cause all this noise?

src/shogun/optimization/FirstOrderSAGCostFunctionInterface.cpp
```diff
@@ -0,0 +1,178 @@
...
+   m_cost_for_ith_point   = cost_for_ith_point;
+   m_total_cost           = total_cost;
+ }
+
+ void FirstOrderSAGCostFunctionInterface::set_training_data(
+   Eigen::Matrix<float64_t, Eigen::Dynamic, Eigen::Dynamic> * X_new,
+   Eigen::Matrix<float64_t, 1, Eigen::Dynamic>* y_new
+ )
+ {
+  REQUIRE(X_new!=NULL, "X must be non NULL");
```
karlnapf: Is this a user facing error?
If no, pls say "No X provided.\n" (also missing newline)

src/shogun/optimization/FirstOrderSAGCostFunctionInterface.cpp
```diff
@@ -0,0 +1,178 @@
...
+   {
+     this->m_trainable_parameters = new_params;
+   }
+ }
+
+ void FirstOrderSAGCostFunctionInterface::set_ith_cost_function(
+   std::vector< std::function < stan::math::var(int32_t) > >* new_cost_f
+ )
+ {
+   REQUIRE(new_cost_f, "The cost function must be a non NULL vector of stan variables");
```
karlnapf: see above, and also for all other error msgs

src/shogun/optimization/FirstOrderSAGCostFunctionInterface.cpp
```diff
@@ -0,0 +1,178 @@
...
+  if (m_index_of_sample>=num_of_samples)return false;
+  ++m_index_of_sample;
+  return true;
+ }
+
+
+SGVector<float64_t> FirstOrderSAGCostFunctionInterface::get_gradient()
+{
+  int32_t num_of_variables = m_trainable_parameters->rows();
+  REQUIRE(num_of_variables > 0, "Number of training parameters must be greater than 0");
```
karlnapf: Pls always print the provided number
"Number of training parameters (%d) must be positive.\n"

src/shogun/optimization/FirstOrderSAGCostFunctionInterface.cpp
```diff
@@ -0,0 +1,178 @@
...
+SGVector<float64_t> FirstOrderSAGCostFunctionInterface::get_average_gradient()
+{
+  int32_t params_num = m_trainable_parameters->rows();
+  SGVector<float64_t> average_gradients(params_num);
+
+  int32_t old_index_sample = m_index_of_sample;
+  int32_t n = get_sample_size();
+  REQUIRE(n>0 , "Number of sample must be greater than 0");
+
+  for(auto i=0; i<n; ++i)
```
karlnapf: minor: `for (auto i : range(n))` for all range based loops

src/shogun/optimization/FirstOrderSAGCostFunctionInterface.cpp
```diff
@@ -0,0 +1,178 @@
...
+  {
+        m_index_of_sample = i;
+        auto grad = get_gradient();
+        for(auto j=0; j<params_num; ++j)
+        {
+          average_gradients[j]+= (grad[j]/n);
```
karlnapf: can you use linalg here pls, rather than those vanilla loops

src/shogun/optimization/FirstOrderSAGCostFunctionInterface.h
```diff
@@ -0,0 +1,173 @@
...
+
+  	/** Get next sample
+  	 *
+  	 * @return false if reach the end of the sample sequence
+  	 * */
+  	virtual bool next_sample();
+
+  	/** Get the SAMPLE gradient value wrt target variables
+  	 *
+  	 * WARNING
```
karlnapf: I dont understand this warning thing. Could you make it more clear?

tests/unit/optimization/FirstOrderSAGCostFunctionInterface_unittest.cc
```diff
@@ -0,0 +1,184 @@
...
+  SGVector<float64_t> ret(params_num);
+  for(auto i=0; i<params_num; ++i)
+  {
+    ret[i] = (*m_trainable_parameters)[i].val();
+  }
+  return ret;
+}
+
+
+TEST(LeastSquareTestCostFunction, ONALINE)
```
karlnapf: whats ONALINE?
Pls try to do slightly more elaborate test names :)

FaroukY: Sorry, meant that the points are on a line (ONALINE)

karlnapf: points_on_a_line

tests/unit/optimization/FirstOrderSAGCostFunctionInterface_unittest.cc
```diff
@@ -0,0 +1,167 @@
...
+	int n = 3;
+	auto X = Matrix<float64_t, Dynamic, Dynamic>();
+	X.resize(2, 3);
+
+	X(0, 0) = 1;
+	X(0, 1) = 1;
+	X(0, 2) = 1;
+	X(1, 0) = 0;
+	X(1, 1) = 1;
+	X(1, 2) = 2;
```
karlnapf: is there a chance of re-using this test data or the general setup? If so it would be good to put it into a function or a fixture. It also makes the test smaller

iglesias: ++

FaroukY:
Alright, addressed all changes.

tests/unit/optimization/FirstOrderSAGCostFunctionInterface_unittest.cc
```diff
@@ -0,0 +1,146 @@
...
+#include "FirstOrderSAGCostFunctionInterface_unittest.h"
+#include <Eigen/Dense>
+
+using namespace shogun;
+using Eigen::Matrix;
+using Eigen::Dynamic;
+using stan::math::var;
+using std::function;
+
+/** This is a temporary fix. The variables are now variables
```
iglesias: What does this mean?

src/shogun/optimization/FirstOrderSAGCostFunctionInterface.h
```diff
@@ -0,0 +1,161 @@
...
+#include <shogun/optimization/FirstOrderSAGCostFunction.h>
+#include <functional>
+#include <vector>
+#include <shogun/lib/SGMatrix.h>
+#include <shogun/lib/SGVector.h>
+
+using Eigen::Dynamic;
+using Eigen::Matrix;
+using stan::math::var;
+using std::function;
```
iglesias: It does not feel good to have type aliases in headers. They also become effective in places where the header is included, no?

src/shogun/optimization/FirstOrderSAGCostFunctionInterface.h
```diff
@@ -0,0 +1,161 @@
...
+
+namespace shogun
+{
+	/** @brief The first order stochastic cost function base class for
+	 * implementing
+	 *  The SAG Cost function
```
iglesias: Style: newline (after implementing) and capitalization (implementing \n The) are odd.

src/shogun/optimization/FirstOrderSAGCostFunctionInterface.h
```diff
@@ -0,0 +1,161 @@
...
+		 * cost_function */
+		FirstOrderSAGCostFunctionInterface(
+		    SGMatrix<float64_t>* X, SGMatrix<float64_t>* y,
+		    Matrix<var, Dynamic, 1>* trainable_parameters,
+		    Matrix<function<var(int32_t)>, Dynamic, 1>* cost_for_ith_point,
+		    function<var(Matrix<var, Dynamic, 1>*)>* total_cost);
+
+		/** Default constructor, use setter helpers to set X, y,
+		*trainable_parameters, and
+		*		cost_function
```
iglesias: This is not informative doc :-) I'd remove very general comments like this one, they add noise.

src/shogun/optimization/FirstOrderSAGCostFunctionInterface.h
```diff
@@ -0,0 +1,161 @@
...
+		 * known
+		 * where \f$f(w)=\frac{ \sum_i^n{ (y_i-w^t x_i)^2 } }{2}\f$
+		 *
+		 * @return average gradient of target variables
+		 */
+		virtual SGVector<float64_t> get_average_gradient();
+
+	protected:
+		/** X is the training data in column major matrix format */
+		SGMatrix<float64_t>* m_X;
```
iglesias: Why are all these members pointers?

src/shogun/optimization/FirstOrderSAGCostFunction.h
```diff
@@ -31,8 +31,8 @@
```
iglesias: As you are already touching this file, it would be nice to do a tiny extra effort and update the Copyright to the more modern shorter version.

src/shogun/optimization/FirstOrderSAGCostFunctionInterface.h
```diff
@@ -0,0 +1,161 @@
...
+	 * The cost function must be Written as a finite sample-specific sum of
+	 * cost.
+	 * For example, least squares cost function,
+	 * \f[
+	 * f(w)=\frac{ \sum_i{ (y_i-w^T x_i)^2 } }{2}
+	 * \f]
+	 * where \f$(y_i,x_i)\f$ is the i-th sample,
+	 * \f$y_i\f$ is the label and \f$x_i\f$ is the features
+	 */
+	class FirstOrderSAGCostFunctionInterface : public FirstOrderSAGCostFunction
```
iglesias: It seems a bit odd that the interface inherits from the class.

iglesias:
I have the feeling that @vigsterkr's first general comment is not addressed. @FaroukY, did you discuss with anybody about it?

tests/unit/optimization/FirstOrderSAGCostFunctionInterface_unittest.cc
vigsterkr: we need to figure out how we can store (if we want a none primitive typed std::vector, like stan::math:var)... hence the temporary story... just add a FIXME/TODO term there and then its good

iglesias:
How when it is needed, I think. We can sync in a few days about what
attributes these cost functions should have, which ones make sense to make
part of the public api.

For instance right now I see that the Interface class even contains the
training data.

src/shogun/optimization/FirstOrderSAGCostFunctionInterface.cpp
```diff
@@ -0,0 +1,149 @@
...
+}
+
+void FirstOrderSAGCostFunctionInterface::begin_sample()
+{
+	m_index_of_sample = 0;
+}
+
+bool FirstOrderSAGCostFunctionInterface::next_sample()
+{
+	auto num_of_samples = get_sample_size();
```
vigsterkr: there's just no need for this variable on the stack... and just extra line of code. just do:

```
if (m_index_of_sample >= get_sample_size())
	return false; 
```

src/shogun/optimization/FirstOrderSAGCostFunctionInterface.cpp
```diff
@@ -0,0 +1,149 @@
...
+	if (m_index_of_sample >= num_of_samples)
+		return false;
+	++m_index_of_sample;
+	return true;
+}
+
+SGVector<float64_t> FirstOrderSAGCostFunctionInterface::get_gradient()
+{
+	int32_t num_of_variables = m_trainable_parameters->rows();
```
vigsterkr: auto

src/shogun/optimization/FirstOrderSAGCostFunctionInterface.cpp
```diff
@@ -0,0 +1,149 @@
...
+
+	for (auto i = 0; i < num_of_variables; ++i)
+		gradients[i] = (*m_trainable_parameters)(i, 0).adj();
+
+	return gradients;
+}
+
+float64_t FirstOrderSAGCostFunctionInterface::get_cost()
+{
+	int32_t n = get_sample_size();
```
vigsterkr: auto

src/shogun/optimization/FirstOrderSAGCostFunctionInterface.cpp
```diff
@@ -0,0 +1,149 @@
...
+}
+
+index_t FirstOrderSAGCostFunctionInterface::get_sample_size()
+{
+	return m_X->num_cols;
+}
+
+SGVector<float64_t> FirstOrderSAGCostFunctionInterface::get_average_gradient()
+{
+	int32_t params_num = m_trainable_parameters->rows();
```
vigsterkr: auto

src/shogun/optimization/FirstOrderSAGCostFunctionInterface.cpp
```diff
@@ -0,0 +1,149 @@
...
+	SGVector<float64_t> average_gradients(params_num);
+
+	int32_t old_index_sample = m_index_of_sample;
```
vigsterkr: auto

src/shogun/optimization/FirstOrderSAGCostFunctionInterface.cpp
```diff
@@ -0,0 +1,149 @@
...
+	int32_t n = get_sample_size();
```
vigsterkr: auto

src/shogun/optimization/FirstOrderSAGCostFunctionInterface.cpp
```diff
@@ -0,0 +1,149 @@
...
+	REQUIRE(
+	    n > 0,
+	    "Number of sample must be greater than 0, you provided no samples");
+
+	for (index_t i = 0; i < n; ++i)
+	{
+		m_index_of_sample = i;
+		auto grad = get_gradient();
```
vigsterkr: do we need this variable?

vigsterkr: just do `average_gradients += get_gradient()`

src/shogun/optimization/FirstOrderSAGCostFunctionInterface.h
vigsterkr: second that this actually contaminates any file that will include this header

src/shogun/optimization/FirstOrderSAGCostFunctionInterface.cpp
```diff
@@ -0,0 +1,149 @@
...
+#include <shogun/optimization/FirstOrderSAGCostFunctionInterface.h>
+#include <shogun/mathematics/Math.h>
+using namespace shogun;
+using stan::math::var;
+using std::function;
+using Eigen::Matrix;
+
+FirstOrderSAGCostFunctionInterface::FirstOrderSAGCostFunctionInterface(
+    SGMatrix<float64_t>* X, SGMatrix<float64_t>* y,
+    Matrix<var, Dynamic, 1>* trainable_parameters,
```
vigsterkr: what's the story behind these being pointers instead of ref?

src/shogun/optimization/FirstOrderSAGCostFunctionInterface.cpp
```diff
@@ -0,0 +1,149 @@
...
+	++m_index_of_sample;
+	return true;
+}
+
+SGVector<float64_t> FirstOrderSAGCostFunctionInterface::get_gradient()
+{
+	int32_t num_of_variables = m_trainable_parameters->rows();
+	REQUIRE(
+	    num_of_variables > 0,
+	    "Number of training parameters must be greater than 0");
```
karlnapf: missing newline. Also pls always print the probided number
"Number of training parameters (%d) must be positive.\n"

src/shogun/optimization/FirstOrderSAGCostFunctionInterface.cpp
```diff
@@ -0,0 +1,149 @@
...
+
+	SGVector<float64_t> gradients(num_of_variables);
+
+	var f_i = (*m_cost_for_ith_point)(m_index_of_sample, 0)(m_index_of_sample);
+
+	stan::math::set_zero_all_adjoints();
+	f_i.grad();
+
+	for (auto i = 0; i < num_of_variables; ++i)
+		gradients[i] = (*m_trainable_parameters)(i, 0).adj();
```
karlnapf: can't this be done without a loop using `std::transform` or something?

src/shogun/optimization/FirstOrderSAGCostFunctionInterface.cpp
```diff
@@ -0,0 +1,149 @@
...
+
+	return gradients;
+}
+
+float64_t FirstOrderSAGCostFunctionInterface::get_cost()
+{
+	int32_t n = get_sample_size();
+	Matrix<var, Dynamic, 1> cost_argument(n);
+	for (auto i = 0; i < n; ++i)
```
karlnapf: as abovek would be better to do this without loop (if easy)
If not, at least do a `for (auto i : range(n))`

src/shogun/optimization/FirstOrderSAGCostFunctionInterface.h
```diff
@@ -0,0 +1,161 @@
...
+	{
+	public:
+		/** Consturctor that initializes X, y, trainable_parameters, and the
+		 * cost_function */
+		FirstOrderSAGCostFunctionInterface(
+		    SGMatrix<float64_t>* X, SGMatrix<float64_t>* y,
+		    Matrix<var, Dynamic, 1>* trainable_parameters,
```
karlnapf: could we maybe alias things like `Matrix<var, Dynamic, 1>` into something more readable?

src/shogun/optimization/FirstOrderSAGCostFunctionInterface.h
```diff
@@ -0,0 +1,161 @@
...
+		 */
+		virtual void begin_sample();
+
+		/** Get next sample
+		 *
+		 * @return false if reach the end of the sample sequence
+		 * */
+		virtual bool next_sample();
+
+		/** Get the SAMPLE gradient value wrt target variables
```
karlnapf: why is this SAMPLE? rather than sample

src/shogun/optimization/FirstOrderSAGCostFunctionInterface.h
```diff
@@ -0,0 +1,161 @@
...
+		 *
+		 * WARNING
```
karlnapf: not sure what the warning is supposed to mean, I commented this before but maybe I missed your reply?

src/shogun/optimization/FirstOrderSAGCostFunctionInterface.h
```diff
@@ -0,0 +1,161 @@
...
+		 */
+		virtual float64_t get_cost();
+
+		/** Get the sample size
+		 *
+		 * @return the sample size
+		 */
+		virtual index_t get_sample_size();
+
+		/** Get the AVERAGE gradient value wrt target variables
```
karlnapf: average

tests/unit/optimization/FirstOrderSAGCostFunctionInterface_unittest.cc
FaroukY: @iglesias  As Viktor said. The original cost function needs a *reference* to an SGVector<float64_t> of the parameters. The parameters have now been changed in the cost function to SGVector<stan::math::var> so we're going to have to change the interface of the minimizers before returning a SGVector<stan::math::var>,

tests/unit/optimization/FirstOrderSAGCostFunctionInterface_unittest.h
```diff
@@ -0,0 +1,35 @@
+/*
+ * This software is distributed under BSD 3-clause license (see LICENSE file).
+ *
+ * Authors: Elfarouk
+ */
+
+#include <shogun/lib/config.h>
+#ifndef FIRSTORDERSAGCOSTFUNCTIONINTERFACE_UNITTEST_H
```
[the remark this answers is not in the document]
FaroukY: Not sure to be honest. I looked at the other files and they all had <one word with no spaces>_UNITTEST_H so I followed the convention. If there is no guideline I'll add more _ between words

src/shogun/optimization/FirstOrderSAGCostFunctionInterface.h
FaroukY: I thought that if you use using std::function; in a header, then you'd need to access it from another file which has the include as FirstOrderSAGCostFunctionInterface::function no? In other words, shouldn't it be under the scope of the .h file?

FaroukY: Actually you're right, I just forgot my c++! Including it in the header file does include it in other files as well. I'll get rid of it.

src/shogun/optimization/FirstOrderSAGCostFunctionInterface.h
FaroukY: Sorry, I ran clang-format on it and didn't check the styling. Will fix the styling of the comments

src/shogun/optimization/FirstOrderSAGCostFunctionInterface.h
```diff
@@ -0,0 +1,161 @@
...
+		/** trainable_parameters are the variables that are optimized for */
+		Matrix<var, Dynamic, 1>* m_trainable_parameters;
+
+		/** cost_for_ith_point is the cost contributed by each point in the
+		 * training data */
+		Matrix<function<var(int32_t)>, Dynamic, 1>* m_cost_for_ith_point;
+
+		/** total_cost is the total cost to be minimized, that in this case is a
+		 * form of sum of cost_for_ith_point*/
+		function<var(Matrix<var, Dynamic, 1>*)>* m_total_cost;
```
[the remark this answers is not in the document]
FaroukY: @iglesias  Not silly at all! I actually thought about this, but I wanted to separate the logic of calculating the cost function from the class so that base class can support any cost function. We can also make a pure virtual method and let the user implement it. Both work.

src/shogun/optimization/FirstOrderSAGCostFunctionInterface.h
FaroukY: @iglesias Just a convention (I try to pass things by reference/pointer instead of value to avoid copying). But I agree I will get rid of SGMatrix's pointers since it already implements that under the hood.

src/shogun/optimization/FirstOrderSAGCostFunctionInterface.h
FaroukY: Hmm, how about FirstOrderSAGArbitraryCostFunction ??

src/shogun/optimization/FirstOrderSAGCostFunctionInterface.cpp
FaroukY: I am not sure what is the convention in Shogun. But I always used pointers in my previous C++ coding (Just to avoid the syntax of reference). If you'd like, I think we can change it to reference here.

src/shogun/optimization/FirstOrderSAGCostFunctionInterface.cpp
FaroukY: Sure, but in this case, I'll just print: "Number of training parameters must be greater than 0, you provided 0".... since you can't have num_of_variables<0

src/shogun/optimization/FirstOrderSAGCostFunctionInterface.cpp
FaroukY: Didn't know this, but Eigen has unaryExpr which allows you to apply any function to all elements in the matrix. So I've removed the vanilla loop and replaced it with a call to that function.

src/shogun/optimization/FirstOrderSAGCostFunctionInterface.cpp
FaroukY: This one is a bid harder to do without a loop. Since we're calling a function inside and passing an argument, Eigen::unaryExpr wouldn't work.

src/shogun/optimization/FirstOrderSAGCostFunctionInterface.h
FaroukY: Just wanted to emphasise it. Changed it to lower case.

src/shogun/optimization/FirstOrderSAGCostFunctionInterface.h
FaroukY: @karlnapf It means we should return the partial derivative of the ith sample with respect to the variables, and not the partial derivative of the total cost with respect with respect to the variables. 

So for example, if f_i(w) = (w^t*X[i]-y[i])^2 is the error of the ith data point, then this function returns the partial derivative of the **current** data sample with respect to w, rather than returning the derivative of Sum_i(f_i(w)) with respect to w.

This function (and get_average_gradient()) is from the base class (FirstOrderSAGCostFunction) and had the warning in its docs. Do you want me to change the warning in both files to something else, and if so, what do you want to clarify?

FaroukY:
Alright, 2nd iteration. Comments are either 1) Addressed or 2) I replied to the comment for some clarification

iglesias:
UNITTEST is good. It is about the FIRSTORDERSAG...

iglesias:
No, the concept is much simpler :-) If you want to use std::function
anywhere (header or implementation file), you just need to include the
functional from the stl.

iglesias:
Making it a member attribute does not separate from the class, right?

iglesias:
That makes sense of course, avoiding unnecessaey copies. But here nothing
is passed, this is just the definition or declaration of the attributes of
a class.

Having these as pointers may make sense. It is of course not always the
wrong choice. Although one should prefer using smart instead of naked
pointers.

Conceptually, whether an attribute should be a pointer, reference, or value
is a question of ownership and life times.

iglesias:
It sounds like you may have been writing mostly C and using a C++ compiler
:-)

Fairly common practice unfortunately. There are more expressive and just
plain better ways of using C++.

iglesias:
Probably not under zero, unless there's a nasty bug corrupting the memory.
I see tht way of being over-careful about writing the error message a form
of defensive programming.

karlnapf:
What is the state here?

FaroukY:
@karlnapf This will be a [WIP] pr. The aim is to provide a very minimal working optimization mini module that works only on stan (Will support 1 minimizer, 1 updater, 1 type of cost functions, and so on. Then other minimizers, updaters, .... will be updated sequentially one by one once we have a "Framework" we agree on. I'll be pushing a new update here very very soon.

src/shogun/optimization/StanFirstOrderSAGCostFunction.cpp
```diff
@@ -0,0 +1,140 @@
...
+	    m_trainable_parameters, m_index_of_sample);
+
+	stan::math::set_zero_all_adjoints();
+	f_i.grad();
+
+	SGVector<float64_t>::EigenVectorXt gradients =
+	    m_trainable_parameters->unaryExpr(
+	        [](stan::math::var x) -> float64_t { return x.adj(); });
+	// clone needed because gradients is local variable
+	return SGVector<float64_t>(gradients).clone();
```
vigsterkr: you could just simply wrap the EigenVectorXt with SGVector and not do cloning...
namely
```
SGVector<float64>(gradients.data(), gradients.size())
```

but i reckon it'd be better that simply just create an
```
SGVector<float64_t> gradients(size);
SGVector<float64_t>::EigenVectorXt mapped_gradients = gradients;
/// now you can use mapped_gradients as an eigen vector

return gradients;
```

src/shogun/optimization/StanFirstOrderSAGCostFunction.cpp
```diff
@@ -0,0 +1,140 @@
...
+#include <shogun/mathematics/Math.h>
+using namespace shogun;
+using stan::math::var;
+using std::function;
+using Eigen::Matrix;
+using Eigen::Dynamic;
+
+StanFirstOrderSAGCostFunction::StanFirstOrderSAGCostFunction(
+    SGMatrix<float64_t> X, SGMatrix<float64_t> y,
+    StanVector* trainable_parameters,
```
vigsterkr: still dont get why we need pointers.

src/shogun/optimization/StanFirstOrderSAGCostFunction.cpp
vigsterkr: you could just simply wrap the EigenVectorXt with SGVector and not do cloning...
namely
```
SGVector<float64>(gradients.data(), gradients.size())
```
but i reckon it'd be better that simply just create an

```
SGVector<float64_t> gradients(size);
SGVector<float64_t>::EigenVectorXt mapped_gradients = gradients;
/// now you can use mapped_gradients as an eigen vector

return gradients;
```

src/shogun/optimization/StanFirstOrderSAGCostFunction.cpp
```diff
@@ -0,0 +1,140 @@
...
+	REQUIRE(
+	    n > 0,
+	    "Number of sample must be greater than 0, you provided no samples");
+
+	for (index_t i = 0; i < n; ++i)
+	{
+		m_index_of_sample = i;
+		average_gradients += get_gradient();
+	}
+	average_gradients.scale(1.0 / n);
```
vigsterkr: plz use `linalg::scale()`

src/shogun/optimization/StanFirstOrderSAGCostFunction.cpp
```diff
@@ -0,0 +1,140 @@
...
+}
+
+index_t StanFirstOrderSAGCostFunction::get_sample_size()
+{
+	return m_X.num_cols;
+}
+
+SGVector<float64_t> StanFirstOrderSAGCostFunction::get_average_gradient()
+{
+	int32_t params_num = m_trainable_parameters->rows();
```
vigsterkr: auto

src/shogun/optimization/StanFirstOrderSAGCostFunction.cpp
```diff
@@ -0,0 +1,140 @@
...
+
+	auto old_index_sample = m_index_of_sample;
+	auto n = get_sample_size();
+	REQUIRE(
+	    n > 0,
+	    "Number of sample must be greater than 0, you provided no samples");
+
+	for (index_t i = 0; i < n; ++i)
+	{
+		m_index_of_sample = i;
```
vigsterkr: mmmm this feels very very strange... as this definitely makes the whole cost function not thread-safe. i'm not saying that they should be, but i'm not convinced atm that this is actually required, or the only way to solve what you want

tests/unit/optimization/StanFirstOrderSAGCostFunction_unittest.cc
```diff
@@ -0,0 +1,139 @@
...
+	return w;
+}
+
+function<var(StanVector*, float64_t)>
+cost_for_ith_datapoint(SGMatrix<float64_t>& X, SGMatrix<float64_t>& y)
+{
+	auto f_i = [X, y](StanVector* w, int32_t idx) {
+		var wx_y = (*w)(0, 0) * X(0, idx) + (*w)(1, 0) * X(1, idx) - y(0, idx);
+		var res = wx_y * wx_y;
+		res /= 2;
```
vigsterkr: nitpicking, but this is the same as `res *= 0.5` except that the former is slower

tests/unit/optimization/StanStochasticMinimizers_unittest.cc
```diff
@@ -0,0 +1,175 @@
...
+}
+
+function<var(StanVector*, float64_t)>
+cost_for_ith_datapoints(SGMatrix<float64_t>& X, SGMatrix<float64_t>& y)
+{
+	auto f_i = [X, y](StanVector* w, int32_t idx) {
+		var wx_y = (*w)(0, 0) * X(0, idx) + (*w)(1, 0) * X(1, idx) +
+		           (*w)(2, 0) * X(2, idx) - y(0, idx);
+		var res = wx_y * wx_y;
+		res /= 2;
```
vigsterkr: same as above

tests/unit/optimization/StanStochasticMinimizers_unittest.cc
```diff
@@ -0,0 +1,175 @@
...
+	auto f_i = cost_for_ith_datapoints(X, y);
+
+	Matrix<function<var(StanVector*, float64_t)>, Dynamic, 1>
+	    cost_for_ith_point =
+	        Matrix<function<var(StanVector*, float64_t)>, Dynamic, 1>::Constant(
+	            n, 1, f_i);
+
+	auto total_cost = get_tot_cost();
+
+	SquareErrorTestCostFunction* fun = new SquareErrorTestCostFunction(
```
vigsterkr: these all could be auto right?

tests/unit/optimization/StanStochasticMinimizers_unittest.cc
```diff
@@ -0,0 +1,175 @@
...
+	    X, y, &w, &cost_for_ith_point, &total_cost);
+
+	SGDMinimizer* opt = new SGDMinimizer(fun);
```
vigsterkr: auto

tests/unit/optimization/StanStochasticMinimizers_unittest.cc
```diff
@@ -0,0 +1,175 @@
...
+
+	ConstLearningRate* rate = new ConstLearningRate();
```
vigsterkr: auto

tests/unit/optimization/StanStochasticMinimizers_unittest.cc
```diff
@@ -0,0 +1,175 @@
...
+	rate->set_const_learning_rate(0.01);
+
+	GradientDescendUpdater* updater = new GradientDescendUpdater();
```
vigsterkr: auto

tests/unit/optimization/StanStochasticMinimizers_unittest.cc
```diff
@@ -0,0 +1,175 @@
...
+		init();
+	}
+	SGMatrix<float64_t> y;
+	SGMatrix<float64_t> x;
+	void init();
+};
+
+void RegressionData::init()
+{
+	// The data is referenced from Wu Lin Unit tests
```
iglesias: Would it make sense to have it in a common place?

tests/unit/optimization/StanStochasticMinimizers_unittest.cc
```diff
@@ -0,0 +1,175 @@
...
+	SGDMinimizer* opt = new SGDMinimizer(fun);
+
+	ConstLearningRate* rate = new ConstLearningRate();
+	rate->set_const_learning_rate(0.01);
+
+	GradientDescendUpdater* updater = new GradientDescendUpdater();
+	opt->set_gradient_updater(updater);
+	opt->set_learning_rate(rate);
+
+	int32_t num_passes = 20;
```
iglesias: It is good to be consistent.

Above you are doing `rate->set_const_learning_rate(0.01)` and here instead you declare and define `num_passes` before passing it to the setter.

tests/unit/optimization/StanStochasticMinimizers_unittest.cc
```diff
@@ -0,0 +1,175 @@
...
+		var wx_y = (*w)(0, 0) * X(0, idx) + (*w)(1, 0) * X(1, idx) +
+		           (*w)(2, 0) * X(2, idx) - y(0, idx);
+		var res = wx_y * wx_y;
+		res /= 2;
+		return res;
+	};
+	return f_i;
+}
+
+function<var(StanVector*)> get_tot_cost()
```
iglesias: Can this `StanVector*` be a `const StanVector&` instead?

tests/unit/optimization/StanStochasticMinimizers_unittest.cc
```diff
@@ -0,0 +1,175 @@
...
+{
+	auto cost = [](StanVector* v) {
+		var total_cost = v->sum();
```
iglesias: If kept as *, there must be a check for nullptr.

tests/unit/optimization/StanStochasticMinimizers_unittest.cc
```diff
@@ -0,0 +1,175 @@
...
+	auto w = generateParameters();
+
+	auto f_i = cost_for_ith_datapoints(X, y);
+
+	Matrix<function<var(StanVector*, float64_t)>, Dynamic, 1>
+	    cost_for_ith_point =
+	        Matrix<function<var(StanVector*, float64_t)>, Dynamic, 1>::Constant(
+	            n, 1, f_i);
+
+	auto total_cost = get_tot_cost();
```
iglesias: Nitpicking, it'd make sense to just have the lambda here.

tests/unit/optimization/StanFirstOrderSAGCostFunction_unittest.cc
```diff
@@ -0,0 +1,139 @@
...
+	StanVector w(2, 1);
+	var m(1), c(0);
+	w(0, 0) = c;
+	w(1, 0) = m;
+
+	return w;
+}
+
+function<var(StanVector*, float64_t)>
+cost_for_ith_datapoint(SGMatrix<float64_t>& X, SGMatrix<float64_t>& y)
```
iglesias: Ok. So, I would like to understand this pattern based on declaring a function that returns a std::function, and the function's definition is a lambda. It feels like it is over-complicated.

Why not simply defining a function? Please explain.

src/shogun/optimization/StanFirstOrderSAGCostFunction.cpp
FaroukY: @vigsterkr  But the data was being destroyed since the variable was local. I tried the second one but when I returned gradients, it just contained garbage values?

vigsterkr: if you return SGVector<float64_t> that should trigger a copy-ctor that should ++ the ref counter and hence shouldn't delete the data itself.

vigsterkr: there should be many examples where this is used in shogun.... e.g. https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/machine/gp/GaussianLikelihood.cpp#L238

tests/unit/optimization/StanStochasticMinimizers_unittest.cc
FaroukY: @iglesias Do you mean to put his name in a more obvious place? If so then I've already put his name in the header (copyrights). No plagiarism here :)

tests/unit/optimization/StanFirstOrderSAGCostFunction_unittest.cc
FaroukY: Just refactoring. I was using the same definition of a lambda twice in the file, so instead I just wrapped the lambda definition in a function: (cost_for_ith_datapoint) and used it in both places that were defining the lambdas so there is less copy paste.

tests/unit/optimization/StanStochasticMinimizers_unittest.cc
FaroukY: Agreed, its only used once here so no need to define a function that wraps the lambda.

tests/unit/optimization/StanStochasticMinimizers_unittest.cc
iglesias: No. That comment refers to all the lines below, where x and y are initialized.

tests/unit/optimization/StanFirstOrderSAGCostFunction_unittest.cc
iglesias: Just use a function that returns a 'var' (instead of std::function). For the definition of the function just use the logic in the body of the lambda.

src/shogun/optimization/StanFirstOrderSAGCostFunction.cpp
FaroukY: This is one of the things I'm discussing with you today.

src/shogun/optimization/StanFirstOrderSAGCostFunction.cpp
```diff
@@ -0,0 +1,113 @@
...
+		average_gradients += get_gradient();
+	}
+  average_gradients.scale(1.0/n);
+  //lingalg::scale(average_gradients, 1.0/n);
```
FaroukY: @vigsterkr  Do I need to set the gpu backend before using linalg::scale ?

src/shogun/optimization/StanFirstOrderSAGCostFunction.h
```diff
@@ -0,0 +1,177 @@
...
+	 * The cost function must be Written as a finite sample-specific sum of
+	 * cost.
+	 * For example, least squares cost function,
+	 * \f[
+	 * f(w)=\frac{ \sum_i{ (y_i-w^T x_i)^2 } }{2}
+	 * \f]
+	 * where \f$(y_i,x_i)\f$ is the i-th sample,
+	 * \f$y_i\f$ is the label and \f$x_i\f$ is the features
+	 */
+	class StanFirstOrderSAGCostFunction : public FirstOrderSAGCostFunction
```
iglesias: Not sure if this relationship between `StanFirstOrderSAGCostFunction` and `FirstOrderSAGCostFunction` makes sense.

src/shogun/optimization/StanFirstOrderSAGCostFunction.h
```diff
@@ -0,0 +1,177 @@
...
+		/** Updates m_trainable_parameters values to m_ref_trainable_parameters
+		 */
+		void update_stan_vectors_to_reference_values();
+
+	protected:
+		/** X is the training data in column major matrix format */
+		SGMatrix<float64_t> m_X;
+
+		/** y is the ground truth, or the correct prediction */
+		SGMatrix<float64_t> m_y;
```
iglesias: `FirstOrderSAGCostFunction` does not have members for the training data. Why are they needed here?

src/shogun/optimization/StanFirstOrderSAGCostFunction.cpp
FaroukY: This should be addressed in the next set of commits.

src/shogun/optimization/StanFirstOrderSAGCostFunction.h
FaroukY: Agreed. I've changed the parent class to be FirstOrderStochasticCostFunction, since all these implemented functions are directly from it. (FirstOrderSAGCostFunction also inherits from FirstOrderStochasticCostFunction so it makes sense here to inherit from it as the Stan version is an alternative of FirstOrderSAGCostFunction).

src/shogun/optimization/StanFirstOrderSAGCostFunction.h
```diff
@@ -0,0 +1,177 @@
...
+
+		/** trainable_parameters are the variables that are optimized for */
+		StanVector& m_trainable_parameters;
```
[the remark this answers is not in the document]
FaroukY: I was just thinking of it as a design problem. As in aggregation vs composition. The cost function doesn't own the parameters, so I felt aggregation works better here (implemented in C++ as a reference) What do you think?

src/shogun/optimization/StanFirstOrderSAGCostFunction.h
FaroukY: @iglesias In the unit test of FirstOrderSAGCostFunction, we implement an additional class called CRegressionExample that wraps FirstOrderSAGCostFunction and contains the training data. I've simply removed the necessity for that class since it just acted as a wrapper around the cost function and data, and have included the data in the loss function, or atleast a reference to the data.

tests/unit/optimization/StanStochasticMinimizers_unittest.cc
FaroukY: @iglesias Do you mean you'd like the data to be in a more clear place (Like the top) rather than in the middle?

src/shogun/optimization/StanFirstOrderSAGCostFunction.h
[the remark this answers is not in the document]
iglesias: I think a reference member should not be used but maybe it is ok.

tests/unit/optimization/StanStochasticMinimizers_unittest.cc
```diff
@@ -0,0 +1,167 @@
...
+using namespace stan::math;
+using std::function;
+
+StanVector generateParameters()
+{
+	StanVector w(3, 1);
+	var x1(0), x2(0), x3(0);
+	w(0, 0) = x1;
+	w(1, 0) = x2;
+	w(2, 0) = x3;
```
iglesias: The syntax to initialize this `StanVector` is a bit awkward. Can you do better? For example with something std::vector-like it could just look like `std::vector<var>{x1, x2, x3}`.

tests/unit/optimization/StanStochasticMinimizers_unittest.cc
```diff
@@ -0,0 +1,167 @@
...
+	return w;
+}
+
+function<var(const StanVector&, float64_t)>
+cost_for_ith_datapoints(SGMatrix<float64_t>& X, SGMatrix<float64_t>& y)
```
iglesias: Refactor.

tests/unit/optimization/StanStochasticMinimizers_unittest.cc
```diff
@@ -0,0 +1,167 @@
...
+	int32_t n = 10;
+
+	RegressionData data;
+	auto X = data.x;
+	auto y = data.y;
+	auto w = generateParameters();
+
+	auto f_i = cost_for_ith_datapoints(X, y);
+
+	Matrix<function<var(const StanVector&, float64_t)>, Dynamic, 1>
```
iglesias: This is a case where auto really makes sense actually.

tests/unit/optimization/StanStochasticMinimizers_unittest.cc
```diff
@@ -0,0 +1,167 @@
...
+	    cost_for_ith_point =
+	        Matrix<function<var(const StanVector&, float64_t)>, Dynamic, 1>::Constant(
+	            n, 1, f_i);
+
+	auto total_cost = [](const StanVector& v) {
+		var total_cost = v.sum();
+		return total_cost;
+	};
+
+	auto fun = new SquareErrorTestCostFunction(
```
iglesias: No deletion for this one and the CostLearningRate?

tests/unit/optimization/StanStochasticMinimizers_unittest.cc
FaroukY: Hmm, well StanVector is just an alias for Eigen<var, Dynamic, 1> so I am not sure I can change the constructor unless I define a wrapper class for Eigen<var, Dynamic, 1>.

tests/unit/optimization/StanStochasticMinimizers_unittest.cc
FaroukY: I wanted to cast here, since the output is not exactly:

Matrix<function<var(const StanVector&, float64_t)>, Dynamic, 1>

tests/unit/optimization/StanStochasticMinimizers_unittest.cc
FaroukY: @iglesias They're handled in the minimizer's destructor.

tests/unit/optimization/StanStochasticMinimizers_unittest.cc
iglesias: Please have a look at initializing Eigen::Matrix in Eigen documentation.

https://eigen.tuxfamily.org/dox/group__TutorialAdvancedInitialization.html

tests/unit/optimization/StanStochasticMinimizers_unittest.cc
iglesias: What do you mean?

src/shogun/neuralnets/StanNeuralLogisticLayer.h
```diff
@@ -0,0 +1,53 @@
...
+{
+/** @brief Neural layer with linear neurons, with a [logistic activation
+ * function](http://en.wikipedia.org/wiki/Logistic_function). can be used as a
+ * hidden layer or an output layer.
+ *
+ * When used as an output layer, a
+ * [squared error measure](http://en.wikipedia.org/wiki/Mean_squared_error) is
+ * used
+ */
+class StanNeuralLogisticLayer : public StanNeuralLinearLayer
```
iglesias: Perhaps it is only a naming issue, but saying that a LogisticLayer is a LinearLayer sounds odd.

src/shogun/neuralnets/StanNeuralLogisticLayer.h
```diff
@@ -0,0 +1,53 @@
...
+{
+public:
+	/** default constructor */
+	StanNeuralLogisticLayer();
+
+	/** Constuctor
```
iglesias: Minor typo: constructor.

src/shogun/neuralnets/StanNeuralLogisticLayer.h
```diff
@@ -0,0 +1,53 @@
...
+	 * parameters of the layer
+	 *
+	 * @param layers Array of layers that form the network that this layer is
+	 * being used with
+	 */
+	virtual void compute_activations(StanVector& parameters,
+			CDynamicObjectArray* layers);
+
+
+	virtual const char* get_name() const { return "NeuralLogisticLayer"; }
```
iglesias: Missing Stan prefixed in the name?

src/shogun/neuralnets/StanNeuralNetwork.h
```diff
@@ -0,0 +1,195 @@
...
+	virtual void disconnect_all();
+
+	/** Initializes the network
+	 *
+	 * @param sigma standard deviation of the gaussian used to randomly
+	 * initialize the parameters
+	 */
+	virtual void initialize_neural_network(float64_t sigma = 0.01f);
+
+	virtual ~CNeuralNetwork();
```
iglesias: Did you mean destructor of `StanNeuralNetwork`?

src/shogun/neuralnets/StanNeuralNetwork.h
```diff
@@ -0,0 +1,195 @@
...
+
+	/** apply machine to data in means of binary classification problem */
+	virtual CBinaryLabels* apply_binary(CFeatures* data);
+	/** apply machine to data in means of regression problem */
+	virtual CRegressionLabels* apply_regression(CFeatures* data);
+	/** apply machine to data in means of multiclass classification problem */
+	virtual CMulticlassLabels* apply_multiclass(CFeatures* data);
```
iglesias: I am not sure these ones should be here. Let's check with @karlnapf or @vigsterkr.

src/shogun/neuralnets/StanNeuralNetwork.cpp
```diff
@@ -0,0 +1,280 @@
...
+ 			layer->compute_activations(get_section(m_params, i), m_layers);
+
+ 		layer->dropout_activations();
+ 	}
+
+ 	return get_layer(j)->get_activations();
+ }
+
+
+ SGVector<float64_t>* StanNeuralNetwork::get_layer_parameters(int32_t i)
```
iglesias: This looks strange. The signature in the header is `StanVector& get_layer_parameters(int32_t i);`.

tests/unit/optimization/StanStochasticMinimizers_unittest.cc
FaroukY: Will be refactoring the stan tests once the neural network is working

src/shogun/neuralnets/StanNeuralLogisticLayer.h
iglesias: Any plans to fix?

src/shogun/optimization/StanFirstOrderSAGCostFunction.h
iglesias: We must be careful and we must no add unnecessary relationships. If it turns out that the training data is a member both in the cost function, and in the neural network, and in etc etc, that is going to cause lot of usage confusion.

It does not sounds unreasonable that the tests have a facility packaged to prepare input data and avoid code duplication (e.g. inside a class a the CRegressionExample you are mentioning). What did you find wrong with it?

src/shogun/optimization/StanFirstOrderSAGCostFunction.h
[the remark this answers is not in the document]
iglesias: Any news about this concern?

FaroukY:
@iglesias 

I previously discussed with Viktor that I'd first finish writing the whole Neural Network module (Which is not already done), then write the unit tests for it and make sure the logic is correct, and then we can go back and refactor all the points addressed previously. I've already addressed some points mentioned, but the main focus was getting a MVP of the Neural network running only on stan for gradient calculation. I'm not ignoring the points, but I'm just more concerned now with finishing the actual logic (Then I can refactor/change names and do the other required changes). 

I'm currently writing the unit tests for the whole neural network code, and this will be the last file to be added to this PR. It should be a standalone PR changing to a tested NN to use stan only.

iglesias:
Hi @FaroukY! That's all right. Though I find that a ~2000 lines patch for a so-called MVP has a bit of over-kill smell.

Still, I do suggest you to keep in mind the comments that have already been made. Just think that you don't want to build upon something that will need to be completely changed from its foundations. I am not necessarily saying that is the case here, but I believe you should keep it mind.

PR #4301, opened 2018-05-24
Use wrap in assignment expr in cpp meta examples

vinx13:


vinx13:
let's address ref count things in transformers first and see how we deal with smart pointers.

karlnapf:
I am confused :) But I will ignore this while it is closed

vinx13:
i've changed ref count things in transformer, and now use a new variable name will work. the input are still valid (and no memory leak using smart pointer :)) after calling apply

PR #4302, opened 2018-05-24
Update notebooks using feature factory

vinx13:
Continue from #4269  #4292 , finally we get rid of `RealFeatures` in all notebooks


karlnapf:
amazing! thanks for this push!

PR #4305, opened 2018-05-24
progress bar in iterative algorithms

shubham808:
implements the progress bar in most of the algorithms from the wiki page

### DIFF
--- a/src/shogun/base/progress.h
+++ b/src/shogun/base/progress.h
@@ -58,10 +58,12 @@
 
 namespace shogun
 {
-//#define progress(...)
-// Progress(std::string(this->get_name())+"::"+std::string(__FUNCTION__),
-//__VA_ARGS__)
-#define progress(...) Progress(std::string(__FUNCTION__), __VA_ARGS__)
+#define SG_PROGRESS(...)                                                       \
+	progress(                                                                  \
+	    std::string(this->get_name()) + "::" + std::string(__FUNCTION__),      \
+	    *this->io, __VA_ARGS__)
+
+#define SG_SPROGRESS(...) progress(__FUNCTION__, __VA_ARGS__)
 
 	/** Possible print modes */
 	enum SG_PRG_MODE
@@ -715,8 +717,8 @@ namespace shogun
 	 * @param	condition	premature stopping condition
 	 */
 	template <typename T>
-	inline PRange<T> Progress(
-	    std::string prefix, Range<T> range, const SGIO& io,
+	inline PRange<T> progress(
+	    std::string prefix, const SGIO& io, Range<T> range,
 	    SG_PRG_MODE mode = UTF8,
 	    std::function<bool()> condition = []() { return true; })
 	{
@@ -736,7 +738,7 @@ namespace shogun
 	 * @param	condition	premature stopping condition
 	 */
 	template <typename T>
-	inline PRange<T> Progress(
+	inline PRange<T> progress(
 	    std::string prefix, Range<T> range, SG_PRG_MODE mode = UTF8,
 	    std::function<bool()> condition = []() { return true; })
 	{
@@ -749,10 +751,23 @@ namespace shogun
 	 * @param condition premature stopping condition
 	 */
 	template <typename T>
-	inline PRange<T> Progress(
+	inline PRange<T> progress(
 	    std::string prefix, Range<T> range, std::function<bool()> condition)
 	{
-		return PRange<T>(range, *sg_io, "PROGRESS: ", UTF8, condition);
+		return PRange<T>(range, *sg_io, prefix, UTF8, condition);
+	}
+	/** Creates @ref PRange given a range and a stopping condition
+	 *
+	 * @param range range used
+	 * @param io SGIO object
+	 * @param condition premature stopping condition
+	 */
+	template <typename T>
+	inline PRange<T> progress(
+	    std::string prefix, const SGIO& io, Range<T> range,
+	    std::function<bool()> condition)
+	{
+		return PRange<T>(range, io, prefix, UTF8, condition);
 	}
 };
 #endif /* __SG_PROGRESS_H__ */
--- a/src/shogun/base/progress.h
+++ b/src/shogun/base/progress.h
@@ -58,6 +58,10 @@
 
 namespace shogun
 {
+//#define progress(...)
+// Progress(std::string(this->get_name())+"::"+std::string(__FUNCTION__),
+//__VA_ARGS__)
+#define progress(...) Progress(std::string(__FUNCTION__), __VA_ARGS__)
 
 	/** Possible print modes */
 	enum SG_PRG_MODE
@@ -711,8 +715,8 @@ namespace shogun
 	 * @param	condition	premature stopping condition
 	 */
 	template <typename T>
-	inline PRange<T> progress(
-	    Range<T> range, const SGIO& io, std::string prefix = "PROGRESS: ",
+	inline PRange<T> Progress(
+	    std::string prefix, Range<T> range, const SGIO& io,
 	    SG_PRG_MODE mode = UTF8,
 	    std::function<bool()> condition = []() { return true; })
 	{
@@ -732,9 +736,8 @@ namespace shogun
 	 * @param	condition	premature stopping condition
 	 */
 	template <typename T>
-	inline PRange<T> progress(
-	    Range<T> range, std::string prefix = "PROGRESS: ",
-	    SG_PRG_MODE mode = UTF8,
+	inline PRange<T> Progress(
+	    std::string prefix, Range<T> range, SG_PRG_MODE mode = UTF8,
 	    std::function<bool()> condition = []() { return true; })
 	{
 		return PRange<T>(range, *sg_io, prefix, mode, condition);
@@ -746,7 +749,8 @@ namespace shogun
 	 * @param condition premature stopping condition
 	 */
 	template <typename T>
-	inline PRange<T> progress(Range<T> range, std::function<bool()> condition)
+	inline PRange<T> Progress(
+	    std::string prefix, Range<T> range, std::function<bool()> condition)
 	{
 		return PRange<T>(range, *sg_io, "PROGRESS: ", UTF8, condition);
 	}
--- a/src/shogun/classifier/AveragedPerceptron.cpp
+++ b/src/shogun/classifier/AveragedPerceptron.cpp
@@ -1,7 +1,7 @@
 /*
  * This software is distributed under BSD 3-clause license (see LICENSE file).
  *
- * Authors: Soeren Sonnenburg, Sergey Lisitsyn, Giovanni De Toni, 
+ * Authors: Soeren Sonnenburg, Sergey Lisitsyn, Giovanni De Toni,
  *          Michele Mazzoni
  */
 
@@ -9,6 +9,7 @@
 #include <shogun/labels/Labels.h>
 #include <shogun/mathematics/Math.h>
 #include <shogun/labels/BinaryLabels.h>
+#include <shogun/base/progress.h>
 #include <shogun/lib/Signal.h>
 
 using namespace shogun;
@@ -69,9 +70,10 @@ bool CAveragedPerceptron::train_machine(CFeatures* data)
 	for (int32_t i=0; i<num_feat; i++)
 		w[i]=1.0/num_feat;
 
-	//loop till we either get everything classified right or reach max_iter
 
-	while (!converged && iter < max_iter)
+  auto pb = progress(range(max_iter));
+	//loop till we either get everything classified right or reach max_iter
+	while ((!converged && iter < max_iter))
 	{
 		COMPUTATION_CONTROLLERS
 		converged=true;
@@ -95,8 +97,9 @@ bool CAveragedPerceptron::train_machine(CFeatures* data)
 			tmp_bias+=bias;
 		}
 		iter++;
+		pb.print_progress();
 	}
-
+	pb.complete();
 	if (converged)
 		SG_INFO("Averaged Perceptron algorithm converged after %d iterations.\n", iter)
 	else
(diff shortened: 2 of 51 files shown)


src/shogun/machine/BaggingMachine.cpp
```diff
@@ -177,7 +178,8 @@ bool CBaggingMachine::train_machine(CFeatures* data)
 	for (index_t i = 0; i < m_num_bags*m_bag_size; ++i)
 		rnd_indicies.matrix[i] = CMath::random(0, m_bag_size-1);
 
-	#pragma omp parallel for
+	auto pb = progress(range(m_num_bags));
```
karlnapf: @geektoni @shubham808 does the progress bar work gracefully from within omp?

geektoni: Yes, it does :)

karlnapf:
Maybe we could print classname::methodname using PRETTY_FUNCTION by default?

geektoni:
Yeah, I agree with PRETTY_FUNCTION too.

karlnapf:
@shubham808 could you look into that. See `SGIO.H` for how it is done for `SG_PRINT`

src/shogun/base/progress.h
```diff
@@ -58,6 +58,10 @@
 
 namespace shogun
 {
+//#define progress(...)
+// Progress(std::string(this->get_name())+"::"+std::string(__FUNCTION__),
```
shubham808: due to a class that use a progress bar being external i have commented it out for now

karlnapf: not sure I understand

geektoni: I think that is caused by the fact that some Shogun's class does not have the `get_name` method, because they do not inherith from SGObject. For those classes I think that using the standard progress bar could be enough.

src/shogun/base/progress.h
```diff
@@ -58,6 +58,10 @@
...
+//__VA_ARGS__)
+#define progress(...) Progress(std::string(__FUNCTION__), __VA_ARGS__)
```
geektoni: As a common practice, shouldn't we name macros with uppercase names (so that we know it is a macro and not a function/method call)? It may be a pain to rename everything again though.

src/shogun/base/progress.h
karlnapf: lets have PROGRESS and SPROGRESS (static version). The static version accepts a string as the first argument, callers have to provide the class name manually in those cases

src/shogun/base/progress.h
```diff
@@ -711,8 +715,8 @@ namespace shogun
 	 * @param	condition	premature stopping condition
 	 */
 	template <typename T>
-	inline PRange<T> progress(
-	    Range<T> range, const SGIO& io, std::string prefix = "PROGRESS: ",
+	inline PRange<T> Progress(
+	    std::string prefix, Range<T> range, const SGIO& io,
```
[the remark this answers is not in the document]
shubham808: because the macro was having trouble infering type if we kept it in between or last

src/shogun/distance/Distance.cpp
```diff
@@ -243,7 +243,7 @@ void CDistance::do_precompute_matrix()
 	SG_FREE(precomputed_matrix);
 	precomputed_matrix=SG_MALLOC(float32_t, num*(num+1)/2);
 
-	for (auto i : progress(range(num), *this->io))
+	for (auto i : SG_PROGRESS(range(num), *this->io))
```
karlnapf: why the io passing here?

src/shogun/features/DotFeatures.cpp
```diff
@@ -63,7 +63,7 @@ void CDotFeatures::dense_dot_range(float64_t* output, int32_t start, int32_t sto
 
 	int32_t num_threads;
 	int32_t step;
-	auto pb = progress(range(num_vectors), *this->io);
+	auto pb = SG_PROGRESS(range(num_vectors), *this->io);
```
karlnapf: why the this->io? can't we do this via the macro as well?

shubham808: we dont always pass *this->io so i guess its better to pass it when we need to

src/shogun/base/progress.h
```diff
@@ -58,6 +58,15 @@
...
 namespace shogun
 {
+#define SG_PROGRESS(...)                                                       \
+	progress(                                                                  \
+	    std::string(this->get_name()) + "::" + std::string(__FUNCTION__),      \
+	    *this->io, __VA_ARGS__)
+
+#define SG_SPROGRESS(...)                                                      \
+	progress(                                                                  \
+	    std::string(this->get_name()) + "::" + std::string(__FUNCTION__),      \
```
karlnapf: I dont understand, there is no `this` in a static context.

src/shogun/classifier/AveragedPerceptron.cpp
```diff
@@ -1,15 +1,16 @@
 /*
  * This software is distributed under BSD 3-clause license (see LICENSE file).
  *
- * Authors: Soeren Sonnenburg, Sergey Lisitsyn, Giovanni De Toni, 
+ * Authors: Soeren Sonnenburg, Sergey Lisitsyn, Giovanni De Toni,
```
karlnapf: could you pls check your editor and make it not do such newline character changes? They pollute the diff

src/shogun/classifier/AveragedPerceptron.cpp
```diff
@@ -69,9 +70,9 @@ bool CAveragedPerceptron::train_machine(CFeatures* data)
 	for (int32_t i=0; i<num_feat; i++)
 		w[i]=1.0/num_feat;
 
-	//loop till we either get everything classified right or reach max_iter
-
-	while (!converged && iter < max_iter)
+	auto pb = SG_SPROGRESS(range(max_iter));
+	// loop till we either get everything classified right or reach max_iter
+	while ((!converged && iter < max_iter))
```
karlnapf: why the additional?

src/shogun/classifier/Perceptron.cpp
```diff
@@ -73,7 +74,7 @@ bool CPerceptron::train_machine(CFeatures* data)
 		bias=0;
 		linalg::add_scalar(w, 1.0 / num_feat);
 	}
-
+	auto pb = SG_SPROGRESS(range(max_iter));
```
karlnapf: SPROGRESS would be used in static contexts only ...

src/shogun/base/progress.h
```diff
@@ -58,6 +58,12 @@
 
 namespace shogun
 {
+#define SG_PROGRESS(...)                                                       \
+	progress(                                                                  \
+	    std::string(this->get_name()) + "::" + std::string(__FUNCTION__),      \
+	    *this->io, __VA_ARGS__)
+
+#define SG_SPROGRESS(...) progress("PROGRESS: ", __VA_ARGS__)
```
shubham808: I think just "PROGRESS" is the simplest version here :)

geektoni:
sooo, this should be good to merge, right? @karlnapf, @shubham808

src/shogun/kernel/string/WeightedDegreePositionStringKernel.cpp
```diff
@@ -1250,37 +1250,37 @@ void CWeightedDegreePositionStringKernel::compute_batch(
...
+			pb.print_progress();
 		    }
 		    pb.complete();
 	}
 #ifdef HAVE_PTHREAD
 	else
 	{
 
-		auto pb = progress(range(num_feat), *this->io);
+		auto pb = SG_PROGRESS(range(num_feat));
```
karlnapf: same here, why not put it into the loop?

src/shogun/kernel/string/WeightedDegreePositionStringKernel.cpp
```diff
@@ -1423,7 +1423,7 @@ float64_t* CWeightedDegreePositionStringKernel::compute_scoring(
 	info.R_k = NULL;
 
 	// === main loop
-	auto pb = progress(range(num_feat * max_degree), *this->io);
+	auto pb = SG_PROGRESS(range(num_feat * max_degree));
 	for( k = 0; k < max_degree; ++k )
```
karlnapf: I wonder whether having only an outer progress would be better?
@geektoni ?

src/shogun/lib/external/libocas.cpp
```diff
@@ -581,7 +581,7 @@ ocas_return_value_T svm_ocas_solver(
   Bi=NULL;
 
   /* Initialize progress bar */
-  auto pb = progress(range(10));
+  auto pb = SG_SPROGRESS(range(10));
```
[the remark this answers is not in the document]
karlnapf: nice!

src/shogun/machine/BaggingMachine.cpp
```diff
@@ -177,7 +178,8 @@ bool CBaggingMachine::train_machine(CFeatures* data)
 	for (index_t i = 0; i < m_num_bags*m_bag_size; ++i)
 		rnd_indicies.matrix[i] = CMath::random(0, m_bag_size-1);
 
-	#pragma omp parallel for
+	auto pb = SG_PROGRESS(range(m_num_bags));
+#pragma omp parallel for
 	for (int32_t i = 0; i < m_num_bags; ++i)
```
karlnapf: if omp is used, I assume the `pb` definition needs to happen before (and cannot be done in the for loop?) @geektoni ?

src/shogun/base/progress.h
karlnapf: Sorry for being picky here, but I realised that this is not nicest.
(This doesnt block a merge though, we can change this later)

 Maybe use the filename (without suffix) instead? @vigsterkr thoughts?

karlnapf: Could use `__FILE__` and something like 
```
std::string remove_extension(const std::string& filename) {
    size_t lastdot = filename.find_last_of(".");
    if (lastdot == std::string::npos) return filename;
    return filename.substr(0, lastdot); 
}
```

karlnapf:
I have a few last Qs that I would like to have resolved before merging, but it is almost good!

src/shogun/kernel/string/WeightedDegreePositionStringKernel.cpp
```diff
@@ -1250,37 +1250,37 @@ void CWeightedDegreePositionStringKernel::compute_batch(
...
-		   params.max_shift = max_shift;
-		   params.shift = shift;
-		   params.vec_idx = vec_idx;
-		   compute_batch_helper((void*)&params);
-
-		   pb.print_progress();
+		auto pb = SG_PROGRESS(range(num_feat));
+		// TODO: replace with the new signal
+		// for (int32_t j=0; j<num_feat && !CSignal::cancel_computations(); j++)
+		for (int32_t j = 0; j < num_feat; j++)
```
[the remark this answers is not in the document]
shubham808: this can be moved in the loop sure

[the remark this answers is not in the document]
karlnapf: let's do it in all the places then (apart from the mp case where I dont know whether that will work)

shubham808:
@karlnapf the CI fails seem unrelated.. i think ?

src/shogun/kernel/string/WeightedDegreePositionStringKernel.cpp
geektoni: I agree. Stacked progress bars are not handled very well at the moment.

src/shogun/machine/BaggingMachine.cpp
geektoni: Yes, the ``pb`` definition has to appear before the ``#pragma``.

geektoni:
@shubham808 yes, the CI errors should be unrelated (I do not know why it took so long to complete).

src/shogun/base/progress.h
karlnapf: this is still an open question

geektoni: I would prefer to see the actual object name, because if I had three progress bar inside three different ``train()`` method, I would like to see to which model (``CPerceptron::train()``, ``CAveragedPerceptron::train()`` etc.) they belong. So it is okay as it is done now :)

vigsterkr:
the bc2c9e50ace0eb64e52ecc49087c2822bcbf0d4d commit will hurt later... that merge should not be merged anywhere... @shubham808 plz fix it

shubham808:
@vigsterkr my bad.. I will fix it

src/shogun/base/progress.h
karlnapf: There is no object for SG_SPROGRESS. We can either make the caller of the macro pass something or use the __FUNCTION__. But I think just using "Progress" is not the best or?

shubham808: I have used the ```__FUNCTION__``` macro here now :)

src/shogun/base/progress.h
```diff
@@ -58,6 +58,12 @@
 
 namespace shogun
 {
+#define SG_PROGRESS(...)                                                       \
+	progress(                                                                  \
+	    std::string(this->get_name()) + "::" + std::string(__FUNCTION__),      \
+	    *this->io, __VA_ARGS__)
+
+#define SG_SPROGRESS(...) progress(__FUNCTION__, __VA_ARGS__)
```
karlnapf: cool!
How does it look like with this? Can you post and example?

shubham808:
CustomKernel::get_kernel_matrix |████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 100.00%   0.0 seconds 

KMeansMiniBatch::minibatch_KMeans |██████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 100.00%   0.0 seconds

@geektoni progress bar 2.0 ;)

karlnapf:
Awesome!

karlnapf:
How does it look for the static version?

Issue #4284, opened 2018-05-14
drop HAVE_CXX11 macro


vigsterkr:
there's still some places where we have an `ifdef HAVE_CXX11`, but since 6.1.0 release shogun requires that your compile actually supports c++11.

just do a `git grep HAVE_CXX11` and eliminate those macro checks.

abhishek-1991:
Hi Viktor,

I would like to take up this task.

Thanks and Regards,
Abhishek.

vigsterkr:
@abhishek-1991 sure thing, since this is a rather large effort, just start with a file where you find this macro and send in patches.

abhishek-1991:
Hi Victor,

Sorry for returning after long time. I was busy else where. I pulled the latest build today but it is giving compilation error for file "LaRank.cpp". I am using g++ 5.4.0

Thanks and Regards,
Abhishek.

PR #4314, opened 2018-05-30
[Issue #4284] Removed HAVE_CX11

abhishek-1991:
Removed #ifndef HAVE_CXX11
Issue #4284


karlnapf:
Thanks for the patch. There should be many more such instances.

PR #4316, opened 2018-05-30
Drop apply_preprocessor and preprocessors states in features

vinx13:


tests/unit/classifier/GaussianProcessClassification_unittest.cc
```diff
@@ -595,11 +595,13 @@ TEST_F(GaussianProcessClassification, apply_preprocessor_and_binary)
 	CRescaleFeatures* preproc=new CRescaleFeatures();
 	preproc->fit(features_train);
 
-	features_train->add_preprocessor(preproc);
-	features_train->apply_preprocessor();
+	features_train =
```
vigsterkr: i know that this does not come from this patch but in case of this unit test do we really need to have `features_train` to have `CDenseFeatures<float64_t>` type?

PR #4317, opened 2018-05-30
Add pipeline

vinx13:


src/shogun/machine/Pipeline.cpp
```diff
@@ -0,0 +1,116 @@
...
+		if (train_require_labels())
+		{
+			REQUIRE(m_labels, "No labels given.\n");
+		}
+
+		for (auto&& stage : m_stages)
+		{
+			if (holds_alternative<CTransformer*>(stage))
+			{
+				CTransformer* transformer = shogun::get<CTransformer*>(stage);
```
vigsterkr: oooootoooo, auto :)

src/shogun/machine/Pipeline.cpp
```diff
@@ -0,0 +1,116 @@
...
+				if (transformer->train_require_labels())
```
vigsterkr: this could be a simple ternary expression:

```
transformer->train_require_labels() ? transformer->fit(data, m_labels) : transformer->fit(data);

File: src/shogun/machine/Pipeline.cpp [hunk 3]
```diff
@@ -0,0 +1,116 @@
...
+		return true;
+	}
+
+	CLabels* CPipeline::apply(CFeatures* data)
+	{
+		for (auto&& stage : m_stages)
+		{
+			if (holds_alternative<CTransformer*>(stage))
+			{
+				CTransformer* transformer = shogun::get<CTransformer*>(stage);
```
vigsterkr: auto

src/shogun/machine/Pipeline.cpp
```diff
@@ -0,0 +1,116 @@
...
+				if (transformer->train_require_labels())
+					transformer->fit(data, m_labels);
+				else
+					transformer->fit(data);
+
+				data = transformer->transform(data);
+			}
+			else
+			{
+				CMachine* machine = shogun::get<CMachine*>(stage);
```
vigsterkr: auto?

vigsterkr:
@vinx13 would be great to add a simple xval cookbook where we use a CPruneVarSubMean transformer before doing PCA and applying k-means on a toy data :D

vigsterkr:
+ adding unit test would be great

vigsterkr:
aaaah and another thing: getter methods for the pipeline elements would be desirable. i.e. get the trained transformer etc.

src/shogun/machine/Pipeline.cpp
```diff
@@ -0,0 +1,149 @@
...
+			    stage);
+
+			if (require_labels)
+				return require_labels;
+		}
+
+		return require_labels;
+	}
+
+	void CPipeline::list_stages() const
```
vigsterkr: how about renaming this to
```
std::string to_string() const;
```
and basically return a string instead of directly writing SG_INFO.... that of course means that SGObject::to_string needs to be virtual

src/shogun/machine/Pipeline.cpp
```diff
@@ -0,0 +1,149 @@
...
+		for (auto i : range(m_stages.size()))
+		{
+			SG_INFO("[%d] ", i);
+			visit(
+			    [this](auto&& object) { SG_INFO("%s\n", object->get_name()); },
+			    m_stages[i]);
+		}
+	}
+
+	CTransformer* CPipeline::get_transformer(size_t index) const
```
vigsterkr: maybe instead of accessing the pipeline elements like this it'd be good to have a way to access them by name? see for example http://scikit-learn.org/stable/tutorial/text_analytics/working_with_text_data.html#evaluation-of-the-performance-on-the-test-set
of course this would mean to be able to add elements as a tuple, say

Pipline().with(("customName", ZeroMean).with....

but that could be that we support both, meaning either having
```
Pipline().with(("customName", ZeroMean).with....
```

and
 
```
Pipline().with(ZeroMean).with....
```

and in the 2nd case we use the get_name() string...

src/shogun/machine/Pipeline.cpp
karlnapf: I like this more as well! users can then print the strings themselves

src/shogun/machine/Pipeline.cpp
```diff
@@ -0,0 +1,176 @@
...
+
+		SG_ERROR("Transformer with name %s not found.\n", name.c_str());
+
+		return nullptr;
+	}
+
+	void CPipeline::check_pipeline()
+	{
+		REQUIRE(!m_stages.empty(), "Pipeline is empty");
+		REQUIRE(
```
vigsterkr: i wonder if we should start using typed exceptions, because for now EVERYTHING is ShogunException. things like `MachineNotTrainedException` etc....
@lisitsyn @karlnapf mm?

src/shogun/machine/Pipeline.cpp
```diff
@@ -0,0 +1,176 @@
...
+	CTransformer* CPipeline::get_transformer(const std::string& name) const
+	{
+		for (auto&& stage : m_stages)
+		{
+			if (stage.first == name &&
+			    holds_alternative<CTransformer*>(stage.second))
+				return shogun::get<CTransformer*>(stage.second);
+		}
+
+		SG_ERROR("Transformer with name %s not found.\n", name.c_str());
```
vigsterkr: i would do here `std::invalid_argument` be the more appropriate...
but in lot of the `REQUIRE()` case it's the same... namely a 'more specific type' of exception would be much more appropriate from a shogun user... if its getting integrated in a larger framework.

tests/unit/machine/Pipeline_unittest.cc
```diff
@@ -0,0 +1,101 @@
...
+	    some<NiceMock<MockCTransformer>>();
+	Some<NiceMock<MockCMachine>> machine = some<NiceMock<MockCMachine>>();
+};
+
+TEST_F(PipelineTest, no_machine)
+{
+	pipeline->with(transformer1);
+
+	auto features = some<NiceMock<MockCFeatures>>();
+	EXPECT_THROW(pipeline->train(features), ShogunException);
```
vigsterkr: the need for a more specific type of exception came up in my head here atm, because expecting there ShogunException is too broad.... it could actually come from a transformer or the machine in the pipeline... right? @karlnapf @lisitsyn @vinx13

src/shogun/machine/Pipeline.cpp
karlnapf: Actually, this was part of the "user experience" GSoC project. So yes makes total sense imo

src/shogun/lib/ShogunException.h
```diff
@@ -48,5 +48,25 @@ class ShogunException: public std::exception
 		/** exception string */
 		std::string msg;
 };
+
+/** @brief Class InvalidStateException defines an exception which is thrown
+ * whenever an object in Shogun in invalid state is used. For example, a machine
+ * is used before training.
+ */
+class InvalidStateException : public ShogunException
```
vigsterkr: i would define these in separate headers and cpp files :)

vigsterkr: and in the case of "a machine is used before training" i would call that exception `MachineNotTrainedException` as invalid state could be a lot of things and the not trained case i believe should be explicitly spelled out for the user :)

vigsterkr:
@lisitsyn @karlnapf so as trying to add pipeline to the swig interface it came out that `with` is a reserved keyword in python. see the design from 2017 ws: https://github.com/shogun-toolbox/shogun/wiki/Hackathon-2017-base-api#pipeline

how about changing `Pipeline.with(Transformer)` to `Pipeline.over(Transformer)`, since `add` and `as` are kind of taken by SGObject, imo `as` could still fly but add would be matched. i'm wondering whether `then` would be a reserved keyword in any languages we support?

and another idea here: imo we should have a `Pipeline` that is inherited from `CMachine`, but the builder should be a separate base class, `PipelineBuilder`.

@vinx13 i would still like to have a builder that supports adding a list of elements. something like: add_stages([transf1, transf1, machine])

vigsterkr:
@karlnapf we are getting travis problems here as well..... this branch is rebased over HEAD od develop, but octave is throwing this error

```
error: in method 'Evaluation_evaluate', argument 2 of type 'shogun::CLabels *' (SWIG_TypeError)

error: called from

    /opt/shogun/build/examples/meta/octave/multiclass/cartree.m at line 39 column 10
```

```
error: in method 'Evaluation_evaluate', argument 2 of type 'shogun::CLabels *' (SWIG_TypeError)

error: called from

    /opt/shogun/build/examples/meta/octave/binary/linear_support_vector_machine.m at line 42 column 10
```

vinx13:
the error also happens on python
```
  4/297 Test #238: generated_python-binary-linear_support_vector_machine ...................................***Failed    0.24 sec

  File "/opt/shogun/build/examples/meta/python/binary/linear_support_vector_machine.py", line 50, in <module>

    accuracy = eval.evaluate(labels_predict, labels_test)

TypeError: in method 'Evaluation_evaluate', argument 2 of type 'shogun::CLabels *'

swig/python detected a memory leak of type 'CBinaryLabels *', no destructor found.
```

PR #4318, opened 2018-05-31
Update classifier and cluster meta examples

vinx13:


src/shogun/clustering/KMeansMiniBatch.cpp
```diff
@@ -147,6 +147,11 @@ void CKMeansMiniBatch::init_mb_params()
 {
 	batch_size=-1;
 	minib_iter=-1;
+
+	SG_ADD(
```
vigsterkr: \o/

src/shogun/clustering/KMeansMiniBatch.cpp
```diff
@@ -147,6 +147,11 @@ void CKMeansMiniBatch::init_mb_params()
...
+		&batch_size, "batch_size", "batch size for mini-batch KMeans",
+		MS_NOT_AVAILABLE);
+	SG_ADD(&minib_iter, "mb_iter", "number of iterations", MS_NOT_AVAILABLE);
```
vigsterkr: i wonder if we should have a better name for this. as this is gonna be the name of the param when you do .put(...) and .get(). maybe `iteration`?

PR #4319, opened 2018-05-31
Refactor linear machine api

vinx13:
[WIP]
Adapt machine api to fit + predict
This need some heavy refactor, so let's start from linear machines and iterate a few times.
The idea is to keep both old and new apis working and then remove old apis gradually. 
`train_machine(CFeatures*, CLabels*)` is added to machines that need labels. The old method, `train_machine(CFeatures*)` will redirect to the new api and pass `m_labels` as the labels argument. In this way, the old api (`set_labels` + `train(CFeatures*)`) still works. 

## Roadmap 
- [x] Add `void fit(CFeatures*)`, `void fit(CFeatures*, CLabels*)` to `CMachine`
- [x] Redirect `bool train_machine(CFeatures*)` to `void train_machine(CFeatures*, CLabels*)` in LinearMachine
- [ ] Calling new api in unittests
- [ ] Move api redirection to `LinearMachine` as a base class method after the new api works for all `LinearMachine` subclasses
- [ ] Make `bool train_machine(CFeatures*)` and `bool train_machine_templated(CFeatures*)` return void (The latter one should be easier as it is added last year and we haven't used in many places)

Known issues for moving to const methods:
* ref counting
* labels factory (e.g. `regression_labels`) doesn't accept const args


src/shogun/classifier/AveragedPerceptron.cpp
```diff
@@ -40,59 +32,51 @@ void CAveragedPerceptron::init()
 	SG_ADD(&learn_rate, "learn_rate", "Learning rate.", MS_AVAILABLE);
 }
 
-bool CAveragedPerceptron::train_machine(CFeatures* data)
+void CAveragedPerceptron::train_machine(CFeatures* features, CLabels* labels)
```
karlnapf: dont we want this to be const parameters?

src/shogun/classifier/AveragedPerceptron.cpp
```diff
@@ -40,59 +32,51 @@ void CAveragedPerceptron::init()
...
 		COMPUTATION_CONTROLLERS
-		converged=true;
+		converged = true;
 		SG_INFO("Iteration Number : %d of max %d\n", iter, max_iter);
 
-		for (int32_t i=0; i<num_vec; i++)
+		for (auto i : range(num_vec))
 		{
-			output[i] = features->dense_dot(i, w.vector, w.vlen) + bias;
+			output[i] = dot_features->dense_dot(i, w.vector, w.vlen) + bias;
```
karlnapf: i think this could be done via `DotIterator`

src/shogun/classifier/AveragedPerceptron.cpp
```diff
@@ -103,14 +87,8 @@ bool CAveragedPerceptron::train_machine(CFeatures* data)
...
 	// calculate and set the average paramter of w, bias
-	for (int32_t i=0; i<num_feat; i++)
-		w[i]=tmp_w[i]/(num_vec*iter);
+	linalg::scale(tmp_w, w, 1.0 / (num_vec * iter));
 	bias=tmp_bias/(num_vec*iter);
 
-	SG_FREE(output);
-	SG_FREE(tmp_w);
-
 	set_w(w);
```
karlnapf: i would prefer to remove this and udpate the state vector itself inside the main loop.
@shubham808 can elaborate as he is doing similar stuff for the perceptron

src/shogun/classifier/Perceptron.cpp
```diff
@@ -44,28 +36,22 @@ CPerceptron::~CPerceptron()
 {
 }
 
-bool CPerceptron::train_machine(CFeatures* data)
+void CPerceptron::train_machine(CFeatures* features, CLabels* labels)
 {
-	ASSERT(m_labels)
+	if (!features->has_property(FP_DOT))
```
karlnapf: why not `as`? would be a bit cleaner. Training is expensive, so the additional costs shouldnt matter or?

src/shogun/machine/Machine.cpp
```diff
@@ -66,6 +66,23 @@ bool CMachine::train(CFeatures* data)
 	return result;
 }
 
+void CMachine::fit(CFeatures* features)
+{
+	REQUIRE(train(features), "Failed to fit machine %s\n", get_name());
```
karlnapf: maybe we should throw an exception rather than this boolean stuff?

vigsterkr: require failed -> sg_io->msg(MSG_ERROR, "blabla") -> https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/io/SGIO.cpp#L125

src/shogun/classifier/Perceptron.cpp
vinx13: we can provide more informative error message here, but `as` is as good

src/shogun/classifier/AveragedPerceptron.cpp
vinx13: of course. unfortunately we can't use const for the time being because there are many non-const methods that should be const logically, e.g. `get_feature_matrix`

src/shogun/machine/Machine.cpp
karlnapf: Sure but I mean, shouldn't the train method itself just throw an exception?
We don't get any context information here. If train would throw an exception, we could catch it here and then say "training failed for reason X". No?

src/shogun/classifier/AveragedPerceptron.cpp
karlnapf: I see, sure
(though get_feature_matrix can be made const without problem)

src/shogun/machine/Machine.cpp
vinx13: i'm going to just throw exceptions in `fit`. but for now train returns boolean, i keep this so that it is consistent that fit always throw exceptions when failure.

karlnapf:
Hi!
Shall we pick this up again?

vinx13:
@karlnapf I found there are too many things involved here. When you change the `Machine` base class, almost everything need to be updated. I would suggest start some small refactor first, such as `Distance` or `Kernel`. We may further discuss on irc.

src/shogun/classifier/svm/LibLinear.cpp
```diff
@@ -67,24 +67,23 @@ CLibLinear::~CLibLinear()
...
+bool CLibLinear::train_machine(CFeatures* features)
 {
+	ASSERT(m_labels);
+	train_machine(features, m_labels);
+	return true;
+}
 
-	ASSERT(m_labels)
-	init_linear_term();
+void CLibLinear::train_machine(CFeatures* features, CLabels* labels)
```
karlnapf: const possible?

karlnapf:
Nice! Let me know when we should have a look at this...

src/shogun/classifier/svm/LibLinear.cpp
vinx13: There are many non-const methods algorithm internal methods. We can either use non-const arguments here, or use `const_cast` to drop const in internal methods

vinx13: @karlnapf One problem ref counting is `IterativeMachine`. We need to increase ref count of features and labels in `init_model`

karlnapf: ah crap yes, the ref counting is non-const....
ok so we need to postpone the const making for now until we have another way to do the reference counting

src/shogun/classifier/LPBoost.cpp
```diff
@@ -100,7 +100,13 @@ float64_t CLPBoost::find_max_violator(int32_t& max_dim)
...
 bool CLPBoost::train_machine(CFeatures* data)
 {
-	ASSERT(m_labels)
+	train_machine(data, m_labels);
+	return true;
+}
+
+virtual void CLPBoost::train_machine(CFeatures* features, CLabels* labels);
+{
+	ASSERT(labels)
```
karlnapf: I think handling those asserts happens (should happen?) in the base class?

src/shogun/base/SGObject.h
```diff
@@ -567,6 +567,20 @@ class CSGObject
 			demangled_type<T>().c_str());
 		return nullptr;
 	}
+
+	template <class T> const T* as() const
```
karlnapf: you are using this below I assume when changing the signatures of some methods?

karlnapf:
Cool some progress :) Let us know how you are getting on here

src/shogun/classifier/LPBoost.cpp
vinx13: Agree, it should be in the base class. I will check if it is possible for now.

src/shogun/base/SGObject.h
vinx13: This is useful for const pointers (`const CFeatures*`). Although we decided to move to const methods later, I leave it here for possible future usage.

src/shogun/regression/LeastSquaresRegression.cpp
```diff
@@ -24,4 +18,3 @@ CLeastSquaresRegression::CLeastSquaresRegression(CDenseFeatures<float64_t>* data
 : CLinearRidgeRegression(0, data, lab)
```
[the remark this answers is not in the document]
karlnapf: @Saurabh7 this would be something for you to check/figure out? I think we have seen and solved something similar before. Unless you have ideas @vinx13 ?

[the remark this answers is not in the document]
vinx13: @karlnapf I only saw this error on CI. I cannot reproduce this locally with docker image.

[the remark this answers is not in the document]
karlnapf: Yes I remember this issue somehow with the same reproducibility issues. Maybe @Saurabh7 remembers

src/shogun/base/SGObject.h
```diff
@@ -567,6 +567,20 @@ class CSGObject
...
+	{
+		auto c = dynamic_cast<const T*>(this);
+		if (c)
+			return c;
+
+		SG_SERROR(
+			"Object of type %s cannot be converted to type %s.\n",
+			demangled_type<std::remove_pointer_t<decltype(this)>>().c_str(),
```
karlnapf: btw can we re-use code from the non-const version in here? Like one calls the other?

src/shogun/machine/FeatureDispatchCRTP.h
```diff
@@ -39,15 +39,24 @@ namespace shogun
 			switch (data->get_feature_type())
 			{
 			case F_DREAL:
-				return this_casted->template train_machine_templated<float64_t>(
+			{
+				this_casted->template train_machine_templated<float64_t>(
```
karlnapf: why this?

src/shogun/machine/IterativeMachine.h
```diff
@@ -109,7 +116,17 @@ namespace shogun
 
 		/** To be overloaded in subclasses to initialize the model for training
 		  */
-		virtual void init_model(CFeatures* data = NULL) = 0;
+		virtual void init_model(CFeatures* data = NULL)
```
karlnapf: @shubham808 you should also review/discuss this as some of your classes are changed (better API)

src/shogun/regression/LeastSquaresRegression.cpp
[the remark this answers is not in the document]
karlnapf: I assume this works now?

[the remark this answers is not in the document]
vinx13: yes, we need to explicit instantiate templates in cpp files to make it linkable

src/shogun/classifier/svm/LibLinear.cpp
```diff
@@ -178,7 +175,7 @@ bool CLibLinear::train_machine(CFeatures* data)
 	}
 	case L2R_L2LOSS_SVC:
 	{
-		fun_obj = new l2r_l2_svc_fun(&prob, Cs);
+		fun_obj = new l2r_l2_svc_fun(&prob, Cs.vector);
```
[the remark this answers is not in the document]
vinx13: using `SGVector` can make it simpler, we can get rid of `SG_MALLOC` and `SG_FREE`

src/shogun/machine/FeatureDispatchCRTP.h
karlnapf: thx

src/shogun/regression/LeastSquaresRegression.cpp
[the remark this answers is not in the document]
Saurabh7: Looks like this is solved.

Very Interesting PR btw ! So we are planning to change all APIs to `.fit ` , `.predict` ?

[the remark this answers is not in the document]
Saurabh7: Looks like this is solved.

Very Interesting PR btw ! So we are planning to change all APIs to ` .fit `, `.predict` ?

src/shogun/classifier/AveragedPerceptron.cpp
vinx13: @karlnapf I think `set_w` has been already called inside the iteration function?

src/shogun/regression/LeastSquaresRegression.cpp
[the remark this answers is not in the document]
karlnapf: Sorry I actually meant to ping @shubham808 ;)
But yes @Saurabh7 we are intending to do that

src/shogun/machine/LinearMachine.cpp
```diff
@@ -134,3 +134,52 @@ CDotFeatures* CLinearMachine::get_features()
 void CLinearMachine::store_model_features()
 {
 }
+
+// TODO: almost the same as CMachine::train, move this method to CMachine after
+// we have done refactoring other classes
+void CLinearMachine::fit(CFeatures* features, CLabels* labels)
+{
+	REQUIRE(train_require_labels(), "Labels are not needed.");
```
karlnapf: typo "not"

src/shogun/machine/LinearMachine.cpp
```diff
@@ -134,3 +134,52 @@ CDotFeatures* CLinearMachine::get_features()
...
+	/* not allowed to train on locked data */
+	REQUIRE(
+	    !m_data_locked, "(%s)::train data_lock() was called, only "
+	                    "train_locked() is possible. Call data_unlock if you "
+	                    "want to call train()\n",
+	    get_name());
+
+	auto sub = connect_to_signal_handler();
+
+	if (support_feature_dispatching())
```
karlnapf: @shubham808 maybe have a read of those changes since you wrote the original code

shubham808: this seems same as CMachine::train so it will not cause any problems but i see it is planned to be moved there after refactor

src/shogun/machine/IterativeMachine.h
shubham808: yeah init_model is taking features and labels as arguments which is more flexible :+1:

karlnapf:
Let's push this thing on  :)

gf712:
keeping this alive as I think this is still the direction we want to go in

karlnapf:
bump

PR #4320, opened 2018-05-31
state update in each iteration for perceptron

shubham808:
We want to update the weights vector in each iteration so that we remain consistent during a paused state

### DIFF
--- a/src/shogun/classifier/Perceptron.cpp
+++ b/src/shogun/classifier/Perceptron.cpp
@@ -67,12 +67,14 @@ bool CPerceptron::train_machine(CFeatures* data)
 	ASSERT(num_vec==train_labels.vlen)
 	SGVector<float64_t> output(num_vec);
 
-	auto w = SGVector<float64_t>(num_feat);
+	SGVector<float64_t> w;
 	if (m_initialize_hyperplane)
 	{
+		w = SGVector<float64_t>(num_feat);
 		set_w(w);
-		w.set_const(1.0 / num_feat);
+
 		//start with uniform w, bias=0
+		w.set_const(1.0 / num_feat);
 		bias=0;
 	}
 	else
--- a/src/shogun/classifier/Perceptron.cpp
+++ b/src/shogun/classifier/Perceptron.cpp
@@ -35,7 +35,9 @@ void CPerceptron::init()
 	max_iter = 1000;
 	learn_rate = 0.1;
 	m_initialize_hyperplane = true;
-	SG_ADD(&max_iter, "initialize_hyperplane", "Whether to initialize hyperplane.", MS_AVAILABLE);
+	SG_ADD(
+	    &m_initialize_hyperplane, "initialize_hyperplane",
+	    "Whether to initialize hyperplane.", MS_AVAILABLE);
 	SG_ADD(&max_iter, "max_iter", "Maximum number of iterations.", MS_AVAILABLE);
 	SG_ADD(&learn_rate, "learn_rate", "Learning rate.", MS_AVAILABLE);
 }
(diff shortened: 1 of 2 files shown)


src/shogun/classifier/Perceptron.cpp
```diff
@@ -35,7 +35,9 @@ void CPerceptron::init()
 	max_iter = 1000;
 	learn_rate = 0.1;
 	m_initialize_hyperplane = true;
-	SG_ADD(&max_iter, "initialize_hyperplane", "Whether to initialize hyperplane.", MS_AVAILABLE);
```
karlnapf: why the change? just by editor accident?

src/shogun/classifier/Perceptron.cpp
```diff
@@ -65,15 +67,19 @@ bool CPerceptron::train_machine(CFeatures* data)
 	ASSERT(num_vec==train_labels.vlen)
 	SGVector<float64_t> output(num_vec);
 
-	SGVector<float64_t> w = get_w();
+	auto w = SGVector<float64_t>(num_feat);
+	set_w(w);
```
karlnapf: this is bad.
You over-write the previous w in any case, which kills the feature of users being able to initialise themselves.
The setting should go into the else branch of the if statement.

src/shogun/classifier/Perceptron.cpp
```diff
@@ -65,15 +67,19 @@ bool CPerceptron::train_machine(CFeatures* data)
...
 	if (m_initialize_hyperplane)
 	{
-		w = SGVector<float64_t>(num_feat);
+		w.set_const(0);
```
karlnapf: set const to 1.0 / num_feat and them remove the `add` below

src/shogun/classifier/Perceptron.cpp
shubham808: see that its &max_iter and not &m_initialize_hyperplane

karlnapf: ah yes, correcting copy paste errors. nice!

src/shogun/classifier/Perceptron.cpp
```diff
@@ -65,15 +67,18 @@ bool CPerceptron::train_machine(CFeatures* data)
 	ASSERT(num_vec==train_labels.vlen)
 	SGVector<float64_t> output(num_vec);
 
-	SGVector<float64_t> w = get_w();
+	auto w = SGVector<float64_t>(num_feat);
```
karlnapf: The allocation should be moved into the if

karlnapf: i.e. just define it here

karlnapf: actually, why dont you just do m_w.resize(num_feats)?

shubham808: m_w is private so we will need to use get_w() everywhere instead... this causes type errors with v.add(gradient, w)

karlnapf: ok cool. But still you can move the allocation inside the if, you dont want to allocate memory if the user has provided intitialization

src/shogun/classifier/Perceptron.cpp
```diff
@@ -65,15 +67,18 @@ bool CPerceptron::train_machine(CFeatures* data)
...
 	if (m_initialize_hyperplane)
 	{
-		w = SGVector<float64_t>(num_feat);
```
karlnapf: keep this line here, and change the above to a definition

tests/unit/classifier/Perceptron_unittest.cc
```diff
@@ -59,3 +59,25 @@ TEST(Perceptron, train)
...
+
+	auto perceptron = some<CPerceptron>(features, labels);
+	perceptron->train();
+
+	auto weights = perceptron->get_w();
+
+	auto perceptron_initialized = some<CPerceptron>(features, labels);
+	perceptron_initialized->set_initialize_hyperplane(false);
+	perceptron_initialized->set_w(weights);
+	perceptron_initialized->set_max_iter(1);
```
karlnapf: yep this seems to make sense

tests/unit/classifier/Perceptron_unittest.cc
```diff
@@ -59,3 +59,25 @@ TEST(Perceptron, train)
...
+
+	perceptron_initialized->train();
+	EXPECT_EQ(perceptron_initialized->get_w(), weights);
```
karlnapf: this checks for ==, can you make it equals?

karlnapf: `EXPECT_TRUE(perceptron_initialized->get_w().equals(weights))`
== is pointer equality (should fail)

shubham808: i see

karlnapf:
Waiting for CI and then good to merge

shubham808:
any changes here ? :)

geektoni:
It seems okay from my side. @karlnapf?

PR #4324, opened 2018-06-02
Cookbook for chi_square distance

FaroukY:


doc/cookbook/source/examples/distance/chi_square.rst
```diff
@@ -0,0 +1,35 @@
+===================
+Chi Square Distance
+===================
+
+The Chi Square Distance for real valued features x and x' extends the concept of :math:`\chi^{2}` metric to negative values.
```
karlnapf: since you are using x and x' downstream, pls put them into (inline) math mode here as well

doc/cookbook/source/examples/distance/chi_square.rst
```diff
@@ -0,0 +1,35 @@
...
+This distance is calculated by the equation:
+
+.. math::
+
+    d(\bf{x},\bf{x'}) = \sum_{i=1}^{n}\frac{(x_{i}-x'_{i})^2}{|x_{i}|+|x'_{i}|} \quad \bf{x},\bf{x'} \in R^{n}
```
karlnapf: n is undefined

iglesias:
Can `RealMatrix distance_matrix_aa = d.get_distance_matrix()` be ported to something like `d.get('distance_matrix')`? @karlnapf

doc/cookbook/source/examples/distance/chi_square.rst
iglesias: We need more clarity here :-) The distance extends the concept of the metric, what does this mean? I think distance and metric are commonly used interchangeably. Explain specifically what is meant by extending to negative values.

karlnapf:
@iglesias no `get_distance_matrix()` is actually not a getter but it does compute something. We will need some hacks from @lisitsyn on registering callback functions as parameters accessible via `get` before that works

doc/cookbook/source/examples/distance/chi_square.rst
karlnapf: Can we have some updates here pls? :)

FaroukY: @iglesias @karlnapf  The chi square distance is defined as:

screenshot from 2018-06-08 15-30-10

This is for only positive values. However, the implementation in shogun is for both positive and negative values, since the denominator is (|x_i|+|y_i|) so it extends the chi square distance to the negative values too. Is this more clear?

doc/cookbook/source/examples/distance/chi_square.rst
FaroukY: Should I add "for any natural number n" ?

doc/cookbook/source/examples/distance/chi_square.rst
iglesias: To some extent. What reference are you using for this classical definition?

From I have read after checking shortly out there, a use case of this Chi-2 distance is to calculate distance between histograms. In that use case the values are positive of course. I suppose that people found other use cases where values are not necessary positive. I suspect that this dichotomy you have observed (definitions with and without the absolute value) may come from there.

It is fine for me if you want to leave the text as it is. I would not play with saying the distance extends the concept of metric to negative values since because I find it rather vague and potentially misleading.

doc/cookbook/source/examples/distance/chi_square.rst
iglesias: I think it should be fine for defining n if you just put that the features belong to R^n the first time you introduce them. By the way, for extra mathy happiness, use \mathbb with R :nerd_face:

karlnapf:
thx

PR #4326, opened 2018-06-04
Add inverse transform in transformer

vinx13:


src/interfaces/swig/Transformer.i
```diff
@@ -5,7 +5,8 @@
  */
 
 /* These functions return new Objects */
-%newobject shogun::CTransformer::transform(CFeatures*);
+%newobject shogun::CTransformer::transform(CFeatures*, bool inplace=true);
+%newobject shogun::CTransformer::inverse_transform(CFeatures*, bool inplace=true);
```
karlnapf: can we rename to `inverse` ... without `transform` ?

vigsterkr:
The idea here was about sklearnish approach

karlnapf:
I see they have that ... well, then it is probably useful. ... ;)

vigsterkr:
Min max scaled target (regression) for which u learn a model but then u wanna actually map back the prediction into the “original scale”

karlnapf:
Ah yes of course!
@vinx13 hey what about a cookbook for that use case? :)

PR #4327, opened 2018-06-04
Feature/pausing serialization test

shubham808:
[WIP] The build on appeveyor is still failing

### DIFF
--- a/src/shogun/lib/StoppableSGObject.h
+++ b/src/shogun/lib/StoppableSGObject.h
@@ -17,7 +17,7 @@ namespace shogun
 {
 #define COMPUTATION_CONTROLLERS                                                \
 	if (cancel_computation())                                                  \
-		continue;                                                              \
+		break;                                                                 \
 	pause_computation();
 
 	/**
--- a/tests/unit/base/trained_model_serialization_test.cc
+++ b/tests/unit/base/trained_model_serialization_test.cc
@@ -16,6 +16,7 @@
 #include <shogun/machine/Machine.h>
 
 using namespace shogun;
+using namespace std;
 
 extern LinearTestEnvironment* linear_test_env;
 extern MultiLabelTestEnvironment* multilabel_test_env;
@@ -87,8 +88,7 @@ protected:
 	}
 
 	bool serialize_machine(
-	    CMachine* cmachine, std::string& filename,
-	    bool store_model_features = false)
+	    CMachine* cmachine, std::string& filename, bool store_model_features)
 	{
 		std::string class_name = cmachine->get_name();
 		filename = "shogun-unittest-trained-model-serialization-" + class_name +
@@ -103,6 +103,30 @@ protected:
 		SG_FREE(file);
 
 		return save_success;
+	}
+
+	bool test_serialization(bool store_model_features = false)
+	{
+		machine->set_labels(train_labels);
+		machine->train(train_feats);
+
+		auto predictions = wrap<CLabels>(machine->apply(test_feats));
+
+		std::string filename;
+		if (!serialize_machine(machine, filename, store_model_features))
+			return false;
+
+		if (!deserialize_machine(deserialized_machine, filename))
+			return false;
+
+		auto deserialized_predictions =
+		    wrap<CLabels>(deserialized_machine->apply(test_feats));
+
+		set_global_fequals_epsilon(1e-7);
+		if (!predictions->equals(deserialized_predictions))
+			return false;
+		set_global_fequals_epsilon(0);
+		return true;
 	}
 
 	bool deserialize_machine(CMachine* cmachine, std::string filename)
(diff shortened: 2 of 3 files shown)


karlnapf:
Ok, let's split this up a bit to make it easier to review.
I suggest a first patch where you take micheles commit and just add another one that makes things work.

Then you modify this PR here to be rebased against the above.
Then we can isolate the appveyor stuff in the other PR.

I will just review the pausing code here for now

src/shogun/lib/StoppableSGObject.h
```diff
@@ -17,7 +17,7 @@ namespace shogun
 {
 #define COMPUTATION_CONTROLLERS                                                \
 	if (cancel_computation())                                                  \
-		continue;                                                              \
+		break;                                                                 \
```
karlnapf: yeah +1 as we discussed

tests/unit/base/trained_model_serialization_unittest.cc
```diff
@@ -0,0 +1,209 @@
...
+{
+	EXPECT_TRUE(this->test_serialization());
+}
+
+template <class T>
+class StoppedMachineSerialization : public TrainedModelSerializationFixture<T>
+{
+};
+
+TYPED_TEST_CASE(StoppedMachineSerialization, StoppableMachineTypes);
```
karlnapf: since this is a test for pausing, maybe we should refactor the test file a bit and either create another file for those tests, or rename the test to something like "all_models_tests" (or some better name).

tests/unit/base/trained_model_serialization_unittest.cc
```diff
@@ -0,0 +1,209 @@
...
+
+TYPED_TEST(StoppedMachineSerialization, Test)
+{
+	int i = 0;
+	std::function<bool()> callback = [&i]() {
+		if (i >= 1)
```
karlnapf: so we are stopping after 1 iteration? seems ok, but can you explain why that choice?

tests/unit/base/trained_model_serialization_unittest.cc
```diff
@@ -0,0 +1,209 @@
...
+		{
+			get_global_signal()->get_subscriber()->on_next(SG_BLOCK_COMP);
+			return true;
+		}
+		i++;
+		return false;
+	};
+
+	this->machine->set_callback(callback);
```
karlnapf: this seems quite clean now, nice

tests/unit/base/trained_model_serialization_unittest.cc
```diff
@@ -0,0 +1,209 @@
...
+	EXPECT_TRUE(this->test_serialization());
```
karlnapf: ok just so I understand. This trains for 1 iteration, then applys and remembers results, then saves-loads and applys, ensures results are the same

tests/unit/base/trained_model_serialization_unittest.cc
karlnapf: is `StoppableMachineTypes` instantiated somewhere? cant find it

tests/unit/base/trained_model_serialization_unittest.cc.py
```diff
@@ -112,30 +107,54 @@ def entry(templateFile, input_file, config_file):
...
+        'CNativeMulticlassMachine': 'Machine',
+        'CLinearMulticlassMachine': 'Machine',
+        'CKernelMachine': 'KernelMachine',
+        'CKernelMulticlassMachine': 'KernelMachine',
+    }
+
+    test_machines_map = {
+        'Machine': [],
+        'KernelMachine': [],
+        'StoppableMachine' : []
```
karlnapf: ok!

tests/unit/base/trained_model_serialization_test.cc.py
```diff
@@ -35,6 +35,17 @@
     'CMulticlassLibSVM' # error C <= 0
 ]
 
+# Algorithms that can be prematurely stopped
```
shubham808: this is list is limited by 2 things 
1. The algorithms that can be systematically tested for serialization
2. The algorithms that do not converge before starting second iteration.

For now I have also not included any KernelMachine algorithms here

tests/unit/base/trained_model_serialization_test.cc
```diff
@@ -105,6 +109,30 @@ class TrainedModelSerializationFixture : public ::testing::Test
 		return save_success;
 	}
 
+	bool test_serialization(bool store_model_features = false)
```
karlnapf: did you yet rebase against micheles merged patch? Surprised to still see this in here

shubham808: yes i did :) this only the part i want to add

karlnapf: why do we need to add this? wasn't there something similar already?

tests/unit/base/trained_model_serialization_test.cc.py
```diff
@@ -35,6 +35,17 @@
...
+STOPPABLE = [
```
karlnapf: I have another Q here: rather than whitelisting stoppable machines, we should backlist the ones that we want to exclude from testing. Or? Otherwise new machines will not be added to this list, and thus, not be tested

tests/unit/base/trained_model_serialization_test.cc
shubham808: removes repetition of the same code this is a cleanup from the previous version :)

tests/unit/base/trained_model_serialization_test.cc.py
shubham808: i think its better to whitelist because blacklisting will mean having to blacklist the same algorithms over again for each test.. for example: we will have to add the algos in the ```IGNORE``` list here as well (since they cannot be initialized in the same way)

karlnapf: sure there will be different blacklists for different tests, but that is fine. Some machines will be serializable but not stoppable and vice versa. As said before, if we have to explicitly add new machines to this list in order to have them tested, this will be forgotten and so test coverage will suffer. Think 3 years ahead. We can maybe have a global ignore list and a test specific one to avoid too much redundancy

shubham808: you are right i did miss that....

shubham808:
as discussed we will not be needing the test right now. I have updated the PR with some small cleanups.

tests/unit/base/trained_model_serialization_test.cc
```diff
@@ -16,6 +16,7 @@
 #include <shogun/machine/Machine.h>
 
 using namespace shogun;
+using namespace std;
```
karlnapf: i usually prefer to avoid this, but it is ok here as well

PR #4330, opened 2018-06-06
helper to serialize all machines into Ascii

shubham808:
[WIP] on pausing we can serialize a model directly

### DIFF
--- a/src/shogun/lib/StoppableSGObject.cpp
+++ b/src/shogun/lib/StoppableSGObject.cpp
@@ -6,6 +6,7 @@
 
 #include <rxcpp/rx-lite.hpp>
 #include <shogun/base/init.h>
+#include <shogun/io/SerializableAsciiFile.h>
 #include <shogun/lib/Signal.h>
 #include <shogun/lib/StoppableSGObject.h>
 
@@ -40,6 +41,16 @@ rxcpp::subscription CStoppableSGObject::connect_to_signal_handler()
 	return get_global_signal()->get_observable()->subscribe(subscriber);
 }
 
+void CStoppableSGObject::serialize_machine()
+{
+	std::string filename = this->get_name();
+	filename += ".txt";
+	auto file = new CSerializableAsciiFile(filename.c_str(), 'w');
+	this->save_serializable(file);
+	file->close();
+	SG_FREE(file);
+}
+
 void CStoppableSGObject::set_callback(std::function<bool()> callback)
 {
 	m_callback = callback;
@@ -60,6 +71,7 @@ void CStoppableSGObject::on_next()
 void CStoppableSGObject::on_pause()
 {
 	m_pause_computation_flag.store(true);
+	this->serialize_machine();
 	on_pause_impl();
 	resume_computation();
 }
--- a/src/shogun/lib/StoppableSGObject.cpp
+++ b/src/shogun/lib/StoppableSGObject.cpp
@@ -41,14 +41,11 @@ rxcpp::subscription CStoppableSGObject::connect_to_signal_handler()
 	return get_global_signal()->get_observable()->subscribe(subscriber);
 }
 
-void CStoppableSGObject::serialize_machine()
+void CStoppableSGObject::serialize_machine(std::string filename)
 {
-	std::string filename = this->get_name();
-	filename += ".txt";
-	auto file = new CSerializableAsciiFile(filename.c_str(), 'w');
-	this->save_serializable(file);
-	file->close();
-	SG_FREE(file);
+	m_serialized_file = new CSerializableAsciiFile(filename.c_str(), 'w');
+	this->save_serializable(m_serialized_file);
+	m_serialized_file->close();
 }
 
 void CStoppableSGObject::set_callback(std::function<bool()> callback)
@@ -71,7 +68,7 @@ void CStoppableSGObject::on_next()
 void CStoppableSGObject::on_pause()
 {
 	m_pause_computation_flag.store(true);
-	this->serialize_machine();
+	this->serialize_machine(std::string(this->get_name()) + ".txt");
 	on_pause_impl();
 	resume_computation();
 }
--- a/src/shogun/lib/StoppableSGObject.h
+++ b/src/shogun/lib/StoppableSGObject.h
@@ -78,6 +78,9 @@ namespace shogun
 		}
 
 	protected:
+		/** Serialize machine to an Ascii file with a default file name */
+		void serialize_machine();
+
 		/** connect the machine instance to the signal handler */
 		rxcpp::subscription connect_to_signal_handler();
 
--- a/src/shogun/lib/StoppableSGObject.h
+++ b/src/shogun/lib/StoppableSGObject.h
@@ -79,7 +79,7 @@ namespace shogun
 
 	protected:
 		/** Serialize machine to an Ascii file with a default file name */
-		void serialize_machine();
+		void serialize_machine(std::string filename);
 
 		/** connect the machine instance to the signal handler */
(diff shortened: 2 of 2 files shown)


src/shogun/lib/StoppableSGObject.cpp
```diff
@@ -40,6 +41,16 @@ rxcpp::subscription CStoppableSGObject::connect_to_signal_handler()
 	return get_global_signal()->get_observable()->subscribe(subscriber);
 }
 
+void CStoppableSGObject::serialize_machine()
```
karlnapf: This is for the user chossing to serialize during pause right?

shubham808: yes but we could use it anywhere else too... easy serialization method

shubham808:
yes thats better

src/shogun/lib/StoppableSGObject.cpp
```diff
@@ -40,6 +41,13 @@ rxcpp::subscription CStoppableSGObject::connect_to_signal_handler()
 	return get_global_signal()->get_observable()->subscribe(subscriber);
 }
 
+void CStoppableSGObject::serialize_machine(std::string filename)
+{
+	m_serialized_file = new CSerializableAsciiFile(filename.c_str(), 'w');
```
shubham808: we will need to discuss this :)

src/shogun/lib/StoppableSGObject.cpp
```diff
@@ -60,6 +68,7 @@ void CStoppableSGObject::on_next()
 void CStoppableSGObject::on_pause()
 {
 	m_pause_computation_flag.store(true);
+	this->serialize_machine(std::string(this->get_name()) + ".txt");
```
geektoni: This is good but in my opinion, the serialization should be placed into ``on_pause_impl()`` instead :) devs may want to provide other behaviours for algorithm instead of serializing them by default (and to do it now they would have to override the ``on_pause`` method, which is no good).

src/shogun/lib/StoppableSGObject.cpp
```diff
@@ -40,6 +41,13 @@ rxcpp::subscription CStoppableSGObject::connect_to_signal_handler()
 	return get_global_signal()->get_observable()->subscribe(subscriber);
 }
 
+void CStoppableSGObject::serialize_machine(std::string filename)
```
vigsterkr: i'm not so sure if get it why there's a specific interface serializing a "machine". especially when `CStoppableSGObject` is actually not even a machine, but an SGObject

src/shogun/lib/StoppableSGObject.cpp
vigsterkr: i second what @geektoni says above.... why do you want to automatically serialize a machine and why that name and why in a format of ascii and why extension of .txt and what happens if the file already exists...

src/shogun/lib/StoppableSGObject.cpp
karlnapf: yep, the name is misleading, just call it "serialize".

src/shogun/lib/StoppableSGObject.cpp
karlnapf: second @geektoni comment

Also agree with @vigsterkr The ascii format should not be compulsory, as discussed in another thread. Same goes for filename. Also some error handling will be needed

shubham808: @vigsterkr the idea is to serialize in Ascii as default when pausing but the user can always change that, infact even specify what type he wants before pausing :) I am still working on it atm

src/shogun/lib/StoppableSGObject.cpp
vigsterkr: @shubham808 it's not about the name. it's more about that `CStoppableSGObject` is a `CSGObject` which already have a serialization functionality in it `CSGObject::save_serializable` `CSGObject::load_serializable` which should be sufficient to save any SGObject-like

src/shogun/lib/StoppableSGObject.cpp
vigsterkr: mmm sorry but if i want to pause my training i dont want an object to touch my FS and do any sort of IO without my explicit request. i dont understand why a pause should trigger IO and serialization of the object itself. if one pauses a procedure, then it just pauses the procedure, having these implicit default actions like writing to file etc. is not something one would like to have - developer perspective. if she/he wants to serialize the paused object, she can do it by explicitly calling the function. it's there

karlnapf: Use case: user runs algorithm and pauses it. Then she want to serialize the intermediate state AND continue training afterwards. Not possible via explicitly calling serializing as that requires training to be stopped fully, and thus it cannot be continued afterwards. So serialization during a pause has to be an option for the user (in addition to do nothing, continue)

karlnapf: Furthermore, we dont want to interface with the user to ask for filenames and formats in the pause. This is why there needs to be a way to specify potential intermediate training serialization format and name for the user before training starts

src/shogun/lib/StoppableSGObject.cpp
karlnapf: IMO this can be a member method as it will depend on the state of the stoppable object, namely the file that might be used for dumping the temporary state

src/shogun/lib/StoppableSGObject.cpp
vigsterkr: why wouldn't you be able to serialize a paused object/method?

src/shogun/lib/StoppableSGObject.cpp
vigsterkr: why a file at all? why not a bytestream? i really dont get this concept atm

src/shogun/lib/StoppableSGObject.cpp
karlnapf: you can, but ,as said above, you have to stop training for that

src/shogun/lib/StoppableSGObject.cpp
karlnapf: because we dont have a bytestream right now. Once we have a generic interface for serializing into whatever, we will use it here

vigsterkr: CFile* ?

karlnapf: yes it should be CFile. @shubham808 has not updated the PR yet after our discussion.

src/shogun/lib/StoppableSGObject.cpp
karlnapf: BTW @shubham808 the file should not be automatically serialized on pausing. Just if the user decides to.

vigsterkr:
```
m = new Machine();

thread 1: m.train();

thread 2: SIG to thread1 pause
thread 2 calls m.save_serializable(...)
does whatever

thread 2 sends a callback to m to continue
```

why wouldn't this work?

karlnapf: It works. The idea here was more that, in a typical use case, a user is just living in one thread. So the idea was to make his life easier and offer him a button to press that dumps the intermediate machine somewhere. It is a smallish detail

vigsterkr: smallish detail but this adds things to the places and functionality places where it does not belong. atm i really dont see this as a good design to go ahead with this as is.

shubham808:
we dont need this anymore.

PR #4331, opened 2018-06-06
Allow SGIO to throw custom exception types

vinx13:
* separate formatting and printing in SGIO::message
* define new macros `REQUIRE_E`, `SG_THROW`, `SG_STHROW` that support using custom exception types


vinx13:
appveyor timeouted, travis build for python failed because too many warnings were thrown

karlnapf:
How are those macros used? Could you write a unit test?

PR #4333, opened 2018-06-07
make xvalidation always leave state of machine/features invariant

karlnapf:
previously, this was only done for number of threads >1 which lead to inconsistent behaviour.
@FaroukY this will fix your problem with the failing integration test from octave (which currently is enforced to use a single thread only)

As long as our features are used in a non-read-only fashion, this unfortunately is the only way of guaranteeing correct results in xvalidation. As soon as we have
''CMachine::train(const CFeatures*, const CLabels*)`, this can be updated in a way that features are shared across folds. And then we can also re-enable multithreaded xvaldiation without getting a huge number of race conditions and memory errors.
@vinx13 we are counting on you ;)


karlnapf:
#4334

PR #4335, opened 2018-06-09
iterative machine

shubham808:
[WIP] This is to see if our concept with IterativeMachine will work. The PR is not yet complete.


src/shogun/classifier/Perceptron.cpp
```diff
@@ -77,42 +52,30 @@ bool CPerceptron::train_machine(CFeatures* data)
...
+		{
+			m_complete = false;
+			const auto gradient = learn_rate * true_label;
+			bias += gradient;
+			v.add(gradient, w);
 		}
-
-		iter++;
 	}
-
```
shubham808: for error checking after loop ops maybe we should add an end_training() method to IterativeMachine

src/shogun/classifier/Perceptron.cpp
```diff
@@ -17,22 +17,14 @@
 using namespace shogun;
 
 CPerceptron::CPerceptron()
-: CLinearMachine()
```
shubham808: Done as a special case of LinearMachine for now (as discussed)

src/shogun/machine/IterativeLinearMachine.cpp
```diff
@@ -0,0 +1,49 @@
...
+bool CIterativeLinearMachine::train(CDotFeatures* f, CLabels* l)
+{
+   init_model(f, l);
+   continue_train(f, l);
+   return m_complete;
+}
+
+void CIterativeLinearMachine::continue_train(CDotFeatures* f, CLabels* l)
+{
+  //resume_computation();
```
shubham808: this will need to be a method to revive computation controllers after cancel flag

shubham808: after fixing this a test like [here](https://gist.github.com/shubham808/c1446f80c0aca3bcfe695c507efed803) should work.

src/shogun/machine/IterativeLinearMachine.h
```diff
@@ -0,0 +1,71 @@
...
+       virtual ~CIterativeLinearMachine();
+       
+       /** Train machine
+        * @params f features
+        * @params l labels 
+        * @return whether training was successful
+        */
+       virtual bool train(CDotFeatures* f, CLabels* l);
+       
+       void register_params();
```
geektoni: I guess that this guy should be `protected`, right?

src/shogun/machine/IterativeLinearMachine.cpp
```diff
@@ -0,0 +1,49 @@
...
+   while (m_current_iteration < m_max_iterations && !m_complete)
+   {
+       COMPUTATION_CONTROLLERS
+       iteration(f, l);
+       m_current_iteration++;
+   }
+}
```
geektoni: You could call here, right before the end of ``continue_train``, the method  `reset_computation_variables` (see [here](https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/lib/StoppableSGObject.h#L85)). So that each time the `continue_train` method finishes, you can call it again.

src/shogun/machine/IterativeLinearMachine.h
```diff
@@ -0,0 +1,71 @@
...
+      /* default constructor */  
+       CIterativeLinearMachine();
+       virtual ~CIterativeLinearMachine();
+       
+       /** Train machine
+        * @params f features
+        * @params l labels 
+        * @return whether training was successful
+        */
+       virtual bool train(CDotFeatures* f, CLabels* l);
```
geektoni: We could enforce constness here with this new design, right? This can be done for all the method below (`continue_train`, `iteration`). @karlnapf

src/shogun/machine/IterativeLinearMachine.cpp
geektoni: See my comment below :)

src/shogun/machine/IterativeLinearMachine.cpp
shubham808: yeah that works

src/shogun/classifier/Perceptron.cpp
```diff
@@ -77,42 +52,30 @@ bool CPerceptron::train_machine(CFeatures* data)
...
+}
+
+void CPerceptron::iteration(CDotFeatures* f, CLabels* l)
+{
+	m_complete = true;
+	SGVector<float64_t> w = get_w();
+	SGVector<float64_t> output(f->get_num_vectors());
+	auto iter_train_labels = binary_labels(l)->get_int_labels().begin();
+	auto iter_output = output.begin();
+	for (const auto& v : DotIterator(f))
```
geektoni: Aren't we doing a sort of batch learning here? This ``iteration`` method will run over all the features so, if we premature stop ``continue_train`` and then we call it again, the `iteration` method will just run over all the features `f` from the start. This would work if we provided to ``iteration`` a set of new features/labels each time ``continue_train`` is called. @karlnapf @shubham808

karlnapf: for Perceptron this should be ok as we are touching *all* data in *every* iteration. So there is no problem with continued algorithms starting to touch the data from beginning again

src/interfaces/swig/Classifier.i
```diff
@@ -36,6 +36,7 @@
 %rename(SVMLightOneClass) CSVMLightOneClass;
 #endif //USE_SVMLIGHT
 %rename(LinearMachine) CLinearMachine;
+%rename(IterativeLinearMachine) CIterativeLinearMachine;
```
karlnapf: this shouldn't be in swig

src/shogun/classifier/Perceptron.cpp
```diff
@@ -17,22 +17,14 @@
 using namespace shogun;
 
 CPerceptron::CPerceptron()
-: CLinearMachine()
+: CIterativeLinearMachine()
 {
 	init();
 }
 
-CPerceptron::CPerceptron(CDotFeatures* traindat, CLabels* trainlab)
```
karlnapf: yes, let's just remove those ctors

src/shogun/classifier/Perceptron.cpp
```diff
@@ -17,22 +17,14 @@
...
-: CLinearMachine()
-{
-	init();
-	set_features(traindat);
-	set_labels(trainlab);
-}
-
 void CPerceptron::init()
 {
-	max_iter = 1000;
```
karlnapf: keep in mind this will blow the integration test

src/shogun/classifier/Perceptron.h
```diff
@@ -69,20 +63,14 @@ class CPerceptron : public CLinearMachine
 
 		/** @return object name */
 		virtual const char* get_name() const { return "Perceptron"; }
-
+		
+		virtual void init_model(CDotFeatures* f, CLabels* l);
```
karlnapf: protected! both of them

src/shogun/machine/IterativeLinearMachine.h
karlnapf: Let's follow shogun's design for now which is to take features here

karlnapf: (and change this later)

src/shogun/machine/IterativeLinearMachine.h
```diff
@@ -0,0 +1,71 @@
...
+       
+       void register_params();
+       
+       /* Continue Training
```
karlnapf: This needs more documentation about when it fails and what it is expected to do. Furthermore, as discussed, it should be in CMachine, virtual, and throwing an error (SG_NOTIMPLEMENTED)

src/shogun/machine/IterativeLinearMachine.h
```diff
@@ -0,0 +1,71 @@
...
+        * @params f features
+        * @params l labels 
+        */
+       void continue_train(CDotFeatures* f, CLabels* l);
+       
+       
+       virtual const char* get_name() const { return "IterativeLinearMachine"; }
```
karlnapf: as this class is abstract, no need for this

src/shogun/machine/IterativeLinearMachine.cpp
```diff
@@ -0,0 +1,49 @@
...
+#include <shogun/lib/config.h>
+
+#include <shogun/lib/common.h>
+#include <shogun/machine/IterativeLinearMachine.h>
+
+using namespace shogun;
+
+CIterativeLinearMachine::CIterativeLinearMachine() : CLinearMachine()
+{
+ m_current_iteration = 0;
```
karlnapf: I think no need for "register params" actually, as we only have one ctor

src/shogun/machine/IterativeLinearMachine.cpp
```diff
@@ -0,0 +1,49 @@
...
+ m_complete = false;
+ register_params();
+}
+
+CIterativeLinearMachine::~CIterativeLinearMachine()
+{
```
karlnapf: SG_UNREF(m_features) (once those are added)

src/shogun/machine/IterativeLinearMachine.cpp
```diff
@@ -0,0 +1,49 @@
...
+}
+
+bool CIterativeLinearMachine::train(CDotFeatures* f, CLabels* l)
+{
+   init_model(f, l);
+   continue_train(f, l);
+   return m_complete;
+}
+
+void CIterativeLinearMachine::continue_train(CDotFeatures* f, CLabels* l)
```
karlnapf: continue_train is actually a better name! :)

src/shogun/machine/IterativeLinearMachine.cpp
```diff
@@ -0,0 +1,49 @@
...
+ register_params();
+}
+
+CIterativeLinearMachine::~CIterativeLinearMachine()
+{
+}
+
+bool CIterativeLinearMachine::train(CDotFeatures* f, CLabels* l)
+{
+   init_model(f, l);
```
karlnapf: you need to set the iteration counter to 0 here, and the completed flag to false

src/shogun/machine/IterativeLinearMachine.cpp
```diff
@@ -0,0 +1,49 @@
...
+   continue_train(f, l);
+   return m_complete;
```
karlnapf: this is not supposed to indicate completeness, but success.
So and "continue" should return a bool

src/shogun/machine/IterativeLinearMachine.cpp
```diff
@@ -0,0 +1,49 @@
...
+ m_current_iteration = 0;
+ m_complete = false;
+ register_params();
+}
+
+CIterativeLinearMachine::~CIterativeLinearMachine()
+{
+}
+
+bool CIterativeLinearMachine::train(CDotFeatures* f, CLabels* l)
```
karlnapf: if you are overloading the base class method I think you need to use CFeatures* here

src/shogun/machine/IterativeLinearMachine.cpp
```diff
@@ -0,0 +1,49 @@
...
+       COMPUTATION_CONTROLLERS
+       iteration(f, l);
+       m_current_iteration++;
+   }
+}
+
+void CIterativeLinearMachine::register_params()
+{
+  SG_ADD(&m_current_iteration, "current_iteration", "current iteration", MS_AVAILABLE);
+  SG_ADD(&m_max_iterations, "max_iteration", "maximum iteration", MS_AVAILABLE);
```
karlnapf: type in description.

model selection should only be available for max_iterations, not for the other two flags

tests/unit/base/trained_model_serialization_test.cc.py
```diff
@@ -15,7 +15,7 @@
     'CMultitaskL12LogisticRegression', 'CMultitaskLeastSquaresRegression',
     'CMultitaskTraceLogisticRegression', 'CMultitaskClusteredLogisticRegression',
     'CLatentSVM', 'CLatentSOSVM', 'CDomainAdaptationSVMLinear',
-    'CLinearLatentMachine',
+    'CLinearLatentMachine','CPerceptron',
```
karlnapf: why this?

src/shogun/machine/IterativeLinearMachine.cpp
shubham808: i see yeah

tests/unit/classifier/Perceptron_unittest.cc
```diff
@@ -68,16 +72,16 @@ TEST(Perceptron, custom_hyperplane_initialization)
...
 	auto weights = perceptron->get_w();
 
-	auto perceptron_initialized = some<CPerceptron>(features, labels);
+	auto perceptron_initialized = some<CPerceptron>();
 	perceptron_initialized->set_initialize_hyperplane(false);
 	perceptron_initialized->set_w(weights);
 	perceptron_initialized->set_max_iter(1);
 
-	perceptron_initialized->train();
+	perceptron_initialized->train(features, labels);
```
karlnapf: pls use the base class method here (this means you need to manually set labels before training)

src/shogun/classifier/Perceptron.cpp
shubham808: this was temp.. we will remove it

tests/unit/base/trained_model_serialization_test.cc.py
shubham808: because is was not being initialized earlier in the same way... should be fine when we overload train_machine and use members instead of args for features

src/shogun/machine/IterativeLinearMachine.cpp
shubham808: how about we just use the ```m_features``` member already present in LinearMachine ?

src/shogun/classifier/Perceptron.cpp
```diff
@@ -78,44 +60,34 @@ bool CPerceptron::train_machine(CFeatures* data)
...
-	while (!converged && iter < max_iter)
+}
+
+void CPerceptron::iteration()
+{
+	bool converged = true;
+	SGVector<float64_t> w = get_w();
+	SGVector<float64_t> output(features->get_num_vectors());
+
+	auto labels = binary_labels(m_labels)->get_int_labels();
```
shubham808: Suprisingly this works but doing .begin() in the same line does not.

tests/unit/classifier/Perceptron_unittest.cc
```diff
@@ -68,16 +72,68 @@ TEST(Perceptron, custom_hyperplane_initialization)
...
-	perceptron_initialized->set_max_iter(1);
+	perceptron_initialized->put<index_t>("max_iteration", 1);
+	perceptron_initialized->set_labels(labels);
 
-	perceptron_initialized->train();
+	perceptron_initialized->train(features);
 	EXPECT_TRUE(perceptron_initialized->get_w().equals(weights));
 }
+
+TEST(Perceptron, prematurely_stop)
```
shubham808: we will need to merge [this](https://github.com/shogun-toolbox/shogun/pull/4327/files#diff-5334f04115d44684a239e44b3ce21fbaL20) before we can run this test otherwise it will just time out in an infinite loop.

src/shogun/machine/IterativeLinearMachine.cpp
karlnapf: no, pls use another member. We want to move towards not storing features in other classes. Call it `m_continue_features` or so

src/shogun/classifier/Perceptron.cpp
karlnapf: needs investigation!

tests/unit/classifier/Perceptron_unittest.cc
karlnapf: on it

examples/meta/src/binary/perceptron.sg
```diff
@@ -11,7 +11,7 @@ Labels labels_test = labels(f_labels_test)
 #![create_features]
 
 #![create_instance]
-Machine perceptron = machine("Perceptron", labels=labels_train, learn_rate=1.0, max_iter=1000)
+Machine perceptron = machine("Perceptron", labels=labels_train, learn_rate=1.0, max_iteration=1000)
```
karlnapf: max_iterations if you rename it (plural)

src/shogun/classifier/Perceptron.cpp
karlnapf: i was actually wrong, if the number is the same this is fine

src/shogun/machine/IterativeLinearMachine.h
```diff
@@ -0,0 +1,68 @@
...
+		virtual bool continue_train();
+
+	protected:
+		/* Single Iteration implemented in an iterative model
+		 * @params f features
+		 * @params l labels
+		 */
+		virtual void iteration() = 0;
+
+		/* init the model
```
karlnapf: Explain the purpose of the method, I.e. Can be overloaded by sublcasses for custom initialisation purposes

src/shogun/machine/IterativeLinearMachine.h
```diff
@@ -0,0 +1,68 @@
...
+		 * @params f features
+		 * @params l labels
+		 */
+		virtual void init_model(CFeatures* data = NULL) = 0;
+
+		/* Maximum Iterations */
+		index_t m_max_iterations;
```
karlnapf: maybe this should be int64

src/shogun/machine/Machine.h
```diff
@@ -333,6 +333,19 @@ class CMachine : public CStoppableSGObject
...
+		/** Continue Training
+		 *
+		 * NOT IMPLEMENTED!
+		 *
+		 * @return whether training was successful
+		 */
+		virtual bool continue_train()
+		{
+			SG_ERROR(
+			    "continue_train is not yet implemented for %s!\n", get_name());
```
karlnapf: Just "Training continuation not supported.\n"

src/shogun/classifier/Perceptron.cpp
shubham808: agreed

src/shogun/classifier/Perceptron.cpp
```diff
@@ -78,44 +61,34 @@ bool CPerceptron::train_machine(CFeatures* data)
...
-	auto pb = SG_PROGRESS(range(max_iter));
-	//loop till we either get everything classified right or reach max_iter
-	while (!converged && iter < max_iter)
+}
+
+void CPerceptron::iteration()
+{
+	bool converged = true;
+	SGVector<float64_t> w = get_w();
+	SGVector<float64_t> output(m_continue_features->get_num_vectors());
```
shubham808: i think this is not so efficient. @karlnapf ?

src/shogun/machine/IterativeLinearMachine.h
```diff
@@ -0,0 +1,68 @@
...
+		  *iterations of training loop.
+		  */
+		virtual void iteration() = 0;
+
+		/** Can be overloaded in subclasses to initialize the model for training
+		  */
+		virtual void init_model(CFeatures* data = NULL) = 0;
+
+		/** Maximum Iterations */
+		int32_t m_max_iterations;
```
shubham808: i think int32_t should be fine or ? also swig has trouble in using "put" with int64_t

src/shogun/classifier/Perceptron.cpp
```diff
@@ -17,56 +17,39 @@
...
+void CPerceptron::init_model(CFeatures* data)
 {
-	ASSERT(m_labels)
-
 	if (data)
 	{
 		if (!data->has_property(FP_DOT))
 			SG_ERROR("Specified features are not of type CDotFeatures\n")
-		set_features((CDotFeatures*) data);
+		m_continue_features = data;
```
karlnapf: whenever you store a reference to a CSGObject, you need to  `SG_REF` it, (and `SG_UNREF` it in the destructor if it is a member)

src/shogun/classifier/Perceptron.cpp
```diff
@@ -17,56 +17,39 @@
...
-bool CPerceptron::train_machine(CFeatures* data)
+void CPerceptron::init_model(CFeatures* data)
 {
-	ASSERT(m_labels)
-
 	if (data)
 	{
 		if (!data->has_property(FP_DOT))
 			SG_ERROR("Specified features are not of type CDotFeatures\n")
-		set_features((CDotFeatures*) data);
```
karlnapf: I think you dont want to delete this.

src/shogun/classifier/Perceptron.cpp
karlnapf: yes, you are here re-allocating the vector in every iterations. This is not good and we need to find a way to recycle things. As we said we will "communicate" over member variables, I guess this needs to be moved there.

src/shogun/classifier/Perceptron.cpp
```diff
@@ -78,44 +61,34 @@ bool CPerceptron::train_machine(CFeatures* data)
...
-		w = get_w();
-	}
-	auto pb = SG_PROGRESS(range(max_iter));
-	//loop till we either get everything classified right or reach max_iter
-	while (!converged && iter < max_iter)
+}
+
+void CPerceptron::iteration()
+{
+	bool converged = true;
```
karlnapf: could we set this to false up here and then set it to true if it is converged rather than the other way around?
Why does this even have to be a local variable? Can't we use the member directly? (and modify it if needed only)

src/shogun/classifier/Perceptron.cpp
```diff
@@ -17,56 +17,39 @@
 
 using namespace shogun;
 
-CPerceptron::CPerceptron()
-: CLinearMachine()
+CPerceptron::CPerceptron() : CIterativeLinearMachine()
 {
 	init();
```
karlnapf: I think since we only have a single (default) constructor, there is no need for this init method, but rather copy its contents into the ctor directly

src/shogun/classifier/Perceptron.h
```diff
@@ -72,28 +60,21 @@ class CPerceptron : public CLinearMachine
...
-		 *
-		 * @param data training data (parameter can be avoided if distance or
-		 * kernel-based classifiers are used and distance/kernels are
-		 * initialized with train data)
-		 *
-		 * @return whether training was successful
-		 */
-		virtual bool train_machine(CFeatures* data=NULL);
+		void init();
+		/** init the model before training begins */
```
karlnapf: pls only document these methods in the base class (properly for developers, with instructions that overloading this methods allows for certain things). No need to re-document them here

src/shogun/classifier/Perceptron.h
```diff
@@ -72,28 +60,21 @@ class CPerceptron : public CLinearMachine
...
+		virtual void init_model(CFeatures* data);
+		/** perform a single iteration over features and update weights */
```
karlnapf: same here, we would only overload the docs if something special happens inside the method (which it doesnt for perceptron)

src/shogun/machine/IterativeLinearMachine.h
```diff
@@ -0,0 +1,68 @@
...
+#include <shogun/lib/SGVector.h>
+#include <shogun/lib/common.h>
+#include <shogun/machine/LinearMachine.h>
+
+namespace shogun
+{
+	class CFeatures;
+	class CLabels;
+
+	class CIterativeLinearMachine : public CLinearMachine
```
karlnapf: some docs describing that this class implements a model whose training can be stopped, and in particular resumed, at any time

src/shogun/machine/IterativeLinearMachine.h
```diff
@@ -0,0 +1,68 @@
...
+
+		/** Train classifier
+		  *
+		  * @param data training data (parameter can be avoided if distance or
+		  * kernel-based classifiers are used and distance/kernels are
+		  * initialized with train data)
+		  *
+		  * @return whether training was successful
+		  */
+		virtual bool train_machine(CFeatures* data = NULL);
```
karlnapf: train_machine is protected btw. Be careful when inheriting to respect the base class structure!

src/shogun/machine/IterativeLinearMachine.h
```diff
@@ -0,0 +1,68 @@
...
+
+		/** Continue Training
```
karlnapf: good doc!

src/shogun/machine/IterativeLinearMachine.h
```diff
@@ -0,0 +1,68 @@
...
+		  * Only possible if a previous CMachine::train call was prematurely
+		  * stopped.
+		  * Throws an error otherwise
+		  *
+		  * @return whether continue training was successful
+		  */
+		virtual bool continue_train();
+
+	protected:
+		/** Can be overloaded by sublcasses to implement custom single
```
karlnapf: "To be overloaded by subclasses to implement a single iteration in the main training loop.".... it is mandatory ;)

src/shogun/machine/IterativeLinearMachine.h
karlnapf: put can be extended easily, but maybe you are right and 2^31 is enough ;)

src/shogun/machine/Machine.h
```diff
@@ -333,6 +333,18 @@ class CMachine : public CStoppableSGObject
 			return false;
 		}
 
+		/** Continue Training
```
karlnapf: This need proper documentation pls here as well, this is user facing.

src/shogun/machine/IterativeLinearMachine.cpp
```diff
@@ -0,0 +1,53 @@
...
+{
+	m_current_iteration = 0;
+	m_complete = false;
+	init_model(data);
+	return continue_train();
+}
+
+bool CIterativeLinearMachine::continue_train()
+{
+	reset_computation_variables();
```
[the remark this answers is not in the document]
geektoni: ``reset_computation_variables`` reset the flags used to prematurely stop the computations (see the `StoppableSGObject` class). If you do not call this and you have prematurely stopped the execution, the next time you will call `train`/`continue_train` those methods won't work because of COMPUTATION_CONTROLLERS (since the flag is still set to false).

tests/unit/classifier/Perceptron_unittest.cc
```diff
@@ -68,16 +72,69 @@ TEST(Perceptron, custom_hyperplane_initialization)
...
+	auto features = wrap(env->get_features_train());
+	auto labels = wrap(env->get_labels_train());
+	auto test_features = wrap(env->get_features_test());
+	auto test_labels = wrap(env->get_labels_test());
+
+	auto perceptron = some<CPerceptron>();
+	perceptron->set_labels(labels);
+	perceptron->train(features);
+
+	auto results = wrap(perceptron->apply(test_features))
```
karlnapf: I think in c++, we can do this a bit simpler
`auto results = perceptron->apply(test_featatures)->as<CBinaryLabels>()->get_labels()`

src/shogun/machine/IterativeLinearMachine.h
geektoni: Yes, I think this is fine here.

tests/unit/classifier/Perceptron_unittest.cc
```diff
@@ -68,16 +72,69 @@ TEST(Perceptron, custom_hyperplane_initialization)
...
+		iter++;
+		return false;
+	};
+	perceptron_stop->set_callback(callback);
+
+	perceptron_stop->put<bool>("compute_bias", false);
+	perceptron_stop->set_labels(labels);
+	perceptron_stop->train(features);
+
+	auto results_itermediate = wrap(perceptron_stop->apply(test_features))
```
karlnapf: see above

tests/unit/classifier/Perceptron_unittest.cc
```diff
@@ -68,16 +72,69 @@ TEST(Perceptron, custom_hyperplane_initialization)
...
+		{
+			get_global_signal()->get_subscriber()->on_next(SG_BLOCK_COMP);
+			return true;
+		}
+		iter++;
+		return false;
+	};
+	perceptron_stop->set_callback(callback);
+
+	perceptron_stop->put<bool>("compute_bias", false);
```
karlnapf: there is no setter? (fine if no)

tests/unit/classifier/Perceptron_unittest.cc
karlnapf: btw `wrap` is a hack we need for the meta examples, but not need to use it in c++

karlnapf: @lisitsyn @vigsterkr somehow the API didn't get simpler with all the latest additions :D

vigsterkr: yeah the wrapping is something that ought to go sooner or later ... one would hope :)

karlnapf: IMO, I would keep the c++ typed API in the unit tests.
`wrap(perceptron->apply_binary(test_features))->get_labels()`

At least simpler than old school (SG_UNREF etc)

tests/unit/classifier/Perceptron_unittest.cc
```diff
@@ -68,16 +72,69 @@ TEST(Perceptron, custom_hyperplane_initialization)
...
+
+TEST(Perceptron, prematurely_stop)
+{
+	auto env = linear_test_env->getBinaryLabelData();
+	auto features = wrap(env->get_features_train());
+	auto labels = wrap(env->get_labels_train());
+	auto test_features = wrap(env->get_features_test());
+	auto test_labels = wrap(env->get_labels_test());
+
+	auto perceptron = some<CPerceptron>();
```
karlnapf: dont we need to set a number of iterations here?

src/shogun/machine/IterativeLinearMachine.cpp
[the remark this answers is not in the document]
karlnapf: I see. Maybe rename and document

tests/unit/classifier/Perceptron_unittest.cc
```diff
@@ -68,16 +72,69 @@ TEST(Perceptron, custom_hyperplane_initialization)
...
+	perceptron_stop->set_callback(callback);
+
+	perceptron_stop->put<bool>("compute_bias", false);
+	perceptron_stop->set_labels(labels);
+	perceptron_stop->train(features);
+
+	auto results_itermediate = wrap(perceptron_stop->apply(test_features))
+	                               ->get<SGVector<float64_t>>("labels");
+
+	EXPECT_FALSE(results.equals(results_itermediate));
```
karlnapf: I am confused why this would be the same: one perceptron was stopped after one iterations, and the other one was trained until convergence no?

src/shogun/machine/IterativeLinearMachine.cpp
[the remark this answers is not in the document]
karlnapf: or maybe I should rather read :)

src/shogun/classifier/Perceptron.cpp
shubham808: we can do directly with member but i think we talked about updating members in the end of the iteration ?

tests/unit/classifier/Perceptron_unittest.cc
shubham808: no there is not

src/shogun/machine/IterativeLinearMachine.h
shubham808: right yeah

tests/unit/classifier/Perceptron_unittest.cc
shubham808: this is a different test

tests/unit/classifier/Perceptron_unittest.cc
shubham808: I train one perceptron for reference result then i trained another to stop after one iteration. the results_intermediate should not be equal to results. but when we continue_train we will converge properly and results_complete will be equal to results.

tests/unit/classifier/Perceptron_unittest.cc
shubham808: @karlnapf there is a setter for this i have just not used it here

src/shogun/classifier/Perceptron.cpp
karlnapf: The converged flag can be set to true by the iteration method. So there is no need to have a variable. Just set it to true when it is appropriate, and leave it unchanged otherwise

tests/unit/classifier/Perceptron_unittest.cc
karlnapf: I see. So this is about testing that one can stop a machine, then continue, and then the results match the process that didnt stop in the middle. OK!

BTW I think the `EXPECT_FALSE(results.equals(results_itermediate));` should be remove as this is not always necessarily the case. The last check is the one we care about.

You can check the status of the flags
m_converged == false
m_current_iteration == whatever

karlnapf: pls name the test appropriately
stop_and_continue_same_as_no_stop or something :)

src/shogun/classifier/Perceptron.cpp
```diff
@@ -78,44 +58,34 @@ bool CPerceptron::train_machine(CFeatures* data)
...
-		converged=true;
-		auto iter_train_labels = train_labels.begin();
-		auto iter_output = output.begin();
-		for (const auto& v : DotIterator(features))
+		const auto true_label = *(iter_train_labels++);
+		auto& predicted_label = *(iter_output++);
+
+		predicted_label = v.dot(w) + bias;
+
+		if (CMath::sign<float64_t>(predicted_label) != true_label)
```
shubham808: the loop here is such that i think we will need a local variable to check when m_complete needs to be changed( that is check if we never enter the if statement here) if that is so i think the converged bool works fine or?

karlnapf: i dont understand, see below

src/shogun/classifier/Perceptron.cpp
```diff
@@ -78,44 +58,34 @@ bool CPerceptron::train_machine(CFeatures* data)
...
-			predicted_label = v.dot(w) + bias;
-
-			if (CMath::sign<float64_t>(predicted_label) != true_label)
-			{
-				converged = false;
-				const auto gradient = learn_rate * true_label;
-				bias += gradient;
-				v.add(gradient, w);
-			}
+			converged = false;
```
karlnapf: why not just use the member here? then ....

karlnapf: there is no need for this statement

karlnapf: just put an `else` branch and set converged (the member) to true in there. done. no need for a variable

karlnapf: ok as discussed, ignore my comments on this :)

src/shogun/machine/IterativeLinearMachine.cpp
```diff
@@ -0,0 +1,54 @@
...
+	auto pb = SG_PROGRESS(range(m_max_iterations));
+	while (m_current_iteration < m_max_iterations && !m_complete)
+	{
+		COMPUTATION_CONTROLLERS
+		iteration();
+		m_current_iteration++;
+		pb.print_progress();
+	}
```
karlnapf: could you print a warning here if max iterations is reached but the complete flag is not yet set (like the one in perceptron)

shubham808: with the end_training method or?

src/shogun/machine/IterativeLinearMachine.cpp
```diff
@@ -0,0 +1,69 @@
...
+	pb.complete();
+
+	if (m_complete)
+		SG_INFO(
+		    "%s algorithm converged after %d iterations.\n", this->get_name(),
+		    m_current_iteration)
+	else
+		SG_WARNING(
+		    "%s algorithm did not converge after %d iterations.\n",
+		    this->get_name(), m_current_iteration)
```
shubham808: i used m_current_iteration here so that even if it is prematurely stopped the warning shows correct iterations.

src/shogun/machine/IterativeLinearMachine.cpp
```diff
@@ -0,0 +1,69 @@
...
+	}
+	pb.complete();
+
+	if (m_complete)
+		SG_INFO(
+		    "%s algorithm converged after %d iterations.\n", this->get_name(),
+		    m_current_iteration)
+	else
+		SG_WARNING(
+		    "%s algorithm did not converge after %d iterations.\n",
```
karlnapf: I would remove "algorithm"

karlnapf: did not converge after the maximum number of %d iterations

src/shogun/machine/IterativeLinearMachine.cpp
karlnapf: Question: I guess we dont want the warning to be shown if the user stopped prematurely. Only if the algorithm actually did that number of iterations, see what I mean?

src/shogun/machine/IterativeLinearMachine.h
```diff
@@ -0,0 +1,67 @@
...
+		/** To be overloaded by sublcasses to implement custom single
+		  * iterations of training loop.
+		  */
+		virtual void iteration() = 0;
+
+		/** To be overloaded in subclasses to initialize the model for training
+		  */
+		virtual void init_model(CFeatures* data = NULL) = 0;
+
+		/** Can be overloaded in subclasses to show more information */
```
karlnapf: and/or clean up temporary states

tests/unit/classifier/Perceptron_unittest.cc
```diff
@@ -68,16 +72,65 @@ TEST(Perceptron, custom_hyperplane_initialization)
...
+	auto features = wrap(env->get_features_train());
+	auto labels = wrap(env->get_labels_train());
+	auto test_features = wrap(env->get_features_test());
+	auto test_labels = wrap(env->get_labels_test());
+
+	auto perceptron = some<CPerceptron>();
+	perceptron->set_labels(labels);
+	perceptron->train(features);
+
+	auto results = perceptron->apply_binary(test_features)->get_labels();
```
karlnapf: you need to keep the reference here and UNREF it, otherwise you will have a memory error
(most getters in shogun increase the reference counter on returning objects)

src/shogun/machine/IterativeLinearMachine.cpp
shubham808: we dont want warning in case of premature stop ?

karlnapf: This means, only showiung the warning if complete is false and m_current_iteration == max_iterations

src/shogun/classifier/Perceptron.h
```diff
@@ -71,29 +59,20 @@ class CPerceptron : public CLinearMachine
...
 
 	protected:
 		/** learning rate */
 		float64_t learn_rate;
-		/** maximum number of iterations */
-		int32_t max_iter;
 
 	private:
+		/** Store output during training */
+		SGVector<float64_t> m_output;
```
karlnapf: as an example for the end_training method, you could overload it here and then free the vector...i.e. overwrite it with empty vector

src/shogun/classifier/Perceptron.cpp
```diff
@@ -17,56 +17,36 @@
...
+void CPerceptron::init_model(CFeatures* data)
 {
-	ASSERT(m_labels)
-
 	if (data)
 	{
 		if (!data->has_property(FP_DOT))
 			SG_ERROR("Specified features are not of type CDotFeatures\n")
 		set_features((CDotFeatures*) data);
+		SG_REF(data);
```
karlnapf: more logical would be to put the SG_REF after the assignment and to REF m_continue_features

karlnapf: actually, you need to do this
```
SG_REF(data)
SG_UNREF(m_continue_features)
m_continue_features = data
```

src/shogun/classifier/Perceptron.cpp
```diff
@@ -78,44 +58,39 @@ bool CPerceptron::train_machine(CFeatures* data)
...
-		SG_INFO("Perceptron algorithm converged after %d iterations.\n", iter)
-	else
-		SG_WARNING("Perceptron algorithm did not converge after %d iterations.\n", max_iter)
+	m_complete = converged;
+}
 
-	return converged;
+void CPerceptron::end_training()
+{
+	m_output = SGVector<float64_t>(0);
```
karlnapf: no need for a 0 btw

src/shogun/machine/IterativeLinearMachine.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+CIterativeLinearMachine::~CIterativeLinearMachine()
+{
+	SG_UNREF(m_continue_features);
+}
+
+bool CIterativeLinearMachine::train_machine(CFeatures* data)
+{
+	m_current_iteration = 0;
+	m_complete = false;
```
karlnapf: also reset the continue features here

src/shogun/machine/IterativeLinearMachine.cpp
```diff
@@ -0,0 +1,71 @@
...
+		    m_current_iteration)
+	else if (!m_complete && m_current_iteration == m_max_iterations)
+	{
+		SG_WARNING(
+		    "%s did not converge after the maximum number of %d iterations.\n",
+		    this->get_name(), m_current_iteration)
+
+		this->end_training();
+	}
+	return m_complete;
```
karlnapf: reset continue features here if algorithm completed naturally

src/shogun/machine/IterativeLinearMachine.cpp
karlnapf: (and store the given features)

src/shogun/machine/IterativeLinearMachine.cpp
karlnapf: I think we should have one. But that should be different. It should just be an SG_INFO message that logs that the user chose to stop X prematurely after Y iterations

shubham808: but the CSignal will show something if we press CTRL+C

src/shogun/machine/IterativeLinearMachine.cpp
shubham808: we do not need to unref when we leave continue_train() right ?

src/shogun/machine/IterativeLinearMachine.cpp
```diff
@@ -0,0 +1,71 @@
...
+	    "Current Iteration of training", MS_NOT_AVAILABLE);
+	SG_ADD(
+	    &m_max_iterations, "max_iterations", "Maximum number of Iterations",
+	    MS_AVAILABLE);
+	SG_ADD(&m_complete, "complete", "Convergence status", MS_NOT_AVAILABLE);
+}
+
+CIterativeLinearMachine::~CIterativeLinearMachine()
+{
+	SG_UNREF(m_continue_features);
```
shubham808: what we could do inside train_machine is: 
sgref(m_continue_features)
continue->train()
if(naturally stopped)
sgunref(m_continue_features)
this will cover our cases and we do not need to unref in destructor then.... will that work ?

src/shogun/machine/IterativeLinearMachine.cpp
karlnapf: Yeah or that. Just something that informs the user that he aborts after X iterations ...

src/shogun/machine/IterativeLinearMachine.cpp
karlnapf: you need to ref when you store a reference, and unref when you forget it

src/shogun/machine/IterativeLinearMachine.cpp
karlnapf: sgref data is not needed if you dont store the pointer (subclasses or base classes might and will take care of this)

calling train machine should first forget old continue features, so the unref seems valid

You will have to try all cases in unit tests to really figure it out, it is sometimes hard to foresee

karlnapf: but the rule is simple: 
* always ref if you store a pointer
* always unref in dtor and when you overwrite it
* if overwriting a pointer, first ref the new one and only then unref the old one (in case the obj is the same you avoid a deletion of the obj this way)

src/shogun/machine/IterativeMachine.h
```diff
@@ -0,0 +1,130 @@
...
+		    name, param,                                                       \
+		    AnyParameterProperties(                                            \
+		        description, ms_available, GRADIENT_NOT_AVAILABLE));           \
+		if (ms_available)                                                      \
+			this->m_model_selection_parameters->add(param, name, description); \
+	}
+
+	/** @brief Class IterativeLinearMachine implements an iterative model
+	 * whose training can be prematurely stopped, and in particular
+	 * resumed, anytime.
```
karlnapf: minor: be resumed

src/shogun/machine/IterativeMachine.h
```diff
@@ -0,0 +1,130 @@
...
+	class CIterativeMachine : public T
+	{
+	public:
+		/** Default constructor */
+		CIterativeMachine() : T()
+		{
+			m_current_iteration = 0;
+			m_complete = false;
+
+			SG_ADD8(
```
karlnapf: There should be no 8 in here. We will have to make this work otherwise somehow

src/shogun/machine/IterativeMachine.h
```diff
@@ -0,0 +1,130 @@
...
+#include <shogun/lib/SGVector.h>
+#include <shogun/lib/common.h>
+#include <shogun/machine/LinearMachine.h>
+
+namespace shogun
+{
+	class CFeatures;
+	class CLabels;
+
+#define SG_ADD8(param, name, description, ms_available)                        \
```
karlnapf: Rather than defining this here, couldn't we just add this `this->` to the original SG_ADD?
@vigsterkr @lisitsyn

src/shogun/machine/IterativeMachine.h
```diff
@@ -0,0 +1,130 @@
...
+	/** @brief Class IterativeLinearMachine implements an iterative model
+	 * whose training can be prematurely stopped, and in particular
+	 * resumed, anytime.
+	 */
+	template <class T>
+	class CIterativeMachine : public T
+	{
+	public:
+		/** Default constructor */
+		CIterativeMachine() : T()
```
karlnapf: What is cool is that we avoid the usual mixin factory macro mess since we only have a default constructor anyways

src/shogun/machine/IterativeMachine.h
```diff
@@ -0,0 +1,130 @@
...
+
+		virtual ~CIterativeMachine()
+		{
+		}
+
+		/** Continue Training
+		  *
+		  * Only possible if a previous CMachine::train call was prematurely
+		  * stopped.
+		  * Throws an error otherwise
```
karlnapf: @vigsterkr would we have an own type of exception here?

src/shogun/machine/IterativeMachine.h
```diff
@@ -0,0 +1,130 @@
...
+		  *
+		  * @return whether continue training was successful
+		  */
+		virtual bool continue_train()
+		{
+			this->reset_computation_variables();
+			while (m_current_iteration < m_max_iterations && !m_complete)
```
karlnapf: progress bar is missing

src/shogun/machine/IterativeMachine.h
```diff
@@ -0,0 +1,130 @@
...
+			{
+				if (this->cancel_computation())
+					break;
+				this->pause_computation();
```
karlnapf: I dont understand why we would repeatedly call pause_computation here. Can you explain?

src/shogun/machine/IterativeMachine.h
lisitsyn: Wow, what's 8 here? :)

I very much prefer getting rid of macro at all, if possible.

tests/unit/classifier/Perceptron_unittest.cc
lisitsyn: I think adding a few helper functions like `do_not_compute_bias(perceptron_stop)` would make code more readable.

tests/unit/classifier/Perceptron_unittest.cc
```diff
@@ -68,16 +72,67 @@ TEST(Perceptron, custom_hyperplane_initialization)
...
+	};
+	perceptron_stop->set_callback(callback);
+
+	perceptron_stop->put<bool>("compute_bias", false);
+	perceptron_stop->set_labels(labels);
+	perceptron_stop->train(features);
+
+	EXPECT_FALSE(perceptron_stop->get<bool>("complete"));
+
+	if (!perceptron_stop->get<bool>("complete"))
```
lisitsyn: This might be `is_complete(perceptron_stop)`

src/shogun/machine/IterativeMachine.h
shubham808: the 8 is to diffrentiate between SG_ADD and the macro here... not anything meaningful really

src/shogun/machine/IterativeMachine.h
shubham808: yes I will need to use the progress object here

src/shogun/machine/IterativeMachine.h
shubham808: this is the implementation of COMPUTATION_CONTROLLERS macro but with ```this->``` appended

src/shogun/machine/Machine.h
karlnapf: I.e. explain what the method does (from a user perspective)

Something in the lines of:
"Continue training.

This method can be used to continue a previously prematurely stopped training call.
It throws an error if the machine does not support this feature.
"

No talking about overloading (that would be developer facing)

src/shogun/machine/Machine.h
```diff
@@ -333,6 +333,21 @@ class CMachine : public CStoppableSGObject
...
+		 * This is overloaded in IterativeLinearMachine
+		 * to resume training of a model that was prematurely stopped.
+		 *
+		 * NOT IMPLEMENTED!
+		 *
+		 * @return whether training was successful
+		 */
+		virtual bool continue_train()
+		{
+			SG_ERROR("Training continuation not supported by this model.\n");
```
karlnapf: Should we have our own exception type for such things?
NotImplementedException
@vigsterkr 
SG_NOTIMPLEMENTED is more for "sorry not YET implemented, i.e. incomplete implementation)", as opposed to. "This machine doesnt support this feature"

tests/unit/classifier/Perceptron_unittest.cc
karlnapf: why an if here?
you could just make the above an `ASSERT_FALSE`

karlnapf:
I like the mixin, and also there wont be any of the instantiation problems (like we have for trees) because Perceptron is actually not templated..

For now, we left `iteration()` without parameters (as of yet, labels and features are stored in CMachine and CLinearMachine anyways). Once we refactor things, the design patterns can be applied here.

Asking for a review @vigsterkr @lisitsyn @iglesias

src/shogun/machine/IterativeMachine.h
karlnapf: same question as for the SG_ADD, can't we just add the "this" to the macro?

src/shogun/machine/IterativeMachine.h
iglesias: Any potential side-effects adding it? Maybe we can add it and run the tests, examples, etc, to get a feeling of it.

karlnapf: acutually no. Just talked with @vigsterkr and we agreed that we should just add the "this" in the orginal definition of the macro

src/shogun/classifier/Perceptron.cpp
```diff
@@ -78,44 +53,39 @@ bool CPerceptron::train_machine(CFeatures* data)
...
 	{
-		COMPUTATION_CONTROLLERS
-		converged=true;
-		auto iter_train_labels = train_labels.begin();
-		auto iter_output = output.begin();
-		for (const auto& v : DotIterator(features))
+		const auto true_label = *(iter_train_labels++);
+		auto& predicted_label = *(iter_output++);
+
+		predicted_label = v.dot(w) + bias;
```
karlnapf: I have a final question here:
`predicted_label` is a scalar, and it seems that the previous iterations ouputs are not used in the current iteration. Why do we need to store the vector of them?
Couldnt we just not have the `m_output` and instead just define a local variable in the iteration method?

src/shogun/classifier/Perceptron.h
```diff
@@ -71,29 +59,21 @@ class CPerceptron : public CLinearMachine
...
 		/** whether the hyperplane should be initialized in train_machine
 		 *
-		 * this allows to initialize the hyperplane externally using set_w and set_b
+		 * this allows to initialize the hyperplane to externally set
```
karlnapf: this is a bit misleading: 
"Flag that determines whether hyper-plane is initialised by the algorithm, or not. The latter allows users to initialize the algorithm by manually setting weights and bias before training."

src/shogun/machine/IterativeMachine.h
```diff
@@ -0,0 +1,127 @@
...
+#include <shogun/lib/SGVector.h>
+#include <shogun/lib/common.h>
+#include <shogun/machine/LinearMachine.h>
+
+namespace shogun
+{
+	class CFeatures;
+	class CLabels;
+
+	/** @brief Class IterativeLinearMachine implements an iterative model
```
karlnapf: You can remove the class name
"Mix-in class that implements an iterative model whose ..."

src/shogun/machine/IterativeMachine.h
```diff
@@ -0,0 +1,127 @@
...
+				m_current_iteration++;
+				pb.print_progress();
+			}
+			pb.complete();
+
+			if (m_complete)
+				SG_SINFO(
+				    "%s converged after %d iterations.\n", this->get_name(),
+				    m_current_iteration)
+			else if (!m_complete && m_current_iteration == m_max_iterations)
```
karlnapf: maybe >= just in case

src/shogun/machine/Machine.h
```diff
@@ -333,6 +333,22 @@ class CMachine : public CStoppableSGObject
 			return false;
 		}
 
+		/** Continue Training
+		 *
+		 * This method can be used to continue a previously prematurely stopped
+		 * training call.
+		 * It throws an error if the machine does not support this feature.
+		 *
+		 * NOT IMPLEMENTED!
```
karlnapf: Could you remove this?
subclasses will inherit the docstring ...

src/shogun/machine/IterativeMachine.h
```diff
@@ -0,0 +1,127 @@
...
+		/** Returns convergence status */
+		bool is_complete()
+		{
+			return m_complete;
+		}
+
+		/** Continue Training
+		  *
+		  * Only possible if a previous CMachine::train call was prematurely
+		  * stopped.
```
karlnapf: remove docstring here so that the one of CMachine is used (and the CMachine one should be polished a bit, see my comment)

src/shogun/machine/Machine.h
karlnapf: we can leave it as is for now

karlnapf: or @shubham808 you can have a look at what is the apprpriate exception
@vinx13 could you explain how to use custom ones?

vinx13: you can use `SG_THROW(exception_type, message...)`, but that's in `feature/transformers` branch that haven't been merged

iglesias:
Implementing the IterativeMachine concept with a mix-in seems like a reasonably idea. Nice job!

shubham808:
Cheers !

src/shogun/machine/Machine.h
```diff
@@ -333,6 +333,22 @@ class CMachine : public CStoppableSGObject
 			return false;
 		}
 
+		/** Continue Training
+		 *
+		 * This method can be used to continue a prematurely stopped
+		 * call to CMachine::train.
+		 * It throws an error if the machine does not support this feature.
+		 *
+		 * NOT IMPLEMENTED!
```
iglesias: As mentioned in another earlier comment from Heiko iirc, is "NOT IMPLEMENTED!" "inherited" in the sublasses documentation (where this method may be actually implemented)?

karlnapf: it is only if sublcasses dont write a new docstring.
But as said, I would remove this here

PR #4337, opened 2018-06-15
Added a new example and cookbook for LeastSquareRegression [ci skip]

FaroukY:
As discussed with Viktor, commit is with ci skip until the cookbook writeup and example are both okay then I'll focus on running the tests and integration tests.

### DIFF
--- a/doc/cookbook/source/examples/regression/least_square_regression.rst
+++ b/doc/cookbook/source/examples/regression/least_square_regression.rst
@@ -1,0 +1,46 @@
+=======================
+Least Square Regression
+=======================
+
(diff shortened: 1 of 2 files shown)


doc/cookbook/source/examples/regression/least_square_regression.rst
```diff
@@ -0,0 +1,46 @@
+=======================
```
karlnapf: Not so sure about this thing as we already have a cookbook for ridge regression

http://shogun.ml/examples/latest/examples/regression/linear_ridge_regression.html

PR #4343, opened 2018-06-23
Lazy any

lisitsyn:


lisitsyn:
@karlnapf I'm in progress yet, just to let you know something is going on ;)

tests/unit/lib/Any_unittest.cc
```diff
@@ -636,3 +636,10 @@ TEST(Any, reset_array2d_reference)
 	EXPECT_EQ(obj->ref_count(), 1);
 	SG_UNREF(obj);
 }
+
+TEST(Any, lazy_simple)
+{
+	auto v = 9;
+	auto any = make_any<int>([=]() { return v; });
+	EXPECT_EQ(any.as<int32_t>(), v);
```
lisitsyn: @karlnapf this thing works now, but it definitely needs more thorough testing

lisitsyn:
@karlnapf an interesting question is whether we want to support the following assignments:

- `std::function<T()> = T`
- `T = std::function<T()>`

tests/unit/lib/Any_unittest.cc
karlnapf: shall we try to test a slightly more elaborate example of registering a member function that returns a constant or something?

karlnapf:
idk, imo we only need the functions to returns stuff. But you are the guru here :D

tests/unit/lib/Any_unittest.cc
lisitsyn: Yeah I can add one.

tests/unit/lib/Any_unittest.cc
```diff
@@ -636,3 +641,40 @@ TEST(Any, reset_array2d_reference)
...
+
+TEST(Any, lazy_assignment_from)
+{
+	auto v = 3;
+	auto any = make_any(0);
+	EXPECT_THROW(
+	    any = make_any<int>([=]() { return v; }), TypeMismatchException);
+}
+
+TEST(Any, lazy_member_function)
```
iglesias: What do you think about a test case that asserts laziness?

lisitsyn: Good idea

karlnapf: Also, we need to keep in mind equals/serialization should exclude lazy computations.

karlnapf: i.e. we need some kind of markup. Otherwise these things will be equaled/cloned/serialized as well....or is that intended? Costs computation and I am not sure it is worth it

karlnapf: ah, clone_from will be a problem as it attempts to write into the member

lisitsyn: That's a bunch of questions :)

- Serialization is tricky. It seems the 'serializator' itself should know
- We can explicitly throw something when trying to do equals
- Clone is cheap but it won't work correctly. It would use the original object

karlnapf: I think then in clone and equals, I would just skip all registered function anys

karlnapf:
https://ci.appveyor.com/project/vigsterkr/shogun/build/2756#L27945

src/shogun/lib/any.h
```diff
@@ -328,9 +336,46 @@ namespace shogun
...
 
+		template <>
+		inline const Empty& value_of(Empty const* ptr)
+		{
+			static Empty empty;
+			return empty;
+		}
+
+		template <class T>
+		struct has_result_type
```
karlnapf: lol :D

tests/unit/lib/Any_unittest.cc
```diff
@@ -89,6 +89,11 @@ class Object : public CSGObject
 	{
 		return new Object();
 	}
+
+	int computed_member() const
+	{
+		return 90210;
```
karlnapf: 42?

karlnapf:
All fixed? :D

lisitsyn:
Not yet, there is a bug in has_type_fallback. And some lazyness flag visible outside is due.

karlnapf:
add the lazyness flag you lazy flagger

lisitsyn:
I am lazy to add the lazyness

tests/unit/lib/Any_unittest.cc
```diff
@@ -678,3 +678,16 @@ TEST(Any, lazy_member_function)
...
+{
+	Any any;
+	EXPECT_THROW(
+	    any.clone_from(make_any<int>([=]() { return 111; })), std::logic_error);
+	EXPECT_TRUE(any.cloneable());
+	EXPECT_TRUE(any.visitable());
+	any = make_any<int>([=]() { return 222; });
+	EXPECT_FALSE(any.cloneable());
+	EXPECT_FALSE(any.visitable());
+	EXPECT_THROW(any.visit(nullptr), std::logic_error);
```
karlnapf: sweet!

src/shogun/base/SGObject.cpp
```diff
@@ -698,6 +698,13 @@ CSGObject* CSGObject::clone()
 		const BaseTag& tag = it.first;
 		const Any& own = it.second.get_value();
 
+		if (!own.cloneable())
+		{
+			SG_SDEBUG(
+			    "Skipping clone of %s::%s of type %s.\n", this->get_name(),
+			    tag.name().c_str(), own.type().c_str());
```
karlnapf: continue!

src/shogun/base/SGObject.cpp
```diff
@@ -993,6 +1000,16 @@ bool CSGObject::equals(const CSGObject* other) const
...
 		const BaseTag& tag = it.first;
 		const Any& own = it.second.get_value();
+
+		if (!own.visitable())
+		{
+			SG_SDEBUG(
+			    "Skipping comparison of %s::%s of type %s as it is "
+			    "non-visitable.\n",
+			    this->get_name(), tag.name().c_str(), own.type().c_str());
+			continue;
```
karlnapf: same here, i would move this down and only say "skipping" to avoid duplicate code

src/shogun/base/SGObject.cpp
lisitsyn: Yeah but since it is here such thing might be useful anyway

PR #4345, opened 2018-06-25
Add string features factory

vinx13:


src/shogun/util/factory.h
```diff
@@ -85,6 +85,26 @@ namespace shogun
...
+	    CFile* file, EAlphabet alpha = DNA,
+	    EPrimitiveType primitive_type = PT_CHAR)
+	{
+		REQUIRE(file, "No file provided.\n");
+		CFeatures* result = nullptr;
+
+		switch (primitive_type)
+		{
+		case PT_CHAR:
+			result = new CStringFeatures<char>(file, alpha);
```
karlnapf: wasnt the idea that we dont want a `string_features` factory, buy instead somehow deal with this inside the `features` factory?

karlnapf: this is something that could be done in the factory, or? (taling a sneak preview)

karlnapf:
Merging, as this allows us to continue adding more examples, and the `string_features` factory can very easily be grep replaced later once there is an appropriate factory.

PR #4348, opened 2018-06-28
Cleanup and optimize transformers

vinx13:
* Use linalg in NormOne
* Optimize PruneVarSubMean

### DIFF
--- a/src/shogun/preprocessor/PruneVarSubMean.cpp
+++ b/src/shogun/preprocessor/PruneVarSubMean.cpp
@@ -102,7 +102,7 @@ void CPruneVarSubMean::cleanup()
 SGMatrix<float64_t>
 CPruneVarSubMean::apply_to_matrix(SGMatrix<float64_t> matrix)
 {
-	ASSERT(m_initialized)
+	REQUIRE(m_initialized, "Transformer has not been fitted.\n");
 
 	auto num_vectors = matrix.num_cols;
 	auto result = matrix;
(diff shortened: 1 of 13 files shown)


karlnapf:
This needs a rebase

karlnapf:
Btw whenever you make something linalg, pls ensure there is some test coverage of the changes lines, so that we at least know he results didn’t change in the integration test examples

tests/unit/preprocessor/NormOne_unittest.cc
```diff
@@ -0,0 +1,69 @@
...
+
+	EXPECT_EQ(feats->get_num_vectors(), num_vectors);
+
+	for (auto i : range(num_vectors))
+	{
+		SGVector<float64_t> v = feats->get_feature_vector(i);
+		EXPECT_EQ(v.vlen, num_features);
+		for (auto j : range(v.vlen))
+		{
+			EXPECT_DOUBLE_EQ(v[j], data[num_features * i + j] / norm[i]);
```
karlnapf: cant we use `operator()` here?

tests/unit/preprocessor/NormOne_unittest.cc
```diff
@@ -0,0 +1,69 @@
...
+
+TEST_F(NormOne, apply_to_vector)
+{
+	transformer->fit(feats);
+
+	for (auto i : range(num_vectors))
+	{
+		SGVector<float64_t> v = feats->get_feature_vector(i);
+		auto result = transformer->apply_to_feature_vector(v);
+		EXPECT_EQ(result.vlen, num_features);
```
karlnapf: `ASSERT_EQ` to avoid segfault

src/shogun/preprocessor/PruneVarSubMean.cpp
```diff
@@ -102,7 +102,7 @@ void CPruneVarSubMean::cleanup()
 SGMatrix<float64_t>
 CPruneVarSubMean::apply_to_matrix(SGMatrix<float64_t> matrix)
 {
-	ASSERT(m_initialized)
+	REQUIRE(m_initialized, "Transformer has not been fitted.\n");
```
karlnapf: offtopic: I would love to move all those general checks to base classes so that they dont get repeated in every specialization....
any thoughts on this

vigsterkr: shouldn't we use a different exception type other than ShogunException

src/shogun/preprocessor/PruneVarSubMean.cpp
```diff
@@ -132,31 +132,18 @@ CPruneVarSubMean::apply_to_matrix(SGMatrix<float64_t> matrix)
 /// result in feature matrix
 SGVector<float64_t> CPruneVarSubMean::apply_to_feature_vector(SGVector<float64_t> vector)
 {
-	float64_t* ret=NULL;
+	REQUIRE(m_initialized, "Transformer has not been fitted.\n");
```
vigsterkr: shouldn't we use a different exception type other than ShogunException

src/shogun/transformer/Transformer.cpp
```diff
@@ -0,0 +1,20 @@
...
+{
+
+	CTransformer::CTransformer() : CSGObject()
+	{
+		SG_ADD(
+		    &m_fitted, "fitted", "Whether the transformer has been fitted.",
+		    MS_NOT_AVAILABLE);
+	}
+
+	void CTransformer::check_fitted() const
```
vigsterkr: i would rather go with the java like: `is_fitted` or `is_fit` depending which english we'd like to go with :) but let's stick to british (non north american english as in case of the exception type naming)

PR #4350, opened 2018-06-29
feature dispatching in base for CLARS

shubham808:
[WIP] This is very minimal example of how we want dispatching to work based on the gist by @karlnapf . All feature classes and types are not covered.

### DIFF
--- a/src/shogun/machine/Machine.h
+++ b/src/shogun/machine/Machine.h
@@ -25,6 +25,20 @@
 namespace shogun
 {
 
+#define TRAIN_DENSE_DISPATHER(train_templated_name) \
+virtual bool train_with_dense(CFeatures* data) \
+{  \
+  switch (data->get_feature_type()) \
+  { \
+    case F_DREAL: \
+      return train_templated_name<float64_t>(data->as<CDenseFeatures<float64_t>>()); \
+		case F_SHORTREAL: \
+      return train_templated_name<float32_t>(data->as<CDenseFeatures<float32_t>>()); \
+		case F_LONGREAL: \
+	    return train_templated_name<floatmax_t>(data->as<CDenseFeatures<floatmax_t>>()); \
+  } \
+}
+
 class CFeatures;
 class CLabels;
 
@@ -315,7 +329,6 @@ class CMachine : public CStoppableSGObject
 
 		virtual const char* get_name() const { return "Machine"; }
 
-	protected:
 		/** train machine
 		 *
 		 * @param data training data (parameter can be avoided if distance or
@@ -328,11 +341,38 @@ class CMachine : public CStoppableSGObject
 		 */
 		virtual bool train_machine(CFeatures* data=NULL)
 		{
-			SG_ERROR("train_machine is not yet implemented for %s!\n",
-					get_name());
+			bool result=false;
+			
+			switch (data->get_feature_class())
+			{
+				case C_DENSE:
+					result=train_with_dense(data);
+					break;
+				case C_STRING:
+					result=train_with_string(data);
+					break;
+				default:
+					SG_ERROR("train_machine is not yet implemented for %s!\n",
+							get_name())
+					break;
+			}
+			return result;
+		}
+		
+	protected:
+		
+		virtual bool train_with_dense(CFeatures* data)
+		{
+			SG_NOTIMPLEMENTED
 			return false;
 		}
-
+		
+		virtual bool train_with_string(CFeatures* data)
+		{
+			SG_NOTIMPLEMENTED
+			return false;
+		}
+		
 		/** Continue Training
 		 *
 		 * This method can be used to continue a prematurely stopped
--- a/src/shogun/regression/LeastAngleRegression.cpp
+++ b/src/shogun/regression/LeastAngleRegression.cpp
@@ -102,37 +102,6 @@ void CLeastAngleRegression::plane_rot(ST x0, ST x1,
 	}
 }
 
-bool CLeastAngleRegression::train_machine(CFeatures* data)
-{
-	if (!data)
-	{
-		REQUIRE(features, "No features provided.\n")
-		REQUIRE(features->get_feature_class() == C_DENSE,
-			"Feature-class (%d) must be of type C_DENSE (%d)\n", features->get_feature_class(), C_DENSE)
-			
-		data = features;
-	}
-	else
-		REQUIRE(data->get_feature_class() == C_DENSE,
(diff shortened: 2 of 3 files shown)


src/shogun/machine/Machine.h
```diff
@@ -25,6 +25,20 @@
 namespace shogun
 {
 
+#define TRAIN_DENSE_DISPATHER(train_templated_name) \
+virtual bool train_with_dense(CFeatures* data) \
+{  \
+  switch (data->get_feature_type()) \
+  { \
+    case F_DREAL: \
+      return train_templated_name<float64_t>(data->as<CDenseFeatures<float64_t>>()); \
```
karlnapf: I guess we could get away with static casting here as well, but I am always more comfortable using a dynamic cast as done in  `as`

src/shogun/machine/Machine.h
```diff
@@ -315,7 +329,6 @@ class CMachine : public CStoppableSGObject
 
 		virtual const char* get_name() const { return "Machine"; }
 
-	protected:
```
karlnapf: you want `train_machine` to be public? why?

src/shogun/machine/Machine.h
```diff
@@ -328,11 +341,38 @@ class CMachine : public CStoppableSGObject
...
+			switch (data->get_feature_class())
+			{
+				case C_DENSE:
+					result=train_with_dense(data);
+					break;
+				case C_STRING:
+					result=train_with_string(data);
+					break;
+				default:
+					SG_ERROR("train_machine is not yet implemented for %s!\n",
```
karlnapf: the error message should change and say that.
"Training %s with %s is not supported.\n", get_name(), data->get_name()

src/shogun/machine/Machine.h
```diff
@@ -328,11 +341,38 @@ class CMachine : public CStoppableSGObject
 		 */
 		virtual bool train_machine(CFeatures* data=NULL)
 		{
-			SG_ERROR("train_machine is not yet implemented for %s!\n",
-					get_name());
+			bool result=false;
+			
+			switch (data->get_feature_class())
```
karlnapf: keep in mind data might be `nullptr`.
This actually raises a problem: as we cannot dispatch in that case ...
I guess we need to add a "default" option for training if features are not provided in the macro

karlnapf: @shubham808 I think for now you can add a "train_without_features" that could be overloaded by subclasses. We can then remove that later

src/shogun/regression/LeastAngleRegression.cpp
```diff
@@ -102,37 +102,6 @@ void CLeastAngleRegression::plane_rot(ST x0, ST x1,
...
-	{
-		REQUIRE(features, "No features provided.\n")
-		REQUIRE(features->get_feature_class() == C_DENSE,
-			"Feature-class (%d) must be of type C_DENSE (%d)\n", features->get_feature_class(), C_DENSE)
-			
-		data = features;
-	}
-	else
-		REQUIRE(data->get_feature_class() == C_DENSE,
-			"Feature-class must be of type C_DENSE (%d)\n", data->get_feature_class(), C_DENSE)
```
karlnapf: nice, so we shaved off all this code in every class that dispatches the templated type for training

src/shogun/machine/Machine.h
shubham808: this is a mistake...  no need for public ;)

src/shogun/machine/Machine.h
```diff
@@ -25,6 +25,20 @@
 namespace shogun
 {
 
+#define TRAIN_DENSE_DISPATHER(train_templated_name) \
+virtual bool train_with_dense(CFeatures* data) \
+{  \
```
karlnapf: As of now, I am quite puzzled how we could combine this idea with iterative machine, since making the `iteration` method virtual *and* templated is not an option, but neither is to do the dispatching inside that method :(

src/shogun/machine/Machine.h
shubham808: default option will also work for features classes not yet in the switch cases, train_without_features  makes more sense

src/shogun/machine/Machine.h
```diff
@@ -25,6 +25,20 @@
 namespace shogun
 {
 
+#define TRAIN_DENSE_DISPATHER(train_templated_name) \
```
karlnapf: typo "DISPATCHER"

src/shogun/machine/Machine.h
karlnapf: Could we make the macro in a way that we can use it do overload the 'train_with_dense' to dispatch the template type, but also that we can use it to overload 'iterate_dense' ?

src/shogun/machine/Machine.h
```diff
@@ -328,11 +341,38 @@ class CMachine : public CStoppableSGObject
...
+					SG_ERROR("train_machine is not yet implemented for %s!\n",
+							get_name())
+					break;
+			}
+			return result;
+		}
+		
+	protected:
+		
+		virtual bool train_with_dense(CFeatures* data)
```
karlnapf: maybe an easier name would be train_dense

src/shogun/machine/Machine.h
```diff
@@ -25,6 +25,20 @@
...
+{  \
+  switch (data->get_feature_type()) \
+  { \
+    case F_DREAL: \
+      return train_templated_name<float64_t>(data->as<CDenseFeatures<float64_t>>()); \
+		case F_SHORTREAL: \
+      return train_templated_name<float32_t>(data->as<CDenseFeatures<float32_t>>()); \
+		case F_LONGREAL: \
+	    return train_templated_name<floatmax_t>(data->as<CDenseFeatures<floatmax_t>>()); \
+  } \
```
karlnapf: needs a default case

src/shogun/machine/Machine.h
shubham808: i think we will need a new macro for iterate_dense

karlnapf: the code is essentially the same, so why not try to create one macro for both.

* accept the to be overloaded name as a parameter, here `train_with_dense`
* accept a variable number of parameters that are passed on to the definition of the overloaded method and the call of the typed implementation

karlnapf: nope, but it would be cool to use the same macro

karlnapf:
Ill close this one as we solved this using CRTP instead of macros

PR #3970, opened 2017-08-23
[FeaturesRefactor] Features::view()/preprocess()

micmn:
[Work in progress]

Continues #3968, only the last commit is relevant.

The design is outlined in [this](https://gist.github.com/micmn/52c82a6f47ebb8b49d35a5a3452becb8) gist.

I might split this into a few PRs that builds on the fundamental changes [1-2-3],
[4] cross validation, [5] `add/remove_subset()` replacement, [6] preprocessors.

Anyway, what's in here:

1. Base methods to construct new features `Features::preprocess()/view()`.

2. Base method to construct new labels `Labels::view()`, this requires a shallow copy 'ctor
like in `Features` classes (`duplicate()` + copy 'ctors),
implemented for (Dense/Binary/Multiclass/Regression)Labels.

3. `DenseFeatures`:
  - on-the-fly evaluation through `get_feature_vector()` method;
  - eager evaluation through `eval()` method;
  - `get_feature_matrix()` constness.

4. `CrossValidation`: use `view()` instead of `add/remove_subset()`.

5. Replace `add/remove_subset()` with `view()` + unit tests failing due to `DenseFeatures` changes [see below]:
  - `StochasticGBMachine`: refactor `get_subset()` and related functions
  to use a new labels object instead of adding the subset to `m_labels`,
  refactor unit tests and add case that covers `subset_frac < 1` branch;
  - `LMNNImpl`: refactor to use `preprocess()` instead of `apply_to_feature_matrix()`;
  - `CARTree`: fix feature matrix constness, minor changes to unit test (`some`);

6. Preprocessors:
  - `NormOne::apply_to_feature_vector()` uses linalg;
  - rewrite `PruneVarSubMean::init()` to use `DotFeatures::get_mean()/get_std()`
  (the latter was added to `DotFeatures`), add unit test;
  - `RandomFourierGaussPreproc`: fix feature matrix constness;
  - rewrite `MultipleProcessors` unit test with eval case;
  - rewrite `LeastAngleRegression.ols_equivalence` unit test to use `preprocess()`

[Misc] Initializer list constructor for `SGMatrix/Vector`
(not a big breakthrough but anyway useful for unit tests with fixed data...)

---
**Unit tests**:
  - 27 - unit-StochasticGBMachine (Fixed)
  - 32 - unit-LeastAngleRegression (Fixed)
  - 46 - unit-LMNNImpl (Fixed)
  - 47 - unit-LMNN (Fixed)
  - 167 - unit-CARTree (Fixed)
[...]

PR #4352, opened 2018-07-03
Add features and labels view

vinx13:
Continue #3970 , but features and labels view return raw pointer now instead of `Some` because that will cause covariant type problem, and `Some` is not available in SWIG

### DIFF
--- a/src/shogun/features/Features.cpp
+++ b/src/shogun/features/Features.cpp
@@ -265,6 +265,24 @@ void CFeatures::unset_property(EFeatureProperty p)
 	properties &= (properties | p) ^ p;
 }
 
+CFeatures* CFeatures::view(const SGVector<index_t>& subset)
+{
+	auto feats_view = this->duplicate();
+	feats_view->add_subset(subset.clone());
+	return feats_view;
+}
+
+CFeatures* CFeatures::view(const std::vector<index_t>& subset)
+{
+	auto feats_view = this->duplicate();
+
+	auto sg_subset = SGVector<index_t>(subset.size());
+	std::copy(subset.cbegin(), subset.cend(), sg_subset.data());
+	feats_view->add_subset(sg_subset);
+
+	return feats_view;
+}
+
 void CFeatures::add_subset(SGVector<index_t> subset)
 {
 	m_subset_stack->add_subset(subset);
--- a/src/shogun/features/Features.cpp
+++ b/src/shogun/features/Features.cpp
@@ -272,17 +272,6 @@ CFeatures* CFeatures::view(const SGVector<index_t>& subset)
 	return feats_view;
 }
 
-CFeatures* CFeatures::view(const std::vector<index_t>& subset)
-{
-	auto feats_view = this->duplicate();
-
-	auto sg_subset = SGVector<index_t>(subset.size());
-	std::copy(subset.cbegin(), subset.cend(), sg_subset.data());
-	feats_view->add_subset(sg_subset);
-
-	return feats_view;
-}
-
 void CFeatures::add_subset(SGVector<index_t> subset)
 {
 	m_subset_stack->add_subset(subset);
--- a/src/shogun/features/Features.cpp
+++ b/src/shogun/features/Features.cpp
@@ -268,7 +268,7 @@ void CFeatures::unset_property(EFeatureProperty p)
 CFeatures* CFeatures::view(const SGVector<index_t>& subset)
 {
 	auto feats_view = this->duplicate();
-	feats_view->add_subset(subset.clone());
+	feats_view->add_subset(subset);
 	return feats_view;
 }
 
--- a/src/shogun/features/Features.cpp
+++ b/src/shogun/features/Features.cpp
@@ -265,13 +265,6 @@ void CFeatures::unset_property(EFeatureProperty p)
 	properties &= (properties | p) ^ p;
 }
 
-CFeatures* CFeatures::view(const SGVector<index_t>& subset)
-{
-	auto feats_view = this->duplicate();
-	feats_view->add_subset(subset);
-	return feats_view;
-}
-
 void CFeatures::add_subset(SGVector<index_t> subset)
 {
 	m_subset_stack->add_subset(subset);
--- a/src/shogun/labels/BinaryLabels.cpp
+++ b/src/shogun/labels/BinaryLabels.cpp
@@ -134,6 +134,11 @@ CBinaryLabels::CBinaryLabels(const CDenseLabels& dense) : CDenseLabels(dense)
 	ensure_valid();
 }
 
+CLabels* CBinaryLabels::duplicate() const
+{
+	return new CBinaryLabels(*this);
+}
+
 namespace shogun
 {
 	Some<CBinaryLabels> binary_labels(CLabels* orig)
--- a/src/shogun/labels/BinaryLabels.h
+++ b/src/shogun/labels/BinaryLabels.h
@@ -96,6 +96,8 @@ public:
 	 */
 	virtual ELabelType get_label_type() const override;
 
+	virtual CLabels* duplicate() const;
+
 	/** Converts all scores to calibrated probabilities by fitting a
 	 * sigmoid function using the method described in
 	 * Lin, H., Lin, C., and Weng, R. (2007).
--- a/src/shogun/labels/BinaryLabels.h
+++ b/src/shogun/labels/BinaryLabels.h
@@ -96,7 +96,7 @@ public:
 	 */
 	virtual ELabelType get_label_type() const override;
 
-	virtual CLabels* duplicate() const;
+	virtual CLabels* duplicate() const override;
 
 	/** Converts all scores to calibrated probabilities by fitting a
 	 * sigmoid function using the method described in
--- a/src/shogun/labels/Labels.cpp
+++ b/src/shogun/labels/Labels.cpp
@@ -82,17 +82,6 @@ CLabels* CLabels::view(const SGVector<index_t>& subset)
 	return labels_view;
 }
 
-CLabels* CLabels::view(const std::vector<index_t>& subset)
-{
-	auto labels_view = this->duplicate();
-
-	auto sg_subset = SGVector<index_t>(subset.size());
-	std::copy(subset.cbegin(), subset.cend(), sg_subset.data());
-	labels_view->add_subset(sg_subset);
-
-	return labels_view;
-}
-
 float64_t CLabels::get_value(int32_t idx)
 {
 	ASSERT(m_current_values.vector && idx < get_num_labels())
--- a/src/shogun/labels/Labels.cpp
+++ b/src/shogun/labels/Labels.cpp
@@ -24,6 +24,13 @@ CLabels::CLabels(const CLabels& orig)
     : CSGObject(orig), m_current_values(orig.m_current_values)
 {
 	init();
+
+	if (orig.m_subset_stack != NULL)
+	{
+		SG_UNREF(m_subset_stack);
+		m_subset_stack = new CSubsetStack(*orig.m_subset_stack);
+		SG_REF(m_subset_stack);
+	}
 }
 
 CLabels::~CLabels()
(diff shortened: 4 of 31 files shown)


src/shogun/features/Features.cpp
```diff
@@ -265,6 +265,24 @@ void CFeatures::unset_property(EFeatureProperty p)
 	properties &= (properties | p) ^ p;
 }
 
+CFeatures* CFeatures::view(const SGVector<index_t>& subset)
+{
+	auto feats_view = this->duplicate();
```
karlnapf: i would prefer a copy ctor here

vinx13: this is a virtual function that calls copy ctor of subclasses

src/shogun/features/Features.cpp
```diff
@@ -265,6 +265,24 @@ void CFeatures::unset_property(EFeatureProperty p)
...
+	feats_view->add_subset(subset.clone());
+	return feats_view;
+}
+
+CFeatures* CFeatures::view(const std::vector<index_t>& subset)
```
karlnapf: this doesnt belong into CFeatures imo, this is conversion code between std::vector and SGVector and therefore should sit in SGVector
`features->view(SGVector<float64_t>(my_std_vector))`

karlnapf: or even done implicitly (if possible)

src/shogun/labels/BinaryLabels.cpp
```diff
@@ -156,6 +156,11 @@ CBinaryLabels::CBinaryLabels(const CDenseLabels& dense) : CDenseLabels(dense)
 	ensure_valid();
 }
 
+CLabels* CBinaryLabels::duplicate() const
+{
+	return new CBinaryLabels(*this);
```
karlnapf: why have this method why not just use the copy ctor?

src/shogun/labels/Labels.cpp
```diff
@@ -68,6 +75,24 @@ CSubsetStack* CLabels::get_subset_stack()
...
+	auto labels_view = this->duplicate();
+	labels_view->add_subset(subset);
+	return labels_view;
+}
+
+CLabels* CLabels::view(const std::vector<index_t>& subset)
+{
+	auto labels_view = this->duplicate();
+
+	auto sg_subset = SGVector<index_t>(subset.size());
```
karlnapf: same comment as above

src/shogun/machine/StochasticGBMachine.cpp
```diff
@@ -187,31 +189,25 @@ bool CStochasticGBMachine::train_machine(CFeatures* data)
 
 	for (auto i : SG_PROGRESS(range(m_num_iter)))
 	{
-		// apply subset
-		if (m_subset_frac!=1.0)
-			apply_subset(feats,interf);
+		const auto result = get_subset(feats, interf);
```
karlnapf: why all these changes? shouldnt we focus on the view first?

src/shogun/machine/StochasticGBMachine.cpp
```diff
@@ -260,10 +257,11 @@ CMachine* CStochasticGBMachine::fit_model(CDenseFeatures<float64_t>* feats, CReg
 	return c;
 }
 
-CRegressionLabels* CStochasticGBMachine::compute_pseudo_residuals(CRegressionLabels* inter_f)
+CRegressionLabels* CStochasticGBMachine::compute_pseudo_residuals(
+    CRegressionLabels* inter_f, CLabels* labs)
 {
-	REQUIRE(m_labels,"training labels not set!\n")
-	SGVector<float64_t> labels=(dynamic_cast<CDenseLabels*>(m_labels))->get_labels();
```
karlnapf: pls avoid such whitespace changes in PRs.
They should be in a sep PR. It just pollutes the diff (i.e. hard to review, takes longer)

src/shogun/machine/StochasticGBMachine.cpp
```diff
@@ -273,8 +271,14 @@ CRegressionLabels* CStochasticGBMachine::compute_pseudo_residuals(CRegressionLab
 	return new CRegressionLabels(residuals);
 }
 
-void CStochasticGBMachine::apply_subset(CDenseFeatures<float64_t>* f, CLabels* interf)
+std::tuple<Some<CDenseFeatures<float64_t>>, Some<CRegressionLabels>,
```
karlnapf: can you explain the reasonaing here?

tests/unit/multiclass/tree/CARTree_unittest.cc
```diff
@@ -519,7 +516,7 @@ TEST(CARTree, form_t1_test)
 	lab[3]=1;
 	lab[4]=0;
 
-	CDenseFeatures<float64_t>* feats=new CDenseFeatures<float64_t>(data);
+	auto feats = some<CDenseFeatures<float64_t>>(data);
```
karlnapf: again, I would prefer those changes to happen in a separate minimal PR

src/shogun/machine/StochasticGBMachine.cpp
vinx13: this is taken from michele's pr, we need small refactor on this to deploy views

src/shogun/machine/StochasticGBMachine.cpp
vinx13: since view create a new instance, using smart pointers prevent ref/unref

karlnapf: doesnt this create the same problem with Some?

src/shogun/machine/StochasticGBMachine.cpp
karlnapf: ok,
I technically would prefer first adding the view and then have a second Pr to deploy it.
But I think others see that differently, so ok leave it in :)

src/shogun/features/Features.cpp
karlnapf: ah sorry I see now, totally justified!

src/shogun/machine/StochasticGBMachine.cpp
vinx13: actually no, this function is only used internally, so we know the static type of return value based on types of args

src/shogun/features/Features.cpp
```diff
@@ -272,17 +272,6 @@ CFeatures* CFeatures::view(const SGVector<index_t>& subset)
 	return feats_view;
 }
 
-CFeatures* CFeatures::view(const std::vector<index_t>& subset)
```
karlnapf: can you explain ?
We need to use the base class interface in things like xvalidation ..

karlnapf: ah sorry nevermind I didnt see the arugment was the std

karlnapf: instead: yes ! :)

src/shogun/features/Features.cpp
```diff
@@ -265,6 +265,13 @@ void CFeatures::unset_property(EFeatureProperty p)
 	properties &= (properties | p) ^ p;
 }
 
+CFeatures* CFeatures::view(const SGVector<index_t>& subset)
+{
+	auto feats_view = this->duplicate();
+	feats_view->add_subset(subset.clone());
```
karlnapf: i wonder, do we want to clone the subset?
I think actually we shouldnt
And also I think that `add_subset` should accept a const vector

karlnapf: but maybe you have a different opinion?

tests/unit/labels/MulticlassLabels_unittest.cc
```diff
@@ -101,3 +102,16 @@ TEST_F(MulticlassLabels, multiclass_labels_from_dense_not_contiguous)
...
+TEST_F(MulticlassLabels, view)
+{
+	auto labels = some<CMulticlassLabels>(labels_true);
+	SGVector<index_t> subset{0, 2};
+	auto labels_subset = wrap(labels->view(subset)->as<CMulticlassLabels>());
+
+	ASSERT_EQ(labels_subset->get_num_labels(), subset.vlen);
+	for (auto i : range(subset.vlen))
+	{
+		EXPECT_EQ(labels_subset->get_int_label(i), labels_true[subset[i]]);
```
karlnapf: maybe we can actually systematically test all access methods of labels and features somehow?
I would also assert that the view's data pointer is the same as the original one (i.e. no copy happened)

src/shogun/features/Features.cpp
vinx13: agree, we don't need clone here

tests/unit/labels/MulticlassLabels_unittest.cc
vinx13: we don't have a way to test different features/labels types, and different access methods now.
and we can't check view's data pointer because that's a private member, calling `get_labels` will create a copy if subset is present

vinx13: we don't have a way to test different features/labels types, and different access methods now.
and we can't check view's data pointer because that's a private member, calling `get_labels` will create a copy if subset is present

src/shogun/lib/View.h
```diff
@@ -0,0 +1,36 @@
...
+namespace shogun
+{
+
+	template <class T>
+	T* view(T* viewable, const SGVector<index_t>& subset)
+	{
+		static_assert(
+		    std::is_base_of<CFeatures, T>::value ||
+		        std::is_base_of<CLabels, T>::value,
+		    "Only CFeatures and CLabels are viewable.");
```
karlnapf: I am not a fan of error messages that contain things that might change in the future, i.e. a new class is made viewable.
I would just state: Class is not viewable. The compiler error will provide the T and also the static assert which will tell the caller what is viewable and what is not. Or?

src/shogun/lib/View.h
```diff
@@ -0,0 +1,36 @@
...
+#include <shogun/features/Features.h>
+#include <shogun/labels/Labels.h>
+#include <shogun/lib/SGVector.h>
+#include <type_traits>
+
+namespace shogun
+{
+
+	template <class T>
+	T* view(T* viewable, const SGVector<index_t>& subset)
```
karlnapf: could we make `viewable` const?

karlnapf: is this used anywhere?
How are we going to do this without multiple inheritance?
Maybe a mixin approach would be better, see IterativeMachine

vinx13: @karlnapf i'm not clear, could you explain the multiple inheritance problem?
currently we expect T to be either features or labels, but we can also use viewable mixin in features and labels

vigsterkr: @karlnapf you mean about the duplciate method or what exactly do you mean by multiple inheritance?

src/shogun/lib/View.h
```diff
@@ -0,0 +1,36 @@
...
+		    std::is_base_of<CFeatures, T>::value ||
+		        std::is_base_of<CLabels, T>::value,
+		    "Only CFeatures and CLabels are viewable.");
+		auto result = viewable->duplicate();
+		result->add_subset(subset);
+		return static_cast<T*>(result);
+	}
+
+	template <class T>
+	T* view(Some<T> viewable, const SGVector<index_t>& subset)
```
vinx13: @karlnapf @vigsterkr maybe we can just return `Some<T>` here?

vigsterkr:
@vinx13 yep cool! this way the c++ code became much much cleaner!

src/shogun/lib/View.h
karlnapf: I think no.

src/shogun/lib/View.h
karlnapf: Yes please! If you pass Some you should get Some

src/shogun/lib/View.h
```diff
@@ -0,0 +1,51 @@
...
+
+	/** Creates a subset view of the viewable object containing the elements
+	 * whose indices are listed in the passed vector
+	 *
+	 * @param viewable pointer to the viewable object
+	 * @param subset subset of indices
+	 * @return new viewable instance
+	 */
+	template <class T>
+	T* view(T* viewable, const SGVector<index_t>& subset)
```
karlnapf: I still vote for const T*

src/shogun/lib/View.h
```diff
@@ -0,0 +1,51 @@
...
+
+	/** Creates a subset view of the viewable object containing the elements
+	 * whose indices are listed in the passed vector
+	 *
+	 * @param viewable pointer to the viewable object
+	 * @param subset subset of indices
+	 * @return new viewable instance
+	 */
+	template <class T>
+	Some<T> view(Some<T> viewable, const SGVector<index_t>& subset)
```
karlnapf: ++

src/shogun/labels/BinaryLabels.h
```diff
@@ -96,6 +96,8 @@ class CBinaryLabels : public CDenseLabels
 	 */
 	virtual ELabelType get_label_type() const;
 
+	virtual CLabels* duplicate() const;
```
vigsterkr: use `override`

src/shogun/labels/Labels.h
```diff
@@ -139,24 +139,33 @@ namespace shogun
 		 */
 		virtual SGVector<float64_t> get_values() const;
 
+		/** shallow-copy of the labels object
+		 *
+		 * @return labels object
+		 */
+		virtual CLabels* duplicate() const
```
vigsterkr: use `override`

PR #4356, opened 2018-07-06
Add Pseudo Inverse to linalg

shubham808:
[WIP] unit test and using for_each is left


geektoni:
Cool! Remeber to add also a unit test to check your implementation (so that we know that it works correctly).

src/shogun/mathematics/linalg/LinalgNamespace.h
```diff
@@ -484,6 +484,44 @@ namespace shogun
 			    ->add_col_vec(A, i, b, result, alpha, beta);
 		}
 
+		/** Calculates pseudo inverse A+ from the singular value
```
karlnapf: This doc is not entirely correct, as this is the self adjoint version, it uses an Eigensolver if the provided psd matrix, not an SVD.

As suggested before, I would put a flag in `self_adjoint` that defaults to false.
This would then use the SVD to compute the singular values (the rest of the code is identical for non-self-adjoint).

If we dont do this, there will be lots of redundant code (basically copy paste for the non self-adjoint version)

karlnapf: (and then rename it to pinv)

src/shogun/mathematics/linalg/LinalgNamespace.h
```diff
@@ -484,6 +484,44 @@ namespace shogun
...
+		 *
+		 * @param A The matrix
+		 * @param result The matrix that saves the result
+		 */
+		template <typename T>
+		void pinv_self_adjoint(const SGMatrix<T>& A, SGMatrix<T>& result)
+		{
+
+			REQUIRE(
+			    result.num_rows == A.num_rows && result.num_cols == A.num_cols,
```
karlnapf: pls check how scipy/numpy pinv behave
They do work for non symmetric matrices.
Best thing is you do unit tests

* symmetric psd matrix, using self_adjoint=true
* square matrix that is not PSD, using self_adjoint=false
* non-square matrix (more rows that cols), using self_adjoint=false
* non-square matrix (more cols than rows), using self_adjoint=false

I suggest using some 3x3 matrix, and then do a unit test that compares against pinv in numpy

karlnapf: ah and also tests that assert for raised errors when self-adjoint=true but a non-square matrix is passed

src/shogun/mathematics/linalg/LinalgNamespace.h
karlnapf: Actually, let's keep both methods in.
So instead of the flag, lets add another method called `pinv`
But make both methods share code via a helper method, they only compute the decomposition themselves

src/shogun/mathematics/linalg/LinalgNamespace.h
```diff
@@ -484,6 +484,70 @@ namespace shogun
...
+		template <typename T>
+		void pinvh(const SGMatrix<T>& A, SGMatrix<T>& result)
+		{
+
+			REQUIRE(
+			    result.num_rows == A.num_rows && result.num_cols == A.num_cols,
+			    "Dimension mismatch! A (%d x %d) vs result (%d x %d).\n",
+			    A.num_rows, A.num_cols, result.num_rows, result.num_cols);
+	
+			REQUIRE(A.num_rows == A.num_cols, "Given Matrix is not square A(%d x %d)", A.num_rows, A.num_cols);
```
karlnapf: matrix, not Matrix

also no need to print A

src/shogun/mathematics/linalg/LinalgNamespace.h
```diff
@@ -484,6 +484,70 @@ namespace shogun
...
+		 * Calculates pseudo inverse A+ from eigen values solved by a self adjoint eigen solver.
+		 * This version returns the result in a newly created matrix.
+		 *
+		 * @param A The matrix
+		 * @return The result matrix
+		 */
+		template <typename T, template <typename> class Container>
+		Container<T> pinvh(const SGMatrix<T>& A)
+		{
+			auto result = A.clone();
```
karlnapf: the memcpy that is part of clone is technically unncessesary here, i.e. the memory initialization of the result

src/shogun/mathematics/linalg/backend/eigen/BasicOps.cpp
```diff
@@ -208,6 +226,46 @@ void LinalgBackendEigen::add_diag_impl(
...
+void LinalgBackendEigen::pinv_impl(
+    const SGMatrix<T>& A, SGMatrix<T>& result) const
+{
+	auto m = CMath::min(A.num_cols, A.num_rows);
+	SGVector<T> s(m);
+	SGMatrix<T> V(A.num_rows, m);
+	typename SGMatrix<T>::EigenMatrixXtMap A_eig = A;
+	typename SGMatrix<T>::EigenMatrixXtMap result_eig = result;
+	
+	svd_impl<T>(A, s, V, true, linalg::SVDAlgorithm::BidiagonalDivideConquer);
```
karlnapf: I think you should offer the solver enum as a parameter in this method, just like svd does

src/shogun/mathematics/linalg/LinalgNamespace.h
```diff
@@ -484,6 +484,70 @@ namespace shogun
...
+		 * User should pass an appropriately pre-allocated memory matrix
+		 * or pass the operand argument A as a result.
+		 *
+		 * @param A The matrix
+		 * @param result The matrix that saves the result
+		 */
+		template <typename T>
+		void pinv(const SGMatrix<T>& A, SGMatrix<T>& result)
+		{
+			infer_backend(A)->pinv(A, result);
```
karlnapf: size checks?

src/shogun/mathematics/linalg/backend/eigen/BasicOps.cpp
```diff
@@ -208,6 +226,46 @@ void LinalgBackendEigen::add_diag_impl(
...
+	typename SGMatrix<T>::EigenMatrixXtMap A_eig = A;
+	typename SGMatrix<T>::EigenMatrixXtMap result_eig = result;
+	
+	eigen_solver_symmetric_impl<T>(A, s, V, m);
+	
+	typename SGVector<T>::EigenVectorXtMap s_eig = s;
+	typename SGMatrix<T>::EigenMatrixXtMap V_eig = V;
+		
+	float64_t pinv_tol = CMath::MACHINE_EPSILON * m * s_eig.real().maxCoeff();
+	s_eig.array() = (s_eig.real().array()>pinv_tol).select(s_eig.real().array().inverse(), 0);
```
karlnapf: maybe this line can be shared somewhere between both pinv versions? it is the same code after all

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
```diff
@@ -465,6 +465,60 @@ TEST(LinalgBackendEigen, SGMatrix_cross_entropy)
...
+	A(1, 2) = -1.0;
+	A(2, 0) = 0.0;
+	A(2, 1) = -1.0;
+	A(2, 2) = 2.0;
+	SGMatrix<float64_t> I(3, 3);
+	linalg::identity(I);
+	SGMatrix<float64_t>B(3, 3);
+	linalg::pinvh(A, B);
+	SGMatrix<float64_t>C = linalg::matrix_prod(A, B);
+	C.display_matrix(C);
```
karlnapf: pls remove this

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
```diff
@@ -465,6 +465,60 @@ TEST(LinalgBackendEigen, SGMatrix_cross_entropy)
...
+	for(auto i:range(3))
+	{
+		for(auto j:range(3))
+		{
+			EXPECT_NEAR(I(i, j), C(i, j), 1e-14);
```
karlnapf: This is not sufficient as an only test. There needs to be a comparison to scipy explicitly, so I suggest you keep this loop here, but also explicitly check the B matrix contents

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
```diff
@@ -465,6 +465,60 @@ TEST(LinalgBackendEigen, SGMatrix_cross_entropy)
...
+	A(1, 0) = -1.0;
+	A(1, 1) = 2.0;
+	A(1, 2) = -1.0;
+	A(2, 0) = 0.0;
+	A(2, 1) = -1.0;
+	A(2, 2) = 2.0;
+	SGMatrix<float64_t> I(3, 3);
+	linalg::identity(I);
+	SGMatrix<float64_t>B(3, 3);
+	linalg::pinvh(A, B);
```
karlnapf: what about naming B as A_pinv?

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
karlnapf: name C as I_check or so, so it is clear it is supposed to be the identity

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
```diff
@@ -465,6 +465,60 @@ TEST(LinalgBackendEigen, SGMatrix_cross_entropy)
...
+	A(0, 2) = 1.0;
+	A(0, 3) = 1.0;
+	A(1, 0) = 5.0;
+	A(1, 1) = 7.0;
+	A(1, 2) = 7.0;
+	A(1, 3) = 9.0;
+	SGMatrix<float64_t> I(2, 2);
+	linalg::identity(I);
+	SGMatrix<float64_t>B(4, 2);
+	linalg::pinvh(A, B);
```
karlnapf: same comments for the test above regarding names, and explicitly checking the output

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
```diff
@@ -465,6 +465,60 @@ TEST(LinalgBackendEigen, SGMatrix_cross_entropy)
...
+		{
+			EXPECT_NEAR(I(i, j), C(i, j), 1e-14);
+		}
+	}
+}
+
+
+TEST(LDA, SGMatrix_pinv)
+{
+	SGMatrix<float64_t>A(2, 4);
```
karlnapf: for the SVD case can you also pls test a 4x2 matrix and a 2x2 (or the one from above)
Also please test that wrong dimensions raise error messages

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
karlnapf: btw you also need to test the other interface where the result matrix is not passed, but generated (here you also need to check that the result has the correct size)

src/shogun/classifier/svm/NewtonSVM.cpp
```diff
@@ -127,7 +127,7 @@ void CNewtonSVM::iteration()
 
 	linalg::add(Xsv2sum, lcrossdiag, Xsv2sum);
 	SGMatrix<float64_t> inverse((x_d + 1), (x_d + 1));
-	SGMatrix<float64_t>::pinv(Xsv2sum.data(), x_d + 1, x_d + 1, inverse.data());
```
karlnapf: you could use the other interface here if you re-allocate the inverse in every iteration (no need to do 2 lines for that)
But in fact since the size is constant, why dont you share the memory of that matrix across iterations. Then it would be a bit more efficient...

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
```diff
@@ -465,6 +465,63 @@ TEST(LinalgBackendEigen, SGMatrix_cross_entropy)
 	EXPECT_NEAR(ref, result, 1e-15);
 }
 
+TEST(LinalgBackendEigen, SGMatrix_pinvh)
+{
+	float64_t A_data[] = {2.0, -1.0, 0.0,
+												-1.0, 2.0, -1.0,
```
karlnapf: the formatting here is a bit all over the place ;)

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
```diff
@@ -465,6 +465,63 @@ TEST(LinalgBackendEigen, SGMatrix_cross_entropy)
...
+													0.0, -1.0, 2.0};
+								
+	float64_t result_data[] = {0.75, 0.5, 0.25,
```
karlnapf: name this A_pinv_scipy or so, so it is clear where this comes from. A comment also will help

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
```diff
@@ -465,6 +465,63 @@ TEST(LinalgBackendEigen, SGMatrix_cross_entropy)
...
+															0.5, 1.0, 0.5,
+															0.25, 0.5, 0.75};
+																		
+	SGMatrix<float64_t>A(A_data, 3, 3, false);
+	SGMatrix<float64_t>result(result_data, 3, 3, false);
+	
+	SGMatrix<float64_t> I(3, 3);
```
karlnapf: is there no API for 
`auto I = linalg::identity(3);`?

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
shubham808: yes yes :) will fix it

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
```diff
@@ -465,6 +465,63 @@ TEST(LinalgBackendEigen, SGMatrix_cross_entropy)
...
+	SGMatrix<float64_t> I(2, 2);
+	linalg::identity(I);
+	SGMatrix<float64_t>A_pinv(2, 4);
+	linalg::pinv(A, A_pinv);
+	SGMatrix<float64_t>I_check = linalg::matrix_prod(A, A_pinv);
+	for(auto i:range(2))
+	{
+		for(auto j:range(2))
+		{
+			EXPECT_NEAR(I(i, j), I_check(i, j), 1e-14);
```
karlnapf: checking against result is missing

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
```diff
@@ -465,6 +465,63 @@ TEST(LinalgBackendEigen, SGMatrix_cross_entropy)
...
+			EXPECT_NEAR(result(i, j), A_pinv(i, j), 1e-14);
+		}
+	}
+}
+
+
+TEST(LinalgBackendEigen, SGMatrix_pinv)
+{
+	float64_t A_data[] = {1, 1, 1, 1,
```
karlnapf: as said, pls test all cases here: square, and both rectangular setups

src/shogun/mathematics/linalg/backend/eigen/BasicOps.cpp
```diff
@@ -208,6 +226,46 @@ void LinalgBackendEigen::add_diag_impl(
...
+	result_eig = V_eig * s_eig.asDiagonal() * V_eig.transpose();
+}
+
+template <typename T>
+void LinalgBackendEigen::pinv_impl(
+    const SGMatrix<T>& A, SGMatrix<T>& result, linalg::SVDAlgorithm alg) const
+{
+	auto m = CMath::min(A.num_cols, A.num_rows);
```
karlnapf: pls use `std::min`

src/shogun/mathematics/linalg/LinalgNamespace.h
```diff
@@ -484,6 +484,76 @@ namespace shogun
...
+		 * @param A The matrix
+		 * @param result The matrix that saves the result
+		 */
+		template <typename T>
+		void pinv(const SGMatrix<T>& A, SGMatrix<T>& result, 
+		SVDAlgorithm alg = SVDAlgorithm::BidiagonalDivideConquer)
+		{
+			REQUIRE(
+					result.num_rows == A.num_cols && result.num_cols == A.num_rows,
+					"Dimension mismatch! result must be of (%d x %d) provided is (%d x %d).\n",
```
karlnapf: Result with capital R

src/shogun/mathematics/linalg/LinalgNamespace.h
```diff
@@ -484,6 +484,76 @@ namespace shogun
...
+		 * Calculates pseudo inverse A+ from eigen values solved by a self adjoint eigen solver.
+		 * This version returns the result in a newly created matrix.
+		 *
+		 * @param A The matrix
+		 * @return The result matrix
+		 */
+		template <typename T>
+		SGMatrix<T> pinvh(const SGMatrix<T>& A)
+		{
+			auto result = A.clone();
```
karlnapf: i am still moaning about the clone, which has a sg_memcpy that is not necessary

src/shogun/mathematics/linalg/LinalgNamespace.h
```diff
@@ -484,6 +484,76 @@ namespace shogun
...
+		 * Calculates pseudo inverse A+ using singular value decomposition.
+		 * This version returns the result in a newly created matrix.
+		 *
+		 * @param A The matrix
+		 * @return The result matrix
+		 */
+		template <typename T>
+		SGMatrix<T> pinv(const SGMatrix<T>& A)
+		{
+			auto result = A.clone();
```
karlnapf: i dont think this can work

src/shogun/mathematics/linalg/LinalgNamespace.h
```diff
@@ -484,6 +484,76 @@ namespace shogun
...
+		SGMatrix<T> pinvh(const SGMatrix<T>& A)
+		{
+			auto result = A.clone();
+			pinvh(A, result);
+			return result;
+		}
+
+		/** Calculates pseudo inverse A+ using singular value decomposition.
+		 * User should pass an appropriately pre-allocated memory matrix
+		 * or pass the operand argument A as a result.
```
karlnapf: "or pass the operand argument A as a result" mean?
I dont think that can work here as the dimension doesnt work out or? It would only work for square matrices
But you need to add a test for this as well
`linalg::pinvh(A, A)`

src/shogun/mathematics/linalg/LinalgNamespace.h
shubham808: yes the dimensions would be incorrect

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
shubham808: no i did not find any

src/shogun/classifier/svm/NewtonSVM.cpp
shubham808: it is only used in newtonsvm it seems :)

src/shogun/mathematics/linalg/backend/eigen/BasicOps.cpp
```diff
@@ -208,6 +226,46 @@ void LinalgBackendEigen::add_diag_impl(
...
+
+template <typename T>
+void LinalgBackendEigen::pinv_impl(
+    const SGMatrix<T>& A, SGMatrix<T>& result) const
+{
+	auto m = std::min(A.num_cols, A.num_rows);
+	typename SGMatrix<T>::EigenMatrixXtMap A_eig = A;
+	typename SGMatrix<T>::EigenMatrixXtMap result_eig = result;
+
+	auto svd_eig = A_eig.jacobiSvd(Eigen::ComputeThinU | Eigen::ComputeThinV);
```
shubham808: we cannot use linalg::svd because it does not compute V matrix which we need to calculate inverse

karlnapf: mmh that is really weird

karlnapf: mmh that is really weird

src/shogun/classifier/svm/NewtonSVM.cpp
karlnapf: ok kill it then :)

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
karlnapf: i guess data type issues, ok leave it as it it

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
```diff
@@ -465,6 +465,116 @@ TEST(LinalgBackendEigen, SGMatrix_cross_entropy)
...
+	SGMatrix<float64_t> I_check = linalg::matrix_prod(A_pinverse, A);
+	for (auto i : range(2))
+	{
+		for (auto j : range(2))
+		{
+			EXPECT_NEAR(identity_matrix(i, j), I_check(i, j), 1e-14);
+		}
+	}
+	// compare with results from scipy
+	EXPECT_NEAR(A_pinverse(0, 0), 1.0, 1e-14);
```
karlnapf: btw there is ASSERT_DOUBLE_EQ

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
```diff
@@ -465,6 +465,116 @@ TEST(LinalgBackendEigen, SGMatrix_cross_entropy)
 	EXPECT_NEAR(ref, result, 1e-15);
 }
 
+TEST(LinalgBackendEigen, SGMatrix_pinv_psd)
```
karlnapf: name it pinvh after the method

and then split this into two tests, one for each method. You could use a fixture to share the data

karlnapf: i take this back. lets just leave it like this

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
```diff
@@ -465,6 +465,116 @@ TEST(LinalgBackendEigen, SGMatrix_cross_entropy)
...
+			EXPECT_NEAR(result(i, j), A_pinv(i, j), 1e-14);
+		}
+	}
+	// no memory errors 
+	EXPECT_NO_THROW(linalg::pinvh(A, A));
+}
+
+TEST(LinalgBackendEigen, SGMatrix_pinv_2x4)
+{
+	SGMatrix<float64_t> A(2, 4);
```
karlnapf: btw you should add a test for each method that complains if wrong dimensional result matrices are passed.
So that we have coverage of the REQUIRE statements

PR #4357, opened 2018-07-06
Disallow put of dynamic parameter

lisitsyn:


src/shogun/base/SGObject.h
```diff
@@ -349,9 +349,17 @@ class CSGObject
 	{
 		if (has_parameter(_tag))
 		{
+			auto parameter_value = get_parameter(_tag).get_value();
+			if (!parameter_value.cloneable())
+			{
+				SG_ERROR(
+					"Cannot put parameter %s::%s. Perhaps it is immutable for "
+					"%s",
```
karlnapf: dont understand the message.
Perhaps it is unmutable for CLASSNAME?

PR #4358, opened 2018-07-06
Feature/lazy get

karlnapf:


karlnapf:
@lisitsyn your stuff in action
@shubham808 @FaroukY @vinx13 here is how you can now port examples that use method calls in machines that are not part of `CMachine`'s interface: you register then as lazy methods and then can just use `get`.

src/shogun/base/SGObject.h
```diff
@@ -726,6 +731,22 @@ class CSGObject
 			std::bind(method, dynamic_cast<const S*>(this));
 		create_parameter(tag, AnyParameter(make_any(bind_method), properties));
 	}
+
+	/** Puts a pointer to a (lazily evaluated) function into the parameter map.
```
karlnapf: @lisitsyn there is the question whether we really want to allow this for non-const methods ---- they might alter the state of the object, which probably is not what you would expect from a `get`

karlnapf: but for this example, due to some weird design, get_cluster_centers is actually non-const and it cannot easily be made const since it calls `get_lhs` on the internal distance object, which increases the reference counter, which cannot be const. gna!
Object counter inside SGObject are not really ideas. ... is there maybe a way to fix this nicely?

PR #4363, opened 2018-07-08
Use map in class_list

lisitsyn:


src/shogun/base/class_list.cpp.templ
```diff
@@ -44,30 +44,20 @@ REPLACE template_definitions THIS
...
-typedef struct
-{
-	const char* m_class_name;
-	create_function m_create_function;
-} class_list_entry_t;
-#endif
+typedef CSGObject* (*CreateFunction)(EPrimitiveType generic);
 
-static class_list_entry_t class_list[] = {
+static const std::map<std::string, CreateFunction> classes = {
```
karlnapf: so this is O(logn) rather than O(n) ?
Good call!

lisitsyn: Yeah map fits better here. It became possible since we use initializer list here {} and C++11 is enabled.

PR #4364, opened 2018-07-08
Build CoreML protobufs

lisitsyn:


lisitsyn:
This might keep in branch but since we merged coreml I'd prefer to merge it also.

karlnapf:
sure :)

PR #4365, opened 2018-07-09
Cleanup non-const methods in dense and combined features

vinx13:
* methods made const: `create_merged_copy`, `get_feature_matrix`, `get_feature_obj`
* dropped public methods: `set_feature_vector`, `reshape`, `steal_feature_matrix`
* made `free_features` protected

### DIFF
--- a/src/shogun/classifier/NearestCentroid.cpp
+++ b/src/shogun/classifier/NearestCentroid.cpp
@@ -29,17 +29,12 @@ namespace shogun{
 
 	CNearestCentroid::~CNearestCentroid()
 	{
-		if(m_is_trained)
-			distance->remove_lhs();
-		else
-			delete m_centroids;
 	}
 
 	void CNearestCentroid::init()
 	{
 		m_shrinking=0;
 		m_is_trained=false;
-		m_centroids = new CDenseFeatures<float64_t>();
 	}
 
 
@@ -64,9 +59,6 @@ namespace shogun{
 		int32_t num_feats = ((CDenseFeatures<float64_t>*) data)->get_num_features();
 		SGMatrix<float64_t> centroids(num_feats,num_classes);
 		centroids.zero();
-
-		m_centroids->set_num_features(num_feats);
-		m_centroids->set_num_vectors(num_classes);
 
 		int64_t* num_per_class = new int64_t[num_classes];
 		for (int32_t i=0 ; i<num_classes ; i++)
@@ -100,12 +92,10 @@ namespace shogun{
 			SGVector<float64_t>::scale_vector(scale,target,num_feats);
 		}
 
-		m_centroids->free_feature_matrix();
-		m_centroids->set_feature_matrix(centroids);
-
+		auto centroids_feats = some<CDenseFeatures<float64_t>>(centroids);
 
 		m_is_trained=true;
-		distance->init(m_centroids,distance->get_rhs());
+		distance->init(centroids_feats, distance->get_rhs());
 
 		SG_FREE(num_per_class);
 
--- a/src/shogun/features/CombinedFeatures.h
+++ b/src/shogun/features/CombinedFeatures.h
@@ -76,7 +76,7 @@ class CCombinedFeatures : public CFeatures
 		}
 
 		/** list feature objects */
-		void list_feature_objs();
+		void list_feature_objs() const;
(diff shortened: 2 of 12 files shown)


lisitsyn:
lgtm!

src/shogun/classifier/NearestCentroid.cpp
```diff
@@ -100,12 +92,10 @@ namespace shogun{
...
 		}
 
-		m_centroids->free_feature_matrix();
-		m_centroids->set_feature_matrix(centroids);
-
+		auto centroids_feats = some<CDenseFeatures<float64_t>>(centroids);
 
 		m_is_trained=true;
-		distance->init(m_centroids,distance->get_rhs());
+		distance->init(centroids_feats, distance->get_rhs());
```
karlnapf: why the change here?

src/shogun/features/CombinedFeatures.h
```diff
@@ -76,7 +76,7 @@ class CCombinedFeatures : public CFeatures
 		}
 
 		/** list feature objects */
-		void list_feature_objs();
+		void list_feature_objs() const;
```
karlnapf: lol, this method should rather be deleted (or return a list of strings or so)

src/shogun/features/DenseFeatures.cpp
```diff
@@ -170,28 +170,6 @@ template<class ST> ST* CDenseFeatures<ST>::get_feature_vector(int32_t num, int32
 	return feat;
 }
 
-template<class ST> void CDenseFeatures<ST>::set_feature_vector(SGVector<ST> vector, int32_t num)
```
karlnapf: it is not used?

src/shogun/features/DenseFeatures.cpp
```diff
@@ -331,16 +309,6 @@ void CDenseFeatures<ST>::copy_feature_matrix(SGMatrix<ST> target, index_t column
 	}
 }
 
-template<class ST> SGMatrix<ST> CDenseFeatures<ST>::steal_feature_matrix()
```
karlnapf: lol another joker method :D

src/shogun/features/DenseFeatures.cpp
```diff
@@ -468,21 +438,6 @@ template<class ST> void CDenseFeatures<ST>::initialize_cache()
 
 template<class ST> EFeatureClass CDenseFeatures<ST>::get_feature_class() const  { return C_DENSE; }
 
-template<class ST> bool CDenseFeatures<ST>::reshape(int32_t p_num_features, int32_t p_num_vectors)
```
karlnapf: this should just be in the matrix itself, i.e. create a new feature instance using the reshaped sgmatrix

src/shogun/features/Features.h
```diff
@@ -259,7 +259,7 @@ class CFeatures : public CSGObject
 		 * @return new feature object which contains copy of data of this
 		 * instance and of given one
 		 */
-		virtual CFeatures* create_merged_copy(CFeatures* other)
+		virtual CFeatures* create_merged_copy(CFeatures* other) const
```
karlnapf: maybe we can rename this to `merge`  at some point later?

karlnapf:
CI is not happy though

src/shogun/features/DenseFeatures.cpp
vinx13: yes, and i think `set_feature_matrix` should be also removed (after some refactor)

src/shogun/classifier/NearestCentroid.cpp
vinx13: since we would like to drop setters in features`set_num_vectors`, `set_num_features`, we can't an empty features and then set feature matrix later, and here there is no need to store it in a member field

src/shogun/features/DenseFeatures.cpp
karlnapf: absolutely, the instance should be tied to the matrix, and otherwise users can create a new instance

karlnapf:
let me know once CI is happy and I will merge

vinx13:
@karlnapf timeout on windows

karlnapf:
restarted.
But I think this is good. Merging

PR #4367, opened 2018-07-12
Support on-the-fly preprocessors in dense features

vinx13:


vinx13:
@vigsterkr could you review this?

PR #4369, opened 2018-07-13
Make trained model testing support single mixins

karlnapf:
@micmn we added mixins which makes the base class checking a bit more complicated.
mixins are defined as `template <class T> CIterativeMachine : public T { ... }`, and the `class Perceptron : public CIterativeMachine<CLinearMachine> { ... }`, so ctags will generate `inherits: CIterativeMachine` and therefore the perceptron would not be included in the tests.

We cannot use your mapping function instead because I dont know what type of machine will lie inside, e.g. `'CiterativeMachine: CMachine'` would not work for example for kernel machines.

So what I did here is to extract the mixin template parameter from the declaration of the class and then substitute the base class with that, but keep the mixin in the `classes` dictionary (we might want to define tests for mixins only later on).

### DIFF
--- a/src/shogun/CMakeLists.txt
+++ b/src/shogun/CMakeLists.txt
@@ -592,6 +592,13 @@ IF (CTAGS_FOUND)
         --c++-kinds=cfgp
         --fields=+im
         --exclude=*.cpp
+        --exclude=third_party
+        --exclude=GoogleMock
+        --exclude=rxcpp
+        --exclude=StanMath
+        --exclude=interfaces
+        --exclude=Eigen3
+        --exclude=external
         --languages=c++
         -R ${CMAKE_SOURCE_DIR})
 
--- a/tests/unit/base/trained_model_serialization_test.cc.py
+++ b/tests/unit/base/trained_model_serialization_test.cc.py
@@ -8,7 +8,7 @@
 # Classes to ignore: mostly because default initialization isn't enough
 # to setup the machine for training (i.e. Multitask and DomainAdaptation),
 # different reasons are given below.
-IGNORE = [
+IGNORE = set([
     # LinearMachines
     'CFeatureBlockLogisticRegression', 'CLibLinearMTL',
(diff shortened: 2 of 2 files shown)


karlnapf:
Eventually, all this kind of stuff should rather be done using an AST parser rather than ctags, e.g. http://clang.llvm.org/docs/LibASTMatchersTutorial.html
but for now, this PR re-adds the classes to the test so we dont have error slipping through

micmn:
Wow last year I did some research on llvm for these unit tests but I totally missed LibASTMatchers :|

Did a quick test with _clang-query_: something like
`match cxxRecordDecl(allOf(hasName("CPerceptron"),isDerivedFrom(hasName("CLinearMachine"))))` should easily do it even for mixins.

karlnapf:
yessss.....
mind giving these automated tests a go with this lib?

karlnapf:
let me know if anything came out of this. would be so much better than using clang

micmn:
Sure, probably the most annoying thing is to figure out how to integrate it
in the build without slowing it down excessively?

Il lun 16 lug 2018 22:50 Heiko Strathmann <notifications@github.com> ha
scritto:

karlnapf:
Indeed,
But we could exclude it from travis, and just run those tests on the buildbot if needed, @vigsterkr what do you think?

micmn:
I tried this tutorial http://clang.llvm.org/docs/LibASTMatchersTutorial.html but it seems that to build the tool you need to compile clang and it takes hours and hours (and gbs of ram/hdd), then I tried linking against pre-built clang libraries but I got errors (maybe there is a reason they make you compile clang from source in the tutorial?)

So if we have to compile clang the only way would be to build a binary for this tool or...?

karlnapf:
yes :( mmmh dont know what to do from here...
@vigsterkr ?

vigsterkr:
Working on it.

You can actually use precompiled llvm and libtooling

PR #4371, opened 2018-07-13
set compute bias to false

shubham808:
Since model computes bias automatically we do not need to set it to mean difference later.


karlnapf:
this needs a data update, obviously

karlnapf:
Can be closed

PR #4373, opened 2018-07-15
Feature/dispatch from recursive mixin

shubham808:
[WIP] This is a minimal example based on the gist by @karlnapf about dispatching feature types through a mixin


src/shogun/machine/TypedMachine.h
```diff
@@ -0,0 +1,70 @@
...
+#include <shogun/lib/common.h>
+#include <shogun/machine/LinearMachine.h>
+
+namespace shogun
+{
+	class CFeatures;
+	class CLabels;
+	
+	template <class P, class T>
+	class CTypedMachine : public T
```
karlnapf: i think the name is not the best. I know we used this in the gist, but maybe we can call this FeatureDispatcherMachine (or something better)

shubham808: yes i agree

src/shogun/machine/TypedMachine.h
```diff
@@ -0,0 +1,70 @@
...
+		{
+		}
+
+	protected:
+		virtual bool train_machine(CFeatures* data = NULL)
+		{
+				if (!data)
+				{	
+					REQUIRE(this->features, "No features provided.\n")	
+					REQUIRE(this->features->get_feature_class() == C_DENSE,
```
karlnapf: So this class is supposed to JUST work with dense features? Then we need one for string and sparse, etc as well....

src/shogun/machine/TypedMachine.h
```diff
@@ -0,0 +1,70 @@
...
+		{
+				if (!data)
+				{	
+					REQUIRE(this->features, "No features provided.\n")	
+					REQUIRE(this->features->get_feature_class() == C_DENSE,	
+					"Feature-class (%d) must be of type C_DENSE (%d)\n", this->features->get_feature_class(), C_DENSE)	
+					
+					data = this->features;	
+				}	
+		else
```
karlnapf: formatting clash here

src/shogun/machine/TypedMachine.h
```diff
@@ -0,0 +1,70 @@
...
+		virtual ~CTypedMachine()
+		{
+		}
+
+	protected:
+		virtual bool train_machine(CFeatures* data = NULL)
+		{
+				if (!data)
+				{	
+					REQUIRE(this->features, "No features provided.\n")
```
karlnapf: what is `this` here? This is inherited from the base class I guess?

src/shogun/machine/TypedMachine.h
shubham808: we can add more cases like we did in macro version and then throw error if nothing matches

src/shogun/machine/TypedMachine.h
shubham808: this would be CTypedMachine<CLARS, CLinearMachine>

src/shogun/machine/TypedMachine.h
```diff
@@ -0,0 +1,70 @@
...
+					REQUIRE(this->features->get_feature_class() == C_DENSE,	
+					"Feature-class (%d) must be of type C_DENSE (%d)\n", this->features->get_feature_class(), C_DENSE)	
+					
+					data = this->features;	
+				}	
+		else	
+			REQUIRE(data->get_feature_class() == C_DENSE,	
+				"Feature-class must be of type C_DENSE (%d)\n", data->get_feature_class(), C_DENSE)	
+		
+			REQUIRE(data->get_num_vectors() == this->m_labels->get_num_labels(), "Number of training vectors (%d) does not match number of labels (%d)\n"
```
karlnapf: if this class is just for dense, then such checks should be one level up in the hierarchy since we dont want to copy paste them around

src/shogun/machine/TypedMachine.h
```diff
@@ -0,0 +1,70 @@
...
+				}	
+		else	
+			REQUIRE(data->get_feature_class() == C_DENSE,	
+				"Feature-class must be of type C_DENSE (%d)\n", data->get_feature_class(), C_DENSE)	
+		
+			REQUIRE(data->get_num_vectors() == this->m_labels->get_num_labels(), "Number of training vectors (%d) does not match number of labels (%d)\n"	
+		, data->get_num_vectors(), this->m_labels->get_num_labels())
+			
+			//check for type of CFeatures, then call the appropriate template method
+			if(data->get_feature_type() == F_DREAL)
```
karlnapf: formatting all over :)

shubham808: yes.. will fix that

src/shogun/machine/TypedMachine.h
```diff
@@ -0,0 +1,70 @@
...
+				return this->template as<P>()->template train_machine_templated<float64_t>(data->as<CDenseFeatures<float64_t>>());
+			else if(data->get_feature_type() == F_SHORTREAL)
+				return this->template as<P>()->template train_machine_templated<float32_t>(data->as<CDenseFeatures<float32_t>>());
```
karlnapf: wow didnt expect we need such things, but yes makes sense if i think about it

src/shogun/machine/TypedMachine.h
shubham808: this will be general but i have only written the dense case

src/shogun/classifier/LDA.cpp
```diff
@@ -46,6 +46,8 @@ void CLDA::init()
 	m_method=AUTO_LDA;
 	m_gamma=0;
 	m_bdc_svd = true;
+  set_compute_bias(false);
```
karlnapf: lets do this in a spearate PR ok?

src/shogun/machine/TypedMachine.h
shubham808: yes this is a bit tricky

karlnapf:
Does all this work actually?

src/shogun/classifier/LDA.cpp
shubham808: actually it was already false we are taking it out of the constructor

shubham808:
CI seems happy :)

src/shogun/machine/TypedMachine.h
karlnapf: i think the problem here is that we force the subclasses to implement the templated train method. So if we add string here, we will get a compile error when it is not implemented for strings....or?

karlnapf:
Cool, can you make sure to extensively test the examples with those two classes (memory, results, etc) Hard to overview otherwise

src/shogun/machine/TypedMachine.h
karlnapf: not so sure about that.
Add a case for string and you will see the problem (see above)

src/shogun/machine/TypedMachine.h
karlnapf: but whose features are this?
CLARS right? So this means that your implementation here depends on the fact that the first template parameter has a member called features. We dont want to enforce that actually.

shubham808: features is CLinearMachine member CLARS do not store them locally... so i think all will have that member like CLDA because it is in machine

src/shogun/machine/TypedMachine.h
lisitsyn: I mean `this->template as<P>` might go into some variable.

shubham808: ah i see :)

src/shogun/machine/TypedMachine.h
karlnapf: my point is that this implementation assumes that T or P has a member called features, i.e. it wont compile otherwise.
I would just enforce features to be passed here tbh.
We wanna to this anyways soon, so all the mixins can already require features to be passed, so they dont need to be members

src/shogun/regression/LeastAngleRegression.h
```diff
@@ -218,18 +208,20 @@ class CLeastAngleRegression: public CLinearMachine
...
+public:
 	/**
 	* A templated specialization of the train_machine method
 	* @param data training data
 	* @see train_machine
 	*/
 	template <typename ST>
 	bool train_machine_templated(CDenseFeatures<ST> * data);
+	
+	virtual bool support_dispatching(){ return true; }
```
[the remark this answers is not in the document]
karlnapf: also, maybe a different name might be nice.

src/shogun/classifier/LDA.cpp
karlnapf: i see, thx!

src/shogun/machine/TypedMachine.h
karlnapf: CMachineFeatureTypeCRTP would be more descriptive, but also more ugly ;)

src/shogun/machine/TypedMachine.h
karlnapf: could you paste the compile error that happens when `train_machine_templated` does not exist in CLARS? Curious how it looks

shubham808: `/home/atom/shogun/src/shogun/machine/TypedMachine.h:49:98: error: ‘class shogun::CLDA’ has no member named ‘train_machine_templated’
     return obj->template train_machine_templated<float64_t>(data->as<CDenseFeatures<float64_t>>());` 
also here ```obj = this->as<P>()```

src/shogun/machine/TypedMachine.h
shubham808: yes i see it :)

src/shogun/machine/TypedMachine.h
karlnapf: this is actually pretty good ...
Just need to make sure in the docs of the CRTP class that we mention that implementing this method is mandatory

src/shogun/regression/LeastAngleRegression.h
```diff
@@ -218,18 +208,20 @@ class CLeastAngleRegression: public CLinearMachine
...
-	/** Initialize and register parameters */
-	void init();
-
+public:
 	/**
 	* A templated specialization of the train_machine method
 	* @param data training data
 	* @see train_machine
 	*/
 	template <typename ST>
```
karlnapf: could you try to make this only available for floating point numbers?
`template <class ST, class = typename std::enable_if<std::is_floating_point<ST>::value>::type>`

src/shogun/machine/TypedMachine.h
lisitsyn: We can use something like static_assert for meaningful messages

shubham808:
as discussed, it makes more sense to dispatch feature classes in CMachine. This is a mix of our previous similar pr and the mixin

src/shogun/machine/FeatureDispatchCRTP.h
```diff
@@ -0,0 +1,120 @@
...
+		}
+
+		virtual ~CDenseFeaturesDispatch()
+		{
+		}
+
+	protected:
+		virtual bool train_dense(CFeatures* data)
+		{
+			REQUIRE(
```
karlnapf: assert for non null pointer

src/shogun/machine/FeatureDispatchCRTP.h
```diff
@@ -0,0 +1,120 @@
...
+			    data->get_num_vectors() == this->m_labels->get_num_labels(),
```
karlnapf: the check should happen inside `CMachine`, otherwise we have to repeat it

src/shogun/machine/FeatureDispatchCRTP.h
```diff
@@ -0,0 +1,120 @@
...
+			    "Number of training vectors (%d) does not match number of "
+			    "labels (%d)\n",
+			    data->get_num_vectors(), this->m_labels->get_num_labels())
+
+			// check for type of CFeatures, then call the appropriate template
+			// method
+			auto obj = this->template as<P>();
```
karlnapf: I think `auto` should be replaced with the real type here to make reading the code easier

src/shogun/machine/FeatureDispatchCRTP.h
shubham808: yes since we are enforcing data should be passed during train

src/shogun/machine/FeatureDispatchCRTP.h
```diff
@@ -0,0 +1,120 @@
...
+			if (data->get_feature_type() == F_DREAL)
```
karlnapf: what about a switch here?

src/shogun/machine/FeatureDispatchCRTP.h
```diff
@@ -0,0 +1,120 @@
...
+#include <shogun/machine/LinearMachine.h>
+
+namespace shogun
+{
+	class CFeatures;
+	class CLabels;
+#define IGNORE_IN_CLASSLIST
+
+	template <class P, class T>
+	IGNORE_IN_CLASSLIST class CDenseFeaturesDispatch : public T
```
karlnapf: the class name should match the filename please.
Or is your plan to put all the dispatcher classes into one file?

src/shogun/machine/FeatureDispatchCRTP.h
lisitsyn: I don't quite agree, the type is already in the right part.

src/shogun/machine/FeatureDispatchCRTP.h
shubham808: yes that is the plan

src/shogun/machine/FeatureDispatchCRTP.h
```diff
@@ -0,0 +1,120 @@
...
+				    data->as<CDenseFeatures<float64_t>>());
+			else if (data->get_feature_type() == F_SHORTREAL)
+				return obj->template train_machine_templated<float32_t>(
+				    data->as<CDenseFeatures<float32_t>>());
+			else if (data->get_feature_type() == F_LONGREAL)
+				return obj->template train_machine_templated<floatmax_t>(
+				    data->as<CDenseFeatures<floatmax_t>>());
+			else
+				SG_SERROR(
+				    "Feature-type (%d) must be of type F_SHORTREAL (%d), "
```
karlnapf: could we introduce something that translates the feature type enum to a string and print the strings here instead?

E.g. `std::string feature_type(EFeatureType)` that returns the actual string, which we then can use to print such things in a human readable way?

karlnapf: I would not print all options here, think if one type is added, then we need to change all error messages.
Just saw that "Training with dense features of provided type %s is not possible"

src/shogun/machine/FeatureDispatchCRTP.h
```diff
@@ -0,0 +1,120 @@
...
+		}
+
+		virtual ~CStringFeaturesDispatch()
+		{
+		}
+
+	protected:
+		virtual bool train_string(CFeatures* data)
+		{
+			REQUIRE(
```
karlnapf: you see, this is copy pasted code that can go to the base class

src/shogun/machine/FeatureDispatchCRTP.h
```diff
@@ -0,0 +1,120 @@
...
+	protected:
+		virtual bool train_dense(CFeatures* data)
+		{
+			REQUIRE(
+			    data->get_num_vectors() == this->m_labels->get_num_labels(),
+			    "Number of training vectors (%d) does not match number of "
+			    "labels (%d)\n",
+			    data->get_num_vectors(), this->m_labels->get_num_labels())
+
+			// check for type of CFeatures, then call the appropriate template
```
karlnapf: remove comment

src/shogun/machine/FeatureDispatchCRTP.h
shubham808: yes i agree :)

src/shogun/machine/FeatureDispatchCRTP.h
```diff
@@ -0,0 +1,120 @@
...
+				    data->as<CStringFeatures<uint8_t>>());
+			else if (data->get_feature_type() == F_CHAR)
+				return obj->template train_machine_templated<char>(
+				    data->as<CStringFeatures<char>>());
+			else if (data->get_feature_type() == F_WORD)
+				return obj->template train_machine_templated<uint16_t>(
+				    data->as<CStringFeatures<uint16_t>>());
+			else
+				SG_SERROR(
+				    "Feature-type (%d) must be of type F_SHORTREAL (%d), "
```
karlnapf: there we go, this is why I dont like to list options in error messages, the text is wrong now.
Just do the msg as above

src/shogun/machine/Machine.h
```diff
@@ -334,8 +334,40 @@ class CMachine : public CStoppableSGObject
 		 */
 		virtual bool train_machine(CFeatures* data=NULL)
 		{
-			SG_ERROR("train_machine is not yet implemented for %s!\n",
-					get_name());
+			bool result = false;
+			if (support_features_dispatching())
```
karlnapf: This doesnt make sense: `train_machine` is overloaded by lots of shogun algorithms, so you cannot use that. You have to do the feature class dispatching in `train`, and it also needs to be optional (for the machines that are not CRTP), and the default option should be to just call `train_machine??`

karlnapf: supports_feature_dispatching

src/shogun/machine/Machine.h
```diff
@@ -334,8 +334,40 @@ class CMachine : public CStoppableSGObject
...
+			return false;
+		}
+
+		virtual bool train_string(CFeatures* data)
+		{
+			SG_NOTIMPLEMENTED
+			return false;
+		}
+
+		virtual bool support_features_dispatching()
```
karlnapf: I am afraid we need one of such checks for each feature class to avoid problems when passing for example passing string features to LARS

tests/unit/classifier/LDA_unittest.cc
```diff
@@ -201,10 +202,11 @@ TEST(LDA, num_classes_in_labels_exception)
 	auto labels = some<CDenseLabels>();
 	labels->set_labels(lab);
 	auto features = some<CDenseFeatures<float64_t>>(feat);
-	auto lda = some<CLDA>(0, features, labels, SVD_LDA);
+	auto lda = some<CLDA>(0, SVD_LDA);
+	lda->set_labels(labels);
 	// should throw an incorrect number of classes exception (expected value is
 	// 2)
-	EXPECT_THROW(lda->train(), ShogunException);
+	EXPECT_THROW(lda->train(features), ShogunException);
```
karlnapf: we need some tests for the dispatcher in a mock class environment, to test that

* features are correctly dispatched, also for multiple feature types
* errors are thrown if the "wrong" type is passed

src/shogun/classifier/LDA.h
```diff
@@ -173,7 +173,7 @@ class CLDA : public CLinearMachine
 		/** @return object name */
 		virtual const char* get_name() const { return "LDA"; }
 
-	protected:
+	public:
```
karlnapf: working with friend classes is infeasible here?

shubham808: i have not tried yet

tests/unit/machine/FeatureDispatchCRTP_unittest.cc
```diff
@@ -0,0 +1,52 @@
...
+	auto mock_machine = some<CMockLinear>();
+	mock_machine->set_labels(some<CBinaryLabels>(labels));
+
+	// cannot train with null features
+	EXPECT_THROW(mock_machine->train(), ShogunException);
+	// features are F_DREAL and C_DENSE
+	EXPECT_TRUE(mock_machine->train(some<CDenseFeatures<float64_t>>(features)));
+	// features are F_SHORTREAL and C_DENSE
+	EXPECT_FALSE(
+	    mock_machine->train(some<CDenseFeatures<float32_t>>(features_short)));
```
shubham808: test for exception when trying to pass StringFeatures is missing

src/shogun/machine/Machine.cpp
```diff
@@ -55,7 +55,29 @@ bool CMachine::train(CFeatures* data)
...
+		    data->get_num_vectors(), this->m_labels->get_num_labels())
+
+		if (support_dense_dispatching() && data->get_feature_class() == C_DENSE)
+			result = train_dense(data);
+		else if (
+		    support_string_dispatching() &&
+		    data->get_feature_class() == C_STRING)
+			result = train_string(data);
+		else
+			SG_ERROR("Training with %s is not implemented!", data->get_name());
```
karlnapf: "Cannot train %s using %s./n". get_name(), data->get_name()

src/shogun/machine/Machine.cpp
```diff
@@ -55,7 +55,29 @@ bool CMachine::train(CFeatures* data)
 	}
 
 	auto sub = connect_to_signal_handler();
-	bool result = train_machine(data);
+	bool result = false;
+
+	if (support_features_dispatching())
+	{
+		REQUIRE(data != NULL, "Features cannot be NULL!");
```
karlnapf: No features provided

karlnapf: pls also assert labels, you call a method on them below

src/shogun/machine/FeatureDispatchCRTP.h
```diff
@@ -0,0 +1,114 @@
...
+				return obj->template train_machine_templated<uint8_t>(
+				    data->as<CStringFeatures<uint8_t>>());
+			case F_CHAR:
+				return obj->template train_machine_templated<char>(
+				    data->as<CStringFeatures<char>>());
+			case F_WORD:
+				return obj->template train_machine_templated<uint16_t>(
+				    data->as<CStringFeatures<uint16_t>>());
+			default:
+				SG_SERROR("Training with StringFeatures of provided type is "
```
karlnapf: of provided type (%s) ..., ptype_name(data->get_ptype) (i dont know how to get the ptype but i think it is possible)

tests/unit/machine/FeatureDispatchCRTP_unittest.cc
```diff
@@ -0,0 +1,52 @@
...
+	CMockLinear() : CDenseFeaturesDispatch<CMockLinear, CMachine>()
+	{
+	}
+	~CMockLinear()
+	{
+	}
+	template <typename T>
+	bool train_machine_templated(CDenseFeatures<T>* data)
+	{
+		if (data->get_feature_type() == F_DREAL)
```
shubham808: is this a good way to test it ?

karlnapf: F_DREAL is hardcoded!

karlnapf: I have doubts now

src/shogun/classifier/LDA.h
```diff
@@ -182,24 +182,22 @@ class CLDA : public CLinearMachine
...
-
-	   	/**
-		 * A templated specialization of the train_machine method
-		 * @param features training data
-		 * @param labels labels for training data
-		 * @see train_machine
-		 */
-		template <typename ST>
-		bool train_machine_templated();
+		template <typename ST, typename U = typename std::enable_if<
```
karlnapf: could you pls use `enable_if_t` ?

karlnapf: `typename U = typename std::enable_if_t<std::is_arithmetic<T>::value>>`

src/shogun/classifier/LDA.h
karlnapf: did you try the friend class thingi? would be nicer to  not have this method public. minor though

src/shogun/classifier/LDA.h
```diff
@@ -182,24 +182,21 @@ class CLDA : public CLinearMachine
...
-	   	/**
-		 * A templated specialization of the train_machine method
-		 * @param features training data
-		 * @param labels labels for training data
-		 * @see train_machine
-		 */
-		template <typename ST>
-		bool train_machine_templated();
+		template <typename ST, typename U = typename std::enable_if_t<std::is_floating_point<ST>::value>>
+		bool train_machine_templated(CDenseFeatures<ST>* data);
```
karlnapf: newline pls

src/shogun/classifier/LDA.h
```diff
@@ -182,24 +182,21 @@ class CLDA : public CLinearMachine
...
+		virtual bool support_dispatching()
```
karlnapf: cant this sit inside the mixin?

src/shogun/classifier/svm/WDSVMOcas.cpp
```diff
@@ -60,8 +60,8 @@ CWDSVMOcas::CWDSVMOcas()
 }
 
 CWDSVMOcas::CWDSVMOcas(E_SVM_TYPE type)
-: CMachine(), use_bias(false), bufsize(3000), C1(1), C2(1),
-	epsilon(1e-3), method(type)
+    : CStringFeaturesDispatch<CWDSVMOcas, CMachine>(), use_bias(false),
```
karlnapf: we would need an integration (or unit) test for this class with string data before we can merge this.
Could you send an example in a separate PR?
@vinx13 what is the state of string feature meta examples?

src/shogun/classifier/LDA.h
shubham808: actually we dont need it :) will remove altogether

src/shogun/classifier/svm/WDSVMOcas.cpp
```diff
@@ -156,21 +156,15 @@ int32_t CWDSVMOcas::set_wd_weights()
...
-	{
-		if (data->get_feature_class() != C_STRING ||
-				data->get_feature_type() != F_BYTE)
-		{
-			SG_ERROR("Features not of class string type byte\n")
-		}
-		set_features((CStringFeatures<uint8_t>*) data);
-	}
+
+	set_features((CStringFeatures<uint8_t>*)data);
```
karlnapf: this will fail if you pass anythign else then uin8_t string features. You will need to assert the above if you want to rely on this: https://github.com/shogun-toolbox/shogun/pull/4373/files#diff-2a3c04af057ddaaf9cfa245feef22d22L167

src/shogun/features/FeatureTypes.cpp
```diff
@@ -0,0 +1,50 @@
...
+			return "SHORTREAL";
+		case F_DREAL:
+			return "DREAL";
+		case F_LONGREAL:
+			return "LONGREAL";
+		case F_ANY:
+			return "ANY";
+		default:
+			SG_SNOTIMPLEMENTED
+			return "UNKNOWN";
```
karlnapf: ++

src/shogun/machine/FeatureDispatchCRTP.h
```diff
@@ -0,0 +1,118 @@
...
+		}
+
+		virtual ~CDenseRealDispatch()
+		{
+		}
+
+	protected:
+		virtual bool train_dense(CFeatures* data)
+		{
+			auto obj = this->template as<P>();
```
karlnapf: I think a more appropriate name would be "this_casted"

src/shogun/machine/FeatureDispatchCRTP.h
```diff
@@ -0,0 +1,118 @@
...
+				    data->as<CDenseFeatures<float64_t>>());
+			case F_SHORTREAL:
+				return obj->template train_machine_templated<float32_t>(
+				    data->as<CDenseFeatures<float32_t>>());
+			case F_LONGREAL:
+				return obj->template train_machine_templated<floatmax_t>(
+				    data->as<CDenseFeatures<floatmax_t>>());
+			default:
+				SG_SERROR(
+				    "Training with DenseFeatures of provided %s type is not "
```
karlnapf: Training with %s of provided type %s is not ....(move type and use get_name of features)

src/shogun/machine/FeatureDispatchCRTP.h
```diff
@@ -0,0 +1,118 @@
...
+				    data->as<CStringFeatures<uint8_t>>());
+			case F_CHAR:
+				return obj->template train_machine_templated<char>(
+				    data->as<CStringFeatures<char>>());
+			case F_WORD:
+				return obj->template train_machine_templated<uint16_t>(
+				    data->as<CStringFeatures<uint16_t>>());
+			default:
+				SG_SERROR(
+				    "Training with StringFeatures of provided %s type is "
```
karlnapf: as above

src/shogun/machine/Machine.h
```diff
@@ -334,8 +334,35 @@ class CMachine : public CStoppableSGObject
...
-			SG_ERROR("train_machine is not yet implemented for %s!\n",
-					get_name());
+			SG_ERROR(
+			    "train_machine is not yet implemented for %s!\n", get_name());
+			return false;
+		}
+
+		virtual bool train_dense(CFeatures* data)
+		{
+			SG_NOTIMPLEMENTED
```
karlnapf: Could you be more explicit here: "Training %s with %s is not supported", get_name(), data->get_name()

src/shogun/regression/LeastAngleRegression.h
```diff
@@ -218,18 +208,24 @@ class CLeastAngleRegression: public CLinearMachine
...
 	* @param data training data
 	* @see train_machine
 	*/
-	template <typename ST>
-	bool train_machine_templated(CDenseFeatures<ST> * data);
+	template <typename ST, typename U = typename std::enable_if_t<
+		                       std::is_floating_point<ST>::value>>
+	bool train_machine_templated(CDenseFeatures<ST>* data);
+
+	virtual bool support_dispatching()
```
karlnapf: could this not sit in the mixin?

tests/unit/utils/Utils.cpp
```diff
@@ -27,6 +28,30 @@ void generate_temp_filename(char* file_name)
 #endif
 }
 
+SGStringList<char> generateRandomStringData(
```
karlnapf: good!

tests/unit/machine/FeatureDispatchCRTP_unittest.cc
```diff
@@ -0,0 +1,81 @@
...
+#include <gtest/gtest.h>
+#include <shogun/base/some.h>
+#include <shogun/features/DenseFeatures.h>
+#include <shogun/features/StringFeatures.h>
+#include <shogun/machine/FeatureDispatchCRTP.h>
+#include <shogun/machine/LinearMachine.h>
+
+using namespace shogun;
+
+class CMockModel : public CDenseRealDispatch<CMockModel, CMachine>
```
karlnapf: MockMachine

tests/unit/machine/FeatureDispatchCRTP_unittest.cc
```diff
@@ -0,0 +1,81 @@
...
+	auto labels = SGVector<float64_t>({1, -1});
+
+	auto mock_machine = some<CMockModel>();
+	mock_machine->set_labels(some<CBinaryLabels>(labels));
+
+	mock_machine->val = features->get_feature_type();
+	EXPECT_TRUE(mock_machine->train(features));
+}
+
+TEST(TrainDense, train_dense_with_int)
```
karlnapf: could you rename the rest to "TrainDenseReal, train_wrong_feature_type"

tests/unit/machine/FeatureDispatchCRTP_unittest.cc
```diff
@@ -0,0 +1,81 @@
...
+	auto features = some<CDenseFeatures<int16_t>>(feats);
+	auto labels = SGVector<float64_t>({1, -1});
+
+	auto mock_machine = some<CMockModel>();
+	mock_machine->set_labels(some<CBinaryLabels>(labels));
+
+	EXPECT_THROW(mock_machine->train(features), ShogunException);
+}
+
+TEST(TrainDense, train_dense_with_string)
```
karlnapf: this should be typed as well (i.e. tried for all feature types for string) or?

src/shogun/machine/FeatureDispatchCRTP.h
```diff
@@ -0,0 +1,118 @@
...
+		{
+		}
+
+	protected:
+		virtual bool train_string(CFeatures* data)
+		{
+			auto obj = this->template as<P>();
+			switch (data->get_feature_type())
+			{
+			case F_BYTE:
```
karlnapf: why did you chose those specific 3 types?

tests/unit/utils/Utils.cpp
karlnapf: maybe template it (with default parameter being char)

tests/unit/machine/FeatureDispatchCRTP_unittest.cc
```diff
@@ -0,0 +1,81 @@
...
+{
+	auto strings = generateRandomStringData(2);
```
karlnapf: you would need to make this templated to make the test templated

src/shogun/classifier/svm/WDSVMOcas.cpp
vinx13: with string feature factory, we can port meta examples now

src/shogun/classifier/LDA.h
karlnapf: I think we do to enter the branch in CMachine::train or?

src/shogun/classifier/svm/WDSVMOcas.cpp
karlnapf: @vinx13 would you mind adding an example for the `CWDSVMOcas` ? :) (no cookbook needed for now)

vinx13: allright

src/shogun/machine/FeatureDispatchCRTP.h
```diff
@@ -0,0 +1,119 @@
...
+				    data->as<CDenseFeatures<float64_t>>());
+			case F_SHORTREAL:
+				return this_casted->template train_machine_templated<float32_t>(
+				    data->as<CDenseFeatures<float32_t>>());
+			case F_LONGREAL:
+				return this_casted
+				    ->template train_machine_templated<floatmax_t>(
+				        data->as<CDenseFeatures<floatmax_t>>());
+			default:
+				SG_OBJ_ERROR(
```
shubham808: travis was causing errors.. i think this should fix or ?

src/shogun/machine/Machine.cpp
```diff
@@ -55,7 +55,29 @@ bool CMachine::train(CFeatures* data)
 	}
 
 	auto sub = connect_to_signal_handler();
-	bool result = train_machine(data);
+	bool result = false;
+
+	if (support_feature_dispatching())
+	{
+		REQUIRE(data != NULL, "Features not provided!");
+		REQUIRE(
```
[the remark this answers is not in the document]
shubham808: i think the check here is okay ? https://github.com/shogun-toolbox/shogun/pull/4373/files#diff-cc285e900e1599bd903dbb6d435f6089L52

tests/unit/machine/FeatureDispatchCRTP_unittest.cc
```diff
@@ -0,0 +1,135 @@
...
+#include <shogun/features/StringFeatures.h>
+#include <shogun/machine/FeatureDispatchCRTP.h>
+#include <shogun/machine/LinearMachine.h>
+
+using namespace shogun;
+
+class CDenseRealMockMachine
+    : public CDenseRealDispatch<CDenseRealMockMachine, CMachine>
+{
+public:
```
shubham808: making it protected caused errors with some<> also train_machine_templated need to be public so i will update it when i look into making it friend

tests/unit/machine/FeatureDispatchCRTP_unittest.cc
shubham808: actually this exception is generated in cmachine before we check for feature type so i think this is okay for a single case or ?

src/shogun/machine/FeatureDispatchCRTP.h
shubham808: int, short int and char seemed okay... but we can add more to it

src/shogun/machine/Machine.h
shubham808: the same error as in CMachine here or ?

src/shogun/machine/Machine.cpp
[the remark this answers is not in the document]
karlnapf: the link doesnt work, where is this?

examples/undocumented/python/classifier_lda.py
```diff
@@ -1,25 +0,0 @@
-#!/usr/bin/env python
```
karlnapf: ok!

src/shogun/machine/FeatureDispatchCRTP.h
```diff
@@ -0,0 +1,121 @@
...
+
+#include <shogun/features/DenseFeatures.h>
+#include <shogun/lib/common.h>
+#include <shogun/machine/LinearMachine.h>
+
+namespace shogun
+{
+	class CFeatures;
+	class CLabels;
+#define IGNORE_IN_CLASSLIST
```
karlnapf: wasnt this defined already?

src/shogun/machine/Machine.h
karlnapf: actually nevermind, this should never be executed anyways

src/shogun/machine/Machine.cpp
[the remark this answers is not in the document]
shubham808: [Machine.cpp line 51](https://github.com/shubham808/shogun/blob/c498f80caafb2e421dfd05f88759d1029fd9bb52/src/shogun/machine/Machine.cpp#L51)

src/shogun/machine/FeatureDispatchCRTP.h
shubham808: no we need this here

src/shogun/classifier/LDA.h
karlnapf: mmh this is not what I meant ... the formatting is messed up now ...

src/shogun/machine/FeatureDispatchCRTP.h
```diff
@@ -35,22 +35,22 @@ namespace shogun
...
 				    data->as<CDenseFeatures<float32_t>>());
 			case F_LONGREAL:
-				return obj->template train_machine_templated<floatmax_t>(
+				return this_casted->template train_machine_templated<floatmax_t>(
 				    data->as<CDenseFeatures<floatmax_t>>());
 			default:
-				SG_SERROR(
-				    "Training with DenseFeatures of provided %s type is not "
-				    "possible!",
+				SG_OBJ_ERROR(this_casted,
```
karlnapf: `"Training %s with %s of type %s is not supported.", get_name(), data->get_name(), feature_type(data->get_feature_type())`

src/shogun/machine/FeatureDispatchCRTP.h
```diff
@@ -83,22 +83,22 @@ namespace shogun
...
-				return obj->template train_machine_templated<uint16_t>(
+				return this_casted->template train_machine_templated<uint16_t>(
 				    data->as<CStringFeatures<uint16_t>>());
 			default:
-				SG_SERROR(
-				    "Training with StringFeatures of provided %s type is "
-				    "not possible!",
+				SG_OBJ_ERROR(this_casted,
+				    "Training with %s of provided type %s is "
+				    "not possible!",data->get_name(),
```
karlnapf: as above

src/shogun/machine/FeatureDispatchCRTP.h
karlnapf: ok!

src/shogun/machine/FeatureDispatchCRTP.h
```diff
@@ -0,0 +1,121 @@
...
+			case F_SHORTREAL:
+				return this_casted->template train_machine_templated<float32_t>(
+				    data->as<CDenseFeatures<float32_t>>());
+			case F_LONGREAL:
+				return this_casted
+				    ->template train_machine_templated<floatmax_t>(
+				        data->as<CDenseFeatures<floatmax_t>>());
+			default:
+				SG_SERROR(
+				    "Training with %s of provided type %s is not "
```
karlnapf: I made a comment on the commit how this error msg should look like,

src/shogun/machine/Machine.cpp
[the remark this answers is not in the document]
karlnapf: this is not enough as it is only checked if `training_requires_labels()` returns true. Your code here can segfault if no labels are provided....you need to enforce labels if dispatching is enabled ...
Since we will enforce labels to be passed soon, you can do that for the case of dispatching

src/shogun/classifier/LDA.h
```diff
@@ -95,6 +95,7 @@ template <class ST> class CDenseFeatures;
 
 class CLDA : public CDenseRealDispatch<CLDA, CLinearMachine>
 {
+	friend class CDenseRealDispatch<CLDA, CLinearMachine>;
```
karlnapf: sorry, we cannot do this unfortunately

karlnapf: sorry, my bad, this is exactly what we wanna do :D

PR #4375, opened 2018-07-16
Add factory of embed string features

vinx13:
add another string features factory that calls `obtain_from_char` internally to translate a string features to high order representation

### DIFF
--- a/src/shogun/util/factory.h
+++ b/src/shogun/util/factory.h
@@ -106,6 +106,50 @@ namespace shogun
 		return result;
 	}
 
+	/** Create embed string features from string char features
+	 * @param features StringCharFeatures
+	 * @param start start
+	 * @param p_order order
+	 * @param gap gap
+	 * @param rev reverse
+	 * @param primitive_type primitive type of the string features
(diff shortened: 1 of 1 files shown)


src/shogun/util/factory.h
```diff
@@ -106,6 +106,50 @@ namespace shogun
 		return result;
 	}
 
+	/** Create embed string features from string char features
```
karlnapf: "embedded"

karlnapf: this is user facing doc, so you need to make it a bit more user friendly

src/shogun/util/factory.h
```diff
@@ -106,7 +106,11 @@ namespace shogun
 		return result;
 	}
 
-	/** Create embed string features from string char features
+	/** Create embedded string features from string char features.
+	 * The new features has the same alphabet as the original features. Data of
+	 * the new features is obtained by calling `obtain_from_char` with the given
```
karlnapf: Much better. One final thing, could you make doxygen link the class method here?

PR #4376, opened 2018-07-17
Drop usage of set_feature_matrix

vinx13:
now the only usage of `set_feature_matrix` is swig https://github.com/shogun-toolbox/shogun/blob/770f748f4fd3b6db66ac08e699924f7152a8557d/src/interfaces/python/DenseFeatures_protocols.i#L661
after that we can drop it from public api


vinx13:
an idea to deal with swig is to use a friend function, so that we can hide `set_feature_matrix` from public api, @karlnapf what do you think

karlnapf:
I think that is a good idea, depending on where this statement will be.
The alternative in such special cases is to use `put<SGMatrix<T>>("feature_matrix", mat)`

PR #4377, opened 2018-07-17
Support cross validation of pipeline

vinx13:
* implemented `store_model_features` in pipeline, which forwards the call to the underlying machine of the pipeline
~* made `store_model_feature` public, such that as a method of base class `CMachine` it is callable from pipeline~
* overrode `clone` to make pipeline cloneable
* overrode `get_machine_problem_type`

### DIFF
--- a/src/shogun/clustering/Hierarchical.h
+++ b/src/shogun/clustering/Hierarchical.h
@@ -108,6 +108,11 @@ class CHierarchical : public CDistanceMachine
 			return false;
 		}
 
+		/** TODO: Ensures cluster centers are in lhs of underlying distance
+		 * Currently: does nothing.
+		 * */
+		virtual void store_model_features();
+
 	protected:
 		/** estimate hierarchical clustering
 		 *
@@ -118,11 +123,6 @@ class CHierarchical : public CDistanceMachine
 		 * @return whether training was successful
 		 */
 		virtual bool train_machine(CFeatures* data=NULL);
-
-		/** TODO: Ensures cluster centers are in lhs of underlying distance
-		 * Currently: does nothing.
-		 * */
-		virtual void store_model_features();
 
 	private:
 		/** Initialize attributes */
--- a/src/shogun/clustering/Hierarchical.h
+++ b/src/shogun/clustering/Hierarchical.h
@@ -108,11 +108,6 @@ class CHierarchical : public CDistanceMachine
 			return false;
 		}
 
-		/** TODO: Ensures cluster centers are in lhs of underlying distance
-		 * Currently: does nothing.
-		 * */
-		virtual void store_model_features();
-
 	protected:
 		/** estimate hierarchical clustering
 		 *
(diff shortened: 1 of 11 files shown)


src/shogun/clustering/Hierarchical.h
```diff
@@ -108,6 +108,11 @@ class CHierarchical : public CDistanceMachine
 			return false;
 		}
 
+		/** TODO: Ensures cluster centers are in lhs of underlying distance
```
karlnapf: could you open an issue with this?

karlnapf:
couldnt we make some friend classes instead of adding the set_store_model features to the interface. This is not meant to be an interface method (definitely not for SWIG, but not even for cpp)

src/shogun/machine/LinearMachine.h
```diff
@@ -142,11 +148,11 @@ class CLinearMachine : public CMachine
 		 */
 		virtual SGVector<float64_t> apply_get_outputs(CFeatures* data);
 
-		/** Stores feature data of underlying model. Does nothing because
-		 * Linear machines store the normal vector of the separating hyperplane
-		 * and therefore the model anyway
-		 */
-		virtual void store_model_features();
+		/** Computes the added bias. The bias is computed
```
karlnapf: wasnt this method remove in develop?

src/shogun/machine/Pipeline.cpp
```diff
@@ -222,4 +223,37 @@ namespace shogun
 	{
 		return shogun::get<CMachine*>(m_stages.back().second);
 	}
+
+	void CPipeline::set_store_model_features(bool store_model)
+	{
+		get_machine()->set_store_model_features(store_model);
```
karlnapf: If Pipeline inherits from CMachine, why cant the method be protected?

src/shogun/machine/Pipeline.cpp
```diff
@@ -222,4 +223,37 @@ namespace shogun
...
+
+	void CPipeline::store_model_features()
+	{
+		get_machine()->store_model_features();
+	}
+
+	CSGObject* CPipeline::clone()
+	{
+		auto result = CMachine::clone()->as<CPipeline>();
+		for (auto&& stage : m_stages)
```
karlnapf: I think we could probably do this via the parameter framework clone, with a few additions for std::types. The class is not serializable anyways.
@lisitsyn what do you think?

src/shogun/machine/Pipeline.cpp
vinx13: it's a protected method from base class, you cannot access it of another instance in the derived class `CPipeline`

src/shogun/clustering/Hierarchical.h
vinx13: this comment as left before, i just copy-pasted to make the method public

vinx13:
@vigsterkr could you check and merge this?

PR #4380, opened 2018-07-20
Add cookbook of cross validation with pipeline

vinx13:
data pr shogun-toolbox/shogun-data#164
also need to merge #4377 first

### DIFF
--- a/doc/cookbook/source/examples/evaluation/cross_validation_pipeline.rst
+++ b/doc/cookbook/source/examples/evaluation/cross_validation_pipeline.rst
@@ -1,6 +1,6 @@
-============================
+==============================
 Cross Validation on a Pipeline
-============================
+==============================
 
 In this example, we illustrate how to use cross-validation with :sgclass:`CPipeline`.
 
--- a/doc/cookbook/source/examples/evaluation/cross_validation_pipeline.rst
+++ b/doc/cookbook/source/examples/evaluation/cross_validation_pipeline.rst
@@ -1,5 +1,5 @@
 ============================
-Cross Validation on Pipeline
+Cross Validation on a Pipeline
 ============================
 
 In this example, we illustrate how to use cross-validation with :sgclass:`CPipeline`.
@@ -9,22 +9,21 @@ For more information on cross-validation, check :doc:`./cross_validation`.
 -------
 Example
 -------
-We'll use as example a binary classification problem solvable by a pipeline consisted of a transformer :sgclass:`CPruneVarSubMean` and a machine :sgclass:`CLibLinear`.
+We demonstrate a pipeline consisting of a transformer :sgclass:`CPruneVarSubMean` for normalizing the features, and a machine :sgclass:`CLibLinear` for binary classification.
 
-Imagine we have files with training data. We create :sgclass:`CDenseFeatures` (here 64 bit floats aka RealFeatures) as
+We create :sgclass:`CFeatures` and :sgclass:`CLabels` via loading from files
 
 .. sgexample:: cross_validation_pipeline:create_features
 
-
-We use :sgclass:`CPruneVarSubMean` to normalize the features and then use :sgclass:`CLibLinear` for classification.
-The transformer and the machine are chained as a :sgclass:`CPipeline`.
+We first chain all transformers, and then finalize the pipeline with the classifier.
 
 .. sgexample:: cross_validation_pipeline:create_pipeline
 
 Next, we initialize a splitting strategy :sgclass:`CStratifiedCrossValidationSplitting` to divide the dataset into :math:`k-` folds for the :math:`k-` fold cross validation.
-We also have to decide on an evaluation criterion class (from :sgclass:`CEvaluation`) to evaluate the performance of the trained models.
+We also have to decide on an evaluation criterion class (see :sgclass:`CEvaluation`) to evaluate the performance of the trained model.
 In this case, we use :sgclass:`CAccuracyMeasure`.
 We then instantiate :sgclass:`CCrossValidation` and set the number of cross validation's runs.
+The pipeline instance behaves just like a :sgclass:`CMachine` and this can be directly passed to :sgclass:`CCrossValidation`.
 
 .. sgexample:: cross_validation_pipeline:create_cross_validation
 
--- a/doc/cookbook/source/examples/evaluation/cross_validation_pipeline.rst
+++ b/doc/cookbook/source/examples/evaluation/cross_validation_pipeline.rst
@@ -23,7 +23,7 @@ Next, we initialize a splitting strategy :sgclass:`CStratifiedCrossValidationSpl
 We also have to decide on an evaluation criterion class (see :sgclass:`CEvaluation`) to evaluate the performance of the trained model.
 In this case, we use :sgclass:`CAccuracyMeasure`.
 We then instantiate :sgclass:`CCrossValidation` and set the number of cross validation's runs.
-The pipeline instance behaves just like a :sgclass:`CMachine` and this can be directly passed to :sgclass:`CCrossValidation`.
+We pass the :sgclass:`CPipeline` instance, which implements the :sgclass:`CMachine` API, to :sgclass:`CCrossValidation`.
 
 .. sgexample:: cross_validation_pipeline:create_cross_validation
 
--- a/doc/cookbook/source/examples/evaluation/cross_validation_pipeline.rst
+++ b/doc/cookbook/source/examples/evaluation/cross_validation_pipeline.rst
@@ -15,7 +15,7 @@ We create :sgclass:`CFeatures` and :sgclass:`CLabels` via loading from files
 
 .. sgexample:: cross_validation_pipeline:create_features
 
-We first chain all transformers, and then finalize the pipeline with the classifier.
+We first chain transformers, and then finalize the pipeline with the classifier.
 
 .. sgexample:: cross_validation_pipeline:create_pipeline
 
--- a/doc/cookbook/source/examples/evaluation/cross_validation_pipeline.rst
+++ b/doc/cookbook/source/examples/evaluation/cross_validation_pipeline.rst
@@ -1,0 +1,34 @@
+============================
+Cross Validation on Pipeline
+============================
+
+In this example, we illustrate how to use cross-validation with :sgclass:`CPipeline`.
+
+For more information on cross-validation, check :doc:`./cross_validation`.
+
+-------
+Example
+-------
+We'll use as example a binary classification problem solvable by a pipeline consisted of a transformer :sgclass:`CPruneVarSubMean` and a machine :sgclass:`CLibLinear`.
+
+Imagine we have files with training data. We create :sgclass:`CDenseFeatures` (here 64 bit floats aka RealFeatures) as
+
+.. sgexample:: cross_validation_pipeline:create_features
+
+
+We use :sgclass:`CPruneVarSubMean` to normalize the features and then use :sgclass:`CLibLinear` for classification.
+The transformer and the machine are chained as a :sgclass:`CPipeline`.
+
+.. sgexample:: cross_validation_pipeline:create_pipeline
+
+Next, we initialize a splitting strategy :sgclass:`CStratifiedCrossValidationSplitting` to divide the dataset into :math:`k-` folds for the :math:`k-` fold cross validation.
+We also have to decide on an evaluation criterion class (from :sgclass:`CEvaluation`) to evaluate the performance of the trained models.
+In this case, we use :sgclass:`CAccuracyMeasure`.
+We then instantiate :sgclass:`CCrossValidation` and set the number of cross validation's runs.
+
+.. sgexample:: cross_validation_pipeline:create_cross_validation
+
+Finally, we evaluate the model and get the results (aka a :sgclass:`CCrossValidationResult` instance).
+
+.. sgexample:: cross_validation_pipeline:evaluate_and_get_result
+
--- a/doc/cookbook/source/examples/evaluation/cross_validation_pipeline.rst
+++ b/doc/cookbook/source/examples/evaluation/cross_validation_pipeline.rst
@@ -23,7 +23,7 @@ Next, we initialize a splitting strategy :sgclass:`CStratifiedCrossValidationSpl
 We also have to decide on an evaluation criterion class (see :sgclass:`CEvaluation`) to evaluate the performance of the trained model.
 In this case, we use :sgclass:`CAccuracyMeasure`.
 We then instantiate :sgclass:`CCrossValidation` and set the number of cross validation's runs.
-We pass the :sgclass:`CPipeline` instance, which implements the :sgclass:`CMachine` API, to :sgclass:`CCrossValidation`.
+We next create a cross-validation instance and pass the generated pipeline.
 
 .. sgexample:: cross_validation_pipeline:create_cross_validation
 
--- a/examples/meta/generator/targets/cpp.json
+++ b/examples/meta/generator/targets/cpp.json
@@ -96,7 +96,8 @@
             "get_int_vector": "$object->get<SGVector<int32_t>>($arguments)",
             "get_real": "$object->get<float64_t>($arguments)",
             "get_real_vector": "$object->get<SGVector<float64_t>>($arguments)",
-            "get_real_matrix": "$object->get<SGMatrix<float64_t>>($arguments)"
+            "get_real_matrix": "$object->get<SGMatrix<float64_t>>($arguments)",
+            "put_machine": "$object->put<CMachine>($arguments)"
         },
         "StaticCall": "C$typeName::$method($arguments)",
         "GlobalCall": "$method($arguments)",
--- a/examples/meta/generator/targets/cpp.json
+++ b/examples/meta/generator/targets/cpp.json
@@ -96,8 +96,7 @@
             "get_int_vector": "$object->get<SGVector<int32_t>>($arguments)",
             "get_real": "$object->get<float64_t>($arguments)",
             "get_real_vector": "$object->get<SGVector<float64_t>>($arguments)",
-            "get_real_matrix": "$object->get<SGMatrix<float64_t>>($arguments)",
-            "put_machine": "$object->put<CMachine>($arguments)"
+            "get_real_matrix": "$object->get<SGMatrix<float64_t>>($arguments)"
         },
         "StaticCall": "C$typeName::$method($arguments)",
         "GlobalCall": "$method($arguments)",
--- a/examples/meta/generator/targets/python.json
+++ b/examples/meta/generator/targets/python.json
@@ -2,7 +2,7 @@
     "Program": "import numpy as np\n$dependencies\n\n$program",
     "Dependencies": {
         "IncludeAllClasses": false,
-        "IncludeInterfacedClasses": true, 
+        "IncludeInterfacedClasses": true,
         "IncludeEnums": true,
         "IncludeGlobalFunctions": true,
         "DependencyListElement": "from shogun import $typeName",
@@ -61,7 +61,8 @@
             "get_int_vector": "$object.get($arguments)",
             "get_real": "$object.get($arguments)",
             "get_real_vector": "$object.get($arguments)",
-            "get_real_matrix": "$object.get($arguments)"
+            "get_real_matrix": "$object.get($arguments)",
+            "put_machine": "$object.put($arguments)"
         },
         "StaticCall": "$typeName.$method($arguments)",
         "GlobalCall": "$method($arguments$kwargs)",
--- a/examples/meta/generator/targets/python.json
+++ b/examples/meta/generator/targets/python.json
@@ -2,7 +2,7 @@
     "Program": "import numpy as np\n$dependencies\n\n$program",
     "Dependencies": {
         "IncludeAllClasses": false,
-        "IncludeInterfacedClasses": true,
+        "IncludeInterfacedClasses": true, 
         "IncludeEnums": true,
         "IncludeGlobalFunctions": true,
         "DependencyListElement": "from shogun import $typeName",
@@ -61,8 +61,7 @@
             "get_int_vector": "$object.get($arguments)",
             "get_real": "$object.get($arguments)",
             "get_real_vector": "$object.get($arguments)",
-            "get_real_matrix": "$object.get($arguments)",
-            "put_machine": "$object.put($arguments)"
+            "get_real_matrix": "$object.get($arguments)"
         },
         "StaticCall": "$typeName.$method($arguments)",
         "GlobalCall": "$method($arguments$kwargs)",
--- a/examples/meta/src/evaluation/cross_validation_pipeline.sg
+++ b/examples/meta/src/evaluation/cross_validation_pipeline.sg
@@ -21,8 +21,8 @@ Machine pipeline = builder.then(svm)
 
 #![create_cross_validation]
 SplittingStrategy strategy = splitting_strategy("StratifiedCrossValidationSplitting", labels=labels_train, num_subsets=2)
-Evaluation evaluation_criteron = evaluation("AccuracyMeasure")
-MachineEvaluation cross = machine_evaluation("CrossValidation", machine=pipeline, features=feats_train, labels=labels_train, splitting_strategy=strategy, evaluation_criteron=evaluation_criteron, autolock=False, num_runs=2)
+Evaluation evaluation_criterion = evaluation("AccuracyMeasure")
+MachineEvaluation cross = machine_evaluation("CrossValidation", machine=pipeline, features=feats_train, labels=labels_train, splitting_strategy=strategy, evaluation_criterion=evaluation_criterion, autolock=False, num_runs=2)
 #![create_cross_validation]
 
 #![evaluate_and_get_result]
--- a/examples/meta/src/evaluation/cross_validation_pipeline.sg
+++ b/examples/meta/src/evaluation/cross_validation_pipeline.sg
@@ -1,0 +1,33 @@
+File f_feats_train = csv_file("../../data/classifier_binary_2d_linear_features_train.dat")
+File f_feats_test = csv_file("../../data/classifier_binary_2d_linear_features_test.dat")
+File f_labels_train = csv_file("../../data/classifier_binary_2d_linear_labels_train.dat")
+File f_labels_test = csv_file("../../data/classifier_binary_2d_linear_labels_test.dat")
+
+#![create_features]
+Features feats_train = features(f_feats_train)
+Features feats_test = features(f_feats_test)
+Labels labels_train = labels(f_labels_train)
+Labels labels_test = labels(f_labels_test)
+#![create_features]
+
+#![create_pipeline
+Transformer subMean = transformer("PruneVarSubMean")
+Machine svm = machine("LibLinear")
+
+PipelineBuilder builder()
+builder.over(subMean)
+Pipeline pipeline = builder.then(svm)
+#![create_pipeline]
+
+#![create_cross_validation]
+StratifiedCrossValidationSplitting splitting_strategy(labels_train, 2)
+Evaluation evaluation_criteron = evaluation("AccuracyMeasure")
+CrossValidation cross(pipeline, feats_train, labels_train, splitting_strategy, evaluation_criteron, False)
+cross.set_num_runs(2)
+#![create_cross_validation]
+
+#![evaluate_and_get_result]
+EvaluationResult result = cross.evaluate() 
+real mean = result.get_real("mean")
+real stddev = result.get_real("std_dev")
+#![evaluate_and_get_result]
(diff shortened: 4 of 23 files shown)


doc/cookbook/source/examples/evaluation/cross_validation_pipeline.rst
```diff
@@ -0,0 +1,34 @@
+============================
+Cross Validation on Pipeline
```
karlnapf: on "a" pipeline

doc/cookbook/source/examples/evaluation/cross_validation_pipeline.rst
```diff
@@ -0,0 +1,34 @@
...
+============================
+
+In this example, we illustrate how to use cross-validation with :sgclass:`CPipeline`.
```
karlnapf: maybe link to some cross-validation cookbook. There is some way to link to notebooks, I think MKL does this

karlnapf: ha it is just below, nevermind :)

doc/cookbook/source/examples/evaluation/cross_validation_pipeline.rst
```diff
@@ -0,0 +1,34 @@
...
+
+For more information on cross-validation, check :doc:`./cross_validation`.
+
+-------
+Example
+-------
+We'll use as example a binary classification problem solvable by a pipeline consisted of a transformer :sgclass:`CPruneVarSubMean` and a machine :sgclass:`CLibLinear`.
```
karlnapf: Slightly weight English:
We demonstrate a pipeline consisting of a transformer ..., and LibLinear for binary classification.
(maybe also link to the liblinear cookbook)

doc/cookbook/source/examples/evaluation/cross_validation_pipeline.rst
```diff
@@ -0,0 +1,34 @@
...
+
+Imagine we have files with training data. We create :sgclass:`CDenseFeatures` (here 64 bit floats aka RealFeatures) as
```
karlnapf: I know this is copy pasted, but this is now outdates when we use factories.
Just say: "We create CFeatures and CLabels via loading from files"

doc/cookbook/source/examples/evaluation/cross_validation_pipeline.rst
```diff
@@ -0,0 +1,34 @@
...
+
+.. sgexample:: cross_validation_pipeline:create_features
+
+
+We use :sgclass:`CPruneVarSubMean` to normalize the features and then use :sgclass:`CLibLinear` for classification.
```
karlnapf: This sentence basically repeats the intro, so I would just remove it

doc/cookbook/source/examples/evaluation/cross_validation_pipeline.rst
```diff
@@ -0,0 +1,34 @@
...
+The transformer and the machine are chained as a :sgclass:`CPipeline`.
```
karlnapf: We create a Cpipeline, and chain the transformer and the classifier.

doc/cookbook/source/examples/evaluation/cross_validation_pipeline.rst
```diff
@@ -0,0 +1,34 @@
...
+
+.. sgexample:: cross_validation_pipeline:create_pipeline
+
+Next, we initialize a splitting strategy :sgclass:`CStratifiedCrossValidationSplitting` to divide the dataset into :math:`k-` folds for the :math:`k-` fold cross validation.
+We also have to decide on an evaluation criterion class (from :sgclass:`CEvaluation`) to evaluate the performance of the trained models.
```
karlnapf: see CEvaluation

karlnapf: last word model (singular)

doc/cookbook/source/examples/evaluation/cross_validation_pipeline.rst
karlnapf: actually, I would mention that "we first chain all transformers, and then finalize the pipeline with the classifier" (since you use "then")

doc/cookbook/source/examples/evaluation/cross_validation_pipeline.rst
```diff
@@ -0,0 +1,34 @@
...
+In this case, we use :sgclass:`CAccuracyMeasure`.
+We then instantiate :sgclass:`CCrossValidation` and set the number of cross validation's runs.
```
karlnapf: Please also mention something like "The pipeline instance behaves just like a machine and this can be directly passed to CCrossValidation"

examples/meta/src/evaluation/cross_validation_pipeline.sg
```diff
@@ -0,0 +1,33 @@
...
+Transformer subMean = transformer("PruneVarSubMean")
+Machine svm = machine("LibLinear")
+
+PipelineBuilder builder()
+builder.over(subMean)
+Pipeline pipeline = builder.then(svm)
+#![create_pipeline]
+
+#![create_cross_validation]
+StratifiedCrossValidationSplitting splitting_strategy(labels_train, 2)
```
karlnapf: dont we have factories for those things?
I would prefer if all newly added examples would fully use the new api so we dont have to refactor later

examples/meta/src/evaluation/cross_validation_pipeline.sg
```diff
@@ -0,0 +1,33 @@
...
+Features feats_test = features(f_feats_test)
+Labels labels_train = labels(f_labels_train)
+Labels labels_test = labels(f_labels_test)
+#![create_features]
+
+#![create_pipeline
+Transformer subMean = transformer("PruneVarSubMean")
+Machine svm = machine("LibLinear")
+
+PipelineBuilder builder()
```
karlnapf: Same here, could you pls use a factory for this. We dont want to use constructors in the examples

examples/meta/src/evaluation/cross_validation_pipeline.sg
```diff
@@ -0,0 +1,33 @@
...
+builder.over(subMean)
+Pipeline pipeline = builder.then(svm)
+#![create_pipeline]
+
+#![create_cross_validation]
+StratifiedCrossValidationSplitting splitting_strategy(labels_train, 2)
+Evaluation evaluation_criteron = evaluation("AccuracyMeasure")
+CrossValidation cross(pipeline, feats_train, labels_train, splitting_strategy, evaluation_criteron, False)
```
karlnapf: factory if possible (this might be more tricky)

examples/meta/src/evaluation/cross_validation_pipeline.sg
vinx13: we don't have factory for `CSplittingStrategy`

examples/meta/src/evaluation/cross_validation_pipeline.sg
vinx13: could you explain how to use factory here?
we could create a factory `PipelineBuilder* pipeline_builder()`, or?

examples/meta/src/evaluation/cross_validation_pipeline.sg
karlnapf: Would you mind creating one? That should be pretty easy!

examples/meta/src/evaluation/cross_validation_pipeline.sg
karlnapf: yes exactly, though I would just name it `pipeline`

karlnapf:
Let me know if the factory creating worked....

examples/meta/src/evaluation/cross_validation_pipeline.sg
vinx13: sure

examples/meta/src/evaluation/cross_validation_pipeline.sg
```diff
@@ -14,7 +14,7 @@ Labels labels_test = labels(f_labels_test)
 Transformer subMean = transformer("PruneVarSubMean")
 Machine svm = machine("LibLinear")
 
-PipelineBuilder builder()
+PipelineBuilder builder = pipeline()
```
karlnapf: exactly like this! :)

karlnapf: BTW why is the C++ ype not just called Pipeline and the Machine called PipelineMachine?

examples/meta/src/evaluation/cross_validation_pipeline.sg
vinx13: @karlnapf since there are multiple subclasses of `CSplittingStrategy`, are we going to do some string comparison by name?

examples/meta/src/evaluation/cross_validation_pipeline.sg
vinx13: i think `Pipeline` and `PipelineMachine` might be a bit confusing, while `PipelineBuilder` can indicate its usage as a builder

examples/meta/src/evaluation/cross_validation_pipeline.sg
karlnapf: check the factory `machine` in `factory.h, you just have to add some macro lines
the call should be `splitting_strategy("StratifiedCrossValidationSplitting", labels=labels_train, k=2)`

examples/meta/src/evaluation/cross_validation_pipeline.sg
karlnapf: TBH, I dont really agree, I think `then` should actually return `CMachine` (since the subsequent object will be used in this fashion) ... why do we need to know that a machine is a pipeline if it behaves as a machine?
And then `Pipeline` is the thing that builds the stuff.
Makes for a cleaner API imo

@vigsterkr @lisitsyn @iglesias what are your thoughts?

vigsterkr: @karlnapf are we talking now about PipelineMachine and PipelineBuilder, or why `then` returns Pipeline* ? :) imo those are different things or?

vigsterkr: in case `then` would return CMachine, then the whole thing becomes opaque, and you would never be able to get out the elements of pipeline...
unless we could register `std::vector<std::pair<std::string, variant<CTransformer*, CMachine*>>>  m_stages;` and then use getter... which is currently a nogo for multiple reasons :)

karlnapf: Sorry I wasnt clear ;)

What I suggested is this: CPipeline inherits from CMachine, so it *is* and behaves as a machine. Therefore I don't see a need to expose a separate class (say to swig). All we do is to call train/etc. I would call this thing CPipelineMachine and hide it from the outside world.

Next, the class that constructs the above machine (whose name is now hidden from the API) could be called CPipeline now, the thing where one can add stages etc and the finalise as a CMachine.

Does it make any sense?

karlnapf: So CPipeline -> CPipelineMachine, CPipelineBuilder ->CPipeline, then returns CMachine

vigsterkr: imo builder is just clearer but i dont have any strong feelings about it...
note my second comment about extraction and observation of pipeline.

karlnapf: yes the stages thing is a problem

vigsterkr: i.e. until we cannot register and cleanly extract stages from a pipeline this sort of explicit exposure is required :) otherwise the whole thing becomes totally opaque once built

karlnapf: got it! thx for clarifying.
I still would slightly prefer different names (CPipelineMachine, CPipeline) but it is only a minor difference, also no strong feelings. Can leave it as it is

examples/meta/src/evaluation/cross_validation_pipeline.sg
vinx13: so we need to do initialization in `init()` method, instead of the constructor
e.g. https://github.com/shogun-toolbox/shogun/blob/1be1dbbf6fdce20e73901d205cee253f2d4cd873/src/shogun/evaluation/StratifiedCrossValidationSplitting.cpp#L21
this need refactor right?

karlnapf: ah yes, this will be moved into the method that is called from outside, i.e. `build_subsets` ... putting it into a helper method makes sense

karlnapf:
rebase and merge :)

doc/cookbook/source/examples/evaluation/cross_validation_pipeline.rst
```diff
@@ -0,0 +1,33 @@
...
+-------
+Example
+-------
+We demonstrate a pipeline consisting of a transformer :sgclass:`CPruneVarSubMean` for normalizing the features, and a machine :sgclass:`CLibLinear` for binary classification.
+
+We create :sgclass:`CFeatures` and :sgclass:`CLabels` via loading from files
+
+.. sgexample:: cross_validation_pipeline:create_features
+
+We first chain all transformers, and then finalize the pipeline with the classifier.
```
iglesias: Is it one as said above? Then change or remove "all".

examples/meta/src/evaluation/cross_validation_pipeline.sg
```diff
@@ -0,0 +1,32 @@
...
+File f_labels_test = csv_file("../../data/classifier_binary_2d_linear_labels_test.dat")
+
+#![create_features]
+Features feats_train = features(f_feats_train)
+Features feats_test = features(f_feats_test)
+Labels labels_train = labels(f_labels_train)
+Labels labels_test = labels(f_labels_test)
+#![create_features]
+
+#![create_pipeline
```
iglesias: Missing ]?

examples/meta/src/evaluation/cross_validation_pipeline.sg
```diff
@@ -0,0 +1,32 @@
...
+Transformer subMean = transformer("PruneVarSubMean")
+Machine svm = machine("LibLinear")
+
+PipelineBuilder builder = pipeline()
+builder.over(subMean)
+Pipeline pipeline = builder.then(svm)
```
iglesias: Why does the API contain these two builder and pipeline concepts? What about just a pipeline where steps are added (and of course the order of addition matters).

vinx13: the idea is to separate construction of pipeline into a single class so that the pipeline object is immutable, and then we don't need to verify elements of pipeline everytime

karlnapf: This is pretty much the same thing that got me confused about the builder vs the pipeline.

vinx13: the second one, `Pipeline` can be `Machine` here because that's enough, we use it as machine in xval. however, if we want to get elements in the pipeline, we still need `Pipeline` type for now

karlnapf: what about adding a factory for it? So users can call
`PipelineMachine(machine)`? 
Alternatively, for the c++ folks, there is `as`

vigsterkr:
We always have these multipurpose classes where multiple functionality/concept is crammed into one class... here’s a clear cut between a builder pattern and cmachine

examples/meta/src/evaluation/cross_validation_pipeline.sg
iglesias: I understand that there might be mutability reason justifying two types. Why both part of the interface though?

vigsterkr:
You need the builder for building, but u would like to query a trained pipeline for various stages of it as well. Or what do you mean?

iglesias:
What I was wondering was why a user needs to use two different classes to
use a pipeline: e.g. the pipeline_builder to build/construct a pipeline
object, and then this pipeline object to do all the preprocessing,
cross-validation, etc, etc, she wants.

Of course I also understand your point. It is not clean if a class
implements many different concepts altogether. It gets confusing quickly,
classes get bloated, etc.

It is just a trade-off I think.

To clarify :-) These were not comments critizing any design choice.
Completely far from that! I asked to gain understanding.

examples/meta/src/evaluation/cross_validation_pipeline.sg
```diff
@@ -0,0 +1,32 @@
...
+Labels labels_test = labels(f_labels_test)
+#![create_features]
+
+#![create_pipeline]
+Transformer subMean = transformer("PruneVarSubMean")
+Machine svm = machine("LibLinear")
+
+PipelineBuilder builder = pipeline()
+builder.over(subMean)
+Machine pipeline = builder.then(svm)
```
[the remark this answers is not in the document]
vinx13: yes this overload the factory name, the factory works in this case. thanks for letting me know.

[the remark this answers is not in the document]
karlnapf: I like the flow of this, so good to merge from my side

karlnapf:
Can you rebase data and then we can merge this as well!

vinx13:
@karlnapf As I can remember, previously we faced the choice on whether pipeline builder should return `CMachine*` or `CPipeline*`. The first choice, returning `CMachine*` and then providing a `pipeline` factory, works though being ugly. According to @vigsterkr 's idea, in my last commit I tried to make `CPipeline` as base class, but got some different errors on travis `  what():  [1;31m[ERROR][0m In file /opt/shogun/src/shogun/base/SGObject.h line 369: Cannot put parameter CrossValidation::machine of type shogun::CMachine*, incompatible provided type shogun::CPipeline*.` Maybe @lisitsyn has idea?

karlnapf:
Ah I remember now.
Well the real question is whether Pipeline should be part of the modular interfaces (i.e. the base types) or not.

karlnapf:
Ok some more thought:
you get the above error because `put` is strict with types: no subclasses allowed.
Since that won't change, there are the options to (braindump)

* hardcode a pipeline in `put` that checks and casts to `CMachine` (ugly and not scalable, but does it need to be?)
* make the builder return a `CMachine` and whoever wants to extract the elements of the pipeline needs to do `.as_pipeline()`. E.g. @gf712 doing the work on openml would have to use `get_name()` rather than `sg.Pipeline` to check the type first and then use the `as`. BTW `as` is something that we discussed at the hackathon in budapest as something that is needed anyways, see [here](https://github.com/shogun-toolbox/shogun/wiki/GSoC_2018_project_usability). As most users don't want to inspect the pipeline they just built but rather call `train` I think this is a reasonable compromise (although not C++ style). (Veto'd by @vigsterkr but also liked by @iglesias )
* template crossvalidation and use type traits to restrict to `CMachine` and `CPipeline` and whatever other learner we have (transformers maybe). Then have a factory for each that instantiates the corresponding xvalidation instance and then accepts pipeline or machine. (unnecessary complex)
* have a factory for cross validation instances that either accepts machine or pipelines. This would allow the set the `machine` member, but `put` still doesn't work for `CPipeline.

vinx13:
I made machine a positional argument of machine_evaluation factory and it works. But I got another error https://dev.azure.com/shogunml/shogun/_build/results?buildId=328&view=logs&jobId=089c709a-44eb-5f6e-96e7-15e9ee1ff5bf&taskId=5899c1e7-1112-57da-0a81-46bbe225a6c3&lineStart=81&lineEnd=82&colStart=1&colEnd=1
@karlnapf Any idea what I missed?

gf712:
> I made machine a positional argument of machine_evaluation factory and it works. But I got another [error](https://dev.azure.com/shogunml/shogun/_build/results?buildId=328&view=logs&jobId=089c709a-44eb-5f6e-96e7-15e9ee1ff5bf&taskId=5899c1e7-1112-57da-0a81-46bbe225a6c3&lineStart=81&lineEnd=82&colStart=1&colEnd=1)
> 
> ```
> splitting_strategy() takes no keyword arguments
> ```
> @karlnapf Any idea what I missed?

@vinx13 I think that is related to me changing python.json in the generator. Basically it doesn't replace the expressions with consecutive put calls, but assumes you are using the python wrapper I wrote. So you need to add splitting_strategy as a string here:
https://github.com/shogun-toolbox/shogun/blob/e01874993fcd5feb836490e5a767123ccc462eca/src/interfaces/python/swig_typemaps.i#L1356-L1365
This allows you to use **kwargs in factories.
I am not sure if that is the issue though, but it's worth a try!

vinx13:
it works, @karlnapf @gf712 can you take a look now?

karlnapf:
Cool! I’ll check it out soon!

src/interfaces/python/swig_typemaps.i
```diff
@@ -1371,9 +1373,9 @@ def _internal_factory_wrapper(object_name, new_name, docstring=None):
     via .put
     """
     _obj = getattr(sys.modules[__name__], object_name)
-    def _internal_factory(name, **kwargs):
+    def _internal_factory(name, *args, **kwargs):
```
gf712: It might be worth mentioning in the docstring that *args are passed to the object at instantiation?
It's a good idea to handle constructors with multiple arguments!

karlnapf: agreed!

src/shogun/util/factory.h
```diff
@@ -267,5 +272,28 @@ namespace shogun
...
+	}
+
+	CMachineEvaluation* machine_evaluation(const std::string& name, CMachine* machine)
+	{
+		auto obj = create_object<CMachineEvaluation>(name.c_str());
+		obj->put("machine", machine);
+		return obj;
+	}
+
+	CMachineEvaluation* machine_evaluation(CSGObject* obj)
```
karlnapf: I am not so keen on this one....but did you actually use it?

vinx13: no...but every other factory has this, do we need it?

karlnapf: Ah! Of course, you can actually even use this as the "as_machine"

examples/meta/generator/targets/python.json
```diff
@@ -61,7 +61,8 @@
             "get_int_vector": "$object.get($arguments)",
             "get_real": "$object.get($arguments)",
             "get_real_vector": "$object.get($arguments)",
-            "get_real_matrix": "$object.get($arguments)"
+            "get_real_matrix": "$object.get($arguments)",
+            "put_machine": "$object.put($arguments)"
```
karlnapf: this will not work as we cannot extract the arguments yet and you would need to pass the second argument to the `machine` factory

karlnapf: we can do this once #4490 is solved

examples/meta/src/evaluation/cross_validation_pipeline.sg
```diff
@@ -0,0 +1,35 @@
...
```
karlnapf: This type here should be "Pipeline" as that is what the builder returns (and now Pipeline is part of swig as well)

examples/meta/src/evaluation/cross_validation_pipeline.sg
```diff
@@ -0,0 +1,35 @@
...
+#![create_pipeline]
+
+#![create_cross_validation]
+SplittingStrategy strategy = splitting_strategy("StratifiedCrossValidationSplitting", labels=labels_train, num_subsets=2)
+Evaluation evaluation_criterion = evaluation("AccuracyMeasure")
+MachineEvaluation cross = machine_evaluation("CrossValidation", features=feats_train, labels=labels_train, splitting_strategy=strategy, evaluation_criterion=evaluation_criterion, autolock=False, num_runs=2)
+cross.put_machine("machine", pipeline)
```
karlnapf: change to `cross.put(machine(pipeline))` and it should work
we can change that to put_machine later once #4490 is in

examples/meta/src/evaluation/cross_validation_pipeline.sg
```diff
@@ -0,0 +1,35 @@
...
+
+PipelineBuilder builder = pipeline()
+builder.over(subMean)
+Machine pipeline = builder.then(svm)
+#![create_pipeline]
+
+#![create_cross_validation]
+SplittingStrategy strategy = splitting_strategy("StratifiedCrossValidationSplitting", labels=labels_train, num_subsets=2)
+Evaluation evaluation_criterion = evaluation("AccuracyMeasure")
+MachineEvaluation cross = machine_evaluation("CrossValidation", features=feats_train, labels=labels_train, splitting_strategy=strategy, evaluation_criterion=evaluation_criterion, autolock=False, num_runs=2)
```
karlnapf: OT: Can I suggest that we don't provide features and labels to crossvalidation but instead as parameters of `CMachineEvaluation::evaluate(CFeatures*, CLabels*) ? Different PR though
@lisitsyn @vigsterkr @iglesias

iglesias: Hey mate, I can't say much based on this snippet. What's your point? Also based on the last line, CrossValidation here seems to be a type of MachineEvaluation, no?

vinx13:
why `shogun/build/examples/meta/cpp/evaluation/cross_validation_pipeline.cpp:61:27: error: ‘class shogun::CSplittingStrategy’ has no member named ‘evaluate’`, has the api been changed?

examples/meta/generator/targets/cpp.json
```diff
@@ -96,7 +96,8 @@
             "get_int_vector": "$object->get<SGVector<int32_t>>($arguments)",
             "get_real": "$object->get<float64_t>($arguments)",
             "get_real_vector": "$object->get<SGVector<float64_t>>($arguments)",
-            "get_real_matrix": "$object->get<SGMatrix<float64_t>>($arguments)"
+            "get_real_matrix": "$object->get<SGMatrix<float64_t>>($arguments)",
+            "put_machine": "$object->put<CMachine>($arguments)"
         },
```
karlnapf: We don’t need this anymore now right?

src/interfaces/python/swig_typemaps.i
karlnapf: Not needed abymore

src/shogun/machine/Pipeline.cpp
```diff
@@ -256,4 +256,9 @@ namespace shogun
 	{
 		return get_machine()->get_machine_problem_type();
 	}
+
```
karlnapf: Can’t this be in factory?

tests/unit/machine/Pipeline_unittest.cc
```diff
@@ -48,7 +48,8 @@ TEST_F(PipelineTest, fit_predict)
 	auto pipeline = some<CPipelineBuilder>()
 	                    ->over(transformer1)
 	                    ->over(transformer2)
-	                    ->then(machine);
+	                    ->then(machine)
+	                    ->as<CPipeline>();
```
karlnapf: Is that needed?

karlnapf:
Something about include paths is broken in the example ...

doc/cookbook/source/examples/evaluation/cross_validation_pipeline.rst
```diff
@@ -0,0 +1,33 @@
+============================
```
karlnapf: minor. the `====` should be as long as the text

doc/cookbook/source/examples/evaluation/cross_validation_pipeline.rst
```diff
@@ -0,0 +1,33 @@
...
+
+We first chain transformers, and then finalize the pipeline with the classifier.
+
+.. sgexample:: cross_validation_pipeline:create_pipeline
+
+Next, we initialize a splitting strategy :sgclass:`CStratifiedCrossValidationSplitting` to divide the dataset into :math:`k-` folds for the :math:`k-` fold cross validation.
+We also have to decide on an evaluation criterion class (see :sgclass:`CEvaluation`) to evaluate the performance of the trained model.
+In this case, we use :sgclass:`CAccuracyMeasure`.
+We then instantiate :sgclass:`CCrossValidation` and set the number of cross validation's runs.
+The pipeline instance behaves just like a :sgclass:`CMachine` and this can be directly passed to :sgclass:`CCrossValidation`.
```
karlnapf: This is not exactly true anymore, as we need to call the factory now. I suggest we just avoid this and say
"We pass the pipeline instance, which behaves just like a :sgclass:`CMachine`, to :sgclass:`CCrossValidation`."

src/shogun/util/factory.h
```diff
@@ -267,5 +274,21 @@ namespace shogun
 		SG_REF(result);
 		return result;
 	}
+
+	/** Create a pipeline builder.
+	 * See also CPipelineBuilder and CPipeline.
+	 * @return new instance of CPipelineBuilder
+	 */
+	CPipelineBuilder* pipeline()
```
karlnapf: @gf712 would you be up for making the definitions of the factory methods a bit neater? :)

karlnapf:
Cool!
Thanks for the big amount of work. It is awesome to have this in now :)

src/shogun/util/factory.h
gf712: sure! what do you have in mind?

karlnapf: Actually, just checked, not sure `factory.h` can even be made that much simpler?

PR #4384, opened 2018-07-23
Simplify LRR via using CDenseFeatures interface for cov, gram, sum

karlnapf:


src/shogun/features/DenseFeatures.cpp
```diff
@@ -1001,6 +1001,42 @@ template< class ST > CDenseFeatures< ST >* CDenseFeatures< ST >::obtain_from_gen
 	return (CDenseFeatures< ST >*) base_features;
 }
 
+template <typename ST>
```
karlnapf: I think it would be even cooler if we moved those methods to `CDotFeatures` and then specialize them in here for batch computation if not subsets are set (otherwise call CDotFeatures::cov which has an iterative implementation)

src/shogun/regression/LinearRidgeRegression.cpp
```diff
@@ -57,32 +83,26 @@ bool CLinearRidgeRegression::train_machine(CFeatures* data)
...
-
-	auto decomposition = linalg::cholesky_factor(kernel_matrix);
-	y = linalg::cholesky_solver(decomposition, y);
-	set_w(y);
-
-	auto x_bar = linalg::colwise_sum(feats_matrix);
-	linalg::scale(x_bar, x_bar, 1.0 / ((float64_t)x_bar.size()));
-	auto intercept = linalg::mean(lab) - linalg::dot(lab, x_bar);
-	set_bias(intercept);
+	auto feats = data->as<CDenseFeatures<float64_t>>();
```
karlnapf: the idea if this patch is that inside the algorithm, we only work against a very high level interface of CFeatures (deals with subsets etc).
For labels, we always get the vector (which creates a copy for subsets). I think since it is just a vector, this is fine, although we could also think about adding an interface to CDot/DenseFeatures that computes the dot product with a CDenseLabels which then is batch without subsets and iterative(no copy) if there are....probably too complicated for the marginal benefits of avoiding a single vector copy

src/shogun/regression/LinearRidgeRegression.cpp
```diff
@@ -57,32 +83,26 @@ bool CLinearRidgeRegression::train_machine(CFeatures* data)
...
+	auto y = regression_labels(m_labels)->get_labels();
+	auto N = feats->get_num_vectors();
+	auto D = feats->get_num_features();
+
+	SGVector<float64_t> w;
+	if (N >= D)
+		w = solve_using_covariance<float64_t>(feats, y, m_tau);
```
karlnapf: also generalise LRR for cases where D>N

src/shogun/regression/LinearRidgeRegression.cpp
```diff
@@ -57,32 +83,26 @@ bool CLinearRidgeRegression::train_machine(CFeatures* data)
...
+	else
+		w = solve_using_gram<float64_t>(feats, y, m_tau);
+
+	set_w(w);
+
+	if (m_use_bias)
```
karlnapf: make bias compitation optional which is inline with other shogun methods

src/shogun/regression/LinearRidgeRegression.cpp
```diff
@@ -57,32 +83,26 @@ bool CLinearRidgeRegression::train_machine(CFeatures* data)
...
+	{
+		auto x_bar = feats->sum();
+		linalg::scale(x_bar, x_bar, ((float64_t)1.0) / N);
+		auto intercept = linalg::mean(y) - linalg::dot(w, x_bar);
```
iglesias: For consistency in the computations of features and labels, is it possible to do some `linalg::mean` of features?

src/shogun/regression/LinearRidgeRegression.cpp
```diff
@@ -57,32 +83,26 @@ bool CLinearRidgeRegression::train_machine(CFeatures* data)
...
-	auto x_bar = linalg::colwise_sum(feats_matrix);
-	linalg::scale(x_bar, x_bar, 1.0 / ((float64_t)x_bar.size()));
-	auto intercept = linalg::mean(lab) - linalg::dot(lab, x_bar);
-	set_bias(intercept);
+	auto feats = data->as<CDenseFeatures<float64_t>>();
+	auto y = regression_labels(m_labels)->get_labels();
+	auto N = feats->get_num_vectors();
+	auto D = feats->get_num_features();
+
+	SGVector<float64_t> w;
```
iglesias: So the model contains both w and b and what we are doing is first find w, ignoring that b exists, and then find b provided the found w.

Is this standard practice in linear regression? I am a bit surprised that the equations to find w do not depend on the b since the gradient of the loss wrt to w has b.

iglesias: Ok. My question was probably wrong. Even if b does not appear explicitly in the expression for w, that does not mean it is not in there. It has just been substituted.

However, then the expression for w should be in principle different for the cases (1) w/ bias; (2) w/o bias. No?

karlnapf: You are right! the expression for w was wrong, it assumed no bias. I am fixing it right now

src/shogun/regression/LinearRidgeRegression.cpp
```diff
@@ -32,8 +32,66 @@ CLinearRidgeRegression::CLinearRidgeRegression(float64_t tau, CDenseFeatures<flo
...
+	}
+
+	SGVector<float64_t> w;
+	if (N >= D)
+	{
+		SGMatrix<T> cov = feats->cov();
+		linalg::add_ridge(cov, m_tau);
+		auto L = linalg::cholesky_factor(cov);
+
+		if (m_use_bias)
```
karlnapf: @iglesias voila! :)
If you could do the math and confirm this, that would be neat! :)

iglesias:
Even nicer exercise for gsocers ;-)

It would fit very nicely in a Jupyter nb.

src/shogun/features/DenseFeatures.cpp
```diff
@@ -19,6 +19,19 @@
 #include <algorithm>
 #include <string.h>
 
+#define ASSERT_FLOATING_POINT                                                  \
```
karlnapf: @lisitsyn @iglesias @vigsterkr our template style for features doesnt allow me to use type traits to define member functions of templated features, as all template types are instantiated. This is why I added this runtime check for floating point numbers. We can remove that once the features are cleaned up a bit. It for now just stops callers from doing nonsense (compute mean on bools or ints)

src/shogun/regression/LinearRidgeRegression.cpp
```diff
@@ -32,57 +32,64 @@ CLinearRidgeRegression::CLinearRidgeRegression(float64_t tau, CDenseFeatures<flo
...
 }
 
-bool CLinearRidgeRegression::train_machine(CFeatures* data)
+template <typename T>
+bool CLinearRidgeRegression::train_machine_templated(
+    const CDenseFeatures<T>* feats)
 {
-    REQUIRE(m_labels,"No labels set\n");
-
-    if (!data)
```
karlnapf: thanks to @shubham808 s work, we can now delete such redundant codes, while making algorithms fully templated

src/shogun/regression/LinearRidgeRegression.h
```diff
@@ -36,11 +37,23 @@ namespace shogun
...
+	 * In practice, an additional bias term
+	 * \f$b=\frac{1}{N}\sum_{i=1}^{N}y_{i}-{\bf w}\cdot\bar{\mathbf{x}}\f$
 	 * for
-	 * \f$\bar{\mathbf{x}}=\frac{1}{N}\sum_{i=1}^{N}{\bf x}_{i}\f$.
+	 * \f$\bar{\mathbf{x}}=\frac{1}{N}\sum_{i=1}^{N}{\bf x}_{i}\f$
+	 * can also be included, which effectively centers the \f$X\f$ before
+	 * computing the solution.
 	 */
-	class CLinearRidgeRegression : public CLinearMachine
+	class CLinearRidgeRegression
```
karlnapf: @shubham808 tada! :)

shubham808: looks smooth :)

src/shogun/regression/LinearRidgeRegression.cpp
```diff
@@ -32,57 +33,64 @@ CLinearRidgeRegression::CLinearRidgeRegression(float64_t tau, CDenseFeatures<flo
...
-	set_w(y);
-
-	auto x_bar = linalg::colwise_sum(feats_matrix);
-	linalg::scale(x_bar, x_bar, 1.0 / ((float64_t)x_bar.size()));
-	auto intercept = linalg::mean(lab) - linalg::dot(lab, x_bar);
-	set_bias(intercept);
+	auto N = feats->get_num_vectors();
+	auto D = feats->get_num_features();
+
+	auto y = regression_labels(m_labels)->get_labels().as<T>();
```
karlnapf: copying the labels vector to get the correctly typed version of y (consequence of float64 labels vs templated algorithm impl)

src/shogun/regression/LinearRidgeRegression.cpp
```diff
@@ -32,57 +33,64 @@ CLinearRidgeRegression::CLinearRidgeRegression(float64_t tau, CDenseFeatures<flo
...
+		if (m_use_bias)
+			SG_NOTIMPLEMENTED
+
+		SGMatrix<T> gram = feats->gram();
+		linalg::add_ridge(gram, (T)m_tau);
+		auto L = linalg::cholesky_factor(gram);
+		auto b = linalg::cholesky_solver(L, y);
+		w = feats->dot(b);
+	}
+	set_w(w.as<float64_t>());
```
karlnapf: copying the labels vector to get the correctly typed version of y (consequence of float64 w vs templated algorithm impl)

karlnapf:
Need to ignore the stylechecker since class_list.py assumes that classes are defined in a single line, so cannot apply this: https://travis-ci.org/shogun-toolbox/shogun/jobs/410368055#L746

src/shogun/regression/LinearRidgeRegression.cpp
```diff
@@ -32,57 +33,64 @@ CLinearRidgeRegression::CLinearRidgeRegression(float64_t tau, CDenseFeatures<flo
...
+	auto N = feats->get_num_vectors();
+	auto D = feats->get_num_features();
+
+	auto y = regression_labels(m_labels)->get_labels().as<T>();
+
+	SGVector<T> x_mean;
+	T y_mean;
+	if (m_use_bias)
+	{
+		x_mean = feats->mean();
```
iglesias: Would a linalg::mean(feats) would make sense from the point of view of API happiness? :-P

src/shogun/regression/LinearRidgeRegression.cpp
```diff
@@ -32,57 +33,64 @@ CLinearRidgeRegression::CLinearRidgeRegression(float64_t tau, CDenseFeatures<flo
...
+		y_mean = linalg::mean(y);
+	}
+
+	SGVector<T> w;
+	if (N >= D)
+	{
+		SGMatrix<T> cov = feats->cov();
+		linalg::add_ridge(cov, (T)m_tau);
```
iglesias: Cast more solicit perhaps? If there has to be cast using T I think it should be more fair to do static_cast.

iglesias: Explicit, not solicit..... Autocorrectors xD

src/shogun/regression/LinearRidgeRegression.cpp
karlnapf: Good point!
So the idea is that features are a quite different type of object than matrices and vectors. Even in the case for dense features, e.g. think of an active subset. This is why such an interface for CDenseFeatures (can be moved up later) makes sense: via only offering an iterator interface and some convenience functions (as opposed to direct matrix access), one can keep things clean.
linalg on the other hand works on linear memory blocks, such as SGMatrix.
The PR is to exemplify these concepts

src/shogun/regression/LinearRidgeRegression.cpp
```diff
@@ -32,57 +33,64 @@ CLinearRidgeRegression::CLinearRidgeRegression(float64_t tau, CDenseFeatures<flo
...
+		if (m_use_bias)
+			linalg::rank_update(cov, x_mean, (T)-N);
```
iglesias: Why the parentheses between T? :-)

src/shogun/regression/LinearRidgeRegression.cpp
```diff
@@ -32,57 +33,65 @@ CLinearRidgeRegression::CLinearRidgeRegression(float64_t tau, CDenseFeatures<flo
...
-
-	auto x_bar = linalg::colwise_sum(feats_matrix);
-	linalg::scale(x_bar, x_bar, 1.0 / ((float64_t)x_bar.size()));
-	auto intercept = linalg::mean(lab) - linalg::dot(lab, x_bar);
-	set_bias(intercept);
+	auto N = feats->get_num_vectors();
+	auto D = feats->get_num_features();
+
+	auto y = regression_labels(m_labels)->get_labels().as<T>();
+	T tau = m_tau;
```
karlnapf: I avoid the static_cast or c-cast now using this explicit thing. Makes it more readable indeed
For the numerical things (1.0/N) I prefer the C cast tbh

src/shogun/regression/LinearRidgeRegression.cpp
iglesias: All righty! Do we have already some case using these feature iterators?

karlnapf: Well `feats->cov()` kind of does something in these lines.
If there is no subset set, it does a speedy matrix multiplication. If there is a subset set, it first copies the matrix into a continuous one and then does the product. An improvement would be to do this without the copy, i.e. an iterative procedure like outlines in `CDotFeatures::cov`. However, DotFeatures are not typed and always return float64, and I am puzzled how to move the `mean, cov, gram` methods into that class ....

src/shogun/features/DenseFeatures.h
```diff
@@ -303,6 +303,46 @@ template<class ST> class CDenseFeatures: public CDotFeatures
...
+	 *
+	 *\f[
+	 * X X^\top
+	 * \f]
+	 *
+	 * where \f$X\f$ is the \f$DxN\f$ dimensional feature matrix with \f$N\f$
+	 * feature vectors of dimension \f$D\f$.
+	 */
+	SGMatrix<ST> cov() const;
+	/** Computes the \f$fNxN\f$ (uncentered, un-normalized) gram matrix of
```
iglesias: Completely minor: missing newline.

src/shogun/features/DenseFeatures.h
```diff
@@ -303,6 +303,46 @@ template<class ST> class CDenseFeatures: public CDotFeatures
...
+	/** Computes the \f$DxD\f$ (uncentered, un-normalized) covariance matrix
+	 *
+	 *\f[
+	 * X X^\top
+	 * \f]
+	 *
+	 * where \f$X\f$ is the \f$DxN\f$ dimensional feature matrix with \f$N\f$
+	 * feature vectors of dimension \f$D\f$.
+	 */
+	SGMatrix<ST> cov() const;
```
iglesias: Would it make sense to call it noncentered_cov or something similar that makes the fact that is not centered explicit?

karlnapf: Since I first met @lisitsyn I am in favour of short function names.
I was thinking about giving it a boolean flag for centring at first but then didnt do it. Could re-introduce it in a case where this is needed.
Generally, I would prefer to move towards making clear cuts of functionality, i.e. have a data centring module come first, and then doing the cov.
A preprocessor could also be responsible for this (at the cost of speedy bulk operations), then the code becomes less convoluted (mean is computed/removed at many places in the code)
What do you think>?

iglesias: I think having a function called covariance that does not compute the covariance can lead to confusion :-P

karlnapf: Well it computes the covariance, just for a certain case of input.
It is also not called covariance_for_column_major_matrix ;)
If you feel strong, I can change it, otherwise I would just outsource mean computations into different code..
On another note, I wanna mimic the `dot` API of DotFeatures, but just templated. What are your thoughts on that?

iglesias:
Of course it is good, no strong feelings :-)

I see the DenseFeatures::dot is a template already, and I am guessing the
call to linalg it does is to another template as well. What's your idea?

karlnapf:
We have `virtual float64_t CDotFeatures::dot=0`, which is a quite cool design, as independently of any feature type, we can write code and algorithms that work with these dot product features.
The problem, however, is the return type. Think about linear regression and `cov`, and 32bit features. We could of course write an iterative cov implementation that fills a 32bit cov matrix using the above API and elementwise casting. We cannot, however, do a batch version without copying the matrix. That is the first issue. The second issue is that the API is not `const`, due to the on-the-fly features that are computed in a non const inplace way.
So maybe we can build a second version of the API (keeping the old intact), that serves a similar purpose, but that is both const and templated. Thoughts?

karlnapf:
@vinx13 do you have any idea on this error: https://travis-ci.org/shogun-toolbox/shogun/jobs/410945599#L2476

vinx13:
i don't have seen that before, but i will try to fix it

karlnapf:
The thing is I cannot reproduce this locally or on the buildbot. Similar error occur when gtest.h is not included as the first include in the unit tests

vinx13:
I can reproduce locally with g++ 5.4.1

iglesias:
Sure, that sounds good. At the very least, to start experiment.

vinx13:
it is
#include <shogun/mathematics/linalg/LinalgNamespace.h> in DenseFeatures.h that causes the error about vairant. but i still don't see why.

karlnapf:
Aaaah it is probably the include order .... Can you try moving it around?

src/shogun/features/DenseFeatures.h
```diff
@@ -11,14 +11,14 @@
...
-#include <shogun/io/File.h>
 #include <shogun/features/DotFeatures.h>
 #include <shogun/features/StringFeatures.h>
+#include <shogun/io/File.h>
+#include <shogun/lib/Cache.h>
 #include <shogun/lib/DataType.h>
-
 #include <shogun/lib/SGMatrix.h>
+#include <shogun/lib/common.h>
+#include <shogun/mathematics/linalg/LinalgNamespace.h>
```
vinx13: this can be moved to .cpp file so that the inclusion order problem can be fixed

karlnapf:
ah yes of course, this was since I had some of the implementations in the header for some short time.
@iglesias it is scary indeed. I think this is due to the non-std template stuff we are doing in CDenseFeatures.....it is definitely not a guard. Same with gtest.h btw, has to be included first otherwise it blows.
@vinx13 thanks for the tip, I have updated, lets see :)

PR #4386, opened 2018-07-27
Feature/neural nets factory and cookbook

shubham808:


karlnapf:
Looks good, how does it work?

examples/meta/src/neural_nets/convolutional_net_classification.sg
```diff
@@ -0,0 +1,46 @@
...
+#![create_instance]
+NeuralNetwork network = neural_network("NeuralNetwork", labels=labels_train, max_num_epochs=6, epsilon=0.0, optimization_method=enum ENNOptimizationMethod.NNOM_GRADIENT_DESCENT, gd_learning_rate=0.01, gd_mini_batch_size=3, max_norm=1.0, dropout_input=0.5)
+network.set_layers(layers_conv)
+network.quick_connect()
+network.initialize_neural_network()
+#![create_instance]
+
+#![train_and_apply]
+network.train(features_train)
+Labels labels_predict = network.apply_multiclass(features_train)
```
karlnapf: apply

examples/meta/src/neural_nets/feedforward_net_classification.sg
```diff
@@ -1,28 +1,31 @@
...
+all_layers.append_element(input)
+NeuralLayer relu = neural_layer("NeuralRectifiedLinearLayer", num_neurons=10)
+all_layers.append_element(relu)
+NeuralLayer softmax = neural_layer("NeuralSoftmaxLayer", num_neurons=2)
+all_layers.append_element(softmax)
 #![add_layers]
 
 #![create_instance]
-NeuralNetwork network(all_layers)
+NeuralNetwork network = neural_network("NeuralNetwork", labels=labels_train)
```
karlnapf: Neural network is a machine, so this could be instantiated using `machine`
This means that the below methods `set_layers`, `quick_connect`, and `initialize_neural_network` all would have to be changed.
* `set_layers` can be done via put (already now!) just needs a dispatcher for the `CDynamicObjectArray` in put.
* `quick_connect`, and `initialize_neural_network` could just be called automatically inside the `train_machine` method which should be a minor refactor

examples/meta/src/neural_nets/feedforward_net_classification.sg
```diff
@@ -37,19 +40,18 @@ network.set_gd_momentum(0.9)
 #![set_parameters]
 
 #![train_and_apply]
-network.set_labels(labels_train)
 network.train(features_train)
-BinaryLabels labels_predict = network.apply_binary(features_test)
+Labels labels_predict = network.apply_binary(features_test)
```
karlnapf: apply

src/shogun/util/factory.h
```diff
@@ -34,6 +36,8 @@ namespace shogun
 	CECOCEncoder* ecoc_encoder(const std::string& name);
 	CECOCDecoder* ecoc_decoder(const std::string& name);
 	CTransformer* transformer(const std::string& name);
+	CNeuralLayer* neural_layer(const std::string& name);
+	CNeuralNetwork* neural_network(const std::string& name);
```
karlnapf: Would rather not have a network factory, but instead use `machine`

karlnapf:
Please also update to only using 2 classes (and then try less data and also pls post the runtime of the meta example)

src/shogun/neuralnets/NeuralNetwork.cpp
```diff
@@ -229,6 +238,15 @@ CDenseFeatures< float64_t >* CNeuralNetwork::transform(
 
 bool CNeuralNetwork::train_machine(CFeatures* data)
 {
+	if (!m_is_connected)
+	{
+		quick_connect();
+	}
+	if (!m_is_initialized)
```
shubham808: this is cleaner since it will avoid multiple calls to initialize_neural_network

examples/meta/src/neural_nets/convolutional_net_classification.sg
```diff
@@ -0,0 +1,43 @@
...
+NeuralLayer conv1 = neural_layer("NeuralConvolutionalLayer", num_maps=10, radius_x=2, radius_y=2, pooling_width=2, pooling_height=2, stride_x=1, stride_y=1)
+layers_conv.append_element(conv1)
+NeuralLayer conv2 = neural_layer("NeuralConvolutionalLayer", num_maps=15, radius_x=2, radius_y=2, pooling_width=2, pooling_height=2, stride_x=1, stride_y=1)
+layers_conv.append_element(conv2)
+NeuralLayer softmax = neural_layer("NeuralSoftmaxLayer", num_neurons=10)
+layers_conv.append_element(softmax)
+#![add_layers]
+
+#![create_instance]
+Machine network = machine("NeuralNetwork", labels=labels_train, max_num_epochs=4, epsilon=0.0, optimization_method=enum ENNOptimizationMethod.NNOM_GRADIENT_DESCENT, gd_learning_rate=0.01, gd_mini_batch_size=3, max_norm=1.0, dropout_input=0.5, layers=layers_conv)
```
karlnapf: nice!

examples/meta/src/neural_nets/feedforward_net_classification.sg
```diff
@@ -1,55 +1,45 @@
...
 
 #![add_layers]
 int num_feats = features_train.get_int("num_features")
-NeuralLayers layers() 
-layers.input(num_feats)
-layers.rectified_linear(10)
-layers.softmax(2)
-DynamicObjectArray all_layers = layers.done()
+DynamicObjectArray all_layers()
+NeuralLayer input = neural_layer("NeuralInputLayer", num_neurons=num_feats)
```
karlnapf: I wonder, could we go with the simpler name `layer`?

src/shogun/base/base_types.h
```diff
@@ -31,7 +31,8 @@ namespace shogun
 	                    std::is_same<CLabels, T>::value ||
 	                    std::is_same<CECOCEncoder, T>::value ||
 	                    std::is_same<CECOCDecoder, T>::value ||
-	                    std::is_same<CMulticlassStrategy, T>::value>
+	                    std::is_same<CMulticlassStrategy, T>::value ||
+	                    std::is_same<CDynamicObjectArray, T>::value>
```
karlnapf: exactly like this!

src/shogun/neuralnets/NeuralNetwork.cpp
karlnapf: yep!

karlnapf: although .... 
why would there be multiple calls? Due to multiple "train" calls?

karlnapf: also would need to make sure that these things are set to false if the network architecture is changed.
So actually, it might be that *always* calling these might be a better idea. The cost is neglectible.
Thoughts?

examples/meta/src/neural_nets/feedforward_net_regression.sg
```diff
@@ -1,54 +1,45 @@
...
 
 #![add_layers]
-int dimensions = features_train.get_int("num_features")
-NeuralLayers layers() 
-layers.input(dimensions)
-layers.rectified_linear(20)
-layers.linear(1)
-DynamicObjectArray all_layers = layers.done()
+int num_feats = features_train.get_int("num_features")
+DynamicObjectArray all_layers()
```
karlnapf: btw instead of using the array type here explicitly, we should be able to use `SGObject::add`, which allows to append to an internal array.
The call should be `machine.add("layers", input);` etc. This means you would first create the network instance, and only then add the layers.
If this works, we dont need to add the array type to shogun's base classes, which I would prefer

karlnapf: see the multiple kernel learning example where the same thing is done

src/shogun/neuralnets/NeuralNetwork.cpp
```diff
@@ -813,9 +829,13 @@ void CNeuralNetwork::init()
 		"Parameters", MS_NOT_AVAILABLE);
 	SG_ADD(&m_param_regularizable, "param_regularizable",
 		"Parameter Regularizable", MS_NOT_AVAILABLE);
-	SG_ADD((CSGObject**)&m_layers, "layers",
-		"DynamicObjectArray of NeuralNetwork objects",
-		MS_NOT_AVAILABLE);
+	SG_ADD(
+	    &m_layers, "layers", "DynamicObjectArray of NeuralNetwork objects",
+	    MS_NOT_AVAILABLE);
+	SG_ADD(&m_auto_quick_initialize, "auto_quick_initialize", "auto_quick_initialize", MS_NOT_AVAILABLE);
```
karlnapf: This needs a better description.
"Whether the network is automatically connected and initialized on training."

src/shogun/neuralnets/NeuralNetwork.h
```diff
@@ -622,6 +626,16 @@ friend class CDeepBeliefNetwork;
 	 */
 	bool m_is_training;
 
+	/** True if the network layers are to be quick connected and initialized
+	 * initial value is False
+	 */
+	bool m_auto_quick_initialize;
```
karlnapf: pls refer to the doxygen methods that are called here

examples/meta/src/neural_nets/feedforward_net_classification.sg
```diff
@@ -1,55 +1,44 @@
...
-network.set_l2_coefficient(0.01)
-network.set_dropout_hidden(0.5)
-network.set_max_num_epochs(50)
-network.set_gd_mini_batch_size(num_feats)
-network.set_gd_learning_rate(0.1)
-network.set_gd_momentum(0.9)
-#![set_parameters]
+#![add_layers]
+NeuralLayer input = layer("NeuralInputLayer", num_neurons=num_feats)
+network.add("layers", input)
```
karlnapf: yes this is much nicer. Should have thought about it straight away :)

examples/meta/src/neural_nets/feedforward_net_classification.sg
```diff
@@ -1,55 +1,44 @@
...
 
-#![set_parameters]
-network.set_l2_coefficient(0.01)
-network.set_dropout_hidden(0.5)
-network.set_max_num_epochs(50)
-network.set_gd_mini_batch_size(num_feats)
-network.set_gd_learning_rate(0.1)
-network.set_gd_momentum(0.9)
-#![set_parameters]
+#![add_layers]
```
karlnapf: if you change these references you will need to change the cookbook ones as well! It refers to these snippets!

examples/meta/src/neural_nets/feedforward_net_regression.sg
```diff
@@ -1,54 +1,44 @@
...
+Labels labels_test = labels(f_labels_test)
 #![create_features]
 
-#![add_layers]
-int dimensions = features_train.get_int("num_features")
-NeuralLayers layers() 
-layers.input(dimensions)
-layers.rectified_linear(20)
-layers.linear(1)
-DynamicObjectArray all_layers = layers.done()
```
karlnapf: Is this taken care of?

karlnapf: ah yes it just returns the DynamicObjectArray so if we set that instead using add, all should be good

src/shogun/neuralnets/NeuralNetwork.cpp
```diff
@@ -772,7 +784,11 @@ void CNeuralNetwork::init()
 	m_lbfgs_temp_inputs = NULL;
 	m_lbfgs_temp_targets = NULL;
 	m_is_training = false;
-
+	m_auto_quick_initialize = false;
+	m_sigma = 0.01f;
+	m_layers=new CDynamicObjectArray();
```
karlnapf: you need to unref this in the destructor

src/shogun/neuralnets/NeuralNetwork.cpp
```diff
@@ -772,7 +784,11 @@ void CNeuralNetwork::init()
 	m_lbfgs_temp_inputs = NULL;
 	m_lbfgs_temp_targets = NULL;
 	m_is_training = false;
-
+	m_auto_quick_initialize = false;
+	m_sigma = 0.01f;
```
karlnapf: the f is not necessary

examples/meta/src/neural_nets/feedforward_net_classification.sg
shubham808: right i will edit the cookbook too :)

examples/meta/src/neural_nets/convolutional_net_classification.sg
```diff
@@ -1,7 +1,7 @@
-File f_feats_train = csv_file("../../data/fm_train_mnist_16x16.dat")
-File f_feats_test = csv_file("../../data/fm_test_mnist_16x16.dat")
-File f_labels_train = csv_file("../../data/labels_train_mnist_16x16.dat")
-File f_labels_test = csv_file("../../data/labels_test_mnist_16x16.dat")
+File f_feats_train = csv_file("../../data/classifier_3class_256d_linear_features_train.dat")
```
karlnapf: I am confused by the "linear" in the filename. 
Also I think it should still say "mnist"
What about `mnist_3class_256d_features_train.dat`

karlnapf:
Happy to merge after the filename change

examples/meta/src/neural_nets/convolutional_net_classification.sg
shubham808: yes that is a better name :)

karlnapf:
Tada :)

Issue #4385, opened 2018-07-25
fix linalg::dger which is way to inefficient


karlnapf:
the lapack routine does this better.
At naive loop implementation would be better than allocating a new matrix here

@vinx13 I assign you, as you started using this, I know it is not your fault but maybe you can make this a bit cleaner?  https://github.com/shogun-toolbox/shogun/commit/ce8c9b009b0d04f4866b0f0e62b2c514ac6f9176#diff-b59cfc6c549dd52160d0ce73b6356b04R589

karlnapf:
At least a bit better:
```
		template <typename T>
		void
		dger(T alpha, const SGVector<T> x, const SGVector<T> y, SGMatrix<T>& A)
		{
			REQUIRE(
			    A.num_rows == A.num_cols, "Matrix A (%d x% d) is not square!\n",
			    A.num_rows, A.num_cols);
			REQUIRE(
				x.vlen == A.num_rows, "Length of vector x (%d) doesn't match "
									  "matrix A (%d)\n",
				x.vlen, A.num_rows);
			REQUIRE(
				y.vlen == A.num_rows, "Length of vector y (%d) doesn't match "
									  "matrix A (%d)\n",
				y.vlen, A.num_rows);

			for (auto j = 0; j < A.num_cols; ++j)
			{
				for (auto i = 0; i < A.num_cols, ++i)
					A(i,j) += x[i]*y[j];
			}
		}
```

karlnapf:
Actually nevermind @vinx13 I'll include that in my latest PR

PR #4387, opened 2018-07-31
add plain and cholesky rank updates

karlnapf:


karlnapf:
fixes #4385 and prepares ground for a linear regression with bias update

karlnapf:
Eigen3 can also do this, but it only updates the lower or upper part of a matrix. We might want to think about introducing types link symmetric matrices where only half of the memory is effectively used and the rest can be used a temp storage by algorithms, like eigen does

karlnapf:
@OXPHOS btw is there a reason why we don't define the methods in linalg using a type trait 

```
template <typename T, typename T2 = typename std::enable_if_t<std::is_arithmetic<T>::value>>
template <typename T, typename T2 = typename std::enable_if_t<std::is_floating_point<T>::value>>
template <typename T, typename T2 = typename std::enable_if_t<std::is_integral<T>::value>>
```
etc

Could this replace the macros to define things for certain ptypes?

OXPHOS:
I can not recall any reason for not using `enable_if`. We actually used it
for one scenario. I can look into it soon. Just finished my defense ;D

karlnapf:
wow congrats! :)

src/shogun/mathematics/linalg/backend/eigen/BasicOps.cpp
```diff
@@ -266,6 +291,25 @@ void LinalgBackendEigen::pinv_impl(
...
 
+template <typename T>
+void LinalgBackendEigen::rank_update_impl(
+    SGMatrix<T>& A, const SGVector<T>& b, T alpha) const
+{
+	T x;
+	T update;
+	for (auto j : range(A.num_cols))
+	{
+		x = b[j];
```
iglesias: Out of curiosity, having the local x is better than just using `b[j]` all over the method? Same question for `update`.

src/shogun/mathematics/linalg/LinalgNamespace.h
```diff
@@ -680,9 +673,63 @@ namespace shogun
...
+			REQUIRE(
+			    L.num_rows == b.size(),
+			    "Vector size (%d) must match matrix size (%dx%d)\n", b.size(),
+			    L.num_rows, L.num_rows);
+			return infer_backend(L, SGMatrix<T>(b))
+			    ->cholesky_rank_update(L, b, alpha, lower);
+		}
+
+		/**
+		 * Updates a matrix \f$A\f$ with a rank one term in-place,If A = LL^T
```
iglesias: Minor: I think for this method the the A = LL^T can be left out.

src/shogun/mathematics/linalg/LinalgNamespace.h
```diff
@@ -680,9 +673,63 @@ namespace shogun
...
+			    "Matrix dimensions (%dx%d) are not square\n", A.num_rows,
+			    A.num_cols);
 			return infer_backend(A)->cholesky_factor(A, lower);
 		}
 
+		/**
+		 * Updates the Cholesky factorization \f$A = L L^{*}\f$ with a rank one
+		 * term in-place.
+		 * If A = LL^T before the update, then after it
+		 * LL^{*} = A + alpha * b b^{*}
```
iglesias: Quite nitpicking: perhaps a different name for the L as they are not the same as the ones above. In wikipedia they use tilde analogously for A. It is completely subjective I think, whether we want the doc to be more math correct, or just reflect what happens in the code. I am happy either way :-)

PR #4392, opened 2018-08-18
Full support for std::vectors in Any [WIP]

lisitsyn:


lisitsyn:
Ok I introduced a test that shows that vectors generally work. What does not work is combinatorial stuff like vectors of SGVectors. Will work on that next

tests/unit/lib/Any_unittest.cc
```diff
@@ -691,3 +691,13 @@ TEST(Any, lazy_cloneable_visitable)
 	EXPECT_FALSE(any.visitable());
 	EXPECT_THROW(any.visit(nullptr), std::logic_error);
 }
+
+TEST(Any, compare_string_vectors)
+{
+	std::string str("some string");
+	std::vector<std::string> lhs{str};
+	std::vector<std::string> rhs{str};
```
karlnapf: sure, we can register and we can do equals, and what happens is that the == operator of std::vector is called. This one uses the == operator on all elements, https://en.cppreference.com/w/cpp/container/vector/operator_cmp

It will not work recursively, i.e. std::vector<SGVector<X>> will also use the == operator of the elements, which in shogun is pointer equality and not .equals

karlnapf:
Discussed this quite a bit with @vigsterkr and we think that we do need vector, list, map, set (at least). But set list and vector would be a great start if map is hard

tests/unit/lib/Any_unittest.cc
```diff
@@ -691,3 +691,25 @@ TEST(Any, lazy_cloneable_visitable)
...
+TEST(Any, compare_object_vectors)
+{
+	auto lhs_obj = std::make_shared<Object>();
+	auto rhs_obj = std::make_shared<Object>();
+	EXPECT_TRUE(lhs_obj->equals(rhs_obj.get()));
+	std::vector<CSGObject*> lhs{lhs_obj.get()};
+	std::vector<CSGObject*> rhs{rhs_obj.get()};
+	Any any_lhs = make_any(lhs);
+	Any any_rhs = make_any(rhs);
+	EXPECT_EQ(any_lhs, any_rhs);
```
karlnapf: wait, did we want == operator for Any? Maybe I forgot this .... ;)
I think we can merge and then I add a vector parameter to our big shogun class test

karlnapf:
https://travis-ci.org/shogun-toolbox/shogun/jobs/426069785#L799

karlnapf:
Soooo. what is the latest here? :)

vigsterkr:
@lisitsyn @karlnapf can i cherry-pick rebase and merge this into develop?

karlnapf:
Yes go for it. Can be extended later on but then we at least have vector

PR #4400, opened 2018-09-25
Fix issue when building without GPL.

geektoni:
If -DLICENSE_GPL_SHOGUN=OFF -DUSE_SVMLIGHT=OFF are provided then the cmake procedure fails because it does not find some SVM meta examples (which are non-existing).

### DIFF
--- a/cmake/FindMetaExamples.cmake
+++ b/cmake/FindMetaExamples.cmake
@@ -35,8 +35,6 @@ function(get_excluded_meta_examples)
       statistical_testing/quadratic_time_maximum_mean_discrepancy.sg
       gaussian_process/classifier.sg
       gaussian_process/regression.sg
-      binary/svmlin.sg
-      binary/svmsgd.sg
       )
 	ENDIF()
 


cmake/FindMetaExamples.cmake
```diff
@@ -35,8 +35,6 @@ function(get_excluded_meta_examples)
       statistical_testing/quadratic_time_maximum_mean_discrepancy.sg
       gaussian_process/classifier.sg
       gaussian_process/regression.sg
-      binary/svmlin.sg
```
karlnapf: this is not the right fix here imo

This list is excluded meta examples (in this case for GPL)

Instead, you need to add the same names to the SVMLight exclusion section.

geektoni: I see what you mean. However, the problem here is caused by the fact that those two files don't exist anymore, so even if I moved them into the SVMLight section I would face the same issue (cmake refuses to run because it doesn't find those files).

karlnapf: I see. I guess they were renamed?

karlnapf: If so, we should just change the fname. IF they were deleted, then this can go..... but if they were just renamed the build should fail as the examples with the new name are not excluded from the tests and then produce a failure due to missing svmlight

geektoni: I did a bit of research around the repo and the only file that matches is an undocumented python exampled (examples/undocumented/python/graphical/svmlin.py). I also tried to search for those files inside git history, but I found nothing. If we look at the buildbot (http://buildbot.shogun-toolbox.org:8080/#/builders/4/builds/175) we see that three months ago everything was working. I tried to search for relevant changes in the last three months, but again I found nothing. I guess we can remove those files since they seem to have been not renamed.

geektoni: The buildbot will spot errors (hopefully) if we missed something anyway.

geektoni:
Sooo, shall we merge this? ;)

karlnapf:
I think we should find out where those lines initially came from and then either confirm that those files were deleted (which commit, why), or find out the new name and then replace the filenames. We had renamed a few examples a while ago and I have the feeling that those for part of that.

geektoni:
Okay, so, I did again a bit of research and I found the real problem. Those two lines were added here https://github.com/shogun-toolbox/shogun/commit/9e5199bb45bef4196f20a14c00426fe400053677 when we merged this PR #4361. The purpose of the PR was to port two undocumented examples (`classifier_svmlin.py` and `classifier_svmsgd.py`) to the meta example system. However, the PR lacks the "translated" `.sg` files for the meta examples :/ therefore, those two files never existed.

The solution here would be to add again those two files (since they are actually useful).

geektoni:
Hey @vinx13! Do you remember those two files (`svmlin.sg` and `svmsgd.sg`)? If yes, do you happen to still have them somewhere? ;)

vinx13:
@geektoni Seems that I didn't include that two files in the pr, I need some time to investigate it

karlnapf:
Thanks for sorting this out guys!!

PR #4404, opened 2018-10-11
Add svmlin & svmsgd meta example

vinx13:
Two files are missing from previous pr #4361, data shogun-toolbox/shogun-data#168
See also #4400


karlnapf:
What’s the errors?

karlnapf:
uff this seems like a python version problem. Can you reproduce this error locally? I bet it is some upgrade to Python 3

vinx13:
some error on windows about variant since 7c7693f81974191b377245ea4f, i guess that's some compiler version issue

karlnapf:
the windows variant error for sure is compiler version
But the Python error, what about that? We should get at least travis green

vinx13:
@karlnapf travis gets green now

karlnapf:
Great, thanks for the ping and let's merge

karlnapf:
@vinx13 would you min checking the variant compiler error on windows? At least googling the source a bit?

vinx13:
the variant error on windows seems a MSVC issue, see https://github.com/mpark/variant/pull/49
I will update version of variant

Issue #4383, opened 2018-07-23
clean up linalg unittest


karlnapf:
E.g. the `Eigen3_operations_unittest.cc` only tests for float64_t.

Make the tests typed so they are executed for every shogun type! See e.g. `SGObjectAll_unittest.cc` for how to do that.

Next, use fixtures to define the necessary matrix objects to avoid the huge amount of redundant code in there

gf712:
Hi! 
Do you still need help with this issue? I just started having a look at how the typed tests work in LDA_unittest.cc. It's pretty neat!

For the linalg tests would you test it just with floats, so it would be a refactor of `Eigen3_operations_unittest.cc`, i.e. the tests are written with floats, so just need to extend to `float32_t` and `floatmax_t` using `TYPED_TEST`. Or, also perform tests with `uint8_t`, `int16_t`, `int32_t` and `int64_t` where possible, which I assume requires a change of the values being tested to work as expected with both int and float types?

karlnapf:
Welcome! :)
Yes totally, this is a good way to get involved with our testing system!
We would want to test the operations for all types that they are supported with. Sometimes that is only floats (some solvers for example), sometimes it is more (add) You can see the definitions [here](https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/mathematics/linalg/LinalgBackendBase.h), `DEFINE_FOR_ALL_PTYPE` and the like. It is quite a mess with Macros and all that in there, but basically every operation is explicitly defined for a number of types. I think the first step would be to define the test types in the unit test locally and refactor, then define them globally somehow so we can re-use things.

karlnapf:
The best thing would be to globally define the types for googletest to match things like all shogun ptypes, all floating point ptypes, all numerical ptypes, etc

gf712:
OK! Thank you. For now I am defining the test case types locally. I agree that it is a bit messy to rewrite the types each time.
I will send a PR when that is done, and then can have a look at how to clean up the code a bit!

karlnapf:
No idea. In fact, we have been moving towards getting rid of `CMath::` functions recently, in favour of using `std::` directly. So there should be no need for `powl` anymore.
@vigsterkr do you know why `pow` is still there, we have removed most others
@vinx13 maybe you know?

vinx13:
i have no idea. but here i think it is ok to rename `powl` to `pow`

karlnapf:
Thanks!

gf712:
[...]

PR #4405, opened 2018-10-15
Refactored linalg operations tests ( addresses #4383)

gf712:
Refactored all the linalg Eigen3 operations in `Eigen3_operations_unittest.cc` as suggested in #4383.

In addtion to typed testing, changes include:
* Decreased tolerance for `float` type tests to work with `float32_t`, `float64_t` and `floatmax_t`
* Different values in tests where the same code has to work with both `int` and `float` types

I did not add `complex128_t`, `bool` or `char` tests at this point.

I just need help writing a test for `SGMatrix_squared_error` as this potentially requires changing CMath::pow (mentioned in #4383). At this point I left the original test.

I know it's quite a large number of lines changed in one PR, but this is mainly a code refactor and the tests are almost identical.

All tests run fine locally!

### DIFF
--- a/tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
+++ b/tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
@@ -11,6 +11,17 @@ using namespace shogun;
 using namespace linalg;
 using namespace Eigen;
 
+// Tolerance values for tests
+template<typename T>
+// default tolerance
+T get_epsilon() {return 0;}
+template<>
+float32_t get_epsilon() {return 1e-5;}
+template<>
+float64_t get_epsilon() {return 1e-9;}
+template<>
+floatmax_t get_epsilon() {return 1e-12;}
+
 template <typename T>
 class LinalgBackendEigenAllTypesTest: public ::testing::Test { };
 template <typename T>
--- a/tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
+++ b/tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
@@ -1633,29 +1633,29 @@ void check_SGMatrix_softmax()
         EXPECT_NEAR(ref[i], A[i], 1e-7);
 }
 
-template <typename ST>
-void check_SGMatrix_squared_error()
-{
-    SGMatrix<ST> A(4, 3);
-    SGMatrix<ST> B(4, 3);
-
-    ST size = A.num_rows * A.num_cols;
-    for (ST i = 0; i < size; i+=2)
-    {
-        A[i] = i / size;
-        B[i] = (i / size) * 2;
-    }
-
-    ST ref = 0;
-    for (index_t i = 0; i < size; i++)
-        ref += CMath::pow(A[i] - B[i], 2);
-    ref *= 0.5;
-
-    printf("%d", ref);
-
-    auto result = linalg::squared_error(A, B);
-        EXPECT_NEAR(ref, result, 1e-15);
-}
+//template <typename ST>
+//void check_SGMatrix_squared_error()
+//{
+//    SGMatrix<ST> A(4, 3);
+//    SGMatrix<ST> B(4, 3);
+//
+//    ST size = A.num_rows * A.num_cols;
+//    for (ST i = 0; i < size; i+=2)
+//    {
+//        A[i] = i / size;
+//        B[i] = (i / size) * 2;
+//    }
+//
+//    ST ref = 0;
+//    for (index_t i = 0; i < size; i++)
+//        ref += CMath::pow(A[i] - B[i], 2);
+//    ref *= 0.5;
+//
+//    printf("%d", ref);
+//
+//    auto result = linalg::squared_error(A, B);
+//        EXPECT_NEAR(ref, result, 1e-15);
+//}
 
 
 template <typename ST>
@@ -2747,4 +2747,26 @@ TYPED_TEST(LinalgBackendEigenAllTypesTest, SGMatrix_zero)
 TYPED_TEST(LinalgBackendEigenNonIntegerTypesTest, SGMatrix_rank_update)
 {
     check_SGMatrix_rank_update<TypeParam>();
+}
+
+// TODO: implement typed testing for SE
+TEST(LinalgBackendEigen, SGMatrix_squared_error)
+{
+    SGMatrix<float64_t> A(4, 3);
+    SGMatrix<float64_t> B(4, 3);
+
+    int32_t size = A.num_rows * A.num_cols;
+    for (float64_t i = 0; i < size; ++i)
+    {
+        A[i] = i / size;
+        B[i] = (i / size) * 0.5;
+    }
+
+    float64_t ref = 0;
+    for (index_t i = 0; i < size; i++)
+        ref += CMath::pow(A[i] - B[i], 2);
+            ref *= 0.5;
+
+    auto result = linalg::squared_error(A, B);
+    EXPECT_NEAR(ref, result, 1e-15);
 }
--- a/tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
+++ b/tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
@@ -11,148 +11,178 @@ using namespace shogun;
 using namespace linalg;
 using namespace Eigen;
 
-TEST(LinalgBackendEigen, SGVector_add)
-{
-	const float64_t alpha = 0.3;
-	const float64_t beta = -1.5;
+template <typename T>
+class LinalgBackendEigenAllTypesTest: public ::testing::Test { };
+template <typename T>
+class LinalgBackendEigenNonComplexTypesTest: public ::testing::Test { };
+template <typename T>
+class LinalgBackendEigenRealTypesTest: public ::testing::Test { };
+template <typename T>
+class LinalgBackendEigenNonIntegerTypesTest: public ::testing::Test { };
 
-	SGVector<float64_t> A(9);
-	SGVector<float64_t> B(9);
+// TODO: make global definitions
+// Definition of the 4 groups of Shogun types (shogun/mathematics/linalg/LinalgBackendBase.h)
+// TODO: add bool, chars and complex128_t types
+typedef ::testing::Types<int8_t, int16_t, uint16_t, int32_t, uint32_t, int64_t, uint64_t, float32_t, float64_t, floatmax_t> AllTypes;
+typedef ::testing::Types<int8_t, int16_t, uint16_t, int32_t, uint32_t, int64_t, uint64_t, float32_t, float64_t, floatmax_t> NonComplexTypes;
+typedef ::testing::Types<float32_t, float64_t, floatmax_t> RealTypes;
+// TODO: add complex128_t type
+typedef ::testing::Types<float32_t, float64_t, floatmax_t> NonIntegerTypes;
 
-	for (index_t i = 0; i < 9; ++i)
-	{
-		A[i] = i;
-		B[i] = 0.5*i;
-	}
+TYPED_TEST_CASE(LinalgBackendEigenAllTypesTest, AllTypes);
+TYPED_TEST_CASE(LinalgBackendEigenNonComplexTypesTest, NonComplexTypes);
+TYPED_TEST_CASE(LinalgBackendEigenRealTypesTest, RealTypes);
+TYPED_TEST_CASE(LinalgBackendEigenNonIntegerTypesTest, NonIntegerTypes);
 
-	auto result = add(A, B, alpha, beta);
 
-	for (index_t i = 0; i < 9; ++i)
-		EXPECT_NEAR(alpha*A[i]+beta*B[i], result[i], 1e-15);
+template <typename ST>
+void check_SGVector_add() {
+
+    const ST alpha = 1;
+    const ST beta = 2;
+
+    SGVector<ST> A(9);
+    SGVector<ST> B(9);
+
+    for (index_t i = 0; i < 9; ++i)
+    {
+        A[i] = i;
+        B[i] = 2*i;
+    }
+
+    auto result = add(A, B, alpha, beta);
+
+    for (index_t i = 0; i < 9; ++i)
+        EXPECT_NEAR(alpha*A[i]+beta*B[i], result[i], 1e-15);
 }
 
-TEST(LinalgBackendEigen, SGMatrix_add)
-{
-	const float64_t alpha = 0.3;
-	const float64_t beta = -1.5;
-	const index_t nrows = 2, ncols = 3;
+template <typename ST>
+void check_SGMatrix_add() {
 
-	SGMatrix<float64_t> A(nrows, ncols);
-	SGMatrix<float64_t> B(nrows, ncols);
+    const ST alpha = 1;
+    const ST beta = 2;
+    const index_t nrows = 2, ncols = 3;
 
-	for (index_t i = 0; i < nrows*ncols; ++i)
-	{
-		A[i] = i;
-		B[i] = 0.5*i;
-	}
+    SGMatrix<ST> A(nrows, ncols);
+    SGMatrix<ST> B(nrows, ncols);
 
-	auto result = add(A, B, alpha, beta);
+    for (index_t i = 0; i < nrows*ncols; ++i)
+    {
+        A[i] = i;
+        B[i] = 2*i;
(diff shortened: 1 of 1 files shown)


tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
```diff
@@ -11,148 +11,178 @@ using namespace shogun;
...
+class LinalgBackendEigenRealTypesTest: public ::testing::Test { };
+template <typename T>
+class LinalgBackendEigenNonIntegerTypesTest: public ::testing::Test { };
 
-	SGVector<float64_t> A(9);
-	SGVector<float64_t> B(9);
+// TODO: make global definitions
+// Definition of the 4 groups of Shogun types (shogun/mathematics/linalg/LinalgBackendBase.h)
+// TODO: add bool, chars and complex128_t types
+typedef ::testing::Types<int8_t, int16_t, uint16_t, int32_t, uint32_t, int64_t, uint64_t, float32_t, float64_t, floatmax_t> AllTypes;
```
karlnapf: yep!

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
```diff
@@ -11,148 +11,178 @@ using namespace shogun;
...
+typedef ::testing::Types<float32_t, float64_t, floatmax_t> NonIntegerTypes;
 
-	for (index_t i = 0; i < 9; ++i)
-	{
-		A[i] = i;
-		B[i] = 0.5*i;
-	}
+TYPED_TEST_CASE(LinalgBackendEigenAllTypesTest, AllTypes);
+TYPED_TEST_CASE(LinalgBackendEigenNonComplexTypesTest, NonComplexTypes);
+TYPED_TEST_CASE(LinalgBackendEigenRealTypesTest, RealTypes);
```
karlnapf: @vigsterkr what's your opinion in making those things globally visible so that we wont have to re-declare them?

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
```diff
@@ -11,148 +11,178 @@ using namespace shogun;
...
+TYPED_TEST_CASE(LinalgBackendEigenNonIntegerTypesTest, NonIntegerTypes);
+
+
+template <typename ST>
+void check_SGVector_add() {
```
karlnapf: What is the motivation to have the "check_" prefix here?

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
```diff
@@ -11,148 +11,178 @@ using namespace shogun;
...
 
-	for (index_t i = 0; i < 9; ++i)
-	{
-		A[i] = i;
-		B[i] = 0.5*i;
-		C[i] = i;
-	}
+    for (index_t i = 0; i < 9; ++i)
+    {
+    A[i] = i;
```
karlnapf: there are some minor formatting issues here.
You an actually automatically format this properly... see the output from the first travis job here:
https://travis-ci.org/shogun-toolbox/shogun/jobs/441713579#L723

karlnapf: This would also make the diff of this PR much smaller, as there are fewer whitespace changes

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
```diff
@@ -11,148 +11,178 @@ using namespace shogun;
...
-	const float64_t alpha = 0.3;
-	const float64_t beta = -1.5;
-	const index_t nrows = 2, ncols = 3;
+template <typename ST>
+void check_SGMatrix_add_in_place() {
 
-	SGMatrix<float64_t> A(nrows, ncols);
-	SGMatrix<float64_t> B(nrows, ncols);
-	SGMatrix<float64_t> C(nrows, ncols);
+    const ST alpha = 1;
```
karlnapf: uh I didn't reallise that some tests actually operated with floating point number constants and we now have to change all those numbers. Quite some work! Thanks for going into this!

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
```diff
@@ -275,1817 +314,2459 @@ TEST(LinalgBackendEigen, add_vector)
...
-	float64_t data[] = {0.8192343,  0.13191962, 0.50888604,
-	                    0.16857468, 0.24107738, 0.89455301,
-	                    0.40657379, 0.07902286, 0.24319651};
-	float64_t result[] = {0.25587541,  -0.1173183, -0.13855711,
-	                      -0.34283925, 0.04378442, 0.29905482,
-	                      0.08696383,  0.07353387, -0.16049771};
+	ST data[] = {0.5, 0.3, 0.4,
+	             0.4, 0.5, 0.3,
+	             0.3, 0.4, 0.5};
+	ST result[] = {0.1, -0.1, 0.0,
```
karlnapf: did you calculate these numbers?

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
```diff
@@ -275,1817 +314,2459 @@ TEST(LinalgBackendEigen, add_vector)
...
 }
 
-TEST(LinalgBackendEigen, SGMatrix_cholesky_llt_lower)
+template <typename ST>
+void check_SGMatrix_cholesky_llt_lower()
 {
 	const index_t size=2;
-	SGMatrix<float64_t> m(size, size);
+	SGMatrix<ST> m(size, size);
+	typedef Matrix<ST, Dynamic, Dynamic> Mxx; // need to adapt the Eigen::Matrix to ST type
```
karlnapf: I think we might have something similar in `SGVector.h`

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
```diff
@@ -275,1817 +314,2459 @@ TEST(LinalgBackendEigen, add_vector)
...
+	SGMatrix<ST> L = cholesky_factor(m);
+
 
-	Map<MatrixXd> map_A(m.matrix,m.num_rows,m.num_cols);
-	Map<MatrixXd> map_L(L.matrix,L.num_rows,L.num_cols);
+	Map<Mxx> map_A(m.matrix,m.num_rows,m.num_cols);
+	Map<Mxx> map_L(L.matrix,L.num_rows,L.num_cols);
 	EXPECT_NEAR((map_A-map_L*map_L.transpose()).norm(),
-		0.0, 1E-15);
+		0.0, 1E-6); // need to change the treshold to work with floats
```
karlnapf: as an idea, we could use a templated function `get_epsilon<T>` that looks up the appropriate epsilons?

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
```diff
@@ -275,1817 +314,2459 @@ TEST(LinalgBackendEigen, add_vector)
...
+
+    auto A2 = A.clone();
+    Map<Mxx> A2_eig(A2.matrix, A2.num_rows, A2.num_cols);
+    A2(0, 0) += b[0] * b[0];
+    A2(0, 1) += b[0] * b[1];
+    A2(1, 0) += b[1] * b[0];
+    A2(1, 1) += b[1] * b[1];
+
+    cholesky_rank_update(U, b, alpha, false);
+    EXPECT_NEAR((A2_eig - U_eig.transpose() * U_eig).norm(), 0.0, 1e-5);  // need to change the treshold to work with floats
```
karlnapf: minor style: we do comments in a line above usually (+ a whitespace line above that)

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
```diff
@@ -275,1817 +314,2459 @@ TEST(LinalgBackendEigen, add_vector)
...
-	                      0.30492837, 0.39708534, 0.40466969,
-	                      0.36415317, 0.04407589, 0.9095746};
-	float64_t data_B[] = {0.76775073, 0.88471312, 0.34795225,
-	                      0.94311546, 0.59630347, 0.65820143};
-	float64_t result[] = {-0.73834587, 4.22750496, -1.37484721,
-	                      -1.14718091, 4.49142548, -1.08282992};
+    const index_t nrows = 2, ncols = 3;
+    SGMatrix<ST> mat(nrows, ncols);
 
-	SGMatrix<float64_t> A(data_A, n, n, fa
```
karlnapf: What is the motivation for putting these one liners in here rather than the test code directly?
How does it look when a test fails? Is it easy to exactly identify where the program is coming from? Sometimes with googltest, it is hard to figure out which case the tester is exactly in

karlnapf: Or in other words: Why not put the test code in there straight away and use `TypeParam` instead of `ST`?

karlnapf: not sure the method is even defined for int types?
Generally, we try to avoid TODOs in the codebase (they never get addressed ;)

gf712:
- Did you run all the tests locally?
Yes, all tests run fine locally, but it would be better to first implementing a default tolerance level for each type to avoid any issues!
- No memory errors?
I had no memory issues when I run with valgrind. But when running with valgrind one of the tests fails (it doesn't fail when just running unittest). In the case of`LinalgBackendEigenNonIntegerTypesTest/2.SGMatrix_pinv_psd, where TypeParam = long double` pinv returns `nan` for all matrix elements.

karlnapf:
Hi Gil,
I am quite busy today, but I will try to get back to this PR tomorrow latest. 

The error when running with valgrind seems strange, this usually happens when there is an uninitialized read somewhere in the code (something which valgrind shows). For the pinv, it might be worth checking how it is implemented and then dig in the code and eigen3 docs to see whether the way it is done works with long (eigen might bail actually). If it is unfixable, we will have to remove the definition for long

karlnapf:
Hi Gil,
I am quite busy today, but I will try to get back to this PR tomorrow latest. 

The error when running with valgrind seems strange, this usually happens when there is an uninitialized read somewhere in the code (something which valgrind shows). For the pinv, it might be worth checking how it is implemented and then dig in the code and eigen3 docs to see whether the way it is done works with long (eigen might bail actually). If it is unfixable, we will have to remove the definition for long

gf712:
Hi, I will have a better look at it today. Github wasn't working properly so I didn't really want to push any code. 
It's just odd that the pinv test runs fine with `long double` without using valgrind..!

gf712:
It seems like after changing some values of the matrix tested by `SGMatrix_pinv_psd` gives the correct answer. However, I'm not sure why this is the case. 
Printing out the individual components of the SVD decomposition shows that for some reason Eigen's `JacobiSVD` returns `nan` values, but only when using valgrind with `long double`, when there is a matrix with this format:
```
[[a b c]
 [b d b]
 [e b f]]
```
By changing one of any `b` values seems to give a valid answer..

karlnapf:
It might be worth filing this as a bug report with the eigen3 guys, they are quick in fixing bugs

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
```diff
@@ -11,189 +11,269 @@ using namespace shogun;
...
+// Tolerance values for tests
+template <typename T>
+// default tolerance
+T get_epsilon()
 {
-	const float64_t alpha = 0.3;
-	const float64_t beta = -1.5;
+	return 0;
+}
+template <>
```
karlnapf: Any chance you can make this more compact using https://en.cppreference.com/w/cpp/types/numeric_limits/epsilon

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
```diff
@@ -2077,15 +2263,561 @@ TEST(LinalgBackendEigen, SGMatrix_rank_update)
...
+	EXPECT_NEAR((A2_eig - A_eig).norm(), 0, get_epsilon<ST>());
+
+	rank_update(A, b, static_cast<ST>(-1));
+	EXPECT_NEAR((A_eig - A_eig).norm(), 0, get_epsilon<ST>());
+}
+
+// test types based on shogun/mathematics/linalg/LinalgBackendBase.h
+TYPED_TEST(LinalgBackendEigenAllTypesTest, SGVector_add)
+{
+	check_SGVector_add<TypeParam>();
```
karlnapf: again the question: if I grep for `check_SGVector_add<`, I only find this occurrence of the call, why not move the code in here? Did I miss a response for this anywhere?

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
karlnapf: does a call to this method with int type even compile?

gf712: Hi, it seems like my previous answer disappeared. There was no specific motivation for this. I just used other code in shogun as a template. The way it was written here is the same as in `shogun/tests/unit/classifier/LDA_unittest.cc`. The error traceback works fine when the test fails.

gf712: Yes, this compiles, but I don't think it might not make much sense...

https://github.com/shogun-toolbox/shogun/blob/f6efd89467a7016be8a50bf3f7c0fdf32b075369/src/shogun/mathematics/linalg/LinalgBackendBase.h#L255-L262

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
karlnapf: I think taking the machine epsilon and then dropping two orders of magnitude is a reasonable idea

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
karlnapf: I was just asking what was the motivation here: why put the code into a call rather than moving it into the test directly? If there is a good reason it is fine of course, but if there is none, then maybe we should move it :)

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
karlnapf: well as long as the eigen3 call goes through it makes sense to offer it :)
But I am surprised ...

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
karlnapf: ok then, let's keep it, no big deal

karlnapf:
What's missing for this to be finished?

gf712:
I think all that is missing is the addition of bool, chars and complex type tests. I am just not sure how those test would look like..? For example you could compile a test with dot product of `bool` and `char ` types but how would that test look like/be interpreted?

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
karlnapf: actually, thought about it again, and I think it would be better to move the code down (for those instances where the function is not re-used anywhere else) to keep the spaghetti factor down a bit. These things tend to be annoying years in the future when refactoring stuff.

gf712: OK, I'll do that!

karlnapf:
We should follow c++ `is_arithmetic`. Then remove the types that eigen3 cannot deal with, or those where it just doesn't make any sense

gf712:
Yes, I agree, in another patch could move the definitions to a header file with definitions that can be accessed by all tests?
The only thing missing here would be the complex type... But might have to write special test cases for these no?

karlnapf:
> Yes, I agree, in another patch could move the definitions to a header file with definitions that can be accessed by all tests?
yes

> The only thing missing here would be the complex type... But might have to write special test cases for these no?
You might have to for some. For now we can leave that and maybe investigate it later (e.g. when someone needs it.

Ok then, I will wait for a few more days for others to comment and then merge this

PR #4406, opened 2018-10-16
Improve (again) Signal handling.

geektoni:
Before this patch, premature stopping could only be achieved by enabling the interactive signal handler. The user now can prematurely stop algorithms by just pressing specific key combinations without passing through the handler.


karlnapf:
Just restarted the build. The patch looks good and IMO we can merge it :)

karlnapf:
How can we make examples to show this kind of stuff to users?

Issue #4382, opened 2018-07-23
LibLibearRegression with bias crashes


karlnapf:
Seems to be a minor dimensional issue.

Investigate and fix, add a meta example with integration data so we have coverage

gf712:
Hi!
I am trying to reproduce this error. Do you see this issue when setting `use_bias` to `true`?

If this is the case I found a couple of issues:
* the `prob` instance is never set to `use_bias`, e.g. `prob.use_bias = get_use_bias();`
* there is dimensionality problem, because `w` is assigned size=`num_features`:
https://github.com/shogun-toolbox/shogun/blob/e338426fc5777386b30eb0adee9e63bf61076e70/src/shogun/regression/svr/LibLinearRegression.cpp#L78-L81
and `prob.n=w.vlen+1` in the case of `use_bias = true`.
This can be fixed in `solve_l2r_l1l2_svr` by defining w_size as follows:
```
if (prob->use_bias)
  w_size = prob->n - 1;
else
  w_size = prob->n;
``` 
So w_size is the size of `w` without bias. And then replace:
https://github.com/shogun-toolbox/shogun/blob/e338426fc5777386b30eb0adee9e63bf61076e70/src/shogun/regression/svr/LibLinearRegression.cpp#L201-L202
with:
```
if (prob->use_bias)
  w.vector[w_size+1]+=beta[i];
```

I hope this makes sense and works (seems to at least run without issues...)!

karlnapf:
* Yes, this is how to reproduce the error!
* Yes, the dimensionality changing is the issue, i.e. `w` contains one more component (this bias)

I thought about the solution, and it seems to solve the problem.
Can I suggest we have a PR, along with a unit test that compares this exact solver and bias configuration with say the one interfaced by sklearn? On a minimal problem (we have environments set up for testing such things, with trainig/test data etc, see some other unit tests, or ask here)

karlnapf:
Nice catch! :)

gf712:
I will submit PR now!

I also had to add a line after training to set the bias. There was an issue with setting the train labels to `prob.y`, the memory was allocated but then the values were never assigned..

I wrote a test based on `shogun/tests/unit/regression/LibSVR_unittest.cc` with a simple equation (`y = 3x + 2`).

PR #4407, opened 2018-10-17
Lr bugfix

gf712:


src/shogun/regression/svr/LibLinearRegression.cpp
```diff
@@ -93,7 +93,10 @@ bool CLibLinearRegression::train_machine(CFeatures* data)
 	}
 	prob.l=num_vec;
 	prob.x=features;
-	prob.y=SG_MALLOC(float64_t, prob.l);
+    prob.y=SG_MALLOC(float64_t, prob.l);
```
vinx13: there are some style issues, we are using tab indent, see https://travis-ci.org/shogun-toolbox/shogun/jobs/442635134#L728

src/shogun/regression/svr/LibLinearRegression.cpp
```diff
@@ -199,7 +209,7 @@ void CLibLinearRegression::solve_l2r_l1l2_svr(SGVector<float64_t>& w, const libl
 		prob->x->add_to_dense_vec(beta[i], i, w.vector, w_size);
 
 		if (prob->use_bias)
-			w.vector[w_size]+=beta[i];
+            w.vector[w_size+1]+=beta[i];
```
vinx13: why `w_size+1`? if w has `w_size+1` elements, the last element is `w.vector[w_size]`

gf712: The original issue described in #4382 is due to a mismatch of `w_size` and `prob.x.num_features`, the former included bias, whereas the latter did not. This results in an assertion error when performing vectorised operations with `prob.x`, i.e. `prob->x->add_to_dense_vec(...)`.

 As a fix I set `w_size` to the size of weights, excluding the bias term. 
```
if (prob->use_bias)
    w_size = prob->n - 1;
else w_size = prob->n;
```
Therefore, the bias is in the `w_size+1` position. It seemed to be the cleanest fix! I could also have written something like this:
```
if (prob->use_bias)
{
    prob->x->add_to_dense_vec(beta[i], i, w.vector, prob->n-1);
    w.vector[prob->n]+=beta[i];
}
else prob->x->add_to_dense_vec(beta[i], i, w.vector, prob->n);
```

src/shogun/regression/svr/LibLinearRegression.cpp
gf712: Thanks! I have to switch my IDE to indent!

src/shogun/regression/svr/LibLinearRegression.cpp
vinx13: Thanks for clarifying. The two code snippets indeed are equivalent, but I am still confused.
When you set `w_size` to the size of weights (without bias), `w.vector[0] ... w.vector[w_size-1]` are weights, and `w.vector[w_size]` is bias. Maybe you have different idea?

gf712: Yes, that is right! I seem to have gotten confused, thanks for pointing that out. The bias worked before, but could lead to undefined behaviour.. I have fixed it now!

src/shogun/regression/svr/LibLinearRegression.cpp
```diff
@@ -123,6 +126,10 @@ bool CLibLinearRegression::train_machine(CFeatures* data)
 	}
 
 	set_w(w);
+	if (get_use_bias())
```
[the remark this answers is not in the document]
gf712: Yes that makes sense!

tests/unit/regression/LibLinearRegression_unittest.cc
```diff
@@ -0,0 +1,128 @@
...
+using namespace shogun;
+
+TEST(LibLinearRegression, lr_with_bias)
+{
+	// using bias
+	bool use_bias = true;
+	// set epsilon to an order of magnitude lower than test tolerance
+	double epsilon = 1E-6;
+
+	/* create some easy regression data: y = 3x + 2 */
```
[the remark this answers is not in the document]
gf712: I have added this in the latest commit. It should work with linear regression problems with one or more features!

tests/unit/environments/LinearTestEnvironment.h
```diff
@@ -46,16 +47,37 @@ class LinearTestEnvironment : public ::testing::Environment
...
+		mOneDimensionalRegressionData = std::shared_ptr<LinearRegressionDataGenerator>(
+				new LinearRegressionDataGenerator(100, coefficients, 0.0, 0.95));
 	}
 
 	std::shared_ptr<GaussianCheckerboard> getBinaryLabelData() const
 	{
 		return mBinaryLabelData;
 	}
 
+	std::shared_ptr<LinearRegressionDataGenerator> getOneDimensionalRegressionData(bool withBias) const
```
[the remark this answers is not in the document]
karlnapf: variables and function names are doing_this_kind_of_naming

tests/unit/regression/LibLinearRegression_unittest.cc
```diff
@@ -0,0 +1,100 @@
...
+#include "environments/LinearTestEnvironment.h"
+
+using namespace shogun;
+
+// get the test environment
+extern LinearTestEnvironment* linear_test_env;
+
+TEST(LibLinearRegression, lr_with_bias)
+{
+	// using bias
```
karlnapf: superfluous

tests/unit/regression/LibLinearRegression_unittest.cc
```diff
@@ -0,0 +1,100 @@
...
+	bool use_bias = true;
+	// set epsilon to an order of magnitude lower than test tolerance
```
karlnapf: can this be made explicit in the code rather than hardcoded?

tests/unit/regression/LibLinearRegression_unittest.cc
```diff
@@ -0,0 +1,100 @@
...
+	double epsilon = 1E-6;
+
+	// create some easy regression data: y = 3x + 2 (defined in LinearTestEnvironment.h)
```
karlnapf: no need for the comment

tests/unit/regression/LibLinearRegression_unittest.cc
```diff
@@ -0,0 +1,100 @@
...
+	std::shared_ptr<LinearRegressionDataGenerator> mockData =
+			linear_test_env->getOneDimensionalRegressionData(use_bias);
+
+	// fetch data from test environment
```
karlnapf: no need for comment

tests/unit/regression/LibLinearRegression_unittest.cc
```diff
@@ -0,0 +1,100 @@
...
+	CDenseFeatures<float64_t>* train_feats = mockData->get_features_train();
+	CDenseFeatures<float64_t>* test_feats = mockData->get_features_test();
+
+	CRegressionLabels* labels_test = mockData->get_labels_test();
+	CRegressionLabels* labels_train = mockData->get_labels_train();
+
+	// train model
```
karlnapf: same here :)

tests/unit/regression/LibLinearRegression_unittest.cc
```diff
@@ -0,0 +1,100 @@
...
+	CLibLinearRegression* lr =
+	    new CLibLinearRegression(1., train_feats, labels_train);
+	lr->set_use_bias(use_bias);
+	lr->set_epsilon(epsilon);
+	lr->train();
+
+	// run predictions
```
karlnapf: and here :) you get the point ...

tests/unit/regression/LibLinearRegression_unittest.cc
```diff
@@ -0,0 +1,100 @@
...
+	// test predictions
+	for (index_t i = 0; i < mockData->get_test_size(); ++i)
+		EXPECT_NEAR(predicted_labels->get_label(i), labels_test->get_label(i), 1E-5);
+
+	/* clean up */
+	SG_UNREF(predicted_labels);
+	SG_UNREF(lr);
+}
+
+TEST(LibLinearRegression, lr_without_bias)
```
karlnapf: I wonder, souldnt we pull out the shared code to avoid this heavy redundancy?

src/shogun/regression/svr/LibLinearRegression.cpp
```diff
@@ -304,8 +315,8 @@ void CLibLinearRegression::solve_l2r_l1l2_svr(SGVector<float64_t>& w, const libl
 		iter++;
 
 		pb.print_absolute(
-		    Gnorm1_new, -CMath::log10(Gnorm1_new),
-		    -CMath::log10(eps * Gnorm1_init), -CMath::log10(Gnorm1_init));
+			Gnorm1_new, -CMath::log10(Gnorm1_new),
```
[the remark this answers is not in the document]
gf712: Yes, this one comes from the clang format when I run it locally with the project format file.. I did find it odd!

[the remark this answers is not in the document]
karlnapf: usually, clang only formats the changes lines, i.e. if you don't touch it wont format. Did you use the correct commit hash for the format? I mean it doesnt matter too much tbh.

karlnapf:
This seems ok to me now. I will wait for another day or so in case somebody else reviews, otherwise it's good to merge. Thanks a ton for the patch! Maybe we can test all linear models in this style once this is merged .... Ill put some links soon

PR #4409, opened 2018-10-23
Meta examples generation bug fix (fixes #4408)

### LINKED ISSUE #4408: Travis bug generating trained_model_serialization_test.h
There is currently an issue building in one of the Travis environments ([example](https://travis-ci.org/shogun-toolbox/shogun/jobs/443619544#L2902)). The issue gives an error like this:
```
[ 58%] Generating trained_model_serialization_test.h
  File "/opt/shogun/tests/unit/base/trained_model_serialization_test.cc.py", line 169, in <module>
    config_file = sys.argv[3]
```
This is because the `examples/meta/tags` file is not passed to trained_model_serialization_test.cc.py, and there are only two out of the three expected args.

The issue is that `CTAGS_FILE` is defined in `shogun/examples/meta/CMakeLists.txt`, which is added in this line:
https://github.com/shogun-toolbox/shogun/blob/e338426fc5777386b30eb0adee9e63bf61076e70/CMakeLists.txt#L516-L518
But the `trained_model_serialization_test.cc.py` command is defined in `tests/meta`, which is added before setting `CTAGS_FILE` to `examples/meta/tags`:
https://github.com/shogun-toolbox/shogun/blob/e338426fc5777386b30eb0adee9e63bf61076e70/CMakeLists.txt#L502-L503

Switching the order of execution of the two blocks where these lines are defined in the project `CMakeLists.txt` fixes the bug (on my machine) and doesn't seem to break anything else.

I'll submit a PR if this makes sense!

gf712:
As explained in #4408 switching execution order defines Cmake variables properly

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -484,20 +484,12 @@ OPTION(TRAVIS_DISABLE_UNIT_TESTS "Disable unit testing to speed up jobs on travi
 OPTION(TRAVIS_DISABLE_META_CPP "Disable cpp meta examples and integration testing to speed up jobs on travis-ci" OFF)
 OPTION(DISABLE_META_INTEGRATION_TESTS "Disable meta integration testing to speed up build" OFF)
 
-IF(EXISTS ${CMAKE_SOURCE_DIR}/examples)
-	IF(ENABLE_TESTING AND NOT BUILD_EXAMPLES)
-		message(STATUS "Tests require (disabled) examples, enabling.")
-	ENDIF()
-	IF(ENABLE_TESTING OR BUILD_EXAMPLES)
-		add_subdirectory(${CMAKE_SOURCE_DIR}/examples)
-	ENDIF()
-
-	IF(BUILD_META_EXAMPLES)
-		# allow meta examples without adding examples dir itself
-		add_subdirectory(${CMAKE_SOURCE_DIR}/examples/meta)
-	ENDIF()
+IF((ENABLE_TESTING AND NOT TRAVIS_DISABLE_UNIT_TESTS AND EXISTS ${CMAKE_SOURCE_DIR}/tests/unit)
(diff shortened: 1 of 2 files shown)


karlnapf:
Travis is alive again! :)

karlnapf:
Seems good to me. I will merge so that we have CI for the other tests. Thanks a lot for this

CMakeLists.txt
```diff
@@ -500,23 +505,23 @@ IF(ENABLE_TESTING)
...
-        ENDIF()
+			add_subdirectory(${CMAKE_SOURCE_DIR}/tests/meta)
+		ENDIF()
 	ENDIF()
 ENDIF()
 
 IF(EXISTS ${CMAKE_SOURCE_DIR}/examples)
 	IF(ENABLE_TESTING AND NOT BUILD_EXAMPLES)
-	    message(STATUS "Tests require (disabled) examples, enabling.")
+		message(STATUS "Tests require (disabled) examples, enabling.")
```
karlnapf: mmh these whitespace changes pollute the diff. Could you remove them?

CMakeLists.txt
```diff
@@ -500,23 +505,23 @@ IF(ENABLE_TESTING)
...
-            add_subdirectory(${CMAKE_SOURCE_DIR}/tests/meta)
-        ENDIF()
+			add_subdirectory(${CMAKE_SOURCE_DIR}/tests/meta)
+		ENDIF()
 	ENDIF()
 ENDIF()
 
 IF(EXISTS ${CMAKE_SOURCE_DIR}/examples)
 	IF(ENABLE_TESTING AND NOT BUILD_EXAMPLES)
-	    message(STATUS "Tests require (disabled) examples, enabling.")
```
karlnapf: these whitespace changes are polluting the diff a bit. Would you mind removing them?

karlnapf: well, next time!

gf712: Hi, sorry about that!

PR #4412, opened 2018-11-07
AnyParameterProperties refactor

gf712:
This PR refactors the AnyParameterProperties class to use a mask (`m_mask_attribute`) to store parameter options:
 * HYPERPARAMETER: the parameter determines how the training is performed, e.g. regularisation
 * GRADIENT_PARAM: the parameter is used in gradient updates
 * MODEL_PARAM: model parameter that is trained, e.g. weight and bias
This makes the use of Enums in `AnyParameter.h` not necessary anymore, and the corresponding getter methods can be replaced with `bool` return type.  
The mask uses bit shift operations to keep information in a hexadecimal representation.

### DIFF
--- a/src/shogun/base/AnyParameter.h
+++ b/src/shogun/base/AnyParameter.h
@@ -28,13 +28,9 @@ namespace shogun
 		GRADIENT_AVAILABLE = 1
 	};
 	/** @brief Class AnyParameterProperties keeps track of parameter properties.
-	 *
-	 * Currently there are three types of parameters:
-	 * -# HYPERPARAMETER: the parameter determines how the training is
-	 * performed, e.g. regularisation
-	 * -# GRADIENT_PARAM: the parameter is used in gradient updates
-	 * -# MODEL_PARAM: model parameter that is trained, e.g. weight and bias
-	 * vectors and bias
+	 * The parameter properties can be either true or false.
+	 * These properties describe if a parameter is for example a hyperparameter
+	 * or if it has a gradient.
 	 */
 	class AnyParameterProperties
 	{
@@ -43,12 +39,20 @@ namespace shogun
 		static const int32_t GRADIENT_PARAM = 0x00000010;
 		static const int32_t MODEL_PARAM = 0x00000100;
 
-		/** default constructor where all parameter types are switched to false
+		/** Default constructor where all parameter properties are false
 		 */
 		AnyParameterProperties() : m_description(), m_mask_attribute(0x00000000)
 		{
 		}
-		/** legacy constructor */
+		/** Constructor
+		 * @param description parameter description
+		 * @param hyperparameter set to true for parameters that determine
+		 * how training is performed, e.g. regularisation parameters
+		 * @param gradient set to true for parameters required for gradient
+		 * updates
+		 * @param model set to true for parameters used in inference, e.g.
+		 * weights and bias
+		 * */
 		AnyParameterProperties(
 		    std::string description,
 		    EModelSelectionAvailability model_selection = MS_NOT_AVAILABLE,
@@ -65,13 +69,15 @@ namespace shogun
 			if (hyperparameter)
 				m_mask_attribute |= HYPERPARAMETER;
 		}
-		/** mask constructor */
-		AnyParameterProperties(std::string description, int32_t mask_attribute)
+		/** Mask constructor
+		 * @param description parameter description
+		 * @param attribute_mask mask encoding parameter properties
+		 * */
 		    : m_description(description)
 		{
 			m_mask_attribute = mask_attribute;
 		}
-		/** copy contructor */
+		/** Copy contructor */
 		AnyParameterProperties(const AnyParameterProperties& other)
 		    : m_description(other.m_description),
 		      m_model_selection(other.m_model_selection),
@@ -79,12 +85,11 @@ namespace shogun
 		      m_mask_attribute(other.m_mask_attribute)
 		{
 		}
-		/** description getter */
+		const std::string& get_description() const
 		std::string get_description() const
 		{
 			return m_description;
 		}
-		/** model selection flag getter */
 		EModelSelectionAvailability get_model_selection() const
 		{
 			EModelSelectionAvailability return_val;
@@ -95,7 +100,6 @@ namespace shogun
 			return return_val;
 		}
 
-		/** gradient flag getter */
 		EGradientAvailability get_gradient() const
 		{
 			EGradientAvailability return_val;
@@ -106,7 +110,6 @@ namespace shogun
 			return return_val;
 		}
 
-		/** hyperparameter flag getter */
 		bool get_hyperparameter() const
 		{
 			bool return_val;
@@ -121,7 +124,7 @@ namespace shogun
 		std::string m_description;
 		EModelSelectionAvailability m_model_selection;
 		EGradientAvailability m_gradient;
-		/** mask to store all param flags*/
+		int32_t m_attribute_mask;
 		int32_t m_mask_attribute;
 	};
 
--- a/src/shogun/base/AnyParameter.h
+++ b/src/shogun/base/AnyParameter.h
@@ -1,3 +1,9 @@
+/*
+ * This software is distributed under BSD 3-clause license (see LICENSE file).
+ *
+ * Authors: Heiko Strathmann, Gil Hoben
+ */
+
 #ifndef __ANYPARAMETER_H__
 #define __ANYPARAMETER_H__
 
@@ -27,7 +33,7 @@ namespace shogun
 	 * -# HYPERPARAMETER: the parameter determines how the training is
 	 * performed, e.g. regularisation
 	 * -# GRADIENT_PARAM: the parameter is used in gradient updates
-	 * -# MODELSELECTION_PARAM: model parameter that is trained, e.g. weight
+	 * -# MODEL_PARAM: model parameter that is trained, e.g. weight and bias
 	 * vectors and bias
 	 */
 	class AnyParameterProperties
@@ -35,7 +41,7 @@ namespace shogun
 	public:
 		static const int32_t HYPERPARAMETER = 0x00000001;
 		static const int32_t GRADIENT_PARAM = 0x00000010;
-		static const int32_t MODELSELECTION_PARAM = 0x00000100;
+		static const int32_t MODEL_PARAM = 0x00000100;
 
 		/** default constructor where all parameter types are switched to false
 		 */
@@ -53,7 +59,7 @@ namespace shogun
 		{
 			m_mask_attribute = 0x00000000;
 			if (model_selection)
-				m_mask_attribute |= MODELSELECTION_PARAM;
+				m_mask_attribute |= MODEL_PARAM;
 			if (gradient)
 				m_mask_attribute |= GRADIENT_PARAM;
 			if (hyperparameter)
@@ -82,7 +88,7 @@ namespace shogun
 		EModelSelectionAvailability get_model_selection() const
 		{
 			EModelSelectionAvailability return_val;
-			if (m_mask_attribute & MODELSELECTION_PARAM)
+			if (m_mask_attribute & HYPERPARAMETER)
 				return_val = EModelSelectionAvailability::MS_AVAILABLE;
 			else
 				return_val = EModelSelectionAvailability::MS_NOT_AVAILABLE;
@@ -169,6 +175,6 @@ namespace shogun
 		Any m_value;
 		AnyParameterProperties m_properties;
 	};
-}
+} // namespace shogun
 
 #endif
--- a/src/shogun/base/AnyParameter.h
+++ b/src/shogun/base/AnyParameter.h
@@ -35,13 +35,14 @@ namespace shogun
 	class AnyParameterProperties
 	{
 	public:
-		static const int32_t HYPERPARAMETER = 0x00000001;
-		static const int32_t GRADIENT_PARAM = 0x00000010;
-		static const int32_t MODEL_PARAM = 0x00000100;
+		static const int32_t HYPER = 1u << 0;
+		static const int32_t GRADIENT = 1u << 1;
+		static const int32_t MODEL = 1u << 2;
 
 		/** Default constructor where all parameter properties are false
 		 */
-		AnyParameterProperties() : m_description(), m_mask_attribute(0x00000000)
+		AnyParameterProperties()
+		    : m_description("No description given"), m_attribute_mask(0)
 		{
 		}
 		/** Constructor
@@ -55,69 +56,50 @@ namespace shogun
 		 * */
 		AnyParameterProperties(
 		    std::string description,
-		    EModelSelectionAvailability model_selection = MS_NOT_AVAILABLE,
+		    EModelSelectionAvailability hyperparameter = MS_NOT_AVAILABLE,
 		    EGradientAvailability gradient = GRADIENT_NOT_AVAILABLE,
-		    bool hyperparameter = false)
-		    : m_description(description), m_model_selection(model_selection),
+		    bool model = false)
+		    : m_description(description), m_model_selection(hyperparameter),
 		      m_gradient(gradient)
 		{
-			m_mask_attribute = 0x00000000;
-			if (model_selection)
-				m_mask_attribute |= MODEL_PARAM;
-			if (gradient)
-				m_mask_attribute |= GRADIENT_PARAM;
-			if (hyperparameter)
-				m_mask_attribute |= HYPERPARAMETER;
+			m_attribute_mask = (hyperparameter << 0 & HYPER) |
+			                   (gradient << 1 & GRADIENT) |
+			                   (model << 2 & MODEL);
 		}
 		/** Mask constructor
 		 * @param description parameter description
 		 * @param attribute_mask mask encoding parameter properties
 		 * */
+		AnyParameterProperties(std::string description, int32_t attribute_mask)
 		    : m_description(description)
 		{
-			m_mask_attribute = mask_attribute;
+			m_attribute_mask = attribute_mask;
 		}
 		/** Copy contructor */
 		AnyParameterProperties(const AnyParameterProperties& other)
 		    : m_description(other.m_description),
 		      m_model_selection(other.m_model_selection),
 		      m_gradient(other.m_gradient),
-		      m_mask_attribute(other.m_mask_attribute)
+		      m_attribute_mask(other.m_attribute_mask)
 		{
 		}
 		const std::string& get_description() const
-		std::string get_description() const
 		{
 			return m_description;
 		}
 		EModelSelectionAvailability get_model_selection() const
 		{
-			EModelSelectionAvailability return_val;
-			if (m_mask_attribute & HYPERPARAMETER)
-				return_val = EModelSelectionAvailability::MS_AVAILABLE;
-			else
-				return_val = EModelSelectionAvailability::MS_NOT_AVAILABLE;
-			return return_val;
+			return static_cast<EModelSelectionAvailability>(
+			    (m_attribute_mask & HYPER) > 0);
 		}
-
 		EGradientAvailability get_gradient() const
 		{
-			EGradientAvailability return_val;
-			if (m_mask_attribute & GRADIENT_PARAM)
-				return_val = EGradientAvailability::GRADIENT_AVAILABLE;
-			else
-				return_val = EGradientAvailability::GRADIENT_NOT_AVAILABLE;
-			return return_val;
+			return static_cast<EGradientAvailability>(
+			    (m_attribute_mask & GRADIENT) > 0);
 		}
-
-		bool get_hyperparameter() const
+		bool get_model() const
 		{
-			bool return_val;
-			if (m_mask_attribute & HYPERPARAMETER)
-				return_val = true;
-			else
-				return_val = false;
-			return return_val;
+			return static_cast<bool>(m_attribute_mask & MODEL);
 		}
 
 	private:
@@ -125,7 +107,6 @@ namespace shogun
 		EModelSelectionAvailability m_model_selection;
 		EGradientAvailability m_gradient;
 		int32_t m_attribute_mask;
-		int32_t m_mask_attribute;
 	};
 
 	class AnyParameter
--- a/src/shogun/base/AnyParameter.h
+++ b/src/shogun/base/AnyParameter.h
@@ -47,14 +47,11 @@ namespace shogun
 	class AnyParameterProperties
 	{
 	public:
-		static const int32_t HYPER = 1u << 0;
-		static const int32_t GRADIENT = 1u << 1;
-		static const int32_t MODEL = 1u << 2;
-
 		/** Default constructor where all parameter properties are false
 		 */
 		AnyParameterProperties()
-		    : m_description("No description given"), m_attribute_mask(0)
+		    : m_description("No description given"),
+		      m_attribute_mask(ParameterProperties())
 		{
 		}
 		/** Constructor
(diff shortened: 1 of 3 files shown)


src/shogun/base/AnyParameter.h
```diff
@@ -21,49 +27,102 @@ namespace shogun
...
+	 * -# MODEL_PARAM: model parameter that is trained, e.g. weight and bias
+	 * vectors and bias
+	 */
 	class AnyParameterProperties
 	{
 	public:
-		AnyParameterProperties()
-		    : m_description(), m_model_selection(MS_NOT_AVAILABLE),
-		      m_gradient(GRADIENT_NOT_AVAILABLE)
+		static const int32_t HYPERPARAMETER = 0x00000001;
```
lisitsyn: What about making it an int enum?

src/shogun/base/AnyParameter.h
```diff
@@ -21,49 +27,102 @@ namespace shogun
...
+			return return_val;
 		}
 
+		/** gradient flag getter */
 		EGradientAvailability get_gradient() const
 		{
-			return m_gradient;
+			EGradientAvailability return_val;
+			if (m_mask_attribute & GRADIENT_PARAM)
+				return_val = EGradientAvailability::GRADIENT_AVAILABLE;
```
lisitsyn: it makes sense to return immediately without return_val

src/shogun/base/AnyParameter.h
karlnapf: Check this
https://stackoverflow.com/questions/4920783/combining-enum-value-using-bitmask

src/shogun/base/AnyParameter.h
```diff
@@ -21,49 +27,102 @@ namespace shogun
 		GRADIENT_NOT_AVAILABLE = 0,
 		GRADIENT_AVAILABLE = 1
 	};
-
+	/** @brief Class AnyParameterProperties keeps track of parameter properties.
+	 *
+	 * Currently there are three types of parameters:
+	 * -# HYPERPARAMETER: the parameter determines how the training is
```
karlnapf: Pls don’t put such things into the docs, they outdated at light speed.
Better describe that there are attributes, maybe an example, but not losing them all

src/shogun/base/AnyParameter.h
```diff
@@ -21,49 +27,102 @@ namespace shogun
...
+	 * vectors and bias
+	 */
 	class AnyParameterProperties
 	{
 	public:
-		AnyParameterProperties()
-		    : m_description(), m_model_selection(MS_NOT_AVAILABLE),
-		      m_gradient(GRADIENT_NOT_AVAILABLE)
+		static const int32_t HYPERPARAMETER = 0x00000001;
+		static const int32_t GRADIENT_PARAM = 0x00000010;
```
karlnapf: You can also use the << operator to shift the 1 a desired number of digits
See the stack overflow post

src/shogun/base/AnyParameter.h
gf712: OK!

src/shogun/base/AnyParameter.h
gf712: I had a look at it, I do that, I just thought that the hexadecimal representation is more readable, no?

karlnapf: at the end of the day, it is style. You can either define the mask as you did with hexadecimal as you did there. Or, instead, you can generate those masks on the fly using a shift operator, and then the `HYPERPARAMETER` constant is the number of bits that you shift, i.e.
```
my_mask &&
(1 << HYPERPARAMETER) &&
(1 << GRADIENT_AVAILABLE) &&
```
 The latter has the advantage that somebody reading the code never is exposed to hexadecimal numbers, nor powers or two (which can be confusing). On the other hand, masking operations (as what I wrote above) becomes a bit more notation heavy. Idk what is best to be honest. I think the shifting is more standard and therefore I think we should follow that.

Ah one thing I think is good about the shifting approach is that you can change the size of the mask easily. You are using a 32bit int above, but you only write down the last 8 bits, which is weird. Imagine you now changed the mask to 64 bit. Then the hexadecimals would be quite long, whereas the shifts wouldnt be.

@lisitsyn @vigsterkr ?

src/shogun/base/AnyParameter.h
```diff
@@ -21,49 +27,102 @@ namespace shogun
...
+		static const int32_t MODEL_PARAM = 0x00000100;
+
+		/** default constructor where all parameter types are switched to false
```
karlnapf: Minor: sentences start with capital letters

src/shogun/base/AnyParameter.h
```diff
@@ -21,49 +27,102 @@ namespace shogun
...
+		 */
+		AnyParameterProperties() : m_description(), m_mask_attribute(0x00000000)
 		{
 		}
+		/** legacy constructor */
```
karlnapf: I would remove that. Until we haven't removed it, it is not legacy.

src/shogun/base/AnyParameter.h
```diff
@@ -21,49 +27,102 @@ namespace shogun
...
+				return_val = false;
+			return return_val;
 		}
 
 	private:
 		std::string m_description;
 		EModelSelectionAvailability m_model_selection;
 		EGradientAvailability m_gradient;
+		/** mask to store all param flags*/
+		int32_t m_mask_attribute;
```
karlnapf: what about calling it "attribute_mask" ? :)

src/shogun/base/AnyParameter.h
```diff
@@ -21,49 +27,102 @@ namespace shogun
...
 		{
+			m_mask_attribute = 0x00000000;
+			if (model_selection)
+				m_mask_attribute |= MODEL_PARAM;
+			if (gradient)
+				m_mask_attribute |= GRADIENT_PARAM;
+			if (hyperparameter)
+				m_mask_attribute |= HYPERPARAMETER;
+		}
+		/** mask constructor */
```
karlnapf: if you add doxygen docs, pls document all parameters
minor: capital M

src/shogun/base/AnyParameter.h
```diff
@@ -21,49 +27,102 @@ namespace shogun
...
 		    std::string description,
 		    EModelSelectionAvailability model_selection = MS_NOT_AVAILABLE,
-		    EGradientAvailability gradient = GRADIENT_NOT_AVAILABLE)
+		    EGradientAvailability gradient = GRADIENT_NOT_AVAILABLE,
+		    bool hyperparameter = false)
 		    : m_description(description), m_model_selection(model_selection),
 		      m_gradient(gradient)
 		{
+			m_mask_attribute = 0x00000000;
+			if (model_selection)
```
karlnapf: wouldnt this work?
```
m_attribute_mask = 
(MODEL_PARAM & model_selection) |
(GRADIENT_PARAM & gradient) |
(HYPERPARAMETER & hyperparameter)
```

karlnapf: or in the new style
```
m_attribute_mask = 
((1<< MODEL_PARAM) & model_selection) |
((1<< GRADIENT_PARAM) & gradient) |
((1<<HYPERPARAMETER) & hyperparameter)
```

src/shogun/base/AnyParameter.h
karlnapf: I would call this GRADIENT_AVAILABLE

src/shogun/base/AnyParameter.h
gf712: Yup, I just noticed that I switched the order...

src/shogun/base/AnyParameter.h
gf712: True, that would be more concise. Is it worth writing a static member that is the default, i.e. 0?

src/shogun/base/AnyParameter.h
karlnapf: can we keep the naming self-consistent
HYPER_PARAM
or
HYPER_PARAMETER
or name the other one
MODELPARAMETER
or rather
MODEL_PARAMETER

Actually, in favour of short and concise names, why dont we use
HYPER
MODEL
GRADIENT
?
And then the constant itself can have a doxygen string to say what that means?

src/shogun/base/AnyParameter.h
```diff
@@ -21,49 +27,102 @@ namespace shogun
...
 		AnyParameterProperties(const AnyParameterProperties& other)
 		    : m_description(other.m_description),
 		      m_model_selection(other.m_model_selection),
-		      m_gradient(other.m_gradient)
+		      m_gradient(other.m_gradient),
+		      m_mask_attribute(other.m_mask_attribute)
 		{
 		}
-
+		/** description getter */
```
karlnapf: superflous doc, as the method name explains it, i would just remove it

src/shogun/base/AnyParameter.h
```diff
@@ -21,49 +27,102 @@ namespace shogun
...
 		std::string get_description() const
```
karlnapf: minor what about returning a const string reference

src/shogun/base/AnyParameter.h
```diff
@@ -21,49 +27,102 @@ namespace shogun
...
 		{
 			return m_description;
 		}
-
+		/** model selection flag getter */
```
karlnapf: same here, would remove the comment

src/shogun/base/SGObject.h
```diff
@@ -676,7 +676,10 @@ class CSGObject
 	void watch_param(
 		const std::string& name, T* value,
 		AnyParameterProperties properties = AnyParameterProperties(
-		    "Unknown parameter", MS_NOT_AVAILABLE, GRADIENT_NOT_AVAILABLE))
+			"Unknown parameter",
+			AnyParameterProperties::HYPERPARAMETER |
```
karlnapf: instead of doing this here, I think it might be better to offer a constructor that does not take any properties, and then set the default values inside there. We need to discuss clever default values as well btw

src/shogun/base/AnyParameter.h
gf712: OK! But hyperparameter is a term that could be misinterpreted if it written as HYPER, no?

src/shogun/base/SGObject.h
```diff
@@ -715,7 +721,10 @@ class CSGObject
 	void watch_param(
 		const std::string& name, T** value, S* rows, S* cols,
 		AnyParameterProperties properties = AnyParameterProperties(
-		    "Unknown parameter", MS_NOT_AVAILABLE, GRADIENT_NOT_AVAILABLE))
+			"Unknown parameter",
```
karlnapf: Same with the description, I would offer a ctor that does not ask for one, and then set it to the empty string or "No description given" in the ctor, but not here in `SGObject.h`

src/shogun/base/AnyParameter.h
karlnapf: If the default is 0, then we don't need that.
I would be explicit about a non-zero default, and add a private `init` method that sets it and that is called by all ctors.

src/shogun/base/AnyParameter.h
karlnapf: I think it is pretty clear since the class is ending with "Parameter", so everything in there is one ....

src/shogun/base/AnyParameter.h
gf712: So would we have something like this:

```
enum ParameterAvailability
{
    HYPER = 0,
    GRADIENT = 1,
    MODEL = 2
};

...

m_attribute_mask =
(hyperparameter<<HYPER) |
(gradient<<GRADIENT) |
(model_selection<<MODEL);
```

And then get the values like this:
```
bool is_hyperparamater()
     return m_attribute_mask & (1<<HYPER)
```

src/shogun/base/AnyParameter.h
gf712: The name MODEL causes conflicts if we use an enum, so will have to use something more specific, or just have these values as static class members of AnyParameterProperties.

src/shogun/base/AnyParameter.h
gf712: Do you mean with a one liner like this? 
```
return static_cast<EGradientAvailability>((m_mask_attribute & GRADIENT_PARAM) > 0);
```

src/shogun/base/AnyParameter.h
gf712: OK! I think you're right, I have changed the class members to use bit shifts, it is a bit cleaner.

src/shogun/base/AnyParameter.h
gf712: Yes, I think we could do that. It seems to be pretty much standard now to return const references for std::string. It can sometimes lead to odd behaviours, but I don't think we need to worry about those.. 
http://thesyntacticsugar.blogspot.com/2011/09/evil-side-of-returning-member-as-const.html

src/shogun/base/AnyParameter.h
karlnapf: yes, please don't define them globally, but as a static member

src/shogun/base/AnyParameter.h
karlnapf: yes, but the first rule of the C++ committee is:
"The C++ language does not prevent people from shooting themselves in their feet" ;)

src/shogun/base/AnyParameter.h
```diff
@@ -28,101 +28,85 @@ namespace shogun
...
 	};
 	/** @brief Class AnyParameterProperties keeps track of parameter properties.
-	 *
-	 * Currently there are three types of parameters:
-	 * -# HYPERPARAMETER: the parameter determines how the training is
-	 * performed, e.g. regularisation
-	 * -# GRADIENT_PARAM: the parameter is used in gradient updates
-	 * -# MODEL_PARAM: model parameter that is trained, e.g. weight and bias
-	 * vectors and bias
+	 * The parameter properties can be either true or false.
```
karlnapf: like!
Maybe also mention descriptions in the first sentence " ... of parameter meta information, such as properties and descriptions"

src/shogun/base/AnyParameter.h
```diff
@@ -28,101 +28,85 @@ namespace shogun
...
 	 */
 	class AnyParameterProperties
 	{
 	public:
-		static const int32_t HYPERPARAMETER = 0x00000001;
-		static const int32_t GRADIENT_PARAM = 0x00000010;
-		static const int32_t MODEL_PARAM = 0x00000100;
+		static const int32_t HYPER = 1u << 0;
+		static const int32_t GRADIENT = 1u << 1;
+		static const int32_t MODEL = 1u << 2;
```
karlnapf: ah wait. the thing I had in mind was 
```
static const int HYPER = 1;
```
and then
```
mask = (1 << HYPER) | (1 << MODEL);
```
This is now difference, as you are effectively storing masks themselves in these const members, not the number of bit shifts. Or am I wrong? Let me stare a bit more to understand

src/shogun/base/AnyParameter.h
```diff
@@ -28,101 +28,85 @@ namespace shogun
...
 		      m_gradient(gradient)
 		{
-			m_mask_attribute = 0x00000000;
-			if (model_selection)
-				m_mask_attribute |= MODEL_PARAM;
-			if (gradient)
-				m_mask_attribute |= GRADIENT_PARAM;
-			if (hyperparameter)
-				m_mask_attribute |= HYPERPARAMETER;
+			m_attribute_mask = (hyperparameter << 0 & HYPER) |
```
karlnapf: I was more thinking
`(1 << HYPER) & model_selection | (1 << GRADIENT) && gradient` etc

src/shogun/base/AnyParameter.h
```diff
@@ -28,101 +28,85 @@ namespace shogun
...
 		EModelSelectionAvailability get_model_selection() const
 		{
-			EModelSelectionAvailability return_val;
-			if (m_mask_attribute & HYPERPARAMETER)
-				return_val = EModelSelectionAvailability::MS_AVAILABLE;
-			else
-				return_val = EModelSelectionAvailability::MS_NOT_AVAILABLE;
-			return return_val;
+			return static_cast<EModelSelectionAvailability>(
+			    (m_attribute_mask & HYPER) > 0);
```
karlnapf: and here I was thinking
`m_attribute_mask & (1<<HYPER)`

karlnapf: wouldn't we want != 0. Also, this is not very explicit ... I know sergey suggested, but I need to think a bit here to understand this so not sure that is the best solution

src/shogun/base/AnyParameter.h
karlnapf: fine as well

src/shogun/base/AnyParameter.h
gf712: I thought about that, but then I thought it was cleaner to have the constructor "know" at what position the bit shift is required? Not sure if that makes sense?

src/shogun/base/AnyParameter.h
gf712: Do you mean:
```
static const int HYPER = 0;
static const int GRADIENT = 1;
static const int MODEL = 2;
```

src/shogun/base/AnyParameter.h
karlnapf: not sure :) haha
we can discuss in irc

src/shogun/base/AnyParameter.h
karlnapf: yes, but start at 1

gf712: OK, if we have:
```
  static const int HYPER = 1;
  static const int GRADIENT = 2;
  static const int MODEL = 3;

  bool model_selection = true;
  bool gradient = true;

  int mask = (1 << HYPER) & model_selection | (1 << GRADIENT) & gradient;
```
Then the above mask will always be false/0 no?

gf712: Do you mean more like this:
```
int mask = (model_selection << HYPER) | (gradient << GRADIENT)
```
which is the equivalent of shift bit at position X (1,2 or 3 in this case) if the function parameter is true.

karlnapf: sure, the thing you wrote is (or should be, not sure what happens when you shift a boolean true) a neater version if what I wrote above.

gf712: Ah now I remember why I did it this way.. Basically if you want to provide an interface to create your own mask like this
```
AnyParameterProperties("My description",  
                                           AnyParameterProperties::MODEL |
                                           AnyParameterProperties::GRADIENT)
```
you need to have the values of the static member with the shift. Unless I am missing something?

karlnapf: Not sure what you mean, but what I have seen (and been doing) would result in
```
AnyParameterProperties("My description",  
                       (1 << AnyParameterProperties::MODEL) |
                       (1 << AnyParameterProperties::GRADIENT))
```
It is not as neat, you are totally right!. BUT it can be hidden quite easily either.
I mean, it is your code, Ill leave it to you. Not sure whether it is possible to say what is better now, we will have to commit to something and then see if it works :) So go ahead with the above solution and then we see where we get

gf712: OK! Btw why would does it need to start at 1, instead of 0?

karlnapf: ah, no that was nonsense (just needed to make sure that you never get an all zero mask)

BTW I also realised why the micro-controller people store the number of bits rather than the full masks: it uses less memory (which is not really relevant here, but still good to think about it for a moment).
```
byte HYPER = 2
...
byte BLA = 255
```
so with an 8bit integer, you can do 255 attributes. If you wanted to store the mask for each attribute instead (each mask is 8bit), you would have to store 255 * 8bit, which is quite a difference , especially if you have only 4k memory as on an AVR ;)

How many parameters are registered during a shogun run? Let's say 100 is realistic. Then we have 100 byte usage if we store only the offset, and 2.5kB if we use the masks. In a data heavy environment, and since we won't register many many more parameters this doesn't matter. For 1000 parameters it is still ok, 10^6 will result in some pretty big differences.

src/shogun/base/AnyParameter.h
lisitsyn: Oh that's too low-level. Can we introduce a constexpr function for that?

src/shogun/base/AnyParameter.h
lisitsyn: Lets typedef the actual type so we can change later.

src/shogun/base/AnyParameter.h
lisitsyn: Yeah maybe with some helper function so it looks shorter? There is some repeated code in these flags.

src/shogun/base/AnyParameter.h
karlnapf: yeah actually good point! :)

karlnapf:
Any news here?  :)

gf712:
I did another round of clean up, but I haven't pushed it yet. Got a bit distracted looking into constexpr, as I have never used them before in my own code.. As for using a typedef I am not sure how it would be used as we are storing data in a int mask..?
 Also if it is too level couldn't we use a `struct` instead with booleans? It would be a bit less space efficient, but would be quite easy to expand in the future?

karlnapf:
I googled a bit and found this, maybe interesting? 
http://blog.bitwigglers.org/using-enum-classes-as-type-safe-bitmasks/

gf712:
So do you think we could just copy this? https://www.justsoftwaresolutions.co.uk/files/bitmask_operators.hpp
It seems to be the original code of that post and seems to cover most operations of interest.. Otherwise he wrote this more complete example.. https://github.com/Dalzhim/ArticleEnumClass-v2

karlnapf:
Yeah ok, I think this is cool (I think the first simple version is fine)
 
* We do explicitly store the mask (1,2,4,8,16, etc), rather than the offset. Easy explicit usage `AnyParameterProperties::MODEL | AnyParameterProperties::HYPER`
* we make this typesafe through the code snippet

gf712:
Is it worth overloading the enum operators in a separate header file or does it make sense to keep it all in `AnyParameter.h`?

gf712:
Is it worth overloading the enum operators in a separate header file or does it make sense to keep it all in `AnyParameter.h`?

karlnapf:
As it is general purpose code, I would copy it to `lib` or so

karlnapf:
E.g. `lib/bitmasking.h`

gf712:
Not too happy with
```

**Comment by gf712** [comment]
Not too happy with:
```
m_attribute_mask = ParameterProperties();
if (hyperparameter)
    m_attribute_mask |= ParameterProperties::HYPER;
if (gradient)
    m_attribute_mask |= ParameterProperties::GRADIENT;
if (model)
    m_attribute_mask |= ParameterProperties::MODEL;
```
but given the restrictions that are imposed in `bitmask_operators.h` it seems to me to be the best solution..

**Comment by karlnapf** [comment]
This comes from the fact that since it is now typesafe, you cannot `&` the mask with a boolean anymore?
It is not a big issue as we want to get rid of those booleans eventually anyways

File: src/shogun/lib/bitmask_operators.h [hunk 19]
```diff
@@ -0,0 +1,110 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY
+// KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE
+// WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR
+// PURPOSE, TITLE AND NON-INFRINGEMENT. IN NO EVENT SHALL THE
+// COPYRIGHT HOLDERS OR ANYONE DISTRIBUTING THE SOFTWARE BE
+// LIABLE FOR ANY DAMAGES OR OTHER LIABILITY, WHETHER IN
+// CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+// CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
```
karlnapf: You could put "modified by G.F" here if you want

src/shogun/base/SGObject.h
```diff
@@ -735,9 +731,9 @@ class CSGObject
 		BaseTag tag(name);
 		AnyParameterProperties properties(
 			"Dynamic parameter",
-			!AnyParameterProperties::HYPER |
-			!AnyParameterProperties::GRADIENT |
-			!AnyParameterProperties::MODEL);
+			ParameterProperties::HYPER |
+			ParameterProperties::GRADIENT |
+            ParameterProperties::MODEL);
```
karlnapf: the whitespaces seem weird here

karlnapf:
Ok, we might have to touch this again at some point, but I think we can move ahead now

karlnapf:
Shall we have a little test with a minimal mock class where we register some parameters with certain properties and make sure everything makes sense?

Next step would be: Let's `observe` every model parameter when `put` is called.

gf712:
> This comes from the fact that since it is now typesafe, you cannot `&` the mask with a boolean anymore?
Yes, and adding `bool` to the allowed types would defeat the point of using `enum class`.

src/shogun/lib/bitmask_operators.h
gf712: Ah ok! I wasn't sure how it works with the boost software license!

src/shogun/base/SGObject.h
gf712: Yup, and should change it to default to `false` instead of all `true`.

PR #4415, opened 2018-11-15
merge

gf712:


karlnapf:
I think we mentioned somewhere in the docs that we want to use rebase based pulling

PR #4417, opened 2018-11-16
SG_ADD refactor

gf712:
Now that we can register parameter properties with a single enum we don't need to have several SG_ADD macros with variable number of parameters.
I replaced with a regular expression the following expressions in the whole project:
- MS_NOT_AVAILABLE()); -> ParameterProperties());
- MS_AVAILABLE()); -> ParameterProperties::HYPER);
- MS_AVAILABLE, GRADIENT_AVAILABLE); -> ParameterProperties::HYPER | ParameterProperties::GRADIENT);

We would have to do the same with the code in the GPL submodule.
Sorry about the large commit, it just seemed like replacing everything with a regex was the simplest!
I also tested the code locally to make sure it works fine, and passed the tests.

### DIFF
--- a/src/shogun/base/SGObject.h
+++ b/src/shogun/base/SGObject.h
@@ -4,7 +4,8 @@
  * Authors: Heiko Strathmann, Soeren Sonnenburg, Sergey Lisitsyn,
  *          Giovanni De Toni, Jacob Walker, Thoralf Klein, Chiyuan Zhang,
  *          Fernando Iglesias, Sanuj Sharma, Roman Votyakov, Yuyu Zhang,
- *          Viktor Gal, Bjoern Esser, Evangelos Anagnostopoulos, Pan Deng
+ *          Viktor Gal, Bjoern Esser, Evangelos Anagnostopoulos, Pan Deng,
+ *          Gil Hoben
  */
 
 #ifndef __SGOBJECT_H__
@@ -59,49 +60,17 @@ template <class T> class SGStringList;
  * Macros for registering parameters/model selection parameters
  ******************************************************************************/
 
-#ifdef _MSC_VER
-
-#define VA_NARGS(...)  INTERNAL_EXPAND_ARGS_PRIVATE(INTERNAL_ARGS_AUGMENTER(__VA_ARGS__))
-#define INTERNAL_ARGS_AUGMENTER(...) unused, __VA_ARGS__
-#define INTERNAL_EXPAND(x) x
-#define INTERNAL_EXPAND_ARGS_PRIVATE(...) INTERNAL_EXPAND(INTERNAL_GET_ARG_COUNT_PRIVATE(__VA_ARGS__, 5, 4, 3, 2, 1, 0))
-#define INTERNAL_GET_ARG_COUNT_PRIVATE(_0_, _1_, _2_, _3_, _4_, _5_, count, ...) count
-
-#else
-
-#define VA_NARGS_IMPL(_1, _2, _3, _4, _5, N, ...) N
-#define VA_NARGS(...) VA_NARGS_IMPL(__VA_ARGS__, 5, 4, 3, 2, 1)
-
-#endif
-
-#define VARARG_IMPL2(base, count, ...) base##count(__VA_ARGS__)
-#define VARARG_IMPL(base, count, ...) VARARG_IMPL2(base, count, __VA_ARGS__)
-#define VARARG(base, ...) VARARG_IMPL(base, VA_NARGS(__VA_ARGS__), __VA_ARGS__)
-
-#define SG_ADD4(param, name, description, ms_available)                        \
+#define SG_ADD(param, name, description, param_properties)                     \
 	{                                                                          \
+		AnyParameterProperties pprop =                                         \
+		    AnyParameterProperties(description, param_properties);             \
 		this->m_parameters->add(param, name, description);                     \
-		this->watch_param(                                                     \
-		    name, param,                                                       \
-		    AnyParameterProperties(                                            \
-		        description, ms_available, GRADIENT_NOT_AVAILABLE));           \
-		if (ms_available)                                                      \
+		this->watch_param(name, param, pprop);                                 \
+		if (pprop.get_model_selection())                                       \
 			this->m_model_selection_parameters->add(param, name, description); \
-	}
-
-#define SG_ADD5(param, name, description, ms_available, gradient_available)    \
-	{                                                                          \
-		this->m_parameters->add(param, name, description);                     \
-		this->watch_param(                                                     \
-		    name, param, AnyParameterProperties(                               \
-		                     description, ms_available, gradient_available));  \
-		if (ms_available)                                                      \
-			this->m_model_selection_parameters->add(param, name, description); \
-		if (gradient_available)                                                \
+		if (pprop.get_gradient())                                              \
 			this->m_gradient_parameters->add(param, name, description);        \
 	}
-
-#define SG_ADD(...) VARARG(SG_ADD, __VA_ARGS__)
 
 /*******************************************************************************
  * End of macros for registering parameters/model selection parameters
--- a/src/shogun/classifier/LDA.cpp
+++ b/src/shogun/classifier/LDA.cpp
@@ -49,9 +49,9 @@ void CLDA::init()
 
 	SG_ADD(
 	    (machine_int_t*)&m_method, "m_method",
-	    "Method used for LDA calculation", MS_NOT_AVAILABLE);
-	SG_ADD(&m_gamma, "m_gamma", "Regularization parameter", MS_AVAILABLE);
-	SG_ADD(&m_bdc_svd, "m_bdc_svd", "Use BDC-SVD algorithm", MS_NOT_AVAILABLE);
+	    "Method used for LDA calculation", ParameterProperties());
+	SG_ADD(&m_gamma, "m_gamma", "Regularization parameter", ParameterProperties::HYPER);
+	SG_ADD(&m_bdc_svd, "m_bdc_svd", "Use BDC-SVD algorithm", ParameterProperties());
 }
 
 CLDA::~CLDA()
--- a/src/shogun/classifier/PluginEstimate.cpp
+++ b/src/shogun/classifier/PluginEstimate.cpp
@@ -22,17 +22,17 @@ CPluginEstimate::CPluginEstimate(float64_t pos_pseudo, float64_t neg_pseudo)
 {
 	SG_ADD(
 	    &m_pos_pseudo, "pos_pseudo", "pseudo count for positive class",
-	    MS_NOT_AVAILABLE);
+	    ParameterProperties());
 	SG_ADD(
 	    &m_neg_pseudo, "neg_pseudo", "pseudo count for negative class",
-	    MS_NOT_AVAILABLE);
+	    ParameterProperties());
 	SG_ADD(
 	    &pos_model, "pos_model", "LinearHMM modelling positive class.",
-	    MS_NOT_AVAILABLE);
+	    ParameterProperties());
 	SG_ADD(
 	    &neg_model, "neg_model", "LinearHMM modelling negative class.",
-	    MS_NOT_AVAILABLE);
-	SG_ADD(&features, "features", "String Features.", MS_NOT_AVAILABLE);
+	    ParameterProperties());
+	SG_ADD(&features, "features", "String Features.", ParameterProperties());
 }
 
 CPluginEstimate::~CPluginEstimate()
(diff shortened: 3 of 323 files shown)


src/shogun/base/SGObject.h
```diff
@@ -4,7 +4,8 @@
  * Authors: Heiko Strathmann, Soeren Sonnenburg, Sergey Lisitsyn,
  *          Giovanni De Toni, Jacob Walker, Thoralf Klein, Chiyuan Zhang,
  *          Fernando Iglesias, Sanuj Sharma, Roman Votyakov, Yuyu Zhang,
- *          Viktor Gal, Bjoern Esser, Evangelos Anagnostopoulos, Pan Deng
+ *          Viktor Gal, Bjoern Esser, Evangelos Anagnostopoulos, Pan Deng,
+ *          Gil Hoben
```
karlnapf: Weclome to SGObject ;)

src/shogun/classifier/LDA.cpp
```diff
@@ -49,9 +49,9 @@ void CLDA::init()
 
 	SG_ADD(
 	    (machine_int_t*)&m_method, "m_method",
-	    "Method used for LDA calculation", MS_NOT_AVAILABLE);
-	SG_ADD(&m_gamma, "m_gamma", "Regularization parameter", MS_AVAILABLE);
-	SG_ADD(&m_bdc_svd, "m_bdc_svd", "Use BDC-SVD algorithm", MS_NOT_AVAILABLE);
+	    "Method used for LDA calculation", ParameterProperties());
+	SG_ADD(&m_gamma, "m_gamma", "Regularization parameter", ParameterProperties::HYPER);
+	SG_ADD(&m_bdc_svd, "m_bdc_svd", "Use BDC-SVD algorithm", ParameterProperties());
```
karlnapf: Mmmh. This is not nice, it would be better if this default was done automatically.
I didn't realise this before, but I think it means we have to have SG_ADD3 (default properties) and SG_ADD4(user specified properties) as macros cannot do default parameters.

src/shogun/base/SGObject.h
```diff
@@ -59,50 +60,18 @@ template <class T> class SGStringList;
  * Macros for registering parameters/model selection parameters
  ******************************************************************************/
 
-#ifdef _MSC_VER
-
-#define VA_NARGS(...)  INTERNAL_EXPAND_ARGS_PRIVATE(INTERNAL_ARGS_AUGMENTER(__VA_ARGS__))
```
karlnapf: While it would be good to get rid of this thing, I think we still need to keep it (see below on default properties)

src/shogun/classifier/LDA.cpp
```diff
@@ -49,9 +49,9 @@ void CLDA::init()
 
 	SG_ADD(
 	    (machine_int_t*)&m_method, "m_method",
-	    "Method used for LDA calculation", MS_NOT_AVAILABLE);
-	SG_ADD(&m_gamma, "m_gamma", "Regularization parameter", MS_AVAILABLE);
-	SG_ADD(&m_bdc_svd, "m_bdc_svd", "Use BDC-SVD algorithm", MS_NOT_AVAILABLE);
+	    "Method used for LDA calculation", ParameterProperties());
+	SG_ADD(&m_gamma, "m_gamma", "Regularization parameter", ParameterProperties::HYPER);
```
karlnapf: this is quite nice now

src/shogun/classifier/PluginEstimate.cpp
```diff
@@ -22,17 +22,17 @@ CPluginEstimate::CPluginEstimate(float64_t pos_pseudo, float64_t neg_pseudo)
 {
 	SG_ADD(
 	    &m_pos_pseudo, "pos_pseudo", "pseudo count for positive class",
-	    MS_NOT_AVAILABLE);
+	    ParameterProperties());
```
karlnapf: again, would be good if default was taken if just nothing was provided

src/shogun/classifier/mkl/MKL.cpp
```diff
@@ -271,21 +271,21 @@ void CMKL::register_params()
 	rho = 0;
 	lp_initialized = false;
 
-	SG_ADD((CMachine**)&svm, "svm", "wrapper svm", MS_NOT_AVAILABLE);
-	SG_ADD(&C_mkl, "C_mkl", "C mkl", MS_NOT_AVAILABLE);
-	SG_ADD(&mkl_norm, "mkl_norm", "norm used in mkl", MS_NOT_AVAILABLE);
-	SG_ADD(&ent_lambda, "ent_lambda", "elastic net sparsity trade-off parameter", MS_NOT_AVAILABLE);
-	SG_ADD(&mkl_block_norm, "mkl_block_norm", "mkl sparse trade-off parameter", MS_NOT_AVAILABLE);
+	SG_ADD((CMachine**)&svm, "svm", "wrapper svm", ParameterProperties());
+	SG_ADD(&C_mkl, "C_mkl", "C mkl", ParameterProperties());
```
karlnapf: this is HYPER

src/shogun/classifier/mkl/MKL.cpp
```diff
@@ -271,21 +271,21 @@ void CMKL::register_params()
...
+	SG_ADD(&mkl_norm, "mkl_norm", "norm used in mkl", ParameterProperties());
+	SG_ADD(&ent_lambda, "ent_lambda", "elastic net sparsity trade-off parameter", ParameterProperties());
```
karlnapf: hyper

src/shogun/classifier/mkl/MKL.cpp
```diff
@@ -271,21 +271,21 @@ void CMKL::register_params()
...
+	SG_ADD(&mkl_block_norm, "mkl_block_norm", "mkl sparse trade-off parameter", ParameterProperties());
```
karlnapf: hyper

src/shogun/machine/LinearMachine.cpp
```diff
@@ -35,11 +35,11 @@ void CLinearMachine::init()
 	bias = 0;
 	features = NULL;
 
-	SG_ADD(&m_w, "w", "Parameter vector w.", MS_NOT_AVAILABLE);
-	SG_ADD(&bias, "bias", "Bias b.", MS_NOT_AVAILABLE);
+	SG_ADD(&m_w, "w", "Parameter vector w.", ParameterProperties());
```
karlnapf: things like this would be MODEL

src/shogun/machine/LinearStructuredOutputMachine.cpp
```diff
@@ -68,7 +68,7 @@ CStructuredLabels* CLinearStructuredOutputMachine::apply_structured(CFeatures* d
 
 void CLinearStructuredOutputMachine::register_parameters()
 {
-	SG_ADD(&m_w, "m_w", "Weight vector", MS_NOT_AVAILABLE);
+	SG_ADD(&m_w, "m_w", "Weight vector", ParameterProperties());
```
karlnapf: model

src/shogun/machine/OnlineLinearMachine.cpp
```diff
@@ -19,10 +19,10 @@ using namespace shogun;
 COnlineLinearMachine::COnlineLinearMachine()
 : CMachine(), bias(0), features(NULL)
 {
-	SG_ADD(&m_w, "m_w", "Parameter vector w.", MS_NOT_AVAILABLE);
-	SG_ADD(&bias, "bias", "Bias b.", MS_NOT_AVAILABLE);
+	SG_ADD(&m_w, "m_w", "Parameter vector w.", ParameterProperties());
```
karlnapf: model, you get the hang :)

src/shogun/machine/gp/GaussianLikelihood.cpp
```diff
@@ -52,7 +52,7 @@ CGaussianLikelihood::CGaussianLikelihood(float64_t sigma) : CLikelihoodModel()
 void CGaussianLikelihood::init()
 {
 	m_log_sigma=0.0;
-	SG_ADD(&m_log_sigma, "log_sigma", "Observation noise in log domain", MS_AVAILABLE, GRADIENT_AVAILABLE);
+	SG_ADD(&m_log_sigma, "log_sigma", "Observation noise in log domain", ParameterProperties::HYPER | ParameterProperties::GRADIENT);
```
karlnapf: nice!

src/shogun/base/SGObject.h
gf712: Haha yes, I thought I would discretely add myself to the list :D

src/shogun/base/SGObject.h
gf712: OK, I will bring this back and add a SG_ADD3!

src/shogun/machine/LinearMachine.cpp
gf712: Both the weight and bias right?

karlnapf: yes. We dont need to fix all the wrongly added parameter in this PR though, just put a few examples in

src/shogun/base/SGObject.h
```diff
@@ -57,10 +57,36 @@ template <class T> class SGStringList;
...
+
+#define VARARG_IMPL2(base, count, ...) base##count(__VA_ARGS__)
+#define VARARG_IMPL(base, count, ...) VARARG_IMPL2(base, count, __VA_ARGS__)
+#define VARARG(base, ...) VARARG_IMPL(base, VA_NARGS(__VA_ARGS__), __VA_ARGS__)
+
+#define SG_ADD3(param, name, description)    \
+	{                                                                          \
+		this->m_parameters->add(param, name, description);                     \
+		this->watch_param(                                                     \
+		    name, param, AnyParameterProperties()); 						   \
```
karlnapf: yes this is good

src/shogun/classifier/LDA.cpp
```diff
@@ -49,9 +49,9 @@ void CLDA::init()
 
 	SG_ADD(
 	    (machine_int_t*)&m_method, "m_method",
-	    "Method used for LDA calculation", ParameterProperties());
+	    "Method used for LDA calculation");
```
karlnapf: yep, much cleaner

src/shogun/converter/LocallyLinearEmbedding.cpp
```diff
@@ -26,11 +26,11 @@ CLocallyLinearEmbedding::CLocallyLinearEmbedding() :
 
 void CLocallyLinearEmbedding::init()
 {
-	SG_ADD(&m_k, "k", "number of neighbors", MS_AVAILABLE);
+	SG_ADD(&m_k, "k", "number of neighbors", ParameterProperties::HYPER);
 	SG_ADD(&m_nullspace_shift, "nullspace_shift",
-      "nullspace finding regularization shift",MS_NOT_AVAILABLE);
+      "nullspace finding regularization shift",ParameterProperties());
```
karlnapf: missed one here :)

src/shogun/converter/ManifoldSculpting.cpp
```diff
@@ -23,11 +23,11 @@ CManifoldSculpting::CManifoldSculpting() :
 
 void CManifoldSculpting::init()
 {
-	SG_ADD(&m_k, "k", "number of neighbors", MS_NOT_AVAILABLE);
+	SG_ADD(&m_k, "k", "number of neighbors");
 	SG_ADD(&m_squishing_rate, "quishing_rate",
-      "squishing rate",MS_NOT_AVAILABLE);
+      "squishing rate",ParameterProperties());
```
karlnapf: and here

karlnapf:
BTW this passed all of our CI
https://dev.azure.com/shogunml/shogun/_build/results?buildId=91

Try to keep an eye on the output of this after any patch is merged (until we have the pre-merge builds back)

gf712:
Yes, I saw that. I think it's because of the multiple constructors, which also have defaults, led to something we didn't expect. So might be worth removing the constructors which accept the old enums since they shouldn't be required anymore?

karlnapf:
Sorry what do you mean? :)

gf712:
We still have these in the code:

```
/** model selection availability */
enum EModelSelectionAvailability
{
	MS_NOT_AVAILABLE = 0,
	MS_AVAILABLE = 1,
};

/** gradient availability */
enum EGradientAvailability
{
	GRADIENT_NOT_AVAILABLE = 0,
	GRADIENT_AVAILABLE = 1
};
```

Might be worth removing this with an update of the hash for the latest commit on the GPL repo?

I am not sure we are talking about the same thing though. I thought you meant you expected some tests to have failed because the GPL repo hasn't been updated yet on the shogun repo? Which they should have, because GPL still uses the old SG_ADD, right?

karlnapf:
ah I see now.

GPL should compile, could you check/make sure?

removing these might cause a lot of compile errors, we should eventually do that (as I said, the modelselection parameters should be removed as well), but maybe for now let's focus on one thing?

gf712:
GPL compiles fine! And the tests pass.

PR #4421, opened 2018-11-22
AnyParameterProperties tests

gf712:
Added tests to check binary logic of AnyParameterProperties.m_attribute_mask.


tests/unit/lib/Any_unittest.cc
```diff
@@ -691,3 +691,65 @@ TEST(Any, lazy_cloneable_visitable)
...
+	    params.get_gradient(), EGradientAvailability::GRADIENT_NOT_AVAILABLE);
+	EXPECT_FALSE(params.get_model());
+}
+
+TEST(AnyParameterProperties, new_api_default)
+{
+	AnyParameterProperties params = AnyParameterProperties();
+
+	EXPECT_TRUE(params.equals(ParameterProperties()));
+	EXPECT_FALSE(params.has_property(ParameterProperties::MODEL));
```
karlnapf: as said before, can we not explicitly test that those are not set, but in fact that *none* are set. Otherwise we will have to modify tests when we add new attributes

tests/unit/lib/Any_unittest.cc
```diff
@@ -691,3 +691,65 @@ TEST(Any, lazy_cloneable_visitable)
...
+	    params.get_model_selection(),
+	    EModelSelectionAvailability::MS_NOT_AVAILABLE);
+	EXPECT_EQ(params.get_gradient(), EGradientAvailability::GRADIENT_AVAILABLE);
+	EXPECT_TRUE(params.get_model());
+
+	EXPECT_TRUE(params.has_property(ParameterProperties::MODEL));
+	EXPECT_TRUE(params.has_property(ParameterProperties::GRADIENT));
+	EXPECT_TRUE(params.equals(
+	    ParameterProperties::MODEL | ParameterProperties::GRADIENT));
+    EXPECT_FALSE(params.equals(ParameterProperties::MODEL));
```
karlnapf: everything else should be false. And that should be tested. Not just these two, same reasoning as above

tests/unit/lib/Any_unittest.cc
```diff
@@ -691,3 +691,65 @@ TEST(Any, lazy_cloneable_visitable)
...
+	    params.get_model_selection(),
+	    EModelSelectionAvailability::MS_NOT_AVAILABLE);
+	EXPECT_EQ(params.get_gradient(), EGradientAvailability::GRADIENT_AVAILABLE);
+	EXPECT_TRUE(params.get_model());
+
+	EXPECT_TRUE(params.has_property(ParameterProperties::MODEL));
+	EXPECT_TRUE(params.has_property(ParameterProperties::GRADIENT));
+	EXPECT_TRUE(params.equals(
+	    ParameterProperties::MODEL | ParameterProperties::GRADIENT));
+	EXPECT_FALSE(params.has_property(ParameterProperties::HYPER));
```
karlnapf: same here

tests/unit/lib/Any_unittest.cc
karlnapf: minor indentation glitch

src/shogun/lib/bitmask_operators.h
```diff
@@ -106,5 +108,11 @@ namespace shogun {
                 static_cast<underlying>(lhs) ^ static_cast<underlying>(rhs));
         return lhs;
     }
+    template<typename E>
```
karlnapf: ++

tests/unit/lib/Any_unittest.cc
gf712: I did think I saw that comment in the previous PR, but the next day the comment disappeared..

src/shogun/base/AnyParameter.h
```diff
@@ -118,6 +118,15 @@ namespace shogun
 			return static_cast<bool>(
 			    m_attribute_mask & ParameterProperties::MODEL);
 		}
+		bool has_property(ParameterProperties other_params) const
+		{
+			return static_cast<bool>(m_attribute_mask & other_params);
+		}
+
+		bool equals(ParameterProperties other_params) const
```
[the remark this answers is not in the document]
karlnapf: first one is definitely useful, so let's have that
Second one I dont know whether we need this? But if we do, then all fields need to be compared of course.

tests/unit/lib/Any_unittest.cc
karlnapf: Yeah that happens on force push. I keep the notification emails to not loose comments when github hides them.

src/shogun/base/AnyParameter.h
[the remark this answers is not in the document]
gf712: I think the second one is a bit of overkill and I cant think of a situation where it would be useful.. I'll just stick to the first one!

tests/unit/lib/Any_unittest.cc
```diff
@@ -708,48 +708,48 @@ TEST(AnyParameterProperties, new_api_default)
...
-	EXPECT_TRUE(params.get_model());
+	EXPECT_EQ(
+	    params.get_gradient(), EGradientAvailability::GRADIENT_NOT_AVAILABLE);
+	EXPECT_FALSE(params.get_model());
 
-	EXPECT_TRUE(params.has_property(ParameterProperties::MODEL));
-	EXPECT_TRUE(params.has_property(ParameterProperties::GRADIENT));
-	EXPECT_TRUE(params.equals(
-	    ParameterProperties::MODEL | ParameterProperties::GRADIENT));
+	EXPECT_FALSE(params.compare_mask(
```
karlnapf: nope, this just checks that *one* is not set.
you want to check that *none* are set. So you need to create an "empty" mask (all zeros) and then call compare_mask

karlnapf: but this empty mask should be created via *not* explicitly listing the available types.
Again, this is to avoid having to touch all these tests in the case when a type is added

gf712: OK! And the `EXPECT_FALSE` after, are they not necessary then? They should be covered by the empty mask.

gf712: But `EXPECT_TRUE(params.compare_mask(ParameterProperties::EMPTY))` is the same as `EXPECT_TRUE(params.compare_mask(ParameterProperties()))`. I can add a EMPTY though to the enum.

karlnapf: you can, and of course it is the same. But then it wouldnt test that the default properties are indeed empty, so the test would not be of an use: It would just test that if you create two properties with the default ctor, they have the same parameters. If you would check for `EMPTY` (or maybe we call it `NONE` (and define it to be `0x0`), then you also assert that the default ctor generates empty parameters, which is exactly what we want here. If people then later start messing around with the ctor (e.g. change the default), the test will fail, so it provides some protection against devs going wild

gf712: Ah ok! But then in that case it might make more sense to instantiate the `m_attribute_mask` class member in the constructors with `ParameterProperties::EMPTY` rather than `ParameterProperties()` no? I just used `ParameterProperties()` because it was the easiest way  to represent zero without any casting.

tests/unit/lib/Any_unittest.cc
```diff
@@ -691,3 +691,61 @@ TEST(Any, lazy_cloneable_visitable)
...
+	EXPECT_EQ(params.get_description(), "test");
+	EXPECT_EQ(
+	    params.get_model_selection(),
+	    EModelSelectionAvailability::MS_NOT_AVAILABLE);
+	EXPECT_EQ(
+	    params.get_gradient(), EGradientAvailability::GRADIENT_NOT_AVAILABLE);
+	EXPECT_FALSE(params.get_model());
+
+	EXPECT_TRUE(params.compare_mask(ParameterProperties::NONE));
+	EXPECT_FALSE(params.has_property(ParameterProperties::HYPER));
```
karlnapf: the last three are nop now

tests/unit/lib/Any_unittest.cc
karlnapf: these last 3 as well

gf712:
Do you think this is ready to be merged? Started working on the new getters in CSGObject using the enums as filters, and was wondering what new methods from this PR I could use.

karlnapf:
Thanks!

PR #4422, opened 2018-11-23
Update variant to fix MSVC build

vinx13:


vinx13:
@karlnapf appveyor config is removed in ed31a10

karlnapf:
Ah ok so @vigsterkr already moved on.
This will be visible in the new MS build pipeline.
@vigsterkr shall we merge this ?

PR #4423, opened 2018-11-23
Add meta example features-char-string

avramidis:
A simple meta example for CStringFeatures<char>. 

I would like to make changes and add output file for an integration test but I am not sure if the current outputs are enough for that. Currently, it stores "max_string_length", "number_of_strings" and "length_of_first_string". I don't think that is possible to practically check all the values of "strings". 

However, if you don't have a better idea I could add eight variables that store the value of the first vector before and after the change to "test".

### DIFF
--- a/examples/meta/generator/targets/cpp.json
+++ b/examples/meta/generator/targets/cpp.json
@@ -45,7 +45,8 @@
     "Type": {
         "RealFeatures": "DenseFeatures<float64_t>",
         "RealSubsetFeatures": "DenseSubsetFeatures<float64_t>",
-        "StringCharFeatures": "CStringFeatures<char>",
+        "StringCharFeatures": "StringFeatures<char>",
+        "StringCharList": "SGStringList<char>",
         "Default": "$typeName",
         "bool": "bool",
         "string": "char*",
@@ -96,7 +97,11 @@
             "get_int_vector": "$object->get<SGVector<int32_t>>($arguments)",
             "get_real": "$object->get<float64_t>($arguments)",
             "get_real_vector": "$object->get<SGVector<float64_t>>($arguments)",
-            "get_real_matrix": "$object->get<SGMatrix<float64_t>>($arguments)"
+            "get_real_matrix": "$object->get<SGMatrix<float64_t>>($arguments)",
+            "get_int_string_list": "$object->get<SGStringList<int32_t>>($arguments)",
+            "get_bool_string_list": "$object->get<SGStringList<bool>>($arguments)",
+            "get_char_string_list": "$object->get<SGStringList<char>>($arguments)",
+            "get_real_string_list": "$object->get<SGStringList<float64_t>>($arguments)"
         },
         "StaticCall": "C$typeName::$method($arguments)",
         "GlobalCall": "$method($arguments)",
--- a/examples/meta/generator/translate.py
+++ b/examples/meta/generator/translate.py
@@ -341,6 +341,10 @@ class Translator:
         if self.targetDict["Dependencies"].get("IncludeGlobalFunctions"):
             dependencies = dependencies.union(globalFunctions)
 
+        if "ExcludeImport" in self.targetDict["Dependencies"]:
+            for item in self.targetDict["Dependencies"].get("ExcludeImport"):
+                dependencies.discard(item)
+
         dependencies = list(dependencies)
 
         translations = list(set(map(self.translateDependencyElement, dependencies)))
--- a/examples/meta/src/features/string_char.sg
+++ b/examples/meta/src/features/string_char.sg
@@ -1,0 +1,11 @@
+File words = csv_file("../../data/words.dat")
+
+#![create_features]
+Features f = string_features(words, enum EAlphabet.RAWBYTE)
+#![create_features]
+
+#![output stat]
+int max_string_length = f.get_int("max_string_length")
+int number_of_strings = f.get_int("num_vectors")
+StringCharList string_list = f.get_char_string_list("string_list")
+#![output stat]
--- a/src/interfaces/java/swig_typemaps.i
+++ b/src/interfaces/java/swig_typemaps.i
@@ -693,11 +693,7 @@ TYPEMAP_SGMATRIX(float64_t, double, Double, jdouble, "toDoubleArray", "()[[D", "
 		JCALL4(Set##JAVATYPE##ArrayRegion, jenv, jarr, 0, str[i].slen, arr);
 		JCALL3(SetObjectArrayElement, jenv, res, i, jarr);
 		JCALL1(DeleteLocalRef, jenv, jarr);
-
-		SG_FREE(str[i].string);
-		SG_FREE(arr);
 	}
-	SG_FREE(str);
 	$result = res;
 }
 
@@ -718,7 +714,7 @@ TYPEMAP_STRINGFEATURES(int64_t, int, Int, jint, "Int[][]", "[[I")
 TYPEMAP_STRINGFEATURES(uint64_t, long, Long, jlong, "Long[][]", "[[J")
 TYPEMAP_STRINGFEATURES(long long, long, Long, jlong, "Long[][]", "[[J")
 TYPEMAP_STRINGFEATURES(float32_t, float, Float, jfloat, "Float[][]", "[[F")
-TYPEMAP_STRINGFEATURES(float64_t, double, Double, jdouble, "Doulbe[][]", "[[D")
+TYPEMAP_STRINGFEATURES(float64_t, double, Double, jdouble, "Double[][]", "[[D")
 
 #undef TYPEMAP_STRINGFEATURES
 
@@ -777,9 +773,7 @@ TYPEMAP_STRINGFEATURES(float64_t, double, Double, jdouble, "Doulbe[][]", "[[D")
 		jstring jstr = (jstring)JCALL1(NewStringUTF, jenv, (char *)str[i].string);
 		JCALL3(SetObjectArrayElement, jenv, res, i, jstr);
 		JCALL1(DeleteLocalRef, jenv, jstr);
-		SG_FREE(str[i].string);
 	}
-	SG_FREE(str);
 	$result = res;
 }
 
--- a/src/interfaces/octave/swig_typemaps.i
+++ b/src/interfaces/octave/swig_typemaps.i
@@ -368,21 +368,44 @@ TYPEMAP_STRINGFEATURES_IN(is_matrix_type() && arg.is_uint16_type, uint16NDArray,
 
 /* output typemap for CStringFeatures */
 %define TYPEMAP_STRINGFEATURES_OUT(type,typecode)
+%typemap(out) shogun::SGStringList<char>
+{
+    shogun::SGString<char>* str = $1.strings;
+    int32_t i, num_strings = $1.num_strings;
+
+    Cell c(num_strings, 1);
+
+    for (i = 0; i < num_strings; i++) {
+        c(i)=std::string(str[i].string);
+    }
+
+    $result = c;
+}
+
 %typemap(out) shogun::SGStringList<type>
 {
-    /* TODO STRING OUT TYPEMAPS */
+    shogun::SGString<type>* str = $1.strings;
+    int32_t i, num_strings = $1.num_strings;
+
+    ColumnVector c(dim_vector(num_strings, 1));
+
+    for (i = 0; i < num_strings; i++) {
+        c(i)=*str[i].string;
+    }
+
+    $result = c;
 }
 %enddef
 
-TYPEMAP_STRINGFEATURES_OUT(char,          charMatrix)
-TYPEMAP_STRINGFEATURES_OUT(uint8_t,       uint8NDArray)
-TYPEMAP_STRINGFEATURES_OUT(int16_t,       int16NDArray)
-TYPEMAP_STRINGFEATURES_OUT(uint16_t,      uint16NDArray)
-TYPEMAP_STRINGFEATURES_OUT(int32_t,       int32NDArray)
-TYPEMAP_STRINGFEATURES_OUT(uint32_t,      uint32NDArray)
-TYPEMAP_STRINGFEATURES_OUT(int64_t,       int64NDArray)
-TYPEMAP_STRINGFEATURES_OUT(uint64_t,      uint64NDArray)
-TYPEMAP_STRINGFEATURES_OUT(float64_t,     Matrix)
+TYPEMAP_STRINGFEATURES_OUT(char,          Cell)
+TYPEMAP_STRINGFEATURES_OUT(uint8_t,       ColumnVector)
+TYPEMAP_STRINGFEATURES_OUT(int16_t,       ColumnVector)
+TYPEMAP_STRINGFEATURES_OUT(uint16_t,      ColumnVector)
+TYPEMAP_STRINGFEATURES_OUT(int32_t,       ColumnVector)
+TYPEMAP_STRINGFEATURES_OUT(uint32_t,      ColumnVector)
+TYPEMAP_STRINGFEATURES_OUT(int64_t,       ColumnVector)
+TYPEMAP_STRINGFEATURES_OUT(uint64_t,      ColumnVector)
+TYPEMAP_STRINGFEATURES_OUT(float64_t,     ColumnVector)
 
 #undef TYPEMAP_STRINGFEATURES_OUT
 
--- a/src/interfaces/swig/shogun.i
+++ b/src/interfaces/swig/shogun.i
@@ -232,6 +232,9 @@ namespace shogun
 %template(get_real) CSGObject::get<float64_t, void>;
 %template(get_int) CSGObject::get<int32_t, void>;
 %template(get_real_matrix) CSGObject::get<SGMatrix<float64_t>, void>;
+%template(get_bool_string_list) CSGObject::get<SGStringList<bool>, void>;
+%template(get_char_string_list) CSGObject::get<SGStringList<char>, void>;
+%template(get_int_string_list) CSGObject::get<SGStringList<int32_t>, void>;
 
 #ifndef SWIGJAVA
 %template(get_real_vector) CSGObject::get<SGVector<float64_t>, void>;
--- a/src/shogun/features/StringFeatures.cpp
+++ b/src/shogun/features/StringFeatures.cpp
@@ -180,7 +180,7 @@ template<class ST> EFeatureClass CStringFeatures<ST>::get_feature_class() const 
 
 template<class ST> EFeatureType CStringFeatures<ST>::get_feature_type() const { return F_UNKNOWN; }
 
-template<class ST> CAlphabet* CStringFeatures<ST>::get_alphabet()
+template<class ST> CAlphabet* CStringFeatures<ST>::get_alphabet() const
 {
 	SG_REF(alphabet);
 	return alphabet;
@@ -983,7 +983,7 @@ template<class ST> bool CStringFeatures<ST>::append_features(SGString<ST>* p_fea
 	return false;
 }
 
-template<class ST> SGStringList<ST> CStringFeatures<ST>::get_features()
+template<class ST> SGStringList<ST> CStringFeatures<ST>::get_string_list() const
 {
 	SGStringList<ST> sl(NULL,0,0,false);
 
@@ -991,7 +991,7 @@ template<class ST> SGStringList<ST> CStringFeatures<ST>::get_features()
 	return sl;
 }
 
-template<class ST> SGString<ST>* CStringFeatures<ST>::get_features(int32_t& num_str, int32_t& max_str_len)
+template<class ST> SGString<ST>* CStringFeatures<ST>::get_features(int32_t& num_str, int32_t& max_str_len) const
 {
 	if (m_subset_stack->has_subsets())
 		SG_ERROR("get features() is not possible on subset")
@@ -1695,6 +1695,8 @@ template<class ST> void CStringFeatures<ST>::init()
 
 	m_parameters->add_vector(&symbol_mask_table, &symbol_mask_table_len, "mask_table", "Symbol mask table - using in higher order mapping");
 	watch_param("mask_table", &symbol_mask_table, &symbol_mask_table_len);
+	watch_method("num_vectors", &CStringFeatures::get_num_vectors);
+	watch_method("string_list", &CStringFeatures::get_string_list);
 }
 
 /** get feature type the char feature can deal with
--- a/src/shogun/features/StringFeatures.h
+++ b/src/shogun/features/StringFeatures.h
@@ -151,7 +151,7 @@ template <class ST> class CStringFeatures : public CFeatures
 		 *
 		 * @return alphabet
 		 */
-		CAlphabet* get_alphabet();
+		CAlphabet* get_alphabet() const;
 
 		/** duplicate feature object
 		 *
@@ -419,10 +419,10 @@ template <class ST> class CStringFeatures : public CFeatures
 		bool append_features(SGString<ST>* p_features, int32_t p_num_vectors,
 				int32_t p_max_string_length);
 
-		/** get_features
-		 * @return features
+		/** get_string_list
+		 * @return string_list
 		 */
-        SGStringList<ST> get_features();
+        SGStringList<ST> get_string_list() const;
 
 		/** get_features
 		 *
@@ -432,7 +432,7 @@ template <class ST> class CStringFeatures : public CFeatures
 		 * @param max_str_len maximal string length (returned)
 		 * @return string features
 		 */
-		virtual SGString<ST>* get_features(int32_t& num_str, int32_t& max_str_len);
+		virtual SGString<ST>* get_features(int32_t& num_str, int32_t& max_str_len) const;
 
 		/** copy_features
 		 *
--- a/src/shogun/lib/SGStringList.cpp
+++ b/src/shogun/lib/SGStringList.cpp
@@ -101,6 +101,24 @@ SGStringList<T> SGStringList<T>::clone() const
 	return SGStringList<T>(strings, num_strings, max_string_length);
 }
 
+template <class T>
+bool SGStringList<T>::equals(const SGStringList<T>& other) const
+{
+	if (this->num_strings!=other.num_strings)
+		return false;
+
+	if (this->max_string_length!=other.max_string_length)
+		return false;
+
+	for (auto i : range(num_strings))
+	{
+		if (!this->strings[i].equals(other.strings[i]))
+			return false;
+	}
+
+	return true;
+}
+
 template class SGStringList<bool>;
 template class SGStringList<char>;
 template class SGStringList<int8_t>;
--- a/src/shogun/lib/SGStringList.h
+++ b/src/shogun/lib/SGStringList.h
@@ -66,6 +66,14 @@ public:
 	 */
 	SGStringList<T> clone() const;
 
+
+	/** Equals method
+	 * @param other SGStringList to compare with
+	 * @return false iff the number of strings, the maximum string length or
+	 * any of the string items are different, true otherwise
+	 */
+	bool equals(const SGStringList<T>& other) const;
+
 protected:
 
 	/** copy data */
--- a/src/shogun/util/factory.h
+++ b/src/shogun/util/factory.h
@@ -120,8 +120,8 @@ namespace shogun
 	}
 
 	CFeatures* string_features(
-	    CFile* file, EAlphabet alpha = DNA,
-	    EPrimitiveType primitive_type = PT_CHAR)
+	    CFile* file, machine_int_t alphabet_type = DNA,
+	    machine_int_t primitive_type = PT_CHAR)
 	{
 		REQUIRE(file, "No file provided.\n");
 		CFeatures* result = nullptr;
@@ -129,7 +129,7 @@ namespace shogun
 		switch (primitive_type)
 		{
 		case PT_CHAR:
-			result = new CStringFeatures<char>(file, alpha);
+			result = new CStringFeatures<char>(file, static_cast<EAlphabet>(alphabet_type));
 			break;
 		default:
 			SG_SNOTIMPLEMENTED
--- a/tests/unit/features/StringFeatures_unittest.cc
+++ b/tests/unit/features/StringFeatures_unittest.cc
@@ -78,28 +78,13 @@ TEST(StringFeaturesTest,copy_subset)
 	SG_UNREF(subset_copy);
 }
 
-TEST(StringFeaturesTest,clone)
+TEST(StringFeaturesTest,equals)
 {
 	SGStringList<char> strings = generateRandomStringData();
 
 	CStringFeatures<char>* f=new CStringFeatures<char>(strings, ALPHANUM);
-	CStringFeatures<char>* f_clone = (CStringFeatures<char> *) f->clone();
-
-	for (index_t i=0; i<f->get_num_vectors(); ++i)
-	{
-		index_t a_len;
-		index_t b_len;
-		bool a_free;
-		bool b_free;
-
-		char * a_vec = f->get_feature_vector(i, a_len, a_free);
-		char * b_vec = f_clone->get_feature_vector(i, b_len, b_free);
-
-		EXPECT_EQ(a_len, b_len);
-		EXPECT_EQ(a_free, b_free);
-		for (index_t j = 0; j < a_len; ++j)
-			EXPECT_EQ(a_vec[j], b_vec[j]);
-	}
+	CStringFeatures<char>* f_clone = (CStringFeatures<char>*)f->clone();
+	EXPECT_EQ(f->equals(f_clone), true);
 
 	SG_UNREF(f);
 	SG_UNREF(f_clone);
--- a/doc/readme/INTERFACES.md
+++ b/doc/readme/INTERFACES.md
@@ -54,7 +54,7 @@ This needs `shogun.oct` to be visible, which is either in `path/to/build/src/int
 
 Running an example:
 
-    python path/to/octave_example.py
+    octave path/to/octave_example.m
 
 ### Ruby
 This needs `shogun.rb` to be visible, which is either in `path/to/build/src/interfaces/ruby_modular/` or in something similar to `path/to/shogun-install/lib/x86_64-linux-gnu/site_ruby`
--- a/examples/meta/generator/parse.py
+++ b/examples/meta/generator/parse.py
@@ -104,7 +104,8 @@ class FastParser:
         'ShortRealMatrix': 'SHOGUNSGTYPE',
         'RealMatrix': 'SHOGUNSGTYPE',
         'LongRealMatrix': 'SHOGUNSGTYPE',
-        'ComplexMatrix': 'SHOGUNSGTYPE'
+        'ComplexMatrix': 'SHOGUNSGTYPE',
+        'StringCharList':'SHOGUNSGTYPE'
     }
 
     t_INTLITERAL = "-?[0-9]+"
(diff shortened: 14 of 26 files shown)


examples/meta/generator/targets/cpp.json
```diff
@@ -45,7 +45,7 @@
     "Type": {
         "RealFeatures": "DenseFeatures<float64_t>",
         "RealSubsetFeatures": "DenseSubsetFeatures<float64_t>",
-        "StringCharFeatures": "CStringFeatures<char>",
+        "StringCharFeatures": "StringFeatures<char>",
```
karlnapf: good catch!

examples/meta/generator/targets/cpp.json
```diff
@@ -101,7 +101,7 @@
         "StaticCall": "C$typeName::$method($arguments)",
         "GlobalCall": "$method($arguments)",
         "Identifier": "$identifier",
-        "Enum":"(machine_int_t)$typeName::$value"
```
karlnapf: and this does not break other examples? I cannot remember why exactly this case was here, but I remember putting in there for a reason...
`ctest -R generated_cpp` should bring clarity

src/shogun/features/StringFeatures.h
```diff
@@ -20,6 +20,7 @@
 #include <shogun/features/Features.h>
 #include <shogun/features/Alphabet.h>
 #include <shogun/lib/SGString.h>
+#include <shogun/lib/SGStringList.h>
```
karlnapf: was there a compile error before ?

karlnapf:
You are right we dont really need any output here, as nothing really happens. THese outputs are for making sure that algorithms dont all of a sudden output different results.

src/shogun/features/StringFeatures.h
avramidis: There was in my case. I use gcc version 7.3.0.

examples/meta/generator/targets/cpp.json
avramidis: I believe you put that to make meta_api-enums work. Removing machine_int_t will cause the example to fail. However, to make the example I added to work you have to remove machine_int_t. I was hoping to see the CI results together and discuss how to proceed.

I prefer to not have to convert the enum types.

This is the output I get:

ctest -R generated_cpp-meta_api-enums -V
UpdateCTestConfiguration  from :/home/elavram/Workfolder/shogun/build/DartConfiguration.tcl
UpdateCTestConfiguration  from :/home/elavram/Workfolder/shogun/build/DartConfiguration.tcl
Test project /home/elavram/Workfolder/shogun/build
Constructing a list of tests
Done constructing a list of tests
Updating test list for fixtures
Added 0 tests to meet fixture requirements
Checking test dependency graph...
Checking test dependency graph end
test 47
    Start 47: generated_cpp-meta_api-enums

47: Test command: /home/elavram/Workfolder/shogun/build/examples/meta/cpp/meta_api/cpp-meta_api-enums
47: Test timeout computed to be: 9.99988e+06
47: [ERROR] In file /home/elavram/Workfolder/shogun/src/shogun/base/SGObject.h line 366: Cannot put parameter LibLinear::liblinear_solver_type of type int, incompatible provided type shogun::LIBLINEAR_SOLVER_TYPE.
47: terminate called after throwing an instance of 'shogun::ShogunException'
47:   what():  [ERROR] In file /home/elavram/Workfolder/shogun/src/shogun/base/SGObject.h line 366: Cannot put parameter LibLinear::liblinear_solver_type of type int, incompatible provided type shogun::LIBLINEAR_SOLVER_TYPE.
47: 
1/1 Test #47: generated_cpp-meta_api-enums .....***Exception: Child aborted  0.16 sec

0% tests passed, 1 tests failed out of 1

Total Test time (real) =   0.18 sec

The following tests FAILED:
	 47 - generated_cpp-meta_api-enums (Child aborted)
Errors while running CTest

avramidis:
Apologies for the late reply.

examples/meta/generator/targets/cpp.json
avramidis: I think a solution to consider is to convert the all "enum" to "enum class" so there will be no automatic convert to int and it will be safer.

avramidis:
Yes, that is the plan. I have started making changes but they will be ready mid next week if that is ok with you.

karlnapf:
Sure of course.

avramidis:
An update for this one - I am working on refactoring the watch_method function to be able to register functions that take arguments. The  features-char-string example needs that. I think made some progress but I need to work a bit more.

@gf712 I believe you will encounter this problem when you start converting undocumented python examples.

karlnapf:
@avramidis we cannot really accept arguments in the watched functions. At least not with the current design, and this was a quite conscious decision. The way around this is to communicate with the classes via put get (to set parameters) and then call helper methods.
If that doesnt help, we might want to think about putting methods into the base interface.

Why does `max_string_length` need parameters?

avramidis:
OK, thanks for the explanation. The max_string_length doesn't need it but the set_feature_vector

https://github.com/shogun-toolbox/shogun/blob/f8a120b25ccd32bdefb47a2e9170f04647050fd8/src/shogun/features/StringFeatures.cpp#L248

karlnapf:
Actually that is a method that we do not want to expose to SWIG, so rather just hide it.
We want feature objects to be immutable eventually, so changing is not allowed, rather construct a new one.

But in cases where we want things like this, it would be best added to the base class.

examples/meta/src/features/string_char.sg
```diff
@@ -0,0 +1,10 @@
+File words = csv_file("../../data/words.dat")
+
+#![create_features]
+Features f = string_features(words, enum EAlphabet.RAWBYTE)
+#![create_features]
+
```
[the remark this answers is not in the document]
avramidis: I am stuck. I am trying to make get_features work. However, it is overloaded and it is used in many undocumented python examples. 

https://github.com/shogun-toolbox/shogun/blob/1c9ac30237f8b9ba5b2f9fe619fa9fabd7ee361f/src/shogun/features/StringFeatures.cpp#L1022

https://github.com/shogun-toolbox/shogun/blob/1c9ac30237f8b9ba5b2f9fe619fa9fabd7ee361f/src/shogun/features/StringFeatures.cpp#L1030

https://github.com/shogun-toolbox/shogun/blob/1c9ac30237f8b9ba5b2f9fe619fa9fabd7ee361f/src/shogun/features/StringFeatures.cpp#L1062

I don't want to put them in the base interface because they are specific to string features.

You said earlier  
> The way around this is to communicate with the classes via put get (to set parameters) and then call helper methods.

put and get for watched parameters is fine but what helper methods you mean?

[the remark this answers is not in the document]
karlnapf: I just spend quite some time trying to register `get_features` with `watch_param`. Tricky one!

You will first need to make the method `const` (both of them as one calls the other). The compiler error you see when trying to register it comes from const, not from overloading (although it looks like that).

Then you will hit another compiler error, which complains about `bool shogun::any_detail::compare` not matching with the call (among others). This comes from the fact that anything that is ever any'ed in Shogun needs to be comparable and cloneable for `SGObject::clone/equals`. This hasn't yet been done for `SGStringList`. You will have to add `equals` and `clone` to `SGStringList` (see SGMatrix/SGVector/ for inspiration). Then those compiler errors should vanish. I don't know what happens then, but it will bring you one step closer to making the `get` work for the list of string features.

avramidis:
Nice, thanks for that. I have a branch for trying to watch methods with
arguments. I was working on that in the weekend. We definitely don't need
something like that, right?

karlnapf:
I think that would be quite hard as then we would need multiple `get` methods with different number of arguments as well or?

karlnapf:
I also think our new API design doesn't need that (as far as I can see yet)

avramidis:
I was trying to do that with variadic templates. I still get some errors
but it can be done I think.

I have to dig more into the new API because it seems I have some knowledge
gaps.

karlnapf:
I mean sure it is possible to offer this, exactly using those flexible templates. But it obscures the `get` API, which quite a cost for what is gained. IF we need such a thing (calling methods by a string, with multiple parameters), then we should add some other function that `get`.... something like `run`. But until we don't have to do that, I would be careful getting involved with such things.

BTW the new API is all moving around. We are trying to figure out what is best still.

karlnapf:
Any news here on this?

avramidis:
yes, I am still working on it but I am close. I had to create clone/equals methods for a couple of more classes which caused some unit tests to fail. I fixed those but now it fails for python/java generated examples. I will push again for feedback.

karlnapf:
ah ok. Yeah just push stuff. It is easiest to discuss it here. Might save you some time if devs point out stuff. Shogun is a beast ;)

avramidis:
The runtime error for the python string_char.sg meta example is:

`290: Traceback (most recent call last):
290:   File "/home/ea461/Workfolder/shogun_project/shogun/build/examples/meta/python/features/string_char.py", line 17, in <module>
290:     features = f.get("get_features")
290:   File "/home/ea461/Workfolder/shogun_project/shogun/build/src/interfaces/python/shogun.py", line 162, in _internal_get_param
290:     raise KeyError("There is no parameter called '{}' in {}".format(name, self.get_name()))
290: KeyError: "There is no parameter called 'get_features' in StringFeatures"`

src/shogun/features/StringFeatures.cpp
```diff
@@ -2081,6 +2086,32 @@ bool CStringFeatures<ST>::obtain_from_char_features(CStringFeatures<CT>* sf, int
 	return true;
 }
 
+template<class ST>
+CStringFeatures<ST>* CStringFeatures<ST>::clone() const
+{
+	return new CStringFeatures<ST>(*this);
```
karlnapf: note that `clone` and `equals` are implemented in `SGObject`, so no need to override them here, unless there are problems with the parameter framework (in which case it would be better to solve them)

src/shogun/features/StringFileFeatures.cpp
```diff
@@ -120,6 +121,32 @@ template <class ST> void CStringFileFeatures<ST>::fetch_meta_info_from_file(int3
...
+template<class ST>
+CStringFileFeatures<ST>* CStringFileFeatures<ST>::clone()
+{
+	CStringFileFeatures<ST>* result;
+
+	if (this->get_alphabet()->get_alphabet()==EAlphabet::NONE)
+	{
+		result = new CStringFileFeatures<ST>();
+	}
+	else
```
karlnapf: as above

src/shogun/lib/SGStringList.cpp
```diff
@@ -93,12 +93,27 @@ SGStringList<T> SGStringList<T>::clone() const
...
+bool SGStringList<T>::equals(const SGStringList<T>& other) const
+{
+	if (this->num_strings!=other.num_strings)
+		return false;
+
+	if (this->max_string_length!=other.max_string_length)
+		return false;
+
+	for (auto i : range(num_strings))
+		return ((SGString<T>)(this->strings[i])).equals(other.strings[i]);
```
karlnapf: sematic error: this returns after having compared the first string in the list. I guess you wanna return false of one string is different...

src/shogun/lib/SGStringList.h
```diff
@@ -66,6 +66,14 @@ template <class T> class SGStringList : public SGReferencedData
 	 */
 	SGStringList<T> clone() const;
 
+
+	/** Equals method to SGStringList
+	 * @param other SGStringList to compare with
+	 * @return false if alphabets and feature vectors are different,
```
karlnapf: "iff alphabets or feature vectors"

karlnapf: also the alphabet is not really compared in the method ...

src/shogun/lib/SGStringList.cpp
avramidis: Oh yes!! Good point :P

tests/unit/features/StringFeatures_unittest.cc
```diff
@@ -104,3 +106,15 @@ TEST(StringFeaturesTest,clone)
 	SG_UNREF(f);
 	SG_UNREF(f_clone);
 }
+
+TEST(StringFeaturesTest,equals)
+{
+	SGStringList<char> strings = generateRandomStringData();
```
karlnapf: we would want a test for `SGStringList` ... all shogun objects are tested for clone and equals in [here](https://github.com/shogun-toolbox/shogun/blob/develop/tests/unit/base/SGObjectAll_unittest.cc#L222)

karlnapf:
it seems weird that this error shows up. Are you sure you are running the same shogun version as the source shows? You could find out via printing all parameter names using `f.parameter_names()`...if it is not there, then your change hasnt made it into the lib yet

avramidis:
It seems to be in the list of the parameter names

290: ('alphabet', 'cache_size', 'features', 'get_features', 'get_num_vectors', 'mask_table', 'max_string_length', 'num_symbols', 'order', 'original_num_symbols', 'preproc', 'preprocess_on_get', 'properties', 'single_string', 'subset_stack')

avramidis:
This is the generated python file

```python
import numpy as np
from shogun import DynamicObjectArray
from shogun import RAWBYTE
from shogun import SerializableAsciiFile
from shogun import csv_file
from shogun import string_features

words = csv_file("../../data/words.dat")

#![create_features]
f = string_features(words, RAWBYTE)
#![create_features]

#![output stat]
max_string_length = f.get("max_string_length")
number_of_strings = f.get("get_num_vectors")
features = f.get("get_features")
#![output stat]

# Serialize output for integration testing (automatically generated)
sg_storage = DynamicObjectArray()
sg_storage_file = SerializableAsciiFile("string_char.dat", 'w')
sg_storage.append_element_int(max_string_length, "max_string_length")
sg_storage.append_element_int(number_of_strings, "number_of_strings")
sg_storage.save_serializable(sg_storage_file)
```

karlnapf:
Ah yes... I know what is wrong
come on irc to discuss more.

src/shogun/features/StringFeatures.cpp
avramidis: There where no problems so I removed the ones I added

tests/unit/features/StringFeatures_unittest.cc
avramidis: I've seen those. These are tests for constructors without arguments, right?

src/shogun/features/StringFeatures.cpp
```diff
@@ -78,19 +78,22 @@ template<class ST> CStringFeatures<ST>::CStringFeatures(CAlphabet* alpha)
...
+	num_vectors = orig.num_vectors;
+	single_string = orig.single_string;
+	length_of_single_string = orig.length_of_single_string;
+	max_string_length = orig.max_string_length;
+	num_symbols = orig.num_symbols;
+	original_num_symbols = orig.original_num_symbols;
+	order = orig.order;
+	preprocess_on_get = false;
+	feature_cache = NULL;
+
```
gf712: Isn't it better practice to do list initialisation, i.e. how it was before? Since this is a copy constructor all the class members should haven been initialised anyway

avramidis: The init() is called in the constructor which resets all the values to default. The unit test we have now doesn't catch his problem.

avramidis: And I am not sure if this is entirely correct because single_string is a pointer. How we have it now creates a shallow copy which I am not sure if it's what we want.

gf712: >The init() is called in the constructor which resets all the values to default. The unit test we have now doesn't catch his problem.

Ah ok! Sorry I didn't see that. 

>And I am not sure if this is entirely correct because single_string is a pointer. How we have it now creates a shallow copy which I am not sure if it's what we want.

True, that doesn't seem like a good idea. I am guessing if you delete the original CStringFeatures the copy has undefined behaviour?

avramidis: Yes, that's right. I will make another PR for that after the merge of this one.

tests/unit/features/StringFeatures_unittest.cc
karlnapf: For now, yes
We might add a version with random content soon

src/shogun/lib/SGStringList.cpp
```diff
@@ -93,12 +93,30 @@ SGStringList<T> SGStringList<T>::clone() const
...
+bool SGStringList<T>::equals(const SGStringList<T>& other) const
+{
+	if (this->num_strings!=other.num_strings)
+		return false;
+
+	if (this->max_string_length!=other.max_string_length)
+		return false;
+
+	for (auto i : range(num_strings))
+	if (((SGString<T>)(this->strings[i])).equals(other.strings[i])==false)
```
karlnapf: indentation problems. The CI build will give you a command to automagically fix this

karlnapf: btw why is the cast to `SGString<T>` necessary?

src/shogun/lib/SGStringList.cpp
```diff
@@ -93,12 +93,30 @@ SGStringList<T> SGStringList<T>::clone() const
...
+		strs[i] = this->strings[i].clone();
 
-	return SGStringList<T>(strings, num_strings, max_string_length);
+	return SGStringList<T>(strs, num_strings, max_string_length);
+}
+
+template <class T>
+bool SGStringList<T>::equals(const SGStringList<T>& other) const
+{
+	if (this->num_strings!=other.num_strings)
```
karlnapf: maybe one thing to add would be if the pointer address and the number of strings is the same, then we can return true straight away

src/shogun/lib/SGStringList.h
```diff
@@ -66,6 +66,14 @@ template <class T> class SGStringList : public SGReferencedData
 	 */
 	SGStringList<T> clone() const;
 
+
+	/** Equals method to SGStringList
```
karlnapf: minor: "Equals method" (no need to write the type)

src/shogun/util/factory.h
```diff
@@ -96,16 +96,16 @@ namespace shogun
 	}
 
 	CFeatures* string_features(
-	    CFile* file, EAlphabet alpha = DNA,
-	    EPrimitiveType primitive_type = PT_CHAR)
+	    CFile* file, machine_int_t alphabet_type = DNA,
```
karlnapf: doesn't swig take care of this automatically? I.e. is this strictly necessary?

src/shogun/util/factory.h
```diff
@@ -96,16 +96,16 @@ namespace shogun
...
+	    machine_int_t primitive_type = PT_CHAR)
 	{
 		REQUIRE(file, "No file provided.\n");
 		CFeatures* result = nullptr;
 
 		switch (primitive_type)
 		{
 		case PT_CHAR:
-			result = new CStringFeatures<char>(file, alpha);
+			result = new CStringFeatures<char>(file, static_cast<EAlphabet>(alphabet_type));
```
karlnapf: is this safe?

tests/unit/features/StringFeatures_unittest.cc
```diff
@@ -81,9 +81,11 @@ TEST(StringFeaturesTest,copy_subset)
 TEST(StringFeaturesTest,clone)
```
karlnapf: This test is kinda bad. Maybe we should just delete it and work on the non-empty tests for all sgobjects instead?

tests/unit/features/StringFeatures_unittest.cc
karlnapf: Same here (remove this). I would however love to see a test for clone/equals of `SGStringList` as those are very low level operations

karlnapf: And we should probably do these tests for all legal template arguments using the googletest magic.
@gf712 can help here

karlnapf:
Cool that it works for Python!! :)

src/shogun/features/StringFeatures.cpp
```diff
@@ -78,19 +78,22 @@ template<class ST> CStringFeatures<ST>::CStringFeatures(CAlphabet* alpha)
...
-	original_num_symbols(orig.original_num_symbols),
-	order(orig.order), preprocess_on_get(false),
-	feature_cache(NULL)
+: CFeatures(orig)
 {
 	init();
 
 	ASSERT(orig.single_string == NULL) //not implemented
 
+	num_vectors = orig.num_vectors;
```
karlnapf: General question. Does any of this affect the core PR changes? Because in general, I would prefer if we did those fixes in a different PR and keep this one strictly for adding the `get_features` of string features .... just to keep it a bit more tidy

examples/meta/src/features/string_char.sg
```diff
@@ -0,0 +1,11 @@
...
+#![output stat]
+int max_string_length = f.get_int("max_string_length")
+int number_of_strings = f.get_int("get_num_vectors")
+CharVector features = f.get_char_string_list("get_features")
```
karlnapf: I think `java.json` needs to be updated to that it doesn't translate `DoubleMatrix` ... java swig in fact has typemaps for `SGStringList`, i.e. it is able to translate this straight to a java data structure, done [here](https://github.com/shogun-toolbox/shogun/blob/develop/src/interfaces/java/swig_typemaps.i#L674):

Noticed that when staring the the CI error [here](https://dev.azure.com/shogunml/shogun/_build/results?buildId=290)

karlnapf:
java error [here](https://dev.azure.com/shogunml/shogun/_build/results?buildId=290&view=logs&jobId=1824f0ca-0d69-5973-dadc-e69062ba0810&taskId=bbb9c12e-ef76-5313-d299-1496a475c33a&lineStart=1621&lineEnd=1622&colStart=1&colEnd=1)
```
2019-01-19T12:45:00.0405452Z /build/examples/meta/java/features/string_char.java:29: error: cannot find symbol
2019-01-19T12:45:00.0405578Z DoubleMatrix features = f.get_char_string_list("get_features");
```
apart from the mentioned DoubleMatrix CharVector issue, I don't understand why the java interface doesn't know the `get_char_string_list` method, as you instantiated those in swig (otherwise they wouldn't work in python)

In fact csharp has the very same problem, method is not defined [see here](https://dev.azure.com/shogunml/shogun/_build/results?buildId=290&view=logs&jobId=d5fcc40e-7c75-58e1-78d9-decf4e025cc0&taskId=ab5441b9-8633-5fb1-9268-94fd6dd39d54&lineStart=1282&lineEnd=1283&colStart=1&colEnd=1)
```
2019-01-19T12:44:18.8254589Z string_char.cs(16,21): error CS1061: Type `Features' does not contain a definition for `get_char_string_list' and no extension method `get_char_string_list' of type `Features' could be found. Are you missing an assembly reference?
```

And so does octave, although here it is a runtime error as obviously nothing is compiled, [see here](https://dev.azure.com/shogunml/shogun/_build/results?buildId=290&view=logs&jobId=c13d9fde-dead-5630-14e6-09ee2f4efd6a&taskId=1be49215-08b6-5b1b-b1a4-97ad6ee93918&lineStart=155&lineEnd=156&colStart=1&colEnd=1)
```
2019-01-19T12:46:14.2967680Z error: member not found
```

src/interfaces/swig/shogun.i
```diff
@@ -233,6 +233,10 @@ namespace shogun
 %template(_get_real) CSGObject::get<float64_t, void>;
 %template(_get_int) CSGObject::get<int32_t, void>;
 %template(_get_real_matrix) CSGObject::get<SGMatrix<float64_t>, void>;
+%template(_get_bool_string_list) CSGObject::get<SGStringList<bool>, void>;
```
karlnapf: ah yes, here we have the offender. You are getting the CI errors since these methods are only defined within  `#ifdef SWIGPYTHON`. @gf712 can give instructions where else to add this.

BTW once #4470 is merged this will lead to a merge conflict in here. Depends on what is done earlier, but it should be easy to resolve it

src/shogun/features/StringFeatures.cpp
```diff
@@ -1731,6 +1734,8 @@ template<class ST> void CStringFeatures<ST>::init()
 
 	m_parameters->add_vector(&symbol_mask_table, &symbol_mask_table_len, "mask_table", "Symbol mask table - using in higher order mapping");
 	watch_param("mask_table", &symbol_mask_table, &symbol_mask_table_len);
+	watch_method("get_num_vectors", &CStringFeatures::get_num_vectors);
```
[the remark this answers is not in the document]
avramidis: I will have to try again the "features" because I was getting a double registration error. But I will let you know.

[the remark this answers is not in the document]
karlnapf: Ah yes. It is actually awkward as "features" is the registered raw array of `SGString` ... sigh.
"get_features" is actually a bad name as it simply converts the array to StringList. "string_list" might be the best name.

Eventually, it would be best to somehow refactor/unify all this. But not for now. Now is just for adding the string list to the API

examples/meta/generator/translate.py
```diff
@@ -90,11 +90,9 @@ def getSGTypeToStoreMethodName(sgType):
         return "real_matrix"
     elif sgType=="FloatMatrix":
         return "float_matrix"
-
```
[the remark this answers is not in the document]
avramidis: Sorry about that. I will have a look at gitkraken.

src/shogun/features/StringFeatures.cpp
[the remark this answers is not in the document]
avramidis: I can change it to string_list for now but this will affect a few files.

src/interfaces/swig/shogun.i
gf712: So now you need to add the method name as a string here and not worry about it.
https://github.com/shogun-toolbox/shogun/blob/13961a1d353414f301370425e124c18a4ce4c3b2/src/interfaces/python/swig_typemaps.i#L1348-L1354
And when you recompile it should be accessible via .get in the Python interface.

gf712: But you still need the template, which should be `%template(get_bool_string_list)`

tests/unit/features/StringFeatures_unittest.cc
gf712: Yup, for this you need to define the types you want to test:
https://github.com/shogun-toolbox/shogun/blob/13961a1d353414f301370425e124c18a4ce4c3b2/tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc#L47-L50
Associate it to a (empty) class:
https://github.com/shogun-toolbox/shogun/blob/13961a1d353414f301370425e124c18a4ce4c3b2/tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc#L26-L29
https://github.com/shogun-toolbox/shogun/blob/13961a1d353414f301370425e124c18a4ce4c3b2/tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc#L59
And then just pass the class to the TYPED_TEST macro
https://github.com/shogun-toolbox/shogun/blob/13961a1d353414f301370425e124c18a4ce4c3b2/tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc#L64
In the test you can access the type with `TypeParam`

src/shogun/util/factory.h
avramidis: I think it is fine since we know the specific kind of the object (i.e. EAlphabet).

tests/unit/features/StringFeatures_unittest.cc
avramidis: It is a good idea, however, let's do that in another PR?

karlnapf: yep

src/shogun/features/StringFeatures.h
avramidis: I removed this. The example works fine when I use SGStringList type to store the result of get_string_list (before it was get_features)

src/shogun/util/factory.h
avramidis: It doesn't compile differently. The other factory methods use the same types I used.

avramidis:
My tests show that it works for C# but fails for JAVA. The problem is that the example code generator for the JAVA example add to the JAVA code a false import. Specifically, it takes the code below and adds an import for the "StringCharList".

https://github.com/shogun-toolbox/shogun/blob/bfd1c318398cebaf7b82c717de55a9bf2cb7cc95/examples/meta/src/features/string_char.sg#L10

The fault added import is

`import org.shogun.StringCharList;`

What I want to do is set the example code generator to not add this import. Currently, I am investigating this but if someone has an idea please help.

karlnapf:
Can you paste the generated code?

avramidis:
> Can you paste the generated code?

import org.jblas.DoubleMatrix;
import org.jblas.FloatMatrix;

import org.shogun.shogun;
import org.shogun.DynamicObjectArray;
import org.shogun.EAlphabet;
import org.shogun.Features;
import org.shogun.File;
import org.shogun.SerializableAsciiFile;
import org.shogun.StringCharList;

public class string_char {
static {
System.loadLibrary("shogun");
}

public static void main(String argv[]) {
shogun.init_shogun_with_defaults();

File words = shogun.csv_file("../../data/words.dat");

//![create_features]
Features f = shogun.string_features(words, EAlphabet.RAWBYTE);
//![create_features]

//![output stat]
int max_string_length = f.get_int("max_string_length");
int number_of_strings = f.get_int("num_vectors");
String[] string_list = f.get_char_string_list("string_list");
//![output stat]

// Serialize output for integration testing (automatically generated)
DynamicObjectArray sg_storage = new DynamicObjectArray();
SerializableAsciiFile sg_storage_file = new SerializableAsciiFile("string_char.dat", 'w');
sg_storage.append_element_int(max_string_length, "max_string_length");
sg_storage.append_element_int(number_of_strings, "number_of_strings");
sg_storage.save_serializable(sg_storage_file);

}
}

karlnapf:
It should be straight forward to change this btw, I’ll ping @sorig tomorrow and describe the problem

karlnapf:
@sorig we basically need to add a blacklist of types to avoid in the includes, something like
```
"Dependencies": {
        "IncludeAllClasses": true, 
        ....
        "SkipImport", ["StringCharList"]
```

karlnapf:
@avramidis curious, what happens if you replace in `java.json` the entry `IncludeAllClasses` with `IncludeInterfacedClasses`?

BTW see `translate.py` for how the json is turned into code

avramidis:
I tried the `IncludeAllClasses` but didn't help. Yes, I will get into `translate.py` to see how I can make it work.

karlnapf:
what I mean explicitly setting it to `False` **and** adding the entry I pasted (interfaced classes) and setting it to true, did you try exactly that?

karlnapf:
can you paste the code listing that is generated if you do that?

avramidis:
I believe I did that and got error. I will do that again and paste the code

avramidis:
Yeah, it cannot generate the examples. I did this change in `java.json`

```
"Dependencies": {
        "IncludeAllClasses": false,
        "IncludeInterfacedClasses": true,
        "IncludeEnums": true,
        "IncludeGlobalFunctions": false,
        "DependencyListElement": "import org.shogun.$typeName;",
        "DependencyListSeparator": "\n"
    },
```

and got errors like this one 

```
cmake-build-release/examples/meta/java/statistical_testing/quadratic_time_maximum_mean_discrepancy.java:26: error: cannot find symbol
Features features_p = shogun.features(f_features_p);
```

karlnapf:
ah yes of course, thanks!

avramidis:
I will work on the `SkipImport` solution then.

karlnapf:
it is pretty easy to add a blacklist to the logic. You would check if the dictionary contains the blacklist key "ExcludeImport", and then assume you get a list of imports....
when all the unions are taken in `translate.py`, you can then remove any blacklisted imports in there
https://github.com/shogun-toolbox/shogun/blob/develop/examples/meta/generator/translate.py#L335

avramidis:
That what I was thinking too. I hope this is the last problem with this PR :)

karlnapf:
```
if "ExcludeImport" in self.targetDict:
            dependencies.remove(self.targetDict["ExcludeImport"]) # something in these lines, after all union

**Comment by karlnapf** [comment]
Hehehe, you will be surprised ;)

**Comment by avramidis** [comment]
OK the JAVA problem is fixed. Tomorrow I will take a look the octave one.

**Comment by karlnapf** [comment]
For octave can you paste the generated code?

**Comment by avramidis** [comment]
Ok. I pushed the changes. I tested octave on my pc and there are no errors.

File: src/interfaces/java/swig_typemaps.i [hunk 22]
```diff
@@ -718,7 +718,7 @@ TYPEMAP_STRINGFEATURES(int64_t, int, Int, jint, "Int[][]", "[[I")
 TYPEMAP_STRINGFEATURES(uint64_t, long, Long, jlong, "Long[][]", "[[J")
 TYPEMAP_STRINGFEATURES(long long, long, Long, jlong, "Long[][]", "[[J")
 TYPEMAP_STRINGFEATURES(float32_t, float, Float, jfloat, "Float[][]", "[[F")
-TYPEMAP_STRINGFEATURES(float64_t, double, Double, jdouble, "Doulbe[][]", "[[D")
+TYPEMAP_STRINGFEATURES(float64_t, double, Double, jdouble, "Double[][]", "[[D")
```
karlnapf: :laughing:

src/shogun/lib/SGStringList.cpp
karlnapf: exactly

karlnapf:
Cool, just checked the changes, all looks good. I will wait for the CI :)

avramidis:
Great! Thank you!

karlnapf:
@avramidis octave failed. Could you pls post the listing of the char features example?

karlnapf:
Error is explained [here](http://www.dm.unibo.it/~achilles/calc/octave.html/Defining-Functions.html)

karlnapf:
And here we go: https://github.com/shogun-toolbox/shogun/blob/develop/src/interfaces/octave/swig_typemaps.i#L369

karlnapf:
Have a look at the [incoming typemap](https://github.com/shogun-toolbox/shogun/blob/develop/src/interfaces/octave/swig_typemaps.i#L270)....looks like we need to write an octave string list typemap to make this work. It is actually quite good that we spotted that as it means one was never ever able to execute string based algorithms from octave ... :/
Let me know if you need help with writing the typemap.

avramidis:
@karlnapf are the changes ok with you?

src/interfaces/octave/swig_typemaps.i
```diff
@@ -368,21 +368,50 @@ TYPEMAP_STRINGFEATURES_IN(is_matrix_type() && arg.is_uint16_type, uint16NDArray,
 
 /* output typemap for CStringFeatures */
 %define TYPEMAP_STRINGFEATURES_OUT(type,typecode)
+%typemap(out) shogun::SGStringList<char>
+{
+    /* TODO STRING OUT TYPEMAPS */
+    shogun::SGString<char>* str = $1.strings;
```
karlnapf: Still todo?

src/interfaces/octave/swig_typemaps.i
```diff
@@ -368,21 +368,50 @@ TYPEMAP_STRINGFEATURES_IN(is_matrix_type() && arg.is_uint16_type, uint16NDArray,
...
+    int32_t i, num_strings = $1.num_strings;
+
+    Cell c(num_strings, 1);
+
+    for (i = 0; i < num_strings; i++) {
+        c(i)=std::string(str[i].string);
+        SG_FREE(str[i].string);
```
karlnapf: Does this maybe cause problems? The original string is destroyed?

avramidis: Correct. I copied these from the other languages interfaces. I removed them.

karlnapf:
Very nice work! This now allows to port quite a few more examples

Issue #4233, opened 2018-04-05
Redundant code in tests for Neural******Layer classes


grig-guz:
A lot of code gets repeated over and over in some tests, see
https://github.com/shogun-toolbox/shogun/blob/c8c66e61ebc8b6e90cbe726df7557c7291310cab/tests/unit/neuralnets/NeuralRectifiedLinearLayer_unittest.cc#L52

https://github.com/shogun-toolbox/shogun/blob/c8c66e61ebc8b6e90cbe726df7557c7291310cab/tests/unit/neuralnets/NeuralRectifiedLinearLayer_unittest.cc#L111

This needs to be updated with test fixtures and templates to avoid repetitiveness.

karlnapf:
go for it!

saatvikshah1994:
I think this issue can be closed now.

karlnapf:
Definitely.

Issue #4232, opened 2018-04-05
Update NeuralLayer classes and tests with linalg


grig-guz:
Multiple NeuralLayer functions and tests use Eigen/vanilla  linear algebra operations. For example, 
in NeuralLinearLayer:
https://github.com/shogun-toolbox/shogun/blob/c8c66e61ebc8b6e90cbe726df7557c7291310cab/src/shogun/neuralnets/NeuralLinearLayer.cpp#L97

or in NeuralRectifiedLinearLayer_unittest:
https://github.com/shogun-toolbox/shogun/blob/c8c66e61ebc8b6e90cbe726df7557c7291310cab/tests/unit/neuralnets/NeuralRectifiedLinearLayer_unittest.cc#L90

This needs to be updated with single linalg calls.

saatvikshah1994:
Could I work on this simultaneously with #4233

karlnapf:
@saatvikshah1994 go for it! :) Looking forward to seeing a patch here

saatvikshah1994:
I think this issue can be closed now.

karlnapf:
well done :)

karlnapf:
ah well, actually we also want to linalg-ize the code itself, not just the tests

PR #4424, opened 2018-11-25
Minimize Neural Layer Unit Test Redundancies

saatvikshah1994:
I've currently made changes for the RectifiedLinearLayer and the Softmax layer. Since its my first task, I wanted to be sure these changes look alright, before fixing the remaining Neural*Layer tests.

Addresses #4233 

I could probably also make the fixes needed to address #4232 as part of this PR if thats alright with you'll?


.gitignore
```diff
@@ -291,6 +291,10 @@ $RECYCLE.BIN/
 # Windows shortcuts
 *.lnk
 
+# CMake specific
```
saatvikshah1994: Sorry, this would be CLion specific will fix.

tests/unit/utils/Utils.h
```diff
@@ -55,4 +58,44 @@ SGStringList<T> generateRandomStringData(
 	return strings;
 }
 
+struct NeuralLayerTestUtil
```
karlnapf: this code is so specific, I suggest we move it into a helper file within `neuralnets`

tests/unit/utils/Utils.h
```diff
@@ -55,4 +58,44 @@ SGStringList<T> generateRandomStringData(
...
+{
+	template <typename T>
+	static SGMatrix<T> create_rand_sgmat(
```
karlnapf: I think this is not a function we should have in neuralnet unittests.
Rather, this should be somewhere where other random number generators sit, like `Random.h` or something more appropriate. Random normal matrices are used in a lot of places in shogun

tests/unit/utils/Utils.h
```diff
@@ -55,4 +58,44 @@ SGStringList<T> generateRandomStringData(
...
+	    int32_t num_rows, int32_t num_cols, T lower_bound, T upper_bound)
+	{
+		auto data_batch = SGMatrix<T>(num_rows, num_cols);
+		for (int32_t i = 0; i < data_batch.num_rows * data_batch.num_cols; i++)
+		{
+			data_batch[i] = CMath::random(lower_bound, upper_bound);
```
karlnapf: actually, there should also be a more efficient way to general an array of random normals. Check the array filling procedures in `Random.h`

tests/unit/utils/Utils.h
```diff
@@ -55,4 +58,44 @@ SGStringList<T> generateRandomStringData(
...
+		return std::make_tuple(data_batch, input_layer);
+	}
+
+	static auto init_neural_linear_layer(
+	    CNeuralLinearLayer* layer, CDynamicObjectArray* layers,
+	    SGVector<int32_t> input_indices, int32_t batch_size, double sigma)
+	{
+		layer->initialize_neural_layer(layers, input_indices);
+		SGVector<float64_t> params(layer->get_num_parameters());
+		SGVector<bool> param_regularizable(layer->get_num_parameters());
```
karlnapf: reg_params

tests/unit/neuralnets/NeuralRectifiedLinearLayer_unittest.cc
```diff
@@ -31,47 +31,49 @@
...
-	CNeuralInputLayer* input = new CNeuralInputLayer (x.num_rows);
-	input->set_batch_size(x.num_cols);
-
-	CDynamicObjectArray* layers = new CDynamicObjectArray();
-	layers->append_element(input);
-
+	SGMatrix<float64_t> x;
+	CNeuralInputLayer* input;
+	std::tie(x, input) =
+	    NeuralLayerTestUtil::create_rand_input_layer<float64_t>(
```
karlnapf: it is nice to pull this kind of stuff out!

tests/unit/neuralnets/NeuralSoftmaxLayer_unittest.cc
```diff
@@ -31,47 +31,46 @@
...
+#include "shogun/lib/SGMatrix.h"
+#include "shogun/lib/SGVector.h"
+#include "shogun/mathematics/Math.h"
+#include "shogun/neuralnets/NeuralInputLayer.h"
+#include "utils/Utils.h"
+#include "gtest/gtest.h"
 
 using namespace shogun;
 
+class NeuralSoftmaxLayerTest : public ::testing::Test
```
karlnapf: this fixture is almost identical to the other one. Is there a reason to have this second one?

tests/unit/neuralnets/NeuralRectifiedLinearLayer_unittest.cc
```diff
@@ -31,47 +31,49 @@
...
-
-	CNeuralInputLayer* input = new CNeuralInputLayer (x.num_rows);
-	input->set_batch_size(x.num_cols);
-
-	CDynamicObjectArray* layers = new CDynamicObjectArray();
-	layers->append_element(input);
-
+	SGMatrix<float64_t> x;
+	CNeuralInputLayer* input;
+	std::tie(x, input) =
```
saatvikshah1994: I had a question about which C++ version is Shogun restricted to?

I can further trim this if C++17 is permitted.

tests/unit/neuralnets/NeuralSoftmaxLayer_unittest.cc
saatvikshah1994: Yep will do so - I'll add a common fixture with the utility functions in `NeuralLayerUtil` right inside the `unit/neuralnets` folder

saatvikshah1994:
I had 2 `C++ in Shogun` related questions:
1. As asked above, upto what version of C++ is permitted?
2. Is there an option to run the tests with Address Sanitizer turned on?

saatvikshah1994:
Addressed all your comments.
P.S. I ran clang-format-3.8 directly(the `check-format.sh` script doesnt work on my Macbook) which ended up changing some additional lines.

karlnapf:
Ok cool, let me check this

tests/unit/neuralnets/NeuralLayerTestFixture.h
```diff
@@ -0,0 +1,99 @@
...
+
+ * 1. Redistributions of source code must retain the above copyright notice,
+ * this list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ *
+ * 3. Neither the name of the copyright holder nor the names of its
+ * contributors may be used to endorse or promote products derived from this
```
karlnapf: we should use the short style license here that is used in most places in shogun

tests/unit/neuralnets/NeuralLayerTestFixture.h
```diff
@@ -0,0 +1,99 @@
...
+#include <memory>
+#include <tuple>
+
+using namespace shogun;
+
+class NeuralLayerTestFixture : public ::testing::Test
+{
+public:
+	template <typename T>
+	SGMatrix<T> create_rand_sgmat(
```
karlnapf: see my earlier comment: this is not nn specific but should be a global shogun option

tests/unit/utils/Utils.h
```diff
@@ -3,12 +3,11 @@
...
  * Authors: Giovanni De Toni, Heiko Strathmann
  */
-#ifndef __UTILS_H__
-#define __UTILS_H__
+#pragma once
 
-#include <shogun/lib/SGMatrix.h>
-#include <shogun/lib/SGStringList.h>
-#include <shogun/lib/SGVector.h>
+#include "shogun/lib/SGMatrix.h"
```
karlnapf: why this?

tests/unit/utils/Utils.h
```diff
@@ -3,12 +3,11 @@
  *
  * Authors: Giovanni De Toni, Heiko Strathmann
  */
-#ifndef __UTILS_H__
-#define __UTILS_H__
+#pragma once
```
karlnapf: we will need to discuss whether we allow this...
@vigsterkr @lisitsyn ?

tests/unit/neuralnets/NeuralLayerTestFixture.h
```diff
@@ -0,0 +1,99 @@
...
+	    int32_t num_rows, int32_t num_cols, T lower_bound, T upper_bound) const
+	{
+		auto data_v = SGVector<T>(num_rows * num_cols);
+		data_v.random(lower_bound, upper_bound);
```
karlnapf: e.g. we could have a random method in SGMatrix (because we have one in vector already)
this code is a bit too much noise for my taste, especially in a test

tests/unit/utils/Utils.h
saatvikshah1994: I believed that its [standard convention](https://stackoverflow.com/a/21594/3656081) - system files <> vs. project files "". Mainly useful since it helps differentiate project includes vs. system includes.
I can revert if Shogun follows a different convention regarding this.

I had made that change earlier when I had added `NeuralLayerUtil`. Decided to leave that in since I felt its a style improvement.

karlnapf: yes but in order to compile the shogun tests we add the include path via `-I` so it by definition is a system include. Check out the other headers ... we are doing this everywhere :)

saatvikshah1994: Oh ok, thanks for clarifying. I should also reverse the `""` includes I've done in the other classes then.
Ill just revert my changes on this file then. If needed even the `#pragma once` can be added later.

tests/unit/neuralnets/NeuralLayerTestFixture.h
saatvikshah1994: This one:
```
/*
 *
 * Authors: ...
 */
```

saatvikshah1994:
Addressed comments - Further reduced a number of redundancies - Added shogun::linalg ops wherever possible.

src/shogun/lib/SGMatrix.cpp
```diff
@@ -227,6 +227,22 @@ void SGMatrix<T>::zero()
 	set_const(static_cast<T>(0));
 }
 
+template <class T>
+void SGMatrix<T>::random(T min_value, T max_value)
+{
+	for (index_t idx = 0; idx < num_rows * num_cols; ++idx)
```
karlnapf: pls, instead of a loop, use the `CRandom::fill_array*` functions that in some situations can be much much faster

tests/unit/neuralnets/NeuralRectifiedLinearLayer_unittest.cc
```diff
@@ -31,172 +31,119 @@
...
-			A_ref(i,j) = biases[i];
-
-			for (int32_t k=0; k<x.num_rows; k++)
-				A_ref(i,j) += weights[i+k*A_ref.num_rows]*x(k,j);
-
-			A_ref(i,j) = CMath::max<float64_t>(0, A_ref(i,j));
-		}
-	}
+	shogun::linalg::add_vector(
+	    shogun::linalg::matrix_prod(weights, x), biases, A_ref);
```
karlnapf: you could store this in a variable as it is not optimized out or anything anyways

tests/unit/neuralnets/NeuralSoftmaxLayer_unittest.cc
```diff
@@ -31,208 +31,147 @@
...
 #include <shogun/lib/SGMatrix.h>
-#include <shogun/mathematics/Math.h>
-#include <gtest/gtest.h>
+#include <shogun/lib/SGVector.h>
+#include <shogun/mathematics/linalg/LinalgNamespace.h>
+#include <shogun/mathematics/linalg/LinalgSpecialPurposes.h>
+#include <shogun/neuralnets/NeuralInputLayer.h>
+#include <shogun/neuralnets/NeuralSoftmaxLayer.h>
+
+#include <memory>
```
karlnapf: do you use anything from this?

saatvikshah1994: no will remove

tests/unit/neuralnets/NeuralRectifiedLinearLayer_unittest.cc
saatvikshah1994: Not sure what you mean? `add_vector` itself stores its result in the position denoted by A_ref and returns void.

src/shogun/lib/SGMatrix.cpp
saatvikshah1994: I was making it consistent with `SGVector` but sure - yes I noticed it uses SIMD instructions.
I think `SGVector::random` should be changed too accordingly - not sure whether should be done in this or a different PR.

karlnapf: 1. yes you are exactly right, this needs template specialization, and a loop fallback is fine imo
Let's do it for SGMatrix::random in this PR
2. I suggest we keep the interface without arguments, i.e. (`SGMatrix::random()` and `SGMatrix::random_normal`), and then the caller is responsible for scaling things to the appropriate bounds, using linalg

very nicely thought trough! :)

saatvikshah1994: Thanks, but I'd like to clarify (2) again. I get the `SGMatrix::random()` function which would just fill the array with a uniform random distribution(using `fill_array*`). For `random_normal` I would need to further apply something like the [Box-Muller transform](https://en.wikipedia.org/wiki/Box%E2%80%93Muller_transform) right? - I could try to do that using the `shogun::linalg` ops though not sure if its feasible with those alone.

karlnapf: @saatvikshah1994 
we have two options here:
1. Make a loop over the CRandom::random_normal insice `SGMatrix::random (since CRandom doesn't support SIMD random normal generation 
2. Implement a random_normal_array which
 a. Has a default implementation with a loop for now
 b. First uses fill_array and then re-uses the box muller trainsform code that is already present (would need some refactoring)

I suggest we keep things simple for now:
* offer `CRandom::random_normal_array` and use a loop for now
* if you want: make another PR once this is merged where you do option 2b from above

We definitely only want `SGMatrix::random_normal` to be a wrapper, i.e. no transforms or other complicated things in there.

saatvikshah1994: @karlnapf thanks this makes sense.
I had one suggestion though - I feel that the entire `random()` and `random_normal()` impl should go in its own separate (issue + PR combo). Reason being that (1) the implementation is less related to the immediate issue of minimizing the test redundancies and (2) the changes should ideally be propagated to `SGVector` as well - overall then a significant portion of this PR would mix up two disparate issues. For this PR, I could just use the original `NeuralLayerFixture::create_rand_matrix` which would then be replaced in the second PR. 

I'm up for whichever way you lean towards.

tests/unit/neuralnets/NeuralLayerTestFixture.h
```diff
@@ -0,0 +1,114 @@
...
+		return params;
+	}
+
+	// dynamic list of layers
+	std::unique_ptr<CDynamicObjectArray> m_layers;
+
+protected:
+	void SetUp() final
+	{
+		CMath::init_random(SEED);
```
karlnapf: do we need to set the seed? would prefer not to (portability)

saatvikshah1994: Not clear to me why this makes it less portable? I kept it to ensure consistency on each run + it was being done earlier :). You've approved the PR so should I change it still?

saatvikshah1994:
Once this is merged, I'll need to sift through the remaining Neural*Layer tests and apply similar changes.

tests/unit/neuralnets/NeuralLayerTestFixture.h
karlnapf: Different environments -> different random numbers for the same seed

If all the tests pass without it, I would prefer without ... let me know.

unit tests are one binary, so if you change the state of libshogun, all other tests might be affected. Even if not, it is better practice to only ensure something if it is needed.

Yes, all approved otherwise! :)

karlnapf:
> Once this is merged, I'll need to sift through the remaining Neural*Layer tests and apply similar changes.

great!

karlnapf:
Thanks for this.
Please keep an eye in the [CI}(https://dev.azure.com/shogunml/shogun/_build?definitionId=2) and the [buildbot](http://buildbot.shogun-toolbox.org:8080/). If anything breaks that wasn't broken before, we should fix it first thing.

saatvikshah1994:
The CI output looks correct but the buildbot seems to have failed here: http://buildbot.shogun-toolbox.org:8080/#/ on multiple machines although the errors reported seem arbitrary.

karlnapf:
Yep CI is indeed fine. Buildbot, just check the [builder list](http://buildbot.shogun-toolbox.org:8080/#/builders) for cases that were green until we merged this PR (last build). They are all fine. The broken builds need fixing, most of them at least, with the recently broken ones having a higher priority. It is usually a tiny bit of work to figure out what causes errors, but we usually have logs to see which commit broke them.

PR #4425, opened 2018-11-29
Added Minkowski distance meta example

avramidis:


karlnapf:
This looks good.
I am just confused, was there an integration testing file before? Usually we only have those if there is an example. 
What was the error when the k was in there?

karlnapf:
for reference https://github.com/shogun-toolbox/shogun-data/pull/170

karlnapf:
ok cool then good catch!

karlnapf:
I'll merge, let's keep an eye on the CI output (as pre-merge CI is down atm)

karlnapf:
https://dev.azure.com/shogunml/shogun/_build/results?buildId=92

avramidis:
Thanks! The CI output seems fine!

karlnapf:
yes all interfaces passed. Thanks for the patch!

BTW I thought again about those int enums. I think it slightly sucks as it is not typesafe. 

It might be worth looking into this at some point, but it is slightly messy
http://www.swig.org/Doc3.0/Java.html#Java_typesafe_enums
http://www.swig.org/Doc3.0/CSharp.html#CSharp
etc

I think before potentially doing anything here, one would try to fool `put` via placing an illegal int value that is fed into an enum, and then check. Technically, if we had proper handling of `default` cases when switching over the enums, all should be good (apart from potentially cryptic error messages). However, knowing the codebase, I am quite sure it will crash in some easy to spot situation ;)

Having said all that, I think it might be best to come back to this at a later point.

avramidis:
Yes, you are right. I think the best action for now would be to do all the examples to identify any potential problems like this enum business.

karlnapf:
BTW here is one thing that is related to examples/API that requires ongoing efforts (not sure there is an issue or not).

We eventually want to get rid of all specialized types in all examples and notebooks. We in fact did this for `CDesneFeatures` aka (SWIG) `RealFeatures`. The factory/put/get API allows to create and modify objects so that neither constructors nor special public methods of such specialized classes are needed.

So you could pick any notebook/example where you find a specialized type, and then refactor the code to use the new put/get API. This will reveal more problems that then can be fixed on the fly.

Once all test code (apart from c++ unittests) is free of the use of the class, we can then remove it from SWIG (the `.i` definition files in `src/interfaces/`). This speeds up the compile time quite drastically.
It can be done one after the other

1. remove all uses of special type in gpl/examples/notebooks
2. grep for the classname and all aliases and potentially goto 1
3. Remove classname from swig definition
4. Ensure all tests/builds pass

PR #4426, opened 2018-11-29
Check type of Any for casting

gf712:
To cast Any back to its original type we need to check each type used in shogun before casting.
I implemented a function that does this check and then executes a lambda when it finds the right type.
It still needs some work:
- [x] handle errors
- [x] add more type maps
- [x] maybe add some sort of check for the lambda, i.e. does it have the expected signature?
- [x] should it be void rather than return `int`? The `int` was just initially used for debugging.

Example usage:

```
#include <iostream>
#include <shogun/base/init.h>
#include <shogun/io/SGIO.h>
#include <shogun/lib/type_case.h>

using namespace shogun;

int main() {

	init_shogun();

	int a = 2.0;
	Any any = make_any(a);

	// declare a lambda that is executed when the original type is found
	// needs this signature
	auto found_it_lambda = [](auto value) {
		std::cout << "It's a " << demangled_type<decltype(value)>() << std::endl;
		std::cout << "Value " << value << std::endl;
	};

	try {
		sg_any_dispatch(any, sg_all_types, found_it_lambda);
	}
	catch (ShogunException &msg) {
		std::cout << msg.what() << std::endl;
	}

	try {
		sg_any_dispatch(any, sg_real_types, found_it_lambda);
	}
	catch (ShogunException &m) {
		std::cout << m.what() << std::endl;
	}
}
```
Expected output:
```
It's a int
Value 2
[ERROR] In file /Users/ghoben/anaconda3/include/shogun/lib/type_case.h line 197: Type int is not part of <long double, double, float>
```


karlnapf:
@lisitsyn check this out

src/shogun/lib/type_case.h
```diff
@@ -0,0 +1,152 @@
...
+		std::type_index t = std::type_index(any.type_info());
+		return map[t];
+	}
+	template <typename T1, typename T2>
+	typename std::enable_if<(not std::is_same<T1, Types0>::value), int>::type
+	recursive_type_finder(const Any& any, TYPE type, T2 func)
+	{
+		if (type == sg_primitive_type<typename T1::Head>::value)
+		{
+			typename T1::Head i;
```
karlnapf: what about a better name than i

Also the typenames could me more meaningful

src/shogun/lib/type_case.h
```diff
@@ -0,0 +1,152 @@
...
+		}
+		else
+		{
+			return recursive_type_finder<typename T1::Tail>(any, type, func);
+		}
+	}
+
+	template <typename T1, typename T2>
+	typename std::enable_if<std::is_same<T1, Types0>::value, int>::type
+	recursive_type_finder(const Any& any, TYPE type, T2 func)
```
karlnapf: what about just type_finder? Or even something else?

src/shogun/lib/type_case.h
```diff
@@ -0,0 +1,152 @@
...
+	{
+		return 0;
+	}
+
+	template <typename T>
+	int type_finder(const Any& any, typemap& typesmap, T func)
```
karlnapf: for_each_type? not appropriate anymore?

src/shogun/lib/type_case.h
```diff
@@ -0,0 +1,152 @@
...
+		PT_UNDEFINED = 16
+	};
+} // namespace shogun
+
+template <typename T>
+struct sg_primitive_type
+{
+};
+
+#define SG_PRIMITIVE_TYPE(T, ptype)                                            \
```
karlnapf: make sure to udef all the macros after using them. We don't want to pollute the scope too much

src/shogun/lib/type_case.h
```diff
@@ -0,0 +1,152 @@
...
+#include <shogun/lib/any.h>
+#include <shogun/lib/type_list.h>
+
+using namespace shogun;
+
+// here we define a list of all types that are implemented in shogun
+// the head of the list returns the type
+typedef Types<
+    bool, char, int8_t, uint8_t, int16_t, uint16_t, int32_t, uint32_t, int64_t,
+    uint64_t, float32_t, float64_t, floatmax_t>::type AllTypes;
```
karlnapf: sg_types
SG_TYPES
?
Also, those are only scalar types. What about adding a matrix type to start seeing what happens in those cases?
We will also need std::vector and friends.....which might show the issues with this approach: combinatorial explosions. 
I guess we could argue that the number of types are limited.

src/shogun/lib/type_case.h
```diff
@@ -0,0 +1,152 @@
...
+
+namespace shogun
+{
+	// here we make types switchable
+	enum class TYPE
+	{
+		PT_BOOL = 1,
```
karlnapf: the PT_ prefix comes from somewhere else?

src/shogun/lib/type_case.h
```diff
@@ -0,0 +1,152 @@
...
+		PT_INT16 = 5,
+		PT_UINT16 = 6,
+		PT_INT32 = 7,
+		PT_UINT32 = 8,
+		PT_INT64 = 9,
+		PT_UINT64 = 10,
+		PT_FLOAT32 = 11,
+		PT_FLOAT64 = 12,
+		PT_FLOATMAX = 13,
+		PT_SGOBJECT = 14,
```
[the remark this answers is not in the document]
karlnapf: uuuh, actually we might need to add all base types here as well....sigh
@lisitsyn any thoughts?

src/shogun/lib/type_case.h
gf712: OK! I just gave them the most standard names because I'm not very innovative with variable names..

src/shogun/lib/type_case.h
gf712: Yes, that could work. Btw do you know why it is considered better to pass a lambda as a rvalue reference rather than as a copy, i.e. `T&& func` rather than `T func`

src/shogun/lib/type_case.h
gf712: yes, that comes from `shogun/tests/unit/base/SGObjectAll_unittest.cc`

src/shogun/io/TBOutputFormat.cpp
```diff
@@ -106,6 +107,13 @@ tensorflow::Event TBOutputFormat::convert_scalar(
 		    "Unsupported type %s", value.first.get_value().type_info().name());
 	}
 
+	auto write_summary = [&](auto type) {
```
karlnapf: Question, could we explicit here with the variables passed on as reference?

src/shogun/lib/type_case.h
```diff
@@ -0,0 +1,165 @@
...
+	}
+	template <typename TypeList, typename Lambda>
+	typename std::enable_if<(not std::is_same<TypeList, Types0>::value), int>::type
+	type_finder(const Any& any, TYPE type, Lambda func)
+	{
+		if (type == sg_primitive_type<typename TypeList::Head>::value)
+		{
+			typename TypeList::Head temporary_type_holder;
+			func(temporary_type_holder);
+			return 1;
```
karlnapf: Is there a specific reason for 0 and 1? As opposed to something else?

src/shogun/lib/type_case.h
```diff
@@ -0,0 +1,165 @@
...
+			ADD_TYPE_TO_MAP(float32_t , TYPE::PT_FLOAT32)
+			ADD_TYPE_TO_MAP(float64_t , TYPE::PT_FLOAT64)
+			ADD_TYPE_TO_MAP(floatmax_t , TYPE::PT_FLOATMAX)
+	};
+	typemap real_types = {
+			ADD_TYPE_TO_MAP(float32_t , TYPE::PT_FLOAT32)
+			ADD_TYPE_TO_MAP(float64_t , TYPE::PT_FLOAT64)
+			ADD_TYPE_TO_MAP(floatmax_t , TYPE::PT_FLOATMAX)
+	};
+	typemap non_integer_types = {
```
karlnapf: Nice this will be usefil

src/shogun/lib/type_case.h
gf712: No, could be a boolean, or could be a void function. This is just a way to capture errors, like in my example above we could do this:
```
if (!for_each_type(any, all_types, found_it_lambda)) {
	std::cout << "FAILED" << std::endl;
}
```
Or could just throw an error from within the function if the type isn't found? Or allow a custom lambda to be executed in that situation?

src/shogun/io/TBOutputFormat.cpp
gf712: Do you mean something like:
```
auto write_summary = [value](auto type) {summaryValue->set_simple_value(
 		    any_cast<decltype(type)>(value.first.get_value()));
 	};
```

karlnapf: like that yes, so that callers are forced to think about the scope

src/shogun/lib/type_case.h
karlnapf: mmmh, we could throw an exception (recently added customized exceptions might be worth checking this out). If not, a boolean might be better than the int

karlnapf: question is whether there will be situations where no type is found but the caller doensnt know that....
If this does not make sense, exception would be best (with a nice error string) imo

gf712: I am trying out SG_ERROR, but it won't work here because it seems to only work with classes that inherit from CSObject.

gf712: Can I just generate an exception inside the function like this:
```
char buffer[256];
sprintf(buffer, "Unsupported type %s", any.type_info().name());
auto msg = std::string(buffer);
throw ShogunException(msg);
``` 

Also I realised that if a type is not in the typemap we can throw an exception immediately. It's only in the case that the type is completely unknown, i.e. not in TYPES, that we will have to go through the list and then throw an exception.

src/shogun/lib/type_case.h
karlnapf: ehm ....it rings some bell, but I can't remember right now... stackoverflow ;)

src/shogun/lib/type_case.h
karlnapf: i see, what about somehow re-using those from `DataType.h`

src/shogun/io/TBOutputFormat.cpp
karlnapf: ok great, `this` is quite redundant anyways, I like the C++ updates :)

src/shogun/lib/type_case.h
karlnapf: @vinx13 could you advise how to best use your new expections?

@gf712 try SG_SERROR for static contexts

Generally, we want to move towards better exceptions, with types. But this is especially true for errors that can be caused by users. This is more an internal error that is caused by a developer...so maybe the SG_SERROR is ok as well

vinx13: Use [`SG_THROW`](https://github.com/shogun-toolbox/shogun/blob/55ebca82b9872cdf5fdb08d5727a3ac92af09a2e/src/shogun/io/SGIO.h#L126) or `SG_STHROW` to throw an exception with custom type. 
The first argument is the type of the exception. The rest of the arguments are the same as `SG_ERROR`
e.g.
`SG_THROW(ShogunException, "...")`

src/shogun/io/TBOutputFormat.cpp
gf712: Yup, now just need to come up with a way to make sure people use this signature... 
Would it be worth having some checks (not sure how)? Or just add somewhere in the documentation?

src/shogun/lib/type_case.h
gf712: Ah cool! Thanks, works as expected!

src/shogun/lib/type_case.h
gf712: Ah seem to be the same! But will have to address the containers..

src/shogun/lib/type_case.h
gf712: Ah actually no, this gives me a segmentation fault:
```
SG_SERROR("Unsupported type %s", any.type_info().name());
```
which I didn't have when I did this:
```
char buffer[256];
sprintf(buffer, "Unsupported type %s", any.type_info().name());
auto msg = std::string(buffer);
throw ShogunException(msg);
```

gf712: ah nvm, I just need to add init_shogun() at the start of my example code!

src/shogun/io/TBOutputFormat.cpp
karlnapf: I think it is fine to let people look at examples and then use the same style. Documentation just outdates, and this is only for developers (who know how to browse the codebase) anyways.
What happens if a different signature is passed. Compile error I guess so that would be fine.

src/shogun/lib/type_case.h
karlnapf: Well we don't want to support *all* the containers. Also the container definitions are part of the old type system and probably will be removed soon (nobody needs strings of SGVectors).
What is more likely is that we support matrix/vector of the ptypes, and then strings...but let's do that later when it is needed.

src/shogun/io/TBOutputFormat.cpp
```diff
@@ -106,6 +107,13 @@ tensorflow::Event TBOutputFormat::convert_scalar(
 		    "Unsupported type %s", value.first.get_value().type_info().name());
 	}
 
+	auto write_summary = [&summaryValue=summaryValue, &value](auto type) {
+		summaryValue->set_simple_value(
+		    any_cast<decltype(type)>(value.first.get_value()));
+	};
+
+	for_each_type(value.first.get_value(), all_types, write_summary);
```
karlnapf: sorry one more thing. Could we make the naming in a way that it is clear that this is a shogun utility function.
Like adding a prefix `sg_` to all functions and global variances that have to do with this

src/shogun/lib/type_case.h
```diff
@@ -0,0 +1,170 @@
...
+#include <shogun/lib/any.h>
+#include <shogun/lib/type_list.h>
+
+using namespace shogun;
+
+namespace shogun
+{
+	typedef Types<
+	    bool, char, int8_t, uint8_t, int16_t, uint16_t, int32_t, uint32_t,
+	    int64_t, uint64_t, float32_t, float64_t, floatmax_t>::type SG_TYPES;
```
karlnapf: so this collects all types that we support going from any to compile time type right?
So if we added SGMatrix<float64_t>, it would be added here or?

src/shogun/lib/type_case.h
```diff
@@ -0,0 +1,170 @@
...
+#undef SG_PRIMITIVE_TYPE
+
+namespace shogun
+{
+	typedef std::unordered_map<std::type_index, TYPE> typemap;
+
+#define ADD_TYPE_TO_MAP(TYPENAME, TYPE_ENUM)                                   \
+	{std::type_index(typeid(TYPENAME)), TYPE_ENUM},
+
+	typemap all_types = {
```
karlnapf: sg_ prefixes would be good I think

karlnapf:
> Expected output:
> 
> ```
> It's a int
> Value 2
> [ERROR] In file /Users/ghoben/anaconda3/include/shogun/lib/type_case.h line 161: Type i is not part of the given typemap
> ```

Just looking at the output. The error message is a bit cryptic. It should state "Type int32_t is not part of <float32_t, float64_t, float128_t>" or something in those lines

src/shogun/io/TBOutputFormat.cpp
gf712: OK! Btw, do you have any rule when to put or leave out functions/typedefs/enums from the namespace?

src/shogun/lib/type_case.h
gf712: Yes, it should be added there. And then in the enum class TYPE and use `SG_PRIMITIVE_TYPE` to map the type to the enum and then to the appropriate map!

src/shogun/lib/type_case.h
gf712: You mean `sg_all_types` right? (as opposed to the name of the macro)

src/shogun/io/TBOutputFormat.cpp
karlnapf: minimal evasion, so only make something as visible/complex as needed, not more

src/shogun/lib/type_case.h
karlnapf: yes

gf712:
Without much more code I can get the message `Type int is not part of <long double, double, float>` 
To get  "Type int32_t is not part of <float32_t, float64_t, float128_t>" we would have to map the demangled strings to those strings, i.e. `{"long double":  "float128_t"}`

src/shogun/lib/type_case.h
```diff
@@ -75,10 +75,49 @@ namespace shogun
 {
 	typedef std::unordered_map<std::type_index, TYPE> typemap;
 
+	namespace type_internal
+	{
+		std::string demangled_type(const char* name)
```
karlnapf: wait, didn't we have a function like this somewhere already?

src/shogun/lib/type_case.h
```diff
@@ -75,10 +75,49 @@ namespace shogun
...
+			    abi::__cxa_demangle(name, nullptr, &length, &status);
+			std::string demangled_string(demangled);
+			free(demangled);
+#else
+			std::string demangled_string(name);
+#endif
+			return demangled_string;
+		}
+
+		std::string print_map(const typemap& map)
```
karlnapf: this is for debugging?
I don't think we really need this in the repo, or?

src/shogun/lib/type_case.h
```diff
@@ -139,30 +173,30 @@ namespace shogun
...
+		if (type == TYPE::PT_UNDEFINED)
 			SG_SERROR(
-			    "Type %s is not part of the given typemap",
-			    any.type_info().name());
-		}
-
-		type_finder<SG_TYPES>(any, type, func);
+			    "Type %s is not part of %s",
+			    type_internal::demangled_type(any.type_info().name()).c_str(),
+			    type_internal::print_map(typesmap).c_str())
```
karlnapf: ah sorry, we actually need it here, nevermind my above comment

src/shogun/lib/type_case.h
```diff
@@ -0,0 +1,204 @@
...
+	typename std::enable_if<std::is_same<TypeList, Types0>::value, void>::type
+	sg_type_finder(const Any& any, TYPE type, Lambda func)
+	{
+		SG_SERROR(
+		    "Unsupported type %s",
+		    type_internal::demangled_type(any.type_info().name()).c_str())
+	}
+
+	template <typename Lambda>
+	void sg_for_each_type(const Any& any, typemap& typesmap, Lambda func)
```
karlnapf: one thing: could you write a clear docstring for this function? So that it will be immediately clear to other devs how to use it (not so much how it works)

src/shogun/lib/type_case.h
gf712: yes, it is similar, but it is a templated function that requires the type. Here we don't have the type, but have the `std::type_info::name` instead (and we cant infer type from that), so I needed this custom function that I hid in this namespace.

src/shogun/lib/type_case.h
karlnapf: can we have one vector/matrix type for illustration?

src/shogun/lib/type_case.h
karlnapf: any way to re-use some redundant code?

karlnapf: especially the guard/macro bit....and only if it doesnt cause more trouble than it solves

gf712: I will give it a go! I can add both implementations in Any.h

src/shogun/lib/type_case.h
```diff
@@ -0,0 +1,204 @@
...
+			ADD_TYPE_TO_MAP(float64_t , TYPE::PT_FLOAT64)
+			ADD_TYPE_TO_MAP(floatmax_t , TYPE::PT_FLOATMAX)
+			ADD_TYPE_TO_MAP(complex128_t, TYPE::PT_COMPLEX128)
+	};
+
+#undef ADD_TYPE_TO_MAP
+	template <typename TypeList, typename Lambda>
+	typename std::enable_if<
+	    (not std::is_same<TypeList, Types0>::value), void>::type
+	sg_type_finder(const Any& any, TYPE type, Lambda func)
```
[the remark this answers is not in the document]
gf712: Good point, I can move it to the other namespace I created? Btw, should the other namespace be nested in shogun (i.e. `shogun::type_internal`), or completely separate?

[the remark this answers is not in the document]
karlnapf: yes move it to the other.
IDK whether they should be nested or not .... I guess so, would there be any case where we would use the latter on its own?

src/shogun/lib/type_case.h
gf712: I put it in Any.cpp as I was getting a duplicate symbol linker error.

src/shogun/lib/type_case.h
gf712: Oh I meant if I should write the signature for the lambda in the description, sorry!

src/shogun/lib/type_case.h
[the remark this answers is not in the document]
gf712: OK! No, I doubt it, it's very specific.

src/shogun/io/TBOutputFormat.cpp
```diff
@@ -106,6 +107,13 @@ tensorflow::Event TBOutputFormat::convert_scalar(
 		    "Unsupported type %s", value.first.get_value().type_info().name());
 	}
 
+	auto write_summary = [&summaryValue=summaryValue, &value](auto type) {
+		summaryValue->set_simple_value(
+		    any_cast<decltype(type)>(value.first.get_value()));
+	};
+
+	sg_for_each_type(value.first.get_value(), sg_all_types, write_summary);
```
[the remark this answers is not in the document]
gf712: All right! I am guessing those tests fit in Any_unittest.cc?

src/shogun/lib/type_case.h
karlnapf: Yes I think this is a good thing to document (we will see whether it outdates)

src/shogun/io/TBOutputFormat.cpp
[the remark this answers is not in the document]
karlnapf: yep, or an own test for the type_finger (whatever the name is)
To add tests, just add another cpp file and re-run cmake to detect it

src/shogun/lib/type_case.h
karlnapf: On the other hand that might be ok as in shogun, we usually only template API function according to the primitive type, not the container type (apart from some Any things)

gf712: Might have to think for it for a while... I cant think of a way how the compiler could do such a check at compile time. Might have to stick to the more naive implementation for now

karlnapf: No I think it is good to offer functions to deal with vector/matrix/std::vector separately... we do this specialization inside Any as well (see equals/clone)

src/shogun/io/TBOutputFormat.cpp
```diff
@@ -112,7 +112,7 @@ tensorflow::Event TBOutputFormat::convert_scalar(
 		    any_cast<decltype(type)>(value.first.get_value()));
 	};
 
-	sg_for_each_type(value.first.get_value(), sg_all_types, write_summary);
+	sg_for_each_scalar_type(value.first.get_value(), sg_all_types, write_summary);
```
karlnapf: mmh this name is a bit of a monster now ;)
I am thinking loud ... sg_dispatch_any, sg_dispatch_any_vector, sg_dispatch_any_matrix, sg_dispatch_any_stl
Don't know if that is even better

src/shogun/lib/type_case.h
```diff
@@ -38,8 +42,10 @@ namespace shogun
 		PT_FLOATMAX = 13,
 		PT_SGOBJECT = 14,
 		PT_COMPLEX128 = 15,
-		PT_UNDEFINED = 16
+		PT_SGVECTOR_FLOAT32 = 16,
```
karlnapf: PT stands for "primitive type" in the old parameter framework. So keeping that in here doesnt make sense, maybe we need a new prefix?

src/shogun/lib/type_case.h
gf712: Ah ok! I do think I found a solution. Can basically have different parameters in the sg_for_each_type to execute lambdas for specific datatypes, and then with some SFINAE magic can execute the right lambda or throw an error (or do nothing). Could a have a function signature like this:

```
template <typename Lambda1 = std::nullptr_t, typename Lambda2 = std::nullptr_t, typename Lambda3 = std::nullptr_t>
void sg_for_each_type(const Any& any, const typemap& typesmap,
            Lambda1 scalar_func = nullptr, Lambda2 vector_func = nullptr, Lambda3 matrix_func = nullptr)
```

gf712:
Getting this error from CI for the lint job:
```
2018-12-04T13:38:11.4740116Z ##[error]/usr/bin/docker: invalid reference format.
2018-12-04T13:38:11.4753526Z ##[error]See '/usr/bin/docker run --help'.
2018-12-04T13:38:11.4754489Z ##[error]/usr/bin/docker failed with return code: 125
```
Is that an issue with the script or is the format not correct in my code?

src/shogun/io/TBOutputFormat.cpp
gf712: Including "any" in the function name might make it easier to find the function in docs! So something along those lines would be better.

src/shogun/lib/type_case.h
gf712: Could just leave without a prefix? Otherwise just use SG_ instead?

src/shogun/lib/type_case.h
```diff
@@ -0,0 +1,369 @@
...
+
+using namespace shogun;
+
+namespace shogun
+{
+	typedef Types<
+		bool, char, int8_t, uint8_t, int16_t, uint16_t, int32_t, uint32_t,
+		int64_t, uint64_t, float32_t, float64_t, floatmax_t, SGVector<int32_t>,
+		SGVector<int64_t>, SGVector<float32_t>, SGVector<float64_t>,
+		SGVector<floatmax_t>, SGMatrix<int32_t>, SGMatrix<int64_t>,
```
karlnapf: those should suffice for now, so nice :)

src/shogun/lib/type_case.h
karlnapf: what about T_

src/shogun/lib/type_case.h
```diff
@@ -0,0 +1,369 @@
...
+		PT_SGMATRIX_FLOAT64 = 22,
+		PT_SGMATRIX_FLOATMAX = 23,
+		PT_SGMATRIX_INT32 = 24,
+		PT_SGMATRIX_INT64 = 25,
+		PT_UNDEFINED = 26
+	};
+
+	enum class CONTAINER_TYPE
+	{
+		SCALAR = 1,
```
karlnapf: CT_ ?

src/shogun/lib/type_case.h
```diff
@@ -0,0 +1,369 @@
...
+		VECTOR = 2,
+		MATRIX = 3
```
karlnapf: if there is another "container type" enum, I don't really get why we then also need to state the containers here?
Having both seems redundant...? Maybe I just don't get it

src/shogun/lib/type_case.h
```diff
@@ -0,0 +1,369 @@
...
+SG_PRIMITIVE_TYPE(SGVector<float32_t>, TYPE::PT_SGVECTOR_FLOAT32)
+SG_PRIMITIVE_TYPE(SGVector<float64_t>, TYPE::PT_SGVECTOR_FLOAT64)
+SG_PRIMITIVE_TYPE(SGVector<floatmax_t>, TYPE::PT_SGVECTOR_FLOATMAX)
+SG_PRIMITIVE_TYPE(SGVector<int32_t>, TYPE::PT_SGVECTOR_INT32)
+SG_PRIMITIVE_TYPE(SGVector<int64_t>, TYPE::PT_SGVECTOR_INT64)
+SG_PRIMITIVE_TYPE(SGMatrix<float32_t>, TYPE::PT_SGMATRIX_FLOAT32)
+SG_PRIMITIVE_TYPE(SGMatrix<float64_t>, TYPE::PT_SGMATRIX_FLOAT64)
+SG_PRIMITIVE_TYPE(SGMatrix<floatmax_t>, TYPE::PT_SGMATRIX_FLOATMAX)
+SG_PRIMITIVE_TYPE(SGMatrix<int32_t>, TYPE::PT_SGMATRIX_INT32)
+SG_PRIMITIVE_TYPE(SGMatrix<int64_t>, TYPE::PT_SGMATRIX_INT64)
```
karlnapf: this is not nice and quite long....
BUT the good thing is that we (shogun framework devs) do this once, and not people who write higher level framework code.

src/shogun/lib/type_case.h
```diff
@@ -0,0 +1,369 @@
...
+		TYPE get_type(const Any& any, const typemap& map)
+		{
+			auto type = std::type_index(any.type_info());
+			auto it = map.find(type);
+
+			return it == map.end() ? TYPE::PT_UNDEFINED : map.at(type);
+		}
+
+		template <typename T>
+		struct is_vector : public std::false_type
```
karlnapf: maybe `is_sg_vector` ?

src/shogun/lib/type_case.h
```diff
@@ -0,0 +1,369 @@
...
+				(std::is_same<VectorLambdaT, std::nullptr_t>::value and
+				 is_vector<T>::value) or
+				(std::is_same<MatrixLambdaT, std::nullptr_t>::value and
+				 is_matrix<T>::value),
+			void>
+		execute_function(
+			T value, ScalarLambdaT scalar_func, VectorLambdaT vector_func,
+			MatrixLambdaT matrix_func)
+		{
+			SG_SWARNING(
```
karlnapf: I feel this should not be a runtime warning?

src/shogun/io/TBOutputFormat.cpp
karlnapf: since you have the vector thingi now in there, maybe also refactor the histogram bit?

karlnapf: I think you might kill me, but maybe "dispatch" should be part of the function name? `sg_any_dispatch` comes to my mind now.... but not sure what is better. What are your toughts?

src/shogun/lib/type_case.h
karlnapf: is this used anywhere?

tests/unit/lib/type_case_unittest.cc
```diff
@@ -0,0 +1,121 @@
...
+	};
+
+private:
+	T value;
+};
+
+TEST(Type_case, positional_lambdas)
+{
+	float32_t a_scalar = 42.0;
+	auto a_vector = SGVector<float32_t>({a_scalar});
```
karlnapf: minor
`SGVector<float32_t> a_vector = {a_scalar}` is a tiny bit shorter ;)

tests/unit/lib/type_case_unittest.cc
```diff
@@ -0,0 +1,121 @@
...
+	auto f_vector = [&counter](auto type) { counter++; };
+	auto f_matrix = [&counter](auto type) { counter++; };
+
+	sg_for_each_type(any_scalar, sg_all_types, f_scalar);
+	EXPECT_EQ(counter, 1);
+
+	sg_for_each_type(any_vector, sg_all_types, nullptr, f_vector);
+	EXPECT_EQ(counter, 2);
+
+	sg_for_each_type(any_matrix, sg_all_types, nullptr, nullptr, f_matrix);
```
karlnapf: this API is not the most elegant....need to discuss this

src/shogun/lib/type_case.h
karlnapf: ++ for static_assert, always nicer

karlnapf:
> Getting this error from CI for the lint job:
> 
> ```
> 2018-12-04T13:38:11.4740116Z ##[error]/usr/bin/docker: invalid reference format.
> 2018-12-04T13:38:11.4753526Z ##[error]See '/usr/bin/docker run --help'.
> 2018-12-04T13:38:11.4754489Z ##[error]/usr/bin/docker failed with return code: 125
> ```
> Is that an issue with the script or is the format not correct in my code?

I think @vigsterkr is working on CI stuff atm, so it might behave weirdly ...

src/shogun/io/TBOutputFormat.cpp
gf712: I was thinking about that, but it uses std::vector instead of SGVector? Is there some implicit casting going on?

src/shogun/lib/type_case.h
gf712: I have been thinking that a lot of these things should be static asserts to make debugging easier... If everything compiles any developer might become more complacent...

src/shogun/lib/type_case.h
gf712: I know.. The only other time that it might be used is if someone wants to create a custom map. Do you mean the list is long or the names?

src/shogun/lib/type_case.h
gf712: Do we need std::vector too?

tests/unit/lib/type_case_unittest.cc
gf712: Hmmm ok! Can have three different functions, it could make my life easier when specialising the helper function templates

src/shogun/io/TBOutputFormat.cpp
gf712: Yes that sounds good! Was just waiting for a final decision before changing the name!

src/shogun/lib/type_case.h
```diff
@@ -0,0 +1,369 @@
...
+	 * @code
+	 * auto f = [&any](auto type) {
+	 *     std::cout << any_cast<decltype(type)>(any);
+	 * };
+	 * @endcode
+	 * The function accepts three lambda expressions for the cases where the
+	 * type is scalar, SGVector or SGMatrix.
+	 *
+	 * @param any Any object
+	 * @param typesmap check the underlying type of Any given this map
```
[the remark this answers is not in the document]
gf712: It is the equivalent to three functions, but could avoid repeating 3 almost equal lines of code with this call.. The nullptr is annoying, it's a shame you cant use python style function calls!

src/shogun/lib/type_case.h
gf712: We could check the signature of the function to see if it is void and has one parameter at compile time.

src/shogun/lib/type_case.h
```diff
@@ -0,0 +1,369 @@
...
+			ADD_TYPE_TO_MAP(floatmax_t , TYPE::PT_FLOATMAX)
+			ADD_TYPE_TO_MAP(complex128_t, TYPE::PT_COMPLEX128)
+	};
+#undef ADD_TYPE_TO_MAP
+
+	/** Checks the underlying type of an Any instance and
+	 * executes the appropriate lambda. The type is checked
+	 * against a typemap and if it isn't found there
+	 * a ShogunException is thrown.
+	 * The lambda function must have a specific signature,
```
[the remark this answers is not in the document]
gf712: I don't think so. Because the compiler knows what type is required by the function inside the lambda, i.e. summaryValue->set_simple_value() should be a float, but at compile time we don't know what the dispatched object is, so we compile functions for each possibility, including vector and matrix. That's why knowing if it's scalar, vector or matrix helps out a lot with SFINAE and avoid a compile error.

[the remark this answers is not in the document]
gf712: The only way would be to check the signature of all function calls inside the lambda and use that to figure out if it's scalar, vector or matrix, but I don't know if that is possible!

src/shogun/io/TBOutputFormat.cpp
karlnapf: mmh phew ...
I would like to think it is possible to refactor this to SGVector?
On the other hand, it might be nice to support std:: structures as well

src/shogun/lib/type_case.h
karlnapf: I like such static assertions a lot!

src/shogun/lib/type_case.h
karlnapf: no just the big stack. But it is OK, as it is better than before

src/shogun/lib/type_case.h
karlnapf: Let's think. The feature of going from any to typed will be used in
* logging (only model parameters, numerical stuff)
* modelselecton (again, mostly model parameters, here we will need `SGObject` as well ...)
* not sure what else?
So with all that in mind, I think those types should be ok. And we can extend on request .... like done for any

tests/unit/lib/type_case_unittest.cc
karlnapf: I wonder: Can't this be implicitly be done by forcing the lambda to have a different signature (including SGVector/matrix) ....
If not, I think having three functions is the best bet, at least API wise. They all could call the function you wrote here internally to avoid redundant code

src/shogun/io/TBOutputFormat.cpp
karlnapf: lol ... let's hope this is the last :D

src/shogun/lib/type_case.h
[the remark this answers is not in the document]
karlnapf: It is clear that the caller needs to know what he wants to dispatch to (primitive,vector/matrix).
BUT, what if the call is the same for all, and then the compiler uses the signature of the lambda to decide. This means we would have to change the signature of course.

* define lambda (with SGVector in signature somehow)
* pass to `sg_any_dispatch`
* compiler sees lambda and knows that it is a vector and does the appropriate call

If that is not possible, I think 3 different functions is the best compromise. Or we ask @lisitsyn if he has an idea (though he is busy at nips until next week)

tests/unit/lib/type_case_unittest.cc
gf712: I think that could be possible, assuming that the compiler can implicitly convert SGVector and SGMatrix specialisations, right? Like in the summaryValue->set_simple_value(float value), I think all scalars can be interpreted as float right?

Btw seems like `std::is_scalar` is quite a broad definition, will have to find a more strict definition like `std::is_arithmetic` (potentially combined with a `std::decay`)

src/shogun/io/TBOutputFormat.cpp
gf712: I think the conversion to std::vector is necessary, because tensor board doesn't know about SGVector, like in this case:
https://github.com/shogun-toolbox/shogun/blob/2f0b9cfe59d18fec38904fbc74ba5aa55a531e33/src/shogun/io/TBOutputFormat.cpp#L125-L135

src/shogun/lib/type_case.h
gf712: Alright, got 1 static assertion working! I didn't work on anything else yet as this took a while to figure out. I also wanted to add a test for static_assert with gtest, and that wasn't easy but now its working! :D

src/shogun/lib/type_case.h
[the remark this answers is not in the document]
gf712: That should be doable! It's a matter of API design, functionally that is almost the same as we have now.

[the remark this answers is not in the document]
gf712: I added some utility structs to figure out if the type is SGVector or SGMatrix, so can use that to execute the correct lambda

gf712:
Things seem to be working as expected, but the compiler generates so many warnings (all expected) because it looks at the function implementation of each type, due to the recursive nature of the check implementation. What is nice is that the way the static assert is written guarantees that the error doesn't get buried in warnings.

src/shogun/lib/type_case.h
gf712: Should I replace the warning with an exception? The idea of the warning was that there could be situations where no lambda implementation exists of the expected type on purpose and an exception could abort a ML run without need.

src/shogun/lib/type_case.h
gf712: Yup, can use that

tests/unit/lib/type_case_unittest.cc
karlnapf: > I think that could be possible, assuming that the compiler can implicitly convert SGVector and SGMatrix specialisations, right?
What do you mean by that? BTW not sure if that helps, but It is definitely possible to extract inner types from templates, see [this ugly thing](https://github.com/shogun-toolbox/shogun/blob/develop/src/interfaces/swig/shogun.i#L151) for example

> Like in the summaryValue->set_simple_value(float value), I think all scalars can be interpreted as float right?
That is true, but the compiler will not implicitly convert `float` to `SGVector`
> 

> Btw seems like `std::is_scalar` is quite a broad definition, will have to find a more strict definition like `std::is_arithmetic` (potentially combined with a `std::decay`). 
Yes definitely, I would have a typedef for that that checks, see [here](https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/base/base_types.h)

src/shogun/lib/type_case.h
karlnapf: > Alright, got 1 static assertion working! I didn't work on anything else yet as this took a while to figure out. I also wanted to add a test for static_assert with gtest, and that wasn't easy but now its working! :D

nice one!

karlnapf: > Should I replace the warning with an exception? The idea of the warning was that there could be situations where no lambda implementation exists of the expected type on purpose and an exception could abort a ML run without need.

I would say yes, with an appropriate string describing what is wrong

karlnapf:
> Things seem to be working as expected, but the compiler generates so many warnings (all expected) because it looks at the function implementation of each type, due to the recursive nature of the check implementation. What is nice is that the way the static assert is written guarantees that the error doesn't get buried in warnings.

is there a way to silence the warnings?

tests/unit/lib/type_case_unittest.cc
```diff
@@ -64,7 +75,7 @@ TEST(Type_case, exception)
 	auto f_scalar = [&counter](auto type) { counter += 1; };
 
 	EXPECT_THROW(
-		sg_for_each_type(any_scalar, sg_real_types, f_scalar), ShogunException);
+	    sg_for_each_type(any_scalar, sg_real_types, f_scalar), ShogunException);
```
karlnapf: quite nice!

gf712:
> is there a way to silence the warnings?

There are two ways, either with compiler flags or add some pragmas (https://stackoverflow.com/a/7159392)

karlnapf:
> > is there a way to silence the warnings?
> 
> There are two ways, either with compiler flags or add some pragmas (https://stackoverflow.com/a/7159392)

ok cool, might be worth looking into the pragma (later); we cannot change the flags

gf712:
> I will to read through this in a quiet minute.

Yup, it's quite hard to understand it. I based it on this talk https://channel9.msdn.com/Events/CPP/CppCon-2016/CppCon-2016-Roland-Bock-How-To-Test-staticassert

> I like the static assertions a lot (though they come at some cost in readability), while being super helpful for downstream development.

Depending on how we continue developing the API we can have more specific static_asserts

> The None{} is API-wise quite similar to nullptr

Yes, it's the same, but felt more natural. Could replace with a macro `#define NONE None{}`?

> I guess the thing I had in mind was overloading in the style we did in put, i.e. overloading via explicitly defining the templated methods for certain types
sg_dispatch(Any, Lambda<T>) // T is primitive
sg_dispatch(Any, Lambda<SGVector<T>>) // T is primitive
or even
sg_dispatch(Any, Lambda<U<T>>) // T is primitive, U is SG*

Ah ok, I will have a look at that. But I thought that was C++17?

karlnapf:
The NONE is not really necessary, as it is all the same: positional arguments.

Check what we did in `put` here:
We define a different put for [objects](https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/base/SGObject.h#L456) and [primitives / vectors](https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/base/SGObject.h#L482), and even more involved for certain swig cases such as [scalars](https://github.com/shogun-toolbox/shogun/blob/develop/src/interfaces/swig/shogun.i#L130), [double matrices 1](https://github.com/shogun-toolbox/shogun/blob/develop/src/interfaces/swig/shogun.i#L152) [double matrices 2](https://github.com/shogun-toolbox/shogun/blob/develop/src/interfaces/swig/shogun.i#L206) etc. The compiler then figures out which call to make, it is not c++17, pretty simple in fact. The internals are not that pretty, but the API is: just a `put`. I wonder whether this is possible with the lambda signature as well so that there is only one `sg_any_dispatch` which is overloaded for the different situations. I am by no means sure this goes, but it might be worth trying

gf712:
Oh I meant the part where we extract the type from the lambda. I am not sure that is possible in C++14, but there might be a work around!

karlnapf:
Yes good idea to have the 3 lambda version be there in case we want to revert to it.
Why not keep it for now, and if you start another approach we can have another PR for that?

gf712:
I got a 1 lambda version working. 
- Checks at compile time:
   1. lambda has only 1 arg
   2. lambda's only arg is a valid type
   3. lambda is void

- Checks at runtime if the Any type is valid and if it's in the map. 

Should I push that version and then can decide what is best?

gf712:
The issue is that you can't convert `SGVector<float32_t>` to `SGVector<float64_t>` and all other variations.. Is that something that could/should be implemented?

karlnapf:
Yes, maybe open another PR with it? We can keep this one as is for now

gf712:
OK, I am happy with the code now! The one lambda version doesn't work as I expected so might have to talk over how that would work at some point.

karlnapf:
Pretty happy here as well. This is definitely merge-able from my side. Clean API as well.
Let's see whether the lambda type extraction gets you somewhere and if not, continue .... We can always change this later as the touch points with the rest of the code are super minimal and are actually easy refactors ...

karlnapf:
We should also move on from this soon actually :) Soon the exciting stuff begins

gf712:
Ah one last thing, should we replace std::is_scalar with std::is_arithmetic?

karlnapf:
what about `is_sg_primitive` ? :D

karlnapf:
No but honestly, I think defining such thins explicitly might be best actually

gf712:
So something that would include these: bool, char, int8_t, uint8_t, int16_t, int32_t, int64_t, float32_t, float64_t, floatmax_t?

karlnapf:
Yes, exactly. Just to have those types centrally defined so we can use them in other templated methods as well. Thanks!

gf712:
OK, should be good to go now!

src/shogun/lib/any.h
```diff
@@ -44,30 +44,22 @@
 #include <string>
 #include <typeinfo>
 #include <type_traits>
-#ifdef HAVE_CXA_DEMANGLE
-#include <cxxabi.h>
-#endif
 
 namespace shogun
 {
+	std::string demangled_type_helper(const char* name);
```
karlnapf: shall we maybe hide this one from outside?

karlnapf:
I somehow thought I merged this already? Shall I, or was something still missing?

gf712:
I think everything is in place and working! It's only a first iteration anyway.

karlnapf:
Ok shall I then? Just had another look and this is definitely something to use for a start

karlnapf:
Ok let’s watch ci. There might be compiler issues on some platforms with this modern stuff

karlnapf:
Awesome!

gf712:
OK! There seem to be some issues with msvc, like #4443.. But it might be an old bug in the compiler that has been fixed in more recent versions.

gf712:
CI seems to be failing during testing but due to `double free or corruption` in other tests..?

gf712:
OK, I'll investigate, but I can't reproduce these issues locally

Issue #4430, opened 2018-12-06
Random matrix Generation


saatvikshah1994:
As discussed with @karlnapf in #4424, there are several improvements/additions that could be made to the random number generation(RNG) API for SGMatrix/SGVector. Neither take advantage of the `CRandom::fill_array` vectorized RNGs. SGMatrix does not support RNG while SGVector simply loops over the entire legnth calling `CMath::random` per index. In summary the proposal is to add the following for both SGMatrix/SGVector:
1. `SG*::random()`: Generates random numbers in the [0, 1] range for float types. Generates numbers in [std::numeric_limits<T>::min,std::numeric_limits<T>::max] for int types. Uses the `fill_array` vectorized RNG wherever possible and falls back to simple for loop otherwise. These will follow a **uniform distribution**.
2. `SG*::random_normal(mu, sigma)`: Uses SG*::random() followed by a possibly vectorized variant of the [Box-Muller transform](https://en.wikipedia.org/wiki/Box%E2%80%93Muller_transform).
3. **To be discussed:** Since `SGVector::random(lb, ub)` is publicly exposed already I dont think it can be removed. A SG*::random(lb, ub) could be additionally supplied which utilizes SG*::random() to generate random numbers more efficiently.

Finally, I also think it might be important to benchmark a simple `for-loop` vs. `vectorized` variant for these cases.

karlnapf:
Really nice write-up!

One comment, 3. we can definitely change the api here ... towards a more standard `random` and `random_normal`, or even better: `rand()`, and `randn()`

saatvikshah1994:
I'd like to work on this as my next issue. One thing I realized while looking at the Random.h is that [0, 1] is not available within [dSFMT](http://www.math.sci.hiroshima-u.ac.jp/~m-mat/MT/SFMT/) used for vectorized double generation. So would have to do [0, 1) instead.

karlnapf:
That should be fine imo as the boundaries are a set of zero measure i.e. `P(x=1)=0` almost surely

PR #4437, opened 2018-12-15
Vectorized RNG

saatvikshah1994:
Addresses #4430 

Only adds Uniform Random capability for now:
- [x] SGMatrix
- [x] SGVector
- [x] Benchmark

Edit: Will add Normal random in a different PR.


src/shogun/lib/SGMatrix.h
```diff
@@ -309,6 +333,68 @@ template<class T> class SGMatrix : public SGReferencedData
...
+		 * fill matrix with uniformly distributed randoms
+		 * Float Types: [0, 1)
+		 * Int Types: std::numeric_limit<T>::(min->max)
+		 * Rest: disabled
+		**/
+		// vectorized available
+		template <typename U>
+		using enable_simd_float = std::enable_if_t<is_any<U, float64_t>::value>;
+		template <typename U = T>
+		auto random() -> enable_simd_float<U>
```
saatvikshah1994: separating declaration and definition would need explicit specialization of each template method signature - which would end up being really long. Since the methods are relatively short it seemed ok to keep both in the header file.

The only disadvantage is that an additional include of Math.h has to be added here

src/shogun/lib/SGMatrix.cpp
```diff
@@ -1193,6 +1192,7 @@ SGVector<T> SGMatrix<T>::get_diagonal_vector() const
 	return diag;
 }
 
+// Explicit Specialization of Templates
```
saatvikshah1994: unneeded - will revert

src/shogun/lib/SGMatrix.h
```diff
@@ -309,6 +333,68 @@ template<class T> class SGMatrix : public SGReferencedData
...
+		{
+			CRandom r{};
+			r.fill_array_co(
+			    static_cast<float64_t*>(matrix), num_rows * num_cols);
```
gf712: Here you could write `static_cast<U*>(matrix), num_rows * num_cols)` instead, no?

src/shogun/lib/SGMatrix.h
saatvikshah1994: Yep - Actually 3 versions are possible which mean the same:
1. https://coliru.stacked-crooked.com/a/bc08035924d3582e
2. https://coliru.stacked-crooked.com/a/7cc61a4adcdd5ded
3. https://coliru.stacked-crooked.com/a/45dba741ddf2d4a0

Mine is a slightly more readable version of (2) :D

src/shogun/lib/SGMatrix.h
saatvikshah1994: Yep I could. Will do so!

Yea I wanted to get the Thumbs up on this version. Adding it for `SGVector` shouldnt take too long on top of this. Dont think there's any other container right?

Also the type trait `is_any` <- I think this should go in some other file where it can be reused. Maybe `common.h`?

src/shogun/lib/SGMatrix.h
gf712: OK! It is more readable indeed! Why do you need the `template <typename U = T>` though? Doesn't `template <typename T>` work?

src/shogun/lib/SGMatrix.h
gf712: I think so, you will need @karlnapf to get back to you on that one though. Btw I think you will need to rename `is_any`, because there is an Any class in shogun, which might cause some confusion. Is `is_any_of` too verbose? 
Btw if you always use the value of `is_any`, instead of its type, you could write a `is_any_v` as a short cut to `is_any<...>::value`. I think that is more conforming with the direction in which c++ is going. (see for example [std::is_void_v](https://en.cppreference.com/w/cpp/types/is_void))

src/shogun/lib/SGMatrix.h
```diff
@@ -26,6 +28,28 @@ namespace Eigen
 	template <class, int, class> class Map;
 }
 
+/**
+ * Useful type traits for template overloads
+ */
+// Checks if any one of the types matches
+// Ref: https://stackoverflow.com/a/17032517/3656081
+template <typename T, typename... Rest>
+struct is_any : std::false_type
```
karlnapf: I wonder whether this is the best place to put this?

src/shogun/lib/SGMatrix.h
```diff
@@ -309,6 +333,68 @@ template<class T> class SGMatrix : public SGReferencedData
 		/** fill matrix with zeros */
 		void zero();
 
+		/**
```
karlnapf: Minor: start sentence with capital letter. I.e. "Fill matrix ... "

src/shogun/lib/SGMatrix.h
```diff
@@ -309,6 +333,68 @@ template<class T> class SGMatrix : public SGReferencedData
...
+		 * fill matrix with uniformly distributed randoms
+		 * Float Types: [0, 1)
+		 * Int Types: std::numeric_limit<T>::(min->max)
+		 * Rest: disabled
```
karlnapf: No need to state the disabled as this is obvious from the signature

src/shogun/lib/SGMatrix.h
```diff
@@ -309,6 +333,68 @@ template<class T> class SGMatrix : public SGReferencedData
...
+		**/
+		// vectorized available
+		template <typename U>
+		using enable_simd_float = std::enable_if_t<is_any<U, float64_t>::value>;
```
karlnapf: Cute!
I am wondering whether we should maybe move all the template magic out of SG* since that is supposed to only be a wrapper class. Maybe Random itself.... or linalg?

src/shogun/lib/SGMatrix.h
karlnapf: +1 for the rename...
Actually, we might even go explicit for now as as there are not THAT many combinations...I like it on the other hand, so could also keep it as a generic tool within the lib (under a different name)

src/shogun/lib/SGMatrix.h
```diff
@@ -309,6 +333,68 @@ template<class T> class SGMatrix : public SGReferencedData
...
+		template <typename U>
+		using enable_nonsimd_float =
+		    std::enable_if_t<is_any<U, float32_t, floatmax_t>::value>;
+		template <typename U = T>
+		auto random() -> enable_nonsimd_float<U>
+		{
+			CRandom r{};
+			// Casting floats upwards or downwards is safe
+			// Ref: https://stackoverflow.com/a/36840390/3656081
+			for (index_t i = 0; i < num_rows * num_cols; i++)
```
karlnapf: std::for_each ?

src/shogun/lib/SGMatrix.h
```diff
@@ -309,6 +333,68 @@ template<class T> class SGMatrix : public SGReferencedData
...
+			CRandom r{};
+			for (index_t i = 0; i < num_rows * num_cols; i++)
+			{
+				matrix[i] = r.random(
+				    std::numeric_limits<T>::min(),
+				    std::numeric_limits<T>::max());
+			}
+		}
+
+		// disabled for the remaining - will throw no matching fn call compile
```
karlnapf: remove

tests/unit/lib/SGMatrix_unittest.cc
```diff
@@ -13,6 +13,22 @@
 
 using namespace shogun;
 
+TEST(SGMatrixTest, vectorized_rng)
+{
+	auto rng_check_float = [](auto elem) {
+		SGMatrix<decltype(elem)> sg_mat(10, 5);
```
karlnapf: would there be a away to check it against the individual calling of random with some seed fixing magic?
If that's a pain, dont bother, this test is more to check that no memory errors happen I guess

src/shogun/lib/SGMatrix.h
gf712: It might be worth having a file with all the utility type_traits, no? Something like `sg_type_traits.h`

src/shogun/lib/SGMatrix.h
saatvikshah1994: SFINAE only works on templated functions. A method inside a template class doesnt count. The function itself has to also be templated. As mentioned in [CPPReference](https://en.cppreference.com/w/cpp/language/sfinae) `This rule applies during overload resolution of function templates`

src/shogun/lib/SGMatrix.h
saatvikshah1994: I think it should stay here - Ive noticed that macros have been used at come places to replicate this functionality which imho is a bad practice. eg. [here](https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/lib/SGVector.cpp#L415). Keeping such examples would help push towards using SFINAE/tag dispatch for overload resolution.

src/shogun/lib/SGMatrix.h
saatvikshah1994: `std::for_each` wont work with a stateful lambda which would be needed here `[&r]{...r.random...}`?

src/shogun/lib/SGMatrix.h
saatvikshah1994: done

tests/unit/lib/SGMatrix_unittest.cc
saatvikshah1994: might be a bit of a chore to implement - especially because I'd have to based on the type use the `random_half_open()` vs. `fill_array_oc` version.

saatvikshah1994:
@karlnapf I've added the task to benchmark - definitely intend to and will look into it after adding the `SGVector` version.

src/shogun/lib/SGMatrix.h
saatvikshah1994: Btw I wrote up a short blog post on this topic - more as a point of reference but [here it is](https://saatvikshah1994.github.io/TemplateOverloadResolution/) in case anyone is interested :)

src/shogun/lib/SGMatrix.h
saatvikshah1994: done

src/shogun/lib/SGMatrix.h
gf712: Ah true! Thanks, I'll have a look!

src/shogun/lib/SGMatrix.h
gf712: I just tried this out on my machine, and both methods seem to take the same amount of time, I am not sure there is a way to speed up either methods.

gf712: And btw, I wouldn't recommend this, but you could do this with `std::for_each`:
```
std::for_each(matrix, matrix+N, 
	[&r, &matrix](auto a){
		static int i = 0; 
		matrix[i] = r.random_half_open(); 
		++i; 
	});
```

src/shogun/lib/SGMatrix.h
karlnapf: Yep!

src/shogun/lib/SGMatrix.h
karlnapf: Ah cool!

src/shogun/lib/SGMatrix.h
karlnapf: Totally right!
If you want to have a stab at doing some of this in a better way, feel free to submit a draft PR for discussing it! Fixing the equals would be a start

tests/unit/lib/SGMatrix_unittest.cc
karlnapf: yeah don't worry then

src/shogun/lib/SGMatrix.h
```diff
@@ -306,9 +308,66 @@ template<class T> class SGMatrix : public SGReferencedData
...
-		/** fill matrix with zeros */
+		/** Fill matrix with zeros */
 		void zero();
 
+		/**
+		 * Fill matrix with uniformly distributed randoms
+		 * Float Types: [0, 1)
+		 * Int Types: std::numeric_limit<T>::(min->max)
+		**/
+		// vectorized available
```
karlnapf: i wonder whether doxygen correctly parses the above docstring if this comment is here?
could even just remove it

src/shogun/lib/SGMatrix.h
```diff
@@ -306,9 +308,66 @@ template<class T> class SGMatrix : public SGReferencedData
...
+		template <typename U>
+		using enable_simd_float = std::enable_if_t<is_any_of_v<U, float64_t>>;
+		template <typename U = T>
+		auto random() -> enable_simd_float<U>
+		{
+			CRandom r{};
```
karlnapf: one thing: we usually allocate sgobjects using `new` for several reasons. It is a style thing but let's adopt that here as well

src/shogun/lib/SGVector.h
```diff
@@ -204,6 +206,63 @@ template<class T> class SGVector : public SGReferencedData
 		/** Fill vector with zeros */
 		void zero();
 
+		/**
```
karlnapf: whitespace issues

src/shogun/lib/sg_type_traits.h
```diff
@@ -0,0 +1,36 @@
...
+#ifndef SHOGUN_SG_TYPE_TRAITS_H
+#define SHOGUN_SG_TYPE_TRAITS_H
+/**
+ * A collection of useful type traits
+ */
+
+// Checks if any one of the types matches
+// Ref: https://stackoverflow.com/a/17032517/3656081
+template <typename T, typename... Rest>
+struct is_any_of : std::false_type
```
karlnapf: maybe `sg_is_any_of` ? To make clear these are shogun traits?

src/shogun/lib/sg_type_traits.h
```diff
@@ -0,0 +1,36 @@
...
+};
+
+template <typename T, typename First>
+struct is_any_of<T, First> : std::is_same<T, First>
+{
+};
+
+template <typename T, typename First, typename... Rest>
+struct is_any_of<T, First, Rest...>
+    : std::integral_constant<bool, std::is_same<T, First>::value ||
```
karlnapf: btw I think I defined `is_sg_base` somewhere...we could move that into this file

karlnapf:
@iglesias maybe you also have comments

src/shogun/lib/SGMatrix.h
```diff
@@ -306,9 +308,66 @@ template<class T> class SGMatrix : public SGReferencedData
...
 		void zero();
 
+		/**
+		 * Fill matrix with uniformly distributed randoms
+		 * Float Types: [0, 1)
+		 * Int Types: std::numeric_limit<T>::(min->max)
+		**/
+		// vectorized available
+		template <typename U>
+		using enable_simd_float = std::enable_if_t<is_any_of_v<U, float64_t>>;
```
iglesias: Just out of curiosity, `std::is_same_v` would make it here as well instead of `is_any_of_v`, yes? No need to update anything.

iglesias:
Nice work @saatvikshah1994, thank you.

I was wondering, would you like to experiment with concepts to implement what you have done here with traits and sfinae in sgvector and sgmatrix? If you're interested we can discuss further what would be the possible advantages and gains. This could be done in a separate branch and discussed in another pull request.

karlnapf:
Will merge this tomorrow if no more input from @lisitsyn @vigsterkr thx!

src/shogun/lib/SGMatrix.h
karlnapf: any thoughts on this @saatvikshah1994 ?

src/shogun/lib/sg_type_traits.h
karlnapf: would still be good to rename or at least discuss

vigsterkr:
Hold your horses plz needs some changes....

src/shogun/lib/sg_type_traits.h
saatvikshah1994: Yep, I'll rename it.
Apologies about the delay! Recently got back from vacation :) - I'll address all the comments this weekend.

src/shogun/lib/SGMatrix.h
saatvikshah1994: done

src/shogun/lib/SGMatrix.h
saatvikshah1994: Sure, I could look up and try to submit a few more once this is done.

src/shogun/lib/SGMatrix.h
saatvikshah1994: Umm, sure, I was trying following the [guidelines](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rr-scoped).
Also I guess instead of `new` I would need to `std::make_unique` since we want to avoid `SG_UNREF`?

src/shogun/lib/SGMatrix.h
saatvikshah1994: Yep, std::transform is indeed equally fast(or even faster at times especially at lower optimization levels). This is done with `std::transform` - the `for_each` version looked a bit hacky :/

src/shogun/lib/SGMatrix.h
saatvikshah1994: done

src/shogun/lib/SGMatrix.h
saatvikshah1994: Yep it would, I just kept it for consistency. Also `is_same_v` isnt part of STL till C++17. So I would have to add that too.

saatvikshah1994:
> Nice work @saatvikshah1994, thank you.
> 
> I was wondering, would you like to experiment with concepts to implement what you have done here with traits and sfinae in sgvector and sgmatrix? If you're interested we can discuss further what would be the possible advantages and gains. This could be done in a separate branch and discussed in another pull request.

Sure, I'm game! Do you want me to create an issue and we can discuss there? Or on IRC first?

src/shogun/lib/sg_type_traits.h
saatvikshah1994: yep, ive moved it over

src/shogun/lib/SGMatrix.h
saatvikshah1994: I added a `sg_is_same_v` for now and have used that. Guess it might be useful at other places too and can be replaced by the STL equivalent when you'll decide to bump up the C++ version.

src/shogun/lib/SGMatrix.h
iglesias: Nice questions, @saatvikshah1994 :-)

I'd say the actual reasons are mostly due to history. Shogun has been around for long time, longer than what we consider today modern C++. Also, I personally think the focus and interests have also moved a bit from "hardcore" machine learning programming to software engineering.

It is of course always good to follow the latest trends and guidelines of C++. I've found however that as a project grows and gets bigger such redesigns and/or refactorings become too time consuming without having a clear value in terms of bringing new features. Technical debt.

iglesias: About your comment on `SGMatrix`'s API. Note that the main SG* guys, `SGMatrix` and `SGVector` are referenced data, so there's no need for handmade memory management with those ;-)

iglesias: About the initialization of `CRandom` that Heiko mentioned. It has to do with objects residing in the stack vs the heap. The rule-of-thumb I started using at some moment was that any `SGObject` has to be heap-allocated.

I forgot the exact technical reason. Stepping through a minimal program with gdb could help understanding. But anyway, very vaguely from what I can remember: it had to do with the reference counting mechanism taking place as scopes end.

Anyhow, I think should be able to follow the CPP core guidelines for this initialization: just wrap your `SGObject` inside a `shogun::Some` (aka `std::shared_ptr`).

iglesias:
Great to know!

Let me start an issue soon and I will mention you there. What I have in
mind is still a work in progress idea ;-)

In the meantime, are you familiar with concepts already? Otherwise I can
recommend a video or two from latest cppcon to get you started.

saatvikshah1994:
Not much no so would surely appreciate any starter tips.

iglesias:
Starters to get introduced to the main ideas:
https://www.youtube.com/watch?v=HddFGPTAmtU
https://en.wikipedia.org/wiki/Concepts_(C%2B%2B)

src/shogun/lib/SGMatrix.h
```diff
@@ -306,9 +309,60 @@ template<class T> class SGMatrix : public SGReferencedData
...
+		 * Fill matrix with uniformly distributed randoms
+		 * Float Types: [0, 1)
+		 * Int Types: std::numeric_limit<T>::(min->max)
+		**/
+		template <typename U>
+		using enable_simd_float = std::enable_if_t<sg_is_same_v<U, float64_t>>;
+		template <typename U = T>
+		auto random() -> enable_simd_float<U>
+		{
+			auto r = std::make_unique<CRandom>();
```
vigsterkr: the problem with having CRandom created here is that you wont ever be able to set seed :(
there's a way to do this, check the usage of the global variable `sg_rand` in the codebase... but i rather would start reviving the `feature/random-refactor` branch and have then this fixed....

vigsterkr: the rest of the patch has the same pattern and it'll be a problem as well there...

saatvikshah1994:
For `SGVector<float64>`: taking a vector of N size as denoted below. Seems like the gains of SIMD start showing only after 1024 elements. 
For `SGMatrix<float64>`: taking a matrix N x N of the N size denoted below.
I'm not very familiar with the algorithm used by `dSFMT`, but seems like tput proportionally improves as the number of elements increase until saturation at ~64K elements. 
In the final cases there is always a drop - my guess is due to first going L1->L2/L3->Memory seemingly.
```
CPU Caches:
  L1 Data 32K (x4)
  L1 Instruction 32K (x4)
  L2 Unified 262K (x4)
  L3 Unified 6291K (x1)
-----------------------------------------------------------------------
Benchmark                                Time           CPU Iterations
-----------------------------------------------------------------------
BM_SGVectorSimpleForRng/256          10686 ns       4824 ns     146526   50.6128MB/s
BM_SGVectorSimpleForRng/1024         31256 ns      19961 ns      33535   48.9237MB/s
BM_SGVectorSimpleForRng/4096        103190 ns      80179 ns       8864    48.719MB/s
BM_SGVectorSimpleForRng/16384       476489 ns     332658 ns       2109   46.9702MB/s
BM_SGVectorSimpleForRng/65536      1731183 ns    1341493 ns        533   46.5899MB/s
BM_SGVectorSimpleForRng/262144     7684643 ns    5338349 ns        129    46.831MB/s
BM_SGVectorSimpleForRng/1048576   28698992 ns   21948063 ns         32   45.5621MB/s
BM_SGVectorSimpleForRng/4194304  124266202 ns   91083250 ns          8   43.9159MB/s
BM_SGVectorSimpleForRng/8388608  183002067 ns  149759250 ns          4   53.4191MB/s
BM_SGVectorSimdRng/256               28129 ns      23222 ns      29822   10.5134MB/s
BM_SGVectorSimdRng/1024              22501 ns      21000 ns      33497   46.5021MB/s
BM_SGVectorSimdRng/4096              37701 ns      30281 ns      25527       129MB/s
BM_SGVectorSimdRng/16384             61909 ns      51976 ns      13567   300.621MB/s
BM_SGVectorSimdRng/65536            137449 ns     125798 ns       5051   496.827MB/s
BM_SGVectorSimdRng/262144           520684 ns     453210 ns       1561   551.621MB/s
BM_SGVectorSimdRng/1048576         1951931 ns    1801798 ns        410   555.001MB/s
BM_SGVectorSimdRng/4194304         8058928 ns    7512247 ns         93   532.464MB/s
BM_SGVectorSimdRng/8388608        15801137 ns   15328644 ns         45   521.899MB/s
BM_SGMatrixSimpleForRng/64           70855 ns      64603 ns      10725   60.4652MB/s
BM_SGMatrixSimpleForRng/128         255030 ns     249715 ns       2717   62.5713MB/s
BM_SGMatrixSimpleForRng/256        1022532 ns     996520 ns        712   62.7183MB/s
BM_SGMatrixSimpleForRng/512        4473191 ns    4014367 ns        177   62.2763MB/s
BM_SGMatrixSimpleForRng/1024      19003502 ns   16312333 ns         45   61.3033MB/s
BM_SGMatrixSimpleForRng/2048      67421134 ns   64075364 ns         11   62.4265MB/s
BM_SGMatrixSimpleForRng/4096     309963506 ns  303301000 ns          2   52.7529MB/s
BM_SGMatrixSimpleForRng/8192    1427790785 ns 1366948000 ns          1   46.8196MB/s
BM_SGMatrixSimdRng/64                22065 ns      21745 ns      31635    179.64MB/s
BM_SGMatrixSimdRng/128               38247 ns      37770 ns      18664   413.689MB/s
BM_SGMatrixSimdRng/256              108321 ns     106161 ns       6338   588.728MB/s
BM_SGMatrixSimdRng/512              378871 ns     369052 ns       1817   677.412MB/s
BM_SGMatrixSimdRng/1024            1519232 ns    1493987 ns        473    669.35MB/s
BM_SGMatrixSimdRng/2048            6198822 ns    6065946 ns        129   659.419MB/s
BM_SGMatrixSimdRng/4096           31814041 ns   31293111 ns         18   511.295MB/s
BM_SGMatrixSimdRng/8192          280005706 ns  274623000 ns          2   233.047MB/s
```

src/shogun/lib/SGMatrix.h
saatvikshah1994: ive addressed it for `SGMatrix`
`SGVector`/`CRandom` are being included in one another, which is causing a problem when using `sg_rand` in the header directly. I'll have a go at it tomorrow.

gf712:
That's really nice! Great work :D

iglesias:
> For `SGVector<float64>`: taking a vector of N size as denoted below. Seems like the gains of SIMD start showing only after 1024 elements.
> For `SGMatrix<float64>`: taking a matrix N x N of the N size denoted below.
> I'm not very familiar with the algorithm used by `dSFMT`, but seems like tput proportionally improves as the number of elements increase until saturation at ~64K elements.
> In the final cases there is always a drop - my guess is due to first going L1->L2/L3->Memory seemingly.
> 
> ```
> CPU Caches:
>   L1 Data 32K (x4)
>   L1 Instruction 32K (x4)
>   L2 Unified 262K (x4)
>   L3 Unified 6291K (x1)
> -----------------------------------------------------------------------
> Benchmark                                Time           CPU Iterations
> -----------------------------------------------------------------------
> BM_SGVectorSimpleForRng/256          10686 ns       4824 ns     146526   50.6128MB/s
> BM_SGVectorSimpleForRng/1024         31256 ns      19961 ns      33535   48.9237MB/s
> BM_SGVectorSimpleForRng/4096        103190 ns      80179 ns       8864    48.719MB/s
> BM_SGVectorSimpleForRng/16384       476489 ns     332658 ns       2109   46.9702MB/s
> BM_SGVectorSimpleForRng/65536      1731183 ns    1341493 ns        533   46.5899MB/s
> BM_SGVectorSimpleForRng/262144     7684643 ns    5338349 ns        129    46.831MB/s
> BM_SGVectorSimpleForRng/1048576   28698992 ns   21948063 ns         32   45.5621MB/s
> BM_SGVectorSimpleForRng/4194304  124266202 ns   91083250 ns          8   43.9159MB/s
> BM_SGVectorSimpleForRng/8388608  183002067 ns  149759250 ns          4   53.4191MB/s
> BM_SGVectorSimdRng/256               28129 ns      23222 ns      29822   10.5134MB/s
> BM_SGVectorSimdRng/1024              22501 ns      21000 ns      33497   46.5021MB/s
> BM_SGVectorSimdRng/4096              37701 ns      30281 ns      25527       129MB/s
> BM_SGVectorSimdRng/16384             61909 ns      51976 ns      13567   300.621MB/s
> BM_SGVectorSimdRng/65536            137449 ns     125798 ns       5051   496.827MB/s
> BM_SGVectorSimdRng/262144           520684 ns     453210 ns       1561   551.621MB/s
> BM_SGVectorSimdRng/1048576         1951931 ns    1801798 ns        410   555.001MB/s
> BM_SGVectorSimdRng/4194304         8058928 ns    7512247 ns         93   532.464MB/s
> BM_SGVectorSimdRng/8388608        15801137 ns   15328644 ns         45   521.899MB/s
> BM_SGMatrixSimpleForRng/64           70855 ns      64603 ns      10725   60.4652MB/s
> BM_SGMatrixSimpleForRng/128         255030 ns     249715 ns       2717   62.5713MB/s
> BM_SGMatrixSimpleForRng/256        1022532 ns     996520 ns        712   62.7183MB/s
> BM_SGMatrixSimpleForRng/512        4473191 ns    4014367 ns        177   62.2763MB/s
> BM_SGMatrixSimpleForRng/1024      19003502 ns   16312333 ns         45   61.3033MB/s
> BM_SGMatrixSimpleForRng/2048      67421134 ns   64075364 ns         11   62.4265MB/s
> BM_SGMatrixSimpleForRng/4096     309963506 ns  303301000 ns          2   52.7529MB/s
> BM_SGMatrixSimpleForRng/8192    1427790785 ns 1366948000 ns          1   46.8196MB/s
> BM_SGMatrixSimdRng/64                22065 ns      21745 ns      31635    179.64MB/s
> BM_SGMatrixSimdRng/128               38247 ns      37770 ns      18664   413.689MB/s
> BM_SGMatrixSimdRng/256              108321 ns     106161 ns       6338   588.728MB/s
> BM_SGMatrixSimdRng/512              378871 ns     369052 ns       1817   677.412MB/s
> BM_SGMatrixSimdRng/1024            1519232 ns    1493987 ns        473    669.35MB/s
> BM_SGMatrixSimdRng/2048            6198822 ns    6065946 ns        129   659.419MB/s
> BM_SGMatrixSimdRng/4096           31814041 ns   31293111 ns         18   511.295MB/s
> BM_SGMatrixSimdRng/8192          280005706 ns  274623000 ns          2   233.047MB/s
> ```

Very nice analysis and results! Nice to see the simd versions achieve up to about 10-fold improvement.

Just to be sure, the results take into account random variation, yes? That is, the code is run many times for each Simple/Simd and size scenario.

Also, can you relabel or clarify the names of the columns? :-) Actually just clarifying the difference between the two time columns should be enough. I am guessing the last column should just be something like 8 bytes times N (for SGVector, N^2 for Matrix) divided by the time in one of the other two columns.

vigsterkr:
This is the output of google’s microbenchmarking tool. :)

And yes the library runs the benchmark codes several times....

src/shogun/lib/SGMatrix.h
saatvikshah1994: done

saatvikshah1994:
@iglesias the column explanations can be found here: https://github.com/google/benchmark/issues/397#issuecomment-305906871
And yes, theres a fix needed in the bmark - I should be multiplying by sizeof(float64_t) in the `SetBytesProcessed` call everywhere.

src/shogun/lib/sg_type_traits.h
karlnapf: Hope you had a good vaccation, and happy new year :)

src/shogun/lib/SGMatrix.h
karlnapf: yes, thx for that, maybe (if not yet done) put a comment that indicated that this can be removed/replaced

karlnapf:
> @karlnapf Im currently blocked on the benchmark by #4452 .

noted! Looks like some sanitizer updates necessary....nice ! :)

src/shogun/lib/SGMatrix.h
karlnapf: nice!

karlnapf:
Really like the results table! Well well done! So useful and a great example of how we can make things faster with some clever thinking about low hanging fruits.

karlnapf:
> > > @karlnapf Im currently blocked on the benchmark by #4452 .
> > 
> > 
> > noted! Looks like some sanitizer updates necessary....nice ! :)
> 
> Actually, I was blocked on Google benchmark not building on Mac/Ubuntu. This has been resolved and hence the analysis table above. The sanitizer problem is a bit tangential and Ive worked around it by using a VM/ @gf712 's docker image. I think this PR is ready to go unless there are some pending comments?

Yes well there are @vigsterkr comments about the seed. Was there any discussion regarding that? I am not updated to the latest I guess

iglesias:
I didn’t see an explanation there about the two results. Maybe I missed
something.

Anyhow, it is not important.

iglesias:
The two time* results.

saatvikshah1994:
> @iglesias the column explanations can be found here: [google/benchmark#397 (comment)](https://github.com/google/benchmark/issues/397#issuecomment-305906871)
> And yes, theres a fix needed in the bmark - I should be multiplying by sizeof(float64_t) in the `SetBytesProcessed` call everywhere.

@iglesias had linked to an answer here by the Google benchmark devs about the time columns. In summary:
1. Time: Avg Wall time per iteration(This is generally a combination of time taken by the CPU + IO)
2. CPU: This is indicative of time taken by the CPU alone. 

Looking at (1) and (2) we can determine where we are IO bound and where CPU bound. 

And yes, what you've written about the last column is correct.

karlnapf:
Hi!
While the concepts PR is great to get going on this new pattern, I still would like to get this one here merged soon. There is the issue with the seed. Any comments on this anyone?

iglesias:
About the seed, I think that was already solved in the latest commit in the PR: 6a97846. See also comments above from @saatvikshah1994 saying it was done.

karlnapf:
Ok let's merge this then :) I will wait one more day and then do it if nobody else complains @lisitsyn @vigsterkr 

I think cool next steps would be:
* actually replace looped rng usage in shogun with this new API
* or write an entrance task for GSoC students to do this

vigsterkr:
@karlnapf i wouldnt put more effort into this for the time being as we wanna drop some of the PRNG stuff as they are flawed by design atm. see `feature/random-refactor`

karlnapf:
I guess the effort is already done? So this is still better than what we have atm. We could stop pushing this forward then of course. But I think can still go in as an example of the overload dispatching?

vigsterkr:
a) where is this code being used? meaning the SGVector/Matrix.random()
b) how do you set the seed of this?

karlnapf:
Alright, agreed actually

karlnapf:
@saatvikshah1994 maybe have a look at the feature branch to get a feeling for what @vigsterkr is talking about

saatvikshah1994:
Sure, I'll take a look

iglesias:
@vigsterkr @karlnapf please guys have a look at the commit 6a97846. I think that one is addressing the seed concern.

karlnapf:
@saatvikshah1994 thanks for the ping. The point beside the sees is that the whole design of the random class in shogun is flawed. This is why Viktor mentioned to not put any more effort into it. That said, we could of course still use your updates until we do a change, but the second point is that the uniform random number generator is not really used within shogun. This doesn't mean the PR is not useful. Quite the opposite is true, it is a really nice demonstration on how to gain speedups with vectorization as well as using better design patterns, so it is very useful for future reference. As said, for now, check the feature branch on how we can improve the random number generation in shogun in general. We should get the same ideas that you developed you in this PR into the feature branch and then put the efforts into pushing this forward rather than tuning deprecated designs.

gf712:
@vigsterkr is this equivalent to what @theartful did?

karlnapf:
Closing due as it is outdated and also we had some discussions whether we want to use this old design in the past

PR #4438, opened 2018-12-16
Fix remaining Neural Layer Tests

saatvikshah1994:
1. Replaced redundancies using fixture class and its helper methods.
2. Replaced manual for loops with shogun::linalg calls wherever feasible.

### DIFF
--- a/tests/unit/neuralnets/NeuralLeakyRectifiedLinearLayer_unittest.cc
+++ b/tests/unit/neuralnets/NeuralLeakyRectifiedLinearLayer_unittest.cc
@@ -30,75 +30,61 @@
  *
  * Written (W) 2015 Sanuj Sharma
  */
-
-#include <shogun/neuralnets/NeuralLeakyRectifiedLinearLayer.h>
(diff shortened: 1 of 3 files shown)


tests/unit/neuralnets/NeuralLeakyRectifiedLinearLayer_unittest.cc
```diff
@@ -30,75 +30,61 @@
  *
  * Written (W) 2015 Sanuj Sharma
  */
-
```
karlnapf: add your name if you like

saatvikshah1994:
Yep no memory errors here.

tests/unit/neuralnets/NeuralLeakyRectifiedLinearLayer_unittest.cc
saatvikshah1994: its fine - i didnt really add any logic just removed boilerplate

karlnapf:
As usual, let’s watch out for the ci abd buildbot

PR #4440, opened 2018-12-17
Removed some uses of m_model_selection_parameters [WIP]

avramidis:
This PR is part of the [Parameter.cpp](https://github.com/shogun-toolbox/shogun/projects/6) project.

In this PR we removed m_model_selection_parameters from SGObject. Also, in the future m_parameters and m_gradient_parameters will be removed. These store models' (hyper)parameters which are estimated during the training phase.

The aim is to use the AnyParameter properties to store the value and properties of these parameters. PRs #4412 and #4426 include code with the first steps towards this aim.

### DIFF
--- a/examples/undocumented/libshogun/parameter_modsel_parameters.cpp
+++ b/examples/undocumented/libshogun/parameter_modsel_parameters.cpp
@@ -23,8 +23,6 @@ void print_message(FILE* target, const char* str)
 
 void print_modsel_parameters(CSGObject* object)
 {
-	SGStringList<char> modsel_params=object->get_modelsel_names();
-
 	SG_SPRINT("Parameters of %s available for model selection:\n",
 			object->get_name());
 
@@ -34,8 +32,6 @@ void print_modsel_parameters(CSGObject* object)
 		/* extract current name, ddescription and type, and print them */
 		const char* name=modsel_params.strings[i].string;
 		index_t index=object->get_modsel_param_index(name);
-		TSGDataType type=object->m_model_selection_parameters->get_parameter(
-				index)->m_datatype;
 		type.to_string(type_string, 100);
 		SG_SPRINT("\"%s\": \"%s\", %s\n", name,
 				object->get_modsel_param_descr(name), type_string);
--- a/src/shogun/base/SGObject.cpp
+++ b/src/shogun/base/SGObject.cpp
@@ -46,6 +46,8 @@
 #include <shogun/multiclass/ecoc/ECOCDecoder.h>
 #include <shogun/multiclass/MulticlassStrategy.h>
 
+#include <shogun/base/init.h>
+#include <shogun/lib/type_case.h>
 
 namespace shogun
 {
@@ -204,7 +206,6 @@ CSGObject::~CSGObject()
 
 	unset_global_objects();
 	delete m_parameters;
-	delete m_model_selection_parameters;
 	delete m_gradient_parameters;
 	delete m_refcount;
 	delete m_subject_params;
@@ -520,7 +521,6 @@ void CSGObject::init()
 	parallel = NULL;
 	version = NULL;
 	m_parameters = new Parameter();
-	m_model_selection_parameters = new Parameter();
 	m_gradient_parameters=new Parameter();
 	m_generic = PT_NOT_GENERIC;
 	m_load_pre_called = false;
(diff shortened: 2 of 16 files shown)


tests/unit/machine/gp/GaussianLikelihood_unittest.cc
```diff
@@ -268,8 +268,6 @@ TEST(GaussianLikelihood,get_first_derivative)
 	// Gaussian likelihood with sigma = 0.13
 	CGaussianLikelihood* likelihood=new CGaussianLikelihood(0.13);
 
-	TParameter* param=likelihood->m_model_selection_parameters->get_parameter("log_sigma");
-
 	SGVector<float64_t> lp_dhyp=likelihood->get_first_derivative(labels, func,
 			param);
```
gf712: I am assuming the compiler complains that `param` isn't declared here right?

src/shogun/base/SGObject.h
```diff
@@ -92,8 +92,6 @@ template <class T> class SGStringList;
 		    AnyParameterProperties(description, param_properties);             \
 		this->m_parameters->add(param, name, description);                     \
 		this->watch_param(name, param, pprop);                                 \
-		if (pprop.get_model_selection())                                       \
-			this->m_model_selection_parameters->add(param, name, description); \
 		if (pprop.get_gradient())                                              \
 			this->m_gradient_parameters->add(param, name, description);        \
 	}
```
gf712: Did you try to remove `m_gradient_parameters` yet? Just curious as to how many things break..

avramidis: No, I didn't remove m_gradient_parameters. Mostly the model selection functionality breaks.

avramidis:
The errors I get is mostly complains from the compiler that the m_model_selection_parameters cannot be found.

gf712:
OK! I think the idea is to now use `CSGObject::get_parameter` instead, and from there you can extract (maybe cast?) the value of the parameter. But I guess we'll discuss that tomorrow!

examples/undocumented/libshogun/parameter_modsel_parameters.cpp
```diff
@@ -23,8 +23,6 @@ void print_message(FILE* target, const char* str)
 
 void print_modsel_parameters(CSGObject* object)
```
karlnapf: we can just drop this function

karlnapf: actually, we can just drop the whole cpp example ... to be replaced with a meta example

Actually 
@gf712 could you try to write a meta example for `base_api` where you show how to extract model selection parameters at some point? mayber after the monkey business is done

src/shogun/base/SGObject.cpp
```diff
@@ -657,27 +573,6 @@ void CSGObject::get_parameter_incremental_hash(uint32_t& hash, uint32_t& carry,
 	}
 }
 
-void CSGObject::build_gradient_parameter_dictionary(CMap<TParameter*, CSGObject*>* dict)
```
karlnapf: this we need to keep working somehow as it is used for the GP framework ... so a replacement needs to be written
The signature will probably change as `TParameter` will be removed and we will use the dispatching code that @gf712 wrote

src/shogun/modelselection/GridSearchModelSelection.cpp
```diff
@@ -68,9 +68,6 @@ CParameterCombination* CGridSearchModelSelection::select_model(bool print_state)
 			current_combination->print_tree();
 		}
 
-		current_combination->apply_to_modsel_parameter(
```
karlnapf: this is the big fish to replace

karlnapf: Could you, instead of removing the method call, just keep it and place a runtime error (SG_NOTIMPLEMENTED) inside the method?

src/shogun/modelselection/ParameterCombination.cpp
```diff
@@ -700,7 +698,7 @@ CParameterCombination* CParameterCombination::copy_tree() const
 
 void CParameterCombination::apply_to_machine(CMachine* machine) const
 {
-	apply_to_modsel_parameter(machine->m_model_selection_parameters);
```
karlnapf: same here. The danger with just removing this line is that we forget to re-add it later when we have a replacement. So rather than changing behaviour of code, let's just put a runtime error

src/shogun/modelselection/ParameterCombination.cpp
```diff
@@ -749,8 +747,6 @@ void CParameterCombination::apply_to_modsel_parameter(
 			{
 				CParameterCombination* child=(CParameterCombination*)
 						m_child_nodes->get_element(i);
-				child->apply_to_modsel_parameter(
```
karlnapf: and here

tests/unit/machine/gp/GaussianLikelihood_unittest.cc
```diff
@@ -268,8 +268,6 @@ TEST(GaussianLikelihood,get_first_derivative)
 	// Gaussian likelihood with sigma = 0.13
 	CGaussianLikelihood* likelihood=new CGaussianLikelihood(0.13);
 
-	TParameter* param=likelihood->m_model_selection_parameters->get_parameter("log_sigma");
```
karlnapf: Looking at the test, this can probably already be replaced with `get`

tests/unit/machine/gp/StudentsTLikelihood_unittest.cc
```diff
@@ -270,9 +270,6 @@ TEST(StudentsTLikelihood,get_first_derivative)
 	// Stundent's-t likelihood with sigma = 0.17, df = 3
 	CStudentsTLikelihood* likelihood=new CStudentsTLikelihood(0.17, 3);
 
-	TParameter* param1=likelihood->m_model_selection_parameters->get_parameter("log_sigma");
```
karlnapf: in all of those actually, `get` should do it

avramidis:
A very dirty update with mistakes. The next step would be to remove Parameter, ParameterCombination ,however, many stuff will break. I will have to change (& simplify) the hyper-parameter optimisation which I prefer to do.

src/shogun/modelselection/ModelSelectionParameter.h
```diff
@@ -0,0 +1,62 @@
...
+#define __MODELSELECTIONPARAMETER_H_
+
+#include <shogun/lib/config.h>
+
+#include <shogun/base/SGObject.h>
+#include <shogun/base/AnyParameter.h>
+
+namespace shogun
+{
+
```
karlnapf: I am not sure I agree with this idea of making a class for modelselection parameters....we will have to discuss a bit here.
But actually, as discussed a few day back in person, let's maybe start with the gradient parameter API refactoring

gf712:
this has already been done

PR #4441, opened 2018-12-17
single get method for python interface

gf712:
Found this [thread](https://github.com/swig/swig/issues/723) where they discuss how to add a Python method (written in python) dynamically even when using swig's -builtins. The python code then checks which CSGObject getter works and returns the one that works. If the parameter does not exist it returns the same `RuntimeError` as we would get with the other getters.

I tested this with Python 3.6, but it should work fine with any other Python 3 version, and should work with Python 2.7.

### DIFF
--- a/examples/meta/generator/targets/python.json
+++ b/examples/meta/generator/targets/python.json
@@ -56,7 +56,12 @@
         "RealLiteral": "$number",
         "FloatLiteral": "$number",
         "MethodCall": {
-            "Default": "$object.$method($arguments)"
+            "Default": "$object.$method($arguments)",
+            "get_int": "$object.get($arguments)",
+            "get_int_vector": "$object.get($arguments)",
+            "get_real": "$object.get($arguments)",
+            "get_real_vector": "$object.get($arguments)",
+            "get_real_matrix": "$object.get$arguments)"
         },
         "StaticCall": "$typeName.$method($arguments)",
         "GlobalCall": "$method($arguments)",
--- a/examples/meta/generator/targets/python.json
+++ b/examples/meta/generator/targets/python.json
@@ -61,7 +61,7 @@
             "get_int_vector": "$object.get($arguments)",
             "get_real": "$object.get($arguments)",
             "get_real_vector": "$object.get($arguments)",
-            "get_real_matrix": "$object.get$arguments)"
+            "get_real_matrix": "$object.get($arguments)"
         },
         "StaticCall": "$typeName.$method($arguments)",
         "GlobalCall": "$method($arguments)",
--- a/examples/undocumented/python/structure_multiclass_bmrm.py
+++ b/examples/undocumented/python/structure_multiclass_bmrm.py
@@ -49,7 +49,7 @@ def structure_multiclass_bmrm(fm_train_real=traindat,label_train_multiclass=labe
 	bmrm_out = sosvm.apply()
 	count = 0
 	for i in range(bmrm_out.get_num_labels()):
-		yi_pred = bmrm_out.get_real_vector("labels_vector")[i]
+		yi_pred = bmrm_out.get("labels_vector")[i]
 		if yi_pred == label_train_multiclass[i]:
 			count = count + 1
 
--- a/examples/undocumented/python/structure_multiclass_bmrm.py
+++ b/examples/undocumented/python/structure_multiclass_bmrm.py
@@ -68,7 +68,7 @@ def structure_multiclass_bmrm(fm_train_real=traindat,label_train_multiclass=labe
 	ppbmrm_out = sosvm.apply()
 	count = 0
 	for i in range(ppbmrm_out.get_num_labels()):
-		yi_pred = ppbmrm_out.get_real_vector("labels_vector")[i]
+		yi_pred = ppbmrm_out.get("labels_vector")[i]
 		if yi_pred == label_train_multiclass[i]:
 			count = count + 1
 
@@ -84,7 +84,7 @@ def structure_multiclass_bmrm(fm_train_real=traindat,label_train_multiclass=labe
 	p3bmrm_out = sosvm.apply()
 	count = 0
 	for i in range(p3bmrm_out.get_num_labels()):
-		yi_pred = p3bmrm_out.get_real_vector("labels_vector")[i]
+		yi_pred = p3bmrm_out.get("labels_vector")[i]
 		if yi_pred == label_train_multiclass[i]:
 			count = count + 1
 
--- a/src/interfaces/swig/SGBase.i
+++ b/src/interfaces/swig/SGBase.i
@@ -474,47 +474,20 @@ namespace shogun
 %pythoncode %{
 def _internal_get_param(self, name):
 
-	try:
-		return self._get(name)
-	except SystemError:
-		pass
-	except Exception:
-		raise
-
-	try:
-		return self._get_real(name)
-	except RuntimeError:
-		pass
-	except Exception:
-		raise
-
-	try:
-		return self._get_int(name)
-	except RuntimeError:
-		pass
-	except Exception:
-		raise
-
-	try:
-		return self._get_real_matrix(name)
-	except RuntimeError:
-		pass
-	except Exception:
-		raise
-
-	try:
-		return self._get_real_vector(name)
-	except RuntimeError:
-		pass
-	except Exception:
-		raise
-
-	try:
-		return self._get_int_vector(name)
-	except RuntimeError:
-		raise KeyError("There is no parameter called '{}' in {}".format(name, self.get_name()))
-	except Exception:
-		raise
+	for f in (self._get,
+			self._get_real,
+			self._get_int,
+			self._get_real_matrix,
+			self._get_real_vector,
+			self._get_int_vector
+			):
+		try:
+			return f(name)
+		except (SystemError, RuntimeError):
+			pass
+		except Exception:
+			raise
+	raise KeyError("There is no parameter called '{}' in {}".format(name, self.get_name()))
 
 _swig_monkey_patch(SGObject, "get", _internal_get_param)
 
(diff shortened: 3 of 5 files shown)


lisitsyn:
I was wondering if it could be some different kind of exception. Does it really has to be RuntimeError?

gf712:
No, it was just for consistency with the current exception. Maybe something like a ValueError would make more sense?

lisitsyn:
Or even `KeyError` as it is some kind of map. I am not sure if it makes sense to change now but I'd definitely change it at some point.

gf712:
True, `KeyError` might be better to indicate why the user is getting an error.

gf712:
Is it worth catching all other errors like this:
```
try:
    return self.get_real(name)
except RuntimeError:
    pass
except Exception:
    raise
```

src/interfaces/swig/SGBase.i
```diff
@@ -446,6 +469,31 @@ namespace shogun
     }
 }
 
+%pythoncode %{
+	def _internal_get_param(self, name):
```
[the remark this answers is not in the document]
gf712: Yes, something similar might be possible with other interpreted languages right?

src/interfaces/swig/SGBase.i
```diff
@@ -446,6 +469,31 @@ namespace shogun
...
+		try:
+			return self.get_real_matrix(name)
+		except RuntimeError:
+			pass
+		try:
+			return self.get_real_vector(name)
+		except RuntimeError:
+			pass
+		try:
+			return self.get_int_vector(name)
```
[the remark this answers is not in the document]
gf712: Ah true! Is the name for the method OK btw?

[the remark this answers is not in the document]
karlnapf: idk...eventually, we just want "get" to be visible to the outside....so I would even prefer not to have the dummy method but just "get"

src/interfaces/swig/SGBase.i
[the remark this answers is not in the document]
karlnapf: definitely, but let's first start with python to see how it goes
Such a nice API simplification.
BTW once this is merged, you can change the [json](https://github.com/shogun-toolbox/shogun/blob/develop/examples/meta/generator/targets/cpp.json#L97) that does the `get_real_vector` names and unify it to a simple get. Then we have only moved the place where "real" "real_vector", etc are explicitly stated, rather than added this at another place in the codebase

src/interfaces/swig/SGBase.i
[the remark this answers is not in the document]
gf712: OK, we can "hide" the other getters by adding a `_` at some point?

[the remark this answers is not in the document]
karlnapf: Check shogun.i where all those methods are defined. We can modify things there

src/interfaces/swig/SGBase.i
```diff
@@ -307,7 +330,7 @@ public void readExternal(java.io.ObjectInput in) throws java.io.IOException, jav
 %ignore sg_print_error;
 %ignore sg_cancel_computations;
 
-
+%rename(_get) get(const std::string&);
```
vigsterkr: guard this for the time being with an
```
#ifdef SWIGPYTHON
#endif
```

examples/meta/generator/targets/python.json
```diff
@@ -56,7 +56,12 @@
         "RealLiteral": "$number",
         "FloatLiteral": "$number",
         "MethodCall": {
-            "Default": "$object.$method($arguments)"
+            "Default": "$object.$method($arguments)",
+            "get_int": "$object.get($arguments)",
+            "get_int_vector": "$object.get($arguments)",
+            "get_real": "$object.get($arguments)",
+            "get_real_vector": "$object.get($arguments)",
+            "get_real_matrix": "$object.get$arguments)"
```
lisitsyn: Typo!

src/interfaces/swig/SGBase.i
```diff
@@ -447,6 +472,52 @@ namespace shogun
 }
 
 %pythoncode %{
+def _internal_get_param(self, name):
+
+	try:
+		return self._get(name)
```
lisitsyn: Can we have something like that?
```
try:
  for f in (
    self._get,
    self._get_real
  ):
    f(name)
except:
  ..
```

examples/meta/generator/targets/python.json
gf712: Thanks!

src/interfaces/swig/SGBase.i
gf712: Yes, done!

gf712: Ah wait, that's a good point putting the loop in the try except block

examples/undocumented/python/structure_multiclass_bmrm.py
```diff
@@ -68,7 +68,7 @@ def structure_multiclass_bmrm(fm_train_real=traindat,label_train_multiclass=labe
 	ppbmrm_out = sosvm.apply()
 	count = 0
 	for i in range(ppbmrm_out.get_num_labels()):
-		yi_pred = ppbmrm_out.get_real_vector("labels_vector")[i]
+		yi_pred = ppbmrm_out.get("labels_vector")[i]
```
karlnapf: very nice that you are fixing those things on the fly as well.
It might be worth grepping for those in the ipython notebooks as well

src/interfaces/swig/SGBase.i
```diff
@@ -237,6 +237,29 @@ public void readExternal(java.io.ObjectInput in) throws java.io.IOException, jav
         }
 %}
 
+// taken from https://github.com/swig/swig/issues/723#issuecomment-230178855
+%typemap(out) void _swig_monkey_patch "$result = PyErr_Occurred() ? NULL : SWIG_Py_Void();"
+%inline %{
+    static void _swig_monkey_patch(PyObject *type, PyObject *name, PyObject *object) {
+        if (PyType_Check(type)) {
+#if PY_VERSION_HEX>=0x03000000
```
karlnapf: omg how lovely :D

karlnapf: what is the perspective on this bit here?

src/interfaces/swig/SGBase.i
```diff
@@ -447,6 +472,25 @@ namespace shogun
...
+	for f in (self._get,
+			self._get_real,
+			self._get_int,
+			self._get_real_matrix,
+			self._get_real_vector,
+			self._get_int_vector
+			):
+		try:
+			return f(name)
+		except (SystemError, RuntimeError):
```
karlnapf: why two errors? What is the context of each?

gf712: self._get returns a SystemError and the other getters return a RuntimeError

gf712: self._get returns a SystemError and the other getters return a RuntimeError

src/interfaces/swig/SGBase.i
```diff
@@ -447,6 +472,25 @@ namespace shogun
...
+			pass
+		except Exception:
+			raise
+	raise KeyError("There is no parameter called '{}' in {}".format(name, self.get_name()))
```
karlnapf: I would prefer if we would somehow connect the error msg with the ones present in the c++ lib, where this error already exists. Further, if done as here, the user should be informed that this error is within the interface rather than core shogun.

src/interfaces/swig/SGBase.i
karlnapf: You know why that is?

karlnapf: And also the question: is that good? Maybe we should change that?

gf712: I will have a better look during the week, but it seems that there are multiple definitions of how exceptions are handled in `swig_typemaps.i`. I will see if I can simplify this!

src/interfaces/swig/SGBase.i
gf712: I think for this purpose it is fine, but in general I don't think it is advisable to change the behaviour of software at runtime... This just seems like the simplest solution without messing with the C++ code.

src/interfaces/swig/SGBase.i
karlnapf: ok, but seperate PR

karlnapf:
Merging, since CI is happy. Let's nevertheless monitor the buildbot

karlnapf:
I actually have a request for something that I think might be cool to have. The final goal would be to actually allow
```
kernel = sg.kernel("GaussianKernel", log_width=1)
```

Currently, we do that in the meta examples, but then translate it to 
```
kernel.put("log_width" = 1)
```
which becomes clumsy for multiple parameters.

If we could add to SWIG
```
def _internal_factory(sef, name, **kwargs):
	"""
	A generic factory that accept kwargs and passes them to shogun via .put
	"""
	obj = getattr(shogun, name)()
	for k,v in kwargs.items():
		obj.put(k, v)
```

and then monkey patch the factory methods, (here `kernel`) that would be much neater. I.e. have a single factory method in Python that passes things on to shogun and calls the factory method with the appropriate name.

Now this doesnt work obviously
```
_swig_monkey_patch(SGObject, "kernel", _internal_factory)
```
as `kernel` is a global function. But I feel this could be easily modified to achieve what I described above? This would further simplify the python interface

gf712:
OK! That would be good. And the only way to do this would be by adding some Python code then?

karlnapf:
Yes, we could use something like https://github.com/lisitsyn/stichwort later on, but for python since we wanna leverage the `**kwargs` we would need some python code. The key here is that we then only offer a single factory that then calls the other ones by the string and `getattr` in order to avoid having to add code for every single factory

karlnapf:
I can do this, but I have little idea how to modify the monkey business to override global functions

karlnapf:
@gf712 as this is API related, could you update the appropriate section in the NEWS file? (with a PR). Something like: added single `get` with variable return types for accessing fields in Python

Issue #4431, opened 2018-12-13
When should we prioritise using the STL algorithms?


gf712:
I am currently writing some new print functions for SGObjects and I need to iterate over a map. There are several ways to do this now. The two ways I would use is a range iterator in a for loop or `std::for_each` with a lambda. In my case this would look like this:
- For loop:
```
for (const auto &it: self->get_map())
{
    auto tag = it.first;
    auto any_pprop = it.second.get_properties();

    if (any_pprop.has_property(pprop)) {
        auto type = demangled_type(it.second.get_value().type_info().name());
        SG_PRINT("\t%s (%s): %s\n", tag.name().c_str(), any_pprop.get_description().c_str(),
                 type.c_str());
        num_param++;
        }
}
```  
- `std::for_each`
```
std::for_each(self->map.begin(), self->map.end(),
              [&pprop, &num_param, &io=io](const auto& it){
                  auto tag = it.first;
                  auto any_pprop = it.second.get_properties();

                  if (any_pprop.has_property(pprop)) {
                      auto type = demangled_type(it.second.get_value().type_info().name());
                      SG_PRINT("\t%s (%s): %s\n", tag.name().c_str(), any_pprop.get_description().c_str(),
                               type.c_str());
                      num_param++;
                  }
              });
```
They are both the same, but the lambda has its own scope which is a bit awkward, as we need to remember to capture `io` when using `SG_PRINT` (or capture `this` by value).
My question is: is it worth always sticking to one over the other, particularly inside the same class, or just do what looks more readable at the time? It would be nice to be consistent, no?

karlnapf:
bringing the io in is awekward indeed. I think this issue is just for printing though, so despite the fact that we usually try to stick to the STL recently, I think you can do an explicit loop here. Or you just bring the io into the scope.
BTW why do you want to print? This should not be necessary as most user API should just return strings (which is supported by swig)

gf712:
This would be a way to replace `print_modsel_params` with something more generic like `print_params(ParameterProperties pprop)`, since we would now have implement a `print_gradient_params` and so on, no?

karlnapf:
Just drop this. We can have the string returning method with a filter and potentially some ready packaged cases for things like hyperparamters

karlnapf:
I think it is best if the caller prints for himself. Think IPython/swig where this happens anyways, and in C++ the caller can then choose where and how to print. Could even use `sg_io` but the point is that there is choice and we don't need to manage this so less code

[...]

PR #4442, opened 2018-12-17
Simple attempt to anyparameters to get model selection parameters info

avramidis:
I am not sure if this works properly because I use Gil's PR before it was merged. I just wanted to check anyparameters for a bit.

### DIFF
--- a/src/shogun/base/SGObject.cpp
+++ b/src/shogun/base/SGObject.cpp
@@ -46,6 +46,9 @@
 #include <shogun/multiclass/ecoc/ECOCDecoder.h>
 #include <shogun/multiclass/MulticlassStrategy.h>
 
+#include <shogun/base/init.h>
+#include <shogun/lib/type_case.h>
+
 
 namespace shogun
 {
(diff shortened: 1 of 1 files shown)


gf712:
For that example I wrote this:
```
for (const auto &it: self->get_map())
{
    auto tag = it.first;
    auto any_pprop = it.second.get_properties();

    if (any_pprop.has_property(pprop)) {
        auto type = demangled_type(it.second.get_value().type_info().name());
        SG_PRINT("\t%s (%s): %s\n", tag.name().c_str(), any_pprop.get_description().c_str(),
                 type.c_str());
        num_param++;
        }
}
```
But I think the plan is to drop the print, see #4431 and #4432.
The `sg_for_each_type` (now probably going to be `sg_any_dispath`) is a bit of an overkill to just print the type, because Any has a type_info() method that returns the std::type_info of the "underlying" type. I think `sg_any_dispath` will be used in very specific cases for loggers like tensorboard where we actually need to cast Any.

avramidis:
Oh, yes, I agree. I wrote this (I think) a couple of weeks ago and pushed it in case it is useful.

karlnapf:
Yeah let's just drop the whole print function and instead return an std::vector of strings. I think @gf712 you have prepared something for that somewhere right?

PR #4444, opened 2018-12-20
NOT A PR

gf712:
trying to find out which test causes memory leaks in azure because on my machine I don't have any issues


karlnapf:
Lol .... remember you might maybe also use docker to get the same environment...did that a few times in the past to reproduce such errors

gf712:
True.. I think it's the static initialisation of maps that can cause some odd behaviour in certain compilers. I also noticed that we use gcc-6.3 in CI, might be worth upgrading to 6.5 no?

gf712:
The issue does seem to be the initialisation of the maps. I will fix this with a proper PR!

PR #4448, opened 2018-12-21
Static initialisation of maps

gf712:
This PR fixes a bug where creating maps in the header file was causing undefined behaviour. The `static` keyword guarantees that the maps are allocated only once. The `const` should allow the compiler to optimise it a bit more.


gf712:
OK! Should be fine now, it worked fine with the docker container. Btw, docker on Mac is very slow compared to ubuntu!

Issue #4439, opened 2018-12-17
Valgrind errors in multiple unit tests


saatvikshah1994:
On the current latest `develop` branch, when running ` valgrind --leak-check=full  ./shogun-unit-test --gtest_filter=*NeuralInputLayer*`(or other filters) Im getting thousands of errors all of which seem to be complaining `Conditional jump or move depends on uninitialised value(s)`. 

When running the same on a version from one month back(commit hash `f6efd89467a7016be8a50bf3f7c0fdf32b075369`) theres no errors.

Here is the valgrind output log:
[valgrind_out.txt](https://github.com/shogun-toolbox/shogun/files/2684720/valgrind_out.txt)

gf712:
Hi, which compiler did you use to compile shogun? I just did the same check and did not get any errors.

saatvikshah1994:
I basically setup a CLion project on my Ubuntu 18.04 machine and tested it on the executable generated there. In that case the above errors show.
But when I build it in a separate build directory there are no errors.  
The CXX compiler used seems to be the same. Since Im not able to reproduce it - I guess it can be closed?

gf712:
What are the compile flags passed by your CLion via cmake? I think it might be worth investigating it now, considering it is a recent change that causes that error and we can narrow it down.

karlnapf:
Many thanks for reporting. Can you track down the exact change that started causing this?

saatvikshah1994:
I thought might be a one off thing, but after deleting and rebuilding valgrind still reports the errors. That said, here is what the cmake configuration step looks like with the CLion build:
[cmake-build-out.txt](https://github.com/shogun-toolbox/shogun/files/2693183/cmake-build-out.txt)

karlnapf:
again, if we want to find out what is going on here, it might be best to find out what commit (or even better line changed) started causing those issues ... whether they are real issues, and what the best solution would be

saatvikshah1994:
Sure, I can try to find out where its failing - Regarding whether its a real issue I'm not sure how to figure out. It does seem awkward that the error only occurs when building with CLion.
A few other observations:
1. This warning always creeps up right at the start: `==25464== Warning: set address range perms: large range [0x8f21000, 0x24c75000) (defined)`.
2. I'm running `valgrind` from a Parallels Virtual machine.

saatvikshah1994:
[...]

PR #4451, opened 2019-01-04
Fixed uninitialised attributes in AnyParameterProperties

gf712:
This fixes issues with the copy constructor. Bugs probably only occurred when compiling in Debug mode (see #4439).


karlnapf:
Thx for this. Seems like the CI is working pre merge now. I will merge this anyways as only a minor change.

PR #4456, opened 2019-01-09
add windows c# ci job

vigsterkr:


gf712:
@vigsterkr is this still relevant?

vigsterkr:
@gf712 i'd say once #4719 is merged/fixed this could be picked up as actually having c# stuff tested on windows would be nice

PR #4467, opened 2019-01-14
Started updating NEWS for 7.0.0 release

gf712:
- [ ] date
- [ ] versions
- [ ] main features (simplified interfaces?)


karlnapf:
It is actually a really good idea to have a PR open with this. We can then all populate this over time...Let's keep it open and add things, usually it takes a few days to make the list ...

karlnapf:
Fixed a memory copying issue in SGVector that affected the correctness of a recent refactor of LinearRidgeRegression [Kirill Kolkir]

Issue #4461, opened 2019-01-13
Add kwargs style factories in Python


karlnapf:

karlnapf:
@sorig Check it out we now have shogun API for `my_kernel = sg.kernel("GaussianKernel", log_width=1.0)`. One thing that would be really awesome was if the meta language examples would make use of this. Currently, it is not possible to modify say `python.json` in order to produce the correct output. E.g.
```
        "KeywordArguments": {
            "List": ", $elements",
            "Element": "$keyword = $expr",
            "Separator": ", ",
            "InitialSeperatorWhenArgs>0": false
        },
```
produces things like
```
svm = machine("LibSVM"), C1 = 1.1, kernel = k
```
Any change you could also have a quick hack on that? 

Obviously, for the other langs, we don't want to change things (i.e. still go with the put cascades for now).

gf712:
Maybe could add a hardcoded list of keywords to ignore? 
I was thinking maybe with a comment in front of the statement in the .sg file (e.g. `//ignore_kwargs:python`), but that would probably make it harder to implement and probably more tedious to maintain?

karlnapf:
I think this is a simple matter of changing the generator a bit so that it can generate the python style kwargs .... let's see what esben makes of it :)

karlnapf:
Cool stuff, I should have thought that you did that :) thanks!

PR #4468, opened 2019-01-14
simplified object instantiation in Python API

gf712:
This PR addresses #4461. 
Still need to think a bit more about docstrings and maybe error handling.
@karlnapf Is this what you were thinking (more or less)?

### DIFF
--- a/src/interfaces/python/swig_typemaps.i
+++ b/src/interfaces/python/swig_typemaps.i
@@ -1292,5 +1292,30 @@ TYPEMAP_SPARSEFEATURES_OUT(PyObject,      NPY_OBJECT)
     PyErr_SetString(PyExc_SystemError, $1.what());
     SWIG_fail;
 }
+%rename(_kernel) kernel;
+
+%pythoncode %{
+def _internal_factory_wrapper(object_name, new_name, docstring=None):
+    """
+    A wrapper that returns a generic factory that
+    accepts kwargs and passes them to shogun.object_name
+    via .put
+    """
+    _obj = getattr(_shogun, object_name)
+    def _internal_factory(name, **kwargs):
+
+        new_obj = _obj(name)
+        for k,v in kwargs.items():
+            new_obj.put(k, v)
+            return new_obj
+    if docstring:
+        _internal_factory.__doc__ = docstring
+    else:
+        _internal_factory.__doc__ = _obj.__doc__.replace(object_name, new_name)
+
+    return _internal_factory
+
+kernel = _internal_factory_wrapper("_kernel", "kernel")
+%}
 
 #endif /* HAVE_PYTHON */
--- a/src/interfaces/python/swig_typemaps.i
+++ b/src/interfaces/python/swig_typemaps.i
@@ -1307,7 +1307,7 @@ def _internal_factory_wrapper(object_name, new_name, docstring=None):
         new_obj = _obj(name)
         for k,v in kwargs.items():
             new_obj.put(k, v)
-            return new_obj
+        return new_obj
     if docstring:
         _internal_factory.__doc__ = docstring
     else:


karlnapf:
Yes, thanks a lot for addressing this so quickly. Checking now

src/interfaces/python/swig_typemaps.i
```diff
@@ -1292,5 +1292,30 @@ TYPEMAP_SPARSEFEATURES_OUT(PyObject,      NPY_OBJECT)
...
+%rename(_kernel) kernel;
+
+%pythoncode %{
+def _internal_factory_wrapper(object_name, new_name, docstring=None):
+    """
+    A wrapper that returns a generic factory that
+    accepts kwargs and passes them to shogun.object_name
+    via .put
+    """
+    _obj = getattr(_shogun, object_name)
```
karlnapf: maybe add a comment that says "Instantiate object via shogun factory, called by string"...this might be something where other stumble over ... or maybe it is self explaining. What do you think?

src/interfaces/python/swig_typemaps.i
```diff
@@ -1292,5 +1292,30 @@ TYPEMAP_SPARSEFEATURES_OUT(PyObject,      NPY_OBJECT)
...
+    def _internal_factory(name, **kwargs):
+
+        new_obj = _obj(name)
+        for k,v in kwargs.items():
```
karlnapf: is this python2 and 3 ompatible? I remember something changed in iterating over dictionaries ...

src/interfaces/python/swig_typemaps.i
```diff
@@ -1292,5 +1292,30 @@ TYPEMAP_SPARSEFEATURES_OUT(PyObject,      NPY_OBJECT)
...
+            new_obj.put(k, v)
+        return new_obj
+    if docstring:
+        _internal_factory.__doc__ = docstring
+    else:
+        _internal_factory.__doc__ = _obj.__doc__.replace(object_name, new_name)
```
karlnapf: ah great catch.
So how does this work? sg.kernel docstring now is taken from the actualy "kernel" method? If so that is super nice and was something I was a bit worried about

karlnapf:
As the CI passed, this means that the `kwargs` example passed ... ah no but there are no kernel instantiations in there. Did you try the behaviour a bit? If so, we can merge straight away and I can use it in #4462

src/interfaces/python/swig_typemaps.i
karlnapf: ah yeah, dont worry about it

src/interfaces/python/swig_typemaps.i
gf712: Yup!
```
In [1]: import shogun as sg                                                                                                                                                                                 

In [2]: kernel = sg.kernel("GaussianKernel", log_width=1.0)                                                                                                                                                 

In [3]: kernel                                                                                                                                                                                              
Out[3]: GaussianKernel(cache_size=10,combined_kernel_weight=1,lhs=null,lhs_equals_rhs=false,log_width=1,m_distance=EuclideanDistance(...),m_precomputed_distance=null,normalizer=IdentityKernelNormalizer(...),num_lhs=0,num_rhs=0,opt_type=0,optimization_initialized=false,properties=,rhs=null)

In [4]: kernel?                                                                                                                                                                                             
Type:        Kernel
String form: GaussianKernel(cache_size=10,combined_kernel_weight=1,lhs=null,lhs_equals_rhs=false,log_width=1,m <...> rmalizer(...),num_lhs=0,num_rhs=0,opt_type=0,optimization_initialized=false,properties=,rhs=null)
File:        ~/anaconda3/envs/Python36/lib/python3.6/site-packages/shogun.py
Docstring:  
The Kernel base class.

Non-mathematically spoken, a kernel is a function that given two input
objects ${\\bf x}$ and ${\\bf x'}$ returns a score describing the
similarity of the vectors. The score should be larger when the objects
are more similar.

It can be defined as

\\[ k({\\bf x},{\\bf x'})= \\Phi_k({\\bf x})\\cdot
\\Phi_k({\\bf x'}) \\]

where $\\Phi$ maps the objects into some potentially high
dimensional feature space.

Apart from the input features, the base kernel takes only one argument
(the size of the kernel cache) that is used to efficiently train
kernel-machines like e.g. SVMs.

In case you would like to define your own kernel, you only have to
define a new compute() function (and the kernel name via get_name()
and the kernel type get_kernel_type()). A good example to look at is
the GaussianKernel.

C++ includes: Kernel.h
```

gf712: But it doesn't make a lot of sense as the signature has changed, and it should probably be `kernel(name, **kwargs) -> Kernel`. That's why added the option to pass a docstring manually

karlnapf: Could we also move the docstring from `sg.kernel` to be the one of the c++ method `kernel` in factory? Or is that hard?

gf712: Do you mean the one in Kernel.h?
```
/** @brief The Kernel base class.
 *
 * Non-mathematically spoken, a kernel is a function
...
```

Issue #4455, opened 2019-01-09
Introducing Concepts


iglesias:
Concepts can help reducing code complexity in templated generic programming, as well as producing simple compiler error messages. In addition, with a slightly long-term view in mind, producing a Concepts library for machine learning could become useful for the C++ community.

The goal of this task is getting hands-on and start playing with Concepts in Shogun.

As a first simple use case, we could start by moving from Traits to Concepts in the [vectorized RNG pull request](https://github.com/shogun-toolbox/shogun/pull/4437/files). For this we need simple concepts such as Integer or RealNumber. Simple concepts like these have most likely already been defined somewhere else. A place to start looking could be the [range library](https://github.com/ericniebler/range-v3). It would be nice to collect other tutorials and libraries for learning purposes.

Of course alternative and additional places where we could continue experimenting more are most welcome.

Detected challenges:
- Compiler support. To avoid overhead during the investigation, at first I'd say is fine if this is configured somewhat a bit "hacky". For example, Concepts support could be kept in a separate experimental branch and if necessary further isolated with some build conf. flag.
- ...

Background:
https://www.youtube.com/watch?v=HddFGPTAmtU
https://en.wikipedia.org/wiki/Concepts_(C%2B%2B)

@saatvikshah1994 

@karlnapf @vigsterkr @lisitsyn share your thoughts :-)

karlnapf:
Would be cool to start playing with this, at least to develop some intuition for what we can do here.
As this didnt make it into C++17, we could maybe (as per suggestion by @gf712 ) use this  google lib that implements some 17/20 features in a c+11 compatible way ...

gf712:
Ah yes, it's https://github.com/abseil/abseil-cpp. But I don't think concepts is something that has been or will be implemented there. This library is more to add C++14/17/(20) STL to C++11 onwards!

iglesias:
All right, it was abseil, sure.

I suggest that at this moment we don't worry too much about this aspect* as this is really an investigation for us to start getting our hands dirty playing with concepts.

* C++ standard support for concepts and the version of it used in Shogun.

saatvikshah1994:
Adding some more reference materials here(apart from what @iglesias recommended) in case anyone is interested:
1. Rough notes I've written up: [ConceptsBasics.pdf](https://github.com/shogun-toolbox/shogun/files/2781073/ConceptsBasics.pdf)
[...]

PR #4477, opened 2019-01-20
Concepts POC

saatvikshah:
First stab at #4455 using Concepts for vectorized rng.

Currently this will only work with GCC which provides Concepts TS support based on certain flags. I've tried to make the impl compatible with either Concepts TS or C++20 concepts.


src/shogun/lib/sg_type_traits.h
```diff
@@ -0,0 +1,71 @@
...
+template <typename T>
+concept Integral = std::is_integral<T>::value;
+
+// Anything that uses the fill_array* macros
+template<typename T>
+concept RngVectorizable = requires(T* container, index_t len, CRandom* r){
+{ r->fill_array(container, len) };
+} || requires(T* container, index_t len, CRandom* r){
+{ r->fill_array_co(container, len) };
+};
```
gf712: So this checks if either `r->fill_array(container, len)` or `r->fill_array(container, len)` are valid? Very cool!

src/shogun/lib/SGVector.h
```diff
@@ -204,6 +206,36 @@ template<class T> class SGVector : public SGReferencedData
 		/** Fill vector with zeros */
 		void zero();
 
+        template <typename U = T> requires Floating<U> && RngVectorizable<U>
+        void random() {
+            auto r = std::make_unique<CRandom>();
+            r->fill_array_co(vector, vlen);
+        }
+
+        template <typename U = T> requires Integral<U> && RngVectorizable<U>
```
gf712: This is definitely more readable!

karlnapf:
Awesome initiative!
I will check this out a bit more soon, first need to read up on Concepts I guess ;)

src/shogun/lib/sg_type_traits.h
saatvikshah1994: Yep it really is :)

src/shogun/lib/SGMatrix.h
```diff
@@ -306,9 +309,44 @@ template<class T> class SGMatrix : public SGReferencedData
...
 		void zero();
 
+		/**
+		 * Fill matrix with uniformly distributed randoms
+		 * Float Types: [0, 1)
+		 * Int Types: std::numeric_limit<T>::(min->max)
+		**/
+		template <typename U = T> requires Floating<U> && RngVectorizable<U>
+		void random() {
+			auto r = std::make_unique<CRandom>();
```
iglesias: Let's do it at once with the approach to set the seed. I believe it is crucial or at least quite helpful for unit testing.

iglesias: Of course here and in the other functions as well.

src/shogun/lib/sg_type_traits.h
```diff
@@ -0,0 +1,71 @@
+/*
```
iglesias: Nitpicking a bit. Maybe we can put the concepts in a separate `sg_concepts.hpp` file or rename to something else that includes traits and concepts.

src/shogun/lib/SGVector.h
iglesias: Yeah, I find it a very nice advantage for programmer happiness. Great stuff for shorter and more clear compiler error messages as well.

src/shogun/lib/sg_type_traits.h
iglesias: Good stuff, thanks a lot for the work!

For my own understanding, do you have a specific reference to another concept that is defined similarly to this one you have created here (for example, a concept defined in the range-v3 library)?

tests/unit/lib/SGVector_unittest.cc
```diff
@@ -9,6 +9,22 @@
 
 using namespace shogun;
 
+TEST(SGVectorTest, vectorized_rng)
+{
+	auto rng_check_float = [](auto elem) {
```
iglesias: This is a bit of a strange construct. You are passing a function argument to actually only used its type information (which in turn requires you to use the decltype). This is asking for a template.

Can you use a template lambda?

Otherwise, I'd prefer an auxiliary function outside the TEST body.

iglesias (changes requested):
I could not go through the CMakeLists.txt yet. @vigsterkr, how does it look to you?

src/shogun/lib/sg_type_traits.h
saatvikshah1994: Most of the standard concepts I found in documentation/talks were not so specialized(even in use case) as this one is. 
I'll go through the ranges-v3 repo once - thats a good idea.

src/shogun/lib/SGMatrix.h
```diff
@@ -306,9 +309,44 @@ template<class T> class SGMatrix : public SGReferencedData
...
 
-		/** fill matrix with zeros */
+		/** Fill matrix with zeros */
 		void zero();
 
+		/**
+		 * Fill matrix with uniformly distributed randoms
+		 * Float Types: [0, 1)
+		 * Int Types: std::numeric_limit<T>::(min->max)
+		**/
```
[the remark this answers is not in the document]
saatvikshah1994: I'll look this up and correct as needed.

src/shogun/lib/SGMatrix.h
```diff
@@ -306,9 +309,44 @@ template<class T> class SGMatrix : public SGReferencedData
...
+		template <typename U = T> requires Floating<U> && RngVectorizable<U>
+		void random() {
+			auto r = std::make_unique<CRandom>();
+			r->fill_array_co(matrix, num_rows * num_cols);
+		}
+
+		template <typename U = T> requires Integral<U> && RngVectorizable<U>
+		void random() {
+            auto r = std::make_unique<CRandom>();
```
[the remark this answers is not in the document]
saatvikshah1994: Will do.

src/shogun/lib/SGMatrix.h
```diff
@@ -306,9 +309,44 @@ template<class T> class SGMatrix : public SGReferencedData
...
+		 * Float Types: [0, 1)
+		 * Int Types: std::numeric_limit<T>::(min->max)
+		**/
+		template <typename U = T> requires Floating<U> && RngVectorizable<U>
+		void random() {
+			auto r = std::make_unique<CRandom>();
+			r->fill_array_co(matrix, num_rows * num_cols);
+		}
+
+		template <typename U = T> requires Integral<U> && RngVectorizable<U>
```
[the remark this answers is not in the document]
saatvikshah1994: Yes! there is - I was hitting a slight problem when trying to set it up though because of the explicit specializations across all types that have been defined. I plan on breaking down the concepts one level more to:
1. Integral/Floating
2. Rngable
3. RngVectorizable
This should ensure that only correct matches occur.

src/shogun/lib/SGMatrix.h
saatvikshah1994: Not sure why it picked up the older commit without the `sg_rand` addition when I merged :/ . Will add that in.

src/shogun/lib/sg_type_traits.h
saatvikshah1994: Completely agreed - Will do so.

tests/unit/lib/SGVector_unittest.cc
saatvikshah1994: IIRC templates lambdas arent supported till C++20. Here Im using concepts with C++14 itself. I'll put an auxiliary function outside sure.

saatvikshah1994:
FYI, Throw-lists(eg. `void fn() throw(ShogunException)`) which are used in a [number of places](https://github.com/shogun-toolbox/shogun/blob/b5345f5771e30e6ddec0a7c456f58b9f0b071868/src/shogun/base/SGObject.h#L329) have been deprecated in C++11 and removed from C++17 onwards [ref](https://stackoverflow.com/questions/13841559/deprecated-throw-list-in-c11). Hit this majorly along with a few other errors when I was trying to compile with C++17 support.

gf712:
> FYI, Throw-lists(eg. `void fn() throw(ShogunException)`) which are used in a [number of places](https://github.com/shogun-toolbox/shogun/blob/b5345f5771e30e6ddec0a7c456f58b9f0b071868/src/shogun/base/SGObject.h#L329) have been deprecated in C++11 and removed from C++17 onwards [ref](https://stackoverflow.com/questions/13841559/deprecated-throw-list-in-c11). Hit this majorly along with a few other errors when I was trying to compile with C++17 support.

Yup, @vigsterkr started removing all the dynamic exceptions in the cpp17 branch!

src/shogun/lib/SGMatrix.h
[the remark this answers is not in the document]
iglesias: Looking forward to see it :-)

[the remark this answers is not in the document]
saatvikshah1994: This is turning out way harder than I imagined :/
So the thing is concepts dont interfere with name mangling - But this also means that during explicit specialization to the compiler it looks like there are two duplicate templated member functions. Here is a minimal reconstruction: https://coliru.stacked-crooked.com/a/9d472eef7579adc9

src/shogun/lib/sg_type_traits.h
saatvikshah1994: So `ranges-v3` uses C++ metaprogramming hacks to get a syntax resembling concepts but not exactly so. A few similar concepts as the one I've used here are: [Integral](https://github.com/ericniebler/range-v3/blob/cf8add9f467aac695b9fa38e168b90df36b1a806/include/range/v3/utility/concepts.hpp#L346) which matches the ones I've setup and [Swappable](https://github.com/ericniebler/range-v3/blob/cf8add9f467aac695b9fa38e168b90df36b1a806/include/range/v3/utility/concepts.hpp#L387) which checks that `swap(x, y)` is valid somewhat similar to `RngVectorizable`.

src/shogun/lib/SGMatrix.h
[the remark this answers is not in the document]
saatvikshah1994: This seems to be valid following the style (1) mentioned [here](http://www.doxygen.nl/manual/docblocks.html)

src/shogun/lib/sg_type_traits.h
iglesias: All right! Thanks for explaining :-)

src/shogun/lib/SGMatrix.h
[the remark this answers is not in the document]
iglesias: That seems interesting indeed. I wonder whether this is actually expected behavior or some (potentially current) limitation or bug. It could be worth opening a conversation about it, if there's none yet (just an e-mail should do or even a short blog post illustrating the issue).

For the purpose of this pull request, I say we keep it as you solved the problem. Very nice job debugging it and explaining here, thank you!

iglesias (approved):

iglesias:
Any preference to organize merging this and the other PR that triggered this experiment, @saatvikshah1994? I am guessing it may make sense to merge the other one first and then rebase the changes here onto it. Still it is fine by me to merge this one directly to the feature branch.

saatvikshah1994:
@iglesias Yes, would prefer to merge the other one first(#4437) and then rebase and merge this. But I'm not so sure what is the plan with respect to #4437 ?

iglesias:
All CI tasks good. @saatvikshah1994, I can't merge from here due to the conflicts marked above. Would you mind doing the manual job to fix the conflicts? I will merge afterwards.

saatvikshah1994:
Sure @iglesias I'll also add more compilation guards to prevent any future issues.

lfjr:
Thanks

lfjr:
Should I hook up the USB to bugger p

lfjr:
Very lopgical and fast aslo telle I  not tells me that I am not wanted here my phone is also tapped okay the keyboards change this is why the passwords can't be remembered accounts are deleted backup codes backup codes everything is deleted and miss miss directed but they totally under on that estimate me I've been doing this my whole life I just got a little rusty with Android 6 of systems has for security have an FBI clearance

karlnapf:
Reported as spam

PR #4479, opened 2019-01-21
Expose AnyParameter map to C++ API and extend in SWIG

gf712:
This PR exposes AnyParameter map to the C++ library, which is hidden in Self in CSGObject. This is done with ~~`std::reference_wrapper<const AnyParameter>`~~ `std::shared_ptr<const AnyParameter>` (which should make it impossible to mess with a model's parameters through the getter?).
It then further extends the Python API to have parameter_name and parameter_type getters. These can be useful to debug from Python.

### DIFF
--- a/src/interfaces/python/swig_typemaps.i
+++ b/src/interfaces/python/swig_typemaps.i
@@ -1287,6 +1287,24 @@ TYPEMAP_SPARSEFEATURES_OUT(floatmax_t,    NPY_LONGDOUBLE)
 TYPEMAP_SPARSEFEATURES_OUT(PyObject,      NPY_OBJECT)
 #undef TYPEMAP_SPARSEFEATURES_OUT
 
+%typemap(out) shogun::CMap<std::string, std::string>* {
+    $result = PyDict_New();
+    for (int i = 0; i<$1->get_array_size(); ++i) {
+#ifdef PYTHON3
+        auto pair = $1->get_node_ptr(i);
+        int py_result = PyDict_SetItem($result, PyUnicode_FromString((pair->key).c_str()),
+        PyUnicode_FromString((pair->data).c_str()));
+#else
+        int py_result = PyDict_SetItem($result, PyString_FromString((pair->key).c_str()),
+        PyString_FromString((pair->data).c_str()));
+#endif
+        if (py_result == -1) {
+            PyErr_SetString(PyExc_TypeError, "Error whilst creating dictionary");
+            SWIG_fail;
+        }
+    }
+}
+
 %typemap(throws) shogun::ShogunException
 {
     PyErr_SetString(PyExc_SystemError, $1.what());
@@ -1417,7 +1435,7 @@ def _internal_get_param(self, name):
         except Exception:
             raise
     if name in self.parameter_names():
-        raise ValueError("The current Python API does not have a getter for '{}' which has type '{}'".format(name, self.parameter_types()[name]))
+        raise ValueError("The current Python API does not have a getter for '{}' of type '{}'".format(name, self.parameter_types()[name]))
     else:
         raise KeyError("There is no parameter called '{}' in {}".format(name, self.get_name()))
 
--- a/src/interfaces/python/swig_typemaps.i
+++ b/src/interfaces/python/swig_typemaps.i
@@ -1417,7 +1417,7 @@ def _internal_get_param(self, name):
         except Exception:
             raise
     if name in self.parameter_names():
-        raise ValueError("The current Python API does not have a getter for '{}'".format(name))
+        raise ValueError("The current Python API does not have a getter for '{}' which has type '{}'".format(name, self.parameter_types()[name]))
     else:
         raise KeyError("There is no parameter called '{}' in {}".format(name, self.get_name()))
 
--- a/src/interfaces/python/swig_typemaps.i
+++ b/src/interfaces/python/swig_typemaps.i
@@ -1287,24 +1287,6 @@ TYPEMAP_SPARSEFEATURES_OUT(floatmax_t,    NPY_LONGDOUBLE)
 TYPEMAP_SPARSEFEATURES_OUT(PyObject,      NPY_OBJECT)
 #undef TYPEMAP_SPARSEFEATURES_OUT
 
-%typemap(out) shogun::CMap<std::string, std::string>* {
-    $result = PyDict_New();
-    for (int i = 0; i<$1->get_array_size(); ++i) {
-        auto pair = $1->get_node_ptr(i);
-#ifdef PYTHON3
-        int py_result = PyDict_SetItem($result, PyUnicode_FromString((pair->key).c_str()),
-        PyUnicode_FromString((pair->data).c_str()));
-#else
-        int py_result = PyDict_SetItem($result, PyString_FromString((pair->key).c_str()),
-        PyString_FromString((pair->data).c_str()));
-#endif
-        if (py_result == -1) {
-            PyErr_SetString(PyExc_TypeError, "Error whilst creating dictionary");
-            SWIG_fail;
-        }
-    }
-}
-
 %typemap(throws) shogun::ShogunException
 {
     PyErr_SetString(PyExc_SystemError, $1.what());
--- a/src/interfaces/python/swig_typemaps.i
+++ b/src/interfaces/python/swig_typemaps.i
@@ -1417,7 +1417,7 @@ def _internal_get_param(self, name):
         except Exception:
             raise
     if name in self.parameter_names():
-        raise ValueError("The current Python API does not have a getter for '{}' of type '{}'".format(name, self.parameter_types()[name]))
+        raise ValueError("The current Python API does not have a getter for '{}' of type '{}'".format(name, self.parameter_type(name)))
     else:
         raise KeyError("There is no parameter called '{}' in {}".format(name, self.get_name()))
 
--- a/src/interfaces/python/swig_typemaps.i
+++ b/src/interfaces/python/swig_typemaps.i
@@ -1290,8 +1290,8 @@ TYPEMAP_SPARSEFEATURES_OUT(PyObject,      NPY_OBJECT)
 %typemap(out) shogun::CMap<std::string, std::string>* {
     $result = PyDict_New();
     for (int i = 0; i<$1->get_array_size(); ++i) {
-#ifdef PYTHON3
         auto pair = $1->get_node_ptr(i);
+#ifdef PYTHON3
         int py_result = PyDict_SetItem($result, PyUnicode_FromString((pair->key).c_str()),
         PyUnicode_FromString((pair->data).c_str()));
 #else
--- a/src/interfaces/swig/SGBase.i
+++ b/src/interfaces/swig/SGBase.i
@@ -379,12 +379,27 @@ namespace std {
 %include <shogun/base/Parallel.h>
 %include <shogun/lib/StoppableSGObject.h>
 
-#ifdef SWIGPYTHON
 namespace shogun
 {
-
     %extend CSGObject
     {
+        std::vector<std::string> parameter_names() const {
+			std::vector<std::string> result;
+			for (auto const& each: $self->get_parameters()) {
+				result.emplace_back(each.first);
+			}
+			return result;
+        }
+
+        CMap<std::string, std::string>* parameter_types() const {
+            auto result = new CMap<std::string, std::string>;
+            for (auto const& each: $self->get_parameters()) {
+                result->add(each.first, each.second.get().get_value().type());
+            }
+            return result;
+        }
+
+#ifdef SWIGPYTHON
         std::string __str__() const
         {
             return $self->to_string();
@@ -491,33 +506,6 @@ namespace shogun
         }
 
         /*int getbuffer(PyObject *obj, Py_buffer *view, int flags) { return 0; }*/
-
-        std::vector<std::string> parameter_names() const {
-			std::vector<std::string> result;
-			for (auto const& each: self->get_parameters()) {
-				result.emplace_back(each.first);
-			}
-			return result;
-        }
-
-		PyObject* parameter_types() const {
-			PyObject* py_result = PyDict_New();
-			for (auto const& each: self->get_parameters()) {
-#ifdef PYTHON3
-				int result = PyDict_SetItem(py_result, PyUnicode_FromString(each.first.c_str()),
-				PyUnicode_FromString(each.second.get().get_value().type().c_str()));
-#else
-				int result = PyDict_SetItem(py_result, PyString_FromString(each.first.c_str()),
-				PyString_FromString(each.second.get().get_value().type().c_str()));
-#endif
-				if (result == -1) {
-					PyErr_SetString(PyExc_TypeError, "Error whilst creating type dictionary");
-					return NULL;
-				}
-			}
-
-			return py_result;
-		}
     }
 }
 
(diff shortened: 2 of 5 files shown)


src/interfaces/swig/SGBase.i
```diff
@@ -485,6 +491,29 @@ namespace shogun
...
+				result.emplace_back(each.first);
+			}
+			return result;
+        }
+
+		PyObject* parameter_types() const {
+			PyObject* py_result = PyDict_New();
+			for (auto const& each: self->get_parameters()) {
+				int result = PyDict_SetItem(py_result,
+											PyUnicode_FromString(each.first.c_str()),
```
lisitsyn: Indentation looks strange over there

gf712: oh wow, it does seem strange..

gf712: I also need to to add PyString_FromString for python 2

gf712:
Btw I did not manage to get SWIG to create a dictionary from a std::map so had to create and populate one manually. This means that the same has to be done for other languages..

src/interfaces/python/swig_typemaps.i
```diff
@@ -1417,7 +1417,7 @@ def _internal_get_param(self, name):
         except Exception:
             raise
     if name in self.parameter_names():
-        raise ValueError("The current Python API does not have a getter for '{}'".format(name))
+        raise ValueError("The current Python API does not have a getter for '{}' which has type '{}'".format(name, self.parameter_types()[name]))
     else:
```
karlnapf: “Of type” maybe?

src/interfaces/swig/SGBase.i
```diff
@@ -485,6 +491,33 @@ namespace shogun
...
+        std::vector<std::string> parameter_names() const {
+			std::vector<std::string> result;
+			for (auto const& each: self->get_parameters()) {
+				result.emplace_back(each.first);
+			}
+			return result;
+        }
+
+		PyObject* parameter_types() const {
+			PyObject* py_result = PyDict_New();
```
karlnapf: I’m unsure about this....
We would want simple c++ code that is then exposed to swig which does the mapping. In principle.
Maybe we should rather work on getting the typemaps to work?

karlnapf: Also much less code....

gf712: I tried to template std::map<std::string, std::string> but SWIG didn't return a dictionary... Which I thought it would?

src/shogun/base/SGObject.cpp
```diff
@@ -975,12 +975,12 @@ std::string CSGObject::to_string() const
 	return ss.str();
 }
 
-std::vector<std::string> CSGObject::parameter_names() const
+std::map<std::string, std::reference_wrapper<const AnyParameter>> CSGObject::get_parameters() const
```
gf712: I just realised that `get_parameters()` seems to be used a few times in shogun's code base, so will need to change this to something else. @karlnapf do you have any suggestions? `parameter_map()`?

src/interfaces/python/swig_typemaps.i
gf712: yup, I had that at first but then for some reason sounded weird, but I agree that is better!

src/shogun/base/SGObject.cpp
karlnapf: I’ll need to check where that is and whether it can be changed.
Maybe we’ll have a snowball effect of moving c code to swig

src/interfaces/swig/SGBase.i
karlnapf: It depends on whether swig actually has typemaps for maps. And I also remember that sometimes swig was a diva with matching those, eg we had to move stuff to shogun.i as otherwise it didn’t match (some order problem)
Even if there are no typemaps for map I think it would be better to just write those ourselves and keep the code general here.

gf712: You mean return a map<string, string> here and then write a %typemap(argout) for each target language that can't translate map<string, string> to a builtin data structure?

src/shogun/base/SGObject.cpp
gf712: I don't think there are many, there is one in neuralnets (and then DeepAutoencoders) and rbm, from what I can tell

src/interfaces/swig/SGBase.i
gf712: So I had a better look into SWIG's std::map wrapper, and it behaves very similarly to a dict (i.e. has an item getter like behaviour, key and values getter). It just looks weird when printed, which could be odd for a user:
```python
print(kernel.parameter_types())
<Swig Object of type 'std::map< std::string,std::string,std::less< std::string >,std::allocator< std::pair< std::string const,std::string > > > *' at 0x1181de490>
```
So should probably customise the map<string, string> typemap right (at least for python)?

src/shogun/base/SGObject.cpp
vigsterkr: the std::map is not supported for all the languages we support in SWIG... that's why actually we have CMap in the codebase :( so either we have to go with our map or.... it's only std::vector that is being implemented in all of the language mappings

gf712: Ah ok! I tried with CMap but I get an error because the copy constructor is implicitly deleted.. I have this
```
        CMap<std::string, std::string> parameter_types() const {
            CMap<std::string, std::string> result;
            for (auto const& each: self->get_parameters()) {
                result[each.first] = each.second.get().get_value().type();
            }
            return result;
        }
``` 
How do I get SWIG to copy CMap, when it wraps result with SwigValueWrapper?

vigsterkr: mmmm i would allocate it on the heap, and return a `CMap<...>*`

vigsterkr: so basically

```
CMap<std::string, std::string>* parameter_types() const {
            auto result = new CMap<std::string, std::string>();
            for (auto const& each: self->get_parameters()) {
                result->add(each.first, each.second.get().get_value().type());
            }
            return result;
        }
```

gf712: Ah thanks! I got it to work :)

vigsterkr: ok so another thing is that maybe we wanna keep the parameter_names function and have an additional one that returns the map?

gf712: Do you mean in libshogun or in SWIG?

vigsterkr: both... so that we still have just a function that returns the params :)

gf712:
Calling SGObject._get from Python with a string that does not correspond to any parameter creates a new parameter of type Shogun::Empty. Is that by design? The issue is that the new error message will always return `Cannot get parameter Perceptron::foo of type shogun::Empty as object, not object type.` when an object doesn't exist.

gf712:
It seems like when  SG_ERROR is called in CSGObject it tries to access `self->map[BaseTag(name)]`, which creates a new entry in `self->map` with key BaseTag(name) if it doesn't exist. I am guessing this was not intended?
https://github.com/shogun-toolbox/shogun/blob/bb54b63dd31845ae61ab62ca30e849e65d86b3b1/src/shogun/base/SGObject.cpp#L1072-L1075

src/shogun/base/SGObject.h
```diff
@@ -533,10 +534,10 @@ class CSGObject
 	 */
 	virtual std::string to_string() const;
 
-	/** Returns set of all parameter names of the object.
+	/** Returns map of all parameter names and AnyParameter references of the object.
 	 *
 	 */
-	std::vector<std::string> parameter_names() const;
+	std::map<std::string, std::shared_ptr<const AnyParameter>> get_parameters() const;
```
vigsterkr: use Some instead of shared_ptr... that you can use in any swig language :)

gf712: I don't think I can because AnyParameter isn't derived from CSGObject, so would have to implement it's own ref counter..

vigsterkr: sooo true 🤦‍♂️

gf712: But this isn't meant to be exposed directly to SWIG anyway, so it should be fine no?

gf712: The idea is to then use this to extend SGObject in shogun.i to get some information about the parameters, i.e. name and type

vigsterkr: ok so then it's a go... but in that case u could use std::map as well ;)

gf712: The CMap is still needed for the SWIG interface in `CMap<std::string, std::string> parameter_types() const`, right?

vigsterkr: what's the value content?

gf712: the std::string is the demangled type for the param (key), which is then used for now in an error message in the Python API

src/interfaces/swig/SGBase.i
karlnapf: one option. The printed thing looks ugly indeed

src/shogun/base/SGObject.cpp
karlnapf: we actually discussed that the parameter names is not part of any libshogun API and thus not really needed from C++ perspective (it used to be when we didnt have non-owning any policies).
So, following the idea that interface code should not part part of libshogun, I think we should have both as part of swig. And there, yes I think we should have both the map and the vector

karlnapf: However, I am against using CMap as this thing should not even be there. Instead, we can have a getter for single parameter descriptions (this is fine with your intended use as well). 
Furthermore, we could then either have Python/specific lang extensions that turn it into a local datastructure (such as an explicit typemap for std::map in Python) or whatever if we want.

I just think it would be so awkward to actually get a CMap instance from the Python interface when asking for descriptions. Then rather have a list comprehension [obj.get_param_descr(name) for name in obj.parameter_names()] or something.

Oh and finally another solution could be to just put the descriptions as a vector in the same order as the parameter names.

All these solutions are ugly :(

src/shogun/base/SGObject.h
karlnapf: hide it from swig as well `#ifndef SWIG // SWIG should skip this part`

karlnapf:
> It seems like when SG_ERROR is called in CSGObject it tries to access `self->map[BaseTag(name)]`, which creates a new entry in `self->map` with key BaseTag(name) if it doesn't exist. I am guessing this was not intended?

LOL nope. Nice catch!

src/shogun/base/SGObject.cpp
karlnapf: update.

Just saw you already wrote a cmap typemap for python, so I guess this already works well.

I am still a bit skeptical as this makes the interface between python and the other langs (that dont have the cmap typemap) different, which is something we usually want to avoid. Having a single getter avoids this problem at the cost of the loop necessary if users wanna get all descriptions .... Not that bad in interactive environments like Python.

At the end of the day, this is a question of taste I guess

src/interfaces/swig/SGBase.i
```diff
@@ -373,12 +379,39 @@ namespace std {
...
-#ifdef SWIGPYTHON
 namespace shogun
 {
-
     %extend CSGObject
     {
+        std::vector<std::string> parameter_names() const {
+            std::vector<std::string> result;
+            for (auto const& each: $self->get_params()) {
+                result.emplace_back(each.first);
```
karlnapf: just curious, I always thought that emblace is more for cases where one calls a constructor when placing the element to avoid creating a temp variable. Is there a benefit here as well?

src/shogun/base/SGObject.cpp
```diff
@@ -1068,11 +1070,19 @@ CSGObject* CSGObject::get(const std::string& name)
...
 
-
-	SG_ERROR(
-	    "Cannot get parameter %s::%s of type %s as object, not object type.\n",
-	    get_name(), name.c_str(),
-	    self->map[BaseTag(name)].get_value().type().c_str());
+	if (self->map.find(BaseTag(name)) != self->map.end())
+	{
+		SG_ERROR(
+		    "Cannot get parameter %s::%s of type %s as object, not object "
```
karlnapf: Not sure I understand this error message, the second part

src/interfaces/swig/SGBase.i
gf712: Ah sorry, that's from me copying code to handle std::reference_wrapper, that should be push_back

src/shogun/base/SGObject.cpp
karlnapf: ah maybe "Not a shogun object base type" or something like this?

src/shogun/base/SGObject.h
```diff
@@ -533,11 +533,13 @@ class CSGObject
...
 
-	/** Returns set of all parameter names of the object.
+	/** Returns map of parameter names and AnyParameter pairs
+	 * of the object.
 	 *
 	 */
-	std::vector<std::string> parameter_names() const;
-
+#ifndef SWIG // SWIG should skip this part
+	std::map<std::string, std::shared_ptr<const AnyParameter>> get_params() const;
```
karlnapf: doxygen string would be good, like what is this method doing.

gf712: Oh it's already there but I put the `#ifndef SWIG` bellow it. I guess it should go above the oxygen string?

src/shogun/base/SGObject.cpp
gf712: "Not a shogun CSGObject type"? Just to be more specific?

gf712: Or does it not make sense here?

src/shogun/base/SGObject.h
karlnapf: ah sorry I didnt see

src/shogun/base/SGObject.cpp
karlnapf: i would just saw shogun as swig users might not know what sgobject is :)

karlnapf:
I'll wait for CI and then merge

vigsterkr:
this broke somehow windows CI

vigsterkr:
@gf712 any ideas why suddenly this error?

gf712:
not sure... the CI passed in the last commit... I'll have a look at the CI

vigsterkr:
@gf712 yep i've got that but then somehow after merging the NewtonSVM started to cry.... and before it wansn't... dunno how this is possible of course..

gf712:
@vigsterkr from the error it seems like NewtonSVM wasn't defined, which is because LAPACK wasn't detected? Shouldn't there be a check to prevent that test from running if LAPACK hasn't been found? Or is there such a check somewhere else? Not sure how that is related though to this PR though

vigsterkr:
mmm its strange that there's no lapack as mkl is there... lemme check maybe something changed with anaconda env

vigsterkr:
on the other hand i've just checked NewtonSVM and it seems it doesn't use lapack anymore.... but lemme check again

vigsterkr:
@gf712 yeah something changed with the anaconda env as cmake now only detects BLAS but not lapack from MKL part :S

gf712:
and then fix anaconda env :p

vigsterkr:
@gf712 https://github.com/shogun-toolbox/shogun/commit/d78717b8bfdcafe37b61e0fc7516f3e4ee6b2e3d should fix the problem, but i have to see what has happened with anaconda env :(

PR #4480, opened 2019-01-22
Fixed copy constructor initialisation bug

avramidis:


src/shogun/features/StringFeatures.cpp
```diff
@@ -78,19 +78,22 @@ template<class ST> CStringFeatures<ST>::CStringFeatures(CAlphabet* alpha)
...
-	order(orig.order), preprocess_on_get(false),
-	feature_cache(NULL)
+: CFeatures(orig)
 {
 	init();
 
 	ASSERT(orig.single_string == NULL) //not implemented
 
+	num_vectors=orig.num_vectors;
+	single_string=orig.single_string;
```
gf712: so should this be `single_string=std::string(orig.single_string)` to get a copy?

avramidis: See this reply https://github.com/shogun-toolbox/shogun/pull/4423#discussion_r249283822

So if we need shallow copy then this is fine. However, my understanding is that in the future we will use std::string instead of char* for strings? Is that correct @karlnapf ?

vigsterkr: as far as i can remember std::string is mapped in every swig interface, so std::string should be preferred yes

avramidis: I pushed one more change. I think we should not change the data type for this PR since it is just for fixing this particular issue.

karlnapf: agreed. Changing strings is a whole big other story

karlnapf:
Are these copy ctors actually used anywhere? Because if not I would almost rather remove them (if everything continues to work)

avramidis:
The first constructor (with the orig) is not used, the other one it does.

karlnapf:
where exactly?

karlnapf:
Because if it was buggy then how did the code behave? why didnt tests fail? etc :)

avramidis:
Was used here https://github.com/shogun-toolbox/shogun/blob/e01874993fcd5feb836490e5a767123ccc462eca/src/shogun/util/factory.h#L108

I don't think there are tests for that.

karlnapf:
uh... quite essential interfacing code ... amazing
So yeah let's keep this fix :)

karlnapf:
Good to be merge from my side

avramidis:
Great!

avramidis:
@karlnapf Do I need to anything extra for the merge?

PR #4486, opened 2019-01-25
clone_impl type trait

vigsterkr:


gf712:
Couldn't you also use std::is_reference and std::is_pointer here, instead of the decltype?
https://github.com/shogun-toolbox/shogun/blob/e8b776ae412987fecd9ae857f4d7d1d71e817257/src/shogun/lib/any.h#L245

https://github.com/shogun-toolbox/shogun/blob/e8b776ae412987fecd9ae857f4d7d1d71e817257/src/shogun/lib/any.h#L252

PR #4487, opened 2019-01-25
Polykernel improve

vigsterkr:
our PolyKernel was rather simple till now :)


karlnapf:
Good addition definitely! Did you use this or how did it come about? CI is unhappy though

PR #4491, opened 2019-01-28
GaussianKernel set/get_width should be non-virutal

vigsterkr:
as none of the children classes overrides it.

here's the difference in number of operations (top with virtual and none inlined bottom non-virtual inlined) when compiled with O3:
https://pastebin.com/1LvR1338


vigsterkr:
@karlnapf btw why do we store the log_width instead of the width itself? :)

karlnapf:
Because of derivatives of the kernel for GP inference
@votjakovr did that ages ago.

PR #4495, opened 2019-01-29
remove DenseFeatures from SWIG

karlnapf:
#4463
@gf712 @avramidis @vinx13 @shubham808


gf712:
So we need to cleanup all the examples in undocumented?

karlnapf:
Either that or better port them to meta examples where possible
The legacy examples are causing a lot of work for quite some time, rather than touching them we should reduce their numbers.
But since that is not feasible, yes :)
BTW you could also start with other classes and remove them from swig, we already did a big push for dense feats a while back.
Having said that, all efforts are welcome...

karlnapf:
Really any non base class should be removed from the swig ‘.i’ files
Best to do it one by one as there are always issues popping up... see my let few prs

src/interfaces/python/swig_typemaps.i
```diff
@@ -848,89 +848,6 @@ static bool spvector_to_numpy(PyObject* &obj, SGSparseVector<type> sg_vector, in
 	}
 %enddef
 
-%typemap(out) shogun::CFeatures*
```
karlnapf: since the features are only returned by base class in the future, we don't really need this anymore

I guess there is a question here: Are feature types part of the core shogun or will they be plugins @lisitsyn @vigsterkr ?
The current approach in the new examples handles all features by `CFeatures` only, but I now just started to wonder whether we really want to do that? Certainly reduces the number of types ... (`RealFeatures, FloatFeatures, ...` for dense, string, sparse each)

karlnapf: going with making them plugins for now ...

src/interfaces/swig/Features.i
```diff
@@ -98,6 +98,21 @@ PROTOCOLS_DENSELABELS(CRegressionLabels, RegressionLabels, float64_t, "d\0", NPY
 
 %include <shogun/features/DataGenerator.h>
 
+/* FIXME: DenseFeatures are still included/defined since certain classes
```
karlnapf: this is a bit annoying that SWIG needs the template instantiation in order to realise base class hierarchies, but we can delete this once the API conversion is completed

src/interfaces/swig/Features.i
```diff
@@ -308,82 +323,6 @@ namespace shogun
 #endif
 }
 
-/* Templated Class DenseFeatures */
-%include <shogun/features/DenseFeatures.h>
-namespace shogun
-{
-#ifdef USE_BOOL
-	// something wrong with git
-	PROTOCOLS_DENSEFEATURES(BoolFeatures, bool, "?\0", NPY_BOOL)
```
karlnapf: this is a long list of code (in swig) that will not be used anymore.....directly editing feature objects in shogun in the style of
```
f = RealFeatures(mat)
f[:,1] = np.array([1,2,3], dtype=np.float64)
```
wont work anymore. But we want to make feature immutable anyways ...

examples/undocumented/python/distance_sparseeuclidean.py
```diff
@@ -1,17 +1,19 @@
...
 def distance_sparseeuclidean (train_fname=traindat,test_fname=testdat):
-	from shogun import RealFeatures, SparseRealFeatures, CSVFile
+	from shogun import SparseRealFeatures
+	import shogun as sg
 
-	realfeat=RealFeatures(CSVFile(train_fname))
-	feats_train=SparseRealFeatures(realfeat)
-	realfeat=RealFeatures(CSVFile(test_fname))
-	feats_test=SparseRealFeatures(realfeat)
+	feats_train=SparseRealFeatures(sg.libsvm_file(train_fname))
```
karlnapf: this is a temporary add so I can make the examples work with sparse files. We would later on just give a filename/file and the factory figures out the type automatically so that we just have `sg.file`

karlnapf:
Reminder for myself, first merge the gpl feature branch before this one

PR #4496, opened 2019-01-30
Revert "constexpr version"

vigsterkr:
This reverts commit 46f7838f3c0815e821a3bd0b8805e152e26e9da7.

due to constexpr after every commit the Version.h will change and thus
the whole ccache is contaminated as Version.h is included in SGObject.h

### DIFF
--- a/src/shogun/base/Version.cpp
+++ b/src/shogun/base/Version.cpp
@@ -16,17 +16,15 @@ using namespace shogun;
 
 namespace shogun
 {
-// TODO: in C++17 inline all of this in header file and remove it from here
-constexpr int64_t Version::version_revision;
-constexpr int32_t Version::version_year;
-constexpr int32_t Version::version_month;
-constexpr int32_t Version::version_day;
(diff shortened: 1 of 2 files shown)


src/shogun/base/Version.h
```diff
@@ -35,60 +34,35 @@ class Version
 	/** print version */
 	static void print_version();
 
-	/** get main version */
-    static constexpr const char* get_version_main() {
```
gf712: need to keep the equivalent of this

src/shogun/base/Version.cpp
```diff
@@ -54,6 +53,61 @@ void Version::print_version()
 #endif
 }
 
+const char* Version::get_version_main()
```
gf712: cool thanks!

src/shogun/base/Version.h
vigsterkr: should be fixed in the commit below...

PR #4500, opened 2019-01-31
Support assigning any from string

lisitsyn:


tests/unit/lib/Any_unittest.cc
```diff
@@ -96,6 +96,29 @@ class Object : public CSGObject
...
+{
+public:
+	WithFromString() : value()
+	{
+	}
+	WithFromString(const std::string& v) : value(v)
+	{
+	}
+
+	static WithFromString from_string(const std::string& str)
```
karlnapf: Do we need to implement this for every enum type? :scream: ?

lisitsyn: Yes. We should use some fancy enum like https://github.com/aantron/better-enums you mentioned.

This is the only scalable way I can think of.

lisitsyn:
We discussed it with Heiko and probably it might be not that useful. 

The similar thing can just happen in `put`.

PR #4504, opened 2019-02-01
Converted preprocessor meta examples

avramidis:
The are two more undocumented examples that I didn't convert because I need the changes from #4423 
If the changes here are fine I can later add the generated data files.


src/shogun/kernel/Chi2Kernel.cpp
```diff
@@ -15,7 +15,7 @@ using namespace shogun;
 void
 CChi2Kernel::init()
 {
-	SG_ADD(&width, "width", "Kernel width.", ParameterProperties::HYPER);
+	SG_ADD(&width, "kernel_width", "Kernel width.", ParameterProperties::HYPER);
```
karlnapf: why this?

src/shogun/preprocessor/FisherLDA.h
```diff
@@ -38,6 +38,7 @@
 
 #include <shogun/features/Features.h>
 #include <shogun/labels/Labels.h>
+#include <shogun/labels/MulticlassLabels.h>
```
karlnapf: not sure why this is needed?

src/shogun/kernel/Chi2Kernel.cpp
avramidis: Was getting and error that was saying it was double defined

src/shogun/preprocessor/FisherLDA.h
avramidis: Could not compile differently. I am not sure why

avramidis:
Great! I will add the data files too.

src/shogun/kernel/Chi2Kernel.cpp
karlnapf: ok

karlnapf:
could you rebase this, as there is a merge conflict right now (data was changed in another commit)

examples/meta/src/preprocessor/fishlda.sg
```diff
@@ -0,0 +1,18 @@
+File data = csv_file("../../data/fm_train_real.dat")
```
karlnapf: could you name this "fisher_lda"?

avramidis:
Ok, no problem.

examples/meta/src/preprocessor/kernelpca.sg
```diff
@@ -0,0 +1,21 @@
+File data = csv_file("../../data/fm_train_real.dat")
```
karlnapf: rename to "kernel_pca"

examples/meta/src/preprocessor/logplusone.sg
```diff
@@ -0,0 +1,30 @@
+File data_train = csv_file("../../data/fm_train_real.dat")
```
karlnapf: rename to "log_plus_one" ... you get the point of how we want the filenames to look like :)

examples/meta/src/preprocessor/randomfouriergausspreproc.sg
```diff
@@ -0,0 +1,32 @@
+File data_train = csv_file("../../data/fm_train_real.dat")
```
karlnapf: random_fourier_features.sg

src/shogun/kernel/Chi2Kernel.cpp
karlnapf: Actually, I have a hard time believing that there was no single test that ran the kernel constructor before ...

EDIT. I just checked, none of the base classes register a parameter called `width`, so why would this error appear?

src/shogun/preprocessor/FisherLDA.h
karlnapf: I have doubts here as well.... the CI never complained about this but we actually have some test for fisher LDA somewhere so it was compiled before

EDIT: Just checked, the class is used but not included. Could you forward declare it instead?

src/shogun/kernel/Chi2Kernel.cpp
avramidis: I am not sure. I will investigate.

avramidis: Changed it back and now it works with "width". Honestly, I don't know why I had an error before...

avramidis:
Ah yes, I know why it fail. Will fix it tomorrow. I forgot to debase why data PR...

src/shogun/kernel/Chi2Kernel.cpp
karlnapf: lol well I guess nevermind then :)

karlnapf:
rebase and merge?

avramidis:
Rebased and tests passed. We could merge it if it's ok with you.

examples/meta/src/preprocessor/fish_lda.sg
```diff
@@ -0,0 +1,18 @@
...
+
+#![create_pca_preprocessor]
+Transformer preproc = transformer("FisherLDA")
+preproc.put("final_dimensions", 1)
+preproc.put("method", enum EFLDAMethod.CANVAR_FLDA)
+preproc.fit(sg_features, sg_labels)
+#![create_pca_preprocessor]
+
+#![transform_features]
+Features featstrans = preproc.transform(sg_features)
```
karlnapf: Just he one more look.
There is the thing that only numerical values/vectors etc are serialised for the testing, not the objects themselves. So in this example here, the transformed features (matrix) are not part of the test. But I think they should

karlnapf: So if you extract the feature matrix, it will be serialised

avramidis: ok

examples/meta/src/preprocessor/fish_lda.sg
```diff
@@ -0,0 +1,22 @@
...
+preproc.put("method", enum EFLDAMethod.CANVAR_FLDA)
+preproc.fit(feats_train, feats_labels)
+#![create_preprocessor]
+
+#![transform_features]
+Features feats_traing_trans = preproc.transform(feats_train)
+#![transform_features]
+
+#![get_output]
+RealMatrix feats_train_trans_matrix = feats_traing_trans.get_real_matrix("feature_matrix")
```
karlnapf: ++
No need to have a marker (`get_output`) as this is not referenced in a cookbook, but no need to change now

avramidis:
There is a problem with the pca results. The cause is the different sign. I could manually fix that so it can pass the CI tests because I cannot do my machine to return the same result as the CI ones. What do you think?

karlnapf:
CI now fails, not sure what is happening there, should be passing if you add the files. Maybe forgot to update the data submodule?

karlnapf:
ah we wrote at the same time ...

karlnapf:
give the manual fix a try. But I bet this has to do with the eigensolver backend. We had sign issues with those before ...

avramidis:
Yes, that is the problem. I am trying different libraries all day but no luck. I will give a try the manual fix.

karlnapf:
ah this is annoying, it is a known problem with the integration tests.... I am not sure we can even fix this. In the unit tests we make the tests invariant for sign...but meta examples we cannot do this.
One option is to make the CI pass and then use those as reference (while they fail for other archs)... maybe give it a try

karlnapf:
If the manual fix doesnt work, don't loose sleep. Then we need to make the eigensolvers signs platform independent first

avramidis:
OK, test passed now.

karlnapf:
Ok let’s see how this goes on the buildbot....

PR #4505, opened 2019-02-02
Check if parameter is sg base

gf712:
Added a no throw overload of SGObject::get(const std::string&) to use in swig. Catching errors in C++ code in swig can cause issues because of the error printing.
The interface method is needed for openml to check if a model parameter is an sgobject, i.e. CKernel or CDistance, and if so it needs to be serialised properly.

### DIFF
--- a/src/interfaces/swig/SGBase.i
+++ b/src/interfaces/swig/SGBase.i
@@ -411,6 +411,18 @@ namespace shogun
             }
         }
 
+        bool parameter_is_sg_base(const std::string& name) const {
+            auto params = $self->get_params();
+            if (params.find(name) != params.end()) {
+                if ($self->get(name, std::nothrow) != nullptr)
+                    return true;
+                else
+                    return false;
+            }
+            else
+                SG_SERROR("There is no parameter called '%s' in %s", name.c_str(), $self->get_name());
+        }
+
 #ifdef SWIGPYTHON
         std::string __str__() const
         {
--- a/src/shogun/base/SGObject.h
+++ b/src/shogun/base/SGObject.h
@@ -444,7 +444,8 @@ public:
 	 * @param name name of the parameter
 	 * @return object parameter
 	 */
-	CSGObject* get(const std::string& name);
+	CSGObject* get(const std::string& name, std::nothrow_t) const noexcept;
+	CSGObject* get(const std::string& name) const noexcept(false);
 
 #ifndef SWIG
 	/** Typed setter for an object class parameter of a Shogun base class type,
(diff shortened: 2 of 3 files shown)


src/interfaces/swig/SGBase.i
```diff
@@ -411,6 +411,18 @@ namespace shogun
             }
         }
 
+        bool parameter_is_sg_base(const std::string& name) const {
```
karlnapf: maybe briefer: `is_sg_base(const std::string& param_name)`? Or is that confusing?

src/shogun/base/SGObject.h
```diff
@@ -588,7 +598,7 @@ class CSGObject
 
 protected:
 	template <typename T>
-	CSGObject* get_sgobject_type_dispatcher(const std::string& name)
+	CSGObject* get_sgobject_type_dispatcher(const std::string& name) const
```
karlnapf: ++

src/shogun/base/SGObject.h
```diff
@@ -440,11 +440,21 @@ class CSGObject
 
 	/** Untyped getter for an object class parameter, identified by a name.
 	 * Will attempt to get specified object of appropriate internal type.
+	 * If this is not possible it will raise a ShogunException.
 	 *
 	 * @param name name of the parameter
 	 * @return object parameter
 	 */
-	CSGObject* get(const std::string& name);
+	CSGObject* get(const std::string& name) const noexcept(false);
```
karlnapf: @lisitsyn comments?

src/interfaces/swig/SGBase.i
karlnapf: actually, could we make this a function in shogun namespace, rather than a member?
Also, would it be possible to do that with templates then? :D

karlnapf: not sure I was thinking straight when I wrote this but I had in mind an overloaded function that you pass something and it will give you a bool and we specialize the template for sg base types. However, I think with swig this will just give headaches, so maybe the current thing is best

karlnapf:
CI error is unrelated... shall we merge it?

src/interfaces/swig/SGBase.i
gf712: Ah ok! But what would the signature be? Can give it a go, this method is tailored for what I need rather than what a user would need, I think..

karlnapf: Like a method with one argument and bool return. It takes a templated argument and returns false. And then it is spezialized for shogun base classes and then returns true. Not sure that would work :)
If it is only you who needs this the current approach is probably fine as well or?

gf712: It might not be worth adding all those templates just for this? Can always change in the future if it turns out to be useful.

karlnapf: yeah nevermind, I was just fantasizing :)

Issue #4388, opened 2018-07-31
Set lhs of distance to cluster centers in Hierarchical clustering in store_model_features


vinx13:
every [`CDistanceMachine`](https://github.com/shogun-toolbox/shogun/blob/58b914a6b326cd5f799727e39737bc4b4b11b286/src/shogun/machine/DistanceMachine.h#L105) has to make sure that cluster centers are in lhs of distance variable, the task is to fix the todo in https://github.com/shogun-toolbox/shogun/blob/58b914a6b326cd5f799727e39737bc4b4b11b286/src/shogun/clustering/Hierarchical.h#L122

see [KMeans](https://github.com/shogun-toolbox/shogun/blob/58b914a6b326cd5f799727e39737bc4b4b11b286/src/shogun/clustering/KMeansBase.cpp#L249) for example

karlnapf:
as if you need details

braceletboy:
Hi @karlnapf ,
I was looking at the Hierarchical.cpp file while trying to look into this issue. I think there is something wrong with this [line](https://github.com/shogun-toolbox/shogun/blob/58b914a6b326cd5f799727e39737bc4b4b11b286/src/shogun/clustering/Hierarchical.cpp#L129).  
I think the condition should be `for(; l<num && l<=merges && k<num_pairs-1; l++)`. I tried it out on a small example manually and checked myself. Correct me if I am wrong. Also, if the condition `num-l>=merges` is true then so is `l<num`.

karlnapf:
This seems very reasonable to me.
Would you like to send a PR with an update.
In addition, we should maybe test some results against another library on a very simple example?

karlnapf:
@braceletboy this is unrelated to this issue though, so pls let's communicate in another issue, or even better, in a PR from your side

braceletboy:
@karlnapf Sure. I am on it.

braceletboy:
@karlnapf , I was trying to use the `set_feature_vector()` function in the `CDenseFeatures` class. Which is documented [here](http://www.shogun-toolbox.org/api/latest/classshogun_1_1CDenseFeatures.html#affd56f6e718d53ead36a9b67e53055cc)

But, I couldn't find it in the source code. Refer [DenseFeatures.h](https://github.com/shogun-toolbox/shogun/blob/58b914a6b326cd5f799727e39737bc4b4b11b286/src/shogun/features/DenseFeatures.h). Has it been removed?

gf712:
[...]

PR #4507, opened 2019-02-03
Set lhs of distance to cluster centers(#4388)

braceletboy:
Set the lhs of underlying distance to cluster centers. This is done
    thourgh the store_model_features function of the CHierarchical
    Class.


vinx13:
why close this one? you can use force push to update this PR

PR #4518, opened 2019-02-08
Created Beginners Guide simplified version of Getting started .

glitch401:
The current Getting started can be too much for those who are getting started with open source, so beginner guide aims at beginners who aren't familiar with the complicated tools such as make.


karlnapf:
Thanks for the initiative.
I am not sure this contains anything that is neither in our readme nor in any first item google search though. In particular the latter is really crucial if people want to get involved, so we usually do not put any instructions on how to install software, environments etc. Those also outdate at light-speed.

This is why I am slightly reluctant to put this into the main repository. One place where it could be useful was if you would put it into a blog-style permanent website.

indranilbiswas7:
Actually you are right ! I know it's nothing new into it that this repositorie's readme doesn't contain , but it's just a simplified form . 

I had a really hard time with installation of shogun , to be honest and I did make tons of Google search regarding the errors I faced , but at last those were foolish mistake by me and my insufficient information / knowledge base .

But I think shogun has a huge potential as it's so much spread out into different interfaces .

Yes I would love to create a blog style Guide which will also be containing FAQs for common errors faced , but I guess I will need some guidance .

karlnapf:
Why guidance? :) You could just write a summary of the experience you had trying to compile it and show how you overcame them. Then you could put that into a github page or something similar. We would definitely put a link in our wiki. See for example the GSOC blogs we have there. Also, we could help you edit a bit before you make it public.

indranilbiswas7:
That was some healthy guidance :D . Working on it already , sir !

karlnapf:
I'll close this for now. A blog post about this is still very welcome! :)

PR #4506, opened 2019-02-02
Fix gcc8 compilation error

avramidis:
This fix is for issue #4419. The solution seems to work for g++ (Ubuntu 8.2.0-7ubuntu1) 8.2.0 but it might need some tinkering.

PR #4519, opened 2019-02-09
Added SGMatrix benchmarks

avramidis:
This was added for examining if #4506 had any impact on the speed. See https://github.com/shogun-toolbox/shogun/pull/4506#pullrequestreview-200222124


karlnapf:
What are the results? :)

Issue #3992, opened 2017-10-30
fix test failing with MKL


vigsterkr:
when using MKL as BLAS/LAPACK backend then the following tests fails:
```	 42 - unit-LogitVGPiecewiseBoundLikelihood (OTHER_FAULT)
	 93 - unit-CSOBI (OTHER_FAULT)
	201 - unit-PCA (Failed)
	206 - unit-FLDATest (Failed)
	233 - unit-Statistics (Failed)
	278 - integration_meta_cpp-converter-ica_sobi (OTHER_FAULT)
	279 - integration_meta_cpp-converter-ica_fast (Failed)
	310 - integration_meta_cpp-regression-krr_nystrom (Failed)
	368 - libshogun-variational_approx_example (OTHER_FAULT)
	426 - generated_cpp-converter-ica_sobi (OTHER_FAULT)
```

vigsterkr:
@karlnapf fyi: with develop now you can compile shogun with conda's mkl... has some bugs but nothing major...

karlnapf:
This is ace! It deserves a NEWS entry I think. Shall I create a build on buildbot?

karlnapf:
Entrance task:

 * install conda
 * find out how to compile with conda's MKL activated
 * run tests

Most of the failed tests are probably due to slightly changes numerical values when using matrix decompositions.
 
 * integration tests: just don't store the failure causing value in a variable
 * unit tests: investigate

vigsterkr:
`integration_meta_cpp-regression-krr_nystrom` should be fixed

newbazz:
Hey I am new to this organisation, I would like to fix this issue.

vigsterkr:
@newbazz welcome! and welcome to ask if you get stuck :)

newbazz:
@vigsterkr can u help me in solving this issue.

vigsterkr:
@newbazz ? as mentioned in my previous post if you wanna fix this start trying to replicate the problem and then tell me where exactly are you stuck

rahul16gupta:
i want to contribute. please help where to start

vigsterkr:
@rahul16gupta great! as mentioned already above you should compile the latest shogun in develop with MKL support and run the unit tests and you'll get the error.
one way to have intel's MKL is to actually use anaconda.

other than that anybody who's trying to fix this issue plz post here where are you stuck exactly.

vigsterkr:
fixed in #4529

PR #4523, opened 2019-02-12
MKL fixes

gf712:
I am not sure what causes this issue, but somehow separating evaluation of unary functions from the main expression fixes the issue.

I did some debugging and it seems like for example the square root operation of a matrix (`.cwiseSqrt()`) makes its rows and cols attributes be 0, which then causes an assertion error when this matrix is used in another operation.

This only happens with MKL backend and is first reported in #3992

### DIFF
--- a/src/shogun/converter/ica/SOBI.cpp
+++ b/src/shogun/converter/ica/SOBI.cpp
@@ -66,8 +66,11 @@ void CSOBI::fit_dense(CDenseFeatures<float64_t>* features)
 	MatrixXd M0 = cor(EX,int(m_tau[0]));
 	EigenSolver<MatrixXd> eig;
 	eig.compute(M0);
-	MatrixXd SPH = (eig.pseudoEigenvectors() * eig.pseudoEigenvalueMatrix().cwiseSqrt() * eig.pseudoEigenvectors ().transpose()).inverse();
-	MatrixXd spx = SPH*EX;
+	MatrixXd EVMsqrt = eig.pseudoEigenvalueMatrix().cwiseSqrt();
+	MatrixXd SPH = (eig.pseudoEigenvectors() * EVMsqrt *
+	                eig.pseudoEigenvectors().transpose())
+	                   .inverse();
+	MatrixXd spx = SPH * EX;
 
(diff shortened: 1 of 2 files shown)


gf712:
If there are any Eigen experts around it would be nice to know what is going wrong here so we can file a ticket at Eigen. From what I can gather is that unary functions (power and sqrt) are not being properly evaluated lazily and have to be force evaluated for these code snippets to work.

src/shogun/converter/ica/SOBI.cpp
```diff
@@ -66,8 +66,11 @@ void CSOBI::fit_dense(CDenseFeatures<float64_t>* features)
 	MatrixXd M0 = cor(EX,int(m_tau[0]));
 	EigenSolver<MatrixXd> eig;
 	eig.compute(M0);
-	MatrixXd SPH = (eig.pseudoEigenvectors() * eig.pseudoEigenvalueMatrix().cwiseSqrt() * eig.pseudoEigenvectors ().transpose()).inverse();
-	MatrixXd spx = SPH*EX;
+	MatrixXd EVMsqrt = eig.pseudoEigenvalueMatrix().cwiseSqrt();
```
karlnapf: just asking, is this covered by any tests that would detect changing results?

gf712: Should be covered [here](https://github.com/shogun-toolbox/shogun/blob/41888fe7c8dc3797063d674f452e13351f321338/tests/unit/converter/ica/SOBI_unittest.cc). That's how I found these bugs!

gf712: Is there a buildbot for MKL backend btw?

vigsterkr:
Add one to azure :)

gf712:
> Add one to azure :)
> […](#)

OK! Just for GCC should be fine right? Might need some help setting it up.

vigsterkr:
Yep but use clang as its faster compiling;)
I’ll b on irc in a bit

vigsterkr:
closing in favour of #4529
@gf712 right?

gf712:
Yup!

PR #4524, opened 2019-02-12
Fixes doxygen issue

gf712:
another instance where doxygen doesn't like parameter packs.


vinx13:
@indranilbiswas7 I think this solves your issue with doxygen

PR #4529, opened 2019-02-14
mkl CI config

gf712:
Adds MKL job to macOS


.ci/ci.yml
```diff
@@ -137,7 +137,15 @@ jobs:
     vmImage: xcode9-macos10.13
 
   steps:
-  - template: ./macos-steps.yml
+    - template: ./macos-steps.yml
+  strategy:
+    matrix:
+      vanilla:
+        parameters:
```
vigsterkr: dunno if this is actually needed :)

.ci/ci.yml
```diff
@@ -137,7 +137,15 @@ jobs:
...
+          mkl: false
+      mkl:
+        parameters:
```
vigsterkr: same as above... you could jsut check for the var itself

.ci/macos-steps.yml
```diff
@@ -20,10 +20,25 @@ steps:
...
+- ${{ if eq(parameters.mkl, 'true') }}:
+    - task: CMake@1
+      displayName: Setup
+      variables:
+        CONDA_DEFAULT_ENV: shogun
+        CONDA_EXE: $CONDA_EXE
+        CONDA_PREFIX_1: $CONDA_PREFIX
+        CONDA_PREFIX: $CONDA_PREFIX/envs/shogun
+        CONDA_PYTHON_EXE: $CONDA_PREFIX/python
+      steps:
```
vigsterkr: tasks are specialized steps.... so basically her eyou need first

```
 - script: conda install mkl-devel
    displayName: asdf
-  - task: CMake@1
....
```

.ci/macos-steps.yml
```diff
@@ -20,10 +23,24 @@ steps:
...
+    - task: CMake@1
+      displayName: Setup
+      variables:
+        CONDA_DEFAULT_ENV: shogun
+        CONDA_EXE: $CONDA_EXE
+        CONDA_PREFIX_1: $CONDA_PREFIX
+        CONDA_PREFIX: $CONDA_PREFIX/envs/shogun
+        CONDA_PYTHON_EXE: $CONDA_PREFIX/python
+      inputs:
+        cmakeArgs: -DBUILD_EXAMPLES=OFF -DBUNDLE_JSON=OFF -DBUNDLE_NLOPT=OFF -DENABLE_TESTING=ON -DENABLE_COVERAGE=OFF -DBUILD_META_EXAMPLES=OFF ..
```
vigsterkr: we might need to add `-DBLA_VENDOR=Intel` to force the usage of mkl

.ci/ci.yml
```diff
@@ -137,7 +137,15 @@ jobs:
     vmImage: xcode9-macos10.13
 
   steps:
-  - template: ./macos-steps.yml
+    - template: ./macos-steps.yml
+
+  strategy:
+    maxParallel: 2
+    matrix:
+      LAPACK:
```
vigsterkr: this should be `ACCELERATE`

.ci/macos-steps.yml
```diff
@@ -1,3 +1,6 @@
+parameters:
```
vigsterkr: no need for this imo

.ci/macos-steps.yml
```diff
@@ -20,10 +23,24 @@ steps:
     brew install ccache pkg-config arpack eigen glpk hdf5 json-c lapack lzo nlopt snappy xz
   displayName: Install dependencies
 
-- task: CMake@1
-  displayName: Setup
-  inputs:
-    cmakeArgs: -DBUILD_EXAMPLES=OFF -DBUNDLE_JSON=OFF -DBUNDLE_NLOPT=OFF -DENABLE_TESTING=ON -DENABLE_COVERAGE=OFF -DBUILD_META_EXAMPLES=OFF ..
+- ${{ if eq(parameters.use_mkl, 'true') }}:
```
gf712: @vigsterkr how do you capture env variables here? $(use_mkl) doesn't seem to do anything...

.ci/macos-steps.yml
gf712: true, I just didn't know if it was needed to capture the env variable, not sure why it isn't working :(

.ci/macos-steps.yml
```diff
@@ -20,10 +23,24 @@ steps:
     brew install ccache pkg-config arpack eigen glpk hdf5 json-c lapack lzo nlopt snappy xz
   displayName: Install dependencies
 
-- task: CMake@1
-  displayName: Setup
-  inputs:
-    cmakeArgs: -DBUILD_EXAMPLES=OFF -DBUNDLE_JSON=OFF -DBUNDLE_NLOPT=OFF -DENABLE_TESTING=ON -DENABLE_COVERAGE=OFF -DBUILD_META_EXAMPLES=OFF ..
+- ${{ if eq('$(use_mkl)', 'true') }}:
```
vigsterkr: this should be:
```
${{ if eq(variables['use_mkl'], 'true') }}:
```

cmake/ShogunFindLAPACK.cmake
```diff
@@ -66,14 +66,14 @@ IF (LAPACK_FOUND)
...
-      FIND_PATH(MKL_INCLUDE_DIR mkl.h)
+      FIND_PATH(MKL_INCLUDE_DIR mkl.h HINTS ${MKL_PATH}/../include/)
       IF(NOT MKL_INCLUDE_DIR)
         MESSAGE(STATUS "Found MKL, but not mkl.h. Make sure that mkl headers are available in order to use MKL as BLAS/Lapack backend for Eigen.")
         SET(ENABLE_EIGEN_LAPACK OFF CACHE BOOL "Use ${ENABLE_EIGEN_LAPACK}" FORCE)
       ELSE()
         MESSAGE(STATUS "Enabling MKL as BLAS/Lapack backend for Eigen.")
         SET(EIGEN_USE_MKL_ALL 1)
-        target_include_directories(shogun PUBLIC ${MKL_INCLUDE_DIR})
+        target_include_directories(libshogun PUBLIC ${MKL_INCLUDE_DIR})
```
gf712: @vigsterkr I think this is bug right?

vigsterkr: we should use the public for shogun when mkl is used because then when you link with shogun.so you need to force that the user links against mkl as well

gf712: So need to have
```
target_include_directories(shogun PUBLIC ${MKL_INCLUDE_DIR})
target_include_directories(libshogun PUBLIC ${MKL_INCLUDE_DIR})
```
?

vigsterkr: i think the includes could be `PRIVATE` as i dont think we expose mkl things in our headers.

vigsterkr: or do we? i'm not so sure.... need to see if any of the headers of shogun exposes mkl in some indirect way..... but definitely libshogun needs that as well

gf712: well can do a bit of trial and error... at least its working now. Maybe can merge #4523 to see if it works with this CI config?

vigsterkr:
@gf712 cherry pick the commits of that branch into this one and we'll see right away ;)

cmake/ShogunFindLAPACK.cmake
gf712: so seems like libshogun can be private, but not shogun. Guess that was expected..

gf712:
> @gf712 cherry pick the commits of that branch into this one and we'll see right away ;)

There is an issue with unit-FLDATest, the values seem to be symmetric though...

gf712:
@vigsterkr seems to be working now. Just need to sort the ccache now right?

PR #4532, opened 2019-02-20
Removed enum EMachineType

mrkarna:
Deleted the lines describing enum EMachineType as it was no where used in the whole code.


karlnapf:
Lol, you cannot just delete this as shogun won't compile anymore. The CI output above shows you some errors as a starting point. But before you start refactoring/deleting all code that uses the MachineType type, pls make sure it is feasible to avoid running into a deadend. If the enum and code based on it is never really used, we can just safely delete it indeed

vigsterkr:
We need to get rid of these due to the plugin based arch... there we cannot rely on these enums... but indeed before you push a pr plz make sure that it compiles locally.

mrkarna:
Sorry, I didn't get that at the first place. Can you please tell an efficient way to remove the enum EMachineType as it is used in many files?
The only way I see is to use a map to replace enum variables by their corresponding integer value. Is this correct way of getting rid of EMachineType enum?

karlnapf:
It needs some intelligent refactoring and hat depends on the code, and we will need to discuss different strategies. Nontrivial.

Replacing them with int is not the way to do it.

mrkarna:
Thanks for your suggestions. Please let me know if there is viable solution. I will try doing that.

PR #4533, opened 2019-02-20
Auto feature cleanup

gf712:
@karlnapf working on a cleaner API for automatic parameter initialisation with @vigsterkr. This could then be eventually exposed to the interfaces.

However this fails [`SGObjectAll_unittest.cc`](https://github.com/shogun-toolbox/shogun/blob/78706b74166454701d303c788824d70440bbbe9d/tests/unit/base/SGObjectAll_unittest.cc#L195) because now the refcount of `this` is 2, instead of 1, since it is referenced inside AutoInit.

### DIFF
--- a/src/shogun/lib/auto_init_factory.h
+++ b/src/shogun/lib/auto_init_factory.h
@@ -1,0 +1,66 @@
+/*
+ * This software is distributed under BSD 3-clause license (see LICENSE file).
+ *
+ * Authors: Gil Hoben
+ */
+
+#ifndef __AUTO_INIT_FACTORY_H__
+#define __AUTO_INIT_FACTORY_H__
+
+#include <shogun/features/DenseFeatures.h>
+#include <shogun/kernel/Kernel.h>
+#include <shogun/lib/abstract_auto_init.h>
+#include <shogun/lib/any.h>
+
+namespace shogun
+{
+	namespace factory
+	{
+		class GammaFeatureNumberInit : public AutoInit
+		{
+			static const char* const kName;
+			static const char* const kDescription;
+
+		public:
+			GammaFeatureNumberInit(CKernel* kernel)
+			    : AutoInit(kName, kDescription)
+			{
+				SG_REF(kernel);
+				m_kernel = kernel;
+			}
+
+			~GammaFeatureNumberInit() override
+			{
+				SG_UNREF(m_kernel);
+			}
+
+			Any operator()() override
+			{
+				Any result;
+				auto m_lhs = m_kernel->get<CFeatures*>("lhs");
+				if (m_lhs->get_feature_class() >=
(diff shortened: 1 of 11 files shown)


src/shogun/lib/auto_init_factory.h
```diff
@@ -0,0 +1,66 @@
...
+
+			~GammaFeatureNumberInit() override
+			{
+				SG_UNREF(m_kernel);
+			}
+
+			Any operator()() override
+			{
+				Any result;
+				auto m_lhs = m_kernel->get<CFeatures*>("lhs");
```
karlnapf: I think here you can use the actual getter for lhs (if it exists) as the kernel type will always be known. This way we know for sure there is no runtime error if the parameter is renamed or a typo happens or whatever?

karlnapf: `CKernel::get_lhs`

karlnapf:
Re the failing unit test, just remove the line that asserts the ref count

vigsterkr:
Its pretty standard to define static vars with camelcase and prefix it with k

karlnapf:
I never saw it in shogun, but I also don't really mind either way.

vigsterkr:
Yeah we dont really use these standard notations... but maybe it’s time to change it... especially that as you can see this is not exposed to anything.

vigsterkr:
We could use std::weak_ptr within the autoinit class for storing reference to the kernel and not increment the ref counter of kernel...

src/shogun/lib/auto_init_factory.h
```diff
@@ -0,0 +1,66 @@
...
+				if (m_lhs->get_feature_class() >=
+				        EFeatureClass::C_STREAMING_DENSE &&
```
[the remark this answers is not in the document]
vigsterkr: i wouldn't be so sure about this.... and `as` is quite costly anyways... typeid hash would be more appropriate.... but again the feature class enum we might not wanna get rid of.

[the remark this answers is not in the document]
karlnapf: I guess costs always are in a context. In this context (kernel machines) they are neglectable. But actually you are right if feature classes stay in the main framework (as opposed to being plugins) we can just keep the enum here.

src/shogun/lib/auto_init_factory.h
gf712: I must have missed it!

gf712:
need to fix the legacy code?

karlnapf:
What do you mean?

gf712:
The python tests failed when they use the kernel constructor with features (as opposed to the factories)

karlnapf:
Ah. Yes pls fix them, still dragging those along for coverage.... but maybe not first thing but only before merging the feature branch into develop

PR #4537, opened 2019-02-21
String enum mapping

gf712:
So here is an attempt to start mapping strings from the interface to enums in the library.

So this works now:
```python
In [1]: import shogun as sg                                                                                                                                                                          

In [2]: lr = sg.machine("LibLinearRegression")                                                                                                                                                       

In [3]: lr                                                                                                                                                                                           
Out[3]: LibLinearRegression(C=1,bias=0,data_locked=false,features=null,labels=null,liblinear_regression_type=1,m_tube_epsilon=0.01,max_iter=10000,max_train_time=0,solver_type=0,store_model_features=false,use_bias=false,w=Vector<double>(0): [])

In [4]: lr.param_options("liblinear_regression_type")                                                                                                                                                
Out[4]: ('L2R_L2LOSS_SVR_DUAL', 'L2R_L1LOSS_SVR_DUAL', 'L2R_L2LOSS_SVR')

In [5]: lr.put("liblinear_regression_type", 'L2R_L2LOSS_SVR_DUAL')                                                                                                                                   

In [6]: lr                                                                                                                                                                                           
Out[6]: LibLinearRegression(C=1,bias=0,data_locked=false,features=null,labels=null,liblinear_regression_type=2,m_tube_epsilon=0.01,max_iter=10000,max_train_time=0,solver_type=0,store_model_features=false,use_bias=false,w=Vector<double>(0): [])

In [7]: lr.get("liblinear_regression_type")                                                                                                                                                         
Out[7]: 1

In [8]: lr.put("liblinear_regression_type", 'L2R_L2LOSS_SVR_DUAL')                                                                                                                                  

In [9]: lr.get("liblinear_regression_type")                                                                                                                                                         
Out[9]: 2
```

Need to think about a neater way to register enums and their mappings though..

### DIFF
--- a/examples/meta/src/base_api/put_get_add.sg
+++ b/examples/meta/src/base_api/put_get_add.sg
@@ -33,8 +33,12 @@ Machine svm = machine("LibSVM")
 svm.put("kernel", k)
 Kernel k2 = as_kernel(svm.get("kernel"))
 svm.put("kernel", k2)
+
+#![the following "libsvm_solver_type" put are equivalent]
 svm.put("libsvm_solver_type", 2)
 svm.put("libsvm_solver_type", "LIBSVM_C_SVC")
+#![option is always returned as a string]
+string option = svm.get_option("libsvm_solver_type")
 
 CSVFile f_feats("../../data/regression_1d_sinc_features_train.dat")
 Features feats = features(f_feats)
--- a/examples/meta/src/base_api/put_get_add.sg
+++ b/examples/meta/src/base_api/put_get_add.sg
@@ -33,6 +33,8 @@ Machine svm = machine("LibSVM")
 svm.put("kernel", k)
 Kernel k2 = as_kernel(svm.get("kernel"))
 svm.put("kernel", k2)
+svm.put("libsvm_solver_type", 2)
+svm.put("libsvm_solver_type", "LIBSVM_C_SVC")
 
 CSVFile f_feats("../../data/regression_1d_sinc_features_train.dat")
 Features feats = features(f_feats)
--- a/src/interfaces/swig/SGBase.i
+++ b/src/interfaces/swig/SGBase.i
@@ -420,25 +420,24 @@ namespace shogun
                     return false;
             }
             else
+            {
                 SG_SERROR("There is no parameter called '%s' in %s", name.c_str(), $self->get_name());
+            }
         }
 
         std::vector<std::string> param_options(const std::string& name) const {
             std::vector<std::string> result;
 
-            auto enum_to_string_map = $self->get_enum_to_string_map();
+            auto param_to_enum_map = $self->get_string_to_enum_map();
 
-            if (enum_to_string_map.find($self->get_name()) == enum_to_string_map.end())
-                SG_SERROR("There are no parameter options for the class '%s'", $self->get_name());
+            if (param_to_enum_map.find(name) == param_to_enum_map.end())
+			{
+                SG_SERROR("There are no options for the parameter '%s'", name.c_str());
+            }
 
-            std::unordered_map<std::string, std::unordered_map<std::string, machine_int_t>> class_to_param = enum_to_string_map[$self->get_name()];
+            auto string_to_enum_map = param_to_enum_map[name];
 
-            if (class_to_param.find(name) == class_to_param.end())
-                SG_SERROR("There is no enum mapping for the parameter '%s'", name.c_str());
-
-            std::unordered_map<std::string, machine_int_t> string_to_enum = class_to_param[name];
-
-            for (auto const& each: string_to_enum)
+            for (auto const& each: string_to_enum_map)
                 result.push_back(each.first);
 
             return result;
--- a/src/interfaces/swig/SGBase.i
+++ b/src/interfaces/swig/SGBase.i
@@ -431,8 +431,8 @@ namespace shogun
             auto param_to_enum_map = $self->get_string_to_enum_map();
 
             if (param_to_enum_map.find(name) == param_to_enum_map.end())
-			{
-                SG_SERROR("There are no options for the parameter '%s'", name.c_str());
+            {
+                SG_SERROR("There are no options for parameter %s::%s", $self->get_name(), name.c_str());
             }
 
             auto string_to_enum_map = param_to_enum_map[name];
--- a/src/interfaces/swig/SGBase.i
+++ b/src/interfaces/swig/SGBase.i
@@ -443,7 +443,7 @@ namespace shogun
             return result;
         }
 
-        std::string get_enum(const std::string& name) const {
+        std::string get_option(const std::string& name) const {
             auto param_to_enum_map = $self->get_string_to_enum_map();
             if (param_to_enum_map.find(name) == param_to_enum_map.end())
             {
--- a/src/interfaces/swig/SGBase.i
+++ b/src/interfaces/swig/SGBase.i
@@ -443,6 +443,15 @@ namespace shogun
             return result;
         }
 
+        std::string get_enum(const std::string& name) const {
+            auto param_to_enum_map = $self->get_string_to_enum_map();
+            if (param_to_enum_map.find(name) == param_to_enum_map.end())
+            {
+                SG_SERROR("There are no options for parameter %s::%s", $self->get_name(), name.c_str());
+            }
+            return $self->string_enum_reverse_lookup(name, $self->get<machine_int_t>(name));
+        }
+
 #ifdef SWIGPYTHON
         std::string __str__() const
         {
--- a/src/interfaces/swig/SGBase.i
+++ b/src/interfaces/swig/SGBase.i
@@ -443,15 +443,6 @@ namespace shogun
             return result;
         }
 
-        std::string get_option(const std::string& name) const {
-            auto param_to_enum_map = $self->get_string_to_enum_map();
-            if (param_to_enum_map.find(name) == param_to_enum_map.end())
-            {
-                SG_SERROR("There are no options for parameter %s::%s", $self->get_name(), name.c_str());
-            }
-            return $self->string_enum_reverse_lookup(name, $self->get<machine_int_t>(name));
-        }
-
 #ifdef SWIGPYTHON
         std::string __str__() const
         {
--- a/src/interfaces/swig/SGBase.i
+++ b/src/interfaces/swig/SGBase.i
@@ -423,6 +423,27 @@ namespace shogun
                 SG_SERROR("There is no parameter called '%s' in %s", name.c_str(), $self->get_name());
         }
 
+        std::vector<std::string> param_options(const std::string& name) const {
+            std::vector<std::string> result;
+
+            auto enum_to_string_map = $self->get_enum_to_string_map();
+
+            if (enum_to_string_map.find($self->get_name()) == enum_to_string_map.end())
+                SG_SERROR("There are no parameter options for the class '%s'", $self->get_name());
+
+            std::unordered_map<std::string, std::unordered_map<std::string, machine_int_t>> class_to_param = enum_to_string_map[$self->get_name()];
+
+            if (class_to_param.find(name) == class_to_param.end())
+                SG_SERROR("There is no enum mapping for the parameter '%s'", name.c_str());
+
+            std::unordered_map<std::string, machine_int_t> string_to_enum = class_to_param[name];
+
+            for (auto const& each: string_to_enum)
+                result.push_back(each.first);
+
+            return result;
+        }
+
 #ifdef SWIGPYTHON
         std::string __str__() const
         {
--- a/src/interfaces/swig/shogun.i
+++ b/src/interfaces/swig/shogun.i
@@ -125,19 +125,16 @@ import org.jblas.*;
 
 %define PUT_ENUM_INT_DISPATCHER(TAG, VALUE)
     auto string_to_enum_map = $self->get_string_to_enum_map();
-
     if (string_to_enum_map.find(TAG.name()) == string_to_enum_map.end()) {
         $self->put(TAG, VALUE);
         return;
     }
-
     auto val = static_cast<machine_int_t>(VALUE);
     auto string_to_enum = string_to_enum_map[TAG.name()];
     auto count = std::count_if(string_to_enum.begin(), string_to_enum.end(),
                                [val](const std::pair <std::string, machine_int_t> &p) {
                                    return val == p.second;
                                });
-
     if (count > 0)
     {
         $self->put(Tag<machine_int_t>(TAG.name()), val);
@@ -273,7 +270,7 @@ namespace shogun
 %template(get_int_vector) CSGObject::get_vector_as_matrix_dispatcher<SGMatrix<int32_t>, int32_t>;
 #endif // SWIGJAVA
 
-%template(put) CSGObject::put<std::string>;
+%template(put) CSGObject::put<std::string, std::string>;
 
 %define PUT_ADD(sg_class)
 %template(put) CSGObject::put<sg_class, sg_class, void>;
--- a/src/interfaces/swig/shogun.i
+++ b/src/interfaces/swig/shogun.i
@@ -261,6 +261,7 @@ namespace shogun
 %template(get_real_matrix) CSGObject::get<SGMatrix<float64_t>, void>;
 %template(get_char_string_list) CSGObject::get<SGStringList<char>, void>;
 %template(get_word_string_list) CSGObject::get<SGStringList<uint16_t>, void>;
+%template(get_option) CSGObject::get<std::string, void>;
 
 #ifndef SWIGJAVA
 %template(get_real_vector) CSGObject::get<SGVector<float64_t>, void>;
--- a/src/interfaces/swig/shogun.i
+++ b/src/interfaces/swig/shogun.i
@@ -243,6 +243,8 @@ namespace shogun
 %template(get_int_vector) CSGObject::get_vector_as_matrix_dispatcher<SGMatrix<int32_t>, int32_t>;
 #endif // SWIGJAVA
 
+%template(put) CSGObject::put<std::string>;
+
 %define PUT_ADD(sg_class)
 %template(put) CSGObject::put<sg_class, sg_class, void>;
 %template(add) CSGObject::add<sg_class, sg_class>;
--- a/src/interfaces/swig/shogun.i
+++ b/src/interfaces/swig/shogun.i
@@ -163,12 +163,12 @@ namespace shogun
 
 		if ($self->has(tag_int32))
 		{
-            PUT_ENUM_INT_DISPATCHER(tag_int32, (int32_t) value);
-        }
+			PUT_ENUM_INT_DISPATCHER(tag_int32, (int32_t) value);
+		}
 		else if ($self->has(tag_int64))
 		{
-            PUT_ENUM_INT_DISPATCHER(tag_int64, (int64_t) value);
-        }
+			PUT_ENUM_INT_DISPATCHER(tag_int64, (int64_t) value);
+		}
 		else if ($self->has(tag_float64))
 			$self->put(tag_float64, (float64_t)value);
 		else
--- a/src/interfaces/swig/shogun.i
+++ b/src/interfaces/swig/shogun.i
@@ -123,11 +123,37 @@
 import org.jblas.*;
 %}
 
+%define PUT_ENUM_INT_DISPATCHER(TAG, VALUE)
+    auto string_to_enum_map = $self->get_string_to_enum_map();
+
+    if (string_to_enum_map.find(TAG.name()) == string_to_enum_map.end()) {
+        $self->put(TAG, VALUE);
+        return;
+    }
+
+    auto val = static_cast<machine_int_t>(VALUE);
+    auto string_to_enum = string_to_enum_map[TAG.name()];
+    auto count = std::count_if(string_to_enum.begin(), string_to_enum.end(),
+                               [val](const std::pair <std::string, machine_int_t> &p) {
+                                   return val == p.second;
+                               });
+
+    if (count > 0)
+    {
+        $self->put(Tag<machine_int_t>(TAG.name()), val);
+    }
+    else
+    {
+        SG_SERROR("There is no option in %s::%s for value %i",
+                $self->get_name(), TAG.name().c_str(), val);
+    }
+%enddef
+
 namespace shogun
 {
 %extend CSGObject
 {
-	template <typename T, typename U= typename std::enable_if_t<std::is_arithmetic<T>::value>>
+	template <typename T, typename U = typename std::enable_if_t<std::is_arithmetic<T>::value>>
 	void put_scalar_dispatcher(const std::string& name, T value)
 	{
 		Tag<T> tag_t(name);
@@ -136,9 +162,13 @@ namespace shogun
 		Tag<float64_t> tag_float64(name);
 
 		if ($self->has(tag_int32))
-			$self->put(tag_int32, (int32_t)value);
+		{
+            PUT_ENUM_INT_DISPATCHER(tag_int32, (int32_t) value);
+        }
 		else if ($self->has(tag_int64))
-			$self->put(tag_int64, (int64_t)value);
+		{
+            PUT_ENUM_INT_DISPATCHER(tag_int64, (int64_t) value);
+        }
 		else if ($self->has(tag_float64))
 			$self->put(tag_float64, (float64_t)value);
 		else
--- a/src/shogun/base/SGObject.cpp
+++ b/src/shogun/base/SGObject.cpp
@@ -1122,3 +1122,5 @@ void CSGObject::push_back(CDynamicObjectArray* array, CSGObject* value)
 	ASSERT(array);
 	array->push_back(value);
 }
+
+stringToEnumMapType CSGObject::m_enum_to_string_map;
--- a/src/shogun/base/SGObject.cpp
+++ b/src/shogun/base/SGObject.cpp
@@ -1122,5 +1122,3 @@ void CSGObject::push_back(CDynamicObjectArray* array, CSGObject* value)
 	ASSERT(array);
 	array->push_back(value);
 }
-
-stringToEnumMapType CSGObject::m_enum_to_string_map;
--- a/src/shogun/base/SGObject.cpp
+++ b/src/shogun/base/SGObject.cpp
@@ -969,7 +969,19 @@ std::string CSGObject::to_string() const
 	{
 		ss << it->first.name() << "=";
 		auto value = it->second.get_value();
-		if (value.visitable())
+		if (m_string_to_enum_map.find(it->first.name()) !=
+		    m_string_to_enum_map.end())
+		{
+			auto param_enum_map = m_string_to_enum_map.at(it->first.name());
+			auto enum_value = any_cast<machine_int_t>(value);
+			auto enum_map_it = std::find_if(
+			    param_enum_map.begin(), param_enum_map.end(),
+			    [&enum_value](const std::pair<std::string, machine_int_t>& p) {
+				    return p.second == enum_value;
+			    });
+			ss << enum_map_it->first;
+		}
+		else if (value.visitable())
 		{
 			value.visit(visitor.get());
 		}
--- a/src/shogun/base/SGObject.cpp
+++ b/src/shogun/base/SGObject.cpp
@@ -1102,8 +1102,7 @@ CSGObject* CSGObject::get(const std::string& name) const noexcept(false)
 		{
 			SG_ERROR(
 			    "Cannot get parameter %s::%s of type %s as object, not a "
-			    "shogun "
-			    "object base type.\n",
+			    "shogun object base type.\n",
 			    get_name(), name.c_str(),
 			    self->map[BaseTag(name)].get_value().type().c_str());
 		}
--- a/src/shogun/base/SGObject.cpp
+++ b/src/shogun/base/SGObject.cpp
@@ -50,8 +50,8 @@ namespace shogun
 {
 
 	typedef std::map<BaseTag, AnyParameter> ParametersMap;
-	typedef std::unordered_map<std::string,
-	                           std::pair<SG_OBS_VALUE_TYPE, std::string>>
+	typedef std::unordered_map<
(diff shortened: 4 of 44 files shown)


karlnapf:
Minor: It would be cool if the visitor pattern that prints the instances would also look up the string for the enum so that users would see that rather than the int

karlnapf:
Same with get actually, so that the enums would be strings only in the interfaces

src/shogun/base/SGObject.h
```diff
@@ -390,9 +393,34 @@ class CSGObject
...
 
+    /** Typed setter for an enum mapped with a string, identified by a name.
+     *
+     * @param name name of the parameter
+     * @param value value of the parameter
+	 */
+	template <class T>
+	void put(const std::string& name, const std::string& value) noexcept(false)
+	{
+		if (m_enum_to_string_map.find(get_name()) == m_enum_to_string_map.end())
```
karlnapf: Would remove and just keep the below error. Same as above,  only confuses users, it’s more an internal thing

src/shogun/base/SGObject.h
```diff
@@ -390,9 +393,34 @@ class CSGObject
...
+
+		auto string_to_enum_map = m_enum_to_string_map[get_name()];
+
+		if (string_to_enum_map.find(name) == string_to_enum_map.end())
+			SG_ERROR("There is no enum mapping for the parameter %s::%s", get_name(), name);
+
+		auto string_to_enum = string_to_enum_map[name];
+
+		if (string_to_enum.find(value) == string_to_enum.end())
+			SG_ERROR("There is no enum mapping for '%s' in %s::%s", value, get_name(), name);
```
karlnapf: Thinking this further: I wouldnt even talk about enum mappings, but just “options”. Users don’t need to know that there is an underlying enum

src/shogun/regression/svr/LibLinearRegression.cpp
```diff
@@ -43,11 +45,19 @@ void CLibLinearRegression::init_defaults()
...
+	SG_ADD(
+		&m_epsilon, "m_tube_epsilon", "svr tube epsilon",
+		ParameterProperties::HYPER);
+	SG_ADD(&m_max_iter, "max_iter", "max number of iterations");
+	SG_ADD(&m_use_bias, "use_bias", "indicates whether bias should be used");
+	SG_ADD(
+		(machine_int_t*)&m_liblinear_regression_type,
+		"liblinear_regression_type", "Type of LibLinear regression.");
+	m_enum_to_string_map[get_name()]["liblinear_regression_type"] = {
+		{"L2R_L2LOSS_SVR", 0},
```
karlnapf: Use the enum value here rather than the int to make it readable, and they are available here

src/shogun/regression/svr/LibLinearRegression.cpp
```diff
@@ -43,11 +45,19 @@ void CLibLinearRegression::init_defaults()
...
+	SG_ADD(&m_C, "C", "regularization constant", ParameterProperties::HYPER);
+	SG_ADD(
+		&m_epsilon, "m_tube_epsilon", "svr tube epsilon",
+		ParameterProperties::HYPER);
+	SG_ADD(&m_max_iter, "max_iter", "max number of iterations");
+	SG_ADD(&m_use_bias, "use_bias", "indicates whether bias should be used");
+	SG_ADD(
+		(machine_int_t*)&m_liblinear_regression_type,
+		"liblinear_regression_type", "Type of LibLinear regression.");
+	m_enum_to_string_map[get_name()]["liblinear_regression_type"] = {
```
karlnapf: Actually, since this happens at runtime now, we could rather have one options map for each class, stored locally, and then a method “add_option(string, int)” that is called this method here. Or even better
“add_options(string param, map<string, int>)” which then can be called by directly writing out the map as {“foo”: SOLVER::FOO, “bar”: SOLVER::BAR} in a single call
I didn’t realise yesterday that the map doesn’t need to be gobal nor static

src/shogun/regression/svr/LibLinearRegression.h
```diff
@@ -147,7 +147,7 @@ class CLibLinearRegression : public CLinearMachine
 protected:
 
 		/** train machine */
-		virtual bool train_machine(CFeatures* data = NULL);
+		bool train_machine(CFeatures* data = NULL) override;
```
karlnapf: Mixing using override and not using it causes warnings iirc

src/shogun/regression/svr/LibLinearRegression.cpp
gf712: Ah yes good point, I didn't quite like the fact that there was this static variable around! So `add_options(string param, map<string, int>)` just sits in SGObject, and then just need a mapping from param to string_enum_map, which makes it much simpler.

src/shogun/regression/svr/LibLinearRegression.h
gf712: ah ok, but at some point would probably start adding override no? Doesn't it come at a cost when you just add more and more virtual class members?

karlnapf: Yes at some point. I would just leave this for another or or the detox gsoc project

src/shogun/regression/svr/LibLinearRegression.cpp
karlnapf: Or maybe even: Do you think using a macro, we could avoid having to write the enum value name AND a string ?
Like
ADD_OPTION(param_name, FOO); which adds the mapping “FOO”->FOO where FOO is one option of the enum?

karlnapf: This means we’d have to write things one by one but on the other hand it avoids the redundancy of explicitly naming the option string

gf712: You should be able to pass the literal variable name in a macro yes. It's a shame there isn't a universal name demangling across compilers, would be quite useful in this case...

karlnapf: yes, c++ lacks quite some things ... :scream:

src/interfaces/swig/SGBase.i
```diff
@@ -420,7 +420,27 @@ namespace shogun
...
+        }
+
+        std::vector<std::string> param_options(const std::string& name) const {
+            std::vector<std::string> result;
+
+            auto param_to_enum_map = $self->get_string_to_enum_map();
+
+            if (param_to_enum_map.find(name) == param_to_enum_map.end())
+			{
+                SG_SERROR("There are no options for the parameter '%s'", name.c_str());
```
karlnapf: super minor: maybe " ..for parameter %s::%s" and print classname

src/shogun/base/SGObject.h
```diff
@@ -100,6 +103,16 @@ template <class T> class SGStringList;
 
 #define SG_ADD(...) VARARG(SG_ADD, __VA_ARGS__)
 
+#define VALUE_TO_STRING_MACRO(s) #s
```
karlnapf: long live the preprocessor :D

karlnapf: it is so funny that this old school stuff still is necessary/useful in places

src/shogun/base/SGObject.h
```diff
@@ -100,6 +103,16 @@ template <class T> class SGStringList;
...
+
+#define SG_ADD_OPTION(param_name, enum_value)                                  \
+	{                                                                          \
+		static_assert(                                                         \
+		    std::is_enum<decltype(enum_value)>::value, "Expected an enum!");   \
+		m_string_to_enum_map[param_name][VALUE_TO_STRING_MACRO(enum_value)] =  \
+		    enum_value;                                                        \
+	}
+
```
karlnapf: pls undefine your helper macro after having used it

src/shogun/base/SGObject.h
```diff
@@ -390,9 +403,38 @@ class CSGObject
...
+				"There are no options for the parameter %s::%s", get_name(),
+				name.c_str());
+		}
+
+		auto string_to_enum = m_string_to_enum_map[name];
+
+		if (string_to_enum.find(value) == string_to_enum.end())
+		{
+			SG_ERROR(
+				"There is no enum mapping for '%s' in %s::%s",
```
karlnapf: could you rather say something like "illegal option %s for parameter %s::%s" ... ?

src/shogun/regression/svr/LibLinearRegression.cpp
```diff
@@ -43,11 +45,18 @@ void CLibLinearRegression::init_defaults()
...
+	SG_ADD(
+		&m_epsilon, "m_tube_epsilon", "svr tube epsilon",
+		ParameterProperties::HYPER);
+	SG_ADD(&m_max_iter, "max_iter", "max number of iterations");
+	SG_ADD(&m_use_bias, "use_bias", "indicates whether bias should be used");
+	SG_ADD(
+		(machine_int_t*)&m_liblinear_regression_type,
+		"liblinear_regression_type", "Type of LibLinear regression.");
+	SG_ADD_OPTION("liblinear_regression_type", L2R_L2LOSS_SVR);
+    SG_ADD_OPTION("liblinear_regression_type", L2R_L1LOSS_SVR_DUAL);
```
karlnapf: whitespace going wrong here

karlnapf: sorry about this, but what about actually merging all those lines into one with a multi-argument macro?
`SG_ADD_OPTION("liblinear_regression_type", L2R_L2LOSS_SVR, L2R_L1LOSS_SVR_DUAL)`

gf712: I am not sure that would work though. Because you would then have a recursive macro, which then needs to be manually coded, i.e. `SG_ADD_OPTION_1`, `SG_ADD_OPTION_2`, `SG_ADD_OPTION_3` right?

src/shogun/base/SGObject.h
gf712: Unfortunately I can't :( the macro is needed by `SG_ADD_OPTION`, which is used later on

src/shogun/base/SGObject.h
gf712: OK! Was just trying to follow the pattern from the other error messages

src/shogun/base/SGObject.h
gf712: yes unfortunately this is the simplest way..

src/shogun/base/SGObject.h
karlnapf: I see, well ok

karlnapf:
I think given that we only ever have a few options, why not do the recursive manual macro thing somewhere hidden? Say for up to 5 for a start? It’s not pretty but it’s only at one place whereas the repeated parametername would be in dozens of places

gf712:
> Same with get actually, so that the enums would be strings only in the interfaces

So for the C++ API we would still get an enum/int with get?

gf712:
> I think given that we only ever have a few options, why not do the recursive manual macro thing somewhere hidden? Say for up to 5 for a start? It’s not pretty but it’s only at one place whereas the repeated parametername would be in dozens of places

hmmm ok, I'll write something for that and think if there is something nicer we could do here which would be more sustainable

src/shogun/base/SGObject.cpp
```diff
@@ -958,7 +958,19 @@ std::string CSGObject::to_string() const
 	{
 		ss << it->first.name() << "=";
 		auto value = it->second.get_value();
-		if (value.visitable())
+		if (m_string_to_enum_map.find(it->first.name()) !=
```
gf712: is this fine for the object representation? It works fine

karlnapf: In c++ it would have to be int as the enum definitions go hide in the plugins. So quite unsafe actually. No harm in keeping the get for that for now but also offer the one that returns a string.... need to see how it plays out

karlnapf: How does the output look like?

gf712: `LibLinearRegression(C=1,bias=0,data_locked=false,features=null,labels=null,liblinear_regression_type=L2R_L2LOSS_SVR,m_tube_epsilon=0.01,max_iter=10000,max_train_time=0,solver_type=0,store_model_features=false,use_bias=false,w=Vector<double>(0): [])
`

src/shogun/base/SGObject.h
gf712: it seems like it might be possible to demangle it https://github.com/google/glog/blob/5c576f78c49b28d89b23fbb1fc80f54c879ec02e/src/demangle.cc#L1042, but would take a while to strip out the code we would need

gf712:
@karlnapf I don't really like the fact that we need to continuously add more macros, but this works for now. Until c++ has support for enum to string conversions, or we add such a library, or write our own implementation, this will have to do...

src/shogun/base/macros.h
```diff
@@ -20,8 +20,95 @@
...
+#else
+
+#define VA_NARGS_IMPL(_1, _2, _3, _4, _5, _6, _7, _8, N, ...) N
+#define VA_NARGS(...) VA_NARGS_IMPL(__VA_ARGS__, 8, 7, 6, 5, 4, 3, 2, 1)
+
+#endif
+
+#define VARARG_IMPL2(base, count, ...) base##count(__VA_ARGS__)
+#define VARARG_IMPL(base, count, ...) VARARG_IMPL2(base, count, __VA_ARGS__)
+#define VARARG(base, ...) VARARG_IMPL(base, VA_NARGS(__VA_ARGS__), __VA_ARGS__)
```
gf712: @karlnapf I added these macros to this header because I need them for SG_ADD_OPTIONS

karlnapf: Sure thing

src/shogun/classifier/svm/LibLinear.cpp
```diff
@@ -52,15 +52,16 @@ void CLibLinear::init()
...
 	SG_ADD(&epsilon, "epsilon", "Convergence precision.");
-	SG_ADD(
-	    &max_iterations, "max_iterations", "Max number of iterations.");
+	SG_ADD(&max_iterations, "max_iterations", "Max number of iterations.");
 	SG_ADD(&m_linear_term, "linear_term", "Linear Term");
 	SG_ADD(
 	    (machine_int_t*)&liblinear_solver_type, "liblinear_solver_type",
 	    "Type of LibLinear solver.");
+	SG_ADD_OPTIONS(
+	    "liblinear_solver_type", L2R_LR, L2R_L2LOSS_SVC_DUAL, L2R_L2LOSS_SVC,
```
karlnapf: This doesn’t look too bad now! :)

src/shogun/base/macros.h
```diff
@@ -20,8 +20,95 @@
...
+
+#define VALUE_TO_STRING_MACRO(s) #s
+
+#define SG_ADD_OPTION(param_name, enum_value)                                  \
+	{                                                                          \
+		static_assert(                                                         \
+		    std::is_enum<decltype(enum_value)>::value, "Expected an enum!");   \
+		m_string_to_enum_map[param_name][VALUE_TO_STRING_MACRO(enum_value)] =  \
```
karlnapf: I think we should also do a runtime assertion that the parameter was actually registered, so that only options for already registered parameters can be added. 
Another point would be to assert that the registered parameter is in fact a machine_int_t so the we cannot register options for other types of parameters

gf712:
I'm also trying to get the tag dispatching to work such that it is possible to get a string for the enums instead of a `int` in the SWIG interfaces..

src/shogun/base/macros.h
gf712: the problem with asserting if a parameter is of type `machine_int_t` is that we can't tell the difference from type `int32_t`, so the check will not work all the time..

gf712:
I am not sure how to get `get` to work with SWIG, because it uses partial specialisation. Basically we would have to use template deduction and SWIG doesn't support it yet. I tried a bit with tag dispatching but the issue is the same, we can only tell if an int should be returned as an int or std::string after checking the enum_string map, at runtime...

karlnapf:
Re asserting the correct type I think one way might be to actually combined the registration with adding the option. Then we can force the right enum type and inside hat method cast it to int, register ad parameter and then attach the options. See my point?

karlnapf:
Re the get o think first checking the map and then the parameters might be ok no?

gf712:
> Re asserting the correct type I think one way might be to actually combined the registration with adding the option. Then we can force the right enum type and inside hat method cast it to int, register ad parameter and then attach the options. See my point?

yup, I was thinking about something along those lines, but then will have to change SG_ADD a bit, and agree on a good signature.

> Re the get o think first checking the map and then the parameters might be ok no?

But need to have a get that has accepts two types: the cast and then the return type, because in this case the value is casted to an int, but then we need to lookup in the map and return a string. this is easy with partial specialisation of get, but SWIG doesn't support that. what I had (and worked fine) was 
```cpp
template <typename T, typename U = void, typename std::enable_if_t<std::is_same<T, machine_int_t>::value && std::is_same<U, std::string>::value>* = nullptr>
    std::string get(const std::string& name) const noexcept(false)
    {
        Tag<T> tag(name);

        if (m_string_to_enum_map.find(name) != m_string_to_enum_map.end())
        {
            return string_enum_reverse_lookup(name, get(tag));
        }
        else
        {
            SG_ERROR("There are no options in %s::%s", get_name(), name.c_str());
        }
        return nullptr;
    }
```
and then for example in the to_string I had something like this:
```
auto value = get<machine_int_t, std::string>(param);
```
but this can't be templated in shogun.i along with the other `get` (I tried %template() get<machine_int_t, std::string>) which I am guessing is related to [this](http://swig.10945.n7.nabble.com/type-traits-handling-and-specialization-td15207.html)

karlnapf:
> > Re asserting the correct type I think one way might be to actually combined the registration with adding the option. Then we can force the right enum type and inside hat method cast it to int, register ad parameter and then attach the options. See my point?
> 
> yup, I was thinking about something along those lines, but then will have to change SG_ADD a bit, and agree on a good signature.
What about just calling sgadd from inside the option macro and remove the other sgadd call?
> 
> > Re the get o think first checking the map and then the parameters might be ok no?
> 
> But need to have a get that has accepts two types: the cast and then the return type, because in this case the value is casted to an int, but then we need to lookup in the map and return a string. this is easy with partial specialisation of get, but SWIG doesn't support that. what I had (and worked fine) was
> 
> ```c++
> template <typename T, typename U = void, typename std::enable_if_t<std::is_same<T, machine_int_t>::value && std::is_same<U, std::string>::value>* = nullptr>
>     std::string get(const std::string& name) const noexcept(false)
>     {
>         Tag<T> tag(name);
> 
>         if (m_string_to_enum_map.find(name) != m_string_to_enum_map.end())
>         {
>             return string_enum_reverse_lookup(name, get(tag));
>         }
>         else
>         {
>             SG_ERROR("There are no options in %s::%s", get_name(), name.c_str());
>         }
>         return nullptr;
>     }
> ```
> 
> and then for example in the to_string I had something like this:
> 
> ```
> auto value = get<machine_int_t, std::string>(param);
> ```
> but this can't be templated in shogun.i along with the other `get` (I tried %template() get<machine_int_t, std::string>) which I am guessing is related to [this](http://swig.10945.n7.nabble.com/type-traits-handling-and-specialization-td15207.html)
Ah I see.... so what is the Second best option? I have the feeling one could do some hacks that involve adding a try error approach in python code, and do the mapping there in the lines how you made get return different types.... while that’s again python specific code it would just be a small extension of what’s there and the other Lang’s have to live it int for now

karlnapf:
Ah sorry phone editing is a pain

gf712:
Well we can extend SGObject in SGBase.i with a get_enum()?

gf712:
@karlnapf so this looks like this:
```python
In [1]: import shogun as sg                                                                                                                                                                                                                                              

In [2]: lr = sg.machine("LibLinearRegression")                                                                                                                                                                                                                           

In [3]: lr.get("liblinear_regression_type")                                                                                                                                                                                                                              
Out[3]: 'L2R_L1LOSS_SVR_DUAL'
```
Just need to check get_enum before the get_int as is done in the python code in `swig_typemaps.i`. Not very elegant, but it works with the current limited support of C++11 in SWIG

src/shogun/base/SGObject.h
```diff
@@ -775,6 +795,16 @@ class CSGObject
...
 	virtual CSGObject* clone();
 
+	/**
+	 * Looks up the option name of a parameter given the enum value.
+	 *
+	 * @param param the parameter name
+	 * @param value the enum value to query
+	 * @return the string representation of the enum (option name)
+	 */
+	std::string string_enum_reverse_lookup(
```
karlnapf: Is this public?

src/shogun/classifier/svm/LibLinear.cpp
```diff
@@ -52,15 +52,16 @@ void CLibLinear::init()
...
 	SG_ADD(&C2, "C2", "C Cost constant 2.", ParameterProperties::HYPER);
-	SG_ADD(
-	    &use_bias, "use_bias", "Indicates if bias is used.");
+	SG_ADD(&use_bias, "use_bias", "Indicates if bias is used.");
 	SG_ADD(&epsilon, "epsilon", "Convergence precision.");
-	SG_ADD(
-	    &max_iterations, "max_iterations", "Max number of iterations.");
+	SG_ADD(&max_iterations, "max_iterations", "Max number of iterations.");
 	SG_ADD(&m_linear_term, "linear_term", "Linear Term");
 	SG_ADD(
```
karlnapf: This would be removed or?

src/shogun/base/SGObject.h
gf712: Yes, because I need it in SWIG

gf712: Can always do something like
```cpp
#ifdef SWIG 
public:
#else 
private:
#endif

**Comment by gf712** [comment]
> Cool this is coming along! :)
> 
> Only thing would be to combine the adding options and registering parameter?

Yes, I was thinking we could have a call to SG_ADD inside SG_ADD_OPTIONS? Otherwise we need to write a SG_ADDN until 12 or so...

> And maybe a meta example coverage (in the base api example) only need to cover setting options as getting is more messy in meta examples, although it should be possible as we can substitute method calls. In fact you could add a method to swig that is ‘get_option’ and then use that in meta examples and substitute it with just get in python

I have the `get_enum`, but I can rename it to `get_option`?

File: src/shogun/classifier/svm/LibLinear.cpp (hunk 16 above)
Comment (gf712) [reply src/shogun/classifier/svm/LibLinear.cpp]: You mean it would be removed if we merge SG_ADD and SG_ADD_OPTIONS?

Comment (karlnapf) [reply src/shogun/classifier/svm/LibLinear.cpp]: Yes, to further remove redundant code

**Comment by karlnapf** [comment]
Option is more user friendly. Python users might not even know what an enum is :D

**Comment by karlnapf** [comment]
Cannot guard with swig as I changed the lib based on how it’s compiled (bad) the method should rather live in the swig parts of is no used from inside

**Comment by gf712** [comment]
> Cannot guard with swig as I changed the lib based on how it’s compiled (bad) the method should rather live in the swig parts of is no used from inside

I use it inside in the `to_string()` method

**Comment by karlnapf** [comment]
Ah of course. Ok then!

**Comment by gf712** [comment]
Combining SG_ADD with SG_ADD_OPTIONS is actually quite difficult because we need to tell apart SG_ADD3 from SG_ADD4 given the argument types. I can't think of a way that doesn't involve `if constexpr` right now... 
Trying something like `SG_ADD_OPTIONS(param, name, description, param_properties, ...)` and then check if param_properties is of type `ParameterProperties` (so have `SG_ADD(param, name, description, param_properties)`) or if it is a regular enum, in which case have `SG_ADD(param, name, description)` and pass `param_properties` onwards to `VARARG(SG_ADD_OPTION, param_properties, __VA_ARGS__)`

**Comment by karlnapf** [comment]
I’m sorry for asking all these things, sounds painful. I don’t have an idea how to do this in a better way right now.
Maybe don’t bother distinguishing the 3 and 4 argument add for options?
It’s not the biggest problem if we didn’t combine them in the end, my idea was morenin the “nice to have if it’s reasonable” corner.
Maybe let it sit for a day ti see whether a good solution comes to mind and if not we keep it as it is?
Anything else that needs to be addressed still? Example maybe?

**Comment by gf712** [comment]
> I’m sorry for asking all these things, sounds painful. I don’t have an idea how to do this in a better way right now.

No problem! I like this problem!

> Maybe don’t bother distinguishing the 3 and 4 argument add for options?

I can do that and then maybe we can improve it in c++17?

> Anything else that needs to be addressed still?

It seems like msvc doesn't like it but I am not sure why yet...

> A meta example would be ace.... to see this in action and to have it covered for all Lang’s ci

Yup, good idea, I had forgotten about that.

**Comment by karlnapf** [comment]
Fixing in C++17 would be good! You could put a note or something and until then we go with some horrible copy paste monster?

**Comment by gf712** [comment]
For the meta examples is there a way to do something like expect_exception or comparing results? I wanted to check if giving a non existing enum raises an exception and then check if changing with a correct enum I get the correct string back

**Comment by karlnapf** [comment]
No, just show how they are used. Ie set an option and retrieve an option. There is a base api example for such things. No need to change the tests

File: src/shogun/base/SGObject.h [hunk 17]
```diff
@@ -390,9 +375,38 @@ class CSGObject
...
 	}
 
+	/** Typed setter for an enum mapped with a string, identified by a name.
+	 *
+	 * @param name name of the parameter
+	 * @param value value of the parameter
+	 */
+	template <class T>
+	void
+	put(const std::string& name, const std::string& value) noexcept(false)
```
karlnapf: Just realising. We might at some point have string parameters....

src/shogun/classifier/LDA.cpp
```diff
@@ -36,22 +36,25 @@ CLDA::CLDA(
...
 	m_bdc_svd = true;
 
 	SG_ADD(
-	    (machine_int_t*)&m_method, "m_method",
-	    "Method used for LDA calculation");
-	SG_ADD(&m_gamma, "m_gamma", "Regularization parameter", ParameterProperties::HYPER);
+	    &m_gamma, "m_gamma", "Regularization parameter",
+	    ParameterProperties::HYPER);
 	SG_ADD(&m_bdc_svd, "m_bdc_svd", "Use BDC-SVD algorithm");
+	SG_ADD_OPTIONS(
```
karlnapf: Looks good!

src/shogun/base/SGObject.h
gf712: Ah but that's fine. We can just check if the parameter is in the map, and if not dispatch it with `CSGObject::put(const Tag<T>& _tag, const T& value)` inside this function.

examples/meta/src/base_api/put_get_add.sg
```diff
@@ -33,6 +33,8 @@ Machine svm = machine("LibSVM")
 svm.put("kernel", k)
 Kernel k2 = as_kernel(svm.get("kernel"))
 svm.put("kernel", k2)
+svm.put("libsvm_solver_type", 2)
```
karlnapf: Should put a comment saying that both are the same

gf712:
> Is the windows fail related?
> Otherwise we can merge once ci is ready. Maybe pester Sergey in irc a bit so he is a look as well

Ahhh it's finally working, this macro business is a pain in msvc.... The issue is unrelated, it seems like cblas.h is not found which I saw in another PR recently. I need to check what is happening there, but it's a CI issue.

examples/meta/src/base_api/put_get_add.sg
gf712: OK! Btw is it possible to test if get is working? I guess I need to add `get_option` to the parser?

karlnapf: Yes should be. Just via changing example and the target Json filed to potentially replace some method call strings

src/interfaces/swig/shogun.i
```diff
@@ -136,9 +159,13 @@ namespace shogun
 		Tag<float64_t> tag_float64(name);
 
 		if ($self->has(tag_int32))
-			$self->put(tag_int32, (int32_t)value);
+		{
+			PUT_ENUM_INT_DISPATCHER(tag_int32, (int32_t) value);
```
karlnapf: ah these are smart, didnt see them before

gf712: I didn't want to use a macro but swig didn't like the templated functions I wrote so I just went with this...

gf712: I could move this to the library actually and then it is possible to this check in C++ as well?

gf712: actually could be a bit of a pain to do it with templating...

karlnapf: i think it is fine, we can do that if needed

src/shogun/kernel/string/WeightedDegreePositionStringKernel.cpp
```diff
@@ -1877,90 +1877,98 @@ void CWeightedDegreePositionStringKernel::load_serializable_post() throw (Shogun
...
+	    "If block computation shall be used.");
+	SG_ADD(
+	    &which_degree, "which_degree",
+	    "The selected degree. All degrees are used by default (for value -1).",
+	    ParameterProperties::HYPER);
+	SG_ADD((CSGObject**)&alphabet, "alphabet", "Alphabet of Features.");
+	SG_ADD_OPTIONS(
+	    (machine_int_t*)&type, "type", "WeightedDegree kernel type.",
+	    ParameterProperties::HYPER, E_WD, E_EXTERNAL, E_BLOCK_CONST,
+	    E_BLOCK_LINEAR, E_BLOCK_SQPOLY, E_BLOCK_CUBICPOLY, E_BLOCK_EXP,
```
lisitsyn: Could it be something like `SG_OPTIONS(E_EXTERNAL, E_BLOCK_CONST, E_BLOCK_LINEAR, E_BLOCK_SQPOLY, E_BLOCK_CUBICPOLY, E_BLOCK_EXP)` for readability?

gf712:
@karlnapf it seems to be working fine now on all platforms. the only issue is that linear regression is sometimes causing issues on clang, but that is an unrelated issue.

PR #4538, opened 2019-02-21
Add string lists to integration testing

karlnapf:
Implicitly tests all in/out typemaps for string lists via the string_char meta example

### DIFF
--- a/examples/meta/generator/translate.py
+++ b/examples/meta/generator/translate.py
@@ -76,7 +76,8 @@ def getBasicTypesToStore():
 
 def getSGTypesToStore():
     """ Returns all SG* types which will be serialized """
-    return ("RealVector","RealMatrix","FloatVector","FloatMatrix")
+    return ("RealVector","RealMatrix","FloatVector","FloatMatrix",
+            "StringCharList")
 
 def getSGTypeToStoreMethodName(sgType):
     """ Translates given SG* type into meta language type """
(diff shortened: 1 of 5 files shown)


karlnapf:
@avramidis this tests your new typemap as well, it worked locally when I ran the integration test in octave, which means that cpp and octave string lists dumped to the test file are the same. This covers both outgoing typemaps (the one that is returned by `get`) and ingoing ones (when the variable is passed to the testing container at the end of each example).

avramidis:
Failed for csharp. I will have to take a look at that.

karlnapf:
Maybe the typemap is buggy. Not even sure who wrote that... good to have it uncovered though!

karlnapf:
It crashes haha.... ill also look into it locally a bit. I think it’s probably the ingoing typemap as the example didn’t crash before this pr when only the outgoing one was used.

karlnapf:
Ok I found the error.
The csharp typemaps for strings are a pandorras box

* length of array is written into first string in array using `sprintf` which is then parsed by csharp to allocate array space
* the error arose from `Marshal.PtrToStringAnsi` expecting a null terminted c-string (as the ANSI in the name suggests) but `%typemap(out) shogun::SGStringList<char>` did not provide that
* The thing is generally a big mess
* The typemap might have worked before if parts of the memory were initialized to 0 (it did for some strings). But it was never functional.
* So here we go: string list support in and out for csharp!

karlnapf:
C sharp passed!

karlnapf:
In a next step we can add int strings to the examples/tests. Having looked at the typemaps, I expect this not not work.
I’ll have a go at this in the next couple of days.

avramidis:
Good work @karlnapf !!! Let me know if you need help with that.

PR #4543, opened 2019-02-26
removed more kernels from SWIG

gf712:
here are a few more kernels removed from swig, just left normalisers, strings and director kernel. I am assuming normalisers and strings are the same principle?


gf712:
It seems like `CAUCKernel::setup_auc_maximization` needs to be exposed somehow to swig?

gf712:
Also need to expose `CAUCKernel::subkernel` as type CKernel, so that this works:
```python
subkernel = sg.kernel("GaussianKernel", log_width=1.7)
subkernel.init(feats_train, feats_train)
auc = sg.kernel("AUCKernel", subkernel=subkernel) 
```
Can we just register CAUCKernel::subkernel` as a CKernel**?
```cpp
SG_ADD((CKernel**) &subkernel, "subkernel", "The subkernel.",
	    ParameterProperties::HYPER);
```
Then the python snippet works...

karlnapf:
Yes expose it.
Function calls either pls lazily or via registering them with get

gf712:
@karlnapf what do you mean with "via registering them with get"?

karlnapf:
The “watch_method” retistration

gf712:
But that only works for functions without any arguments no?

karlnapf:
Yes

gf712:
But `CAUCKernel::setup_auc_maximization` takes one argument that is only known at runtime.

karlnapf:
Sorry I wasn’t clear earlier:

* either lazily. That is the method is called only in the public init function of the kernel when the argument is known. We have done this in many places to solve similar problems.

* or force the user to call it via registering the method and then have something like ‘kernel.run(“bla”, arg)’ This would require adding facility for that as watch_method currently doesn’t allow this.

I’d prefer the first option as the second opens door to all sorts of hackery.

See what I mean now?

gf712:
> Sorry I wasn’t clear earlier:
> 
> * either lazily. That is the method is called only in the public init function of the kernel when the argument is known. We have done this in many places to solve similar problems.
> * or force the user to call it via registering the method and then have something like ‘kernel.run(“bla”, arg)’ This would require adding facility for that as watch_method currently doesn’t allow this.
> 
> I’d prefer the first option as the second opens door to all sorts of hackery.
> 
> See what I mean now?

Ah ok I understand what you mean now. The issue with the first solution is that `CAUCKernel::setup_auc_maximization` calls init(lhs, rhs). So I am not sure how to do this?
Because I am assuming you meant this:
```
bool CAUCKernel::init(CFeatures* l, CFeatures* r)
{
        setup_auc_maximization(); // this would use a registered labels parameter
	CDotKernel::init(l, r);
	init_normalizer();
	return true;
}
```
but it wouldn't work...

karlnapf:
Then we might need the second option.... where is this used? Maybe there are other ways around this?

gf712:
Can only find it being used here https://github.com/shogun-toolbox/shogun/blob/e3e8cd197b4dde5a6c75136bea54e7fdc141a947/examples/meta/src/kernel/auc.sg#L9-L10
Can't think of anything right now... In my opinion this should be static, but that doesn't help in this situation...

karlnapf:
I think in this case the solution is easy. Give the kernel an additional member for the labels that are normally passed as the parameter of the method. Then inside `init`, you can put a runtime assertion that the labels have been set. So then users have to do
```
k = kernel("AUCKernel"
k.init(lhs, rhs) # runtime error
k.put("auc_maximization_labels", my_labels) # maybe a different name?
k.init(lhs,rhs) # works
```

gf712:
> I think in this case the solution is easy. Give the kernel an additional member for the labels that are normally passed as the parameter of the method. Then inside `init`, you can put a runtime assertion that the labels have been set. So then users have to do
> 
> ```
> k = kernel("AUCKernel"
> k.init(lhs, rhs) # runtime error
> k.put("auc_maximization_labels", my_labels) # maybe a different name?
> k.init(lhs,rhs) # works
> ```

but from what I understand is that `my_labels` is used to generate `lhs` and `rhs` right? So the second k.init(lhs, rhs) shouldn't work, but the first one could in theory... Basically there are two options to initialise:
 - with lhs, rhs:
```python
k = kernel("AUCKernel")
k.init(lhs, rhs)
```
 - with labels:
```python
k = kernel("AUCKernel")
k.put("auc_maximization_labels", my_labels)
k.init(lhs, rhs) # this works but you're overriding rhs, lhs generated with my_labels when function setup_auc_maximization(my_labels) is called
```

Init is called from here (inside setup_auc_maximization):
https://github.com/shogun-toolbox/shogun/blob/66e9864175f998050e4af1ab1c825bb94581ed95/src/shogun/kernel/AUCKernel.cpp#L110-L111

karlnapf:
Just looked at the method (didnt before, and didnt realise what it did)....
Not sure how to solve this as this call is basically needed and not part of the kernel API.
We cannot add the call to the API neither. The only thing we could do is to add the mentioned `void CSGObject::run(std::string)` which runs a method without. Then we can make the labels that are needed for this method a member that users can set using `put`. Then usage would look like

```
k = kernel("AUCKernel", auc_maximization_labels=my_labels)
k.run("setup_auc_maximization") # replaces the std init call
mat = k.get_kernel_matrix()
```

karlnapf:
A bit hacky, but it was clear that such a problem would occur at some point (and that we need to run methods by strings) ... @lisitsyn ?

gf712:
@karlnapf but this `run` method hasn't been implemented right? I can't find it anywhere..

karlnapf:
Yes would have to something new, inspired by the watch_method, or actually reusing that. Maybe it is just a matter of adding the run for registered methods with void return type. Ask @lisitsyn in irc he will have comments

gf712:
> Yes would have to something new, inspired by the watch_method, or actually reusing that. Maybe it is just a matter of adding the run for registered methods with void return type. Ask @lisitsyn in irc he will have comments

OK, I will ping him. I was just thinking maybe this can be done with the new fancy AUTO property? So when you compute the kernel matrix you call `init_auto_params` and this runs a function like `CAUCKernel::setup_auc_maximization`?

gf712:
Nevermind, that would require lhs and rhs to be marked as AUTO which doesn't work, or could be a bit hacky.

src/shogun/base/SGObject.h
```diff
@@ -817,13 +817,28 @@ class CSGObject
...
 		create_parameter(tag, AnyParameter(make_any(bind_method), properties));
 	}
+
+	/** Puts a pointer to a (lazily evaluated) function into the parameter map.
+	 *
+	 * @param name name of the parameter
+	 * @param method pointer to the method
+	 */
+	template <typename T, typename S>
+	void watch_method(const std::string& name, T (S::*method)())
```
gf712: @lisitsyn is it OK to have a watch_method for non const functions?

examples/meta/src/kernel/auc.sg
```diff
@@ -1,13 +1,13 @@
...
-BinaryLabels labels_train(f_labels_train)
+Labels labels_train = labels(f_labels_train)
 
 Kernel subkernel = kernel("GaussianKernel", log_width=1.7)
 subkernel.init(feats_train, feats_train)
 
-AUCKernel auc(0, subkernel)
-auc.setup_auc_maximization(labels_train)
+Kernel auc = kernel("AUCKernel", subkernel=subkernel, labels=labels_train)
+auc.get_int("setup_auc_maximization")
```
gf712: @karlnapf it is a bit awkward that it isn't `void`.. is it ok?

karlnapf: Lol yeah that’s possible. But I think it is awkward indeed. Having the get compute something lazily is ok imo as the user wants to have something. But here the user wants to execute something so I think there should be a special method for that, hence “run” which indicates a void method

src/shogun/base/SGObject.h
karlnapf: Ah I didn’t think about the function being non const of course. Another reason to have a special directive (run) as it might change the state of the object.

src/shogun/kernel/AUCKernel.cpp
```diff
@@ -16,8 +16,10 @@ using namespace shogun;
 void
 CAUCKernel::init()
 {
-	SG_ADD((CSGObject**) &subkernel, "subkernel", "The subkernel.",
+	SG_ADD((CKernel**) &subkernel, "subkernel", "The subkernel.",
 	    ParameterProperties::HYPER);
+	SG_ADD((CLabels**) &labels, "labels", "The labels.");
```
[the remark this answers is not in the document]
gf712: yup, sorry need to change that

src/shogun/base/SGObject.h
gf712: @karlnapf it seems to be straight forward, just have to decide how two void functions compare, i.e. if `f1` and `f2` are void functions is `f1() == f2()`?

src/shogun/base/SGObject.h
```diff
@@ -963,6 +991,14 @@ class CSGObject
 	 */
 	AnyParameter get_parameter(const BaseTag& _tag) const;
 
+	/** Getter for a class function, identified by a BaseTag.
+	 * Throws an exception if the class does not have such a parameter.
+	 *
+	 * @param _tag name information of parameter
+	 * @return value of the parameter identified by the input tag
+	 */
+	AnyParameter get_function(const BaseTag& _tag) const;
```
gf712: this function getter should guarantee that a registered non const class function cannot be run with `get` but instead has to be invoked using `run`

src/shogun/lib/any.h
```diff
@@ -273,6 +270,14 @@ namespace shogun {
 			return compare(lhs(), rhs());
 		}
 
+		template <>
+		bool compare_impl<void>(
+				maybe_most_important, const std::function<void()>& lhs,
+				const std::function<void()>& rhs)
+		{
+			return true;
```
gf712: @karlnapf for now I put `std::function<void()> == std::function<void()>` is true, but can argue either way tbh..

gf712:
ok, I changed things around, it was getting too complicated to cast a void function back, because any returns the value after invoking the function, rather than returning the function pointer. So I just created a void `run` method that checks the return value of a bool() function that is stored as Any value.

examples/meta/src/kernel/auc.sg
```diff
@@ -1,13 +1,13 @@
...
-BinaryLabels labels_train(f_labels_train)
+Labels labels_train = labels(f_labels_train)
 
 Kernel subkernel = kernel("GaussianKernel", log_width=1.7)
 subkernel.init(feats_train, feats_train)
 
-AUCKernel auc(0, subkernel)
-auc.setup_auc_maximization(labels_train)
+Kernel auc = kernel("AUCKernel", subkernel=subkernel, labels=labels_train)
+auc.run("setup_auc_maximization")
```
karlnapf: :dancer:

src/shogun/base/AnyParameter.h
```diff
@@ -39,7 +39,8 @@ namespace shogun
 		GRADIENT = 1u << 1,
 		MODEL = 1u << 2,
 		AUTO = 1u << 10,
-		READONLY = 1u << 11
+		READONLY = 1u << 11,
+		RUNFUNCTION = 1u << 12
```
karlnapf: we dont have a property for dynamic parameters (those that are registered with watch_method).
Should be unified or?

gf712: this makes a distinction between const functions, i.e. can be run with get, and non-const function which have side effects, so have to be called with run

examples/undocumented/python/kernel_combined.py
```diff
@@ -14,7 +14,7 @@ def kernel_combined (fm_train_real=traindat,fm_test_real=testdat,fm_train_dna=tr
 	from shogun import StringCharFeatures, CombinedFeatures, DNA
 	import shogun as sg
 
-	kernel=CombinedKernel()
+	kernel=sg.CombinedKernel()
```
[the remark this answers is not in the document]
gf712: I think the append function of CombinedKernel would have to be exposed, no?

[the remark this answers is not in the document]
gf712: nevermind, I figured it out, need to use add :D

gf712:
> You can use .add to append to array parameters, see mkl meta example
> OT: you think the name is good? rather “append” would be better?

I think if add always appends to the end of the array it should be called `append`

gf712:
@karlnapf I think this is ready to be merged

examples/undocumented/python/graphical/statistics_linear_time_mmd.py
```diff
@@ -44,9 +44,9 @@ def linear_time_mmd_graphical():
 	sigmas=[2**x for x in range(-3,10)]
 	widths=[x*x*2 for x in sigmas]
 	print "kernel widths:", widths
-	combined=CombinedKernel()
+	combined=sg.kernel("CombinedKernel")
 	for i in range(len(sigmas)):
-		combined.append_kernel(GaussianKernel(10, widths[i]))
+		combined.append_kernel(sg.kernel("GaussianKernel", log_width=widths[i]))
```
karlnapf: Add?

src/shogun/base/AnyParameter.h
karlnapf: Ah ok, maybe put some docstring that makes this clear?

src/shogun/kernel/AUCKernel.cpp
```diff
@@ -1,121 +1,123 @@
...
 
 CAUCKernel::~CAUCKernel()
 {
 	SG_UNREF(subkernel);
+	SG_UNREF(labels)
 	cleanup();
 }
 
-CLabels* CAUCKernel::setup_auc_maximization(CLabels* labels)
+bool CAUCKernel::setup_auc_maximization()
```
karlnapf: Not sure I understood this earlier... why does this return a book? Ie what happens with the return value when calling run?

src/shogun/base/SGObject.cpp
```diff
@@ -1067,7 +1089,7 @@ void CSGObject::init_auto_params()
 
 CSGObject* CSGObject::get(const std::string& name, index_t index) const
 {
-	auto* result = sgo_details::get_by_tag(this, name, std::move(sgo_details::GetByNameIndex(index)));
+	auto* result = sgo_details::get_by_tag(this, name, sgo_details::GetByNameIndex(index));
 	if (!result && has(name))
```
karlnapf: :saint:

src/shogun/base/SGObject.h
```diff
@@ -635,6 +635,21 @@ class CSGObject
...
+	/**
+	 *
+	 * @param name name of the parameter
+	 * @return value of the parameter corresponding to the input name and type
+	 */
+	void run(const std::string& name) const noexcept(false)
+	{
+		Tag<bool> tag(name);
+		auto param = get_function(tag);
+		if (!any_cast<bool>(param.get_value()))
```
karlnapf: Is that really a good idea? This Boolean business? Then we need to make sure that methods are actually written in a way that a false indicates a failure?

gf712: well the issue is that we can't store void functions with any, so this was a work around. Adding support for that would involve quite a bit of redesign I think..

karlnapf: I wonder: maybe this could be done with a return type that is just used for this case?

karlnapf: ok let's use it like that for now, and then refactor if it causes problems.

PR #4544, opened 2019-02-27
Remove Loss from swig

vinx13:


gf712:
Can you also remove from `Loss_includes.i`? Or does that break things?

PR #4547, opened 2019-02-28
Use MKL in windows build

gf712:
This fixes the blas header issue in the windows CI. There seems to be an issue with the include directories if not specified. This PR explicitly gives the path to the MKL libraries to Cmake which then can find the blas header file.


gf712:
@vigsterkr mkl_rt.dll causes a linker error, so I will just use the mkl_corelib instead.

PR #4548, opened 2019-03-01
Feature/correct initials

karlnapf:
For discussing on the diff

### DIFF
--- a/src/shogun/features/DotFeatures.cpp
+++ b/src/shogun/features/DotFeatures.cpp
@@ -43,12 +43,12 @@ CDotFeatures::CDotFeatures(CFile* loader)
 	init();
 }
 
-float64_t CDotFeatures::dense_dot_sgvec(int32_t vec_idx1, SGVector<float64_t> vec2)
+float64_t CDotFeatures::dense_dot_sgvec(int32_t vec_idx1, SGVector<float64_t> vec2) const
 {
 	return dense_dot(vec_idx1, vec2.vector, vec2.vlen);
 }
 
-void CDotFeatures::dense_dot_range(float64_t* output, int32_t start, int32_t stop, float64_t* alphas, float64_t* vec, int32_t dim, float64_t b)
+void CDotFeatures::dense_dot_range(float64_t* output, int32_t start, int32_t stop, float64_t* alphas, float64_t* vec, int32_t dim, float64_t b) const
 {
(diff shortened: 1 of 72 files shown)


src/shogun/features/DotFeatures.cpp
```diff
@@ -200,6 +200,44 @@ SGVector<float64_t> CDotFeatures::get_mean()
...
+	{
+		auto global_mean = linalg::sum(mean) / dim;
+		linalg::set_const(mean, global_mean);
+	}
+
+	linalg::scale(mean, mean, -1.0);
+	SGVector<float64_t> std(dim);
+	for (index_t i = 0; i < num; ++i)
+	{
+		auto m = mean.clone();
```
karlnapf: I think there might be a bug here: when the columnwise std-dev is computed, the column-wise mean would have to be substracted instead of the global one. Or am I confused?

vigsterkr: yeah.... and in case if colwise = true mean is the colwise mean

vigsterkr: note i use the same container (mean) to get the std-dev.... avoid code redundancy and since i need to use add_to_dense_vector....

src/shogun/lib/auto_initialiser.h
```diff
@@ -0,0 +1,64 @@
...
+			~GammaFeatureNumberInit() override = default;
+
+			Any operator()() override
+			{
+				REQUIRE(m_kernel != nullptr, "m_kernel is not pointing to a CKernel object");
+
+				auto lhs = m_kernel->get_lhs();
+				switch (lhs->get_feature_class())
+				{
+					case EFeatureClass::C_DENSE:
```
karlnapf: minor: might be worth to adding a filtering method to those enums, so that we can ask questions like ("is this a CDotFeatures subclass?"), then these checks are  a bit less verbose

src/shogun/features/DotFeatures.cpp
vigsterkr: see https://github.com/shogun-toolbox/shogun/pull/4548/files#diff-e3b574b8a2ef609bae51ca0e7298680cR214

karlnapf: nice! :)

PR #4549, opened 2019-03-01
Read only property

gf712:
@karlnapf Is this what you had in mind?

### DIFF
--- a/examples/meta/src/gaussian_process/regression.sg
+++ b/examples/meta/src/gaussian_process/regression.sg
@@ -11,7 +11,8 @@ Labels labels_test = labels(f_labels_test)
 #![create_features]
 
 #![create_appropriate_kernel_and_mean_function]
-Kernel k = kernel("GaussianKernel", lhs=features_train, rhs=features_train, log_width=0.0)
+Kernel k = kernel("GaussianKernel", log_width=0.0)
+k.init(features_train, features_train)
 MeanFunction mean = gp_mean("ZeroMean")
 #![create_appropriate_kernel_and_mean_function]
 
--- a/src/shogun/base/AnyParameter.h
+++ b/src/shogun/base/AnyParameter.h
@@ -37,7 +37,8 @@ namespace shogun
 		HYPER = 1u << 0,
 		GRADIENT = 1u << 1,
 		MODEL = 1u << 2,
-		AUTO = 1u << 10
+		AUTO = 1u << 10,
+		CONST = 1u << 11
 	};
 
 	enableEnumClassBitmask(ParameterProperties);
--- a/src/shogun/base/AnyParameter.h
+++ b/src/shogun/base/AnyParameter.h
@@ -38,7 +38,7 @@ namespace shogun
 		GRADIENT = 1u << 1,
 		MODEL = 1u << 2,
 		AUTO = 1u << 10,
-		CONST = 1u << 11
+		READONLY = 1u << 11
 	};
 
 	enableEnumClassBitmask(ParameterProperties);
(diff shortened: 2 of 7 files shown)


gf712:
We can't call the enum field `CONST` because msvc thinks we are using the keyword `const`...

gf712:
Should a constant parameter be `put`-able once for initialisation, or would it always be read-only?

src/shogun/base/AnyParameter.h
```diff
@@ -37,7 +37,8 @@ namespace shogun
 		HYPER = 1u << 0,
 		GRADIENT = 1u << 1,
 		MODEL = 1u << 2,
-		AUTO = 1u << 10
+		AUTO = 1u << 10,
+		CONST = 1u << 11
```
lisitsyn: `READ_ONLY` might be a better name to avoid clashes

karlnapf:
I think always read only, never put able. And then we try to use factories.... let’s see if that approach works at least. Can you mark feature matrix of dense feats also read only?

src/shogun/base/AnyParameter.h
gf712: yup, I didn't realise that it could clash with C++ keywords...

gf712:
So should the kernel ,for example, only be initialised with `init(lhs,rhs)`, instead of `sg.kernel("name", lhs=lhs, rhs=rhs)`?

karlnapf:
Exactly! That’s the the point

examples/meta/src/gaussian_process/regression.sg
```diff
@@ -11,7 +11,8 @@ Labels labels_test = labels(f_labels_test)
 #![create_features]
 
 #![create_appropriate_kernel_and_mean_function]
-Kernel k = kernel("GaussianKernel", lhs=features_train, rhs=features_train, log_width=0.0)
+Kernel k = kernel("GaussianKernel", log_width=0.0)
```
karlnapf: yes!

src/shogun/kernel/Kernel.cpp
```diff
@@ -922,7 +922,8 @@ void CKernel::register_params()   {
 	    "Cache size in MB.");
 
 	SG_ADD(
-	    &lhs, "lhs", "Feature vectors to occur on left hand side.");
+		&lhs, "lhs", "Feature vectors to occur on left hand side.",
+		ParameterProperties::READONLY);
```
karlnapf: rhs should also be read-only

vigsterkr:
@karlnapf i'm just wondering in case of DenseFeatures the m_feature could be defined as a `const` and that should have the same effect... or?

karlnapf:
Unfortunately no. It was my preferred solution as well but the any framework makes the compiler freak out if we do that (e.g clone, put) and then when I talked to @lisitsyn and he had some other reasons why not to do that. This pr here is the result of our discussion

lisitsyn:
@vigsterkr I am just afraid adding constness would make things more complex. It could make sense in pure C++ but since we pass the runtime boundary in the interfaces it might be not that good.

lisitsyn:
Ah and this also nearly doubles the amount of generated template code.

src/shogun/kernel/Kernel.cpp
```diff
@@ -920,11 +920,14 @@ void CKernel::save_serializable_post() throw (ShogunException)
...
-	SG_ADD(&rhs, "rhs", "Feature vectors to occur on right hand side.");
 	SG_ADD(
-	    &lhs_equals_rhs, "lhs_equals_rhs",
-	    "If features on lhs are the same as on rhs.");
+		&lhs, "lhs", "Feature vectors to occur on left hand side.",
+		ParameterProperties::READONLY);
+	SG_ADD(
+		&rhs, "rhs", "Feature vectors to occur on right hand side.",
+		ParameterProperties::READONLY);
+	SG_ADD(&lhs_equals_rhs, "lhs_equals_rhs",
```
karlnapf: also readonly!

karlnapf:
CI problems?

PR #4550, opened 2019-03-01
Observable LARS and minimal example.

geektoni:
@karlnapf This PR contains the first start towards an observable LARS, plus a minimal example which shows the usage of it. Moreover, I've added an observer which works as a logger on screen.

### DIFF
--- a/examples/undocumented/libshogun/regression_lars_observable_path.cpp
+++ b/examples/undocumented/libshogun/regression_lars_observable_path.cpp
@@ -1,0 +1,74 @@
+/*
+* Written (W) 2019 Giovanni De Toni
+*/
+
+#include <shogun/features/DenseFeatures.h>
+#include <shogun/lib/SGMatrix.h>
+#include <shogun/regression/LeastAngleRegression.h>
+#include <shogun/labels/RegressionLabels.h>
+#include <shogun/mathematics/linalg/LinalgNamespace.h>
+#include <shogun/preprocessor/PruneVarSubMean.h>
+#include <shogun/preprocessor/NormOne.h>
+#include <shogun/lib/parameter_observers/ParameterObserverLogger.h>
+#include <shogun/base/init.h>
+#include <shogun/lib/common.h>
+
+using namespace shogun;
+
+// Data generator
+void generate_data_n_greater_d(SGMatrix<float64_t> &data, SGVector<float64_t> &lab)
+{
+	data(0,0)=0.044550005575722;
+	data(1,0)=-0.433969606728583;
+	data(2,0)=-0.397935396933392;
+	data(0,1)=-0.778754072066602;
+	data(1,1)=-0.620105076569903;
+	data(2,1)=-0.542538248707627;
+	data(0,2)=0.334313094513960;
+	data(1,2)=0.421985645755003;
+	data(2,2)=0.263031426076997;
+	data(0,3)=0.516043376162584;
+	data(1,3)=0.159041471773470;
+	data(2,3)=0.691318725364356;
+	data(0,4)=-0.116152404185664;
+	data(1,4)=0.473047565770014;
+	data(2,4)=-0.013876505800334;
+
+	lab[0]=-0.196155100498902;
+	lab[1]=-5.376485285422094;
+	lab[2]=-1.717489351713958;
+	lab[3]=4.506538567065213;
+	lab[4]=2.783591170569741;
+}
+
+int main() {
+
+	init_shogun_with_defaults();
+
+	SGMatrix<float64_t> data(3,5);
+	SGVector<float64_t> lab(5);
+	generate_data_n_greater_d(data, lab);
+
+	// Set the parameter observer
+	Some<ParameterObserverLogger> logger = some<ParameterObserverLogger>();
+
+	// Set features and labels
+	Some<CDenseFeatures<float64_t>> features= some<CDenseFeatures<float64_t>>(data);
+	Some<CRegressionLabels> labels= some<CRegressionLabels>(lab);
+
+	// Create LARS object
+	Some<CLeastAngleRegression> lars= some<CLeastAngleRegression>();
+
+	// Subscribe the observer to the LARS object
+	lars->subscribe_to_parameters(logger);
+
+	// Set labels and train
+	lars->set_labels((CLabels*) labels);
+	lars->train(features);
+
+	SGVector<float64_t> active3=SGVector<float64_t>(lars->get_w_for_var(3));
+	SGVector<float64_t> active2=SGVector<float64_t>(lars->get_w_for_var(2));
+	SGVector<float64_t> active1=SGVector<float64_t>(lars->get_w_for_var(1));
+
+	return 0;
+}
--- a/src/shogun/lib/parameter_observers/ObservedValue.h
+++ b/src/shogun/lib/parameter_observers/ObservedValue.h
@@ -70,7 +70,7 @@ namespace shogun
 		 * @param value Any-wrapped value of the param
 		 */
 		ObservedValue(
-		    int64_t step, std::string& name, Any value, SG_OBS_VALUE_TYPE type)
+		    int64_t step, std::string name, Any value, SG_OBS_VALUE_TYPE type)
 		    : m_step(step), m_name(name), m_value(value), m_type(type)
 		{
 		}
@@ -108,7 +108,7 @@ namespace shogun
 		 * Set the param's name
 		 * @param name
 		 */
-		void set_name(const std::string& name)
+		void set_name(const std::string name)
 		{
 			m_name = name;
 		}
@@ -157,7 +157,7 @@ namespace shogun
 		* @return an ObservedValue object initialized
 		*/
 		static ObservedValue
-		make_observation(int64_t step, std::string& name, Any value)
+		make_observation(int64_t step, std::string name, Any value)
 		{
 			return ObservedValue(step, name, value, TENSORBOARD);
 		}
--- a/src/shogun/lib/parameter_observers/ParameterObserverLogger.cpp
+++ b/src/shogun/lib/parameter_observers/ParameterObserverLogger.cpp
@@ -1,0 +1,34 @@
+/*
+* Written (W) 2019 Giovanni De Toni
+*/
+
+#include <shogun/lib/parameter_observers/ParameterObserverLogger.h>
+//#include <shogun/lib/type_case.h>
+
+using namespace shogun;
+
+ParameterObserverLogger::ParameterObserverLogger() {}
+
+ParameterObserverLogger::ParameterObserverLogger(std::vector<std::string> &parameters)
+		: ParameterObserverInterface(parameters) {}
+
+void ParameterObserverLogger::on_next(const TimedObservedValue &value) {
+
+	//auto f_print = [&value](auto _disp_value) {
+	//	SG_SPRINT("[%l] Received %s", convert_to_millis(value.second), _disp_value)
+	//};
+	// Dispatch the value and print it
+	//sg_any_dispatch(value.first.get_value(), sg_all_typemap, f_print, f_print, f_print);
+
+	SG_PRINT("[%l] Received an observed value with name \"%s\"",
+			 convert_to_millis(value.second), value.first.get_name().c_str());
+
+}
+
+void ParameterObserverLogger::on_error(std::exception_ptr ptr) {
+
+}
+
+void ParameterObserverLogger::on_complete() {
+
+}
--- a/src/shogun/lib/parameter_observers/ParameterObserverLogger.cpp
+++ b/src/shogun/lib/parameter_observers/ParameterObserverLogger.cpp
@@ -20,8 +20,7 @@ void ParameterObserverLogger::on_next(const TimedObservedValue &value) {
 	// Dispatch the value and print it
 	//sg_any_dispatch(value.first.get_value(), sg_all_typemap, f_print, f_print, f_print);
 
-	SG_PRINT("[%l] Received an observed value with name \"%s\"",
-			 convert_to_millis(value.second), value.first.get_name().c_str());
+	SG_PRINT("[%l] Received an observed value\n",convert_to_millis(value.second));
 
 }
 
--- a/src/shogun/lib/parameter_observers/ParameterObserverLogger.h
+++ b/src/shogun/lib/parameter_observers/ParameterObserverLogger.h
@@ -1,0 +1,40 @@
+/*
+* Written (W) 2019 Giovanni De Toni
+*/
+
+#ifndef SHOGUN_PARAMETEROBSERVERLOGGER_H
+#define SHOGUN_PARAMETEROBSERVERLOGGER_H
+
+#include <shogun/base/SGObject.h>
+#include <shogun/lib/parameter_observers/ParameterObserverInterface.h>
+
+namespace shogun {
+	/**
+	 * This class implements a logger which will basically print on screen all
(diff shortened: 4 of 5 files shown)


examples/undocumented/libshogun/regression_lars_observable_path.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+	// Create LARS object
+	Some<CLeastAngleRegression> lars= some<CLeastAngleRegression>();
+
+	// Subscribe the observer to the LARS object
+	lars->subscribe_to_parameters(logger);
+
+	// Set labels and train
+	lars->set_labels((CLabels*) labels);
+	lars->train(features);
```
geektoni: The output of this example is the following:
```
root@2bec4daebb48:~/shogun/examples/undocumented/libshogun# ./example
[%] Received an observed value
[%] Received an observed value
[%] Received an observed value
[WARN] In file /root/shogun/src/shogun/regression/LeastAngleRegression.cpp line 357: Convergence level (0.458377) not below tolerance (0.000000) after 3 iterations.
```
Ideally, each time an observed value is capture by the logger, it will print it on the terminal.

src/shogun/lib/parameter_observers/ParameterObserverLogger.cpp
```diff
@@ -0,0 +1,33 @@
...
+ParameterObserverLogger::ParameterObserverLogger(std::vector<std::string> &parameters)
+		: ParameterObserverInterface(parameters) {}
+
+void ParameterObserverLogger::on_next(const TimedObservedValue &value) {
+
+	//auto f_print = [&value](auto _disp_value) {
+	//	SG_SPRINT("[%l] Received %s", convert_to_millis(value.second), _disp_value)
+	//};
+	// Dispatch the value and print it
+	//sg_any_dispatch(value.first.get_value(), sg_all_typemap, f_print, f_print, f_print);
```
geektoni: The original idea of this logger was to print on screen (or on file) the string representation of the observed value.
However, I do not know if it is possible within Shogun (do we have a sort of `to_string()` method for all Shogun's objects?) or if it is feasible without using hacky ways.

src/shogun/regression/LeastAngleRegression.cpp
```diff
@@ -316,6 +316,9 @@ bool CLeastAngleRegression::train_machine_templated(CDenseFeatures<ST>* data)
 					map_Xa.block(0, i_kick+1, numRows, numCols-i_kick).eval();			
 		}
 
+		// Observe this value directly
+		observe(ObservedValue::make_observation(nloop, std::string("betas"), make_any(beta)));
```
geektoni: The issue I'm facing here is how do we explain/show to the user that he can observe this value `betas`? I thought that we could register somehow some "fake" parameters inside the class which cannot be updated/obtained with `put()`/`get()`, but they can only be observed by the user (e.g. `SG_ADD(&beta, "betas", "Weights of the model while training", ParameterProperties::OBSERVABLE);`). @karlnapf what do you think?

src/shogun/lib/parameter_observers/ParameterObserverLogger.cpp
```diff
@@ -0,0 +1,33 @@
+/*
+* Written (W) 2019 Giovanni De Toni
+*/
+
+#include <shogun/lib/parameter_observers/ParameterObserverLogger.h>
+//#include <shogun/lib/type_case.h>
```
geektoni: As a side not, I am not able to include this header file at all (that's why everything down is commented out). I keep getting this strange linker error here:
```
/usr/bin/ld.gold: error: CMakeFiles/libshogun.dir/lib/parameter_observers/ParameterObserverLogger.cpp.o: multiple definition of 'shogun::type_internal::print_map[abi:cxx11](std::unordered_map<std::type_index, shogun::TYPE, std::hash<std::type_index>, std::equal_to<std::type_index>, std::allocator<std::pair<std::type_index const, shogun::TYPE> > > const&)'
/usr/bin/ld.gold: CMakeFiles/libshogun.dir/io/TBOutputFormat.cpp.o: previous definition here
/usr/bin/ld.gold: error: CMakeFiles/libshogun.dir/lib/parameter_observers/ParameterObserverLogger.cpp.o: multiple definition of 'shogun::type_internal::get_type(shogun::Any const&, std::unordered_map<std::type_index, shogun::TYPE, std::hash<std::type_index>, std::equal_to<std::type_index>, std::allocator<std::pair<std::type_index const, shogun::TYPE> > > const&)'
/usr/bin/ld.gold: CMakeFiles/libshogun.dir/io/TBOutputFormat.cpp.o: previous definition here
```
Maybe the solution is straightforward but I can't see it :(

examples/undocumented/libshogun/regression_lars_observable_path.cpp
```diff
@@ -0,0 +1,74 @@
...
+#include <shogun/preprocessor/PruneVarSubMean.h>
+#include <shogun/preprocessor/NormOne.h>
+#include <shogun/lib/parameter_observers/ParameterObserverLogger.h>
+#include <shogun/base/init.h>
+#include <shogun/lib/common.h>
+
+using namespace shogun;
+
+// Data generator
+void generate_data_n_greater_d(SGMatrix<float64_t> &data, SGVector<float64_t> &lab)
```
karlnapf: couldnt we use some environment for regression? Then we'd have less code here

karlnapf: ah sorry this is libshogun example. What about actually loading regression data from a file?

examples/undocumented/libshogun/regression_lars_observable_path.cpp
```diff
@@ -0,0 +1,74 @@
...
+	data(2,4)=-0.013876505800334;
+
+	lab[0]=-0.196155100498902;
+	lab[1]=-5.376485285422094;
+	lab[2]=-1.717489351713958;
+	lab[3]=4.506538567065213;
+	lab[4]=2.783591170569741;
+}
+
+int main() {
```
karlnapf: Thinking even further, this example could be a meta example in fact! All of this is possible with the swig API or? Because if so, I would much prefer doing that to a libshogun example (we would like to get rid of as many as we can)

examples/undocumented/libshogun/regression_lars_observable_path.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+	init_shogun_with_defaults();
+
+	SGMatrix<float64_t> data(3,5);
+	SGVector<float64_t> lab(5);
+	generate_data_n_greater_d(data, lab);
+
+	// Set the parameter observer
+	Some<ParameterObserverLogger> logger = some<ParameterObserverLogger>();
```
karlnapf: in order to make this meta example compatible we would ne a factory for observers is `factory.h`, have a look, it is pretty self explaining. And then `ParameterObserver` needs to be exposed to swig (the other sublcasses shouldn't for now)

examples/undocumented/libshogun/regression_lars_observable_path.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+	// Set features and labels
+	Some<CDenseFeatures<float64_t>> features= some<CDenseFeatures<float64_t>>(data);
+	Some<CRegressionLabels> labels= some<CRegressionLabels>(lab);
+
+	// Create LARS object
+	Some<CLeastAngleRegression> lars= some<CLeastAngleRegression>();
+
+	// Subscribe the observer to the LARS object
+	lars->subscribe_to_parameters(logger);
```
karlnapf: i think the method name is a bit awkward. What about just `subscribe`?

examples/undocumented/libshogun/regression_lars_observable_path.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+	// Set labels and train
+	lars->set_labels((CLabels*) labels);
```
karlnapf: here you would use put, see other meta examples

examples/undocumented/libshogun/regression_lars_observable_path.cpp
```diff
@@ -0,0 +1,74 @@
...
+	lars->train(features);
+
+	SGVector<float64_t> active3=SGVector<float64_t>(lars->get_w_for_var(3));
```
karlnapf: no need to have this in the example, irrelevant or?

src/shogun/lib/parameter_observers/ObservedValue.h
```diff
@@ -70,7 +70,7 @@ namespace shogun
 		 * @param value Any-wrapped value of the param
 		 */
 		ObservedValue(
-		    int64_t step, std::string& name, Any value, SG_OBS_VALUE_TYPE type)
+		    int64_t step, std::string name, Any value, SG_OBS_VALUE_TYPE type)
```
karlnapf: const?

karlnapf: what is up with the SG_OBS_VALUE_TYPE here?

src/shogun/lib/parameter_observers/ParameterObserverLogger.cpp
karlnapf: @gf712 any ideas?

src/shogun/lib/parameter_observers/ParameterObserverLogger.cpp
```diff
@@ -0,0 +1,33 @@
...
+
+using namespace shogun;
+
+ParameterObserverLogger::ParameterObserverLogger() {}
+
+ParameterObserverLogger::ParameterObserverLogger(std::vector<std::string> &parameters)
```
karlnapf: we won't have these ctors available in the meta examples since it will be instantiated with a factory which calls the default ctor. So instead, you would use the `add` api (see the multiple kernel learning examples for how it works) to add strings to an internal vector of parameter names

src/shogun/lib/parameter_observers/ParameterObserverLogger.cpp
karlnapf: we do! see for example #4537 where @gf712 changed it to have nice strings for enum/options

karlnapf: For now, this printing is fine, but generally, we won't to have an abstraction layer for where the generated string is going (we can do that later)

src/shogun/lib/parameter_observers/ParameterObserverLogger.h
```diff
@@ -0,0 +1,40 @@
...
+
+#ifndef SHOGUN_PARAMETEROBSERVERLOGGER_H
+#define SHOGUN_PARAMETEROBSERVERLOGGER_H
+
+#include <shogun/base/SGObject.h>
+#include <shogun/lib/parameter_observers/ParameterObserverInterface.h>
+
+namespace shogun {
+	/**
+	 * This class implements a logger which will basically print on screen all
```
karlnapf: "This class implements a logger which prints all observed updates" .. much simpler

src/shogun/regression/LeastAngleRegression.cpp
karlnapf: Keep in mind registering parameters happens at *runtime*....so in fact you could only register the parameter in the observer once you receive it? Kinda crazy but might work :) Might require some changes. Let me know if you see what I mean

src/shogun/lib/parameter_observers/ParameterObserverLogger.cpp
gf712: Hmmm ok, there are two ways to solve this I think. Either move the definition of both those function into a translation unit (keep declaration where it is though), or mark them as `inline` in `type_case.h` (but inline is not enforceable though, so could break in some compilers/compiler versions). I can do that in a separate PR? It would be just good to check if that is the issue.

gf712: Actually `inline` is fine here because the linker will always see it as inline (and it won't be an ODR violation), whether or not the compiler decides to inline it in the machine code.

src/shogun/lib/parameter_observers/ParameterObserverLogger.cpp
geektoni: Cool! For now, I will leave it just a simple debug observer which will be also useful to check that we are getting the correct values out of the machines.

src/shogun/regression/LeastAngleRegression.cpp
geektoni: I was thinking more about registering inside the machine directly the parameter we are going to observe. For LARS it could be something like:
```c++
void CLeastAngleRegression::init()
{
	m_lasso = true;
	m_max_nonz = 0;
	m_max_l1_norm = 0;
	m_epsilon = CMath::MACHINE_EPSILON;
	SG_ADD(&m_epsilon, "epsilon", "Epsilon for early stopping", ParameterProperties::HYPER);
	SG_ADD(&m_max_nonz, "max_nonz", "Max number of non-zero variables", ParameterProperties::HYPER);
	SG_ADD(&m_max_l1_norm, "max_l1_norm", "Max l1-norm of estimator", ParameterProperties::HYPER);
	SG_ADD(&m_lasso, "lasso", "Max l1-norm of estimator", ParameterProperties::HYPER);

        SG_ADD(&beta, "betas", "Weights of the model while training", ParameterProperties::OBSERVABLE);
}
```
The `betas` parameter will not be modifiable by `put()`\`get()` and it can only be captured using an observer. Therefore, the user may be able to see which kind of information can be extracted from the model at runtime using an observer. 

Obviously, I do not know if SG_ADD is the right method, but I would like to signal somehow the presence of these emitted values.

See what I mean?

src/shogun/lib/parameter_observers/ParameterObserverLogger.cpp
geektoni: Okay then, I may try to fix it here in this PR.

examples/undocumented/libshogun/regression_lars_observable_path.cpp
geektoni: Yes, it could be done. Actually, this example was taken directly from one of the unit tests, so it was not using any data files.

examples/undocumented/libshogun/regression_lars_observable_path.cpp
geektoni: In principle, everything you see there is portable to a meta example. I'll try to see if I can convert it (maybe after we polish this one).

examples/undocumented/libshogun/regression_lars_observable_path.cpp
geektoni: mmh cool. I'll add factories in another PR.

examples/undocumented/libshogun/regression_lars_observable_path.cpp
geektoni: mmh it can be done. I have no strong feelings about it.

src/shogun/lib/parameter_observers/ObservedValue.h
geektoni: SG_OBS_VALUE_TYPE will be removed. I will update this when the other PR goes in.

src/shogun/lib/parameter_observers/ParameterObserverLogger.cpp
geektoni: Issue solved.

geektoni:
Closed in favour of #4618.

Issue #4488, opened 2019-01-25
Use iterative machine mixin in more algorithms


shubham808:
This is a very good issue for those who want to be involved in the project [inside the black box II](https://github.com/shogun-toolbox/shogun/wiki/GSoC_2019_inside_the_black_box-II)

In order to make algorithms stoppable, we need to execute each iteration and then save the state in member variables. If a user decides to stop the process at any iteration, control is returned back. This means the user is free to continue, serialize, test etc. on the pre-trained model
See [Perceptron](https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/classifier/Perceptron.cpp#L35) and [NewtonSVM](https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/classifier/svm/NewtonSVM.cpp#L50) as examples.
We want all iterative algorithms to apply [this](https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/machine/IterativeMachine.h) mixin
#### First Steps
* Identify the initialization and iteration phase(what occurs in the while loop) of the algorithm
* The initialization part (mostly above the while loop) goes into the ```init_model()``` method
* The iteration function will implement a single iteration of the algorithm, so we will move the contents of the while loop into ```iteration``` function
* Create new member variables if required to share data between ```init_model``` and ```iteration```
* Delete local copies of variables that represent state(like weights and bias)
* Finally write tests to see if everything is consistent :) see [NewtonSVM unit tests](https://github.com/shogun-toolbox/shogun/blob/develop/tests/unit/classifier/NewtonSVM_unittest.cc) and [Perceptron unit tests](https://github.com/shogun-toolbox/shogun/blob/develop/tests/unit/classifier/Perceptron_unittest.cc#L92)

Initially we want this done for a single Linear Machine algorithm ([CAveragedPerceptron](https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/classifier/AveragedPerceptron.cpp) is a good place to start)
See the [iterative machine guide](https://shubham808.github.io/blog/gsoc'18/IterativeMachine) for more information

souvik3333:
Hi @shubham808, I am new here. I read the wiki for inside the black box II project and I am really interested to work on it. This issue would be a great introduction to the project. So, Can I work on this?

shubham808:
Yes go ahead :)

han0305:
I am new to open source and I will like to work on this issue.

samdbrice:
[...]

PR #4551, opened 2019-03-02
Use iterative machine mixin in AveragedPerceptron

theartful:
@shubham808 This PR addresses the project 'inside the black box II' and issue #4488.
I still need to write some unit tests, but I would like to get confirmation that this is the right way, since I'm still new to this project.


theartful:
clone_equals_empty test in SGObjectAll_unittest fails for CAveragedPerceptron after the commit, I'm trying to figure out why. Any help is appreciated.

src/shogun/classifier/AveragedPerceptron.cpp
```diff
@@ -14,105 +15,99 @@
...
+			const auto gradient = learn_rate * true_label;
+			bias += gradient;
+			v.add(gradient, w);
 		}
-		iter++;
-		pb.print_progress();
+		// Add current w to tmp_w, and current bias to tmp_bias
+		// To calculate the sum of each iteration's w, bias
+		for (int32_t j = 0; j < num_feat; j++)
+			tmp_w[j] += w[j];
```
karlnapf: You can use a linalg::add variation here

src/shogun/classifier/AveragedPerceptron.cpp
```diff
@@ -14,105 +15,99 @@
...
-	float64_t* output=SG_MALLOC(float64_t, num_vec);
-
+	tmp_w = SGVector<float64_t>(num_feat);
+	
 	//start with uniform w, bias=0, tmp_bias=0
-	bias=0;
-	float64_t tmp_bias=0;
-	for (int32_t i=0; i<num_feat; i++)
-		w[i]=1.0/num_feat;
+	bias = 0, tmp_bias = 0;
```
karlnapf: Pls separate line

src/shogun/classifier/AveragedPerceptron.cpp
```diff
@@ -14,105 +15,99 @@
...
+	float64_t learning_rate
+) : CIterativeMachine<CLinearMachine>()
 {
+	if (trainlab) set_labels(trainlab);
+	if (traindat) init_model((CFeatures*) traindat);
+	
+	set_max_iterations(max_iters);
+	set_learn_rate(learning_rate);
+	
+	SG_ADD(&learn_rate, "learn_rate", "Learning rate.", ParameterProperties::HYPER);
```
karlnapf: This needs to happen in every ctor. Most importantly the default one. So best to move those to a method

src/shogun/classifier/AveragedPerceptron.cpp
```diff
@@ -14,105 +15,99 @@
...
-	bool converged=false;
-	int32_t iter=0;
-	SGVector<int32_t> train_labels = binary_labels(m_labels)->get_int_labels();
-	int32_t num_feat=features->get_dim_feature_space();
-	int32_t num_vec=features->get_num_vectors();
 
-	ASSERT(num_vec==train_labels.vlen)
+		SG_REF(data);
+		SG_UNREF(m_continue_features);
+		m_continue_features = data->as<CDotFeatures>();
```
karlnapf: @shubham808 is this necessary in subclasses?

src/shogun/classifier/AveragedPerceptron.cpp
```diff
@@ -14,105 +15,99 @@
...
-		w[i]=tmp_w[i]/(num_vec*iter);
-	bias=tmp_bias/(num_vec*iter);
-
-	SG_FREE(output);
-	SG_FREE(tmp_w);
-
-	set_w(w);
-
-	return converged;
+	for (int32_t i = 0; i < num_feat; i++)
```
karlnapf: Linalg here as well pls

src/shogun/classifier/AveragedPerceptron.h
```diff
@@ -35,12 +36,20 @@ class CAveragedPerceptron : public CLinearMachine
 		/** default constructor */
 		CAveragedPerceptron();
 
+		CAveragedPerceptron(CDotFeatures* traindat, CLabels* trainlab);
```
karlnapf: No need to add those ctors really as it all will be done via default ctor and then put calls

src/shogun/classifier/AveragedPerceptron.cpp
shubham808: yeah we should do this in each subclass

src/shogun/classifier/AveragedPerceptron.cpp
```diff
@@ -35,84 +26,74 @@ CAveragedPerceptron::~CAveragedPerceptron()
 
 void CAveragedPerceptron::init()
 {
-	max_iter = 1000;
-	learn_rate = 0.1;
-	SG_ADD(&max_iter, "max_iter", "Maximum number of iterations.", ParameterProperties::HYPER);
+	set_max_iter(1000);
```
shubham808: yes this is better

shubham808:
good first pr :) 
a unit test is also necessary since we need to make sure what we did actually works, see consistency tests [here](https://github.com/shogun-toolbox/shogun/blob/develop/tests/unit/classifier/NewtonSVM_unittest.cc#L23)

src/shogun/classifier/AveragedPerceptron.h
```diff
@@ -52,37 +47,33 @@ class CAveragedPerceptron : public CLinearMachine
 		/// set learn rate of gradient descent training algorithm
 		inline void set_learn_rate(float64_t r)
 		{
-			learn_rate=r;
+			learn_rate = r;
 		}
 
 		/// set maximum number of iterations
 		inline void set_max_iter(int32_t i)
```
shubham808: i dont think we need this setter here @karlnapf  ?

src/shogun/classifier/AveragedPerceptron.cpp
theartful: I did the required changes, and brought back the assertions. Averaged perceptron is pretty much the same as normal perceptron (except for averaging), so I imitated what was done in CPerceptron class. I think it can benefit from the changes you mentioned (using linalg and putting assertions), I can do them after this request.

src/shogun/classifier/AveragedPerceptron.h
theartful: I did the required changes, and brought back the assertions. Averaged perceptron is pretty much the same as normal perceptron (except for averaging), so I imitated what was done in CPerceptron class. I think it can benefit from the changes you mentioned (using linalg and putting assertions), I can do them after this request.

theartful:
@shubham808 consistency_training_tests has that test that makes sure that two models: one with max_iterations = 1 and the other stopped after an iteration, yield the same results. This isn't true here, since the one with max_iterations=1 has different weights because end_training() method is called after training is finished, which changes the weights to the final form (performs averaging).
For this to work, end_trainig() has to be called when training is stopped for whatever reason. But this will require us to copy intermediate weights, so that training can be resumed with the intermediate weights instead of the averaged ones.

src/shogun/classifier/AveragedPerceptron.cpp
karlnapf: I don’t think it’s good as then we’d have the same code in every subclass. Plus if someone forgets to do this.... can’t this be something that is done automatically?

src/shogun/classifier/AveragedPerceptron.h
karlnapf: nope, but we also don't need to delete those (yet). But for sure we dont need to add new ones

tests/unit/classifier/AveragedPeceptron_unittest.cc
```diff
@@ -0,0 +1,86 @@
...
+#include <shogun/classifier/AveragedPerceptron.h>
+#include <shogun/evaluation/ContingencyTableEvaluation.h>
+#include <shogun/features/DenseFeatures.h>
+#include <shogun/labels/BinaryLabels.h>
+
+using namespace shogun;
+
+extern LinearTestEnvironment* linear_test_env;
+
+TEST(AveragedPerceptron, continue_training_consistency)
```
karlnapf: Re the rest. Good that you have added it. One thing I don't really like is that this is almost copy pasted from the another test, i.e. we have lots of redundant code. The other test was more a proof of concept how to test those things. I would much rather have a single test that covers all these classes, either templated (with a googletest type list that it is instantiate for), or via runtime iteration through the appropriate models (faster). Have a look at `SGObject` unittests for *all* sgobjects (there is an iterator), and also have a look at the serialized model unit tests.
I think we should sort this out in this PR here ...

karlnapf:
> @shubham808 consistency_training_tests has that test that makes sure that two models: one with max_iterations = 1 and the other stopped after an iteration, yield the same results. This isn't true here, since the one with max_iterations=1 has different weights because end_training() method is called after training is finished, which changes the weights to the final form (performs averaging).
> For this to work, end_trainig() has to be called when training is stopped for whatever reason. But this will require us to copy intermediate weights, so that training can be resumed with the intermediate weights instead of the averaged ones.
> Edit: My suggestion won't work since end_training() is also used to free used resources by other algorithms. So maybe something like pause_training()?
> Edit2: In this implementation of AveragedPerceptron, weights are summed across iterations, then averaged. Maybe this can happen online: mean_{n+1} = mean_{n} + (x - mean_{n}) / (n + 1)

We had this issue before. I dont remember how we exactly solved it. A running average seems to work here and seems like a very good idea. Make sure to pick a numerically stable one (might require some googling). Could you add a way to update a mean vector/scalar in `linalg` in a separate PR? Then you can rebase this PR here to use it. This would be better than just adding the explicit code in here and then it could be used in other algorithms (LARS comes to mind where it is done explicitly as well)

tests/unit/classifier/AveragedPeceptron_unittest.cc
karlnapf: Let's use the test as is for now, since it allows us to fix the error you mentioned, until we have solved the mean issues, and then once that works we can update it.

src/shogun/classifier/AveragedPerceptron.cpp
shubham808: the problem is we wont know what to cast data to as it is only known in subclass so doing this here makes more sense

karlnapf: But we had this discussion before no?: there is no reason why the type of continue features is DotFestures. It should be just Features. The reference should be stored automatically in the base class, and the subclasses can check/assert the correct type at runtime in init_model (but not store them here, just check) and then the iteration method can do a static cast as that’s safe then. This copy pasting of ref/unref code is not good, error prone.
See what I mean?

karlnapf: Maybe I’m confused about this. Need to check he code my computer ... will do soon. But pls also see if the above makes sense

karlnapf: Just checked. The continue features are of type Features. This means that all that needs to happen here is the runtime assertion (as is fine, though there also is the enum, which is faster) that the features have the correct type. The code for storing the reference should be in IterativeMachine. Might need some refactoring for the perceptron as well ...

karlnapf: No need for the as cast btw here as the property is already asserted above

tests/unit/classifier/AveragedPeceptron_unittest.cc
```diff
@@ -0,0 +1,86 @@
...
+
+    // reference for completly trained model
+	auto results = avrg_perceptron->apply(test_features);
+
+	index_t iter = 0;
+
+	// prematurely stopped at 1 iterations
+	auto avrg_perceptron_stop = some<CAveragedPerceptron>();
+
+	std::function<bool()> callback = [&iter]() {
```
gf712: with lambdas in general you should use `auto` instead of a function pointer with `std::function`. Here it doesn't matter too much as it is a test but in the library code use `auto` for lambdas when you can.

src/shogun/classifier/AveragedPerceptron.h
```diff
@@ -52,37 +47,33 @@ class CAveragedPerceptron : public CLinearMachine
...
+		virtual void end_training();
 
 	protected:
 		/** learning rate */
 		float64_t learn_rate;
-		/** maximum number of iterations */
-		int32_t max_iter;
+
+	private:
+		float64_t tmp_bias;
```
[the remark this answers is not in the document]
theartful: Current name is `cached_bias` and `cached_w`, which is still bad I think. I will think of something better.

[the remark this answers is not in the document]
karlnapf: what about just saying what it is? w_averaged etc?

src/shogun/classifier/AveragedPerceptron.cpp
```diff
@@ -35,84 +23,74 @@ CAveragedPerceptron::~CAveragedPerceptron()
...
-
-	//start with uniform w, bias=0, tmp_bias=0
-	bias=0;
-	float64_t tmp_bias=0;
-	for (int32_t i=0; i<num_feat; i++)
-		w[i]=1.0/num_feat;
+	cached_w = SGVector<float64_t>(num_feat);
+	// start with uniform w, bias=0, tmp_bias=0
+	bias = 0;
+	cached_bias = 0;
```
karlnapf: actually this name is fine imo

src/shogun/classifier/AveragedPerceptron.cpp
```diff
@@ -35,84 +23,74 @@ CAveragedPerceptron::~CAveragedPerceptron()
...
-				tmp_w[j]+=w[j];
-			tmp_bias+=bias;
+			converged = false;
+			const auto gradient = learn_rate * true_label;
+			cached_bias += gradient;
+			feature.add(gradient, cached_w);
 		}
-		iter++;
-		pb.print_progress();
+		linalg::update_mean(w, cached_w, num_prev_weights);
```
karlnapf: sweet, much nicer to read!

src/shogun/classifier/Perceptron.cpp
```diff
@@ -39,10 +39,7 @@ void CPerceptron::init_model(CFeatures* data)
 		if (!data->has_property(FP_DOT))
 			SG_ERROR("Specified features are not of type CDotFeatures\n")
 		set_features((CDotFeatures*) data);
-
-		SG_REF(data);
-		SG_UNREF(m_continue_features);
-		m_continue_features = data->as<CDotFeatures>();
+		set_continue_features(data);
```
karlnapf: could we make this happen in the IterativeMachine class without the dev having to specify this in every use of the mixin?

tests/unit/utils/SGObjectIterator.h
```diff
@@ -0,0 +1,164 @@
...
+		return ptype;                                                          \
+	}                                                                          \
+                                                                               \
+	template <>                                                                \
+	std::string sg_primitive_type_string<T>()                                  \
+	{                                                                          \
+		return std::string(XSTRING__(ptype));                                  \
+	}
+
+SG_PRIMITIVE_TYPE(bool, PT_BOOL);
```
karlnapf: @gf712 do we have a central map for these?

tests/unit/utils/SGObjectIterator.h
```diff
@@ -0,0 +1,164 @@
...
+
+/** Helper to write c++11 style loops in tests that cover all Shogun classes */
+template <typename T>
+class sg_object_iterator
+{
+public:
+	sg_object_iterator(
+	    std::set<std::string> class_names = std::set<std::string>())
+	{
+		if (class_names.empty())
```
karlnapf: Id prefer offering two constructors here, one for each case, but that is a minor detail

tests/unit/utils/SGObjectIterator.h
```diff
@@ -0,0 +1,164 @@
+#ifndef __SG_OBJECT_ITERATOR_H__
```
karlnapf: good idea moving this into a new file!

tests/unit/machine/IterativeMachine_unittest.cc
```diff
@@ -0,0 +1,82 @@
...
+#include <shogun/base/some.h>
+#include <shogun/classifier/AveragedPerceptron.h>
+#include <shogun/classifier/Perceptron.h>
+#include <shogun/classifier/svm/NewtonSVM.h>
+#include <shogun/machine/IterativeMachine.h>
+#include <shogun/machine/LinearMachine.h>
+
+using namespace shogun;
+
+std::set<std::string> sg_linear_machines = {"Perceptron", "AveragedPerceptron",
```
karlnapf: ok! This is definitely better than before. I have a question: If a test fails, is it easy to see which class is the offender?

Filtering the tests obviously is not possible via this though. Compile time is an advantage to the typed version of this we had before (we define a bunch of types and then essentially template the test). This then allows for filtering. Maybe in a follow up PR, we can do something about that. I think the current blocker was that it is hard to find all subclasses automatically. However, if we simply define a type list (as you have done here with strings), that is also acceptable for now.

@gf712 might also be interesting for you

tests/unit/machine/IterativeMachine_unittest.cc
```diff
@@ -0,0 +1,82 @@
...
+			}
+			iter++;
+			return false;
+		};
+		machine_stop->set_callback(callback);
+		machine_stop->set_labels(labels);
+		machine_stop->train(features);
+
+		machine_iters->set_labels(labels);
+		machine_iters->put<int32_t>("max_iterations", max_iters);
```
karlnapf: there should be yet a typed setter for this or?

karlnapf: since this helps to make the tests for reliable (no runtime error when setting things as here), we can keep the setter (even though it will be hidden from users in the future)

src/shogun/classifier/AveragedPerceptron.h
shubham808: we can make this new setter typed

shubham808:
@theartful  Nice clean up for the tests, CI error seems unrelated 
about the setting of continue features maybe we could try doing that [here](https://github.com/shogun-toolbox/shogun/blob/fcd4799c5b90d68a191471803dd8a40329821d51/src/shogun/machine/IterativeMachine.h#L98) ?

tests/unit/machine/IterativeMachine_unittest.cc
theartful: It should be easy using `SCOPED_TRACE`. 
Hard coding the linear machines may not be so scaleable. Can't the `class_list` script be extended to detect hierarchies?
I don't get your point about filtering, and why templated tests can be better.

theartful:
Unrelated, but for some reason the LibLinearRegression test on MacOS passed :"D

tests/unit/machine/IterativeMachine_unittest.cc
karlnapf: Cool add the scoping!
Yes we can do this extension, but for now the explicit list is fine.
Compiled types can be helpful but for now this is good!

karlnapf:
Cool ci passed so this is almost good to go. I’ll do one more detailed read through later and then we can merge

shubham808:
@theartful lets check for memory leaks of the new test and then this is good to go from my side :+1:

tests/unit/utils/SGObjectIterator.h
gf712: all the supported types should go here
https://github.com/shogun-toolbox/shogun/blob/801bdbaee76c95ff77a243a8717f0ecefbf66dc0/src/shogun/lib/sg_types.h#L41
and all the any supported types are here
https://github.com/shogun-toolbox/shogun/blob/801bdbaee76c95ff77a243a8717f0ecefbf66dc0/src/shogun/lib/type_case.h#L118

theartful: @gf712 `DataType.h` seems to be obsolete, but I can't use `sg_types` and `type_case` yet since the `class_list` script (which has the `create` function) uses the old code.

theartful:
@shubham808 no memory leaks
```
$ valgrind ./shogun-unit-test --gtest_filter=IterativeMachine
==16774== Memcheck, a memory error detector
==16774== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==16774== Using Valgrind-3.13.0 and LibVEX; rerun with -h for copyright info
==16774== Command: ./shogun-unit-test --gtest_filter=IterativeMachine
==16774== 
Note: Google Test filter = IterativeMachine
[==========] Running 0 tests from 0 test cases.
[==========] 0 tests from 0 test cases ran. (18 ms total)
[  PASSED  ] 0 tests.
==16774== 
==16774== HEAP SUMMARY:
==16774==     in use at exit: 244 bytes in 4 blocks
==16774==   total heap usage: 56,722 allocs, 56,718 frees, 4,652,814 bytes allocated
==16774== 
==16774== LEAK SUMMARY:
==16774==    definitely lost: 0 bytes in 0 blocks
==16774==    indirectly lost: 0 bytes in 0 blocks
==16774==      possibly lost: 0 bytes in 0 blocks
==16774==    still reachable: 244 bytes in 4 blocks
==16774==         suppressed: 0 bytes in 0 blocks
==16774== Rerun with --leak-check=full to see details of leaked memory
==16774== 
==16774== For counts of detected and suppressed errors, rerun with: -v
==16774== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 0 from 0)
```

tests/unit/utils/SGObjectIterator.h
```diff
@@ -0,0 +1,164 @@
...
+#define __SG_OBJECT_ITERATOR_H__
+
+#include <iterator>
+#include <set>
+#include <string>
+#include <shogun/base/class_list.h>
+
+using namespace shogun;
```
vigsterkr: @theartful plz never do this in a header file...

karlnapf: my bad didn't see it

theartful: ouch! sorry for that

theartful: @vigsterkr https://github.com/shogun-toolbox/shogun/blob/79fa98a85aed7eb88be30884349dd0666567ac5a/tests/unit/utils/SGObjectIterator.h#L9
shouldn't it be namespace shogun?

PR #4552, opened 2019-03-02
Refactor ObservedValue.

geektoni:
Clean up the `ObservedValue` class from useless enums since the Any conversion will be done by the observers with dynamic dispatching.


src/shogun/base/SGObject.cpp
```diff
@@ -829,10 +806,7 @@ void CSGObject::list_observable_parameters()
 	SG_PRINT("------");
 	for (auto const& x : param_obs_list->get_list())
 	{
-		SG_PRINT(
-		    "%s [%s]: %s\n", x.first.c_str(),
-		    param_obs_list->type_name(x.second.first).c_str(),
-		    x.second.second.c_str());
```
karlnapf: could we remove this method and instead make one that returns `std::vector<std::string>` like we recently did for parameter names?

src/shogun/lib/parameter_observers/ParameterObserverCV.cpp
```diff
@@ -74,9 +71,10 @@ void CParameterObserverCV::on_next(const shogun::TimedObservedValue& value)
 	}
 	else
 	{
-		SG_SERROR(
-		    "ParameterObserverCV: The observed value received is not of "
-		    "type CrossValidationStorage\n");
+		SG_SPRINT(
```
karlnapf: why print and not error? also why not `REQUIRE`?

src/shogun/base/SGObject.cpp
geektoni: It is fine with me. It will be more consistent with the general Shogun's behaviour.

src/shogun/lib/parameter_observers/ParameterObserverCV.cpp
geektoni: It is because we do not want to make the application fail if the received value is not of the requested type. Let's imagine that I have several parameter observers attached to the same machine, let's call them `x` and `y`. They may receive the same emitted values from the machine, let's say A and B (with a different type). However, `x` may care only about A and `y` may care only about B. If `x` receives B, I do not want to make it crash the entire application. It just have to ignore that emitted value an process the next one.
See what I mean?

I can elaborate further in IRC if needed.

src/shogun/base/SGObject.cpp
```diff
@@ -817,23 +795,19 @@ class CSGObject::ParameterObserverList
...
-    const std::string& name, const SG_OBS_VALUE_TYPE type,
-    const std::string& description)
+    const std::string& name, const std::string& description)
 {
-	param_obs_list->register_param(name, type, description);
+	param_obs_list->register_param(name, description);
 }
 
-void CSGObject::list_observable_parameters()
+std::vector<std::string> CSGObject::list_observable_parameters()
```
geektoni: This guy here returns directly a vector of strings (with the name of the observable parameters), but I am not sure if this what you intended @karlnapf.

src/shogun/lib/parameter_observers/ObservedValue.h
```diff
@@ -69,10 +64,12 @@ namespace shogun
...
 		 */
-		ObservedValue(
-		    int64_t step, std::string& name, Any value, SG_OBS_VALUE_TYPE type)
-		    : m_step(step), m_name(name), m_value(value), m_type(type)
+		ObservedValue(int64_t step, std::string& name, Any value)
+		    : m_step(step), m_name(name), m_value(value)
 		{
+			SG_ADD(&m_step, "step", "Observation step", ParameterProperties::NONE);
+			SG_ADD(&m_name, "name", "Observed value name", ParameterProperties::NONE);
+			SG_ADD(&m_value, "value", "Observed value", ParameterProperties::NONE);
```
geektoni: This should enable `get` for the value observed. However, I fear that this is not the solution either since we would just be able to "get" again an `Any` value.

src/shogun/base/SGObject.cpp
karlnapf: sorry. I was distracted by the name. What about observable_names ... which is inline with recent changes for normal parameters. Minor though

src/shogun/lib/parameter_observers/ObservedValue.cpp
```diff
@@ -0,0 +1,13 @@
+#include <shogun/lib/parameter_observers/ObservedValue.h>
+
+using namespace shogun;
+
+ObservedValue::ObservedValue(int64_t step, std::string& name, Any value)
+		    : m_step(step), m_name(name), m_value(value)
+{
+	SG_ADD(&m_step, "step", "Observation step", ParameterProperties::NONE);
+	SG_ADD(&m_name, "name", "Observed value name", ParameterProperties::NONE);
+	SG_ADD(&m_value, "value", "Observed value", ParameterProperties::NONE);
```
karlnapf: for that to work, you need to template the class (best is subclass as discussed) and then register the typed parameter.

src/shogun/lib/parameter_observers/ObservedValue.cpp
```diff
@@ -0,0 +1,15 @@
...
+#include <shogun/base/Parameter.h>
+#include <shogun/lib/parameter_observers/ObservedValue.h>
+
+using namespace shogun;
+
+ObservedValue::ObservedValue(int64_t step, char * name)
+	: CSGObject(), m_step(step), m_name(name)
+{
+	SG_ADD(&step, "step", "Step", ParameterProperties::NONE);
+	SG_ADD(name, "name", "Name of the observed value", ParameterProperties::NONE);
```
geektoni: Here `name` is actually a `char *`. From what I could see, `std::string` is not accepted by the parameter framework (?).

src/shogun/io/TBOutputFormat.cpp
```diff
@@ -99,19 +99,19 @@ tensorflow::Event TBOutputFormat::convert_vector(
 	tensorflow::Event e;
 	std::time_t now_t = convert_to_millis(value.second);
 	e.set_wall_time(now_t);
-	e.set_step(value.first.get_step());
+	e.set_step(value.first->get<int64_t>("step"));
 
 	tensorflow::Summary* summary = e.mutable_summary();
 	auto summaryValue = summary->add_value();
-	summaryValue->set_tag(value.first.get_name());
+	summaryValue->set_tag(value.first->get<std::string>("name"));
```
geektoni: I was not able to get any `std::string` (or even `char*`) using `get`, therefore this line (and some others) causes an error. However, it should be pretty reasonable to do that or?

tests/unit/io/TBOutputFormat_unittest.cc
```diff
@@ -154,20 +155,20 @@ TEST(TBOutputFormatTest, fail_convert_scalar)
 	test_case_scalar_error<complex128_t>(1);
 };
 
-TYPED_TEST(TBOutputFormatTest, convert_all_types_histo)
+TYPED_TEST(TBOutputFormatTest, DISABLED_convert_all_types_histo)
```
geektoni: These tests are disabled because the parameter framework does not support the `std::vector` type.

src/shogun/lib/parameter_observers/ObservedValue.cpp
karlnapf: Check the definition of the macro
But I think we can add something to use string here as well

tests/unit/io/TBOutputFormat_unittest.cc
karlnapf: In what sense? 
Actually there is a pr by Sergey that Viktor said he wants to merge so that should allow vectors

src/shogun/io/TBOutputFormat.cpp
```diff
@@ -46,21 +46,21 @@ using namespace shogun;
 
 #define CHECK_TYPE(type)                                                       \
 	else if (                                                                  \
-	    value.first.get_value().type_info().hash_code() ==                     \
```
karlnapf: Are this macros still needed with the any dispatch?

src/shogun/lib/parameter_observers/ObservedValue.h
```diff
@@ -37,146 +37,81 @@
...
+
 		/** ObservedValue step (used by Tensorboard to print graphs) */
 		int64_t m_step;
 		/** Parameter's name */
-		std::string m_name;
-		/** Parameter's value */
-		Any m_value;
-		/** ObservedValue type */
-		SG_OBS_VALUE_TYPE m_type;
+		char * m_name;
```
karlnapf: Std string ?

src/shogun/lib/parameter_observers/ParameterObserverCV.cpp
```diff
@@ -57,27 +56,14 @@ CParameterObserverCV::~CParameterObserverCV()
...
-	{
-		CrossValidationStorage* recalled_value =
-		    any_cast<CrossValidationStorage*>(value.first.get_value());
-		SG_REF(recalled_value);
-
-		/* Print information on screen if enabled*/
-		if (m_verbose)
-			print_observed_value(recalled_value);
+	/* Print information on screen if enabled*/
+	if (m_verbose)
```
karlnapf: I think printing on screen is generally a bad thing to do. Rather print to a stream or shogun io or something....

src/shogun/lib/parameter_observers/ObservedValueTemplated.h
```diff
@@ -0,0 +1,40 @@
...
+
+		/**
+		 * Constructor
+		 * @param step step
+		 * @param name the observed value's name
+		 * @param value the observed value
+		 */
+		ObservedValueTemplated(int64_t step, char * name, T value)
+				: ObservedValue(step, name) {
+			SG_ADD(&v, "value", "Value of the observation", ParameterProperties::NONE);
```
karlnapf: Magic! :)

src/shogun/lib/parameter_observers/ObservedValueTemplated.h
```diff
@@ -0,0 +1,40 @@
...
+		}
+
+		~ObservedValueTemplated() {};
+
+		Any get_any() {
+			return make_any(v);
+		}
+
+	private:
+		T v;
```
karlnapf: Pls use a more descriptive name

src/shogun/lib/parameter_observers/ObservedValueTemplated.h
```diff
@@ -0,0 +1,40 @@
...
+#ifndef SHOGUN_OBSERVEDVALUETEMPLATED_H
+#define SHOGUN_OBSERVEDVALUETEMPLATED_H
+
+#include <shogun/base/Parameter.h>
+#include <shogun/lib/parameter_observers/ObservedValue.h>
+
+namespace shogun {
+
+	template<class T>
+	class ObservedValueTemplated : public ObservedValue {
```
karlnapf: Docs for devs are needed here

src/shogun/lib/parameter_observers/ObservedValueTemplated.h
```diff
@@ -0,0 +1,40 @@
...
+		 */
+		ObservedValueTemplated(int64_t step, char * name, T value)
+				: ObservedValue(step, name) {
+			SG_ADD(&v, "value", "Value of the observation", ParameterProperties::NONE);
+		}
+
+		~ObservedValueTemplated() {};
+
+		Any get_any() {
+			return make_any(v);
```
karlnapf: What about storing the any and returning a reference? Not sure about overhead ...

src/shogun/io/TBOutputFormat.cpp
geektoni: The issue is caused by this definition of `put` here (https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/base/SGObject.h#L391). Basically, when we try to extract a `string`, the `put` method will check only the options added with `SG_ADD_OPTION` and it will not consider parameter which are of type `string`.

src/shogun/lib/parameter_observers/ObservedValue.cpp
geektoni: The class `Parameter` does not have any method to store an `std::string` object. I guess we could add it. Should it be straightforward, right?

tests/unit/io/TBOutputFormat_unittest.cc
geektoni: Same here, the `Parameter` class does not have any `add` method to add `std::vector` objects. Moreover, `Any` still not support `std::vector` (however, this should be fixed by Viktor's PR).

src/shogun/io/TBOutputFormat.cpp
geektoni: No, they are not needed. Will remove them.

src/shogun/lib/parameter_observers/ObservedValue.h
geektoni: See issues above.

src/shogun/lib/parameter_observers/ObservedValueTemplated.h
geektoni: :dancer: :dancer: :dancer:

src/shogun/lib/parameter_observers/ObservedValueTemplated.h
geektoni: This is more of a convenience method to make some things works. I'm trying to see if I can remove it somehow :/

src/shogun/io/TBOutputFormat.cpp
karlnapf: Yes there was a discussion about this in the pr that added it. Solution: check the options first and if they don’t work just call the standard out method.
@gf712

src/shogun/lib/parameter_observers/ObservedValue.cpp
karlnapf: Let’s not touch the old parameter code. Instead, don’t use sgadd here but rather register with the watch method used inside sgadd. This means that the parameter will not be serialised for now but that’s fine for these observer classes imo.
You *can* register strings with the new framework

tests/unit/io/TBOutputFormat_unittest.cc
karlnapf: Not quite (the pr disables testing so I’m not sure what will happen with it) . But what you could do is to cherry pick Sergey’s commits that add std vector support in your pr here.

karlnapf: Oh and then also here: don’t use sgadd but directly register the std vector

src/shogun/lib/parameter_observers/ObservedValue.h
karlnapf: Solution: don’t use sgadd but register with new framework directly

src/shogun/lib/parameter_observers/ObservedValue.cpp
```diff
@@ -7,9 +7,10 @@
 
 using namespace shogun;
 
-ObservedValue::ObservedValue(int64_t step, char * name)
+ObservedValue::ObservedValue(int64_t step, std::string name)
 	: CSGObject(), m_step(step), m_name(name)
 {
 	SG_ADD(&step, "step", "Step", ParameterProperties::NONE);
-	SG_ADD(name, "name", "Name of the observed value", ParameterProperties::NONE);
+	this->watch_param("name", &name,
```
karlnapf: Cool! Does it work?

geektoni: It seems so. Still checking though...

src/shogun/base/SGObject.cpp
```diff
@@ -794,7 +794,7 @@ class CSGObject::ParameterObserverList
 	ObsParamsList m_list_obs_params;
 };
 
-void CSGObject::register_observable_param(
+void CSGObject::register_observable(
```
karlnapf: +1

src/shogun/lib/parameter_observers/ObservedValue.cpp
geektoni: It works indeed :D

src/shogun/lib/parameter_observers/ParameterObserverCV.cpp
geektoni: SG_PRINT is fine? It should be okay since it uses the `SGIO` object which could be defined by the user.

src/shogun/lib/parameter_observers/ObservedValueTemplated.h
geektoni: I guess this method will need to stay here for a bit longer. The only place where is needed is here https://github.com/shogun-toolbox/shogun/pull/4552/files#diff-d4ce27ca628f4ba5cacbb66e265af3d0L114, but it is not completely trivial to remove it.

geektoni:
@karlnapf Apart from some minor stuff, I think this guy here is 90% complete.

src/shogun/base/SGObject.cpp
```diff
@@ -816,24 +794,20 @@ class CSGObject::ParameterObserverList
...
-	SG_INFO("List of observable parameters of object %s\n", get_name());
-	SG_PRINT("------");
+	std::vector<std::string> list;
 	for (auto const& x : param_obs_list->get_list())
 	{
-		SG_PRINT(
-		    "%s [%s]: %s\n", x.first.c_str(),
-		    param_obs_list->type_name(x.second.first).c_str(),
-		    x.second.second.c_str());
+		list.push_back(x.first);
```
karlnapf: I think this loop could be replaced with a one liner from the std lib

src/shogun/evaluation/CrossValidation.cpp
```diff
@@ -108,10 +110,10 @@ CEvaluationResult* CCrossValidation::evaluate_impl()
 		SG_DEBUG("result of cross-validation run %d is %f\n", i, results[i])
 
 		/* Emit the value*/
-		std::string obs_value_name{"cross_validation_run"};
-		ObservedValue cv_data{i, obs_value_name, make_any(storage),
-		                      CROSSVALIDATION};
-		observe(cv_data);
+		observe(
+		    ObservedValue::make_observation<CrossValidationStorage*>(
```
karlnapf: What about renaming make_observation to observe (shorter, neater)

src/shogun/lib/parameter_observers/ObservedValue.cpp
```diff
@@ -0,0 +1,21 @@
...
+using namespace shogun;
+
+ObservedValue::ObservedValue(int64_t step, std::string name)
+    : CSGObject(), m_step(step), m_name(name)
+{
+	SG_ADD(&m_step, "step", "Step", ParameterProperties::NONE);
+	this->watch_param(
+	    "name", &m_name,
+	    AnyParameterProperties(
+	        "Name of the observed value", ParameterProperties::NONE));
```
karlnapf: Isn’t NONE the default? If not shouldn’t it? @gf712

src/shogun/lib/parameter_observers/ObservedValue.h
```diff
@@ -1,182 +1,92 @@
...
-		 */
-		void set_type(const SG_OBS_VALUE_TYPE type)
+		* Return a any version of the stored type.
+		* @return the any value.
+		*/
+		virtual Any get_any()
 		{
-			m_type = type;
+			SG_NOTIMPLEMENTED
+			return make_any(nullptr);
```
karlnapf: Did you check the overhead of this

@lisitsyn thought?

src/shogun/lib/parameter_observers/ObservedValueTemplated.h
```diff
@@ -0,0 +1,67 @@
...
+
+#include <shogun/base/Parameter.h>
+#include <shogun/lib/parameter_observers/ObservedValue.h>
+
+namespace shogun
+{
+
+	/**
+	 * Template ObservedValue which is used to store the real value we
+	 * want to send to the parameter observers. Therefore, we will be able
```
karlnapf: Make this shorter: “Templated specialisation of ObservedValue that stores the actual data.”

src/shogun/lib/parameter_observers/ParameterObserverCV.cpp
```diff
@@ -123,14 +110,14 @@ void CParameterObserverCV::print_machine_information(CMachine* machine) const
 	{
 		CLinearMachine* linear_machine = (CLinearMachine*)machine;
 		linear_machine->get_w().display_vector("Learned Weights = ");
-		SG_SPRINT("Learned Bias = %f\n", linear_machine->get_bias())
+		SG_PRINT("Learned Bias = %f\n", linear_machine->get_bias())
```
karlnapf: This is not good. The class won’t be available here anymore in the near future. Need to do either visitor pattern or virtual methods but no explicit casting

src/shogun/lib/parameter_observers/ParameterObserverCV.cpp
```diff
@@ -177,7 +164,7 @@ CrossValidationStorage* CParameterObserverCV::get_observation(int run) const
 	return obs;
 }
 
-const int CParameterObserverCV::get_num_observations() const
+const size_t CParameterObserverCV::get_num_observations() const
```
karlnapf: Please use the shogun int32_t to avoid sign warnings (need to explicitly cast and take care or overflows)

src/shogun/evaluation/CrossValidation.cpp
geektoni: I would prefer to leave it as it is, since it may cause confusion if we changed it to `observe`. Namely, `observe` emit a value, `make_observation` is a utility that creates an observable object.

src/shogun/lib/parameter_observers/ParameterObserverCV.cpp
geektoni: What do you mean here by `won't be available`?

src/shogun/lib/parameter_observers/ObservedValue.cpp
geektoni: Yes, it is indeed the default value.

src/shogun/evaluation/CrossValidation.cpp
karlnapf: Good point :)

src/shogun/lib/parameter_observers/ParameterObserverCV.cpp
karlnapf: Class definition bit within scope. But actually that is not true, sorry.
However I generally preferd if you used the to string method of the respective class, or a new method that prints this info, better style than having all the classes listed in here

src/shogun/lib/parameter_observers/ParameterObserverCV.cpp
```diff
@@ -164,7 +164,7 @@ CrossValidationStorage* CParameterObserverCV::get_observation(int run) const
 	return obs;
 }
 
-const size_t CParameterObserverCV::get_num_observations() const
+const int32_t CParameterObserverCV::get_num_observations() const
```
karlnapf: Need some checks here as sizet is unsigned

src/shogun/lib/parameter_observers/ParameterObserverCV.cpp
geektoni: I guess this will need a bit more of discussion. It may also be a good example of a good entrance task though :)

src/shogun/lib/parameter_observers/ParameterObserverCV.cpp
geektoni: What kind of checks? This method should just return the total number of observations, or am I missing something? :/

src/shogun/lib/parameter_observers/ParameterObserverCV.cpp
karlnapf: A simple solution would be to add `print_machine_information` to `CMachine`. But then again, we have `to_string` actually. Of course there could be a curated version, for machines that return only the relevant fields. Whatever we do, I think we certainly should not dispatch machine types in here.

src/shogun/lib/parameter_observers/ParameterObserverCV.cpp
karlnapf: `size_t` is unsigned so it can represent numbers that cannot be converted to `int32_t`. That is minor as there likely won't be that many observations. But the compiler will generate warnings about such implicit conversions

src/shogun/lib/parameter_observers/ParameterObserverCV.cpp
geektoni: I think that it would be better to do this kind of task in another PR, since it is somewhat related to refactoring the observers (which it is not related to the `ObservedValue`). Plus, I would like to merge this soon such to keep working on the other tasks.

src/shogun/lib/parameter_observers/ParameterObserverCV.cpp
geektoni: Okay, so, since `std::vector::size` returns a `size_t` value, I will keep also the return type of the function equal to `size_t`.

src/shogun/lib/parameter_observers/ParameterObserverCV.cpp
karlnapf: Ok sure, but let’s not forget about it

geektoni: Sure!

karlnapf:
It depends a bit: if you only ever use the size to iterate over indices used for a std vector it is fine. If the indices are used in any other shogun data structure, it won’t be
Why don’t we just introduce a conversion tool that can be used in other places as well.... the problem popped up before in the testing framework

src/shogun/util/utils.h
```diff
@@ -0,0 +1,38 @@
...
+	 *	int a = size_t_to_int32_cast(o);
+	 * } catch (std::exception e)
+	 * {
+	 *	std::cout << e.what() << std::endl;
+	 * }
+	 * @endcode
+	 * @param v original value
+	 * @return casted value
+	 */
+	int32_t size_t_to_int32_cast(size_t v) {
```
geektoni: @karlnapf were you thinking about something like this?

karlnapf: Force it inline

geektoni: Updated.

src/shogun/lib/parameter_observers/ParameterObserverCV.cpp
```diff
@@ -177,7 +165,12 @@ CrossValidationStorage* CParameterObserverCV::get_observation(int run) const
 	return obs;
 }
 
-const int CParameterObserverCV::get_num_observations() const
+const int32_t CParameterObserverCV::get_num_observations() const
 {
-	return m_observations.size();
+	try {
+		return size_t_to_int32_cast(m_observations.size());
```
karlnapf: Exactly in those lines! Would be good to print some debug info on the actual values maybe?

karlnapf:
Flying now. Will check in detail later today

src/shogun/lib/parameter_observers/ObservedValueTemplated.h
```diff
@@ -0,0 +1,64 @@
...
+		~ObservedValueTemplated(){};
+
+		/**
+		 * @copydoc ObservedValue::get_any()
+		 * This method returns an Any reference of the observed
+		 * value stored by this class.
+		 */
+		virtual Any get_any()
+		{
+			return make_any(m_observed_value);
```
karlnapf: Sorry the storing any comment should have been here

geektoni:
> I might have asked this before but forgot: why not store the any in the class? Isn’t there overhead?

No, there should not be too much overhead. I think that it will just save us from calling `make_any` when we need the untyped value. I do not remember either which usage we had in mind such to make us keep also the any :P

karlnapf:
Ah whoops just saw Viktors comment. I think you should cherry pick it or something. Better to have a well behaved central solution

vigsterkr:
just needs a rebase... pushed that little code to develop

karlnapf:
@vigsterkr it seems like deadbeef has everything we always wanted ;) thanks for pointing it out

src/shogun/lib/parameter_observers/ObservedValue.h
```diff
@@ -81,6 +82,8 @@ namespace shogun
 		int64_t m_step;
 		/** Parameter's name */
 		std::string m_name;
+        /** Untyped value */
+        Any m_any_value;
```
geektoni: @karlnapf This way we will store directly also the `Any` reference (I don't know why the formatting is messed up :/).

src/shogun/lib/parameter_observers/ObservedValue.h
```diff
@@ -1,182 +1,95 @@
...
-		SG_OBS_VALUE_TYPE get_type() const
+		* Helper method to generate an ObservedValue.
+		* @param step the step
+		* @param name the param's name we are observing
+		* @param value the param's value
+		* @return an ObservedValue object initialized
+		*/
+		template <class T>
+		static Some<ObservedValue>
+		make_observation(int64_t step, std::string name, T value)
```
karlnapf: this should also be hidden from swig or?

src/shogun/lib/parameter_observers/ObservedValue.h
```diff
@@ -1,182 +1,95 @@
...
-		 * @param type type of this observed value
-		 */
-		void set_type(const SG_OBS_VALUE_TYPE type)
+		* Return a any version of the stored type.
+		* @return the any value.
+		*/
+		virtual Any get_any()
 		{
-			m_type = type;
+			return m_any_value;
```
karlnapf: return value or reference?

src/shogun/lib/parameter_observers/ObservedValue.h
karlnapf: Store value or reference?

src/shogun/lib/parameter_observers/ParameterObserverCV.cpp
```diff
@@ -177,7 +165,13 @@ CrossValidationStorage* CParameterObserverCV::get_observation(int run) const
 	return obs;
 }
 
-const int CParameterObserverCV::get_num_observations() const
+const int32_t CParameterObserverCV::get_num_observations() const
 {
-	return m_observations.size();
+	try {
```
karlnapf: do you need to try catch that here? We could just make the exception string in viktors code more verbose (printing the provided value and the ranges or something? Not sure I get why the error msg is here ...?

karlnapf: This is in fact quite a bit neater. We should have thought of this in the first place :)

karlnapf: Ok good, but this now deals with the issue properly which is always better than adding more and more hacks

geektoni: sooo is that try block okay? LoL

src/shogun/lib/parameter_observers/ObservedValue.h
geektoni: I would say by value because it is more self-contained.

src/shogun/lib/parameter_observers/ObservedValue.h
geektoni: Yes, it should be.

src/shogun/lib/parameter_observers/ObservedValue.h
geektoni: I guess that return by reference would be okay.

src/shogun/lib/parameter_observers/ParameterObserverCV.cpp
karlnapf: Ah ok it is a warning, so just leave this, nevermind my comment :)

src/shogun/lib/parameter_observers/ObservedValueTemplated.h
```diff
@@ -0,0 +1,55 @@
...
+		 * @param name the observed value's name
+		 * @param value the observed value
+		 */
+		ObservedValueTemplated(int64_t step, std::string name, T value)
+		    : ObservedValue(step, name), m_observed_value(value)
+		{
+			this->watch_param(
+			    "value", &m_observed_value,
+			    AnyParameterProperties(
+			        "Value of the observation", ParameterProperties::NONE));
```
karlnapf: Actually, I think the ctor should also accept a description of the observation, rather than just putting it "Value of the observation". So when observations are made (especially for the custom observations like the CV one), users can actually get a one-liner what this is about? What are your thoughts on that?

karlnapf:
Alright, before we merge, the CI errors needs to be addressed ...

src/shogun/lib/parameter_observers/ObservedValueTemplated.h
geektoni: Ideally, the `ObservedValue` has already a name which should refer to its content and it should be enough. The `m_observed_value` instead is the actual observed value, so its description specifies only that if we do `get("value")` we will get the observed data.

geektoni:
Okay, it should be good now. Let's see if CI likes it.

src/shogun/lib/parameter_observers/ObservedValueTemplated.h
karlnapf: Not sure I agree....think a single name "weights"...what weights would those be? Better would be "weights" and "MKL kernel weights of the corresponding x-validation fold". No?

geektoni: mmh yeah, I guess it can make it more understandable. I'll update the constructor to take also a description.

geektoni: Updated.

geektoni:
@karlnapf Should be okay now? :)

src/shogun/lib/parameter_observers/ObservedValueTemplated.h
```diff
@@ -0,0 +1,55 @@
...
+		 * @param value the observed value
+		 */
+		ObservedValueTemplated(int64_t step, std::string name, std::string description, T value)
+		    : ObservedValue(step, name), m_observed_value(value)
+		{
+			this->watch_param(
+			    "value", &m_observed_value,
+			    AnyParameterProperties(
+			        description, ParameterProperties::NONE));
+            m_any_value = make_any(m_observed_value);
```
karlnapf: whitespace is weird here

geektoni: I'll do a round of code formatter

PR #4554, opened 2019-03-03
Add cumulative and exponential moving averages functions

theartful:
The functions uses the form $average_n = alpha * x + (1 - alpha) * average_{n-1}$, which performs additional multiplication compared to $average_n = average_{n-1} + alpha * (x - average_{n-1})$. But its rounding error should be less as alpha gets smaller.
Function names probably should change.


karlnapf:
Closed?

PR #4555, opened 2019-03-03
Add cumulative moving average function

theartful:
Moving average is implemented in the form average_n = (1-alpha)*average_{n-1} + alpha*x. This has an additional multiplication compared to average_n = average_{n-1} + alpha*(x-average_n), but it should has less rounding error when alpha is small (not sure).

### DIFF
--- a/src/shogun/mathematics/linalg/LinalgNamespace.h
+++ b/src/shogun/mathematics/linalg/LinalgNamespace.h
@@ -1507,6 +1507,36 @@ namespace shogun
 			return infer_backend(a)->mean(a);
 		}
 
+		/** Method that updates moving mean vector with new datum point.
+		 *
+		 * @param cma the previous moving mean
+		 * @param datum new datum point
+		 * @param n number of previous data points including the new datum point
+		 */
+		template <typename T>
+		void update_mean(SGVector<T>& cma, const SGVector<T>& datum, int32_t n)
+		{
+			REQUIRE(n > 0, "Number of data points (%d) must be at least 1", n);
+			T alpha = (T)(1.0) / n;
+			T beta = 1 - alpha;
+			add(datum, cma, cma, alpha, beta);
+		}
+
+		/** Method that updates moving mean scalar with new datum point.
+		 *
+		 * @param cma the previous moving mean
+		 * @param datum new datum point
+		 * @param n number of previous data points including the new datum point
+		 */
+		template <typename T>
+		void update_mean(T& cma, const T datum, int32_t n)
+		{
+			REQUIRE(n > 0, "Number of data points (%d) must be at least 1", n);
+			T alpha = (T)(1.0) / n;
+			T beta = 1 - alpha;
+			cma = alpha * datum + beta * cma;
+		}
+
 		/**
 		 * Method that computes the standard deviation of vectors or matrices
 		 * composed of real numbers.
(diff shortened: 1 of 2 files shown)


src/shogun/mathematics/Statistics.h
```diff
@@ -44,6 +45,83 @@ class CStatistics: public CSGObject
...
+		{
+			REQUIRE(
+				n > 0,
+				"Number of data points (%d) must be at least 1",
+				n
+			);
+			T alpha = (T)(1.0) / n;
+			T beta = 1 - alpha;
+			linalg::add<T>(datum, cma, result, alpha, beta);
+		}
```
gf712: You should do a round of [clang-format](https://clang.llvm.org/docs/ClangFormat.html) at some point to fix the indents!

src/shogun/mathematics/Statistics.h
```diff
@@ -44,6 +45,83 @@ class CStatistics: public CSGObject
...
+	 *
+	 * @param cma the previous moving average
+	 * @param datum new datum point
+	 * @param n number of data points including the new datum point
+	 * @param result the vector that saves the result 
+	 */
+	template<class T>
+		static void cumulative_moving_average(
+			const SGVector<T>& cma, const SGVector<T>& datum,
+			int32_t n, SGVector<T>& result)
```
gf712: Wouldn't it be better to return the result rather than modify the reference?

karlnapf: Moving averages imply that something is modified so I think is good to just modify the given vector wih the new datum. Not even sure we need the result parameter. Would just call the method “update_mean” and then have only 3 parameters: the mean, the new vector, and n

theartful:
It seems that C's "complex.h" in both clang and gcc used in CI had macros that interfered with the codebase and caused build failures. This showed when I made "Statistics.h" include the linalg namespace. Luckily C++'s "complex" doesn't have this problem. Lapack is the only library that uses "complex.h" but it has the option to use anything else
`/* Complex type (single precision) */`
`#ifndef lapack_complex_float`
`#include <complex.h>`
`#define lapack_complex_float    float _Complex`
`#endif`
`/* Complex type (double precision) */`
`#ifndef lapack_complex_double`
`#include <complex.h>`
`#define lapack_complex_double   double _Complex`
`#endif`
 
So 
`#include <complex>`
`#define lapack_complex_float std::complex<float>`
`#define lapack_complex_double std::complex<double>
`
before including "Eigen/Eigen" seems to fix the problem for clang and gcc. Still need to figure out why testing fails in macOS.

karlnapf:
Solution: move this to the linalg code. Sorry I didn’t realise earlier that you defined this in the statistics module. Check out how things like the mean are defined in there

theartful:
I think it would be better if the main problem is solved. In clang for example, <complex.h> defines a 'I' for complex numbers and google tests uses I as a template parameter name. Check this issue https://github.com/Reference-LAPACK/lapack/issues/320

karlnapf:
Yeah you are right of course. My term solution wasn’t appropriate ;)
I’ll check into this later (on mobile atm), maybe @gf718 also has an idea.

Nevertheless, the averaging functions should live in the linalg module

theartful:
Is the plan to get rid of Statistics?

PR #4557, opened 2019-03-04
Add factory for the ParameterObservers.

geektoni:
This PR adds a factory for the parameter observers and it contains also some refactoring of the interface and the code itself (to make it consistent with the rest of Shogun). 

* Rename ParameterObserverInterface to ParameterObserver for consistency;
* Make ParameterObserver inherith from SGObject;
* Convert usages in the meta examples;
* Split on_next into on_next and on_next_impl;
* Add more methods to ParameverObserver to get a single observation;
* Refactor ParameterObserverCV;
* Fix SWIG interfaces;
* Fix format;


src/shogun/lib/parameter_observers/ParameterObserver.h
```diff
@@ -103,17 +134,32 @@ namespace shogun
...
+
 		/**
 		 * List of parameter's names we want to monitor
 		 */
 		std::vector<std::string> m_parameters;
 
+		/**
+		 * Observations recorded each time we compute on_next()
+		 */
+		std::vector<Any> m_observations;
```
geektoni: All the observer will store the values they capture. For now, they will be kept as `Any` values since they may be really different from each other (a scalar, a vector, etc.). Does it seem clean enough?

examples/meta/src/evaluation/cross_validation_multiple_kernel_learning_weights_storage.sg
```diff
@@ -32,7 +32,7 @@ MachineEvaluation cross = machine_evaluation("CrossValidation", machine=svm, fea
 #![create_cross_validation]
 
 #![create_observer]
-ParameterObserverCV mkl_obs(True)
+ParameterObserver mkl_obs = parameter_observer("ParameterObserverCV")
```
geektoni: This works perfectly. However, some problems arises when trying to get an observation out of the observer:
```c++
CrossValidationStorage obs = mkl_obs.get_observation(0)
CrossValidationFoldStorage fold = obs.get_fold(0)
```
The call to ```mkl_obs.get_observation(0)``` extract one `Any` value. Is there any way to cast it somehow to its proper type by using the meta language? :/ @karlnapf any ideas?

karlnapf: The way to do that is via ‘get’ which is available for all registered parameters

src/shogun/lib/parameter_observers/ParameterObserver.h
karlnapf: Not sure. We know that all observed values have the same type or? So could template this? Not sure...

examples/meta/src/evaluation/cross_validation_multiple_kernel_learning_weights_storage.sg
karlnapf: so The objects that are exposed to the api need to have parameters registered with the appropriate type. Seems like we are pushing the problem around, need to think a bit about this

src/interfaces/swig/shogun.i
```diff
@@ -295,6 +295,7 @@ PUT_ADD(CEvaluation)
 PUT_ADD(CSVM)
 PUT_ADD(CMeanFunction)
 PUT_ADD(CLikelihoodModel)
+PUT_ADD(ParameterObserver)
```
karlnapf: Do we ever need to put a parameter observer?

src/shogun/lib/parameter_observers/ParameterObserver.cpp
```diff
@@ -67,4 +66,12 @@ bool ParameterObserverInterface::filter(const std::string& param)
 		if (v == param)
 			return true;
 	return false;
+}
+
+Any ParameterObserver::get_observation(int i)
+{
+	REQUIRE(
+	    i < m_observations.size(), "The id of the observation is to great.");
```
karlnapf: Typo. Also could you use a more standard error msg?

src/shogun/lib/parameter_observers/ParameterObserver.h
```diff
@@ -87,12 +88,42 @@ namespace shogun
 		 */
 		virtual bool filter(const std::string& param);
 
+		/**
+		 * Get the i-th observation recorded. It will be presented
+		 * to the user as an Any value which need to be casted.
+		 * @param i id of the observation
+		 * @return Any value with the observation
+		 */
+		virtual Any get_observation(int i);
```
karlnapf: Users should never receive an any, it is internal. Rather, they can access parameters with get

src/shogun/lib/parameter_observers/ParameterObserver.h
geektoni: Okay, so I've come up with another idea which I think it's the most viable option (and also it does not expose `Any` to the final users). Basically, the observers will have an `std::vector<ObservedValue>` parameter which will contain all the observation captured so far (we will not use the `Any`). If I want to access it from the interface I can do:
```c++
ParameterObserver pm = parameter_observer("ParameterObserverCV")
ObservedValue obs = pm.get_observations(1)
```
`ObservedValue` will contain both the observed `Any` value and the name of it as registered parameters. Therefore, we could extract the value itself from the `ObservedValue` like:
```c++
obs.get_real("value") // if we know that the value is real
obs.get_real("value") // if it is a string...
```
The only possible issue is that inside `ObservedValue` we would need to register (with `SG_ADD`) an `Any` value, since we cannot know a-priori which will be the type of the observed value.

examples/meta/src/evaluation/cross_validation_multiple_kernel_learning_weights_storage.sg
geektoni: See above discussion.

src/interfaces/swig/shogun.i
geektoni: Not sure. If we use the `subscribe` method of `SGObject` it may not be needed.

src/shogun/lib/parameter_observers/ParameterObserver.h
geektoni: See above discussion.

src/shogun/lib/parameter_observers/ParameterObserver.h
geektoni: In order to expose the observations (aka `std::vector<ObservedValue>`) we can:
  * Use the current method (`get_observations(int i)`) which won't need `std::vector<ObservedValue>` to be registered with `SG_ADD`;
  * Use `get` and add a new SWIG getter like `%template(get_observation) CSGObject::get<std::vector<ObservedValue>, void>;`. We would need also to add a new type called `ObservedVector` which will correspond to `std::vector<ObservedValue>`.

karlnapf: I think in our discussion, we discussed the first option you described. I think it might work!

src/interfaces/swig/ParameterObserver.i
```diff
@@ -14,6 +14,7 @@
 
 namespace shogun {
 
+%template(SomeObservedValue) Some<ObservedValue>;
 %template(get_CV_storage) CSGObject::get<CrossValidationStorage*, void>;
```
geektoni: Since I need to expose Some<ObservedValue> to SWIG, I've tried to add it to the build. However, I keep getting these warnings:
```
/root/shogun/src/shogun/base/some.h:71: Warning 508: Declaration of 'get' shadows declaration accessible via operator->(),              
/root/shogun/src/shogun/base/SGObject.h:487: Warning 508: previous declaration of 'get'.                                                        
/root/shogun/src/shogun/base/some.h:71: Warning 508: Declaration of 'get' shadows declaration accessible via operator->(),                      
/root/shogun/src/shogun/base/SGObject.h:496: Warning 508: previous declaration of 'get'.                                                        
/root/shogun/src/shogun/base/some.h:24: Warning 509: Overloaded method shogun::Some< shogun::ObservedValue >::Some(shogun::Some< shogun::Observe
dValue > &&) effectively ignored,                                                                                                               
/root/shogun/src/shogun/base/some.h:20: Warning 509: as it is shadowed by shogun::Some< shogun::ObservedValue >::Some(shogun::Some< shogun::Obse
rvedValue > const &).                                                                                                                           
/root/shogun/src/shogun/base/some.h:46: Warning 503: Can't wrap 'operator shogun::ObservedValue*' unless renamed to a valid identifier. 
```
Moreover, the entire compilations fails with this obscure and mystical errors from SWIG:
```
/root/shogun/build/src/interfaces/python/shogunPYTHON_wrap.cxx: In function 'PyObject* _wrap_SomeObservedValue_put__SWIG_2(PyObject*, PyObject*)
':                                                                                                                                      
/root/shogun/build/src/interfaces/python/shogunPYTHON_wrap.cxx:837537:0: error: 'shogun_ObservedValue_put_scalar_dispatcher_Sl_int32_t_Sc_int32_
t_Sg___SWIG_2' was not declared in this scope                                                                                                   
         shogun_ObservedValue_put_scalar_dispatcher_Sl_int32_t_Sc_int32_t_Sg___SWIG_2((shogun::ObservedValue*)(arg1)->operator ->(),(std::string
 const &)*arg2,arg3);  
```
Any clue about what is going on? @karlnapf

karlnapf: Don’t expose some to swig yet. We can do that soon but it’s not ready. You will need to use the old skool approach for now

karlnapf: So just observed value raw pointer in the api

examples/meta/generator/targets/cpp.json
```diff
@@ -102,7 +102,8 @@
             "get_int_string_list": "$object->get<SGStringList<int32_t>>($arguments)",
             "get_char_string_list": "$object->get<SGStringList<char>>($arguments)",
             "get_word_string_list": "$object->get<SGStringList<uint16_t>>($arguments)",
-            "get_option": "$object->get<std::string>($arguments)"
+            "get_option": "$object->get<std::string>($arguments)",
+            "get_CV_storage": "$object->get<CrossValidationStorage*>($arguments)"
```
karlnapf: This type should be hidden from swig / interfaces or?

examples/meta/src/evaluation/cross_validation_multiple_kernel_learning_weights_storage.sg
```diff
@@ -46,8 +46,9 @@ real stddev = result.get_real("std_dev")
 #![get_results]
 
 #![get_fold_machine]
-CrossValidationStorage obs = mkl_obs.get_observation(0)
-CrossValidationFoldStorage fold = obs.get_fold(0)
+ObservedValue obs_v = mkl_obs.get_observation(0)
+CrossValidationStorage stor = obs_v.get_CV_storage("value")
+CrossValidationFoldStorage fold = stor.get_fold(0)
```
karlnapf: IMO, we souldnt add this many new types to the API. Instead can't we use `put/get` to retreive this info?

`mkl_obs.get_observation(0)..get("value").get("fold", 0)`
Even better if the `get("value") wouldnt be there but the values would be directly accessible from the value
`mkl_obs.get_observation(0).get("fold", 0)`

@lisitsyn we need a `get` for array elements, similar to the `add` we have. Maybe we can just do this as above, with the additional `int` parameter in get?

src/interfaces/swig/shogun.i
karlnapf: ok then remove this here

src/interfaces/swig/ParameterObserver.i
```diff
@@ -1,19 +1,19 @@
...
-%include <shogun/lib/parameter_observers/ParameterObserverTensorBoard.h>
-%include <shogun/lib/parameter_observers/ParameterObserverScalar.h>
-%include <shogun/lib/parameter_observers/ParameterObserverHistogram.h>
-#endif // HAVE_TFLOGGER
+%include <shogun/lib/parameter_observers/ObservedValue.h>
+%include <shogun/lib/parameter_observers/ParameterObserver.h>
+
+namespace shogun {
+
+%template(get_CV_storage) CSGObject::get<CrossValidationStorage*, void>;
```
karlnapf: as said above, I think we should try to hide this type

examples/meta/generator/targets/cpp.json
geektoni: It is needed if we want to make the user able to extract it from the parameter map with `get`.

karlnapf: But why does the specialised class need be exposed and not the base class? Get is sgobject

karlnapf: My point is: the CV storage type is not needed in the interfaces, just a getter for the base type and then the user can use ‘get’

geektoni: That's because `CrossValidationStorage` has some methods to store/get objects (e.g. `get_fold()`) which are not present in SGObject. I guess that I could remove all those methods and use directly the tag framework. We could then remove `get_CV_storage`.

examples/meta/src/evaluation/cross_validation_multiple_kernel_learning_weights_storage.sg
geektoni: `get("fold", 0)` would be super useful indeed.

examples/meta/generator/targets/cpp.json
karlnapf: Yes we actually need to do this. Tag framework is the way to go. Reason is all these things will be plugins at some point and we also want to make swig api stable

karlnapf: Yes we actually need to do this. Tag framework is the way to go. Reason is all these things will be plugins at some point and we also want to make swig api stable

examples/meta/src/evaluation/cross_validation_multiple_kernel_learning_weights_storage.sg
karlnapf: Yep let’s add it. I’ll be travelling tomorrow and will have a look.

karlnapf:
Note convert is checking for overflows as well (some inline function)

src/shogun/base/SGObject.cpp
```diff
@@ -771,12 +772,33 @@ void CSGObject::subscribe_to_parameters(ParameterObserverInterface* obs)
...
+	// Insert the subscription in the list
+	m_subscriptions.insert(
+			std::make_pair<int64_t, rxcpp::subscription>(
+					std::move(m_next_subscription_index),
+					std::move(subscription)));
+
+	return m_next_subscription_index++;
+}
+
+void CSGObject::unsubscribe(int64_t subscription_index) {
```
karlnapf: `subscription_index` -> `index`

src/shogun/base/SGObject.cpp
```diff
@@ -771,12 +772,33 @@ void CSGObject::subscribe_to_parameters(ParameterObserverInterface* obs)
...
+
+	// Check if we have such subscription
+	auto it = m_subscriptions.find(subscription_index);
+	if (it == m_subscriptions.end())
+		SG_ERROR("The object %s does not have any registered parameter observer with index %i",
+		  this->get_name(), subscription_index);
+
+	// Unsubscribe from this SGObject and delete it from the map
```
karlnapf: minor: superflous comment, code is self explaining.

src/shogun/base/SGObject.cpp
```diff
@@ -1076,6 +1098,8 @@ CSGObject* CSGObject::get(const std::string& name, std::nothrow_t) const
 		return result;
 	if (auto* result = get_sgobject_type_dispatcher<CLabels>(name))
 		return result;
+	if (auto* result = get_sgobject_type_dispatcher<CMachine>(name))
```
karlnapf: there will be a merge error when I merge #4581 (soon). Just to warn you :)

src/shogun/base/some.h
```diff
@@ -241,6 +241,7 @@ namespace shogun
 		return Some<T>::from_raw(nullptr);
 	}
 
+#ifndef SWIG
```
karlnapf: ++

src/shogun/evaluation/CrossValidation.cpp
```diff
@@ -172,8 +168,8 @@ float64_t CCrossValidation::evaluate_one_run(
 			/* evtl. update xvalidation output class */
 			CrossValidationFoldStorage* fold = new CrossValidationFoldStorage();
 			SG_REF(fold)
-			fold->set_run_index(index);
-			fold->set_fold_index(i);
+			fold->put("current_run_index", (index_t)index);
```
karlnapf: maybe remove "current_" prefix?
Sorry, I am all in favour of removing redundancy from names these days :)

src/shogun/lib/observers/ParameterObserver.h
```diff
@@ -82,12 +83,77 @@ namespace shogun
...
 		virtual bool filter(const std::string& param);
 
+		/**
+		 * Get a vector with all the observations matching a certain name
+		 * and a certain name (not SG_REF).
+		 * @tparam T the type of the observations
+		 * @param name the name of the observations
+		 * @return vector with the observations casted to the requested type
+		 */
+		std::vector<ObservedValue*> get_observations(std::string name)
```
karlnapf: who would call this? Swig?

src/shogun/lib/observers/ParameterObserver.h
```diff
@@ -82,12 +83,77 @@ namespace shogun
...
+
+		/**
+		 * Return a single observation from the received ones (not SG_REF).
+		 * @tparam T the type of the observation
+		 * @param i the index
+		 * @return the observation casted to the requested type
+		 */
+		ObservedValue * get_observation(size_t i)
+		{
+			REQUIRE(i>=0 && i<this->get_num_observations(), "Observation index (%i) is out of bound.", i);
```
karlnapf: print the (upper) bound

src/shogun/lib/observers/ParameterObserver.h
```diff
@@ -82,12 +83,77 @@ namespace shogun
...
+			return this->m_observations[i].get();
+		};
+
+		/**
+		 * Get the total number of observation received.
+		 * @return number of obsevation received.
+		 */
+		const size_t get_num_observations() const
```
karlnapf: could you make all public interfaces use `index_t` rather than `size_t` (To avoid problems with mixing those types)

src/shogun/lib/observers/ParameterObserver.h
karlnapf: Because it will cause problems ... ;)

geektoni: Yeah, this was a draft of a possible design for getting the observations. I thought that SWIG would have caused problems ;)

examples/meta/src/evaluation/cross_validation_multiple_kernel_learning_weights_storage.sg
karlnapf: Get for arrays is now availavle

src/shogun/lib/observers/ParameterObserver.h
karlnapf: swig should use the new `get(name, index)`.
For that `ObservedValue` will need to added to the get dispatcher (you already added `CMachine`)

examples/meta/generator/targets/cpp.json
karlnapf: updates on this?

examples/undocumented/python/evaluation_cross_validation_multiclass_storage.py
```diff
@@ -67,15 +67,15 @@ def evaluation_cross_validation_multiclass_storage (traindat=traindat, label_tra
...
     cross_validation.subscribe_to_parameters(multiclass_storage)
 
     # perform cross-validation
     result=cross_validation.evaluate()
 
     # get first observation and first fold
-    obs = multiclass_storage.get_observations()[0]
-    fold = obs.get_folds_results()[0]
+    obs = multiclass_storage.get_observation(0).get_CV_storage("value")
+    fold = obs.get_fold(0)
```
karlnapf: `get("fold")` ?

src/shogun/evaluation/CrossValidationStorage.h
```diff
@@ -56,106 +56,10 @@ namespace shogun
...
-		/**
-		 * Get ground truth (correct labels for this fold)
-		 * @return ground truth
-		 */
-		CLabels* get_test_true_result() const;
-
-		/**
-		 * Get the evaluation result of this fold
-		 * @return evaluation result
-		 */
```
karlnapf: all this removed code makes swig more lightweight which is good! :)

src/shogun/lib/observers/ParameterObserverLogger.cpp
```diff
@@ -0,0 +1,41 @@
...
+void ParameterObserverLogger::on_complete() {
+
+}
+
+void ParameterObserverLogger::on_next_impl(const TimedObservedValue &value) {
+
+	auto name = value.first->get<std::string>("name");
+	auto any_val = value.first->get_any();
+	auto print_func = [&](auto v){
+		SG_PRINT("[%l] Received a value called %s which contains: %s", convert_to_millis(value.second),
```
karlnapf: nice!
Maybe we can soon start thinking about a first blog post that showcases exactly this logger for say the Perceptron?

src/shogun/lib/observers/ParameterObserverLogger.h
```diff
@@ -0,0 +1,41 @@
...
+#define SHOGUN_PARAMETEROBSERVERLOGGER_H
+
+#include <shogun/lib/observers/ParameterObserver.h>
+
+namespace shogun {
+
+	/**
+	 * This class implements a logger which prints all observed updates.
+	 */
+	class ParameterObserverLogger : public ParameterObserver {
```
karlnapf: I would go for "ParameterLogger" but that is taste I guess. You will have noticed that I like short names these days. Really inspired by Sergey on that :)

examples/meta/generator/targets/cpp.json
geektoni: I am trying to see if I can use just the standard `get()` and remove this guy. Currently I am trying to add the `CrossValidationStorage` to this method [here](https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/base/SGObject.h#L1100). After that it should work.

src/shogun/lib/observers/ParameterObserverLogger.cpp
geektoni: Sure! The CI has some problems when compiling these lambdas, though, but I think I know how to make it work again.

src/shogun/lib/observers/ParameterObserverLogger.h
geektoni: mmh I would be in favour of keeping the "Observer" somewhere, so it becomes clear about what is this object :)

src/shogun/base/SGObject.h
```diff
@@ -1112,6 +1123,10 @@ CSGObject* get_dispatch_all_base_types(const CSGObject* obj, const std::string&
 		return result;
 	if (auto* result = get_if_possible<CFeatures>(obj, name, how))
 		return result;
+	if (auto* result = get_if_possible<CMachine>(obj, name, how))
+		return result;
+	if (auto* result = get_if_possible<CEvaluationResult>(obj, name, how))
```
geektoni: @karlnapf This brings us close to remove `get_CV_storage`.

examples/meta/src/evaluation/cross_validation_multiple_kernel_learning_weights_storage.sg
```diff
@@ -46,9 +46,10 @@ real stddev = result.get_real("std_dev")
 #![get_results]
 
 #![get_fold_machine]
-CrossValidationStorage obs = mkl_obs.get_observation(0)
-CrossValidationFoldStorage fold = obs.get_fold(0)
-Machine machine_mkl = as_machine(fold.get_trained_machine())
+ObservedValue obs_v = mkl_obs.get_observation(0)
+EvaluationResult cross_validation_run = obs_v.get("cross_validation_run")
+EvaluationResult fold = cross_validation_run.get("folds", 0)
```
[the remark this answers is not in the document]
karlnapf: that is nice!

examples/undocumented/libshogun/evaluation_cross_validation_mkl_weight_storage.cpp
```diff
@@ -54,11 +54,11 @@ SGMatrix<float64_t> calculate_weights(
 	for (int o = 0; o < obs.get_num_observations(); o++)
 	{
 		auto obs_storage = obs.get_observation(o);
-		for (int i = 0; i < obs_storage->get_num_folds(); i++)
+		for (int i = 0; i < obs_storage->get<index_t("num_folds"); i++)
```
karlnapf: minor: `auto` and `range`

examples/undocumented/python/evaluation_cross_validation_multiclass_storage.py
```diff
@@ -67,20 +67,20 @@ def evaluation_cross_validation_multiclass_storage (traindat=traindat, label_tra
...
-    obs = multiclass_storage.get_observations()[0]
-    fold = obs.get_folds_results()[0]
+    obs = multiclass_storage.get_observation(0).get("cross_validation_run")
+    fold = obs.get("folds", 0)
 
     # get fold ROC for first class
     eval_ROC = ROCEvaluation()
-    pred_lab_binary = MulticlassLabels.obtain_from_generic(fold.get_test_result()).get_binary_for_class(0)
-    true_lab_binary = MulticlassLabels.obtain_from_generic(fold.get_test_true_result()).get_binary_for_class(0)
+    pred_lab_binary = MulticlassLabels.obtain_from_generic(fold.get("test_result")).get_binary_for_class(0)
```
karlnapf: These labels are a bit confusing don't you think?
Maybe we can call them "training" "validation" and "ground_truth" ?

src/interfaces/swig/shogun.i
```diff
@@ -171,7 +171,7 @@ namespace shogun
 		else
 			$self->put(tag_t, value);
 	}
-
```
karlnapf: minor: pls dont commit a file if only whitespace changes

src/shogun/base/SGObject.cpp
```diff
@@ -540,6 +542,9 @@ void CSGObject::init()
 	m_subject_params = new SGSubject();
 	m_observable_params = new SGObservable(m_subject_params->get_observable());
 	m_subscriber_params = new SGSubscriber(m_subject_params->get_subscriber());
+	m_next_subscription_index = 0;
+
+	watch_method("total_subscriptions", &CSGObject::get_total_subscriptions);
```
karlnapf: minor: "num_subscriptions" is both shorter and more informative

src/shogun/base/SGObject.h
```diff
@@ -441,15 +441,15 @@ class CSGObject
 	/** Typed array getter for an object array class parameter of a Shogun base class
 	* type, identified by a name and an index.
 	*
-	* Raises an error if parameter does not exist.
+	* It does not throw any errors if it does not work.
```
karlnapf: minor: awkward english due to double negation. "Returns nullptr if parameter of desired type does not exist"

src/shogun/base/SGObject.h
```diff
@@ -441,15 +441,15 @@ class CSGObject
...
 	*
 	* @param name name of the parameter array
 	* @param index index of the element in the array
 	* @return desired element
 	*/
 	template <class T,
 			  class X = typename std::enable_if<is_sg_base<T>::value>::type>
-	T* get(const std::string& name, index_t index) const
+	T* get(const std::string& name, index_t index, std::nothrow_t) const
```
karlnapf: Was it needed? If so, where?

src/shogun/base/SGObject.h
```diff
@@ -951,6 +967,16 @@ class CSGObject
 	Unique<ParameterObserverList> param_obs_list;
 
 protected:
+
+	/**
+	 * Return total subscriptions
+	 * @return total number of subscriptions
+	 */
+	size_t get_total_subscriptions() const
```
karlnapf: index_t please!

src/shogun/base/SGObject.h
```diff
@@ -1095,12 +1125,8 @@ template <typename T>
...
 {
 	CSGObject* result = nullptr;
-	try
-	{
-		// there is no "has" that checks for array types, so check implicitly
-		result = obj->get<T>(name, how.m_index);
-	}
-	catch (const std::exception&) {}
+	// there is no "has" that checks for array types, so check implicitly
+	result = obj->get<T>(name, how.m_index, std::nothrow);
```
karlnapf: ah I see. Fine then!

src/shogun/base/SGObject.h
karlnapf: I saw below. OK!

src/shogun/base/SGObject.h
```diff
@@ -1112,6 +1138,12 @@ CSGObject* get_dispatch_all_base_types(const CSGObject* obj, const std::string&
 		return result;
 	if (auto* result = get_if_possible<CFeatures>(obj, name, how))
 		return result;
+	if (auto* result = get_if_possible<CMachine>(obj, name, how))
+		return result;
+	if (auto* result = get_if_possible<CLabels>(obj, name, how))
```
karlnapf: Rather than extending this list, we eventually want to have a single line call here that utilises a global list of base classes. But maybe we can do that another in another PR
@gf712 knows what to do :)

src/shogun/base/base_types.h
```diff
@@ -52,7 +53,8 @@ namespace shogun
 	                    std::is_same<CLikelihoodModel, T>::value ||
 	                    std::is_same<CMeanFunction, T>::value ||
 	                    std::is_same<CLossFunction, T>::value ||
-	                    std::is_same<CTokenizer, T>::value>
+	                    std::is_same<CTokenizer, T>::value ||
+						std::is_same<CEvaluationResult, T>::value>
```
karlnapf: whitespace

src/shogun/evaluation/CrossValidation.cpp
```diff
@@ -97,9 +98,9 @@ CEvaluationResult* CCrossValidation::evaluate_impl()
 		SG_DEBUG("Creating CrossValidationStorage.\n")
 		CrossValidationStorage* storage = new CrossValidationStorage();
 		SG_REF(storage)
-		storage->set_num_runs(m_num_runs);
-		storage->set_num_folds(m_splitting_strategy->get_num_subsets());
-		storage->set_expose_labels(m_labels);
+		storage->put("num_runs", utils::safe_convert<index_t>(m_num_runs));
+		storage->put("num_folds", utils::safe_convert<index_t>(m_splitting_strategy->get_num_subsets()));
+		storage->put("expose_labels", m_labels);
```
karlnapf: weird name. What does it mean? Maybe just "labels" ? :)

src/shogun/evaluation/CrossValidation.cpp
```diff
@@ -108,8 +109,8 @@ CEvaluationResult* CCrossValidation::evaluate_impl()
 		SG_DEBUG("result of cross-validation run %d is %f\n", i, results[i])
 
 		/* Emit the value */
-		observe(
-		    i, "cross_validation_run", "One run of CrossValidation", storage);
+		observe(i, "cross_validation_run", "One run of CrossValidation",
+				dynamic_cast<CEvaluationResult*>(storage));
```
karlnapf: `->as<CEvaluationResult>()`

src/shogun/evaluation/CrossValidation.cpp
```diff
@@ -168,8 +169,8 @@ float64_t CCrossValidation::evaluate_one_run(
 			/* evtl. update xvalidation output class */
 			CrossValidationFoldStorage* fold = new CrossValidationFoldStorage();
 			SG_REF(fold)
-			fold->set_run_index(index);
-			fold->set_fold_index(i);
+			fold->put("run_index", (index_t)index);
```
karlnapf: just "run"

src/shogun/evaluation/CrossValidation.cpp
```diff
@@ -168,8 +169,8 @@ float64_t CCrossValidation::evaluate_one_run(
...
+			fold->put("fold_index", i);
```
karlnapf: just "fold" (imo, you can object of course ;) )

src/shogun/evaluation/CrossValidation.cpp
```diff
@@ -183,9 +184,9 @@ float64_t CCrossValidation::evaluate_one_run(
 			    m_splitting_strategy->generate_subset_indices(i);
 
 			/* evtl. update xvalidation output class */
-			fold->set_train_indices(inverse_subset_indices);
+			fold->put("train_indices", inverse_subset_indices);
 			auto fold_machine = (CMachine*)m_machine->clone();
-			fold->set_trained_machine(fold_machine);
+			fold->put("trained_machine", fold_machine);
```
karlnapf: same here just "machine" Keep in mind that these fields will (hopefully) have meaningful descriptions that would in this case say "Machine trained on training set"

src/shogun/evaluation/CrossValidationStorage.cpp
```diff
@@ -187,23 +97,32 @@ operator==(const CrossValidationFoldStorage& rhs) const
...
+}
+
 /** CrossValidationStorage **/
 
-CrossValidationStorage::CrossValidationStorage() : CSGObject()
+CrossValidationStorage::CrossValidationStorage() : CEvaluationResult()
 {
 	m_num_runs = 0;
 	m_num_folds = 0;
 	m_expose_labels = NULL;
```
karlnapf: I see the name was my idea :see_no_evil: maybe just change it :)

src/shogun/evaluation/CrossValidationStorage.h
```diff
@@ -39,6 +39,7 @@
 #include <shogun/base/SGObject.h>
 #include <shogun/lib/SGVector.h>
 #include <vector>
+#include "EvaluationResult.h"
```
karlnapf: please use <> includes

src/shogun/evaluation/EvaluationResult.h
```diff
@@ -20,7 +20,9 @@ namespace shogun
 enum EEvaluationResultType
 {
 	CROSSVALIDATION_RESULT=0,
-	GRADIENTEVALUATION_RESULT=1
+	GRADIENTEVALUATION_RESULT=1,
+	CROSSVALIDATIONSTORAGE_RESULT=2,
+	CROSSVALIDATIONFOLDSTORAGE_RESULT=3
```
karlnapf: is this used anywhere? I thought we dont need to dispatch the type since we use "get"?

src/shogun/lib/observers/ParameterObserver.cpp
```diff
@@ -33,37 +33,39 @@
...
-#include <shogun/lib/observers/ParameterObserverInterface.h>
+#include <shogun/lib/observers/ParameterObserver.h>
+#include <shogun/base/Parameter.h>
 
 using namespace shogun;
 
-ParameterObserverInterface::ParameterObserverInterface() : m_parameters()
+ParameterObserver::ParameterObserver() : m_observed_parameters(), m_subscription_id(-1)
 {
+	SG_ADD(&m_subscription_id, "subscription_id", "Id of the subscription to an object.");
```
karlnapf: -> "id" :D

src/shogun/lib/observers/ParameterObserver.h
karlnapf: Is this even needed?

src/shogun/lib/observers/ParameterObserver.h
```diff
@@ -82,12 +84,77 @@ namespace shogun
...
+			return result;
+		}
+
+		/**
+		 * Return a single observation from the received ones (not SG_REF).
+		 * @tparam T the type of the observation
+		 * @param i the index
+		 * @return the observation casted to the requested type
+		 */
+		ObservedValue * get_observation(size_t i)
```
karlnapf: index_t

src/shogun/lib/observers/ParameterObserver.h
```diff
@@ -82,12 +84,77 @@ namespace shogun
...
+		{
+			REQUIRE(i>=0 && i<this->get_num_observations(), "Observation index (%i) is out of bound (total observations %i)", i, this->get_num_observations());
+			return this->m_observations[i].get();
+		};
+
+		/**
+		 * Get the total number of observation received.
+		 * @return number of obsevation received.
+		 */
+		const size_t get_num_observations() const
```
karlnapf: index_t ?

src/shogun/lib/observers/ParameterObserverLogger.cpp
karlnapf: cool lets do that once this is merged

src/shogun/lib/observers/ParameterObserverLogger.cpp
```diff
@@ -0,0 +1,52 @@
...
+	auto name = value.first->get<std::string>("name");
+	auto any_val = value.first->get_any();
+
+	auto pf_n = [&](auto v){
+		SG_PRINT("[%l] Received a value called %s which contains: %s", convert_to_millis(value.second),
+				 name.c_str(), std::to_string(v).c_str());
+	};
+
+	auto pf_sgvector = [&](auto v){
+		//SG_PRINT("[%l] Received a value called %s which contains: %s", convert_to_millis(value.second),
```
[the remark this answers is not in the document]
karlnapf: cool! Please dont forget though :)

examples/undocumented/python/evaluation_cross_validation_multiclass_storage.py
geektoni: I would prefer something like:
-  test_result --> predicted_labels;
-  test_true_result --> ground_truth_labels;

src/shogun/base/SGObject.h
geektoni: This change was done to throw correctly the exceptions when the `get(call, index)` does not work. The previous implementation was causing errors with the interfaces.

src/shogun/base/base_types.h
geektoni: This will be fixed by the code formatter.

examples/undocumented/python/evaluation_cross_validation_multiclass_storage.py
karlnapf: predicted, ground_truth? :)
In my opinion, the label should be short and details (such as that this is labels) should be in the type and the description.But if you dont like it go for it

src/shogun/base/SGObject.h
karlnapf: ah good catch then!

src/shogun/base/base_types.h
karlnapf: but you need to apply it or?

src/shogun/evaluation/CrossValidation.cpp
geektoni: I would prefer to keep the `_index` part because it gives more information about what the user will get. I mean, if I do like `get("fold")`, what will I get? A complex object, an `int64_t`? with index is more "obvious" imho

src/shogun/evaluation/CrossValidation.cpp
geektoni: same as above :) even if we have the descriptions

src/shogun/evaluation/CrossValidationStorage.cpp
geektoni: I do like it! :D I would keep it as it is

src/shogun/evaluation/CrossValidation.cpp
karlnapf: The way I see this is: Labels should be short to reduce typing. The user once needs to read the description (thats why we have it!) and then wants to type less. Better than alway forcing people to type imo. And also less redundant. It is a taste thing though I appreciate that. Fine if we disagree, your call and we can also change it later. I just give my opinion :)

src/shogun/evaluation/EvaluationResult.h
geektoni: We still have some `obtain_from_generic` which use those enums. Maybe they can be removed though

karlnapf: yeah just remove that

src/shogun/lib/observers/ParameterObserver.cpp
geektoni: same as above for the `trained_machine` stuff. I imagine the user asking himself "id of what?" when it is trying to get it.

src/shogun/base/base_types.h
geektoni: That will be the last thing. I do not want to pollute the diffs with whitespaces changes ;)

src/shogun/evaluation/CrossValidation.cpp
geektoni: I would prefer to change the naming at the end then :)

karlnapf: list is growing :) but yeah good point!

examples/meta/src/evaluation/cross_validation_multiple_kernel_learning_weights_storage.sg
```diff
@@ -46,9 +46,10 @@ real stddev = result.get_real("std_dev")
 #![get_results]
 
 #![get_fold_machine]
-CrossValidationStorage obs = mkl_obs.get_observation(0)
-CrossValidationFoldStorage fold = obs.get_fold(0)
-Machine machine_mkl = as_machine(fold.get_trained_machine())
+ObservedValue obs_v = mkl_obs.get_observation(0)
+EvaluationResult cross_validation_run = obs_v.get("cross_validation_run")
```
geektoni: I've seen that Java and C# tests fail because they have no way to implicitly convert `SGObject` to `EvaluationResult`. Should I add a new method `as_evaluation_result()`? @karlnapf @vigsterkr

geektoni: and btw, where are `as_kernel`/`as_machine` defined? :/ because I'm not able to find them inside shogun apparently...

geektoni: nevermind. found them

examples/undocumented/python/evaluation_cross_validation_mkl_weight_storage.py
```diff
@@ -18,13 +18,14 @@
...
     from shogun import machine_evaluation
-    from shogun import ParameterObserverCV
+    from shogun import parameter_observer
     from shogun import ContingencyTableEvaluation, ACCURACY
     from shogun import splitting_strategy
     from shogun import BinaryLabels
     from shogun import CombinedFeatures
     from shogun import CombinedKernel
     from shogun import MKLClassification
+    from shogun import as_kernel, as_machine
```
karlnapf: rather than importing all these names here, use `sg.as_kernel` etc

examples/undocumented/python/evaluation_cross_validation_mkl_weight_storage.py
```diff
@@ -71,11 +72,11 @@ def evaluation_cross_validation_mkl_weight_storage(traindat=traindat, label_trai
...
-        obs = mkl_storage.get_observation(obs_index)
-        for fold_index in range(obs.get_num_folds()):
-            fold = obs.get_fold(fold_index)
-            machine = MKLClassification.obtain_from_generic(fold.get_trained_machine())
-            w = machine.get_kernel().get_subkernel_weights()
+        obs = mkl_storage.get_observation(obs_index).get("cross_validation_run")
+        for fold_index in range(obs.get("num_folds")):
+            fold = obs.get("folds", fold_index)
+            machine = as_machine(fold.get("trained_machine"))
+            w = as_kernel(machine.get("kernel")).get_subkernel_weights()
```
karlnapf: `get_subkernel_weights` is not part of `CKernel` or?
Also, I don't think you need to do the cast here, just can do
`obs.get("folds", fold_index).get("trained_machine").get("subkernel_weights_or_whatever_the_name_is")`

src/interfaces/swig/shogun.i
karlnapf: better not to commit this file

src/shogun/base/SGObject.h
```diff
@@ -951,6 +967,16 @@ class CSGObject
 	Unique<ParameterObserverList> param_obs_list;
 
 protected:
+
+	/**
+	 * Return total subscriptions
+	 * @return total number of subscriptions
+	 */
+	index_t get_total_subscriptions() const
```
karlnapf: `get_num_subscriptions` -- to be consistent with the string registered in "watch_method"

src/shogun/base/SGObject.h
```diff
@@ -1095,12 +1125,8 @@ template <typename T>
 CSGObject* get_if_possible(const CSGObject* obj, const std::string& name, GetByNameIndex how)
 {
 	CSGObject* result = nullptr;
-	try
-	{
-		// there is no "has" that checks for array types, so check implicitly
-		result = obj->get<T>(name, how.m_index);
-	}
-	catch (const std::exception&) {}
+	// there is no "has" that checks for array types, so check implicitly
```
karlnapf: this comment can go if the check is not implicit anymore I guess :)

src/shogun/evaluation/CrossValidation.cpp
```diff
@@ -331,7 +333,8 @@ float64_t CCrossValidation::evaluate_one_run(
 			SG_UNREF(labels);
 			SG_UNREF(evaluation_criterion);
 			SG_UNREF(result_labels);
-			SG_UNREF(fold)
+			SG_UNREF(true_labels);
+			SG_UNREF(fold);
```
karlnapf: did you valgrind the updated code somehow to avoid memory leaks here?

src/shogun/evaluation/CrossValidationStorage.cpp
```diff
@@ -187,68 +97,41 @@ operator==(const CrossValidationFoldStorage& rhs) const
...
+
 /** CrossValidationStorage **/
 
-CrossValidationStorage::CrossValidationStorage() : CSGObject()
+CrossValidationStorage::CrossValidationStorage() : CEvaluationResult()
 {
 	m_num_runs = 0;
 	m_num_folds = 0;
-	m_expose_labels = NULL;
+	m_original_labels = NULL;
```
karlnapf: ++

src/shogun/evaluation/CrossValidationStorage.cpp
```diff
@@ -187,68 +97,41 @@ operator==(const CrossValidationFoldStorage& rhs) const
...
-	    &m_num_runs, "m_num_runs", "The total number of cross-validation runs",
+	    &m_num_runs, "num_runs", "The total number of cross-validation runs",
 	    ParameterProperties::HYPER);
 	SG_ADD(
-	    &m_num_folds, "m_num_folds",
+	    &m_num_folds, "num_folds",
 	    "The total number of cross-validation folds", ParameterProperties::HYPER);
 	SG_ADD(
-	    (CSGObject**)&m_expose_labels, "m_expose_labels",
+		(CLabels**)&m_original_labels, "labels",
```
karlnapf: is that cast necessary? the type should already be CLabels or?

src/shogun/evaluation/CrossValidationStorage.cpp
```diff
@@ -260,7 +143,7 @@ bool CrossValidationStorage::operator==(const CrossValidationStorage& rhs) const
 {
 	auto member_vars = m_num_runs == rhs.m_num_runs &&
 	                   m_num_folds == rhs.m_num_folds &&
-	                   m_expose_labels->equals(rhs.m_expose_labels);
+	                   m_original_labels->equals(rhs.m_original_labels);
```
karlnapf: why do we need this operator?

src/shogun/evaluation/CrossValidation.cpp
geektoni: Yes, I did. Actually, I had to place those lines to fix a memory leak.

src/shogun/evaluation/CrossValidationStorage.cpp
geektoni: I guess it was needed for some old unit test which is not there anymore. I'll check if I can remove it safely...

geektoni: They can be removed indeed.

src/shogun/evaluation/CrossValidation.cpp
```diff
@@ -97,9 +98,9 @@ CEvaluationResult* CCrossValidation::evaluate_impl()
 		SG_DEBUG("Creating CrossValidationStorage.\n")
 		CrossValidationStorage* storage = new CrossValidationStorage();
 		SG_REF(storage)
-		storage->set_num_runs(m_num_runs);
-		storage->set_num_folds(m_splitting_strategy->get_num_subsets());
-		storage->set_expose_labels(m_labels);
+		storage->put("num_runs", utils::safe_convert<index_t>(m_num_runs));
```
karlnapf: minor: inside C++ code, I am not the biggest fan of using put with strings if the actual typed C++ API is available (setters). But not need to change that now

PR #4558, opened 2019-03-05
Fix complex/lapack bug

theartful:
This PR addresses the issue with complex.h interfering with the codebase. The header is included by lapacke, which is included by eigen3.h.

### DIFF
--- a/src/shogun/mathematics/eigen3.h
+++ b/src/shogun/mathematics/eigen3.h
@@ -8,12 +8,18 @@
 #ifndef EIGEN3_H_
 #define EIGEN3_H_
 
+#ifdef __cplusplus
+#include <complex>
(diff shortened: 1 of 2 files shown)


vigsterkr:
@theartful can you please point us to the situation when the error you are trying to fix came up.... a test or dev env

gf712:
@vigsterkr I think this is relate to something discussed in https://github.com/shogun-toolbox/shogun/pull/4555

theartful:
@vigsterkr The error happens whenever "complex.h" is included and any parameter named "I" is defined (which is used as a template parameter name in google tests). I wrote a small test that defined that parameter after including lapacke, check the build log here https://dev.azure.com/shogunml/shogun/_build/results?buildId=783

PR #4559, opened 2019-03-06
Remove m_gradient_parameters WIP

avramidis:
WIP to remove m_gradient_parameters form SGObject. Because this parameter is of type Parameter it affects a big number of files. I use AnyParameter to pass the parameters to all the affected functions. 

There are a lot of issues that need to be fixed in this PR such as in type_case.h, makes things simpler, etc.
I added some functions in SGObject that such as get_parameter_by_name and get_parameters_by_property. These probably need to change name/location or get rid of them if there is a better solution.

Now I am at the point of changing the model selection classes. After all these are done with this PR we will be able to remove the Parameter class, the m_parameters and  m_model_selection_parameters variables.

### DIFF
--- a/src/shogun/base/SGObject.cpp
+++ b/src/shogun/base/SGObject.cpp
@@ -35,6 +35,7 @@
 #include <algorithm>
 #include <unordered_map>
 #include <memory>
+#include <vector>
 
 #include <shogun/machine/Machine.h>
 #include <shogun/distance/Distance.h>
@@ -46,6 +47,10 @@
 #include <shogun/multiclass/ecoc/ECOCDecoder.h>
 #include <shogun/multiclass/MulticlassStrategy.h>
 
+#include <shogun/machine/gp/LikelihoodModel.h>
+#include <shogun/machine/gp/MeanFunction.h>
+
+#include <shogun/lib/type_case.h>
 
 namespace shogun
 {
@@ -538,24 +543,39 @@ void CSGObject::print_modsel_params()
 {
 	SG_PRINT("parameters available for model selection for %s:\n", get_name())
 
-	index_t num_param=m_model_selection_parameters->get_num_parameters();
+	Any any;
+	auto found_it_lambda = [&any](auto type) {
+		std::cout << "Type " << demangled_type<decltype(type)>() << std::endl;
+	};
+			
+	SG_PRINT("Model selection parameters:\n");
+
+	index_t num_param = 0;
+	for (auto it = self->map.begin(); it != self->map.end(); ++it )
+	{
+		if (it->second.get_properties().get_model_selection() == 1)
+		{
+			num_param++;
+			
+			// Get the name of the parameter
+			std::cout << "\t " << it->first.name();
+			// Get the description of the parameter
+			std::cout << " (" << it->second.get_properties().get_description() << "): ";
+			// Get the type of the parameter
+			any = it->second.get_value();
+			try 
+			{
+				shogun::sg_any_dispatch(any, shogun::sg_all_typemap, found_it_lambda);
+			}
+			catch (ShogunException &msg) 
+			{
+				std::cout << msg.what() << std::endl;
+			}
+		}
+	}
 
 	if (!num_param)
 		SG_PRINT("\tnone\n")
-
-	for (index_t i=0; i<num_param; i++)
-	{
-		TParameter* current=m_model_selection_parameters->get_parameter(i);
-		index_t  l=200;
-		char* type=SG_MALLOC(char, l);
-		if (type)
-		{
-			current->m_datatype.to_string(type, l);
-			SG_PRINT("\t%s (%s): %s\n", current->m_name, current->m_description,
-					type);
-			SG_FREE(type);
-		}
-	}
 }
 
 SGStringList<char> CSGObject::get_modelsel_names()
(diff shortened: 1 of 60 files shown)


src/shogun/base/base_types.h
```diff
@@ -48,7 +49,8 @@ namespace shogun
 	                    std::is_same<CDifferentiableFunction, T>::value ||
 	                    std::is_same<CInference, T>::value ||
 	                    std::is_same<CLikelihoodModel, T>::value ||
-	                    std::is_same<CMeanFunction, T>::value>
+	                    std::is_same<CMeanFunction, T>::value ||
+	                    std::is_same<CKernelNormalizer, T>::value>
```
avramidis: @gf712 I was trying here to add an sg_type_map for Shogun objects. I am not sure if I am going to need it in the end.

src/shogun/evaluation/GradientResult.h
```diff
@@ -92,20 +92,20 @@ class CGradientResult : public CEvaluationResult
...
 
 			// get parameter name
-			const char* param_name=param_node->key->m_name;
+			const char* param_name=param_node->key->get_properties().get_name().c_str();
 
 			// get object name
-			const char* object_name=
-				m_parameter_dictionary->get_element(param_node->key)->get_name();
+//			const char* object_name=
+//				m_parameter_dictionary->get_element(param_node->key)->get_name();
```
avramidis: Here I need the name of the object the parameter belongs to. AnyParameter doesn't have this information so I will have to think of a way to make it work.

src/shogun/lib/type_case.h
```diff
@@ -115,6 +118,10 @@ namespace shogun
 	{                                                                          \
 	};
 
+        SG_ADD_PRIMITIVE_TYPE(CSGObject*, TYPE::T_SGOBJECT)
+        SG_ADD_PRIMITIVE_TYPE(CKernel*, TYPE::T_CKERNEL)
```
avramidis: @gf712 here too is my try to add objects. This works which allowed me to continue the change but obviously I will have to change it so it will be object type (or similar) and not primitive.

src/shogun/lib/type_case.h
```diff
@@ -464,6 +471,12 @@ static const typemap sg_non_integer_typemap = {
 		ADD_TYPE_TO_MAP(floatmax_t , TYPE::T_FLOATMAX)
 		ADD_TYPE_TO_MAP(complex128_t, TYPE::T_COMPLEX128)
 };
+static const typemap sg_object_typemap = {
+        ADD_TYPE_TO_MAP(CSGObject* , TYPE::T_SGOBJECT)
+        ADD_TYPE_TO_MAP(CKernel* , TYPE::T_CKERNEL)
+        ADD_TYPE_TO_MAP(CLikelihoodModel*, TYPE::T_CLIKELIHOOD_MODEL)
+        ADD_TYPE_TO_MAP(CMeanFunction*, TYPE::T_CMEANFUNCTION)
```
avramidis: @gf712 Here too the object typemaps stuff.

src/shogun/machine/gp/EPInferenceMethod.cpp
```diff
@@ -494,22 +495,28 @@ SGVector<float64_t> CEPInferenceMethod::get_derivative_wrt_inference_method(
...
+
 	SGVector<float64_t> result;
-	int64_t len=const_cast<TParameter *>(param)->m_datatype.get_num_elements();
-	result=SGVector<float64_t>(len);
+	auto param_value = param->get_value();
+	int64_t len;
+    auto f_scalar = [&len](auto value) { len=1; };
+    auto f_vector = [&len, &param_value](auto value) { len=param_value.as<SGVector<float64_t>>().vlen; };
+    auto f_matrix = [&len, &param_value](auto value) { len=param_value.as<SGMatrix<float64_t>>().num_rows*param_value.as<SGMatrix<float64_t>>().num_cols; };
+    sg_any_dispatch(param->get_value(), sg_all_typemap, f_scalar, f_vector, f_matrix);
```
avramidis: This is needed many times so maybe we could move it in Any?

src/shogun/lib/type_case.h
gf712: Yes, just change it to `SG_ADD_OBJECT`

src/shogun/lib/type_case.h
gf712: yup, that's fine!

src/shogun/machine/gp/EPInferenceMethod.cpp
gf712: You mean the lambdas? I think it's better to leave them where they are. It's a zero cost abstraction anyway and improves readability

src/shogun/base/SGObject.cpp
```diff
@@ -657,27 +677,65 @@ void CSGObject::get_parameter_incremental_hash(uint32_t& hash, uint32_t& carry,
...
+		auto temp_val = get(parameter->first);
+		temp_val->build_gradient_parameter_dictionary(dict);
+    };
+
+    for (auto i:results)
+    {
+        dict->add(&(i.second), this);
+    }
+
+    results = get_parameters_by_property(ParameterProperties::HYPER);
```
gf712: this could be `results = self->filter(ParameterProperties::HYPER)` and drop the `get_parameters_by_property`?

src/shogun/base/SGObject.cpp
```diff
@@ -657,27 +677,65 @@ void CSGObject::get_parameter_incremental_hash(uint32_t& hash, uint32_t& carry,
...
+    for (auto i:results)
+    {
+        try
+        {
+			parameter = &i;
+            shogun::sg_any_dispatch(i.second.get_value(), shogun::sg_object_typemap, found_it_lambda);
```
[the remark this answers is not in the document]
gf712: why is it in a try catch?

avramidis:
So the idea is not only to remove the m_gradient_parameters. I think this could affect the other m_*_parameters variables. I will look at the problem again and I will try to suggest a new way for this.

About the AnyParameter and the name variable I found it odd that they name is not there but the description is. I understand that the name is somewhere else but it doesn't seem complete without it since the name is a parameter of the property. However, this is not important now.

I will try to make a new framework for the gradients but please if anyone has any idea/pointers for that please let me know.

src/shogun/base/SGObject.cpp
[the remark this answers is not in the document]
avramidis: Sorry, I was testing something and I forgot about it.

karlnapf:
The name position needs discussion definitely. Don't go out too far for the gradients yet, I was more fishing for ideas of how all this would word. What is a good way to enable manual gradient implementations of parameters? I don't like this dictionary thing a lot. I think we should start from the algorithm dev perspective here. Maybe the pattern we use is the best we can come up with but since these changes require touching a lot of stuff I think it would be good to challenge that a bit.

On the other hand, if your approach works and all the old code stuff works, that is a way forward as well, and can/should well be merged in order to just remove the `m_gradient_parameters`. Maybe we can do two things: 
* discuss all the comments raised in the PR to slowly move that forward
* draft an API to generally retrieve manually implemented gradients

gf712:
@karlnapf I don't know how the derivatives are calculated but I am assuming it's something done at the end of each iteration?
Can we not do something like the auto parameter stuff? But in this case at the end of each iteration we call all the lambdas/functions to compute the derivative and update the respective parameters.

avramidis:
Sorry for the late reply. I wasn't sure how I want to proceed. I propose to continue with "discussing all the comments raised in the PR to slowly move that forward". This PR needs a bit of work but at least we can get rid of the old m_gradient_parameters soon. After that we can start an issue to discuss and evaluate how we can change the gradient calculation mechanism. I would prefer not to be in a hurry to make large changes on that.

karlnapf:
> @karlnapf I don't know how the derivatives are calculated but I am assuming it's something done at the end of each iteration?
> Can we not do something like the auto parameter stuff? But in this case at the end of each iteration we call all the lambdas/functions to compute the derivative and update the respective parameters.

yes that sounds good. So devs could then "attach" gradient computation methods to parameters? In the sense of
```
// register gradient of `foo` wrt to `bar` which is a matrix
add_gradient("foo", "bar", SGMatrix<float64_t> my_gradient_implementation);
```
and then algorithms could call
```
SGMatrix<float64_t> grad = obj.gradient("foo", "bar");
```

and if an algorithm needs gradients of all parameters it could do
```
auto grad_params = obj.gradient_names("bar"); // returns all parameters that have a derivative wrt to "bar"
for (param : grad_params)
     step += param.gradient("bar");
```
In those lines?

karlnapf:
> Sorry for the late reply. I wasn't sure how I want to proceed. I propose to continue with "discussing all the comments raised in the PR to slowly move that forward". This PR needs a bit of work but at least we can get rid of the old m_gradient_parameters soon. After that we can start an issue to discuss and evaluate how we can change the gradient calculation mechanism. I would prefer not to be in a hurry to make large changes on that.

yes I agree. I think it might be best to first refactor before changing the design. So let's proceed with that. No rush needed indeed. So what is needed for the old gradients to work with the new system?

karlnapf:
actually can you clean up the PR a bit and make it smaller? It is super hard to review now. CI doesnt have to pass for now so you can hide using the new approach in all places. Just a single working example to start with is fine for discussion

avramidis:
I am not sure if it's possible but I will try. Basically most of the changes is replacing TParameter with AnyParameter. I understand that is difficult to see what is going on with so many changes. I will try to find a very isolated example that uses this.

Also, please don't try to review it in depth now. Currently, it is very sloppy. I have the PR so it can be seen that someone is working on it and discuss high level solutions.

gf712:
This has already been done

PR #4564, opened 2019-03-09
Add kernel meta examples

avramidis:


avramidis:
Still WIP until the data are merged

examples/meta/src/kernel/anova.sg
```diff
@@ -0,0 +1,20 @@
+File traindat = csv_file("../../data/fm_train_real.dat")
+File testdat = csv_file("../../data/fm_test_real.dat")
+
+#![create_features]
+Features feats_train = features(traindat)
+Labels feats_test = features(testdat)
+#![create_features]
+
+#![create_kernel]
+Kernel anova_kernel = kernel("ANOVAKernel")
```
karlnapf: could you use the new kwargs style here?
`Kernel k = kernel("ANOVEKernel", cardinality=2)`

examples/meta/src/kernel/anova.sg
```diff
@@ -0,0 +1,20 @@
...
+anova_kernel.put("cardinality", 2)
+anova_kernel.put("cache_size", 10)
+anova_kernel.init(feats_train, feats_train)
+#![create_kernel]
+
+#![get_kernel_features]
+RealMatrix km_test_0 = anova_kernel.get_kernel_matrix()
```
karlnapf: these names are a bit odd. What about `K_train_train` and `K_train_test`

examples/meta/src/kernel/chi2.sg
```diff
@@ -0,0 +1,20 @@
...
+
+#![create_features]
+Features feats_train = features(traindat)
+Labels feats_test = features(testdat)
+#![create_features]
+
+#![create_kernel]
+Kernel chi2_kernel = kernel("Chi2Kernel")
+chi2_kernel.put("width", 1.4)
+chi2_kernel.put("cache_size", 10)
```
karlnapf: no need to put those in the examples

avramidis:
I made the changes but I cannot change the status of the PR to ready for review. I guess only members can do that?

karlnapf:
It’s your or so I think you have to do that. Seems I can merge anyways, will check it biw

avramidis:
Now I can see the "Ready for review" button. Probably Github was glitching yesterday.

examples/meta/src/kernel/circular.sg
```diff
@@ -0,0 +1,19 @@
+File traindat = csv_file("../../data/fm_train_real.dat")
+File testdat = csv_file("../../data/fm_test_real.dat")
+
+#![create_features]
+Features feats_train = features(traindat)
+Features feats_test = features(testdat)
+#![create_features]
+
+#![create_kernel]
+Distance d = distance("EuclideanDistance", lhs=feats_train, rhs=feats_train)
```
karlnapf: Can you mark left and right hand side parameters as read only? And then we force users to use the unit function, just like done in CKernel.
Sorry I didn’t say before we just added this option

avramidis:
Changing the left and right hand parameters as read only break a lot of distance examples. @karlnapf better to do that in another PR.

karlnapf:
Conflicts... cannot merge

avramidis:
Ok. I will probably do it tomorrow.

avramidis:
Should be fine now

PR #4566, opened 2019-03-10
std::vector based DynamicObjectArray

vigsterkr:


vigsterkr:
probably CI will fail on all integration tests.... expected

karlnapf:
We cannot do that until serialisation works with the new parameter framework. These changes mean that the class skips the data array when serialised. So all Integration tests become nop

vigsterkr:
@karlnapf yes, thank you... i've just asked a review from @lisitsyn

src/shogun/lib/DynamicObjectArray.h
```diff
@@ -413,13 +301,16 @@ class CDynamicObjectArray : public CSGObject
...
 
 		/** shuffles the array (not thread safe!) */
-		inline void shuffle() { m_array.shuffle(); }
+		SG_FORCED_INLINE void shuffle() { std::random_shuffle(m_array.begin(), m_array.end()); }
 
 		/** shuffles the array with external random state */
-		inline void shuffle(CRandom * rand) { m_array.shuffle(rand); }
+		SG_FORCED_INLINE void shuffle(CRandom * rand)
+		{
+		//	m_array.shuffle(rand);
```
lisitsyn: Is that intentional?

vigsterkr: yes yes... i'm more wondering just about the implementation of clone

src/shogun/lib/DynamicObjectArray.h
```diff
@@ -35,153 +37,52 @@ class CDynamicObjectArray : public CSGObject
 		CDynamicObjectArray()
 		: CSGObject(), m_array()
 		{
-			dim1_size=1;
-			dim2_size=1;
-			dim3_size=1;
-
+			dim1_size = 1;
```
lisitsyn: I am slightly missing the point of having only 2 dimensions now. Could you explain?

vigsterkr: DynProg.cpp ... shitty stuff. :)
on the end of the day if we could use some we could drop this whole thing and have std::vector<Some<CSGObject>> and not have this whole DynamicObjectArray...

vigsterkr: i could keep the dim3 but never used and since on the end of the day we really would like to deprecate this object (see std::vector<Some<CSGobject>>)

karlnapf:
> @karlnapf yes, thank you... i've just asked a review from @lisitsyn 

You’re welcome 💋

karlnapf:
Do you guys mind if I push the commit by @lisitsyn into develop?

vigsterkr:
Sure cherry pick it as this one is already rebase fixed for develop... though i would carve out the clone support as well as without that cmp works but not clone...

vigsterkr:
closing in favour of #4573

PR #4567, opened 2019-03-10
Fix #4353

### LINKED ISSUE #4353: Clean up CFeatures::dense_dot implementations
Currently, there is a ton of copy pasted code in `CFeatures::dense_dot` (in the `.cpp`).

This task is to use C++ (11) features to replace this with a templated version that avoids the heavily redundant code.
Or even use linalg as that will be faster.

Ask if you need a pointer to get started

shiyi001:
Remove redundant code in src/shogun/features/DenseFeatures.cpp and fix #4353 .

### DIFF
--- a/src/shogun/features/DenseFeatures.cpp
+++ b/src/shogun/features/DenseFeatures.cpp
@@ -632,9 +632,29 @@ GET_FEATURE_TYPE(F_LONGREAL, floatmax_t)
 
 template<typename ST>
 float64_t CDenseFeatures<ST>::dense_dot(int32_t vec_idx1,
-        const float64_t* vec2, int32_t vec_len2) const
+        const float64_t* vec2, int32_t vec2_len) const
 {
     ASSERT(vec2_len == num_features)
+
+	int32_t vlen;
+	bool vfree;
+	ST* vec1 = get_feature_vector(vec_idx1, vlen, vfree);
+
+	ASSERT(vlen == num_features)
+	float64_t result = 0;
+
+	for (int32_t i = 0; i < num_features; i++)
+		result += vec1[i] * vec2[i];
+
+	free_feature_vector(vec1, vec_idx1, vfree);
+
+	return result;
+}
+
+template<> float64_t CDenseFeatures<bool>::dense_dot(
+		int32_t vec_idx1, const float64_t* vec2, int32_t vec2_len) const
+{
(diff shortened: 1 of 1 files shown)


shiyi001:
@karlnapf @vinx13

vigsterkr:
@shiyi001 looks good! :) if nobody has any objections i'll squashmerge this

src/shogun/features/DenseFeatures.cpp
```diff
@@ -630,194 +630,15 @@ GET_FEATURE_TYPE(F_DREAL, float64_t)
...
-template<> float64_t CDenseFeatures<uint64_t>::dense_dot(
-		int32_t vec_idx1, const float64_t* vec2, int32_t vec2_len) const
-{
-	ASSERT(vec2_len == num_features)
+    ASSERT(vec2_len == num_features)
 
 	int32_t vlen;
 	bool vfree;
-	uint64_t* vec1 = get_feature_vector(vec_idx1, vlen, vfree);
+	ST* vec1 = get_feature_vector(vec_idx1, vlen, vfree);
```
karlnapf: minor: formatting whitespace issues

src/shogun/features/DenseFeatures.cpp
```diff
@@ -830,20 +651,20 @@ template<> float64_t CDenseFeatures<uint64_t>::dense_dot(
...
 	bool vfree;
-	float32_t* vec1 = get_feature_vector(vec_idx1, vlen, vfree);
+	bool* vec1 = get_feature_vector(vec_idx1, vlen, vfree);
 
 	ASSERT(vlen == num_features)
 	float64_t result = 0;
 
 	for (int32_t i = 0; i < num_features; i++)
-		result += vec1[i] * vec2[i];
+		result += vec1[i] ? vec2[i] : 0;
```
karlnapf: minor: I wonder whether this additional implementation for `bool` could be avoided via wrapping the memory block with `SGVector<T>(vec1, vlen, false)` and then using `linalg::dot`? (since that might have the dot product for booleans implemented, but not sure actually

src/shogun/features/DenseFeatures.cpp
vigsterkr: i wonder why the clang-format job is not complaining....?
@geektoni

PR #4568, opened 2019-03-11
Fix comparison of vector element

avramidis:


avramidis:
I am not sure if that is a problem but it seems to me.

iglesias:
It seems like a reasonable fix.

I don't have all the context though, this was introduced in d1571c4e80060ce5e753caf361889e285d1e9bc0, @vinx13?

Issue #4482, opened 2019-01-23
remove EMachineType and related functions


vigsterkr:
https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/machine/Machine.h#L32

this enum needs to go because this will not be used (or usable when the plugin framework is in place).

mrkarna:
Hi Viktor,
Should I take up this task of removing all enum data types with their integer value?
Thanks

vigsterkr:
@mrkarna try to give it a go and send in a pr

mrkarna:
@vigsterkr Do I need to remove only EMachineType or all the three enums (EMachineType, ESolverType and EProblemType).
Thanks

anis-agwan:
is this issue fixed or needs to be done?

PR #4571, opened 2019-03-12
Remove EMachineType and related functions

theartful:
This PR addresses the issue #4482. I must say, this PR seems like a naive implementation of what was wanted, so sorry for potentially wasting reviewers' time.
Basically, EMachineType is never used almost anywhere in a meaningful way (each machine had a machine type getter, and never used), so I just deleted all of the occurrences.
Build fails on CI since the rest of uses are in the `gpl` submodule (getters in SVM classes). Works fine locally.

### DIFF
--- a/src/shogun/classifier/NearestCentroid.cpp
+++ b/src/shogun/classifier/NearestCentroid.cpp
@@ -8,6 +8,7 @@
 #include <shogun/labels/MulticlassLabels.h>
 #include <shogun/features/Features.h>
 #include <shogun/features/FeatureTypes.h>
+#include <shogun/features/iterators/DotIterator.h>
 
 
 
@@ -43,7 +44,6 @@ namespace shogun{
 		ASSERT(m_labels)
 		ASSERT(m_labels->get_label_type() == LT_MULTICLASS)
 		ASSERT(distance)
-		ASSERT( data->get_feature_class() == C_DENSE)
 		if (data)
 		{
 			if (m_labels->get_num_labels() != data->get_num_vectors())
@@ -54,34 +54,33 @@ namespace shogun{
 		{
 			data = distance->get_lhs();
 		}
-		int32_t num_vectors = data->get_num_vectors();
-		int32_t num_classes = ((CMulticlassLabels*) m_labels)->get_num_classes();
-		int32_t num_feats = ((CDenseFeatures<float64_t>*) data)->get_num_features();
-		SGMatrix<float64_t> centroids(num_feats,num_classes);
+		ASSERT(data->get_feature_class() == C_DENSE)
+
+		auto* multiclass_labels = m_labels->as<CMulticlassLabels>();
+		auto* dense_data = data->as<CDenseFeatures<float64_t>>();
+
+		int32_t num_classes = multiclass_labels->get_num_classes();
+		int32_t num_feats = dense_data->get_num_features();
+
+		SGMatrix<float64_t> centroids(num_feats, num_classes);
 		centroids.zero();
 
-		int64_t* num_per_class = new int64_t[num_classes];
-		for (int32_t i=0 ; i<num_classes ; i++)
-		{
-			num_per_class[i]=0;
-		}
+		SGVector<int64_t> num_per_class(num_classes);
+		num_per_class.zero();
 
-		for (int32_t idx=0 ; idx<num_vectors ; idx++)
+		auto iter_labels = multiclass_labels->get_int_labels().begin();
+		for(const auto& current : DotIterator(dense_data))
 		{
-			int32_t current_len;
-			bool current_free;
-			int32_t current_class = ((CMulticlassLabels*) m_labels)->get_label(idx);
-			float64_t* target = centroids.matrix + num_feats*current_class;
-			float64_t* current = ((CDenseFeatures<float64_t>*)data)->get_feature_vector(idx,current_len,current_free);
-			SGVector<float64_t>::add(target,1.0,target,1.0,current,current_len);
+			const auto current_class = *(iter_labels++);
+			auto target = centroids.get_column_vector(current_class);
+			auto target_vec = SGVector<float64_t>(target, num_feats);
+			current.add(1, target_vec);
 			num_per_class[current_class]++;
-			((CDenseFeatures<float64_t>*)data)->free_feature_vector(current, current_len, current_free);
 		}
-
 
 		for (int32_t i=0 ; i<num_classes ; i++)
 		{
-			float64_t* target = centroids.matrix + num_feats*i;
+			auto* target = centroids.get_column_vector(i);
 			int32_t total = num_per_class[i];
 			float64_t scale = 0;
 			if(total>1)
@@ -96,8 +95,6 @@ namespace shogun{
 
 		m_is_trained=true;
 		distance->init(centroids_feats, distance->get_rhs());
-
-		SG_FREE(num_per_class);
 
 		return true;
 	}


src/shogun/classifier/mkl/MKL.cpp
```diff
@@ -433,12 +433,6 @@ bool CMKL::train_machine(CFeatures* data)
 
 	if (interleaved_optimization)
 	{
-		if (svm->get_classifier_type() != CT_LIGHT && svm->get_classifier_type() != CT_SVRLIGHT)
-		{
-			SG_ERROR("Interleaved MKL optimization is currently "
-					"only supported with SVMlight\n");
-		}
-
```
theartful: Multikernel classes are the only classes that make use of the machine type. They use it for error checking.

theartful:
There is a pull request already addressing the issue. My mistake.

src/shogun/classifier/mkl/MKL.cpp
karlnapf: So what is the replacement here?

theartful: Dynamic casting would solve the problem, but I think it might not be the best solution since it breaks abstraction. Adding specific enum for SVMs would work, but it might defeat the point of removing machine type.

karlnapf: Add a virtual method `bool CSVM::supports_mkl()` that returns `false` by default, and then `CSVMLight` overrides this returning true. Then you can all that method

theartful: The problem here is whether MKL can make interleaving optimization using a certain svm class. I'm not sure, but this seems like information concerning the mkl class, not the svm itself, judging from the error message.

src/shogun/classifier/mkl/MKLOneClass.cpp
```diff
@@ -20,6 +20,6 @@ float64_t CMKLOneClass::compute_sum_alpha()
 
 void CMKLOneClass::init_training()
 {
+	ASSERT(dynamic_cast<CLibSVMOneClass*>(svm) != nullptr)
```
theartful: dynamic casting looks ugly to me. previous line is much cleaner
`ASSERT(svm->get_classifier_type() == CT_LIBSVMONECLASS)`

src/shogun/classifier/mkl/MKL.cpp
karlnapf: No matter how the direction is, we will need to add something to the API/parameters of those classes to communicate. See also my other comment

src/shogun/classifier/mkl/MKLOneClass.cpp
karlnapf: The issue is that we want to go for a plugin architecture of shogun soon. This means that the specialization of classes are not available/visible to the base classes anymore. Only base class interfaces will be. Therefore, this stuff will need to be communicated via the API or `get` calls

theartful: This sounds much cleaner! So this should allow people to create their custom SVMs for example using some factory interface?

src/shogun/classifier/mkl/MKL.cpp
theartful: Sorry, bear with me. I still think that something like `bool CSVM::supports_mkl()` implemented in derived classes is not the best way. Maybe we could add a property like `SVMType` and let the MKL class has a list of types that he supports this optimization on.
So for example, if MKL is a separate plugin, it can be updated to support more types without bothering the other classes.
I don't have a clear picture about this plugin interface, so if what I say doesn't make sense, say so :"D

src/shogun/classifier/mkl/MKLOneClass.cpp
karlnapf: Exactly, new SVMs would just have to serve the base class API. This is the reason why we cannot have central structures for those plugins/classes anymore -- they will only be known at runtime

theartful: I use the name instead of the enum now
```C++
ASSERT(std::string(svm->get_name()) == "LibSVMOneClass")
```

karlnapf: ah I am not so sure about this. I know it works but it is asking for trouble. As far as I can seem the SVM is never explicitly casted to LibSVMOneClass (or am I wrong?) So I would really prefer some method in `CSVM` that indicates support for whatever the MKL class wants to do/check.

theartful: I can add a method that indicates one_class, but then `SVMLightOneClass` should also satisfy this, and it's not supported judging from this line 'ASSERT(svm->get_classifier_type() == CT_LIBSVMONECLASS)'

karlnapf: The reasons why this assertion is in here seem dubious to me. What you say makes sense, checking that the SVM is one class. So let's just add that method to CSVM.

src/shogun/classifier/mkl/MKLClassification.cpp
```diff
@@ -42,11 +42,12 @@ CMKLClassification* CMKLClassification::obtain_from_generic(CMachine* machine)
 {
 	if (machine == NULL)
 		return NULL;
-
-	if (machine->get_classifier_type() != CT_MKLCLASSIFICATION)
-		SG_SERROR("Provided machine is not of type CMKLClassification!")
-
+		
 	CMKLClassification* casted = dynamic_cast<CMKLClassification*>(machine);
```
karlnapf: just remove the obtain_from_generic method completely. Wherever it was called, replace the call with `->as<CMKLClassification>()`

src/shogun/classifier/svm/SVMLight.h
```diff
@@ -595,6 +589,9 @@ class CSVMLight : public CSVM
 	float64_t *c, int32_t totdoc,int32_t iteration, int32_t *inconsistent,
 	int32_t *docs,float64_t *aicache, float64_t* maxdiff);
 
+  /** @return whether mkl interleaving optimization is possible or not */
```
karlnapf: i think there are some whitespace issues here

src/shogun/regression/svr/MKLRegression.cpp
```diff
@@ -46,6 +46,6 @@ void CMKLRegression::init_training()
 {
 	ASSERT(m_labels && m_labels->get_num_labels())
 	ASSERT(svm)
-	ASSERT(svm->get_classifier_type() == CT_SVRLIGHT)
+	ASSERT(std::string(svm->get_name()) == "SVRLight")
```
karlnapf: pls also use your interleaved optimization method here, not the name

theartful:
This is weird, the pr closed after force-pushing.

src/shogun/classifier/svm/LibSVMOneClass.h
```diff
@@ -32,11 +32,8 @@ class CLibSVMOneClass : public CSVM
 		CLibSVMOneClass(float64_t C, CKernel* k);
 		virtual ~CLibSVMOneClass();
 
-		/** get classifier type
-		 *
-		 * @return classifier type LIBSVMONECLASS
-		 */
-		virtual EMachineType get_classifier_type() { return CT_LIBSVMONECLASS; }
+		/** @return whether this svm is one-class or multi-class */
```
karlnapf: minor maybe change to "one-class or not"
(since we can have binary and multi class)

src/shogun/classifier/svm/SVM.h
```diff
@@ -242,6 +242,21 @@ class CSVM : public CKernelMachine
...
 				(CMKL* mkl, const float64_t* sumw, const float64_t suma));
 
+		/** @return whether mkl interleaving optimization is possible or not */
+		virtual bool supports_interleaving() const { return false; }
+
+		/** @return whether this svm is one-class or multi-class */
+		virtual bool is_one_class() const { return false; }
+
+		/** @return whether this svm is multiclass or one-class*/
+		bool is_multiclass() const { return !is_one_class(); }
```
karlnapf: why this?

src/shogun/classifier/svm/SVM.h
```diff
@@ -242,6 +242,21 @@ class CSVM : public CKernelMachine
...
+
+		/** @return whether this svm solves a regression problem or classification */
+		virtual bool is_regression() const { return false; }
+
+		/** @return whether this svm solves a classification problem or regression */
+		bool is_classification() const { return !is_regression(); }
```
karlnapf: this is not used, so rather remove it

theartful: it's used in MKLRegression

src/shogun/classifier/svm/SVM.h
theartful: I was thinking it maybe should be used in MKLMulticlass?

karlnapf: no that would be quite different ...

src/shogun/classifier/svm/SVM.h
karlnapf: I cannot find it in the diff

theartful: I thought I would ask first. Currently `MKLClassification` doesn't check that the SVM is not a regression one.

src/shogun/classifier/svm/SVM.h
theartful: As I understand it, the one class SVMs, are used with the MKLOneClass, and similarily MKLMulticlass is used with the rest, so MKLMulticlass should check for this condition. If that's not the case, the it's useless

src/shogun/classifier/svm/SVM.h
karlnapf: Actually, I just rememberd that we have an enum for this: `EProblemType` defined in `CMachine.h`  ... this can be reused here rather than adding some of the methods (interleaved needs to stay I guess)

vigsterkr:
@theartful great effort but here's some comment:
instead of the `is_*` function, could you please do the following:
`virtual uint32_t supports() const`
in the unsigned int we store bitmask, that describes the machine in question.

something along this line
```
enum Input {
 DenseFeatures 1 << 1;
SparseFeatures 1 << 2;
...
```

similar to the machine type and then just use bitmask operations to set the unit32_t

src/shogun/classifier/svm/SVM.h
theartful: This can be replaced by EProblemType as well, since it has unary, binary, and multiclass.

src/shogun/classifier/svm/LibSVMOneClass.h
```diff
@@ -21,6 +21,10 @@ namespace shogun
 class CLibSVMOneClass : public CSVM
 {
 	public:
+
+		/** problem type */
+		MACHINE_PROBLEM_TYPE(PT_CLASS);
```
theartful: not sure about this. does PT_CLASS means its one class classification?

vigsterkr: yes

vigsterkr: it is a binary classifier

src/shogun/classifier/mkl/MKLOneClass.cpp
```diff
@@ -21,5 +21,8 @@ float64_t CMKLOneClass::compute_sum_alpha()
 void CMKLOneClass::init_training()
 {
 	ASSERT(svm)
-	ASSERT(svm->get_classifier_type() == CT_LIBSVMONECLASS)
+	ASSERT(svm->get_machine_problem_type() == PT_CLASS)
+	// TODO: FIXME this currently supports LibSVMOneClass only
+	// should also support SVMLightOneClass
+	ASSERT(std::string(svm->get_name()) != "SVMLightOneClass")
```
vigsterkr: i would refrain myself for this type checking... you can use `std::type_index(typeid)`

src/shogun/classifier/svm/SVM.h
```diff
@@ -242,6 +242,9 @@ class CSVM : public CKernelMachine
 		void set_callback_function(CMKL* m, bool (*cb)
 				(CMKL* mkl, const float64_t* sumw, const float64_t suma));
 
+		/** @return whether mkl interleaving optimization is possible or not */
+		virtual bool supports_interleaving() const { return false; }
```
vigsterkr: goes into supports

src/shogun/regression/svr/MKLRegression.cpp
```diff
@@ -46,6 +46,11 @@ void CMKLRegression::init_training()
 {
 	ASSERT(m_labels && m_labels->get_num_labels())
 	ASSERT(svm)
-	ASSERT(svm->get_classifier_type() == CT_SVRLIGHT)
+	ASSERT(svm->get_machine_problem_type() == PT_REGRESSION)
+	
+	// TODO: FIXME MKLRegression currently only supports LibSVR
+	// should also support SVRLight
+	ASSERT(std::string(svm->get_name()) != "SVRLight")
```
vigsterkr: please dont do type checking like this...

src/shogun/regression/svr/SVRLight.h
```diff
@@ -186,6 +180,13 @@ class CSVRLight: public CSVMLight
 			int32_t *docs,float64_t *aicache,
 			float64_t* maxdiff);
 
+		/** @return whether this svm solves a regression problem or classification */
+		virtual bool is_regression() const
```
vigsterkr: supports()

theartful: this should go. I don't think supports is needed now, only "supports_interleaving" is left, and the rest comes from EProblemType

vigsterkr: `supports_interleaving` should go and add the described supports() function...

vigsterkr: there we can add what sort of input type it supports etc.... way more extensible than having a simple function that returns a boolean for a certain small thing

vigsterkr: and that interface should actually go into CMachine

src/shogun/classifier/mkl/MKLOneClass.cpp
karlnapf: will that work if the type `SVMLightOneClass` is not known? (plugins)

src/shogun/classifier/svm/LibSVMOneClass.h
karlnapf: any idea why we have `PT_BINARY` and `PT_CLASS`?

src/shogun/regression/svr/SVRLight.h
karlnapf: it is what EProblemType is for I think

src/shogun/classifier/mkl/MKLOneClass.cpp
vigsterkr: this class already obviously depends on `SVMLightOneClass` so yes then it should work...

src/shogun/regression/svr/SVRLight.h
theartful: The CMachine is general purpose, but supports_interleaving only concerns MKL, is this ok? Also problem type and label type currently have their own enums (ELabelType and EMachineType), and they correspond. Sorry just want to make sure

vigsterkr: you can encode in one bit whether a machine is interleaving or not...

src/shogun/classifier/mkl/MKLOneClass.cpp
karlnapf: The dependency is kinda unnecessary through the ctor only if you look at it (and should be removed imo). In general, the MKL stuff should work with pretty much any SVM solver as long as there is some interface defined.

vigsterkr: well mkl doesn't work with anything else but svmlight* stuff ... just saying.

src/shogun/classifier/svm/LibSVMOneClass.h
theartful: `PT_CLASS` doesn't have its own `apply` version, and it makes tests fail if used.

src/shogun/classifier/mkl/MKLClassification.cpp
```diff
@@ -36,17 +36,5 @@ void CMKLClassification::init_training()
...
-	if (machine == NULL)
-		return NULL;
-
-	if (machine->get_classifier_type() != CT_MKLCLASSIFICATION)
-		SG_SERROR("Provided machine is not of type CMKLClassification!")
-
-	CMKLClassification* casted = dynamic_cast<CMKLClassification*>(machine);
-	SG_REF(casted)
-	return casted;
+	ASSERT(svm->get_machine_problem_type() != PT_REGRESSION)
```
karlnapf: Why assert that it is NOT regression? Rather than asserting it is a certain type?

src/shogun/machine/Machine.cpp
```diff
@@ -310,3 +306,23 @@ CLatentLabels* CMachine::apply_locked_latent(SGVector<index_t> indices)
...
+}
+
+bool CMachine::supports(uint64_t property_mask)
+{
+	return (m_supports & property_mask) == property_mask;
+}
+
+void CMachine::add_support(uint64_t property_mask)
+{
+	m_supports |= property_mask;
```
karlnapf: @gf712 will have some comments

src/shogun/machine/Machine.cpp
```diff
@@ -310,3 +306,23 @@ CLatentLabels* CMachine::apply_locked_latent(SGVector<index_t> indices)
...
+}
+
+void CMachine::remove_support(uint64_t property_mask)
```
karlnapf: Not sure we need this one?

src/shogun/machine/Machine.cpp
```diff
@@ -26,6 +26,7 @@ CMachine::CMachine()
 	    &m_store_model_features, "store_model_features",
 	    "Should feature data of model be stored after training?");
 	SG_ADD(&m_data_locked, "data_locked", "Indicates whether data is locked");
+	SG_ADD(&m_supports, "m_supports", "Bitset representing machine properties");
```
karlnapf: This is a property that shouldn’t be serialised as it remains unchanged during the instance life. So we can have a property for that maybe?
@gf712

src/shogun/regression/svr/MKLRegression.cpp
```diff
@@ -46,6 +47,10 @@ void CMKLRegression::init_training()
 {
 	ASSERT(m_labels && m_labels->get_num_labels())
 	ASSERT(svm)
-	ASSERT(svm->get_classifier_type() == CT_SVRLIGHT)
+#ifdef USE_SVMLIGHT
+	ASSERT(typeid(*svm) == typeid(CSVRLight))
+#else
+	SG_ERROR("MKLRegression only works with SVRLight")
```
karlnapf: “Shogun needs to be compiled with svmlight enabled in order to use %s”

src/shogun/machine/Machine.cpp
gf712: maybe we can write some sort of base class to handle the bit logical operations in shogun to avoid having to rewrite these?

gf712: and that can then be used inside AnyParameterProperties and here? Just thinking that at some point someone will make a mistake.. But these look fine

karlnapf: Very good idea

src/shogun/machine/Machine.cpp
gf712: also I think the operation should be like here https://github.com/shogun-toolbox/shogun/blob/6434fe2df14bb0dbc34c27b2da4c7147f67b1159/src/shogun/base/AnyParameter.h#L136
You should use the complement

src/shogun/machine/Machine.cpp
gf712: yes, that should be easy to add, just do it here:
https://github.com/shogun-toolbox/shogun/blob/6434fe2df14bb0dbc34c27b2da4c7147f67b1159/src/shogun/base/AnyParameter.h#L34-L42
and add the property in SG_ADD, i.e. `SG_ADD(..., ParameterProperties::NOT_SERIALISABLE)`

karlnapf: Actually the best thing would be to just not register it I now realised. No get/set/clone/equals/save needed

src/shogun/machine/Machine.cpp
theartful: in case a child class doesn't support a property of its parent class

src/shogun/regression/svr/MKLRegression.cpp
```diff
@@ -46,6 +47,10 @@ void CMKLRegression::init_training()
 {
 	ASSERT(m_labels && m_labels->get_num_labels())
 	ASSERT(svm)
-	ASSERT(svm->get_classifier_type() == CT_SVRLIGHT)
+#ifdef USE_SVMLIGHT
+	ASSERT(typeid(*svm) == typeid(CSVRLight))
```
vigsterkr: this is not gonna work... see the ref manual of c++ for details:
https://en.cppreference.com/w/cpp/language/typeid

in short use this
```
std::type_index(typeid(*svm)) == std::type_index(typeid(CSVRLight))
```

theartful: done. should I send PR in the gpl submodule, or is this still not ready?

vigsterkr: yeah it'd be good to see how the CI react on this change

src/shogun/machine/Machine.h
```diff
@@ -116,6 +51,12 @@ enum EProblemType
 	PT_CLASS = 5
 };
 
+/** SVM properties */
+enum SVMProperties : uint64_t
```
vigsterkr: the problem with this design is that actually this encourages that there are as many enums (and values) as many you like... this will cause problems when we have collision in bit values across enums and then the supports will report false values....

theartful: Should we "namespace" them. For example, any general machine attribute will have specific number of bits from 0 to m. Any svm machine can take from m to n, and so on.

src/shogun/classifier/mkl/MKLOneClass.cpp
```diff
@@ -21,5 +23,7 @@ float64_t CMKLOneClass::compute_sum_alpha()
 void CMKLOneClass::init_training()
 {
 	ASSERT(svm)
-	ASSERT(svm->get_classifier_type() == CT_LIBSVMONECLASS)
+	ASSERT(
+	    std::type_index(typeid(*svm)) ==
+	    std::type_index(typeid(CLibSVMOneClass)))
```
vinx13: `ASSERT(svm->as<CLibSVMOneClass>())` ?

src/shogun/regression/svr/MKLRegression.cpp
```diff
@@ -46,6 +48,11 @@ void CMKLRegression::init_training()
 {
 	ASSERT(m_labels && m_labels->get_num_labels())
 	ASSERT(svm)
-	ASSERT(svm->get_classifier_type() == CT_SVRLIGHT)
+#ifdef USE_SVMLIGHT
+	ASSERT(std::type_index(typeid(*svm)) == std::type_index(typeid(CSVRLight)))
```
vinx13: `ASSERT(svm->as<CSVRLight>())`

src/shogun/classifier/mkl/MKLOneClass.cpp
vigsterkr: this is faster.

src/shogun/regression/svr/MKLRegression.cpp
vigsterkr: same as above... faster

src/shogun/classifier/mkl/MKLOneClass.cpp
vinx13: okay

karlnapf:
Giving up on this one? :)

PR #4572, opened 2019-03-14
Cleanup the Parameter class

theartful:
I cleaned up the parameter class by using templates and SFINAE instead of defining all these functions.

### DIFF
--- a/src/shogun/lib/DataType.h
+++ b/src/shogun/lib/DataType.h
@@ -12,11 +12,20 @@
 #include <shogun/io/SGIO.h>
 #include <shogun/lib/common.h>
 #include <shogun/lib/config.h>
+#include <shogun/lib/type_case.h>
+
 
 #define PT_NOT_GENERIC	PT_SGOBJECT
 
 namespace shogun
 {
+
+class CSGObject;
+template <class ST> class SGString;
+template <class T> class SGMatrix;
+template <class T> class SGSparseMatrix;
+template <class T> class SGVector;
+template <class T> class SGSparseVector;
 
 #ifndef DOXYGEN_SHOULD_SKIP_THIS
 enum EContainerType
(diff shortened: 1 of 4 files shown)


src/shogun/lib/DataType.h
```diff
@@ -40,22 +49,75 @@ enum EStructType
...
+	PT_UINT32=(int)TYPE::T_UINT32,
+	PT_INT64=(int)TYPE::T_INT64,
+	PT_UINT64=(int)TYPE::T_UINT64,
+	PT_FLOAT32=(int)TYPE::T_FLOAT32,
+	PT_FLOAT64=(int)TYPE::T_FLOAT64,
+	PT_FLOATMAX=(int)TYPE::T_FLOATMAX,
+	PT_SGOBJECT=(int)TYPE::T_SGOBJECT,
+	PT_COMPLEX128=(int)TYPE::T_COMPLEX128,
+	PT_UNDEFINED=(int)TYPE::T_UNDEFINED
+};
```
theartful: Not sure about this bit. I wanted to reuse `sg_type` without having to redefine it for new enum values.

src/shogun/lib/DataType.h
```diff
@@ -40,22 +49,75 @@ enum EStructType
...
+
+namespace type_internal
+{
+	template<template<class> class CT, class T>
```
theartful: Also not sure of polluting the `type_internal` namespace like this is ok

src/shogun/lib/DataType.h
```diff
@@ -40,22 +49,75 @@ enum EStructType
...
+	template <typename T>                                                                \
+	struct is_sg_container<CT, T> : public std::true_type                             \
+	{                                                                          \
+	};
+
+	#define SG_ADD_STRUCT_TYPE(ST, type_)                                         \
+	template <typename T>                                                                \
+	struct sg_struct_type<ST, T>                                                          \
+	{                                                                          \
+		static constexpr EStructType stype = type_;                                   \
```
gf712: maybe just type instead of stype? I think it would make it a bit cleaner

src/shogun/lib/type_case.h
```diff
@@ -158,7 +157,7 @@ namespace shogun
 			return msg;
 		}
 
-		TYPE get_type(const Any& any, const typemap& map)
+		static inline TYPE get_type(const Any& any, const typemap& map)
```
gf712: Thanks for that, it was a bug that I had missed before!

theartful:
No. The Parameter class didn't use templates, so there was a whole lot of functions that I deleted. I'm currently trying to solve the whole lot of test failures that this caused :"D

src/shogun/lib/DataType.h
theartful: I mimicked the `ptype` in `sg_primitive_type`

gf712: ah ok, maybe that should be changed too then. Imo it's better to have a consistent member `type` in these structs

gf712:
> No. The Parameter class didn't use templates, so there was a whole lot of functions that I deleted. I'm currently trying to solve the whole lot of test failures that this caused :"D

Ah ok! Is there any chance you can unindent the code in the namespace to get a smaller diff. Right now it's really hard to see what you changed.. It's a bit annoying that certain parts of the code weren't properly formatted before :(

theartful:
Sure. Once this passes tests locally I will push the changes without the formatting (I had problems with local tests before, so sorry for prematurely making pull requests)

karlnapf:
Hey!
Thanks for the initiative. Keep in mind we soon want to delete the whole class, so please don’t spend too much time on it. Smaller clean ups are still appreciated though

karlnapf:
And as @gf712 said, it’s not possible to review this due to the large diff. It is often good to split prs into two where one doesn’t have functional changes but only formatting, and then another one that does the changes.

PR #4575, opened 2019-03-15
Make a function in 'type_case.h' static inline

theartful:


src/shogun/lib/type_case.h
```diff
@@ -145,7 +145,7 @@ namespace shogun
 #undef SG_ADD_SGVECTOR_TYPE
 #undef SG_ADD_SGMATRIX_TYPE
 
-		std::string print_map(const typemap& map)
+		static inline std::string print_map(const typemap& map)
```
karlnapf: if you want to inline things use `SG_FORCE_INLINE`

karlnapf:
thanks for this!

@gf712 have a look as well

src/shogun/lib/type_case.h
gf712: this is an inline specifier for the linker, the code itself doesn't have to be inlined, so don't have to force it

gf712:
> thanks for this!
> 
> @gf712 have a look as well

Yup, this is the issue that @geektoni had in his PR

src/shogun/lib/type_case.h
karlnapf: I see thanks!

karlnapf:
Ill wait for CI and then merge if there are no complaints

src/shogun/lib/type_case.h
vigsterkr: `SG_FORCED_INLINE` plz

src/shogun/lib/type_case.h
```diff
@@ -158,7 +158,7 @@ namespace shogun
 			return msg;
 		}
 
-		TYPE get_type(const Any& any, const typemap& map)
+		static inline TYPE get_type(const Any& any, const typemap& map)
```
vigsterkr: same here...

src/shogun/lib/type_case.h
theartful: @vigsterkr this is to satisfy the one definition rule for the linker. should it be `SG_FORCED_INLINE`ed?

vigsterkr: yes plz force it

gf712:
@karlnapf this should be good to go

geektoni:
What's the status of this one? Because I would really love to have it merged soon :P

PR #4576, opened 2019-03-16
Remove converter from swig

vinx13:


karlnapf:
Approval modulo python failed

PR #4577, opened 2019-03-16
Convert multiclass labels to binary labels

gf712:
@karlnapf this is what I did in the openml branch to get binary features from multiclass. Is this what you meant?

### DIFF
--- a/src/shogun/labels/BinaryLabels.cpp
+++ b/src/shogun/labels/BinaryLabels.cpp
@@ -12,6 +12,8 @@
 #include <shogun/lib/SGVector.h>
 #include <shogun/mathematics/Statistics.h>
 
+#include <set>
+
 using namespace shogun;
 
 CBinaryLabels::CBinaryLabels() : CDenseLabels()
@@ -139,8 +141,28 @@ CLabels* CBinaryLabels::duplicate() const
 	return new CBinaryLabels(*this);
 }
 
-namespace shogun
-{
+
+namespace shogun {
+	SG_FORCED_INLINE Some<CBinaryLabels> from_multiclass(CMulticlassLabels* orig) {
+		auto result_vector = orig->get_labels();
+		std::set<int32_t> unique(result_vector.begin(), result_vector.end());
+		auto min = *std::min_element(unique.begin(), unique.end());
+
+		ASSERT(unique.size() == 2);
+
+		SGVector<int32_t> converted(result_vector.vlen);
+		std::transform(
+				result_vector.begin(), result_vector.end(), converted.begin(),
+				[&min](int32_t old_label) {
+					if (old_label == min)
+						return -1;
+					else
+						return 1;
+				});
+
+		return some<CBinaryLabels>(converted);
+	}
+
 	Some<CBinaryLabels> binary_labels(CLabels* orig)
 	{
 		REQUIRE(orig, "No labels provided.\n");
@@ -150,6 +172,8 @@ namespace shogun
 			{
 			case LT_BINARY:
 				return Some<CBinaryLabels>::from_raw((CBinaryLabels*)orig);
+			case LT_MULTICLASS:
+				return from_multiclass((CMulticlassLabels*)orig);
 			default:
 				SG_SNOTIMPLEMENTED
 			}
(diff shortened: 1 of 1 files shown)


src/shogun/labels/BinaryLabels.cpp
```diff
@@ -139,8 +141,28 @@ CLabels* CBinaryLabels::duplicate() const
...
-namespace shogun
-{
+
+namespace shogun {
+	SG_FORCED_INLINE Some<CBinaryLabels> from_multiclass(CMulticlassLabels* orig) {
+		auto result_vector = orig->get_labels();
+		std::set<int32_t> unique(result_vector.begin(), result_vector.end());
+		auto min = *std::min_element(unique.begin(), unique.end());
+
+		ASSERT(unique.size() == 2);
```
karlnapf: User facing error msg pls

karlnapf: I.e. in the lines of `REQUIRE("%s must contain less than two unique labels, contains %d. Cannot convert to binary labels.\n", orig->get_name(), unique.size());`

src/shogun/labels/BinaryLabels.cpp
```diff
@@ -139,8 +141,28 @@ CLabels* CBinaryLabels::duplicate() const
...
+
+		SGVector<int32_t> converted(result_vector.vlen);
+		std::transform(
```
karlnapf: I wonder whether we *always* should print a warning with the conversion table in case we convert (especially if we change the representation). As doing it implicit might lead to confusion. Thoughts?

gf712: I think we need something like that. Because ::apply will return labels with either -1/+1, which is probably not what the user wanted... But then warnings can be easily missed or be too verbose if you do something like a gridsearch

karlnapf: Actually, technically we only need one unique label. For training that never makes sense. But I think evaluation methods might complain since they also convert. Think test labels all are positive (totally reasonable). So this needs a modification.

karlnapf: Ok so this brings up another issue: What if the training labels are somewhat different than the testing labels...will the conversions be the same? If not, what do we do?

karlnapf: I think the grid-search thing would need to convert only once at the beginning anyways as otherwise  things will be inefficient ... and also might be buggy (see comment on conversion mapping invariance). So for those one off cases, I would rather warn the user or?

gf712: OK, I will add a warning. I think in openml this will generate lots of warnings when they do the learning for each fold, but I can disable warnings there. I will add the warning with a mapping.

karlnapf: Mmmh
I wonder whether we could do the conversion on the base class just like we did the dense feature type dispatcher.
Because you are of course right

src/shogun/labels/BinaryLabels.cpp
```diff
@@ -141,6 +143,32 @@ CLabels* CBinaryLabels::duplicate() const
 
 namespace shogun
 {
+	SG_FORCED_INLINE Some<CBinaryLabels>
+	from_multiclass(CMulticlassLabels* orig)
+	{
+		auto result_vector = orig->get_labels();
+		std::set<int32_t> unique(result_vector.begin(), result_vector.end());
```
vigsterkr: do we need ordering? :) i guess unordered_set would just do it right?

vigsterkr: or we dont need the next line, and just take the `.begin()` or?

karlnapf: Might be good in order to make sure we have some invariance if the mapping to permutations of the label vector?

src/shogun/labels/BinaryLabels.cpp
gf712: I think this would be best, and would avoid any confusions

karlnapf: ok, ideas how to get going with that?

gf712:
@karlnapf do you think it still makes sense to merge this (after rebase)?

karlnapf:
Another one that is better solved by @LiuYuHui label encoder (which should be deployed soon!)

PR #4579, opened 2019-03-18
Automatic estimation of GaussianKernel log_width

gf712:
Estimate log_width as described in https://arxiv.org/pdf/1707.07269.pdf


src/shogun/lib/auto_initialiser.h
```diff
@@ -29,36 +34,93 @@ namespace shogun
 
 			~GammaFeatureNumberInit() override = default;
 
-			Any operator()() override
+			Any operator()() final
```
gf712: @vigsterkr shouldn't this be `final` rather than `override`?

src/shogun/lib/auto_initialiser.h
```diff
@@ -29,36 +34,93 @@ namespace shogun
...
-								dot_features->get_std(false)[0]));
-					}
-					default:
-						return make_any(
-							1.0 /
-							static_cast<float64_t>(
-								((CDotFeatures*)lhs)->get_dim_feature_space()));
+				case EFeatureClass::C_DENSE:
+				{
+					auto dense_features = (CDenseFeatures<float64_t>*)lhs;
```
gf712: I changed this to use the Eigen version of standard deviation because it should be more efficient?

src/shogun/lib/auto_initialiser.h
```diff
@@ -29,36 +34,93 @@ namespace shogun
...
+					for (int i = 0; i < lhs->get_num_vectors(); ++i) {
+						for (int j = i + 1; j < lhs->get_num_vectors(); ++j) {
+							result.set_element(pdist.get_element(i, j), idx);
+							++idx;
+						}
+					}
+					return make_any(std::log(linalg::median(result) / 2.0) / 2.0);
+				}
+				default:
+					return make_any(1.0);
```
gf712: need to change this to std::log(0.5) / 2.0, like the default value in GaussianKernel

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
```diff
@@ -1466,6 +1468,56 @@ TYPED_TEST(LinalgBackendEigenAllTypesTest, SGMatrix_std_deviation)
...
+TYPED_TEST(LinalgBackendEigenAllTypesTest, SGVector_median)
+{
+	SGVector<TypeParam> vec(9);
+	for (index_t i = 0; i < vec.size(); ++i)
+		vec[i] = i;
+
+	std::random_device rd;
+	std::mt19937 g(rd());
+
+	std::shuffle(vec.begin(), vec.end(), g);
```
gf712, line 1662: @vigsterkr I wasn't sure how to do a shuffle with SG_* so I just used the stl

src/shogun/lib/auto_initialiser.h
```diff
@@ -29,36 +34,93 @@ namespace shogun
...
+					auto result = SGVector<float64_t>((lhs->get_num_vectors()*lhs->get_num_vectors()-lhs->get_num_vectors())/2);
+					// copy upper triangular wihout a particular order
+					index_t idx = 0;
+					for (int i = 0; i < lhs->get_num_vectors(); ++i) {
+						for (int j = i + 1; j < lhs->get_num_vectors(); ++j) {
+							result.set_element(pdist.get_element(i, j), idx);
+							++idx;
+						}
+					}
+					return make_any(std::log(linalg::median(result) / 2.0) / 2.0);
```
gf712: could maybe add a static function in GaussianKernel to do the `std::log(width / 2.0) / 2.0` conversion?

src/shogun/lib/auto_initialiser.h
```diff
@@ -29,36 +31,107 @@ namespace shogun
...
-							static_cast<float64_t>(
-								((CDotFeatures*)lhs)->get_dim_feature_space()));
+					return make_any(
+					    std::log(linalg::median(result) / 2.0) / 2.0);
+				}
+				default:
+				{
+#ifdef __builtin_log
+					constexpr float64_t default_value =
+					    __builtin_log(0.5) / 2.0;
```
gf712: saving some precious clock cycles...

src/shogun/lib/auto_initialiser.h
```diff
@@ -29,36 +31,107 @@ namespace shogun
 
 			~GammaFeatureNumberInit() override = default;
 
-			Any operator()() override
+			Any operator()() final
 			{
-				REQUIRE(m_kernel != nullptr, "m_kernel is not pointing to a CKernel object");
+				REQUIRE(
+				    m_kernel != nullptr,
+				    "m_kernel is not pointing to a CKernel object");
```
karlnapf: Can you change to: no kernel provided

karlnapf:
Awesome very useful

karlnapf:
Ah this might blow some Integration tests as default value was changed

karlnapf:
I like the idea of adding a static function to convert between prameteisations of the kernel

src/shogun/lib/auto_initialiser.h
```diff
@@ -29,36 +31,107 @@ namespace shogun
...
+				auto lhs = m_kernel->get_lhs();
+				auto rhs = m_kernel->get_rhs();
+				switch (lhs->get_feature_class())
+				{
+				case EFeatureClass::C_DENSE:
+				{
+					auto distance_kernel = CEuclideanDistance();
+					distance_kernel.init(lhs, rhs);
+					auto pdist = distance_kernel.get_distance_matrix();
+					auto result = SGVector<float64_t>(
```
vigsterkr: do we need to copy this... i mean currently you already rely on sgvector being on cpu memory so just then take the eigen map vector of it...

gf712: What do you mean? I still need to get everything from one of the triangles. The triangular view from Eigen just returns a matrix with 0s in the lower/upper triangle, I still need the non zero values?

src/shogun/features/DenseFeatures.cpp
```diff
@@ -829,7 +829,15 @@ SGVector<ST> CDenseFeatures<ST>::mean() const
 }
 
 template <typename ST>
-SGVector<float64_t > CDenseFeatures<ST>::std(bool colwise) const
+ST CDenseFeatures<ST>::median() const
+{
+	ASSERT_FLOATING_POINT
```
karlnapf: do you need this, median of features rather than the matrix?

src/shogun/kernel/GaussianKernel.h
```diff
@@ -120,7 +120,29 @@ class CGaussianKernel: public CShiftInvariantKernel
 	 */
 	SG_FORCED_INLINE float64_t get_width() const
 	{
-		return std::exp(m_log_width * 2.0) * 2.0;
+		return CGaussianKernel::from_log_width(m_log_width);
+	}
+
+	/**
+	 * Converts width to log domain.
```
karlnapf: this is not quite precise. I would actually ditch the log in the conversion and rather convert from `$-(x-y)^2/\gamma$` parametrisation to `$(-(x-y)^2 / {2\sigma^2})$` (and back), i.e. there must be a square root and a division by two. Then add a log around this explicitly. The log parametrisation has different reasons, and also putting/removing a log is trivial and doesnt twist your brain.

karlnapf: Oh and whatever it does, make sure to add the math in the doxygen so it is clear

src/shogun/lib/auto_initialiser.h
karlnapf: technically might would be possible to compute the median of upper/lower triangular values without copying them. But I am not so sure that is worth it?

karlnapf: However, I think what I would like to see is some analogue to the scipy way of computing pairwise distances, with a function that returns the distances as a vector, and then an "expansion" function. This way you can call the vector function here and then sort that directly, and you save yourself from ever storing the other half of the matrix. This also solves Viktor's comment above.

src/shogun/lib/auto_initialiser.h
```diff
@@ -29,36 +32,102 @@ namespace shogun
...
+					    (lhs->get_num_vectors() * lhs->get_num_vectors() -
+					     lhs->get_num_vectors()) /
+					    2);
+					// copy upper triangular wihout a particular order
+					index_t idx = 0;
+					for (int i = 0; i < lhs->get_num_vectors(); ++i)
```
karlnapf: `for (auto i : range(lhs->get_num_bvectors()))` ?

src/shogun/lib/auto_initialiser.h
```diff
@@ -29,36 +32,102 @@ namespace shogun
...
 					{
-						auto dot_features = (CDotFeatures*)lhs;
-						return make_any(
-						    1.0 /
-						    (static_cast<float64_t>(
-								dot_features->get_dim_feature_space()) *
-								dot_features->get_std(false)[0]));
+						for (int j = i + 1; j < lhs->get_num_vectors(); ++j)
```
karlnapf: It would be cool if we could not have explicit loops for traversing triangular matrices everywhere, but rather do this via a utility iterator or something. See comment on the pdist as vector above

src/shogun/mathematics/linalg/LinalgNamespace.h
```diff
@@ -1538,20 +1538,52 @@ namespace shogun
...
+		 * composed of real numbers.
+		 *
+		 * @param a SGVector or SGMatrix
+		 * @return The whole matrix median
+		 */
+		template <
+		    typename T, template <typename> class Container,
+		    typename std::enable_if_t<!std::is_same<T, complex128_t>::value>* =
+		        nullptr>
+		T median(const Container<T>& a)
```
karlnapf: some food for thought on fast median computation: https://www.stat.cmu.edu/~ryantibs/papers/median.pdf

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
```diff
@@ -1466,6 +1468,56 @@ TYPED_TEST(LinalgBackendEigenAllTypesTest, SGMatrix_std_deviation)
 	    result.get_element(0), 2.581988897471611, get_epsilon<TypeParam>());
 }
 
+TYPED_TEST(LinalgBackendEigenAllTypesTest, SGMatrix_median)
+{
+	const index_t nrows = 3, ncols = 3;
+	SGMatrix<TypeParam> mat(nrows, ncols);
+	for (index_t i = 0; i < mat.size(); ++i)
+		mat[i] = i;
```
karlnapf: I would generate a set of numbers with multiples to be a more "realistic" case, but I guess that is minor

gf712: I just shuffled the values around a bit to make it more realistic, but I can add a multiple there.

src/shogun/lib/auto_initialiser.h
gf712: I guess we could add a function in CDistance that does that? And then add a static function that  can do the conversion to square form?

src/shogun/features/DenseFeatures.cpp
gf712: I wanted to add it, but then I realised that the iterators go along each training example, rather than column, so it became a bit of a pain without copying data...

src/shogun/mathematics/linalg/LinalgNamespace.h
gf712: I guess we can add the option when the exact value isn't required? I am just thinking that the stl implementation must be quite efficient already no?

src/shogun/lib/auto_initialiser.h
gf712: and the new class member function would do this https://stackoverflow.com/a/13079806 ?

gf712:
> Unit test for the bandwidth?

Yes, I still need to add that

src/shogun/lib/auto_initialiser.h
karlnapf: Yes. In fact the method for the full matrix would use the condensed form and expand ir

src/shogun/mathematics/linalg/LinalgNamespace.h
karlnapf: Yes just wanted to share it

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
karlnapf: Probably fine

src/shogun/lib/auto_initialiser.h
gf712: do all the distances in shogun satisfy the triangle inequality and are symmetric? i.e. d(A,B) == d(B,A)

gf712: nvm, they do, the base class has this written :D

karlnapf:
Updates? :)

gf712:
> Updates? :)

Ah yes, I meant to ask if for the unit tests I should then explicitly instantiate everything with log_width 1, which was the old default? or change all the values with automatic width deduction?

karlnapf:
I think it is safe to change to auto. (No idea what CI will say)

gf712:
> I think it is safe to change to auto. (No idea what CI will say)

OK, I'll change the labels and kernel matrix values in the tests where necessary then

karlnapf:
this should go in before the release :)

gf712:
need to rebase and finish this

src/shogun/kernel/GaussianKernel.h
gf712: I am not sure I remember what this comment was about. Shouldn't we have this helper to convert width <-> log_width?

src/shogun/features/DenseFeatures.cpp
gf712: removed as I ended up just using linalg

gf712:
@karlnapf do you remember/know if this could be used with the squared euclidean distance? If so can reuse the distance object from the GaussianKernel when computing the median

src/shogun/base/class_list.cpp.py
```diff
@@ -62,7 +62,8 @@
         "NullFileSystem", "FilterVisitor", "RandomMixin", "MaxCrossValidation",
         "StreamingDataFetcher", "MaxMeasure", "MaxTestPower",
         "MedianHeuristic", "WeightedMaxMeasure", "WeightedMaxTestPower",
-        "Seedable", "ShogunEnv", "ShapeVisitor", "FeatureImportanceTree"]
+        "Seedable", "ShogunEnv", "ShapeVisitor", "FeatureImportanceTree",
```
karlnapf, line 65: what were these for again?

src/shogun/features/DenseFeatures.cpp
karlnapf: ok makes more sense imo

src/shogun/kernel/GaussianKernel.h
karlnapf: yes, and also maybe converting between th edifferent parametrisations (see above)
Just to keep the code readable ....

karlnapf:
> @karlnapf do you remember/know if this could be used with the squared euclidean distance? If so can reuse the distance object from the GaussianKernel when computing the median

what exactly do you mean?

src/shogun/base/class_list.cpp.py
gf712: Need to remove `GaussianWidthAutoInit` from class_list.cpp (we did the same with the gamma "GammaFeatureNumberInit"), as it is not an SGObject

gf712:
> > @karlnapf do you remember/know if this could be used with the squared euclidean distance? If so can reuse the distance object from the GaussianKernel when computing the median
> 
> what exactly do you mean?

can we use the `dist` object used here https://github.com/shogun-toolbox/shogun/blob/1414930317f942965607bd4caa887c814d9d09fd/src/shogun/kernel/GaussianKernel.cpp#L139-L141 to compute the pairwise distance (I guess we need to then apply the sqrt, but that would be for one scalar, the median)? And then use the median of that to compute this heuristic? If so, this would make this log_width estimation much cheaper because then we reuse the `dist` object in the GaussianKernel and we can reuse the precomputed lhs and rhs

src/shogun/lib/auto_initialiser.h
```diff
@@ -7,61 +7,121 @@
...
+					    GaussianKernel::to_log_width(m_kernel.get_width());
+					return make_any(default_value);
+				}
 				}
 			}
 
 		private:
-			Kernel* m_kernel;
-			SG_DELETE_COPY_AND_ASSIGN(GammaFeatureNumberInit);
+			GaussianKernel& m_kernel;
```
gf712, line 139: I made this change so we can avoid a check for nullptr (a reference can never be null so we know that what we passed to the constructor has to be a valid object). Ideally this would be `const` but get_lhs/rhs are not const

src/shogun/kernel/GaussianKernel.h
gf712: I see, I was completely missing the point, I was just seeing the log :D So this should be a function to convert sigma <-> gamma and the log is handled internally?

gf712: should we register `length_scale`/`width` instead of `log_width` then? And then just have a callback around `width` that updates `log_width` when width is changed? And `log_width` is hidden from the user?

src/shogun/io/serialization/JsonSerializer.cpp
```diff
@@ -173,6 +173,11 @@ class JSONWriterVisitor : public AnyVisitor
 		SG_DEBUG("writing std::string with value {}", v->c_str());
 		m_json_writer.String(v->c_str());
 	}
+	void on(AutoValueEmpty* v) override
+	{
+		SG_DEBUG("writing empty auto value");
+		m_json_writer.String("auto");
```
gf712, line 179: @karlnapf this is unlikely to happen, but a user might serialize an object which has not been fitted, so need to handle the default value of automatic parameter instantiation

karlnapf: mmmmh....I guess auto can appear in many places so it is just about justifiable? Not sure....

karlnapf: I guess yes :)

src/shogun/kernel/GaussianKernel.h
karlnapf: I'd prefer registering the bandwidth itself (the actual quantity from the formula) 
Beware that this requires some refactoring work ....

src/shogun/io/serialization/JsonSerializer.cpp
gf712: The other question is what happens with get? In python I just return the value if it has been computed or the string "auto" if not. But in C++ this is not possible, unless we return a variant.. And will have issues with the other interfaces.. I was thinking if maybe it should return NaN if the auto value has not been set, or maybe just an error?

gf712: This is what python looks like right now
```python
In [1]: import shogun as sg                                                                                                                                                                                 

In [2]: k = sg.create_kernel("GaussianKernel")                                                                                                                                                              

In [3]: k                                                                                                                                                                                                   
Out[3]: GaussianKernel(num_subscriptions={function}, cache_size=10, lhs=null, rhs=null, lhs_equals_rhs=false, num_lhs=0, num_rhs=0, combined_kernel_weight=1, optimization_initialized=false, properties=0, normalizer=IdentityKernelNormalizer(...), opt_type=FASTBUTMEMHUNGRY, m_distance=EuclideanDistance(...), m_precomputed_distance=null, log_width=auto)

In [4]: k.get("log_width")                                                                                                                                                                                  
Out[4]: 'auto'
```

karlnapf: I think I'd prefer an error, unified for all interfaces... only the to_string method should print the auto ...

karlnapf: The alternative would be to return nan, but I dont like that too much I think

gf712: > I think I'd prefer an error, unified for all interfaces... only the to_string method should print the auto ...

Cool, makes my life easier :D

> The alternative would be to return nan, but I dont like that too much I think

Yea the issue with that is that there is no standard for non floating point representations of nan. And I don't even know if all languages we support have implemented NaN according to the ISO standard

src/shogun/io/serialization/BitseryVisitor.h
```diff
@@ -74,7 +74,13 @@ namespace shogun
...
 				void on(float64_t* v) override
 				{
-					m_s.value8b(*v);
+					if (static_cast<T*>(this)->m_auto_value.has_value())
+					{
+						*v = *static_cast<T*>(this)->m_auto_value;
+						static_cast<T*>(this)->m_auto_value.reset();
+					}
+					else
+						m_s.value8b(*v);
```
gf712, line 83: @vigsterkr I really don't like this but I need a way to read an optional value.. Basically I store "auto", without initialisation, as a nullptr (like SGObject). If the value is nullptr all good (the variant is active on the empty value type), if not then I need to somehow store the value so that the visitor goes through it (to avoid repositioning the buffer). If you checkout in any.h how I implemented the visitor for std::variant to be used in auto init you'll see why I need this foobarness. I am sure there is a better way, but I can't think of it... Any ideas?

src/interfaces/swig/SGBase.i
```diff
@@ -364,7 +364,14 @@ INTERFACE_BASETYPE get(const std::string& name) const
...
 
         auto visitor = VISITOR_NAME(result);
-        find_iter->second->get_value().visit(&visitor);
+        try {
+            find_iter->second->get_value().visit(&visitor);
+        }
+        catch (const std::bad_optional_access&) {
+            error("The value of parameter {}::{} is automatically inferred during model fitting, "
+                  "and is currently not set. Either set a value with put or call get after model fitting.",
+                  $self->get_name(), name);
```
gf712: @karlnapf is this error message fine?

src/shogun/base/SGObject.h
```diff
@@ -774,6 +799,19 @@ class SGObject: public std::enable_shared_from_this<SGObject>
...
 			}
 		}
+		else if constexpr (is_auto_value_v<T>)
+		{
+			using ReturnType = traits::get_variant_type_t<0, T>;
+			Any::register_visitor<T, ParameterGetterInterface<ReturnType, T>>(
+				[](T* value, auto* visitor)
+				{
+					if (std::holds_alternative<AutoValueEmpty>(*value))
+						throw std::bad_optional_access{};
```
gf712, line 814: this exception does not have a message, but it is caught in `get` which then throw a ShogunException with meaning

src/shogun/kernel/GaussianKernel.cpp
```diff
@@ -6,37 +6,44 @@
...
+
+	auto dist=std::make_shared<EuclideanDistance>();
+	dist->set_disable_sqrt(true);
+	m_distance=dist;
+
+	SG_ADD(
+	    &m_log_width, "log_width", "Kernel width in log domain",
+	    ParameterProperties::AUTO | ParameterProperties::HYPER |
+	        ParameterProperties::GRADIENT,
+	    std::make_shared<params::GaussianWidthAutoInit>(*this, 1.0));
```
gf712, line 30: @karlnapf if for some reason it is not possible to compute the median heuristic it defaults to 1, which is the current default

src/shogun/kernel/SigmoidKernel.h
```diff
@@ -84,12 +90,12 @@ class SigmoidKernel: public DotKernel
 		 */
 		virtual float64_t compute(int32_t idx_a, int32_t idx_b)
 		{
-			return tanh(gamma*DotKernel::compute(idx_a,idx_b)+coef0);
+			return tanh(std::get<float64_t>(m_gamma)*DotKernel::compute(idx_a,idx_b)+coef0);
```
gf712, line 93: this should not cause any runtime overhead

src/interfaces/swig/Boost.i
```diff
@@ -1,35 +0,0 @@
...
- * The views and conclusions contained in the software and documentation are those
- * of the authors and should not be interpreted as representing official policies,
- * either expressed or implied, of the Shogun Development Team.
- */
-
-/* Remove C Prefix */
-%shared_ptr(shogun::StochasticGBMachine)
-
-/* Include Class Headers to make them visible from within the target language */
-%include <shogun/machine/StochasticGBMachine.h>
```
gf712, line 35: whoops, somehow this commit got mixed up. But the file is not needed anymore anyway from what i recall

src/shogun/kernel/GaussianKernel.cpp
karlnapf: 1 for bandwidth or for log_width?

karlnapf: but yeah it is good. Not sure why it wouldnt work?

src/interfaces/swig/SGBase.i
karlnapf: yes it is good. Maybe some nitpicking:

* don't mention method names (put), but just say "set a value", "read value after fitting"
* fit -> train (more shogun-ish)
* inferred -> is automatically computed using HEURISTIC_NAME ... (if the name is accessible at this stage)

src/shogun/base/SGObject.h
karlnapf: maybe add a comment with that?

src/shogun/kernel/GaussianKernel.cpp
karlnapf: see other comment I made

gf712: 1 for bandwidth, which is the current default. It might not work if the features are not dense (there is a check for it in the auto initialiser). But I also don't know if the kernel works with non dense features?

src/shogun/kernel/GaussianKernel.h
gf712: I'll can refactor that in a different PR, or start an issue

src/shogun/kernel/GaussianKernel.cpp
karlnapf: not sure, but it doesnt make sense as the euclidean distance doesnt mean much in high dimensions

gf712: I guess this is more of a software completeness thing. If the algorithm doesn't work because of the feature type then it should be handled somewhere else (by the user or the library)

karlnapf: agree

gf712:
Shall I merge this?

src/shogun/kernel/GaussianKernel.cpp
```diff
@@ -60,7 +67,15 @@ std::shared_ptr<SGObject >GaussianKernel::shallow_copy() const
 	// this assert is to avoid any subclass of GaussianKernel accidentally called
 	// with the implement here
 	ASSERT(typeid(*this) == typeid(GaussianKernel))
-	auto ker = std::make_shared<GaussianKernel>(cache_size, get_width());
+	std::shared_ptr<GaussianKernel> ker;
```
karlnapf: I think you could just remove this method (is it even used?)

src/shogun/kernel/GaussianKernel.h
```diff
@@ -167,13 +192,9 @@ class GaussianKernel: public ShiftInvariantKernel
...
 	virtual float64_t distance(int32_t idx_a, int32_t idx_b) const;
 
-private:
-	/** register parameters and initialize with defaults */
-	void register_params();
-
 protected:
 	/** width */
-	float64_t m_log_width;
+	AutoValue<float64_t> m_log_width = AutoValueEmpty{};
```
karlnapf, line 194: so all auto parameters now need to have this type?

tests/unit/base/SGObject_unittest.cc
```diff
@@ -575,3 +575,11 @@ TEST(SGObject, constrained_parameter)
     EXPECT_EQ(obj->get<int32_t>("constrained_parameter"), 1);
 }
 
+
+TEST(SGObject, auto_parameter)
+{
+    auto obj = std::make_shared<MockObject>();
+    EXPECT_THROW(obj->get<int32_t>("auto_parameter"), ShogunException);
+    obj->put("auto_parameter", 1);
```
karlnapf: class member of c++ method?

karlnapf:
This is a really cool PR! :)

src/shogun/kernel/GaussianKernel.h
gf712: Yes, and need to be initialised as empty. I mean the AutoValueEmpty can be assigned in the parameter registration, which might be safer?

src/shogun/kernel/GaussianKernel.cpp
gf712: I'll have a look, it did look very odd to me, as this is essentially done by the copy constructor..

tests/unit/base/SGObject_unittest.cc
gf712: you mean test for both? I thought the SGObject unittests were all done with put/get?

src/shogun/kernel/GaussianKernel.h
gf712: but yea in a nutshell this type will handle serde correctly and put/get, as well as check when auto initialisation is required

karlnapf: sounds good to me

tests/unit/base/SGObject_unittest.cc
karlnapf: no, what I mean is we don't want to add code with hard-coded parameter names in the tests. Instead we want to use those names from public static class members, or simply use the c++ api. So could add those names to the mock class?

src/shogun/kernel/GaussianKernel.cpp
karlnapf: I dont think we ever really used those things, they were introduced as some form of hackery in some early gsoc.

tests/unit/base/SGObject_unittest.cc
gf712: Ah I see, I'll add that

src/shogun/kernel/GaussianKernel.cpp
gf712: OK, I'll delete this function

gf712: actually there is a handful of places that has this so I'll start an issue and can be done in different PR

PR #4581, opened 2019-03-19
[WIP] implementation of array SGObject::get

karlnapf:
***NO_CI***

@geektoni I will try to finish this tomorrow. Let me know if you have thoughts

### DIFF
--- a/examples/meta/src/statistical_testing/linear_time_maximum_mean_discrepancy.sg
+++ b/examples/meta/src/statistical_testing/linear_time_maximum_mean_discrepancy.sg
@@ -42,15 +42,15 @@ mmd.set_train_test_ratio(1)
 #![select_kernel_single]
 mmd.set_kernel_selection_strategy(enum EKernelSelectionMethod.KSM_MAXIMIZE_POWER)
 mmd.select_kernel()
-Kernel learnt_kernel_single = mmd.get_kernel()
+Kernel learnt_kernel_single = as_kernel(mmd.get("kernel"))
 real width = learnt_kernel_single.get_real("log_width")
 #![select_kernel_single]
 
 #![select_kernel_combined]
 mmd.set_kernel_selection_strategy(enum EKernelSelectionMethod.KSM_MAXIMIZE_POWER, True)
 mmd.select_kernel()
-CombinedKernel learnt_kernel_combined = CombinedKernel:obtain_from_generic(mmd.get_kernel())
-RealVector weights = learnt_kernel_combined.get_subkernel_weights()
+Kernel learnt_kernel_combined = as_kernel(mmd.get("kernel"))
+RealVector weights = learnt_kernel_combined.get_real_vector("subkernel_log_weights")
 #![select_kernel_combined]
 
 #![perform_test_optimized]
--- a/examples/meta/src/statistical_testing/quadratic_time_maximum_mean_discrepancy.sg
+++ b/examples/meta/src/statistical_testing/quadratic_time_maximum_mean_discrepancy.sg
@@ -56,15 +56,15 @@ int num_runs = 1
 int num_folds = 3
 mmd.set_kernel_selection_strategy(enum EKernelSelectionMethod.KSM_CROSS_VALIDATION, num_runs, num_folds, alpha)
 mmd.select_kernel()
-Kernel learnt_kernel_single = mmd.get_kernel()
+Kernel learnt_kernel_single = as_kernel(mmd.get("kernel"))
 real width = learnt_kernel_single.get_real("log_width")
 #![select_kernel_single]
 
 #![select_kernel_combined]
 mmd.set_kernel_selection_strategy(enum EKernelSelectionMethod.KSM_MAXIMIZE_MMD, True)
 mmd.select_kernel()
-CombinedKernel learnt_kernel_combined = CombinedKernel:obtain_from_generic(mmd.get_kernel())
-RealVector weights = learnt_kernel_combined.get_subkernel_weights()
+Kernel learnt_kernel_combined = as_kernel(mmd.get("kernel"))
+RealVector weights = learnt_kernel_combined.get_real_vector("subkernel_log_weights")
 #![select_kernel_combined]
 
 #![perform_test_optimized]
--- a/src/interfaces/python/swig_typemaps.i
+++ b/src/interfaces/python/swig_typemaps.i
@@ -1415,7 +1415,7 @@ for getter in _GETTERS:
     _rename_python_function(_shogun.SGObject, getter, _private_getter)
     _internal_getter_methods.append(_shogun.SGObject.__dict__[_private_getter])
 
-def _internal_get_param(self, name):
+def _internal_get_param(self, name, *args):
     """
     Returns the value of the given parameter.
     The return type depends on the parameter,
@@ -1425,7 +1425,7 @@ def _internal_get_param(self, name):
 
     for getter in _internal_getter_methods:
         try:
-            return getter(self, name)
+            return getter(self, name, *args)
         except SystemError:
             pass
         except Exception:
--- a/src/shogun/base/SGObject.cpp
+++ b/src/shogun/base/SGObject.cpp
@@ -1092,47 +1092,36 @@ void CSGObject::init_auto_params()
 	}
 }
 
+CSGObject* CSGObject::get(const std::string& name, index_t index) const
+{
+	auto* result = sgo_details::get_by_tag(this, name, std::move(sgo_details::GetByNameIndex(index)));
+	if (!result && has(name))
+	{
+		SG_ERROR(
+			"Cannot get array parameter %s::%s[%d] of type %s as object.\n",
+			get_name(), name.c_str(), index,
+			self->map[BaseTag(name)].get_value().type().c_str());
+	}
+	return result;
+}
+
 CSGObject* CSGObject::get(const std::string& name, std::nothrow_t) const
     noexcept
 {
-	if (auto* result = get_sgobject_type_dispatcher<CDistance>(name))
-		return result;
-	if (auto* result = get_sgobject_type_dispatcher<CKernel>(name))
-		return result;
-	if (auto* result = get_sgobject_type_dispatcher<CFeatures>(name))
-		return result;
-	if (auto* result = get_sgobject_type_dispatcher<CLabels>(name))
-		return result;
-	return nullptr;
+	return sgo_details::get_by_tag(this, name, std::move(sgo_details::GetByName()));
 }
 
 CSGObject* CSGObject::get(const std::string& name) const noexcept(false)
 {
 	auto* result = get(name, std::nothrow);
-	if (result == nullptr)
+	if (!result && has(name))
 	{
-		if (self->map.find(BaseTag(name)) != self->map.end())
-		{
-			SG_ERROR(
-			    "Cannot get parameter %s::%s of type %s as object, not a "
-			    "shogun object base type.\n",
-			    get_name(), name.c_str(),
-			    self->map[BaseTag(name)].get_value().type().c_str());
-		}
-		else
-		{
-			SG_ERROR(
-			    "There is no parameter '%s' in %s.\n", name.c_str(),
-			    get_name());
-		}
+		SG_ERROR(
+			"Cannot get parameter %s::%s of type %s as object.\n",
+			get_name(), name.c_str(),
+			self->map[BaseTag(name)].get_value().type().c_str());
 	}
 	return result;
-}
-
-void CSGObject::push_back(CDynamicObjectArray* array, CSGObject* value)
-{
-	ASSERT(array);
-	array->push_back(value);
 }
 
 std::string CSGObject::string_enum_reverse_lookup(
@@ -1146,4 +1135,5 @@ std::string CSGObject::string_enum_reverse_lookup(
 		    return p.second == enum_value;
 	    });
 	return enum_map_it->first;
+
 }
--- a/src/shogun/base/SGObject.h
+++ b/src/shogun/base/SGObject.h
@@ -45,6 +45,17 @@ class CSerializableFile;
 class ParameterObserverInterface;
 class CDynamicObjectArray;
 
+#ifndef SWIG
+#ifndef DOXYGEN_SHOULD_SKIP_THIS
+namespace sgo_details
+{
+template <typename T1, typename T2>
+bool dispatch_array_type(const CSGObject* obj, const std::string& name,
+		T2&& lambda);
+}
+#endif // DOXYGEN_SHOULD_SKIP_THIS
+#endif // SWIG
+
 template <class T, class K> class CMap;
 
 struct TParameter;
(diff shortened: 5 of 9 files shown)


src/shogun/base/SGObject.h
```diff
@@ -476,6 +476,47 @@ class CSGObject
...
+			return (T*) get(array, index);
+		}
+
+		Tag<std::vector<T*>> tag_vector(name);
+		if (has(tag_vector))
+		{
+			// TODO this needs test once we have std::vector parameters
+			SG_NOTIMPLEMENTED
+			auto array = get(tag_vector);
+			return array.element_at(index);
```
geektoni: If this is an `std::vector`, then it won't have the `element_at` method, just `at` (if I understand correctly what it is happening here).

src/shogun/base/SGObject.cpp
```diff
@@ -1135,6 +1135,12 @@ void CSGObject::push_back(CDynamicObjectArray* array, CSGObject* value)
 	array->push_back(value);
 }
 
+CSGObject* CSGObject::get(CDynamicObjectArray* array, index_t index)
+{
+	ASSERT(array);
+	return array->get_element(index);
```
geektoni: Is this throwing any errors to the user if the supplied index is out-of-bound/negative?

src/shogun/base/SGObject.h
karlnapf: yep! copy paste error (didnt compile yet)

src/shogun/base/SGObject.cpp
karlnapf: yes, `CDynamicObjectArray` takes care of that. We might want to catch the expection and make a pretty printed version since this is user facing though. Will add that

karlnapf:
> Nice one! I wonder if it would make sense to also add a `get(elem, x, y)` call for `SGMatrix`, since we are doing it for vectors.

Not sure I understand. `SGMatrix` is only templated for primitives (scalar), so in that case we would just get the whole matrix. This array thing is more for dynamically growing structures (of objects, see the type trait), so this is always linear

geektoni:
> > Nice one! I wonder if it would make sense to also add a `get(elem, x, y)` call for `SGMatrix`, since we are doing it for vectors.
> 
> Not sure I understand. `SGMatrix` is only templated for primitives (scalar), so in that case we would just get the whole matrix. This array thing is more for dynamically growing structures (of objects, see the type trait), so this is always linear

I was thinking about use cases in which we may need to get a single value from an SGMatrix with `get()`. If we supply to the user `get(name, row, col)` then he could get directly a single entry from the matrix. I thought that since we are doing it also for vectors, it could be nice to have a uniform interface which covers also matrices. :) just for consistency

src/shogun/base/SGObject.cpp
```diff
@@ -1135,6 +1135,12 @@ void CSGObject::push_back(CDynamicObjectArray* array, CSGObject* value)
 	array->push_back(value);
 }
 
+CSGObject* CSGObject::get(CDynamicObjectArray* array, index_t index)
```
vigsterkr: why does this need to be part of the SGObject interface?

vigsterkr: this is similar to the CSGObject.push_back that i dont get either why it's part of SGObject

src/shogun/base/SGObject.h
```diff
@@ -969,6 +1010,7 @@ class CSGObject
 
 private:
 	void push_back(CDynamicObjectArray* array, CSGObject* value);
```
vigsterkr: why in SGObject?

src/shogun/base/SGObject.h
```diff
@@ -476,6 +476,47 @@ class CSGObject
...
 
+	/** Typed array getter for an object class parameter of a Shogun base class
+	* type, identified by a name and an index
+	*
+	* @param name name of the parameter array
+	* @param index index of the element in the array
+	*/
+	template <class T,
+			  class X = typename std::enable_if<is_sg_base<T>::value>::type>
+	T* get(const std::string& name, index_t index)
```
vigsterkr: this code is an almost copy-paste of the
```
	template <class T,
	void add(const std::string& name, T* value)
```

would be nice to unify.... and note: std::list, std::deque....

karlnapf:
> > > Nice one! I wonder if it would make sense to also add a `get(elem, x, y)` call for `SGMatrix`, since we are doing it for vectors.
> > 
> > 
> > Not sure I understand. `SGMatrix` is only templated for primitives (scalar), so in that case we would just get the whole matrix. This array thing is more for dynamically growing structures (of objects, see the type trait), so this is always linear
> 
> I was thinking about use cases in which we may need to get a single value from an SGMatrix with `get()`. If we supply to the user `get(name, row, col)` then he could get directly a single entry from the matrix. I thought that since we are doing it also for vectors, it could be nice to have a uniform interface which covers also matrices. :) just for consistency

it is not for SGVector, just for `std::vector<CSGObject>` things..... `get` will not work to retreive from SGVector of primitives, just for base classes

src/shogun/base/SGObject.h
karlnapf: yep agree, would be cool to do that. This was mostly to get something working for Giovanni, but I can look into doing this more general as well

src/shogun/base/SGObject.cpp
karlnapf: it can live anywhere. Just wanted to avoid including the DynamicObjectArray header file. Suggestions ?

vigsterkr: why SGObject and not DynamicObjectArray...
this could be a static method of DynamicObjectArray or just a namespace function....

vigsterkr: in latter case you dont need to include the header file...

karlnapf: cool!

geektoni:
Ahh I see now. I thought it could be also applied to something more general than Shogun classes (I've misread the type trait apparently). Nevermind then.

src/shogun/base/SGObject.h
```diff
@@ -443,39 +443,44 @@ class CSGObject
...
+	{
+		CSGObject* result = nullptr;
+
+		auto get_lambda = [&index, &result](auto& array) {
+			result = array.at(index);
+		};
+		if (execute_dispatch_array_type<T>(this, name, get_lambda))
+			// have to type check as CDynamicObjectArray might return any base type
+			if dynamic_cast<T>(result)
```
vigsterkr: .as?

src/shogun/base/SGObject.h
```diff
@@ -443,39 +446,44 @@ class CSGObject
...
+	T* get(const std::string& name, index_t index)
+	{
+		CSGObject* result = nullptr;
+
+		auto get_lambda = [&index, &result](auto& array) {
+			result = array.at(index);
+		};
+		if (execute_dispatch_array_type<T>(this, name, get_lambda))
+			// have to type check as CDynamicObjectArray might return any base type
+			if (dynamic_cast<T>(result))
```
vigsterkr: and actually better that dyncast... use typeid typeindex combo... :)

src/shogun/base/SGObject.h
```diff
@@ -1011,5 +1018,54 @@ class CSGObject
...
+	{
+		T* result = obj->get<T*>(name);
+		SG_REF(result)
+		return (CSGObject*)result;
+	}
+
+	return nullptr;
+}
+
+template <typename T>
```
karlnapf: @gf712 I would like to unify those two functions, to accept a get lambda that has the index parameter bound. However, the `get` call inside that lambda is templated. Not sure that is possible?

src/shogun/base/SGObject.cpp
```diff
@@ -1092,16 +1092,24 @@ void CSGObject::init_auto_params()
 	}
 }
 
+CSGObject* CSGObject::get(const std::string& name, index_t index) const
+{
+	if (auto* result = get_sgobject_type_dispatcher<CKernel>(this, name, index))
```
karlnapf: The reason I'd like to unify is that I want to use the dispatching function here and in the standard get, with a different lambda that binds the index.

karlnapf: Next question would be to avoid listing the base types twice

src/shogun/base/SGObject.h
karlnapf: works now

src/shogun/base/SGObject.h
```diff
@@ -1011,5 +1021,64 @@ class CSGObject
...
+CSGObject* get(const CSGObject* obj, const std::string& name, ByNameIndex i)
+{
+	return obj->get<T>(name, i.m_index);
+}
+
+template<typename T>
+CSGObject* get_dispatch_all_base_types(const CSGObject* obj, const std::string& name,
+		T&& how)
+{
+	if (auto* result = get<CKernel>(obj, name, how))
```
karlnapf: this is a single central list of trying different get versions for all base types

src/shogun/base/SGObject.h
```diff
@@ -1011,5 +1021,64 @@ class CSGObject
 	/** Subscriber used to call onNext, onComplete etc.*/
 	SGSubscriber* m_subscriber_params;
 };
+
+#ifndef SWIG
```
vigsterkr: could we move this whole thing into a separate header... as sgboject.h is already 1k+ LoC

src/shogun/base/SGObject.h
```diff
@@ -1011,5 +1021,64 @@ class CSGObject
...
+};
+
+template <typename T>
+CSGObject* get(const CSGObject* obj, const std::string& name, ByName)
+{
+	return obj->get<T*>(name);
+}
+
+template <typename T>
+CSGObject* get(const CSGObject* obj, const std::string& name, ByNameIndex i)
```
vigsterkr: detail namespace or less general name

lisitsyn: I guess `GetByName`, `GetByNameAndIndex`.

This can't be detail namespace as it is directly exposed.

vigsterkr: i mean it can be still called detail::get? :)

src/shogun/base/SGObject.h
karlnapf: Yes good point

src/shogun/base/SGObject.h
karlnapf: Yes was thinking that as well. Next step

karlnapf: I have put it all in `sgobject_get_details` now

src/shogun/CMakeLists.txt
```diff
@@ -12,6 +12,16 @@ include(CheckCXXSourceCompiles)
 include(CMakePackageConfigHelpers)
 
 FILE(GLOB_RECURSE LIBSHOGUN_SRC *.${EXT_SRC_CPP} *.${EXT_SRC_C})
+FOREACH(src ${LIBSHOGUN_SRC})
```
karlnapf: @vigsterkr that is a local hack I am using to enforce compiling `SGObject.cpp` happens as first CXX. Very useful to get compiler errors quickly when playing with `SGObject.h`, as otherwise I always need to wait for a bunch of other classes to be compiled first. Not sure something like this should be in, but I put it here for discussion anyways.

vigsterkr: @karlnapf if you would use ninja then you could call the specific object build.... this is a hack that nobody else needs and your problem could be solved by using ninja.

src/shogun/base/SGObject.h
```diff
@@ -45,6 +47,15 @@ class CSerializableFile;
 class ParameterObserverInterface;
 class CDynamicObjectArray;
 
+#ifndef SWIG
+namespace sgobject_get_details
```
vigsterkr: this namespace name is way too ugly...

src/shogun/base/SGObject.h
```diff
@@ -1011,5 +1024,69 @@ class CSGObject
 	/** Subscriber used to call onNext, onComplete etc.*/
 	SGSubscriber* m_subscriber_params;
 };
+
+#ifndef SWIG
+
+namespace sgobject_get_details
```
vigsterkr: other header file? sgobject_utils.h or anything that is not making this file bigger

karlnapf: actually realised this might be tricky as I am calling sgobject methods so need the type, but also need to include the header files.

examples/meta/src/statistical_testing/quadratic_time_maximum_mean_discrepancy.sg
```diff
@@ -56,7 +56,7 @@ int num_runs = 1
 int num_folds = 3
 mmd.set_kernel_selection_strategy(enum EKernelSelectionMethod.KSM_CROSS_VALIDATION, num_runs, num_folds, alpha)
 mmd.select_kernel()
-Kernel learnt_kernel_single = mmd.get_kernel()
```
karlnapf: @lambday here is an example of why the testing framework needs refactoring. I keep on having to do such hacks

karlnapf:
anyone has ideas how to make the compilers in the CI happy? Looks like I broke the C++14 standard?

src/interfaces/python/swig_typemaps.i
```diff
@@ -1415,7 +1415,7 @@ for getter in _GETTERS:
     _rename_python_function(_shogun.SGObject, getter, _private_getter)
     _internal_getter_methods.append(_shogun.SGObject.__dict__[_private_getter])
 
-def _internal_get_param(self, name):
+def _internal_get_param(self, name, *args):
```
karlnapf: @gf712 I added this to account for the index arguments. You agree with it?

examples/meta/src/statistical_testing/linear_time_maximum_mean_discrepancy.sg
```diff
@@ -42,15 +42,15 @@ mmd.set_train_test_ratio(1)
...
+Kernel learnt_kernel_single = mmd.get("kernel")
 real width = learnt_kernel_single.get_real("log_width")
 #![select_kernel_single]
 
 #![select_kernel_combined]
 mmd.set_kernel_selection_strategy(enum EKernelSelectionMethod.KSM_MAXIMIZE_POWER, True)
 mmd.select_kernel()
-CombinedKernel learnt_kernel_combined = CombinedKernel:obtain_from_generic(mmd.get_kernel())
-RealVector weights = learnt_kernel_combined.get_subkernel_weights()
+Kernel learnt_kernel_combined = mmd.get("kernel")
```
karlnapf: @remyabel

remyabel:
> auto parameters are not part of c++14 but Concepts TS ... interesting

`auto` parameters in C++14 is only for generic lambdas. `auto` parameters for regular functions is not allowed (but is a GCC extension.)

examples/meta/src/statistical_testing/linear_time_maximum_mean_discrepancy.sg
```diff
@@ -42,15 +42,15 @@ mmd.set_train_test_ratio(1)
...
+RealVector weights = learnt_kernel_combined.get_real_vector("subkernel_log_weights")
```
karlnapf: @lambday I had to change the getter to the log weight field of `CCombinedKernel` as the method is non-const and therefore not exposable to the new `get` at the moment. Will this work to get the weights your framework is setting?

src/interfaces/python/swig_typemaps.i
gf712: seems fine to me!

karlnapf:
> > auto parameters are not part of c++14 but Concepts TS ... interesting
> 
> `auto` parameters in C++14 is only for generic lambdas. `auto` parameters for regular functions is not allowed (but is a GCC extension.)

yep I realised...thx

src/shogun/base/SGObject.cpp
```diff
@@ -1092,26 +1092,23 @@ void CSGObject::init_auto_params()
...
+	return sgo_details::get_by_tag(this, name, std::move(sgo_details::GetByName()));
 }
 
 CSGObject* CSGObject::get(const std::string& name) const noexcept(false)
 {
 	auto* result = get(name, std::nothrow);
 	if (result == nullptr)
 	{
-		if (self->map.find(BaseTag(name)) != self->map.end())
+		if (has(name))
```
karlnapf: @gf712 btw I changed this as I think it is cleaner to read. Maybe we want to update the implementation of that at some point ?

src/shogun/base/SGObject.cpp
```diff
@@ -1092,26 +1092,23 @@ void CSGObject::init_auto_params()
...
-		return result;
-	if (auto* result = get_sgobject_type_dispatcher<CFeatures>(name))
-		return result;
-	if (auto* result = get_sgobject_type_dispatcher<CLabels>(name))
-		return result;
-	return nullptr;
+	return sgo_details::get_by_tag(this, name, std::move(sgo_details::GetByName()));
 }
 
 CSGObject* CSGObject::get(const std::string& name) const noexcept(false)
```
karlnapf: oh and btw do you still need the non-exception version of this for the openml? I remember I asked you but I forgot what you said?

karlnapf:
I think this works now. Let me know any feedback, otherwise Ill add some docs and merge
(postponing the problem of splitting SGObject.h code into multiple files)

src/shogun/base/SGObject.cpp
gf712: I am not sure which implementation you mean?

src/shogun/base/SGObject.cpp
gf712: Yes, I think that is still needed because there are issues catching errors in swig code

src/shogun/base/SGObject.cpp
karlnapf: the one of `has`

src/shogun/base/SGObject.cpp
karlnapf: ok cool Ill keep it then

karlnapf:
Finally !

src/shogun/base/SGObject.h
```diff
@@ -1011,5 +1029,85 @@ class CSGObject
...
+	}
+	catch (const std::exception&) {}
+	return result;
+}
+
+template<typename T>
+CSGObject* get_dispatch_all_base_types(const CSGObject* obj, const std::string& name,
+		T&& how)
+{
+	if (auto* result = get_if_possible<CKernel>(obj, name, how))
```
karlnapf: @gf712 I would love to centralise this list via using some structure of base types we have in shogun. Was there something in place for that already?

gf712: I was trying to get them all in this [file](https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/lib/sg_types.h).

karlnapf: ok thanks will add that later, but the PR is too big already anyways :)

gf712: @karlnapf Why did you not add CDistance and CLabels?

karlnapf: Forgot! Thanks. Will send a fix soon.

karlnapf: actually it would be good to have a dispatcher for base types, just like we have for `Any`.
@geektoni added `CMachine` in another PR recently, and I am sure there will be dispatching happening at other places in the future as well.
I mean for now this is the centralised place which is re-used at least.
Maybe we can write an entrance task, and label it for the detox project?

Issue #4561, opened 2019-03-07
fix liblinearregression tests


vigsterkr:
the unit test of LibLinearRegression fails every now and then, that causes false alarms in our CI

avramidis:
Can we have artifacts if it's possible to see why it fails? I cannot make it fail locally.

gf712:
the weights learnt go off quite a bit sometimes. It seems to be an issue with something related to RNG. I tested for UB, but couldn't find anything. 

It's also just an issue with clang, which makes it even weirder. If you have clang compile shogun with it and run the test a few times and it will very likely fail occasionally.. 

I think the algorithm expects it to be monotonically decreasing, but sometimes that doesn't happen and it stops training too early, other times it will reach maximum iterations and be stuck in a local minimum with a greater loss... It could also be an issue with the data which I am not sure is always the same when it is generated...

avramidis:
Thank you! I will try with clang.

PR #4582, opened 2019-03-19
Fix liblinear regression bug

theartful:
This PR addresses the issue #4561. The problem was an uninitialized variable, so it created unpredictable behaviour.

### DIFF
--- a/src/shogun/regression/svr/LibLinearRegression.cpp
+++ b/src/shogun/regression/svr/LibLinearRegression.cpp
@@ -38,6 +38,7 @@ void CLibLinearRegression::init_defaults()
 {
 	set_C(1.0);
 	set_epsilon(1e-2);
+	set_tube_epsilon(0);
 	set_max_iter(10000);
 	set_use_bias(false);
 	set_liblinear_regression_type(L2R_L1LOSS_SVR_DUAL);
@@ -103,7 +104,6 @@ bool CLibLinearRegression::train_machine(CFeatures* data)
 	}
 	prob.l=num_vec;
 	prob.x=features;
-	prob.y=SG_MALLOC(float64_t, prob.l);
 	auto labels = regression_labels(m_labels);
 	prob.y = labels->get_labels();
 


src/shogun/regression/svr/LibLinearRegression.cpp
```diff
@@ -38,6 +38,7 @@ void CLibLinearRegression::init_defaults()
 {
 	set_C(1.0);
 	set_epsilon(1e-2);
+	set_tube_epsilon(0);
```
theartful: sklearn uses 0.1, but the test won't pass using this value (error > 1e-5)

src/shogun/regression/svr/LibLinearRegression.cpp
```diff
@@ -103,7 +104,6 @@ bool CLibLinearRegression::train_machine(CFeatures* data)
 	}
 	prob.l=num_vec;
 	prob.x=features;
-	prob.y=SG_MALLOC(float64_t, prob.l);
```
theartful: memory leak?

src/shogun/regression/svr/LibLinearRegression.cpp
gf712: Ah nice! If you want you can adjust the parameters around a bit? Thanks anyway!

src/shogun/regression/svr/LibLinearRegression.cpp
karlnapf: think so

karlnapf:
Thanks a lot!

theartful:
https://github.com/shogun-toolbox/shogun/blob/9adbb15ae0bd1bc6d3a6cec0cc8edd8021e28d41/src/shogun/regression/svr/LibLinearRegression.cpp#L47-L60
@karlnapf sorry for any inconvenience. I think there might be something wrong here. `m_epsilon` is not the tube epsilon. It's used in the TRON optimization algo (tolerance), which optimizes the tubed L1/L2 error.

karlnapf:
Not sure how it can about. Wanna send a correction?

PR #4584, opened 2019-03-21
Fix parameter registration in liblinear regression

theartful:
Sklearn and liblinear use `m_tube_epsilon=0.1` as default, so I used this as the default value


theartful:
0.1 is also a reasonable value for TRON's epsilon, so nothing to change there. Should I change  `init_defaults` to initialize `m_tube_epsilon` to 0.1 instead of 0, and change the tests (since they fail as they don't achieve the required error)?

karlnapf:
Thanks!

PR #4585, opened 2019-03-22
add treenode struct members to the tag framework

vigsterkr:
another way


vigsterkr:
@lisitsyn instead of massive CSGObject inheritance.... i know that it's a bit ugly to expose watch_param to public in CTreeMachineNode. any better ideas (other than inheritance?)

karlnapf:
++

lisitsyn:
@vigsterkr Looks good! I don't have any concrete suggestion how to not make it public though.

PR #4586, opened 2019-03-22
Create unit tests for utils::safe_convert function

### LINKED ISSUE #4580: create unit tests for utils::safe_convert function
create all the possible combination for primitive type conversion to check that utils::safe_convert is acting as it should

remyabel:
Fixes #4580

### DIFF
--- a/tests/unit/utils/safe_convert_unittest.cc
+++ b/tests/unit/utils/safe_convert_unittest.cc
@@ -1,0 +1,99 @@
+/*
+ * This software is distributed under BSD 3-clause license (see LICENSE file).
+ *
+ * Authors: Tommy Nguyen
+ *
+ */
+#include "sg_gtest_utilities.h"
+
+#include <cmath>
+#include <gtest/gtest.h>
+#include <shogun/util/converters.h>
+#include <vector>
+
+using namespace shogun;
+
+template <typename T>
+class SafeConvertRealTypesTest : public ::testing::Test
+{
+};
+
+template <typename T>
+class SafeConvertNonComplexTest : public ::testing::Test
+{
+};
+
+template <typename T>
+class SafeConvertNonfiniteTest : public ::testing::Test
+{
+};
+
+SG_TYPED_TEST_CASE(SafeConvertNonComplexTest, sg_non_complex_types);
+SG_TYPED_TEST_CASE(SafeConvertRealTypesTest, sg_real_types);
+SG_TYPED_TEST_CASE(SafeConvertNonfiniteTest, sg_real_types);
+
+TYPED_TEST(SafeConvertNonfiniteTest, identity_convert)
+{
+	const TypeParam iNAN = NAN;
+	const TypeParam iINF = INFINITY;
+	const TypeParam iMinusINF = -INFINITY;
+
+	EXPECT_TRUE(std::isnan(static_cast<TypeParam>(utils::safe_convert<TypeParam>(iNAN))));
+	EXPECT_TRUE(std::isinf(static_cast<TypeParam>(utils::safe_convert<TypeParam>(iINF))));
+	EXPECT_TRUE(std::isinf(static_cast<TypeParam>(utils::safe_convert<TypeParam>(iMinusINF))));
+}
+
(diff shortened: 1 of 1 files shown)


vigsterkr:
@remyabel good start, but couple of things:
1. it would be nice to test non finite values as well, like NaN, -INF, INF
2. please use gtest's typed tests https://github.com/google/googletest/blob/master/googletest/docs/advanced.md#type-parameterized-tests 
there are couple of examples how to use this among the unit tests.

and why the msvc build failed i dont know...

remyabel:
> it would be nice to test non finite values as well, like NaN, -INF, INF

Should it pass or fail? I did test `INFINITY` and it did not throw, because the test condition is `std::isfinite(x) &&`.

vigsterkr:
it should return NAN, INF, -INF in the given type... so you'll have to use isnan, isinf etc function to test the correctness.... so for example double -> float

remyabel:
Actually I don't think `safe_convert` works with float/double, since `std::make_signed/unsigned` requires T be integral.

vigsterkr:
It works as in another branch it does the convert finite double to float... nonfinite should work but a test would be nice

remyabel:
Yeah then I'm missing something obvious. This code:

```c++
   EXPECT_NO_THROW(utils::safe_convert<TypeParam>(42.0));
   EXPECT_EQ(utils::safe_convert<TypeParam>(42.0), 42.0);
```

results in this error:

```
/usr/include/c++/8/type_traits: In instantiation of ‘struct std::make_unsigned<float>’:
/home/tom/second_drive/shogun/src/shogun/util/converters.h:38:16:   required from ‘typename std::enable_if<(std::is_signed<_Tp>::value && std::is_unsigned<J>::value), I>::type shogun::utils::safe_convert(J) [with I = float; J = unsigned int; typename std::enable_if<(std::is_signed<_Tp>::value && std::is_unsigned<J>::value), I>::type = float]’
/home/tom/second_drive/shogun/tests/unit/utils/safe_convert_unittest.cc:37:2:   required from ‘void SafeConvertAllTypesTest_finite_conversions_Test<gtest_TypeParam_>::TestBody() [with gtest_TypeParam_ = float]’
/home/tom/second_drive/shogun/tests/unit/utils/safe_convert_unittest.cc:32:1:   required from here
/usr/include/c++/8/type_traits:1636:62: error: invalid use of incomplete type ‘class std::__make_unsigned_selector<float, false, false>’
     { typedef typename __make_unsigned_selector<_Tp>::__type type; };
                                                              ^~~~
/usr/include/c++/8/type_traits:1593:11: note: declaration of ‘class std::__make_unsigned_selector<float, false, false>’
     class __make_unsigned_selector;
```

`std::make_unsigned<float>` is undefined behavior. Not sure what you're doing differently.

remyabel:
OK I see the problem now. The problem is `safe_convert<float>(32u)`. `std::is_signed<float>` returns true. So we should do a `static_assert` inside that SFINAE overload to make disallow float, et al. so it doesn't get passed to `std::make_unsigned`.

vigsterkr:
ok here are the tests that i had in mind:
https://repl.it/repls/CompatibleLightcoralMicroscope

and i see what you mean by having problem converting unsigned value to float, although i dont know if that use-case ever makes any sense, but sure the traits can be fixed to support that as well

vigsterkr:
@remyabel pushed some changes into develop that fixes the unsigned -> floating point conversion

remyabel:
Thanks. I attempted to implement your suggestions but left the last test alone. I thought about adding some machinery for getting the next higher type (int32_t -> int64_t) for the out of range tests, but there is probably a simpler approach.

remyabel:
FYI, the MSVC build is failing because of:

```
error C2668: 'fpclassify': ambiguous call to overloaded function
```

I cannot reproduce this in [godbolt](https://gcc.godbolt.org/z/ckDtm1) (which uses v19.16). Hopefully a `static_cast` fixes the issue.

karlnapf:
closed?

daniyalyasin93:
Is this closed? @vigsterkr

karlnapf:
think the issue is still open... simply grep for the calls and see if you can find unit tests

PR #4588, opened 2019-03-24
Set `BUILD_META_EXAMPLES` OFF as a default.

novate:
If set `BUILD_META_EXAMPLES` `ON` as a default, first-time builders could get error in `cmake` if they don't have `python-ply` and `ctags`.
set that OFF initially, it could be **smoothier** for first-time builders to get **error free** when `cmake`. And, just as `INSTALL.md` said, if you want to build meta examples, you can manually set it on by `-DBUILD_META_EXAMPLES=ON`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -288,7 +288,7 @@ STRING(REGEX REPLACE "[0-9]*.[0-9]*.([0-9]*)" "\\1" SHOGUN_VERSION_PATCH "${VERS
 
 ################# EXAMPLES ##################
 OPTION(BUILD_EXAMPLES "Build Examples" ON)
-OPTION(BUILD_META_EXAMPLES "Generate API examples from meta-examples" ON)
+OPTION(BUILD_META_EXAMPLES "Generate API examples from meta-examples" OFF)
 # note the examples dir is added below after tests have been defined
 
 ################# DATATYPES #################


karlnapf:
Thanks for the suggestion. But I don't see the point in doing that. If someone has trouble interpreting error messages such as "install python-ply", then they better not compile the project from scratch but install using `conda`.

novate:
Thank you for your reply. Install python-ply or ctags may not be a problem. My point is that, in INSTALL.md, it’s initially said that “The standard GNU/Linux tools and Python are minimal requirements to compile Shogun.” the reminder of install ply/ctags is at the middle of the chapter.
Actually we can make BUILD_META_EXAMPLES initially be OFF, because maybe not everyone need meta examples I guess?

karlnapf:
the examples are useful for people wanting to run/build some own code, that is why they are on. The error message is very clear and it is easy to turn these examples off

novate:
I agree with you that meta examples are useful, and if build meta examples are ON as default, I suggest that we add a reminder in `install.md` at the beginning of the ‘build from source’ illustration, to remind new builders install ply and ctags at the very beginning.

lisitsyn:
@karlnapf I think @novate is right. It is worth to consider enabling it _if possible_ but it is best to have the build work without any additional settings.

novate:
I don't know why such a minor change could fail the linux checks. Guess from the logs, `ctags` is needed in the unit test. So if I would like to make this change, we should also not build the related unit test?

Maybe modifying `1INSTALL.md` as I said could be easier?

karlnapf:
the ci build flags would need to be changed (so that the examples are built). This is done in the azure config file
Same for buildbot (private repo)

I think it might be tricky to enable it dependencies are met Sergey, and this will just cause more confusion imo. We can disable them. But again I don’t really see the point because there are very clear error messages (improvement welcome!) and if people don’t read those or the readme they won’t build/hack shogun anyways

novate:
I agree, every hacker need meta examples, so let's keep it.
BTW, we can add a reminder at Compiling manually - Requirements in `INSTALL.md` to remind people install `python-ply` and `ctags` before the initial compile.

karlnapf:
sure thing, as said, PR welcome :)

novate:
Solved in #4616

PR #4590, opened 2019-03-25
Create ObservedValue from AnyParameter and extend put()

geektoni:
* Add method to generate an `ObservedValue` from an `AnyParameter`;
* Extend `put()` to emit also an ObservedValue;


src/shogun/base/SGObject.h
```diff
@@ -373,6 +373,7 @@ class CSGObject
 			}
 			ref_value(value);
 			update_parameter(_tag, make_any(value));
+			observe(ObservedValue::make_observation<T>(1, _tag.name(), get_parameter(_tag)));
```
geektoni: This line here has a little problem. Since `ObservedValue` is a forward declaration, I cannot access its methods/constructor. However, I cannot `#include ObservedValue.h` here since it causes a bit of a mess with SGObject itself (it is a cyclic dependency problem). Moreover, since this method is templated I can't place it into a `.cpp` file. Therefore, this line as it is does not compile. :(

Any suggestions about how to fix it? @karlnapf @vigsterkr @lisitsyn

karlnapf: Move into a namespace method that accepts the SGObject as input

geektoni: ehm I actually cannot see exactly the solution you're proposing. Could you elaborate a bit more?

karlnapf: You would not call the ctor/method here (so you need the definitions) but rather a method that just returns what you want (so forward declaration is ok) and define this method in another file to avoid the cyclic dependency.

geektoni: I've tried a bit yesterday evening but I think it is not the proper solution to this problem. Since `make_observation` is templated, even if I create another file for it, it is not possible to split the declaration and definition. Moreover, even if I were to create a method like this (which indeed compiles since we use only pointers):
```c++
ObservedValue * make_observation(...)
{
    return new ObservedValueTemplated<T>(...);
}
```
I would still need to include `<shogun/lib/parameter_observer/ParameterObserverTemplated.h>` in each cpp file which uses SGObject (basically 99% of Shogun's files), since otherwise the compiler will not be able to resolve the forward declaration of `ObservedValueTemplated`.

geektoni:
@karlnapf I did some other experiments, but I am afraid that with the current design of the `ObservedValue` it is not possible to extend `put()` such to emit a value each time it is used.

src/shogun/base/SGObject.h
karlnapf: I think you might be able to define a templated function in SGObject.h, and then put the implementation into the SGObject.cpp file and use `SG_FORCE_INLINED`. Then you can include the observed value header in SGObject.cpp while forward declaring it in SGObject.h

karlnapf: Just tried this:

SGObject.cpp
```
namespace shogun
{
template <class T>
SG_FORCED_INLINE void test()
{
	T* a=nullptr;
}
}
```

SGObjecth
```
template <class T>
void test();
```

and then call the function e.g. inside `CSGObject::get<T>` which compiles and links fine.

geektoni: I've just tried it in every possible manner to make it works, but, in the end, we will have always linker errors (especially when one tries to use the library). That's happens even if the method is represented as inline.

src/shogun/base/SGObject.h
```diff
@@ -1022,5 +997,187 @@ class CSGObject
...
 	SGSubscriber* m_subscriber_params;
 };
+
+	template <class T>
+	class ObservedValueTemplated;
+
+	/**
+	 * Observed value which is emitted by algorithms.
+	 */
+	class ObservedValue : public CSGObject
```
geektoni: @karlnapf as I mentioned before, I think the best way to solve the cyclic dependency problem is to add the ObservedValue class here in SGObject.h. By just adding 100~ lines more we can easily use the observable utilities (and we may be able to get rid of all those `make_observation` methods). @vigsterkr any opinion on this?

src/shogun/base/SGObject.h
```diff
@@ -1022,5 +1010,158 @@ class CSGObject
...
+		 * Constructor
+		 * @param step step
+		 * @param name the observed value's name
+		 * @param value the observed value
+		 */
+		ObservedValueTemplated(int64_t step, std::string name, std::string description, T value)
+				: ObservedValue(step, name), m_observed_value(value)
+		{
+			this->watch_param(
+					name, &m_observed_value,
```
geektoni: The name of the ObservedValue now matches the tag's one. The new behaviour is:
```c++
ObservedValue observation;

// Instead of using get<T>("value") we do
observation->get<SGVector<float64_t>>("w")
```

src/shogun/base/SGObject.h
```diff
@@ -966,6 +934,26 @@ class CSGObject
...
+	/**
+	 * Observe a parameter value given some information
+	 * @tparam T value of the parameter
+	 * @param step step
+	 * @param name name of the observed value
+	 * @param description description
+	 * @param value observed value
+	 */
+	template <class T>
+	void observe(int64_t step, std::string name, std::string description, T value);
```
geektoni: A more compact method to observe a value.

src/shogun/base/SGObject.h
```diff
@@ -966,6 +934,26 @@ class CSGObject
...
+
+	/**
+	 * Observe a registered tag.
+	 * @tparam T type of the tag
+	 * @param step step
+	 * @param name tag's name
+	 */
+	template <class T>
+	void observe_tag(int64_t step, std::string name);
```
geektoni: This can be used to observe registered tags instead. I used a different name so it is clear what it is doing.

src/shogun/base/SGObject.h
```diff
@@ -348,39 +348,7 @@ class CSGObject
...
-			update_parameter(_tag, make_any(value));
-		}
-		else
-		{
-			SG_ERROR(
-				"Parameter %s::%s does not exist.\n", get_name(),
-				_tag.name().c_str());
-		}
-	}
+	void put(const Tag<T>& _tag, const T& value, int64_t step=-1) noexcept(false);
```
geektoni: Since `ObservedValue` needs also a `step` parameter, I've extended `put` such to accept it.

src/shogun/base/SGObject.h
```diff
@@ -1022,5 +1010,158 @@ class CSGObject
...
+				SG_ERROR(
+						"Cannot put parameter %s::%s of type %s, incompatible "
+								"provided type %s.\n",
+						get_name(), _tag.name().c_str(), exc.actual().c_str(),
+						exc.expected().c_str());
+			}
+			ref_value(value);
+			update_parameter(_tag, make_any(value));
+
+			observe_tag<T>(step, _tag.name());
```
geektoni: :dancer: :dancer:

geektoni:
@karlnapf I've updated the PR with the last changes.  Let me know what do think about them. It should be good to go.

Once it is merged, I will update also #4592 and #4557

src/shogun/base/SGObject.h
karlnapf: that is nice!

src/shogun/base/SGObject.h
karlnapf: I think the name can be overloaded, i like doxygen docs for this and the interface is typed so all good...names for indicating types/behaviour is more python style imo ;)

src/shogun/base/SGObject.h
karlnapf: no, let's not do that please. `put` is a separate API of the base class, so it shouldnt be influenced by this under-the-hood code

src/shogun/base/SGObject.h
karlnapf: :rocket:  :)

src/shogun/base/SGObject.cpp
```diff
@@ -21,7 +21,7 @@
 #include <shogun/lib/SGMatrix.h>
 #include <shogun/lib/SGStringList.h>
 #include <shogun/lib/SGVector.h>
-#include <shogun/lib/parameter_observers/ParameterObserverInterface.h>
+#include <shogun/lib/observers/ParameterObserverInterface.h>
```
karlnapf: ++

src/shogun/base/SGObject.h
karlnapf: I think not having the value as argument kind of makes clear that something under the hood will be observed

src/shogun/base/SGObject.h
```diff
@@ -1022,5 +1015,161 @@ class CSGObject
...
+			SG_ERROR(
+				"Cannot put parameter %s::%s of type %s, incompatible "
+				"provided type %s.\n",
+				get_name(), _tag.name().c_str(), exc.actual().c_str(),
+				exc.expected().c_str());
+		}
+		ref_value(value);
+		update_parameter(_tag, make_any(value));
+
+		observe_tag<T>(step, _tag.name());
```
karlnapf: need to find another solution with the `step`, it cannot be part of the put API

src/shogun/base/SGObject.cpp
```diff
@@ -1119,4 +1119,12 @@ std::string CSGObject::string_enum_reverse_lookup(
...
 	return enum_map_it->first;
-}
\ No newline at end of file
+}
+
+ObservedValue::ObservedValue(int64_t step, std::string name)
+    : CSGObject(), m_step(step), m_name(name), m_any_value(Any())
+{
+	SG_ADD(&m_step, "step", "Step");
+	this->watch_param(
```
karlnapf: do we need to store this? Is this ever used with a put/get, by user or developer? If no, remove

src/shogun/base/SGObject.h
```diff
@@ -1022,5 +1015,161 @@ class CSGObject
...
+	 * @param name the observed value's name
+	 * @param value the observed value
+	 */
+	ObservedValueTemplated(
+		int64_t step, std::string name, std::string description, T value)
+		: ObservedValue(step, name), m_observed_value(value)
+	{
+		this->watch_param(
+			name, &m_observed_value,
+			AnyParameterProperties(description, ParameterProperties::NONE));
```
karlnapf: pls add the read-only property here so that users cannot change it

src/shogun/base/SGObject.h
karlnapf: Maybe we do this
```
int step = -1;
Tag<int> tag("current_iteration") // basically refers to iterative machine which is what we wanna use for algorithms with iteration counters
if (has(tag))
  step = get(tag);
```
A hacky solution. This could be more elaborate, for example machines with iteration counters could register methods/lambdas with the base class upon construction to share the iteration.
@gf712 @lisitsyn thoughts?

geektoni: What about adding a singleton counter class? The only thing it will do is provide an incremental number every time it is called. A sample usage could be like:
```c++
int64_t step = IncrementalCounter::get_instance().next_step();
observe_tag<T>(step, _tag.name())
```
Therefore, all the emitted values will be somehow syncronized.

src/shogun/base/SGObject.cpp
geektoni: The name is needed because the observers use it to filter the `ObservedValue` they receive such to keep only the ones they are looking for (e.g I'm looking for "w", therefore I will process only those `ObservedValue` which have it as a name ).

src/shogun/base/SGObject.h
geektoni: Update to observe().

src/shogun/base/SGObject.h
karlnapf: thread safety would be a concern with singletons. The auto counter raises issues when more than one thing are observed. 

The problem with my above idea is runtime costs. We could later do some benchmarking how much overhead the `has`/`get` solution above causes. If it is significant, I would suggest to add a step protected member variable in `CMachine`. But let's do that in another PR so we can move on here, just do the `get` thingi.

```
SG_FORCED_INLINE int get_step()
{
int step = -1;
Tag<int> tag("current_iteration") // could be changed later to local field
if (has(tag))
  step = get(tag);
return step;
}
```
and then use that, which gives us options to overload/virtual it later on, or use a local variable

geektoni: Updated with `get_step()`.

karlnapf:
Nice! Curious did you try it with iterative machine? Perceptron implements it

geektoni:
@karlnapf Ideally this one needs to be merged before the others. I guess that this is complete now.

geektoni:
I think this is ready to be merged. All CIs passed.

vigsterkr:
could the target branch of this be something like feature/observables ?

karlnapf:
Could you make get_step protected? Then also no need to hide it from swig. Otherwise we can merge

geektoni:
> Could you make get_step protected? Then also no need to hide it from swig. Otherwise we can merge

Sure!

karlnapf:
And I guess discuss with Viktor about develop vs feature branch. I’m good with both

src/shogun/base/SGObject.h
```diff
@@ -996,6 +989,21 @@ class CSGObject
...
 		const std::string& name, const std::string& description);
 
+	/**
+	 * Get the current step for the observed values.
+	 */
+#ifndef SWIG
+	SG_FORCED_INLINE int64_t get_step()
+	{
+		int64_t step = -1;
+		Tag<int64_t> tag("current_iteration");
```
vigsterkr: this way of doing things will tie and observed value to be 'stepped' by `current_iteration` object tag... which should not be the case **always**. could be but not necessary

src/shogun/base/SGObject.h
```diff
@@ -1120,5 +1128,161 @@ CSGObject* get_by_tag(const CSGObject* obj, const std::string& name,
...
+	 * Destructor
+	 */
+	~ObservedValue(){};
+
+#ifndef SWIG
+	/**
+	* Return a any version of the stored type.
+	* @return the any value.
+	*/
+	virtual Any& get_any()
```
vigsterkr: what happens in swig?

src/shogun/base/SGObject.h
karlnapf: This was preliminary.
Let’s make it virtual then!?

src/shogun/base/SGObject.h
karlnapf: It’s simply not needed there

vigsterkr: so how do i get the observed value in swig?

src/shogun/base/SGObject.h
vigsterkr: noup... it's the wrong place imo. discussed with @geektoni he'll try to address it later

src/shogun/base/SGObject.h
karlnapf: Get

karlnapf: That’s the whole point of this pr

vigsterkr: this was already true prior pr, as observevalue is sgobject-like already... alas it wasn't registered... but imo it's really an unnecessary overhead to have this whole thing sgobject-like for the sake of registering and having the getter interface

src/shogun/base/SGObject.h
```diff
@@ -987,6 +959,27 @@ class CSGObject
...
+	 * Observe a parameter value given some information
+	 * @tparam T value of the parameter
+	 * @param step step
+	 * @param name name of the observed value
+	 * @param description description
+	 * @param value observed value
+	 */
+	template <class T>
+	void
+	observe(int64_t step, std::string name, std::string description, T value);
```
vigsterkr: does this actually changes anything in SGObject? if not then `const` it

src/shogun/base/SGObject.h
```diff
@@ -987,6 +959,27 @@ class CSGObject
...
+
+	/**
+	 * Observe a registered tag.
+	 * @tparam T type of the tag
+	 * @param step step
+	 * @param name tag's name
+	 */
+	template <class T>
+	void observe(int64_t step, std::string name);
```
vigsterkr: does this actually changes anything in SGObject? if not then `const` it

src/shogun/base/SGObject.h
```diff
@@ -996,6 +989,21 @@ class CSGObject
 	void register_observable(
 		const std::string& name, const std::string& description);
 
+	/**
+	 * Get the current step for the observed values.
+	 */
+#ifndef SWIG
+	SG_FORCED_INLINE int64_t get_step()
```
vigsterkr: does this actually changes anything in SGObject? if not then `const` it

src/shogun/base/SGObject.h
vigsterkr: does this actually changes in the object? if not then `const` it

src/shogun/base/SGObject.cpp
```diff
@@ -1108,5 +1108,12 @@ std::string CSGObject::string_enum_reverse_lookup(
 		    return p.second == enum_value;
 	    });
 	return enum_map_it->first;
+}
 
+ObservedValue::ObservedValue(int64_t step, std::string name)
```
vigsterkr: const inputs

src/shogun/base/SGObject.h
vigsterkr: const inputs

src/shogun/base/SGObject.h
vigsterkr: const inputs

src/shogun/base/SGObject.h
```diff
@@ -1120,5 +1128,161 @@ CSGObject* get_by_tag(const CSGObject* obj, const std::string& name,
...
+ */
+class ObservedValue : public CSGObject
+{
+public:
+	/**
+	 * Constructor
+	 * @param step step
+	 * @param name name of the observed value
+	 */
+	ObservedValue(int64_t step, std::string name);
```
vigsterkr: const inputs

src/shogun/base/SGObject.h
```diff
@@ -1120,5 +1128,161 @@ CSGObject* get_by_tag(const CSGObject* obj, const std::string& name,
...
+
+public:
+	/**
+	 * Constructor
+	 * @param step step
+	 * @param name the observed value's name
+	 * @param value the observed value
+	 */
+	ObservedValueTemplated(
+		int64_t step, std::string name, std::string description, T value)
```
vigsterkr: const inputs

geektoni:
Once CI becomes green and if there are no objections I'll merge this in the feature branch.

src/shogun/base/SGObject.h
karlnapf: should have said series of pr. I agree I dont like the overhead as well. I would like to carve that functionality out of SGObject into a base class that just serves the parameter registration/access purposes. But this is of course out of scope for this little PR here.
The getter interface is a huge advantage as otherwise we once again need to expose/implement/swig-expose a lot of types. Let's discuss this next time on video

src/interfaces/swig/ParameterObserver.i
```diff
@@ -1,19 +1,19 @@
 %rename(ParameterObserverCV) CParameterObserverCV;
 
 %{
-    #include <shogun/lib/parameter_observers/ParameterObserverInterface.h>
-	#include <shogun/lib/parameter_observers/ParameterObserverCV.h>
+    #include <shogun/lib/observers/ParameterObserverInterface.h>
+	#include <shogun/lib/observers/ParameterObserverCV.h>
```
karlnapf: do we still need these interfaces?

karlnapf: Also indentation

src/shogun/base/SGObject.cpp
```diff
@@ -1108,5 +1108,12 @@ std::string CSGObject::string_enum_reverse_lookup(
 		    return p.second == enum_value;
 	    });
 	return enum_map_it->first;
+}
 
+ObservedValue::ObservedValue(const int64_t step, const std::string name)
```
karlnapf: string should be const reference

src/interfaces/swig/ParameterObserver.i
geektoni: They will be removed here #4557

src/shogun/base/SGObject.h
```diff
@@ -996,6 +988,22 @@ class CSGObject
 	void register_observable(
 		const std::string& name, const std::string& description);
 
+	/**
+	 * Get the current step for the observed values.
+	 */
+#ifndef SWIG
+	SG_FORCED_INLINE int64_t get_step() const
```
geektoni: @karlnapf If I do not guard it against SWIG this will generate an error when building the interfaces. I have no idea why since the method is protected. It may be caused by the fact that it is used inside `put()`.

karlnapf: Even if it is protected?

karlnapf: Also what’s the error? :)

src/shogun/base/SGObject.h
karlnapf: What’s the solution here then?

src/shogun/base/SGObject.h
geektoni:
```
/build/opt/include/shogun/base/SGObject.h:994: Error: Syntax error - possibly a missing semicolon.
src/interfaces/python/CMakeFiles/_interface_python.dir/build.make:61: recipe for target 'src/interfaces/python/shogunPYTHON_wrap.cxx' failed
make[2]: *** [src/interfaces/python/shogunPYTHON_wrap.cxx] Error 1
CMakeFiles/Makefile2:86: recipe for target 'src/interfaces/python/CMakeFiles/_interface_python.dir/all' failed
make[1]: *** [src/interfaces/python/CMakeFiles/_interface_python.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
```

karlnapf: it is not protected. SWIG should ignore protected methods. If it is not possible to do that then guarding is fine

PR #4592, opened 2019-03-26
Make classifiers observable.

geektoni:
@karlnapf This is how it would look like with the current design. (Related to #4590)

### DIFF
--- a/src/shogun/classifier/AveragedPerceptron.cpp
+++ b/src/shogun/classifier/AveragedPerceptron.cpp
@@ -89,7 +89,12 @@ void CAveragedPerceptron::iteration()
 		}
 		linalg::update_mean(w, cached_w, num_prev_weights);
 		linalg::update_mean(bias, cached_bias, num_prev_weights);
+
+		observe<SGVector<float64_t>>(m_current_iteration, "w");
+		observe<float64_t>(m_current_iteration, "bias");
+
 		num_prev_weights++;
 	}
+
 	m_complete = converged;
 }
--- a/src/shogun/classifier/Perceptron.cpp
+++ b/src/shogun/classifier/Perceptron.cpp
@@ -73,8 +73,9 @@ void CPerceptron::iteration()
 		{
 			converged = false;
 			const auto gradient = learn_rate * true_label;
-			bias += gradient;
+			put("bias", bias+gradient);
 			v.add(gradient, w);
+			put("w", w);
 		}
 	}
 	m_complete = converged;


src/shogun/classifier/AveragedPerceptron.cpp
```diff
@@ -91,5 +92,9 @@ void CAveragedPerceptron::iteration()
 		linalg::update_mean(bias, cached_bias, num_prev_weights);
 		num_prev_weights++;
 	}
+
+	observe(ObservedValue::make_observation(m_current_iteration, "w", "Model weights", w));
```
geektoni: Without `put()` we would need to do it the "old way".

src/shogun/classifier/AveragedPerceptron.cpp
```diff
@@ -9,6 +9,7 @@
 #include <shogun/features/iterators/DotIterator.h>
 #include <shogun/labels/Labels.h>
 #include <shogun/mathematics/linalg/LinalgNamespace.h>
+#include <shogun/lib/parameter_observers/ObservedValueTemplated.h>
```
karlnapf: I wonder whether we could have some kind of meta include file that has a nicer name? `shogun/lib/observers/observers.h` (minor)
Also I suggest to rename that dir `parameter_observers` -> `observers` (even more minor)

src/shogun/classifier/AveragedPerceptron.cpp
karlnapf: I see. Well that would be the thing for xvalidation then, but definitely not for the weights of the perceptron (those should go via put)

src/shogun/lib/parameter_observers/ObservedValueTemplated.h
```diff
@@ -39,6 +39,20 @@ namespace shogun
...
+		 * Constructor which takes AnyParameterProperties for the observed value
+		 * @param step step
+		 * @param name the observed value's name
+		 * @param value the observed value
+		 * @param properties properties of that observed value
+		 */
+		ObservedValueTemplated(int64_t step, std::string name, T value, AnyParameterProperties properties)
+		: ObservedValue(step, name), m_observed_value(value)
+		{
+			this->watch_param("value", &m_observed_value, properties);
```
karlnapf: I was thinking that the name goes in here where "value" currently is? Or am I confusing two things there?

tests/unit/lib/parameter_observers/ObservedValue_unittest.cc
```diff
@@ -15,4 +15,29 @@ TEST(ObservedValue, set_correct)
 	EXPECT_EQ(obs->get<int64_t>("step"), 1);
 	EXPECT_EQ(obs->get<std::string>("name"), "test");
 	EXPECT_EQ(obs->get<int32_t>("value"), 42);
+};
+
+TEST(ObservedValue, set_correct_properties)
+{
+	AnyParameterProperties any_prop("test description",
+										   ParameterProperties::MODEL);
```
karlnapf: the indentation is weird here

src/shogun/classifier/Perceptron.cpp
```diff
@@ -77,6 +78,10 @@ void CPerceptron::iteration()
 			v.add(gradient, w);
 		}
 	}
+
+	observe(ObservedValue::make_observation(m_current_iteration, "w", "Model weights", m_w));
+	observe(ObservedValue::make_observation(m_current_iteration, "bias", "Model bias", bias));
```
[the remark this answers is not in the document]
geektoni: Unfortunately, this is not possible since otherwise, we would need to have a templated `observe` which brings back all the usual problem we are facing with `put()`.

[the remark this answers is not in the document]
karlnapf: now that the put is "solved", is it reasonable to ask about this again? :)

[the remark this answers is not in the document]
geektoni: For sure ;) now everything is possible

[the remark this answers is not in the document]
karlnapf: haha, so yeah I think it would be a bit neater.
One thing we need to keep in mind is that if we make `observe` templated, then it cannot be virtual anymore so we cannot really overload it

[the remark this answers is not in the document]
geektoni: Ideally, `observe` does not need to be overloaded.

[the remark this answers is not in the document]
karlnapf: ok, then why not give that a try?

src/shogun/lib/parameter_observers/ObservedValueTemplated.h
geektoni: Here we have a variable `name` which identifies the content of the ObservedValue and the registered parameter `value` which instead store the actual value.

karlnapf: ah I see of course, two layers. This will be more clear when there is an example...ill check it again then

src/shogun/classifier/Perceptron.cpp
```diff
@@ -73,8 +73,9 @@ void CPerceptron::iteration()
 		{
 			converged = false;
 			const auto gradient = learn_rate * true_label;
-			bias += gradient;
+			put("bias", bias+gradient);
```
geektoni: This is the updated version of perceptron with the new `put()`.

src/shogun/classifier/AveragedPerceptron.cpp
```diff
@@ -89,7 +89,12 @@ void CAveragedPerceptron::iteration()
 		}
 		linalg::update_mean(w, cached_w, num_prev_weights);
 		linalg::update_mean(bias, cached_bias, num_prev_weights);
+
+		observe<SGVector<float64_t>>(get_step(), "w");
```
geektoni: Regarding the AveragedPerceptron, I have used directly the `observe` method since we are using a linalg call to update both w and the bias. I thought it did not make any sense to use `put()` here since the values are updated in place (and it is more efficient). I think that in many other places around Shogun we will find these kinds of situations.  What do you think? @karlnapf

karlnapf: Yep I think that is good. As long as the algorithm is updating its members rather than some temporary variable.

karlnapf: I am confused about the call, wouldnt you have to pass the `w` vector to the `observe` call? Or does this call mean "observe a model parameter 'w'" ?

karlnapf: One thing, the `get_step()` is not that nice here, maybe we could do that in the same way as in put? I.e. add inferring the step to the observe method? (We can keep the overloaded version that accepts a step, but I feel it would be verbose to do that in here)

src/shogun/classifier/Perceptron.cpp
karlnapf: looks good!

src/shogun/base/SGObject.h
```diff
@@ -1022,5 +1028,161 @@ class CSGObject
...
+
+template <class T>
+void CSGObject::observe(
+	int64_t step, std::string name, std::string description, T value)
+{
+	auto obs = some<ObservedValueTemplated<T>>(step, name, description, value);
+	this->observe(obs);
+}
+
+template <class T>
```
karlnapf: aha here we go, didnt see that earlier

src/shogun/base/SGObject.h
```diff
@@ -966,6 +938,27 @@ class CSGObject
...
+	 * @param name name of the observed value
+	 * @param description description
+	 * @param value observed value
+	 */
+	template <class T>
+	void
+	observe(int64_t step, std::string name, std::string description, T value);
+
+	/**
+	 * Observe a registered tag.
```
karlnapf: minor: "registered parameter"

Tag is the identifier for parameter

src/shogun/base/SGObject.h
```diff
@@ -966,6 +938,27 @@ class CSGObject
...
+	 * @tparam T type of the tag
+	 * @param step step
+	 * @param name tag's name
+	 */
+	template <class T>
+	void observe(int64_t step, std::string name);
```
karlnapf: not sure why we would need this one? To avoid passing the value? Is that an advantage?

src/shogun/base/SGObject.h
```diff
@@ -966,6 +938,27 @@ class CSGObject
...
+	 * Observe a parameter value given some information
+	 * @tparam T value of the parameter
+	 * @param step step
+	 * @param name name of the observed value
+	 * @param description description
+	 * @param value observed value
+	 */
+	template <class T>
+	void
+	observe(int64_t step, std::string name, std::string description, T value);
```
karlnapf: btw wouldnt `const T& value` be more appropriate here?

src/shogun/classifier/AveragedPerceptron.cpp
geektoni: @karlnapf I've updated these calls. Do you still have doubts about them? :)

PR #4594, opened 2019-03-28
to_string some class

vigsterkr:
because having to string is good in repl of interfaces.

### DIFF
--- a/src/shogun/base/AnyParameter.h
+++ b/src/shogun/base/AnyParameter.h
@@ -12,6 +12,7 @@
 #include <shogun/lib/bitmask_operators.h>
 
 #include <memory>
+#include <list>
 
 namespace shogun
 {
@@ -40,6 +41,14 @@ namespace shogun
 		AUTO = 1u << 10,
 		READONLY = 1u << 11
 	};
+
+	static const std::list<std::pair<ParameterProperties, std::string>> kParameterPropNames = {
+		{ParameterProperties::NONE, "NONE"},
+		{ParameterProperties::HYPER, "HYPER"},
+		{ParameterProperties::GRADIENT, "GRADIENT"},
+		{ParameterProperties::MODEL, "MODEL"},
+		{ParameterProperties::AUTO, "AUTO"},
+		{ParameterProperties::READONLY, "READONLY"}};
 
 	enableEnumClassBitmask(ParameterProperties);
 
@@ -134,6 +143,22 @@ namespace shogun
 		void remove_property(const ParameterProperties other)
 		{
 			m_attribute_mask &= ~other;
+		}
+		std::string to_string() const
+		{
+			std::stringstream ss;
+			ss << "Description: "<< m_description << " with attributes: [";
+			bool first_attrib = true;
+			for (const auto& it: kParameterPropNames)
+			{
+				if (has_property(it.first))
+				{
+					ss << (first_attrib ? "": " | ") << it.second;
+					first_attrib = false;
+				}
+			}
+			ss << "]";
+			return ss.str();
 		}
 
 	private:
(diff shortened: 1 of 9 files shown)


src/shogun/lib/SGMatrix.cpp
```diff
@@ -506,270 +506,54 @@ void SGMatrix<T>::remove_column_mean()
...
-		for (int64_t j=0; j<cols; j++)
-			SG_SPRINT("%s\t%d%s", prefix, matrix[j*rows+i],
-				j==cols-1? "" : ",");
-		SG_SPRINT("%s]%s\n", prefix, i==rows-1? "" : ",")
-	}
-	SG_SPRINT("%s]\n", prefix)
-}
-
-template <>
-void SGMatrix<uint32_t>::display_matrix(
```
karlnapf: lol, these methods are just insane

src/shogun/base/AnyParameter.h
```diff
@@ -31,14 +32,17 @@ namespace shogun
...
+	enum class ParameterProperties: std::uint32_t
 	{
 		NONE = 0,
 		HYPER = 1u << 0,
 		GRADIENT = 1u << 1,
 		MODEL = 1u << 2,
 		AUTO = 1u << 10,
-		READONLY = 1u << 11
+		READONLY = 1u << 11,
+		Last,
```
gf712: might be worth adding a doxygen for this no? otherwise Last and First seems a bit random

src/shogun/base/AnyParameter.h
```diff
@@ -31,14 +32,17 @@ namespace shogun
...
+		ALL = std::numeric_limits<std::uint32_t>::max(),
```
gf712: Did this work well then? Is there anyway to infer `uint32_t` from the inheritance btw?

src/shogun/base/AnyParameter.h
```diff
@@ -135,6 +139,20 @@ namespace shogun
...
+		std::string to_string() const
+		{
+			std::stringstream ss;
+			ss << "Description: "<< m_description << " with attributes: [";
+			for (auto it = begin<ParameterProperties>(); it != end<ParameterProperties>(); ++it)
+			{
+				if (it != begin<ParameterProperties>())
+					ss << ", ";
+				if (has_property(*it))
+					ss << static_cast<std::underlying_type_t<ParameterProperties>>(*it);
```
gf712: Should do casting with `underlying_type` in the other methods, i.e. `EGradientAvailability get_gradient() const`

gf712: Also might be worth printing the bit position rather than the int representation no?

src/shogun/base/AnyParameter.h
gf712: should probably add a unittest for this?

src/shogun/base/AnyParameter.h
vigsterkr: i would rather like to have a string representation of the flags :)

gf712: in that case shouldn't you keep a map with a readable string representation, i.e. {ParameterProperties::HYPER: "hyperparameter"} and print the values?

gf712: because right now this is just writing the integers right?

src/shogun/base/AnyParameter.h
```diff
@@ -10,8 +10,10 @@
 #include <shogun/lib/abstract_auto_init.h>
 #include <shogun/lib/any.h>
 #include <shogun/lib/bitmask_operators.h>
+#include <shogun/util/iterators.h>
```
gf712: can remove this right?

src/shogun/util/iterators.h
```diff
@@ -33,6 +33,9 @@
 #ifndef SHOGUN_ITERATORS_H_
 #define SHOGUN_ITERATORS_H_
 
+#include <iterator>
+#include <type_traits>
```
gf712: and this?

PR #4596, opened 2019-03-29
DO NOT MERGE

karlnapf:


src/shogun/base/SGObject.cpp
```diff
@@ -1119,4 +1119,13 @@ std::string CSGObject::string_enum_reverse_lookup(
...
-}
\ No newline at end of file
+}
+
+namespace shogun
+{
+template <class T>
+SG_FORCED_INLINE void test()
+{
+	T* a;
```
karlnapf: implementation depends on a header file that cannot be included in `SGObject.h` as it itself includes `SGObject.h`

PR #4597, opened 2019-03-29
fix return type of lambda

gf712:
This is needed to raise compile time error when passing a constant as a non const reference to lambda.

This works when any is const:
```
auto lambda = [](auto const& val) {}
auto lambda = [](auto val) {}
```
This [should] fail when any is const
```
auto lambda = [](auto& val) {}
```
The latter does not cause a compile time error and leads to unexpected behaviour at runtime


karlnapf:
scary ;)

gf712:
@karlnapf this seems to be good to go. It seems like static assert tests in gtest are not available for msvc so I removed them.

PR #4598, opened 2019-03-31
Refactor GridSearch and parameter tree

gf712:
Proposed API (in python):

OPTION 1: change nested object parameter values with a string 
```python
import numpy as np   
import shogun as sg   
svm = sg.machine("LibSVM", kernel=kernel)   
ps = sg.GridParameters(svm) 
ps.attach("kernel::log_width", np.array([1., 2.])) 
ps.attach("kernel::combined_kernel_weight", np.array([1., 2.])) 
ps.attach("C1", np.array([1., 2.])) 
ps.attach("C2", np.array([1., 2.])) 
for x in range(16):
    print(ps.next_combination()) 
```
OPTION 2: change nested object parameters by manipulating the nested node directly, before adding it to the tree
```python
import numpy as np   
import shogun as sg   
svm = sg.machine("LibSVM")
ps = sg.GridParameters(svm) 
kernel1 = sg.kernel("PolyKernel") 
node1 = sg.GridParameters(kernel1) 
node1.attach("degree", np.array([1.,2,3])) 
ps.attach("kernel", node1) 
ps.attach("C1", np.array([1., 2.])) 
ps.attach("C2", np.array([1., 2.]))
for x in range(12):
    print(ps.next_combination())
```
OPTION 1 and 2: combine both options, but option 1 becomes unavailable when there are several nodes representing a single object, i.e. kernel=[GaussianKernel, PolyKernel]
```python
import numpy as np 
import shogun as sg 

kernel1 = sg.kernel("GaussianKernel")  
kernel2 = sg.kernel("PolyKernel") 
svm = sg.machine("LibSVM", kernel=kernel1)

node = sg.GridParameters(svm)
node.attach("kernel::log_width", np.array([1.]))
node.attach("C1", np.array([1., 2., 3.])) 
node.attach("C2", np.array([1., 2., 3.])) 


node2=sg.GridParameters(kernel2)
node2.attach("degree", np.array([1.,2.]))

node.attach("kernel", node2)

for x in range(20):
    print(node.next_combination()) 
```

- [x] implement option 1
- [x] implement option 2 
- [x] implement GridSearch engine


karlnapf:
Would it be possible to merge `GridParameters` and `GridSearch`?

gf712:
Working on that bit! But yes that would work, I called it ModelSelectionTree for now

src/shogun/base/SGObject.h
```diff
@@ -1022,5 +1029,123 @@ class CSGObject
...
+					if (i < v->vlen - 1)
+						stream() << ",";
+				}
+				stream() << "]";
+			}
+		}
+
+	private:
+		std::stringstream* m_stream;
+	};
```
gf712: Reminder to self, once #4594 is done can put this back into place and use SGVector to_string method

src/shogun/modelselection/parameter_tree.h
```diff
@@ -0,0 +1,259 @@
...
+		 */
+		static const std::string delimiter;
+
+		/**
+		 * Internal method to get the string representation of each value in the
+		 * tree
+		 * @param ss
+		 * @param visitor
+		 */
+		void to_string_helper(
```
gf712: and when #4594 is done also refactor this

karlnapf:
> Working on that bit! But yes that would work, I called it ModelSelectionTree for now

Actually I think GridSearch is pretty good name wise :) No need to make "tree" part if it imo

karlnapf:
I think you forgot to update the gpl submodule ...

gf712:
@karlnapf 
this is how I see it working for now:
```python
import numpy as np 
import shogun as sg 
svm = sg.machine("LibSVM", kernel=kernel) 
ps = sg.ModelSelectionTree(svm) 
ps.attach("GaussianKernel::log_width", np.array([1., 2., 3.])) 
ps.attach("C1", np.array([1., 2., 3.])) 
ps.next_combination()
ps.next_combination()
...
```
And `next_combination` is called within GridSearch, so have:
```python
import numpy as np 
import shogun as sg 
svm = sg.machine("LibSVM", kernel=kernel) 
gs = sg.GridSearch(svm) 
gs.attach("GaussianKernel::log_width", np.array([1., 2., 3.])) 
gs.attach("C1", np.array([1., 2., 3.]))
gs.train(...)
```
Working on the second part now where multiple kernels could be added

src/interfaces/swig/ModelSelection.i
```diff
@@ -10,6 +10,8 @@
 %newobject CParameterCombination::leaf_sets_multiplication();
 %newobject CModelSelectionParameters::get_combinations();
 %newobject CModelSelectionParameters::get_single_combination();
+%newobject ParameterNode::attach();
```
karlnapf: I get that we need attach, but why would we need to call `next_combination` from swig?

src/interfaces/swig/shogun.i
```diff
@@ -300,5 +300,17 @@ PUT_ADD(CTokenizer)
 %template(kernel) kernel<float64_t, float64_t>;
 %template(features) features<float64_t>;
 
+%template(attach) ParameterNode::attach<int32_t>;
```
karlnapf: best way to test this is to add a meta example

src/interfaces/swig/ModelSelection.i
gf712: that's just for debugging right now

src/shogun/lib/type_case.h
```diff
@@ -90,30 +88,30 @@ namespace shogun
...
-	struct is_sg_vector<T> : public std::true_type                             \
-	{                                                                          \
-	};
-#define SG_ADD_SGMATRIX_TYPE(T, type_)                                         \
-	SG_ADD_TYPE(T, type_)                                                      \
-	template <>                                                                \
-	struct is_sg_matrix<T> : public std::true_type                             \
-	{                                                                          \
-	};
+	#define SG_ADD_TYPE(T, type_)                                                  \
```
karlnapf: maybe do this into a separate PR?

src/shogun/modelselection/parameter_tree.cpp
```diff
@@ -0,0 +1,473 @@
...
+		if (auto* value = m_parent->get(param.first, std::nothrow))
+		{
+			create_node(param.first, value);
+			SG_UNREF(value);
+		}
+	}
+	m_current_node = m_nodes.begin();
+}
+
+void ModelSelectionTree::reset()
```
karlnapf: what is that needed for?

src/shogun/modelselection/parameter_tree.cpp
```diff
@@ -0,0 +1,473 @@
...
+				SG_PRINT("updating m_current_param from %s::%s to %s::%s\n", get_parent_name(), param.c_str(), get_parent_name(), m_current_param.c_str())
+			}
+			tree->attach(
+					param, *any_cast<decltype(val.begin())>(m_param_begin[param]));
+			return;
+		}
+
+		tree->attach(
+				param, *any_cast<decltype(val.begin())>(m_param_iter[param]));
+		SG_PRINT("CURRENT: %s, PARAM: %s\n", m_current_param.c_str(), param.c_str())
```
karlnapf: you can make those prints DEBUG and then leave them in actually, might be useful for future debugging?

src/shogun/modelselection/parameter_tree.cpp
gf712: resets the internal state of a node, so that it goes back from the beginning when called from the parent node.

src/shogun/modelselection/parameter_tree.cpp
gf712: yup, I'll do that in the end, I just don't want to set to debug level because then it becomes too verbose

src/shogun/modelselection/parameter_tree.h
```diff
@@ -0,0 +1,259 @@
...
+		virtual void create_node(const std::string& name, CSGObject* obj);
+#endif
+
+		/**
+		 * Attach a new node at a given location in the tree
+		 * defined by the parameter name
+		 * @param name
+		 * @param node
+		 */
+		ParameterNode* attach(
```
karlnapf: I would hide most things from swig in here, e.g. this one

src/shogun/modelselection/parameter_tree.h
```diff
@@ -0,0 +1,259 @@
...
+		 * Attach a new value at a given location in the tree
+		 * defined by the parameter name
+		 * @param name
+		 * @param node
+		 */
+		template <typename T>
+		ParameterNode* attach(const std::string& param, T value)
+		{
+			if (!set_param_helper(param, make_any(value)))
+				SG_SERROR("Could not attach %s", param.c_str());
```
karlnapf: if this is user facing, maybe put some more infos in there? So if users chain multiple attachments, they know which one failed from the error message and see what has been built so far?

src/shogun/modelselection/parameter_tree.h
gf712: This will be needed to add another GridSearch instance

src/shogun/modelselection/parameter_tree.h
```diff
@@ -0,0 +1,259 @@
...
+#include <shogun/base/SGObject.h>
+#include <shogun/lib/SGVector.h>
+#include <shogun/lib/type_case.h>
+#include <shogun/machine/Machine.h>
+#include <unordered_set>
+
+namespace shogun
+{
+
+	class ParameterNode : public CSGObject
```
karlnapf: I wonder ... does this need to be a subclass of SGObject? Would it ever need to be serialized? Would it ever need to be cloned/equal? Probably not or?

karlnapf: It would be good to have some kind of high level explain of the parts involved in here and how they play together, in some doxygen comment. For future development of the stuff. Not now, but definitely later

gf712:
I will have to go through the reference to check how smart pointers and stl containers handle race conditions, but I *think* it should be fine

karlnapf:
Also in terms of instantiating the combinations.... I would imagine that a central generator creates combinations which are then sent to workers to solve, and in particular multiple combinations need to be instantiated before it is clear which ones came back yet....  but ok, let's think about that later :)

gf712:
> Also in terms of instantiating the combinations.... I would imagine that a central generator creates combinations which are then sent to workers to solve, and in particular multiple combinations need to be instantiated before it is clear which ones came back yet.... but ok, let's think about that later :)

I guess we could keep a buffer that feeds the workers? So would always have a pool of at least N jobs left, if possible, where N >= number_of_workers.

gf712:
@karlnapf This is how I am writing the API for option 2:

```python
import numpy as np 
import shogun as sg 

kernel1 = sg.kernel("GaussianKernel")  
kernel2 = sg.kernel("PolyKernel") 
svm = sg.machine("LibSVM", kernel=kernel1)

node = sg.GridParameters(svm)
# option 1 works here -> only one node in kernel parameter
node.attach("kernel::log_width", np.array([1.,2.,3.]))

node2=sg.GridParameters(kernel2)
node2.attach("degree", np.array([1.,2.,3.]))

node.attach("kernel", node2)

# raises exception -> ambiguous
node.attach("kernel::log_width", np.array([1.,2.,3.])) 
```

examples/meta/generator/targets/cpp.json
```diff
@@ -12,6 +12,7 @@
     "Init": {
         "Construct": "auto $name = some<C$typeName>($arguments)$kwargs",
         "Copy": "auto $name = wrap($expr)$kwargs",
+        "CopyRaw": "auto $name = $expr$kwargs",
```
gf712: probably not the best name as nothing is being copied

examples/meta/src/model_selection/tree_builder.sg
```diff
@@ -0,0 +1,24 @@
...
+ParameterNode k1_node = new ParameterNode(k1)
+ParameterNode k2_node = new ParameterNode(k2)
+#![create_child_nodes]
+
+#![attach_child_nodes]
+k2_node.attach("degree", 2)
+node.attach("kernel", k1_node)
+node.attach("kernel::log_width", 10)
+node.attach("kernel", k2_node)
+#![attach_child_nodes]
```
gf712: @karlnapf here is a simple example of how a tree can be built

src/shogun/modelselection/NewGridSearch.h
```diff
@@ -0,0 +1,354 @@
...
+		 * Internal method to get the current parameter combination
+		 * @return
+		 */
+		virtual ParameterNode* get_current();
+
+		/** vector of child nodes */
+		std::map<std::string, std::vector<std::shared_ptr<ParameterNode>>>
+		    m_nodes;
+		/** pointer to object that is being mimicked */
+		CSGObject* m_parent;
```
gf712: @vigsterkr do you think i should use a raw pointer? This is pointing to a model that is passed to the constructor (where i increment the counter). Would it make sense to use `std::unique_ptr`? Right now I unreference this pointer in the destructor.

src/shogun/modelselection/NewGridSearch.h
```diff
@@ -0,0 +1,354 @@
...
+		explicit GridSearch(CSGObject* model) : GridParameters(model)
+		{
+		}
+
+		const char* get_name() const noexcept override
+		{
+			return "GridSearch";
+		}
+
+		void train(CFeatures* data);
```
gf712: need to think about how this works here. right now the labels have to be `put` in the model passed to the GridSearch constructor

src/shogun/modelselection/NewGridSearch.h
```diff
@@ -0,0 +1,354 @@
...
+
+	private:
+		SGVector<float64_t> m_scores;
```
gf712: need to think about evaluation and how to store each model's performance

src/shogun/modelselection/NewGridSearch.h
karlnapf: Should we really have a "train" method here? It is not model training what this does but parameter learning. So "learn" would be a more appropriate name. Or "select_model" or whatever.
Also I find it weird that this does not return anything? Shouldnt this return some kind of representation of the best parameters learned?

src/shogun/modelselection/NewGridSearch.h
karlnapf: Check out the modelselection code we have. There is a criterion, which is either maximised or minimised. And then there is an "evaluation" object that determines how the instantiated models are evaluated (xvalidation, etc).
I think this modular setup is quite important

* modelselection (how are the parameters to be tried generated? grid-search? gradient search?)
* model (train, apply)
* criterion (accuracy? AUC?)
* evaluation (how is the model evaluated? xvalidation? test set? marginal likelihood? theoretical measures? other?)
* parameters (these are the best parameters found, or some intermediate best results)

examples/meta/src/model_selection/grid_search.sg
```diff
@@ -0,0 +1,58 @@
...
+#![create_machine]
+
+#![create_master_node]
+GridSearch node = new GridSearch(svm)
+RealVector c1_param(3)
+c1_param[0] = 0.1
+c1_param[1] = 1
+c1_param[2] = 10
+RealVector c2_param(3)
+c2_param[0] = 0.1
```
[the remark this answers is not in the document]
gf712: I mean one can be a reference to the other... But is it always the case?

src/shogun/base/SGObject.h
vigsterkr: merged :)

examples/meta/generator/parse.py
```diff
@@ -251,6 +253,10 @@ def p_enum(self, p):
         "enum : ENUMKEYWORD identifier DOT identifier"
         p[0] = {"Enum": [p[2], p[4]]}
 
+    def p_new(self, p):
```
karlnapf: i think we should move this to a different PR (including a meta example for testing) to de-clutter the PR a bit?

examples/meta/generator/targets/cpp.json
karlnapf: change it? :)

examples/meta/src/model_selection/grid_search.sg
```diff
@@ -0,0 +1,71 @@
...
+RealVector width_param(3)
+width_param[0] = 0.1
+width_param[1] = 1
+width_param[2] = 2
+node.attach("kernel::log_width", width_param)
+node.attach("kernel", k2_node)
+#![attach_child_nodes]
+
+#![train_and_apply]
+Machine best_model = as_machine(node.train(features_train, labels_train, True))
```
karlnapf: so with this approach, the user gets back a parametrized machine that she can extract the best parameters from?

examples/meta/src/model_selection/grid_search.sg
```diff
@@ -0,0 +1,71 @@
...
+RealVector c2_param(3)
+c2_param[0] = 0.1
+c2_param[1] = 1
+c2_param[2] = 10
+node.attach("C1", c1_param)
+node.attach("C2", c2_param)
+#![create_master_node]
+
+#![create_child_nodes]
+GridParameters k1_node = new GridParameters(k1)
```
karlnapf: I am still a bit confused why the type of the base node and the sub-nodes is different?

examples/meta/src/model_selection/grid_search.sg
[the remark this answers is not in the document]
karlnapf: Actually let's discuss this later as things get out of hand complexity wise otherwise :)

examples/meta/src/model_selection/tree_builder.sg
```diff
@@ -0,0 +1,24 @@
...
+#![create_master_node]
+
+#![create_child_nodes]
+ParameterNode k1_node = new ParameterNode(k1)
+ParameterNode k2_node = new ParameterNode(k2)
+#![create_child_nodes]
+
+#![attach_child_nodes]
+k2_node.attach("degree", 2)
+node.attach("kernel", k1_node)
```
karlnapf: I am just having the crazy idea of re-using `put/add` here: 
The constructor `new ParameterNode(k1)` returns an object that has parameters registered inside (it checks which parameters the provided SGObject has and registers the same in the returned object). Then users just would do `k1_node.add("kernel", 2)` for adding to the options, or `k1_node.put("kernel", 2)` if they want a single value?
I am not sure whether it is really a good idea to do this since it has other implications, but it would save you the pain from implementing "attach" and adding all the typed versions for it, which is kinda re-implementing a lot of stuff in the swig interface (e.g. the java hacks). Thoughts?

examples/meta/src/model_selection/tree_builder.sg
```diff
@@ -0,0 +1,24 @@
...
+node.attach("kernel::log_width", 10)
```
karlnapf: this obviously wouldnt work then

src/shogun/base/SGObject.h
```diff
@@ -1041,6 +1048,124 @@ class CSGObject
...
+			to_string(v);
+		}
+		virtual void on(SGVector<double>* v)
+		{
+			to_string(v);
+		}
+		virtual void on(SGMatrix<int>* mat)
+		{
+			to_string(mat);
+		}
```
karlnapf: maybe also move this whole blob to another PR?

src/shogun/modelselection/NewGridSearch.cpp
```diff
@@ -0,0 +1,891 @@
...
+
+using namespace shogun;
+
+const std::string ParameterNode::delimiter = "::";
+
+ParameterNode::ParameterNode(CSGObject* model)
+{
+	SG_REF(model)
+	m_parent = model;
+	SG_SDEBUG("%s refcount: %d\n", model->get_name(), m_parent->ref_count())
```
karlnapf: FYI if you set the debug level to the lowest, you will see whenever ref-counts are changed

src/shogun/modelselection/NewGridSearch.cpp
```diff
@@ -0,0 +1,891 @@
...
+	for (auto const& param : m_parent->get_params())
+	{
+		// check the param is neither of type CLabels or CFeatures
+		if ((std::type_index(param.second->get_value().type_info()) !=
```
karlnapf: maybe have a central list of the types to exclude somewhere?
But thinking about this, the parameter nodes should only contain parameters that were registered as hyper-parameters in the classes that are passed. Then no filtering would be needed as CFeatures would never be marked as hyper-parameter for modelselection?

src/shogun/modelselection/NewGridSearch.cpp
```diff
@@ -0,0 +1,891 @@
...
+ParameterNode::ParameterNode(
+    CSGObject* model, ParameterProperties param_properties)
+{
+	SG_REF(model)
+	m_parent = model;
+
+	for (auto const& param : m_parent->get_params())
+	{
+		// check the param is neither of type CLabels or CFeatures
+		if ((std::type_index(param.second->get_value().type_info()) !=
```
karlnapf: redundant with above (although as said I think this can go)

src/shogun/modelselection/NewGridSearch.cpp
```diff
@@ -0,0 +1,891 @@
...
+{
+	return attach(param, node.get());
+}
+
+ParameterNode*
+ParameterNode::attach(const std::string& param, ParameterNode* node)
+{
+	auto* node_copy = new ParameterNode(*node);
+	if (!set_param_helper(param, node_copy))
+		SG_SERROR("Could not attach %s\n", param.c_str());
```
karlnapf: would be good to see a reason if this is user-facing

src/shogun/modelselection/NewGridSearch.cpp
```diff
@@ -0,0 +1,891 @@
...
+			}
+			if (node->second.size() > 1)
+			{
+				ss << "]";
+			}
+			if (std::next(node) != m_nodes.end())
+				ss << ", ";
+		}
+	}
+	ss << "))";
```
karlnapf: just curious, could you paste an output of this?

src/shogun/modelselection/NewGridSearch.cpp
```diff
@@ -0,0 +1,891 @@
...
+		SG_SDEBUG(
+		    "ParameterNode::set_param_helper - Looking for param in nodes: "
+		    "%s\n",
+		    param.c_str())
+		for (auto& node : m_nodes)
+		{
+			if (node.second.size() > 1)
+			{
+				SG_SERROR(
+				    "Ambiguous call! More than one %s found in "
```
karlnapf: this is impossible or? Parameters cannot be registered twice...

src/shogun/modelselection/NewGridSearch.cpp
```diff
@@ -0,0 +1,891 @@
...
+	auto tree = new ParameterNode(m_parent);
+
+	std::string param;
+
+	auto get_lambda = [&tree, &param](auto val) { tree->attach(param, val); };
+
+	for (const auto& param_pair : m_param_mapping)
+	{
+		param = param_pair.first;
+		sg_any_dispatch(
```
karlnapf: cool that this is re-used here!

src/shogun/modelselection/NewGridSearch.cpp
```diff
@@ -0,0 +1,891 @@
...
+}
+
+CSGObject* ParameterNode::to_object(const std::unique_ptr<ParameterNode>& tree)
+{
+	SG_SDEBUG("CSGObject* ParameterNode::to_object(const "
+	          "std::unique_ptr<ParameterNode>& tree)\n")
+	SG_SDEBUG(
+	    "REFCOUNT %s %d\n", tree->m_parent->get_name(),
+	    tree->m_parent->ref_count())
+	auto* result = tree->m_parent->clone();
```
karlnapf: I am worried that this also clones the training data when dealing with machines? Because that should not happen

src/shogun/modelselection/NewGridSearch.cpp
```diff
@@ -0,0 +1,891 @@
...
+	SG_SDEBUG("%s: %d\n", result->get_name(), result->ref_count())
+	auto result_params = result->get_params();
+	for (const auto& node : tree->m_nodes)
+	{
+		auto type_index_i =
```
karlnapf: this type of dispatching should be centralised or?

src/shogun/modelselection/NewGridSearch.cpp
```diff
@@ -0,0 +1,891 @@
...
+	}
+}
+
+GridParameters::GridParameters(CSGObject* model)
+    : m_first(true), m_node_complete(false)
+{
+	SG_REF(model)
+	m_parent = model;
+	SG_SDEBUG("%s refcount: %d\n", m_parent->get_name(), m_parent->ref_count())
+	// TODO: once all hyperparameters are flagged properly replace getter with
```
karlnapf: ah I see. I would do that now already, then we can have an entrance task for students to make things model-selectable ...

src/shogun/modelselection/NewGridSearch.cpp
```diff
@@ -0,0 +1,891 @@
...
+	//  get_params.model(ParameterProperties::HYPER)
+	for (auto const& param : model->get_params())
+	{
+		// check the param is neither of type CLabels or CFeatures
+		if ((std::type_index(param.second->get_value().type_info()) !=
+		     std::type_index(typeid(CLabels*))) &&
```
karlnapf: redundant and remove again :)

src/shogun/modelselection/NewGridSearch.cpp
```diff
@@ -0,0 +1,891 @@
...
+			}
+			break;
+		}
+		else
+			sg_any_dispatch(
+			    param_pair->second, sg_vector_typemap, shogun::None{},
+			    get_lambda);
+	}
+
+	return tree;
```
karlnapf: the method is a bit large. Maybe can be split further?

src/shogun/modelselection/NewGridSearch.cpp
```diff
@@ -0,0 +1,891 @@
...
+	while (!is_complete())
+	{
+		// delete old parameter combination and replace with new one
+		SG_SDEBUG("GridSearch::train - GETTING NEW COMBINATION\n")
+		parameter_combination.reset(get_next());
+		SG_SDEBUG("GridSearch::train - GOT NEW COMBINATION\n")
+
+		// the machine is returned with a ref_count of 1 because of clone
+		auto* tmp =
+		    ParameterNode::to_object(parameter_combination)->as<CMachine>();
```
karlnapf: this might create some inefficiencies or, mentioned above: cloning the data?

src/shogun/modelselection/NewGridSearch.cpp
```diff
@@ -0,0 +1,891 @@
...
+			    "GridSearch::train - "
+			    "trying combination: %s\n",
+			    parameter_combination->to_string().c_str())
+			SG_SPRINT(
+			    "GridSearch::train - "
+			    "trying model: %s\n",
+			    eval_machine->get<CMachine*>("machine")->to_string().c_str())
+		}
+
+		// note that this may implicitly lock and unlock the machine
```
karlnapf: I think we should totally remove this concept of locking machines for now tbh. We could add that back later. It is also not something that would be sitting inside GridSearch, but rather in the cross-validation codes, since that is where the issue of precomputing stuff when re-training multiple times using the same parameters happens

src/shogun/modelselection/NewGridSearch.cpp
```diff
@@ -0,0 +1,891 @@
...
+			best_combination = std::move(parameter_combination);
+
+			SG_REF(result);
+			SG_UNREF(best_result);
+			best_result = result;
+		}
+		SG_UNREF(result);
+	}
+
+	if (verbose)
```
karlnapf: Would remove this verbose option. We can soon just make it observable ...

src/shogun/util/factory.h
```diff
@@ -312,9 +295,7 @@ namespace shogun
 	 */
 	CPipelineBuilder* pipeline()
 	{
-		auto result = new CPipelineBuilder();
```
karlnapf: this is also probably material for another PR, not this one :)

examples/meta/generator/parse.py
gf712: Yup, I'll do that. We probably need one for constructors, i.e. std::make_shared and one for functions that return raw pointers, i.e. std::shared_ptr

examples/meta/src/model_selection/grid_search.sg
gf712: yes, that was just for convenience, because I can't return raw pointers right now from a function call (I can from a constructor)

karlnapf: so in general it would return parameters? because this returning a machine might be a bit sketchy (see my cloning comments)

examples/meta/src/model_selection/grid_search.sg
gf712: Hmm, maybe can discuss better this week on irc or something. But basically I don't want each node to have a train method, so I attach nodes to the GridSearch, which has can then be trained.

examples/meta/src/model_selection/grid_search.sg
gf712: yea, I had it returning a parameter combination first, but then I couldn't get the meta examples to work properly for cpp

examples/meta/src/model_selection/grid_search.sg
karlnapf: yeah ok let's discuss in irc. But re your comment I would say that the tree would be something that is passed to the grid-search strategy and the latter has the train method and problem solved or?

src/shogun/base/SGObject.h
gf712: this has to be reset, with viktors PR I can now get the string from the vector

src/shogun/modelselection/NewGridSearch.cpp
gf712: Yea, true, it just had this as SG_SPRINT, and then replaced it with DEBUG in the end

src/shogun/modelselection/NewGridSearch.cpp
gf712: True, eventually that is the goal!

src/shogun/modelselection/NewGridSearch.cpp
gf712: yea, I need to fix the error message a bit. It shouldn't reach this code in theory

karlnapf: Then it could be a plain ASSERT in fact or? But that's minor

src/shogun/util/factory.h
gf712: yup, I just needed to fix this because it was causing "false positive" memory leaks

src/shogun/modelselection/NewGridSearch.cpp
gf712: but you can have several kernels in a node, i.e. svm with gaussian and poly kernel

src/shogun/modelselection/NewGridSearch.cpp
gf712: yea, its really convenient, mostly because it defines a behaviour for 3 generic types: scalar, vector or matrix. whereas the whole visitor pattern would require writing all different scalar types, vector types and so on.. each time it is needed

src/shogun/modelselection/NewGridSearch.cpp
karlnapf: ah yes or course! my bad!

src/shogun/modelselection/NewGridSearch.cpp
gf712: A node could look like this: 
```
ParameterNode(LibSVM(kernel=[GaussianKernel(log_width=10), PolyKernel(degree=2)]))
```
A GridSearch could look like this:
```
GridSearch(LibSVM(C1=[0.1, 1, 10], C2=[0.1, 1, 10], kernel=[GaussianKernel(log_width=[0.1, 1, 2]), PolyKernel(degree=[1, 2, 3])]))
```

examples/meta/src/model_selection/grid_search.sg
```diff
@@ -26,14 +26,8 @@ Evaluation evaluation_criterion = evaluation("AccuracyMeasure")
...
 GridSearch node = new GridSearch(svm, splitting, evaluation_criterion)
-RealVector c1_param(3)
-c1_param[0] = 0.1
-c1_param[1] = 1
-c1_param[2] = 10
-RealVector c2_param(3)
-c2_param[0] = 0.1
-c2_param[1] = 1
-c2_param[2] = 10
+RealVector c1_param([0.1, 1.0, 10.0])
```
karlnapf: :rocket:

src/shogun/modelselection/NewGridSearch.cpp
karlnapf: sweet! really nice

gf712:
keeping it alive for another 180 days, might pick up during GSoC

gf712:
for reference here is a much simpler cartesian product implementation where we can just use visitor patterns (`std::string` becomes `shogun::Any`):

```cpp
std::vector<std::vector<std::string>> cartesian_product(
    const std::vector<std::vector<std::string>>& inputs) {
    std::vector<std::vector<std::string>> result;
    if (std::all_of(inputs.begin(), inputs.end(), [](const auto& el) {return el.empty();})) {
        return result;
    }

    for (auto& el: inputs.front()) {
        result.push_back({el});
    }

    auto size = std::accumulate(inputs.begin(), inputs.end(), 1,
        [](const auto& lhs, const auto& rhs){
            return lhs * rhs.size();
    });

    std::vector<std::vector<std::string>> temp;
    temp.reserve(size);

    for (auto i = 1; i < inputs.size(); ++i) {
        temp.clear();
        for (const auto& e: result) {
            for (const auto& f: inputs[i]) {
                temp.emplace_back(e).push_back(f);
            }
        }
        result = std::move(temp);
    }
    return result;
}

int main()
{
    std::vector<std::string> a {"a1","a2","a3"};
    std::vector<std::string> b {"b1","b2","b3"};
    std::vector<std::string> c {"c1","c2"};

    auto result = cartesian_product({a, b, c});

    std::cout << "Result: \n";
    for (const auto& vec: result) {
        for (const auto& el: vec)
            std::cout << el << ',';
        std::cout << '\n';
    }
}
```

The generator style approach one is much nicer, but it is currently flawed and would need more fixing.

PR #4603, opened 2019-04-10
Fixes memory leak in target languages

gf712:
Right now there is a ref counting issue in the interface languages. Both the factory and some increment the ref counter, but when ~Some is called it only decreases the refcount once. This means the refcount stays at 1, and the object is never destroyed, resulting in a leak.


vigsterkr:
"superseded" (love this one still;) by #4604

Issue #4321, opened 2018-05-31
Factory methods should be registered as %newobject in SWIG interface


vigsterkr:
see http://www.swig.org/Doc3.0/Customization.html `12.2 Object ownership and %newobject`

vigsterkr:
@karlnapf maybe @shubham808 could take a go on this. this is a blocker for 6.2.0

shubham808:
sure :) I am looking into it

karlnapf:
ah yes very good point. @shubham808 put it on your todo list. But dont do it this week pls, there is too many other things that need attention first. This can count as a "meta example patch" so just do it instead of porting an example next week or so

vigsterkr:
this is not needed anymore since shared_ptr...

PR #4604, opened 2019-04-10
factories are factories

vigsterkr:
they are always gonna love :)

### DIFF
--- a/examples/meta/src/regression/multiple_kernel_learning.sg
+++ b/examples/meta/src/regression/multiple_kernel_learning.sg
@@ -25,8 +25,8 @@ combined_kernel.init(features_train, features_train)
 #![create_combined_train]
 
 #![train_mkl]
-SVM binary_svm_solver = as_svm(machine("SVRLight"))
-Machine mkl = machine("MKLRegression", svm=binary_svm_solver, kernel=combined_kernel, labels=labels_train)
+Machine binary_svm_solver = machine("SVRLight")
+Machine mkl = machine("MKLRegression", svm=as_svm(binary_svm_solver), kernel=combined_kernel, labels=labels_train)
 mkl.train()
 #![train_mkl]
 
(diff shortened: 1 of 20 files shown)


vigsterkr:
this is an attempt to fix #4321

gf712:
erf, seems like it isn't that straightforward...

gf712:
create_object in `class_list.h` also increments the ref count

vigsterkr:
imo that refcount++ inn create_object should be dropped... but i'm gonna test it on donbot

vigsterkr:
this became a memory leaking cleanup branch...

examples/meta/src/evaluation/cross_validation_multiple_kernel_learning_weights_storage.sg
```diff
@@ -21,8 +21,8 @@ combined_kernel.add("kernel_array", k_3)
 #![create_kernels]
 
 #![create_classifier]
-SVM libsvm = as_svm(machine("LibSVM"))
-Machine svm = machine("MKLClassification", kernel=combined_kernel, interleaved_optimization=False, svm=libsvm)
+SVM libsvm = machine("LibSVM")
```
vigsterkr: for some reason if `as_svm` is being used here the ref counting is broken :S

examples/meta/src/regression/multiple_kernel_learning.sg
```diff
@@ -25,8 +25,8 @@ combined_kernel.init(features_train, features_train)
 #![create_combined_train]
 
 #![train_mkl]
-SVM binary_svm_solver = as_svm(machine("SVRLight"))
-Machine mkl = machine("MKLRegression", svm=binary_svm_solver, kernel=combined_kernel, labels=labels_train)
+SVM binary_svm_solver = machine("SVRLight")
```
vigsterkr: for some reason if `as_svm` is being used here the ref counting is broken :S

src/interfaces/swig/factory.i
```diff
@@ -1,3 +1,28 @@
+%newobject shogun::distance(std::string name);
```
gf712: doesn't it matter that it is `const std::string&`?

examples/meta/src/regression/multiple_kernel_learning.sg
gf712: broken in which way? too high or too low?

Issue #4178, opened 2018-02-16
Remove NArray from ruby typemap


karlnapf:
Ruby these days supports matrices natively, so no need to do this with the unmaintained narray.

This task is to re-write the typemap such that the new in-built matrices/vectors are used.

The meta example integration tests serve as a reference -- they can only pass if the typemap works correctly. The task is quite nontrivial, but there is a lot of existing code to guide the way, so still a good first issue

FaroukY:
@karlnapf I am interested in this issue. 

For array, we can replace NArray.sint($arguments) for example with Array.new($arguments){0}, is that okay?

I also had a look at Ruby's Matrix library (The one in std-lib now) and it looks like the matrix is actually immutable, so once it is initialized, it can't be assigned since the [] operator is private. (The author of Ruby said he will make it mutable in the future)

One workaround to this is to use:
A = Array.new( num_of_rows ){Array.new( num_of_columns) {0} }

Would that be okay to replace NMatrix.xxx()? And if so, I'd have to slightly edit the translate.py file so that it passes in addition to $numRows, $numColumns in addition to $arguments so that I can set the ruby.json as:

Array.new( $numRows ){Array.new( $numColumns ) {0} }

If that sounds good then let me know and I'll start working on it.

karlnapf:
Hey. 
Before you check out the meta examples, you will have to look into the typemaps.
I suggest you modify one of the generated meta example code listings by hand and then execute them by hand to see how you will have to modify the grammar. But the meta examples are not the problem here, it is easy to update that once the typemaps use the native ruby matrix lib.
Check  `swig_typemaps.i` in `interfaces/ruby`

songxujing:
@FaroukY 
Hello, I'm a shogun beginner and I want to use this tool to do multiple kernel learning. But when I use SVMLight, I get an error. Later I learned that I needed to compile manually, but I encountered a lot of errors in the compilation process. I hope you can help me and write a detailed compilation step for me. I am ubuntu16.04+python2.7+shogun

karlnapf:
@songxujing pls dont spam this thread which is about something else. Pls use the mailing list or open a new issue.

PR #4606, opened 2019-04-11
added list initialisation

gf712:
list initialisation to replace starting empty vector and the replace elements


examples/meta/generator/targets/octave.json
```diff
@@ -12,7 +12,6 @@
             "InitialSeperatorWhenArgs>0": false
         },
         "string": "$name = $arguments",
-        "CharVector": "$name = zeros(1, $arguments, 'char')",
```
gf712: this doesn't work in octave (4.4.1)

examples/meta/generator/targets/ruby.json
```diff
@@ -33,6 +33,18 @@
...
+        "ByteVector": "$name = NArray['byte', [$arguments])",
+        "WordVector": "$name = NArray['byte', [$arguments])",
+        "IntVector": "$name = NArray['sint', [$arguments])",
+        "LongIntVector": "$name = NArray['int', [$arguments])",
+        "ULongIntVector": "$name = NArray['int', [$arguments])",
+        "ShortRealVector": "$name = NArray['sfloat', [$arguments]))",
+        "RealVector": "$name = NArray['float', [$arguments])",
+        "LongRealVector": "$name = NArray['float', [$arguments])",
+        "ComplexVector": "NArray['scomplex', [$arguments])"
+    },
```
gf712: I tested this out with the most recent numo/NArray version

gf712: ah actually, what type are the zero constructors? I can't find it anywhere

examples/meta/generator/translate.py
```diff
@@ -458,6 +461,42 @@ def translateStatement(self, statement):
...
+        typeKey = list(typeDict.keys())[0]
+
+        self.variableTypes[nameString] = typeDict
+
+        if list(initialisation.keys())[0] == "ArgumentList":
+            template = Template(self.targetDict["Init"]["Construct"])
+            # custom SGType construction
+            if typeKey == "ShogunSGType"\
+                and init[0][typeKey] in self.targetDict["ListInit"]:
+                template = Template(self.targetDict["ListInit"][init[0][typeKey]])
```
gf712: This should probably be moved out of the condition. This substitution only works if there is an implementation of the list initialiser of the target language

examples/meta/generator/targets/csharp.json
```diff
@@ -32,6 +32,19 @@
...
     },
+    "ListInit": {
+        "BoolVector": "var $name = new bool[] {$arguments}",
+        "CharVector": "var $name = new char[] {$arguments}",
+        "ByteVector": "var $name = new byte[] {$arguments}",
+        "WordVector": "var $name = new ushort[] {$arguments}",
+        "IntVector": "var $name = new int[] {$arguments}",
+        "LongIntVector": "var $name = new long[] {$arguments}",
+        "ULongIntVector": "var $name = new ulongint[] {$arguments}",
+        "ShortRealVector": "var $name = new float[] {$arguments}",
```
gf712: need to fix c#, floats need to have a 'f' suffix, i.e. 1.0, should become 1.0f ...

examples/meta/generator/targets/java.json
```diff
@@ -38,6 +38,18 @@
         "RealMatrix": "DoubleMatrix $name = new DoubleMatrix($arguments)",
         "LongRealMatrix": "DoubleMatrix $name = new DoubleMatrix($arguments)"
     },
+    "ListInit": {
+        "BoolVector": "BoolVector $name = new BoolVector({$arguments})",
```
gf712: not sure how to do this one

karlnapf: Don’t know either, last resort would be to translate it to explicit assignment although that would suck

gf712:
> Like it! I guess now it’s done work to make it work with all the langs...
> 
> It would be cool if you could also try to make it work with R and ruby. While those meta examples are currently disabled in the build, they in principle work (r doesn’t have static calls, ruby doesn’t have overloaded methods/functions) since both support list inits. I think it would be sufficient to look at generates output and locally try it that works

OK, I can have a look into R. With ruby I am not sure how it would work, as the old NArray doesn't seem to have a list initialiser? The new NArray library does seem to have it though. But I am not sure which one shogun supports?

examples/meta/generator/targets/java.json
```diff
@@ -34,10 +34,22 @@
         "IntMatrix": "DoubleMatrix $name = new DoubleMatrix($arguments)",
         "LongIntMatrix": "DoubleMatrix $name = new DoubleMatrix($arguments)",
         "ULongIntMatrix": "DoubleMatrix $name = new DoubleMatrix($arguments)",
-        "ShortRealMatrix": "DoubleMatrix $name = new DoubleMatrix($arguments)",
+        "ShortRealMatrix": "FloatMatrix $name = new FloatMatrix($arguments)",
```
gf712: I am assuming this should be `FloatMatrix`?

karlnapf:
> I figured out how to make it work, but there is no support for bools in ruby's old NArray library, so it might be worth upgrading to numo-array

Shogun still uses the old one and we have an issue for upgrading [here](#4178). It is some work as the typemaps will have to be re-written. That is cumbersome as ruby is currently disabled due to #4177.
So don't loose your sleep over this I'd say. Or

examples/meta/generator/targets/java.json
karlnapf: not sure what the typemap exactly does...but I think our current Java bindings treat everything as double matrix. Might be worth cleaning that up at some point soon

gf712:
> Shogun still uses the old one and we have an issue for upgrading [here](#4178). It is some work as the typemaps will have to be re-written. That is cumbersome as ruby is currently disabled due to #4177.
> So don't loose your sleep over this I'd say. Or

I am thinking that maybe I skip boolean for now? There is still some lack of support for it in java, ruby and r (at least) so can address that separately?

examples/meta/generator/targets/ruby.json
karlnapf: what do you mean?

examples/meta/src/meta_api/list_init.sg
```diff
@@ -1,5 +1,3 @@
-BoolVector bool_vector([True, False])
```
karlnapf: ok!

karlnapf:
Soooo merge? :)

gf712:
Yup, seems fine to me. When there is a need I can add matrix initialisation

examples/meta/generator/targets/ruby.json
gf712: sorry, this was when I was confused as to how to construct with a list of args in ruby. was simpler than I thought! :D

PR #4608, opened 2019-04-12
Translator doesn't find method calls within method calls

gf712:
There seems to be an issue with the recursive lookup of keywords in the program dictionary during translation. The lookup stops when it finds a keyword, but should instead continue as there may be more nested keywords of interest.
Example:
```
Machine svm = machine("...", svm=as_svm(libsvm))
```
Old behaviour ignores the `as_svm` globalCall, which causes it not to add `as_svm` import.


gf712:
@vigsterkr this fixes the issue you mentioned on irc

PR #4611, opened 2019-04-16
Exception fix

gf712:
Fixes two things:
 - SGObject::get noexcept version calling a function that throws an error 
 - ShogunException c++11 compliance (which requires exceptions to be nothrow copy constructible)
see https://wiki.sei.cmu.edu/confluence/display/cplusplus/ERR60-CPP.+Exception+objects+must+be+nothrow+copy+constructible

### DIFF
--- a/src/shogun/lib/exception/ShogunException.h
+++ b/src/shogun/lib/exception/ShogunException.h
@@ -9,14 +9,13 @@
 
 #include <shogun/lib/config.h>
 
-#include <exception>
+#include <stdexcept>
 #include <string>
 
 namespace shogun
 {
 	/** @brief Class ShogunException defines an exception which is thrown
-	 * whenever an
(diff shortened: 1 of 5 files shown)


src/shogun/lib/exception/ShogunException.h
```diff
@@ -33,21 +32,11 @@ namespace shogun
...
-		virtual ~ShogunException();
-
-		virtual const char* what() const noexcept override;
+		const char* what() const noexcept override;
 
 	private:
-		/** exception string */
-		std::string msg;
+		/** exception object */
+		std::runtime_error msg;
```
vigsterkr: we could just inherit from `std::runtime_error` or? i mean that `ShogunException: public std::runtime_error` and then we are good :)

PR #4612, opened 2019-04-17
isfinite for msvc

gf712:
`std::isfinite` in msvc only accepts floats (otherwise there is an `ambiguous call to overloaded function` compiler error), so need to use `_isfinite` from cfloat to implicitly covert non-floats to floats.

### DIFF
--- a/src/shogun/util/converters.h
+++ b/src/shogun/util/converters.h
@@ -12,6 +12,15 @@
 
 #include <shogun/lib/common.h>
 
+#ifdef _MSC_VER
+#include <cfloat>
+// TODO: microsoft should really start supporting c++11
+#define IS_FINITE(x) _isfinite(x)
+#else
+#include <cmath>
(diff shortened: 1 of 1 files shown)


src/shogun/util/converters.h
```diff
@@ -12,6 +12,15 @@
 
 #include <shogun/lib/common.h>
 
+#ifdef _MSC_VER
+#include <cfloat>
+// TODO: microsoft should really start supporting c++11
+#define IS_FINITE(x) _isfinite(x)
+#else
+#include <cmath>
```
gf712: @vigsterkr this should solve your ruby issue

gf712: otherwise you might have to include header at the top of the swig file, like [here](https://github.com/robotology/idyntree/blob/devel/bindings/iDynTree.i#L29)

PR #4615, opened 2019-04-18
Fix observers for the interfaces.

geektoni:
* Prepend a C in front of concrete ParameterObserver classes;
* Fix printing inside ParameterObserverLogger;

### DIFF
--- a/src/shogun/lib/observers/ParameterObserverLogger.cpp
+++ b/src/shogun/lib/observers/ParameterObserverLogger.cpp
@@ -10,24 +10,24 @@
 
 using namespace shogun;
 
-ParameterObserverLogger::ParameterObserverLogger() {}
+CParameterObserverLogger::CParameterObserverLogger() {}
 
-ParameterObserverLogger::ParameterObserverLogger(std::vector<std::string> &parameters) : ParameterObserver(
+CParameterObserverLogger::CParameterObserverLogger(std::vector<std::string> &parameters) : ParameterObserver(
 		parameters) {}
 
-ParameterObserverLogger::~ParameterObserverLogger() {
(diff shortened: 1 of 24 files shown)


src/shogun/lib/observers/ParameterObserverLogger.cpp
```diff
@@ -47,6 +47,11 @@ void ParameterObserverLogger::on_next_impl(const TimedObservedValue &value) {
 		//		 name.c_str(), v.to_string().c_str());
 	};
 
-	sg_any_dispatch(any_val, sg_all_typemap, pf_n, pf_sgvector, pf_sgmatrix);
-
+	try {
```
geektoni: Not sure if this is the best way to handle it. The problems here happen when the observed value is not a "primitive" type, but something more (e.g. `CFeatures`). Instead of throwing an exception, I thought it would have been much better to catch it, show to the user a nice message and keep processing other emitted values. Thoughts on this @karlnapf?

karlnapf: Can’t we print it using the to string method in this case?

karlnapf: Could we have something like
```
template <class T>
std::string to_string(T v)
```
that is then tag dispatched for `CSGObject`, primitives, and `SGVector` & co?. We then could stringize anything.
Alternatively, why doesnt `Any` have a to_string method? I am not up to the latest here @lisitsyn @vigsterkr @gf712

geektoni: Solved by #4617.

geektoni:
This is complete now. I'll merge tomorrow @karlnapf.

PR #4618, opened 2019-04-24
Make LARS observable.

geektoni:
* Remove getters/setters from LARS;
* Make LARS observable;
* Add a meta example for the observable version;
* Add `get_long_int` to the meta language (aka `get<int64_t>()`);

### DIFF
--- a/examples/meta/generator/targets/cpp.json
+++ b/examples/meta/generator/targets/cpp.json
@@ -108,6 +108,7 @@
         "MethodCall": {
             "Default": "$object->$method($arguments)",
             "get_int": "$object->get<int32_t>($arguments)",
+            "get_long_int": "$object->get<int64_t>($arguments)",
             "get_int_vector": "$object->get<SGVector<int32_t>>($arguments)",
             "get_real": "$object->get<float64_t>($arguments)",
             "get_real_vector": "$object->get<SGVector<float64_t>>($arguments)",
--- a/examples/meta/generator/targets/cpp.json
+++ b/examples/meta/generator/targets/cpp.json
@@ -108,7 +108,7 @@
         "MethodCall": {
             "Default": "$object->$method($arguments)",
             "get_int": "$object->get<int32_t>($arguments)",
-            "get_long_int": "$object->get<int64_t>($arguments)",
+            "get_long": "$object->get<int64_t>($arguments)",
             "get_int_vector": "$object->get<SGVector<int32_t>>($arguments)",
             "get_real": "$object->get<float64_t>($arguments)",
             "get_real_vector": "$object->get<SGVector<float64_t>>($arguments)",
--- a/examples/meta/generator/targets/cpp.json
+++ b/examples/meta/generator/targets/cpp.json
@@ -116,7 +116,8 @@
             "get_int_string_list": "$object->get<SGStringList<int32_t>>($arguments)",
             "get_char_string_list": "$object->get<SGStringList<char>>($arguments)",
             "get_word_string_list": "$object->get<SGStringList<uint16_t>>($arguments)",
-            "get_option": "$object->get<std::string>($arguments)"
+            "get_option": "$object->get<std::string>($arguments)",
+            "get_string": "$object->get<std::string>($arguments)"
         },
         "StaticCall": "C$typeName::$method($arguments)",
         "GlobalCall": "$method($arguments)",
--- a/examples/meta/generator/targets/python.json
+++ b/examples/meta/generator/targets/python.json
@@ -71,6 +71,7 @@
         "MethodCall": {
             "Default": "$object.$method($arguments)",
             "get_int": "$object.get($arguments)",
+            "get_long_int": "$object.get($arguments)",
             "get_int_vector": "$object.get($arguments)",
             "get_real": "$object.get($arguments)",
             "get_real_vector": "$object.get($arguments)",
--- a/examples/meta/generator/targets/python.json
+++ b/examples/meta/generator/targets/python.json
@@ -71,7 +71,7 @@
         "MethodCall": {
             "Default": "$object.$method($arguments)",
             "get_int": "$object.get($arguments)",
-            "get_long_int": "$object.get($arguments)",
+            "get_long": "$object.get($arguments)",
             "get_int_vector": "$object.get($arguments)",
             "get_real": "$object.get($arguments)",
             "get_real_vector": "$object.get($arguments)",
--- a/examples/meta/generator/targets/python.json
+++ b/examples/meta/generator/targets/python.json
@@ -78,7 +78,8 @@
             "get_real_matrix": "$object.get($arguments)",
             "get_char_string_list": "$object.get($arguments)",
             "get_word_string_list": "$object.get($arguments)",
-            "get_option": "$object.get($arguments)"
+            "get_option": "$object.get($arguments)",
+            "get_string": "$object.get($arguments)"
         },
         "StaticCall": "$typeName.$method($arguments)",
         "GlobalCall": "$method($arguments$kwargs)",
--- a/examples/meta/src/observers/least_angle_regression.sg
+++ b/examples/meta/src/observers/least_angle_regression.sg
@@ -42,7 +42,7 @@ ObservedValue observation = logger.get_observation(num_observation-1)
 
 #![read_observation_information]
 string observation_name = observation.get_option("name")
-int observation_step = observation.get_long_int("step")
+int observation_step = observation.get_long("step")
 RealVector beta_path = observation.get_real_vector("beta_path")
 #![read_observation_information]
 
@@ -50,14 +50,6 @@ RealVector beta_path = observation.get_real_vector("beta_path")
 lars.unsubscribe(logger)
 #![unsubscribe_observer]
 
-#![evaluate_error]
-Evaluation eval = evaluation("MeanSquaredError")
-real mse = eval.evaluate(labels_predict, labels_test)
-#![evaluate_error]
-
 #[!extract_w]
 RealVector weights = lars.get_real_vector("w")
 #[!extract_w]
-
-# integration testing variables
-RealVector output = labels_predict.get_real_vector("labels")
--- a/examples/meta/src/observers/least_angle_regression.sg
+++ b/examples/meta/src/observers/least_angle_regression.sg
@@ -42,6 +42,7 @@ ObservedValue observation = logger.get_observation(num_observation-1)
 
 #![read_observation_information]
 string observation_name = observation.get_string("name")
+string observation_description = observation.get_description("beta_path")
 int observation_step = observation.get_long("step")
 RealVector beta_path = observation.get_real_vector("beta_path")
 #![read_observation_information]
--- a/examples/meta/src/observers/least_angle_regression.sg
+++ b/examples/meta/src/observers/least_angle_regression.sg
@@ -41,7 +41,7 @@ ObservedValue observation = logger.get_observation(num_observation-1)
 #![extract_last_observation]
 
 #![read_observation_information]
-string observation_name = observation.get_option("name")
+string observation_name = observation.get_string("name")
 int observation_step = observation.get_long("step")
 RealVector beta_path = observation.get_real_vector("beta_path")
 #![read_observation_information]
--- a/examples/meta/src/observers/least_angle_regression_observable.sg
+++ b/examples/meta/src/observers/least_angle_regression_observable.sg
@@ -1,0 +1,63 @@
+File f_feats_train = csv_file("../../data/regression_1d_linear_features_train.dat")
+File f_feats_test = csv_file("../../data/regression_1d_linear_features_test.dat")
+File f_labels_train = csv_file("../../data/regression_1d_linear_labels_train.dat")
+File f_labels_test = csv_file("../../data/regression_1d_linear_labels_test.dat")
+
+#![create_features]
+Features features_train = features(f_feats_train)
+Features features_test = features(f_feats_test)
+Labels labels_train = labels(f_labels_train)
+Labels labels_test = labels(f_labels_test)
+#![create_features]
+
+#![preprocess_features]
+Transformer SubMean = transformer("PruneVarSubMean")
+Transformer Normalize = transformer("NormOne")
+SubMean.fit(features_train)
+Features pruned_features_train = SubMean.transform(features_train)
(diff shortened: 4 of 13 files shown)


examples/meta/src/regression/least_angle_regression_observable.sg
```diff
@@ -0,0 +1,53 @@
...
+
+#![create_features]
+Features features_train = features(f_feats_train)
+Features features_test = features(f_feats_test)
+Labels labels_train = labels(f_labels_train)
+Labels labels_test = labels(f_labels_test)
+#![create_features]
+
+#![preprocess_features]
+Transformer SubMean = transformer("PruneVarSubMean")
```
karlnapf: minor: pls don't mix CamelCase and camel_case variable names

src/shogun/regression/LeastAngleRegression.cpp
```diff
@@ -341,18 +341,22 @@ bool CLeastAngleRegression::train_machine_templated(CDenseFeatures<ST>* data)
 			va.push_back((float64_t) m_beta_path_t[i][p]);			
 		}
 		m_beta_path.push_back(va);
+
+		SGVector<float64_t> obs_beta_path(va.size());
+		std::copy(va.begin(), va.end(), obs_beta_path.vector);
+		observe(i, "beta_path", "Beta path", obs_beta_path);
```
karlnapf: `va` has the new values concatenated in every iteration, i.e. the old values still stay the same. So if we observe them, the last observation is sufficient to have all elements in the path or?
Then the question is: why do we even need to observe them? Or am I lost (again)? ;)

geektoni: My idea was to expose the different betas value such to produce this kind of visualizations https://en.wikipedia.org/wiki/Least-angle_regression#/media/File:Larsdiabetes.png. One could still do it by using the method `get_w_for_var()`, though.

geektoni:
any ideas why this PR does not trigger the CI? :/ @vigsterkr @karlnapf

gf712:
Sometimes the CI gets confused, so you need to trigger it again. At least that is what has happened to me before

geektoni:
> Sometimes the CI gets confused, so you need to trigger it again. At least that is what has happened to me before

LoL I'll try to make it start then

examples/meta/src/regression/least_angle_regression_observable.sg
karlnapf: IMO we should have one or two examples that are specifically covering the observable framework. E.g. one for observing model parameters and one for observing special parameters (xvalidation). Show the API to retreive, list observable parameters, etc. Everything that is needed. So those examples would look quite different to the usual cookbook examples.
So it is fine to have this separate, but please put it under its own folder that is called something like "framework"

src/shogun/regression/LeastAngleRegression.cpp
karlnapf: Ah I got confused, the betas are stored in `m_beta_path`, not in `va` (which just contains the last one).

OK then.

Playing devils advocat here: in order to produce the visualization, all you would need is `m_beta_path` at the end (contains all betas). So storing those betas inside the observer doesnt really add anything.

karlnapf: Oh one thing: I dont really like the `std::copy` inside here. I think the copy should happen inside the observer that wants to store values. Less redundant. Is that possible?

geektoni: > Ah I got confused, the betas are stored in `m_beta_path`, not in `va` (which just contains the last one).
> 
> OK then.
> 
> Playing devils advocat here: in order to produce the visualization, all you would need is `m_beta_path` at the end (contains all betas). So storing those betas inside the observer doesnt really add anything.

That's true, but you need to wait until the training is finished to read the complete `m_beta_path`. I would like to give to the user the ability to write "interactive" models. That means that I could write a script which trains LARS and that shows me a plot with the LARS paths in real-time (while the machine is training). This could be achieved by providing an observer which executes custom code given by the user (like a lambda). However, if LARS does not emit the paths while it is fitted then this kind of things are not possible.

geektoni: > Oh one thing: I dont really like the `std::copy` inside here. I think the copy should happen inside the observer that wants to store values. Less redundant. Is that possible?

I had to do it since`va` is a `std::vector` and it is not recognized by the `sg_any_dispatch`. I agree when you say that it's redundant though. Let me see if I can find a better solution.

karlnapf: aaaah
@gf712 is there something we can maybe do about this?

examples/meta/src/regression/least_angle_regression_observable.sg
geektoni: Okay then, I'll try to move them to a separate section.

examples/meta/src/observers/least_angle_regression_observable.sg
```diff
@@ -0,0 +1,63 @@
...
+Labels labels_predict = lars.apply(normalized_features_test)
+#![train_and_apply]
+
+#![extract_last_observation]
+int num_observation = logger.get_int("num_observations")
+ObservedValue observation = logger.get_observation(num_observation-1)
+#![extract_last_observation]
+
+#![read_observation_information]
+string observation_name = observation.get_option("name")
```
geektoni: Currently, this is the way to get an `std::string` object from the meta language. Could it be worth to rename it to `get_string` instead?

examples/meta/generator/targets/cpp.json
```diff
@@ -108,6 +108,7 @@
         "MethodCall": {
             "Default": "$object->$method($arguments)",
             "get_int": "$object->get<int32_t>($arguments)",
+            "get_long_int": "$object->get<int64_t>($arguments)",
```
karlnapf: `get_long`

examples/meta/src/observers/least_angle_regression_observable.sg
```diff
@@ -0,0 +1,63 @@
+File f_feats_train = csv_file("../../data/regression_1d_linear_features_train.dat")
```
karlnapf: filename -> `least_angle_regression` (observable is already in the dir name)

examples/meta/src/observers/least_angle_regression_observable.sg
karlnapf: no, this is not for strings. Strings need to be added...but you have a template here....
Note that the c++ code that returns the string option might need to be patched to "catch" a potential string parameter before checking for the option

examples/meta/src/observers/least_angle_regression_observable.sg
```diff
@@ -0,0 +1,63 @@
...
+int observation_step = observation.get_long_int("step")
+RealVector beta_path = observation.get_real_vector("beta_path")
+#![read_observation_information]
+
+#![unsubscribe_observer]
+lars.unsubscribe(logger)
+#![unsubscribe_observer]
+
+#![evaluate_error]
+Evaluation eval = evaluation("MeanSquaredError")
```
karlnapf: this is not needed or?

src/shogun/regression/LeastAngleRegression.cpp
karlnapf: ok that seems addressed.
RE the real time, I like the idea and think it is good
But maybe write about this somewhere (meta example, blog post, etc)

examples/meta/src/observers/least_angle_regression_observable.sg
geektoni: Actually, the `get<std::string>` code works nicely already for extracting `std::string` registered parameters. The only question here would be if it makes sense, from the meta example point of view, to add a new method `get_string` which does the same thing as `get_option` but it is different in the sense that it signals that we are trying to get an `std::string` and not an option.

src/shogun/regression/LeastAngleRegression.cpp
geektoni: This code is not needed anymore since LARS works natively with `SGVector` now.

src/interfaces/swig/shogun.i
```diff
@@ -172,6 +172,13 @@ namespace shogun
 			$self->put(tag_t, value);
 	}
 
+	/* get method for string only available from the interfaces to disambiguate
+	 * it from get_option */
+	std::string get_string(const std::string& name) const
```
geektoni: @karlnapf I think this is the best solution for provide also a `get_string` method from the interfaces.

src/interfaces/swig/shogun.i
```diff
@@ -172,6 +172,13 @@ namespace shogun
 			$self->put(tag_t, value);
 	}
 
+	/* get method for string only available from the interfaces to disambiguate
```
karlnapf: no need for this comment... everything in here is only available from the interfaces :)

src/interfaces/swig/shogun.i
karlnapf: I think this is fine!

examples/meta/generator/targets/python.json
```diff
@@ -71,13 +71,15 @@
...
+            "get_long": "$object.get($arguments)",
             "get_int_vector": "$object.get($arguments)",
             "get_real": "$object.get($arguments)",
             "get_real_vector": "$object.get($arguments)",
             "get_real_matrix": "$object.get($arguments)",
             "get_char_string_list": "$object.get($arguments)",
             "get_word_string_list": "$object.get($arguments)",
-            "get_option": "$object.get($arguments)"
+            "get_option": "$object.get($arguments)",
+            "get_string": "$object.get($arguments)"
```
karlnapf: you will need to update the other .json files as well (see CI)

examples/meta/src/observers/least_angle_regression.sg
```diff
@@ -0,0 +1,55 @@
...
+Labels labels_predict = lars.apply(normalized_features_test)
+#![train_and_apply]
+
+#![extract_last_observation]
+int num_observation = logger.get_int("num_observations")
+ObservedValue observation = logger.get_observation(num_observation-1)
+#![extract_last_observation]
+
+#![read_observation_information]
+string observation_name = observation.get_string("name")
```
karlnapf: nice!
maybe also get description to show the users?

src/shogun/regression/LeastAngleRegression.cpp
```diff
@@ -105,7 +105,7 @@ void CLeastAngleRegression::plane_rot(ST x0, ST x1,
 template <typename ST, typename U>
 bool CLeastAngleRegression::train_machine_templated(CDenseFeatures<ST>* data)
 {
-	std::vector<std::vector<ST>> m_beta_path_t;		
+	std::vector<SGVector<ST>> m_beta_path_t;
```
karlnapf: what is the reason for this? just asking :)

geektoni: I have made this change to observe more easily the `m_beta_path` variable. Since it was an `std::vector` I was forced to convert it to `SGVector` before using `observe` and that was not nice.

examples/meta/generator/targets/python.json
geektoni: The CI's errors are unrelated to this (the python issue will be fixed here #4626). Actually, I have no idea why the build process fails on MacOS :/

src/shogun/regression/LeastAngleRegression.cpp
karlnapf: ok thx. I think that is fine. maybe even make more sense as the inner vectors are of fixed size

examples/meta/generator/targets/python.json
karlnapf: ah actually I was talking nonsense, since `get_string` is part of swig so all other langs will use exactly that call, nervermind :)

examples/meta/src/observers/least_angle_regression.sg
geektoni: Apparently, we left out the `description` from the `ObservedValue`. I don't remember exactly why we did that.

src/interfaces/swig/shogun.i
geektoni: I've updated the comment such to explain why we have `get_string` defined here and not as a `%template` directive.

geektoni:
@karlnapf  If there are no more comments I'll merge this later today (the Python error will be fixed by #4626).

examples/meta/src/observers/least_angle_regression.sg
karlnapf: ah yes, because we wanted to use the description of the parameter in the original class itself. Maybe we can add that here?

src/shogun/base/SGObject.h
```diff
@@ -287,6 +287,13 @@ class CSGObject
 	 */
 	Version* get_global_version();
 
+	/**
+	 * Return the description of a registered parameter given its name
+	 * @param name parameter's name
+	 * @return description of the parameter as a string
+	 */
+	std::string get_description(const std::string& name) const;
```
geektoni: @karlnapf Since there was no way to extract the description of a parameter easily, I moved this method from the SWIG interface to the actual `SGObject.h` (which was originally added by @gf712). I guess it makes sense since it is wayy much simpler now. For instance, if we look at C++:
```c++
object.get_parameters().find(name)->second.get()->get_properties()->get_description()
```
versus:
```c++
object.get_description(name)
```

src/shogun/base/SGObject.cpp
```diff
@@ -547,6 +547,17 @@ void CSGObject::init()
 	watch_method("num_subscriptions", &CSGObject::get_num_subscriptions);
 }
 
+std::string CSGObject::get_description(const std::string &name) const
+{
+	auto it = this->get_params().find(name);
```
gf712: why `this->`?

geektoni: I usually use it to show that `get_params` is a member function and not a method declared elsewhere :) it is more a personal choice rather than a standard (shouldn't be impacting the performance or?).

gf712: OK! No it doesn't affect performance, I was just wondering :)

src/shogun/base/SGObject.h
karlnapf: ok.... we kinda saw it coming :D

PR #4619, opened 2019-04-24
💥 💃 

vigsterkr:
because it's 2019


vigsterkr:
this is still WIP... so dont expect things to work :)

src/shogun/base/DynArray.h
```diff
@@ -446,16 +446,16 @@ template <class T> class DynArray
...
 			for (index_t i=0; i<=current_num_elements-1; ++i)
 				std::swap(
-					array[i], 
+					array[i],
 					array[CMath::random(i, current_num_elements - 1)]);
 		}
 
 		/** randomizes the array with external random state */
-		void shuffle(CRandom * rand)
+		void shuffle(std::shared_ptr<CRandom> rand)
```
gf712: @vigsterkr these things should be `const std::shared_ptr<T>&` IMO, otherwise you are incrementing the ref count and then decreasing it at the end of the function

vigsterkr: problem? :)

vigsterkr: i mean that in this case this should actually work correctly as there's gonna be somebody on the end who holds the ref for this obj.... until that object gets destructed.

lisitsyn:
Does it work in the end?

vigsterkr:
@lisitsyn yeah managed to get swig and the library work.... just unit tests and some other parts needs gradual fixing... but mostly it work already

vigsterkr:
ah yeah gpl part is not at all touched.... :)

vigsterkr:
@karlnapf @lisitsyn @gf712 ok so this is it... we currently have some problems, including `put`, but more or less this is going to be it. have some peep on it when u got some time :)

src/shogun/base/SGObject.h
```diff
@@ -542,34 +474,7 @@ class CSGObject
...
-
-	/** Typed appender for an object class parameter of a Shogun base class
-	* type,
-	* identified by a name.
-	*
-	* @param name name of the parameter
-	* @param value value of the parameter
-	*/
-	template <class T, class = typename std::enable_if_t<is_sg_base<T>::value>>
-	void add(const std::string& name, Some<T> value)
```
karlnapf: What about those? No more add?

vigsterkr: mmm i think that's one of the accidental deletion of rebase... :S will look into it as soon as i've got other blockers fixed thnx for reporting

karlnapf: It’s tested in meta examples btw.... mkl and put-get-add

karlnapf:
Yo!

I have some changes in my label mapping PR that will cause more conflicts. Shall I wait to rebase against this one, or will it take more time to be finished?

vigsterkr:
@karlnapf @lisitsyn @gf712 so this is it. mostly things pass, there's a lot of thing that is gone. i know it's a big patch... only two things are failing:
 * MSVC complaining about a type trait
 * when installing the R interface R runs out of memory when running `preparing package for lazy loading`

gf712:
So things that have been removed are:
 - Lua support (currently no shared_ptr typemap in swig)
 - Drop swig 3 support (in favour of 4)
 - Old parameter framework
 - Model selection module (depended too much on old Parameter framework)
 - Old implementations of things that are now in the standard, like Set and CMap

.ci/ci.yml
```diff
@@ -89,11 +89,6 @@ jobs:
         CXX: "clang++"
         cmakeOptions: '$(commonSWIGCMakeFlags) -DINTERFACE_R=ON'
         interfaceName: 'r'
-      lua:
```
vigsterkr, line 92: bye bye lua for now...

src/shogun/lib/any.h
```diff
@@ -1274,15 +1290,30 @@ namespace shogun
 	inline void register_casts()
 	{
 		using Derived = std::remove_pointer_t<T>;
-		if constexpr (std::is_base_of_v<CSGObject, Derived>)
+		if constexpr (std::is_base_of_v<SGObject, Derived>)
```
vigsterkr, line 1293: @lisitsyn do we wanna have support for SGO* ? as you can see the same thing is copypasted below for shared_ptr<SGO-like>

.ci/ci.yml
karlnapf: didnt work anyways

karlnapf:
> @karlnapf @lisitsyn @gf712 so this is it. mostly things pass, there's a lot of thing that is gone. i know it's a big patch... only two things are failing:
> 
>     * MSVC complaining about a type trait
> 
>     * when installing the R interface R runs out of memory when running `preparing package for lazy loading`

It is quite hard to read through all of the changes, but I think it is totally worth dropping some of the legacy stuff gil mentioned above in favour of this new pointer system. Either we don't use/need those old things, they didnt work (lua), or we are thinking about rewriting them anyways (modsel). So green light from my side regarding that list. I will try to scroll through the code over the next days but but happy for you to push the button.... Big changes always come with collateral damage... But that is even minimal here.... :) Awesome work!

karlnapf:
And I guess if something is terribly wrong, we will figure it out.... it's not like we will want to revert going back to raw pointers ....

vigsterkr:
so now the only remaining issue is R interface, that seems more and more that it actually not working :( so the question now is whether we wanna drop the R interface for the time being (and hope one day the issue will be fixed in SWIG) or don't merge this branch until there's no solution for R....

vigsterkr:
currently this issue is the blocker for the R interface: https://github.com/swig/swig/issues/1660

karlnapf:
> so now the only remaining issue is R interface, that seems more and more that it actually not working :( so the question now is whether we wanna drop the R interface for the time being (and hope one day the issue will be fixed in SWIG) or don't merge this branch until there's no solution for R....

Since R is deactivated since a long time (although after last release iirc, but definitely since we moved to tags in develop due to the static call issue), I would think it is OK to drop the interface for now as a known issue and then try to recover it. But at least then we can move on and this branch doesnt become stale?

vigsterkr:
@karlnapf actually R is now working in develop thnx to the work of @theartful :) hence i'm a bit reluctant to disable it :(

vigsterkr:
it's really cool that actually all the meta tests are passing for R interface: https://dev.azure.com/shogunml/shogun/_build/results?buildId=2364&view=logs&jobId=207c801f-b63c-5ea7-d82b-e9be88ff7c06&taskId=4b28f5e4-7c4b-5b88-2a7d-4649b8b72bdd&lineStart=4&lineEnd=4&colStart=1&colEnd=12

karlnapf:
Pretty cool indeed I had forgotten, should keep it in that state I guess.

I guess we can see and wait what happens with this bug for a bit and then discuss again?

ghost:
This is wonderful, it's great to see this finally finished! I think we should go for it, and leave R until swig solves the problem, if it was a problem on their part.

karlnapf:
Discussed with Gil today and it seems that if we (temporarily) hide the crtp and Mixin types from r then it works. Also not that we want to remove all these types anyways and use factories...

vigsterkr:
Noup it does not work. The mentioned bug filed under swig project shows that even though if we hide those things it still fails to run... casting in R interface seems to be not supported with sharedptrs... but see the simple example in that bug and you’ll understand whats the actual problem...

vigsterkr:
so this is done on my end. R needed to be disabled as it cannot handle shared_ptr casting atm.
if anybody has any objections plz raise it today (next 24h) as i'd like to move on and merge this one before develop diverges again a lot and makes rebasing a job that takes more than a day

karlnapf:
no objections

vigsterkr:
Ooh
You can dance
You can jive
Having the time of your life
Ooh, see that girl
Watch that scene
Dig in the dancing queen

PR #4621, opened 2019-04-26
[WIP] Refactor SGObject using mixins 

theartful:
This is still work in progress for a mixins draft, and not intended to be merged. I will continue working on it next Wednesday.

- [x] Turn SGObject into one big mixin
- [x] Divide the common interface into a mixin (get_name(), global variables (io, version, ...))
- [ ] Divide the any parameter/observable interface into a mixin
- [ ] Use the parameter mixin for ObservedValue
- [ ] Investigate what to do next!


src/shogun/base/SGObject.h
```diff
@@ -25,6 +25,7 @@
 #include <shogun/lib/config.h>
 #include <shogun/lib/exception/ShogunException.h>
 #include <shogun/lib/tag.h>
+#include <shogun/lib/mixins.h>
```
vigsterkr: i would move this to util

src/shogun/base/SGObject.cpp
```diff
@@ -691,19 +651,20 @@ void CSGObject::build_gradient_parameter_dictionary(CMap<TParameter*, CSGObject*
...
 }
 
-CSGObject* CSGObject::clone() const
+template <typename M>
+typename CSGObjectBase<M>::Derived* CSGObjectBase<M>::clone() const
 {
 	SG_DEBUG("Starting to clone %s at %p.\n", get_name(), this);
 	SG_DEBUG("Constructing an empty instance of %s.\n", get_name());
-	CSGObject* clone = create_empty();
+	Derived* clone = create_empty();
```
vigsterkr: you can auto this (unless you used a regex/sed :P)

src/shogun/base/SGObject.cpp
```diff
@@ -1083,14 +1058,16 @@ bool CSGObject::equals(const CSGObject* other) const
 	return true;
 }
 
-CSGObject* CSGObject::create_empty() const
+template <typename M>
+typename CSGObjectBase<M>::Derived* CSGObjectBase<M>::create_empty() const
 {
-	CSGObject* object = create(this->get_name(), this->m_generic);
+	Derived* object = create(this->get_name(), this->m_generic);
```
vigsterkr: auto as above... further refacts will be helped by this

src/shogun/base/SGObject.h
```diff
@@ -452,16 +458,16 @@ class CSGObject
...
 	T* get(const std::string& name, index_t index, std::nothrow_t) const
 	{
-		CSGObject* result = nullptr;
+		Derived* result = nullptr;
 
 		auto get_lambda = [&index, &result](auto& array) {
 			result = array.at(index);
 		};
-		if (sgo_details::dispatch_array_type<T>(this, name, get_lambda))
+		if (sgo_details::dispatch_array_type<T>((Derived*) this, name, get_lambda))
```
vigsterkr: could you do explicit cast with c++ styles? like static_cast<T>()... again later refactor of code will be made easier this way

karlnapf:
Great initiative, thanks for that! I will have a closer look over the next few days!

src/shogun/base/SGObject.cpp
theartful: you caught me :"D

vigsterkr: k then for the time being leave it... will change it with a libtooling script

theartful: don't worry about it, i'll change them, they are only a couple of instances

karlnapf:
This is quite a big PR now, hard to oversee. Would you mind explaining the high level changes you are doing here? This would save us (me at least :D) some time digging through all the code. What shall we look at first design wise?

theartful:
Ideally there will be a bunch of mixins, and a family of SGObject classes. Each SGObject is a composition of these mixins, and nothing else. Each mixin can't assume anything about the SGObject containing it but for the other mixins it depends on. To access a function of another mixin you have get a reference of it using the "mutate" function. A mixin can also access the type of the containing SGObject using the "derived_t" type function inside the mutator.
This will require that each function accepting an SGObject as an argument or returning an SGObject has to be templated, and guarded with checks that it contains the mixins the function needs. A problem comes in mind with the class_list 'create' function (which creates an instance of any class and returns an SGObject).
What I did was:
1. Get rid of "set_generic" specialized template functions inside SGObject, since you can't specialize template functions inside template classes without first specializing the class.
2. Create two mixins to test things out: HouseKeeper (horrible name) which contains reference counting and global variables (io/version/parallel), and other general purpose functions. And SGObjectBase which contains everything else.
3. Now SGObject is a composition of these two.

karlnapf:
Before you press ahead with this type of design, make sure to look at some of the SWIG interfacing stuff. 

In particular, see how SWIG works with templates (messy). There is a few implications, for example if you make the parameter framework a mixin (put/get/etc), then the Shogun base class should have it added (and so all Shogun classes have a common put/get/add). If we don't do that, then the SWIG interface will have to be re-generated for every class where the mixin is used. While it is fine to do that in certain exceptions (we had the example of ObservedValue, where we would like to access "get" without it being a SGObject), we cannot do it for every class as the SWIG compile time will explode.

Also look at a few cases where we actually exploit the fact that all classes in Shogun inherit from SGObject, (class_list for example, used in factories).

karlnapf:
IDK....phew this will get pretty mesy.
What about for now we start with something less general and make it work, then make it more general.
Like for example moving all the ParameterMap stuff (self, put/get) into a separate class that SGObject inherits from, and then go from there. That was the main motivation:

* to split the code in SGObject into multiple files
* to be able to have classes that have the parameter framework without the need to be an SGObject

vigsterkr:
@theartful which swig have you used? in case you haven't tried it with 4.0 plz do so... it was released very recently.... although idk if variadic template support has been added, but there were a lot of improvements... if it works, it's fine to depend on 4.x

theartful:
@vigsterkr There are a couple of problems:
1. separating the get/add mixin is weird. The untyped get functions
```C++
CSGObject* get(const std::string& name) const noexcept(false);
CSGObject* get(const std::string& name, std::nothrow_t) const noexcept;
```
return CSGObject, which depends on the `get`/`add` mixin, which in turn depends on CSGObject.
3. relating to 1. The dispatching code in the get/add mixin depends on CSGObject
4. there is no clear division between `put`/`get`/`add` and `observe`/`watch_param` mixins. They both depend on each other (since `put` uses `observe`) .
5. separating ObservedValue in a different file is difficult. The circular dependency thing, but now with the mixins.

karlnapf:
> @vigsterkr @karlnapf There are a couple of problems:
> 
> 1. separating the get/add mixin is weird. The untyped get functions
> 
> ```c++
> CSGObject* get(const std::string& name) const noexcept(false);
> CSGObject* get(const std::string& name, std::nothrow_t) const noexcept;
> ```
> 
> return CSGObject, which depends on the `get`/`add` mixin, which in turn depends on CSGObject.

Yes these are problematic anyways. We might add them to the swig c++ code?

> 3. relating to 1. The dispatching code in the get/add mixin depends on CSGObject

Yes, that code is another candidate to maybe be removed. We could instead offer something in swig like `get_kernel` which maps to c++ `get<CKernel> so we would have
```
k = m.get_kernel("kernel")
k.init(....)
```
instead of (what we currently have)
```
k = as_kernel(m.get("kernel"))
k.init(....)
```
It has the disadvantage that we need swig get- functions for all base types. But is has the advantage that there is no more dynamic dispatching in c++. And more importantly, it doesn't really make sense as in most use cases, there will be a cast afterwards (see lines above)
You could try to do this in a separate sub-PR

> 4. there is no clear division between `put`/`get`/`add` and `observe`/`watch_param` mixins. They both depend on each other (since `put` uses `observe`) .

Ah that is true. I was initially thinking that the observable stuff simply depends on put/get, but I then forgot that put actually calls observe. I guess the current design means that we need to group them. But I wonder whether that is the best idea ...

> 5. separating ObservedValue in a different file is difficult. The circular dependency thing, but now with the mixins.

I think this is solved once 3. is sorted?

PR #4625, opened 2019-05-02
Fix overflow issue with large matrices (fixes #4595)

### LINKED ISSUE #4595: MMD unbiased estimate issue
I run the following code which computes MMD unbiased estimate multiple times. Since both the distributions are Normal(0,1) and the estimate is unbiased, I expect the average output to be 0. However, it always prints the exact same number (-0.6577018139068969)

```
import numpy as np
from torch.distributions.normal import Normal
import shogun as sg

for _ in range(100):
    X = Normal(0,1).sample((7000,)).numpy().reshape(-1,1)
    Y = Normal(0,1).sample((7000,)).numpy().reshape(-1,1)
    mmd = sg.QuadraticTimeMMD()
    mmd.set_p(sg.RealFeatures(X.T.astype(np.float64)))
    mmd.set_q(sg.RealFeatures(Y.T.astype(np.float64)))
    mmd.set_kernel(sg.GaussianKernel(32, 1))
    mmd.set_statistic_type(sg.ST_UNBIASED_FULL)
    stat = mmd.compute_statistic()
    print('stat = ', stat)
```

karlnapf:


src/shogun/statistical_testing/internals/mmd/ComputeMMD.h
```diff
@@ -108,13 +108,14 @@ struct ComputeMMD
 		const BlockXt& b_xy=map.block(m_n_x, 0, m_n_y, m_n_x);
 
 		terms_t terms;
-		terms.diag[0]=b_x.diagonal().sum();
-		terms.diag[1]=b_y.diagonal().sum();
-		terms.diag[2]=b_xy.diagonal().sum();
+		// cast to 64 bit to avoid overflows
```
karlnapf: kernel matrices are often stored as `float32_t`, so the `.sum()` calls over the whole matrix here overflow

src/shogun/statistical_testing/internals/mmd/ComputeMMD.h
```diff
@@ -108,13 +108,14 @@ struct ComputeMMD
...
+		terms.diag[0]=b_x.diagonal(). template cast<float64_t>().sum();
+		terms.diag[1]=b_y.diagonal(). template cast<float64_t>().sum();
+		terms.diag[2]=b_xy.diagonal(). template cast<float64_t>().sum();
 
-		terms.term[0]=(b_x.sum()-terms.diag[0])/2+terms.diag[0];
-		terms.term[1]=(b_y.sum()-terms.diag[1])/2+terms.diag[1];
-		terms.term[2]=b_xy.sum();
+		terms.term[0]=(b_x. template cast<float64_t>().sum()-terms.diag[0])/2+terms.diag[0];
+		terms.term[1]=(b_y. template cast<float64_t>().sum()-terms.diag[1])/2+terms.diag[1];
+		terms.term[2]=b_xy. template cast<float64_t>().sum();
```
gf712: so is this happening without copying the matrix?

karlnapf: idk :D

karlnapf: since I don't store the result in a variable, `eval` should not be called, and the `sum` should cast element by element in the loop. You have an idea how to check this easily/quickly?

gf712: I am not sure, from the docs it seems like you can force lazy evaluation by using `.noalias` when you store the temporary values (see https://eigen.tuxfamily.org/dox/TopicLazyEvaluation.html)

karlnapf: Ah yes I remember reading that page. I understand it like this: These docs also say that everything is lazy apart from 3 cases which are all different to this one. So merging.

PR #4626, opened 2019-05-02
Fix observe() method when called for SGVector/SGMatrix.

geektoni:
* Fix `observe()` method when called for `SGVector`/`SGMatrix`. The given `SGVector/SGMatrix` was not cloned correctly (since copying them just copy the pointer to the raw data and it does not make a clone of the data itself).

<del>I moved some useful type traits from `type_case.h` to `base_types.h` since I did not want to pollute too much `SGObject.h` with unnecessary headers. </del>

Any comments on this change? @karlnapf @vigsterkr @gf712

### DIFF
--- a/src/shogun/base/SGObject.h
+++ b/src/shogun/base/SGObject.h
@@ -30,6 +30,7 @@
 #include <unordered_map>
 #include <utility>
 #include <vector>
+#include <shogun/lib/type_case.h>
 
 /** \namespace shogun
  * @brief all of classes and functions are contained in the shogun namespace
--- a/src/shogun/base/SGObject.h
+++ b/src/shogun/base/SGObject.h
@@ -30,7 +30,6 @@
 #include <unordered_map>
 #include <utility>
 #include <vector>
-#include <shogun/lib/type_case.h>
 
 /** \namespace shogun
  * @brief all of classes and functions are contained in the shogun namespace
--- a/src/shogun/base/SGObject.h
+++ b/src/shogun/base/SGObject.h
@@ -30,6 +30,7 @@
 #include <unordered_map>
 #include <utility>
 #include <vector>
+#include <shogun/lib/type_case.h>
 
 /** \namespace shogun
  * @brief all of classes and functions are contained in the shogun namespace
@@ -1011,7 +1012,16 @@ protected:
 	 * @param step step
 	 * @param name tag's name
 	 */
-	template <class T>
+	template<class T,
+			typename std::enable_if_t<
+				!type_internal::is_sg_vector<T>::value &&
+				!type_internal::is_sg_matrix<T>::value>* = nullptr>
+	void observe(const int64_t step, const std::string& name) const;
+
+	template<class T,
+			typename std::enable_if_t<
+					type_internal::is_sg_vector<T>::value ||
+					type_internal::is_sg_matrix<T>::value>* = nullptr>
 	void observe(const int64_t step, const std::string& name) const;
 
 	/**
@@ -1313,7 +1323,10 @@ void CSGObject::observe(
 	this->observe(obs);
 }
 
-template <class T>
+	template<class T,
+			typename std::enable_if_t<
+					!type_internal::is_sg_vector<T>::value &&
+					!type_internal::is_sg_matrix<T>::value>* = nullptr>
 void CSGObject::observe(const int64_t step, const std::string& name) const
 {
 	auto param = this->get_parameter(BaseTag(name));
@@ -1321,5 +1334,19 @@ void CSGObject::observe(const int64_t step, const std::string& name) const
 		step, name, any_cast<T>(param.get_value()), param.get_properties());
 	this->observe(obs);
 }
+
+template<class T,
+		typename std::enable_if_t<
+				type_internal::is_sg_vector<T>::value ||
+				type_internal::is_sg_matrix<T>::value>* = nullptr>
+void CSGObject::observe(const int64_t step, const std::string& name) const
+{
+	auto param = this->get_parameter(BaseTag(name));
+	auto obs = some<ObservedValueTemplated<T>>(
+			step, name, any_cast<T>(param.get_value()).clone(), param.get_properties());
+	this->observe(obs);
+}
+
+
 }
 #endif // __SGOBJECT_H__
--- a/src/shogun/base/SGObject.h
+++ b/src/shogun/base/SGObject.h
@@ -30,7 +30,6 @@
 #include <unordered_map>
 #include <utility>
 #include <vector>
-#include <shogun/lib/type_case.h>
 
 /** \namespace shogun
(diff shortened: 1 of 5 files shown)


src/shogun/base/base_types.h
```diff
@@ -68,6 +70,121 @@ namespace shogun
...
+		T_SGVECTOR_FLOATMAX = 18,
+		T_SGVECTOR_INT32 = 19,
+		T_SGVECTOR_INT64 = 20,
+		T_SGMATRIX_FLOAT32 = 21,
+		T_SGMATRIX_FLOAT64 = 22,
+		T_SGMATRIX_FLOATMAX = 23,
+		T_SGMATRIX_INT32 = 24,
+		T_SGMATRIX_INT64 = 25,
+		T_UNDEFINED = 26
+	};
```
gf712: Do you think it would be worth adding all the combinations of containers and primitive types, or some macro that would do that? Because right now stuff like `SGVector<bool>` isn't included (whether or not we actually use it is another question...)

src/shogun/base/base_types.h
```diff
@@ -68,6 +70,121 @@ namespace shogun
...
+
+		template <typename T>
+		struct is_sg_vector : public std::false_type
+		{
+		};
+
+		template <typename T>
+		struct is_sg_matrix : public std::false_type
+		{
+		};
```
gf712: Remember that these structs do not check all possible `T`/`SGVector<T>`/`SGMatrix<T>`, this is to check if the type is registered at runtime to then use the dispatching. To check if a type is `SGVector<T>` you should do something like:
```
template <typename T>
struct is_sg_vector
{
    using type = std::false_type;
};
template <typename T>
struct is_sg_vector<SGVector<T>>
{
  using type = std::true_type;
};
```

gf712: (That's why these things were hidden in their own namespace)

src/shogun/base/base_types.h
geektoni: I guess it would make sense if we were to use it somewhere in Shogun. For instance, if we find a machine which needs to emit `SGVector<bool>` then we will add it.

src/shogun/base/base_types.h
geektoni: Argh you are indeed right. Those need to be left in their namespace. In the end, I just need to check if it is an SGVector/SGMatrix (an not if they match also with a type `T`).

src/shogun/base/SGObject.h
```diff
@@ -1306,13 +1315,30 @@ void CSGObject::observe(
...
 {
 	auto param = this->get_parameter(BaseTag(name));
 	auto obs = some<ObservedValueTemplated<T>>(
 		step, name, any_cast<T>(param.get_value()), param.get_properties());
 	this->observe(obs);
 }
+
+template<class T,
+		typename std::enable_if_t<
+				is_sg_vector<T>::value ||
```
karlnapf: what about std::vector?

src/shogun/base/SGObject.h
```diff
@@ -1306,13 +1315,30 @@ void CSGObject::observe(
...
+				is_sg_matrix<T>::value>*>
+void CSGObject::observe(const int64_t step, const std::string& name) const
+{
+	auto param = this->get_parameter(BaseTag(name));
+	auto obs = some<ObservedValueTemplated<T>>(
```
karlnapf: I also wonder: couldn't we do this inside the `ObservedValueTemplated` class?

src/shogun/base/SGObject.h
karlnapf: btw this is very similar to the problem we have in `any.h` with `clone` and `equals`. There we do the following: we check (in order) whether a type has `->clone`, then `.clone`, and if not we use the copy ctor. We also cover cases such as raw arrays and `std::vector`. I think an approach like this would be better here as well, as otherwise we will start patching and patching the template signature here, causing a change in `SGObject` every time....

geektoni: Do you mean something like this https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/lib/any.h#L276?

geektoni: > what about std::vector?

`std::vector` has a copy constructor so there is no problem. The issue was with `SGVector` since its copy constructor does not copy also the memory, but it copies only the reference to the memory (hence we have multiple `SGVector`s that point to the same location).

src/shogun/base/SGObject.h
geektoni: I would prefer to keep a separate method to construct an `ObservedValueTemplated` object from a `Tag`.

geektoni: If you are talking about a method to construct a new `ObservedValueTemplated` :)

src/shogun/base/SGObject.h
geektoni: @karlnapf have a look at this one: https://github.com/geektoni/shogun/commit/0a9a92e8012fc7352e4d757cb85f0681145854fa. This is basically the `any.h` code, but it should do exactly what we are looking for. Still WIP though. Feel free to criticize it :)

karlnapf: Cool! :) That seems ok to me. The only critique I have is now we have the code twice within the codebase.... maybe you can make use of that in `any.h` too and delete it there? Or instead use the implementation from the `any_detail` namespace (since any is included in your file)
If not, also the filename can maybe just be `clone.h`

geektoni: The `any.h` implementation uses some casts to `void*` and it also uses a `void** storage` argument to store the cloned value. These are things which are not exactly needed by our use case, so I thought it would have been better to have a separate `clone` which can be used from Shogun (and that is not `Any` specific).

karlnapf: I see, seems reasonable ... :)

geektoni:
@karlnapf any final comments before I merge this? :)

tests/unit/base/MockObject.h
```diff
@@ -316,6 +316,13 @@ namespace shogun
 			watch_method("some_method", &CMockObject::some_method);
 		}
 
+		virtual CSGObject* create_empty() const
```
karlnapf: curios: why did you add this here

karlnapf:
good!

tests/unit/base/MockObject.h
geektoni: Basically, since `observe()` emit a clone of the given value, we happens to call `MockObject->clone()`. However, the standard `SGObject` implementation of `clone()` assumes that there is a corresponding `create()` method, for that element, inside `class_list.h`. This is not true for `MockObject` since it is used only for testing purpose. Therefore, I needed to override `create_empty` such to return a new instance without relying on the `class_list.h` methods.

karlnapf: ah yep, I remember that stuff now. thanks!

PR #4627, opened 2019-05-02
arff deserialiser

gf712:
Simple ARFF deserialiser. See the spec [here](https://waikato.github.io/weka-wiki/arff_stable/).

@karlnapf shall I add a ARFF file as an example and a unittest/meta example?


src/shogun/io/ARFFFile.cpp
```diff
@@ -0,0 +1,263 @@
...
+	{
+		elem.erase(
+		    std::remove_if(
+		        elem.begin(), elem.end(),
+		        [](auto& v) { return v == ',' || v == '{' || v == '}'; }),
+		    elem.end());
+	}
+	for (auto iter = line.begin(); iter != line.end(); ++iter)
+	{
+		if (iter->front() == '\'' || iter->front() == '\"')
```
gf712: @vigsterkr does stuff like this work independent of platform?

src/shogun/io/ARFFFile.cpp
```diff
@@ -0,0 +1,263 @@
...
+	process_chunk(read_relation, check_relation, true);
+
+	auto read_attributes = [this]() {
+		if (string_to_lower(m_current_line.substr(
+		        0, strlen(m_attribute_string))) == m_attribute_string)
+		{
+			std::vector<std::string> elems;
+			auto innner_string =
+			    m_current_line.substr(strlen(m_attribute_string));
+			split(innner_string, " ,\t\r\f\v", std::back_inserter(elems));
```
gf712: @vigsterkr and this?

gf712:
@karlnapf how does shogun handle missing data values? It is possible to have NaN values ("?" in arff files) ?

karlnapf:
> @karlnapf shall I add a ARFF file as an example and a unittest/meta example?

Yes you can add it under the toy data in `data/toy` for examples.
For unit test, I would rather not base it on a file ... although there seems to be no other sensible way or?

karlnapf:
> @karlnapf how does shogun handle missing data values? It is possible to have NaN values ("?" in arff files) ?

idk to be honest :D we might have to think/work a bit for that

karlnapf:
Nice!

gf712:
> For unit test, I would rather not base it on a file ... although there seems to be no other sensible way or?

For unittest I could try and use strings? I can then feed the stream directly, rather than a file. Would just have to refactor the class a bit though by adding a constructor for streams rather than file names

gf712:
So datetime somewhat works, but might have to add strptime definition for the windows build? (see https://stackoverflow.com/a/321877). The conversion to the c++ convention of datetime format needs to be tested more too. I convert dates to unix timestamps btw, because it is the only representation that makes sense to me that can be stored in a numeric matrix

gf712:
@karlnapf I am not sure how to do the parsing of strings though, for that we would almost need a `SGMatrix<variant>`

karlnapf:
> So datetime somewhat works, but might have to add strptime definition for the windows build? (see https://stackoverflow.com/a/321877). The conversion to the c++ convention of datetime format needs to be tested more too. I convert dates to unix timestamps btw, because it is the only representation that makes sense to me that can be stored in a numeric matrix

Great. Ill check it out

karlnapf:
> @karlnapf I am not sure how to do the parsing of strings though, for that we would almost need a `SGMatrix<variant>`

Uh that will be tricky with parameter framework. Though we have started using variant in the pipeline...although that causes all sort of headaches (needs an override clone/equals, but no idea how to make put/get work atm)

gf712:
@karlnapf @vigsterkr so the return type is now `CCombinedFeatures`, and I use `std::vector<variant>` internally.
Current features:
1. handle numeric data with CDenseFeatures
2. handle dates (excluding timezone name parsing because it requires libtz from the date library to be compiled into a library)
3. handle strings with StringList and then StringFeatures

The missing data can be ignored for now I think, and just be skipped by the parser, and then one day we can add imputers. 
The relational attributes I still need to look into. 
And then I'll add a meta example too and that's it I think. @karlnapf can you think of an interesting dataset that shogun could handle?
At some point I can add a ARFFWriter, but it isn't really needed right now, and can be done once the new serialisation stuff is ready.

gf712:
also the error seems to be due to `AveragedPerceptron`? I am guessing there's some UB going on

src/shogun/io/ARFFFile.h
```diff
@@ -0,0 +1,544 @@
...
+			return line.find_first_not_of(" \t\r\f\v") == std::string::npos;
+		}
+
+		/**
+		 * Checks if the character in the lhs is in the rhs
+		 * @param lhs the single character to find
+		 * @param rhs a string with the characters to compare against the lhs
+		 * @return whether the character of the lhs is in the rhs
+		 */
+		SG_FORCED_INLINE bool char_in_string(char lhs, const std::string& rhs)
```
karlnapf: minor: maybe `contains`? since the types are already in the signature...

src/shogun/io/ARFFFile.h
```diff
@@ -0,0 +1,544 @@
...
+			}
+		}
+
+		/**
+		 * Returns a string in lowercase.
+		 *
+		 * @param line string to process
+		 * @return lowercase string
+		 */
+		SG_FORCED_INLINE std::string string_to_lower(const std::string& line)
```
karlnapf: same here: `to_lower`?

tests/unit/io/ARFFFile_unittest.cc
```diff
@@ -0,0 +1,180 @@
...
+
+	auto parser = std::make_unique<ARFFDeserializer>(s);
+	parser->read();
+	auto result = parser->get_features();
+	ASSERT_EQ(result->get_num_feature_obj(), 2);
+
+	auto col1 =
+	    dynamic_cast<CDenseFeatures<float64_t>*>(result->get_feature_obj(0));
+	auto col2 =
+	    dynamic_cast<CDenseFeatures<float64_t>*>(result->get_feature_obj(1));
```
karlnapf: maybe `->as()` since otherwise you need to nullptr check?

gf712: yup

src/shogun/io/ARFFFile.h
gf712: done

src/shogun/io/ARFFFile.h
gf712: I agree!

gf712:
@karlnapf I will skip the relational attributes for now, they don't seem to be used much and I need this parser to finish writing the openml stuff. Is it ok to merge this?

src/shogun/io/ARFFFile.cpp
```diff
@@ -0,0 +1,357 @@
...
+					date::sys_seconds t;
+					std::istringstream ss(elems[i]);
+					if (date_pos == m_date_formats.end())
+						SG_SERROR(
+						    "Unexpected date value \"%s\" on line %d.\n",
+						    elems[i].c_str(), m_line_number);
+					ss >> date::parse(*date_pos, t);
+					if (bool(ss))
+					{
+						auto value_timestamp = t.time_since_epoch().count();
```
karlnapf, line 375: I wonder whether this is a good thing to do ...

src/shogun/io/ARFFFile.cpp
```diff
@@ -0,0 +1,357 @@
...
+		case Attribute::Real:
+		case Attribute::Date:
+		case Attribute::Nominal:
+		{
+			auto casted_vec = shogun::get<std::vector<float64_t>>(vec);
+			SGMatrix<float64_t> mat(1, row_count);
+			memcpy(
+			    mat.matrix, casted_vec.data(),
+			    casted_vec.size() * sizeof(float64_t));
+			auto* feat = new CDenseFeatures<float64_t>(mat);
```
karlnapf: do we want everything to be float64?

src/shogun/io/ARFFFile.cpp
```diff
@@ -0,0 +1,357 @@
...
+			SGStringList<char> strings(row_count, max_string_length);
+			for (int j = 0; j < row_count; ++j)
+			{
+				SGString<char> current(max_string_length);
+				memcpy(
+				    current.string, casted_vec[j].data(),
+				    (casted_vec.size() + 1) * sizeof(char));
+				strings.strings[j] = current;
+			}
+			auto* feat = new CStringFeatures<char>(strings, EAlphabet::RAWBYTE);
```
karlnapf: nice that this works for strings.... what about sparse data?

src/shogun/io/ARFFFile.h
```diff
@@ -0,0 +1,545 @@
...
+	/** Contains miscellaneous string manipulation functions using the STL and
+	 * Java to C++ date format utilities */
+	namespace arff_detail
+	{
+		/**
+		 * Checks if string is blank
+		 * @param line to check
+		 * @return bool whether line is empty
+		 */
+		SG_FORCED_INLINE bool is_blank(const std::string& line)
```
karlnapf, line 35: stupid question, isnt there some minimal lib to do these things?
Since technically all this should be tested, or?

src/shogun/io/ARFFFile.h
```diff
@@ -0,0 +1,545 @@
...
+	 */
+	class ARFFDeserializer
+	{
+	private:
+		/**
+		 * The attributes supported in the ARFF format
+		 */
+		enum class Attribute
+		{
+			Numeric = 0,
```
karlnapf: ah ok so no sparse ...

src/shogun/io/ARFFFile.h
```diff
@@ -0,0 +1,545 @@
...
+			Integer = 1,
+			Real = 2,
+			String = 3,
+			Date = 4,
+			Nominal = 5
```
karlnapf: I guess we should add these type of feautres to shogun ... would be another argument to use it over some of the other libs that need everything in real matrix form

src/shogun/io/ARFFFile.cpp
gf712: mmmh, I can add `int64_t` too when a value is specifically `integer`

src/shogun/io/ARFFFile.h
gf712: I am not sure if there is such a library.. But these are quite generic functions that should probably be moved to another file

src/shogun/io/ARFFFile.cpp
karlnapf: my question is more: what if users want float32/int32 or whatever .... see loading csv files

src/shogun/io/ARFFFile.h
karlnapf: ++

karlnapf:
re dataset ... maybe some example from a tutorial?

cmake/external/date.cmake
```diff
@@ -0,0 +1,16 @@
+include(ExternalProject)
+set(DATE_PREFIX ${CMAKE_BINARY_DIR}/date)
+set(DATE_SOURCE_DIR "${THIRD_PARTY_DIR}/date")
+set(DATE_INCLUDE_DIR "${DATE_SOURCE_DIR}/include")
+ExternalProject_Add(
+        date
+        PREFIX ${DATE_PREFIX}
+        SOURCE_DIR ${DATE_SOURCE_DIR}
+        GIT_REPOSITORY https://github.com/HowardHinnant/date.git
```
karlnapf, line 9: seems ok to me to have this

src/shogun/io/ARFFFile.cpp
gf712: ah sure, well the user would have to specify int and float type imo, i.e. float32/int32 or float32/int64

cmake/external/date.cmake
gf712: yea, this will be part of the stl in C++20, so might as well start using it

src/shogun/io/ARFFFile.cpp
karlnapf: at least there should be an API for that. And then some kind of gatekeeper function offers that to users ?

cmake/external/date.cmake
karlnapf: hahahaha
Ah man, finally in c++20, c++ will have nice date parsing... things are changing fast these days ;)

src/shogun/io/ARFFFile.cpp
```diff
@@ -179,6 +179,12 @@ void ARFFDeserializer::read()
 	};
 	process_chunk(read_attributes, check_attributes, true);
 
+	auto pos = m_stream->tellg();
+	auto approx_data_line_count = std::count(std::istreambuf_iterator<char>(*m_stream),
+	        std::istreambuf_iterator<char>(), '\n');
+	reserve_vector_memory(approx_data_line_count);
```
karlnapf: nice! :)

src/shogun/io/ARFFFile.cpp
```diff
@@ -0,0 +1,363 @@
...
+		if (SG_UNLIKELY(m_current_line.substr(0, 1) == m_comment_string))
+			return;
+		// it's the data string (i.e. @data"), does not provide information
+		if (SG_UNLIKELY(
+		        to_lower(m_current_line.substr(0, strlen(m_data_string))) ==
+		        m_data_string))
+		{
+			return;
+		}
+		// assumes that until EOF we should expect comma delimited values
```
karlnapf, line 307: maybe a check with a user facing error msg since users might screw that up?

src/shogun/io/ARFFFile.cpp
karlnapf: Until we have meta-data support, we can do it like this. But maybe we can offer a utility function (swig-ed) that can convert from this format to a date string and vice versa?

src/shogun/io/ARFFFile.cpp
karlnapf: as discussed in IRC, I think it might make sense to template the primitive type, and then give a API that takes an enum (the usual `PT_FLOAT64` etc) to select how to cast the numbers

src/shogun/io/ARFFFile.cpp
karlnapf: ok as discussed can easily be added later

karlnapf: oh btw same thing here with the alphabet.... there might be multiple desired options ... Shogun can also deal with strings of uint16_t etc

src/shogun/io/ARFFFile.h
karlnapf: I actually wonder whether it even might make sense to SWIG a few of those ... but I am not 100% sure. What do you think?

karlnapf: probably there will be a lot of name clashes ...

karlnapf: but moving to a sep file is a good idea to keep this one shorter

src/shogun/io/ARFFFile.h
```diff
@@ -0,0 +1,569 @@
...
+			return cpp_time;
+		}
+	}  // namespace arff_detail
+#endif // SWIG
+	/**
+	 * ARFFDeserializer parses files in the ARFF format.
+	 * For information about this format see
+	 * https://waikato.github.io/weka-wiki/arff_stable/
+	 */
+	class ARFFDeserializer
```
karlnapf, line 363: I guess you mentioned this already, but wouldn't it be good if this class followed some kind of standard API base class thing?

src/shogun/io/ARFFFile.h
karlnapf: in shogun, most enums are CAPITAL, maybe follow that here as well?

src/shogun/io/ARFFFile.h
```diff
@@ -0,0 +1,569 @@
...
+			VectorResizeVisitor visitor{line_count};
+			for (auto& vec : m_data_vectors)
+				shogun::visit(visitor, vec);
+		}
+
+		/**
+		 * Visitor pattern to reserve memory for a std::vector
+		 * wrapped in a variant class.
+		 */
+		struct VectorResizeVisitor
```
karlnapf: I guess this should be hidden from SWIG as well?

src/shogun/io/ARFFFile.h
```diff
@@ -0,0 +1,569 @@
...
+				v.reserve(m_size);
+			}
+			int m_size;
+		};
+
+		/**
+		 * Visitor pattern to determine size of a std::vector
+		 * wrapped in a variant class.
+		 */
+		struct VectorSizeVisitor
```
karlnapf: and this?

src/shogun/io/ARFFFile.h
karlnapf: ah sorry this is private...why does the above need to be hidden then?

src/shogun/io/ARFFFile.cpp
gf712: yup, that's a good idea! You mean a class function that has something like `get_time_as_string()`?

src/shogun/io/ARFFFile.cpp
gf712: yes, I was trying to find that in csv but couldn't find it? I thought they return type depended on the getter used? i.e. `get_vector(T*& vector, int32_t& len);`

src/shogun/io/ARFFFile.h
gf712: I don't think they should be in swig, they are so short that it is better for users to write their own implementations of these functions in their own language no?

src/shogun/io/ARFFFile.h
gf712: true, ill change that

src/shogun/io/ARFFFile.h
gf712: it doesn't, I should even move this to the cpp file

src/shogun/io/ARFFFile.cpp
karlnapf: yes something like that...also maybe a vectorized one that converts a whole vector/matrix

karlnapf: (because the arff parser generates exactly that)

src/shogun/io/ARFFFile.cpp
karlnapf: I meant the factory, e.g. `CFeatures* features(CFile* file, machine_int_t primitive_type = PT_FLOAT64)`

src/shogun/io/ARFFFile.h
karlnapf: yeah ok

gf712:
> * using CombinedFeatures is the best idea? I am torn....see discussion in irc. But maybe leave it as it is for now
@karlnapf I have been thinking about this, I think CList is better, and then we can always add a CCombinedFeatures constructor with CList? The thing with CList is that is can be used with `CDenseFeatures<ST>::create_merged_copy(CList* others)` to "cast" a list of features to `CDenseFeatures`

src/shogun/io/ARFFFile.cpp
gf712: I put a check for element count that should catch any weird behaviour here

src/shogun/io/ARFFFile.h
gf712: I am not sure about the API base class. It'll be the same as @vigsterkr serialisation stuff. and then I can rewrite this a bit

tests/unit/io/ARFFFile_unittest.cc
```diff
@@ -0,0 +1,258 @@
...
+constexpr EPrimitiveType convert_type_to_enum<float64_t>()
+{
+	return EPrimitiveType::PT_FLOAT64;
+}
+
+template <>
+constexpr EPrimitiveType convert_type_to_enum<floatmax_t>()
+{
+	return EPrimitiveType::PT_FLOATMAX;
+}
```
gf712, line 68: @karlnapf I added these utility functions might be worth putting them somewhere else?

gf712: or it might already exist somewhere else?

src/interfaces/swig/shogun.i
```diff
@@ -82,8 +82,8 @@
 %include "Boost_includes.i"
 
 %include "SGBase.i"
-%include "IO.i"
 %include "Library.i"
+%include "IO.i"
```
gf712: @karlnapf @vigsterkr is it ok to swap the order? now that I depend on CList Library.i needs to come before so that swig knows of its existence.. It also means that Features.i should be above IO.i, but I am guessing that would cause issues?
Should probably rethink how these are split, since SGIO.h is used everywhere, but parsers aren't..

src/shogun/io/ARFFFile.h
```diff
@@ -0,0 +1,713 @@
...
+		    m_nominal_attributes;
+
+		/** the parsed features */
+		std::vector<std::shared_ptr<CFeatures>> m_features;
+	};
+
+	class ARFFSerializer
+	{
+	public:
+		ARFFSerializer(
```
gf712, line 681: @karlnapf I had some time so I wrote this (pretty bad) serialiser.. Would it make sense to add a visitor pattern to SGObjects? Or at least for features?

PR #4628, opened 2019-05-03
openml integration in shogun

gf712:
adds [OpenML](https://www.openml.org/home) natively to shogun.

TODO:
- [ ] write get/post
- [ ] write OpenML flown class 
- [ ] expose OpenML to all interfaces

### DIFF
--- a/src/shogun/io/OpenmlFlow.cpp
+++ b/src/shogun/io/OpenmlFlow.cpp
@@ -1,0 +1,100 @@
+/*
+ * This software is distributed under BSD 3-clause license (see LICENSE file).
+ *
+ * Authors: Gil Hoben
+ */
+
+#ifdef HAVE_CURL
+
+#include <shogun/io/OpenmlFlow.h>
+#include "OpenmlFlow.h"
+
+
+using namespace shogun;
+
+size_t writer(char *data, size_t size, size_t nmemb, std::string* buffer_in)
+{
+	// adapted from https://stackoverflow.com/a/5780603
+	// Is there anything in the buffer?
+	if (buffer_in->empty())
+	{
+		// Append the data to the buffer
+		buffer_in->append(data, size * nmemb);
+
+		return size * nmemb;
+	}
+
+	return 0;
+}
+
+const char* OpenMLReader::xml_server = "https://www.openml.org/api/v1/xml";
+const char* OpenMLReader::json_server = "https://www.openml.org/api/v1/json";
+const char* OpenMLReader::dataset_description = "/data/{}";
+const char* OpenMLReader::list_data_qualities = "/data/qualities/list";
+const char* OpenMLReader::data_features = "/data/features/{}";
+const char* OpenMLReader::list_dataset_qualities = "/data/qualities/{}";
+const char* OpenMLReader::list_dataset_filter = "/data/list/{}";
+const char* OpenMLReader::flow_file = "/flow/{}";
+
+const std::unordered_map<std::string, std::string>
+    OpenMLReader::m_format_options = {{"xml", xml_server},
+                                      {"json", json_server}};
+const std::unordered_map<std::string, std::string>
+    OpenMLReader::m_request_options = {
+
+        {"dataset_description", dataset_description},
+        {"list_data_qualities", list_data_qualities},
+        {"data_features", data_features},
+        {"list_dataset_qualities", list_dataset_qualities},
+        {"list_dataset_filter", list_dataset_filter},
+		{"flow_file", flow_file}};
+
+OpenMLReader::OpenMLReader(const std::string& api_key) : m_api_key(api_key)
+{
+}
+
+void OpenMLReader::post(const std::string& request, const std::string& data)
+{
+}
+
+void OpenMLReader::openml_curl_request_helper(const std::string& url)
+{
+	CURL* curl_handle = nullptr;
+
+	curl_handle = curl_easy_init();
+
+	if (!curl_handle)
+	{
+		SG_SERROR("Failed to initialise curl handle.")
+		return;
+	}
+
+	curl_easy_setopt(curl_handle, CURLOPT_URL, url.c_str());
+	curl_easy_setopt(curl_handle, CURLOPT_HTTPGET,1);
+	curl_easy_setopt(curl_handle, CURLOPT_WRITEFUNCTION, writer);
+	curl_easy_setopt(curl_handle, CURLOPT_WRITEDATA, &m_curl_response_buffer);
+
+	CURLcode res = curl_easy_perform(curl_handle);
+
+	openml_curl_error_helper(res);
+
+	curl_easy_cleanup(curl_handle);
+}
+
+void OpenMLReader::openml_curl_error_helper(CURLcode code) {
+
+}
+
+
+void OpenMLFlow::download_flow()
+{
+
+	auto reader = OpenMLReader(m_api_key);
+	auto return_string = reader.get("flow_file", "json", m_flow_id);
+}
+
+void OpenMLFlow::upload_flow(const OpenMLFlow& flow)
+{
+}
+
+#endif // HAVE_CURL
--- a/src/shogun/io/OpenmlFlow.h
+++ b/src/shogun/io/OpenmlFlow.h
@@ -1,0 +1,154 @@
+/*
+ * This software is distributed under BSD 3-clause license (see LICENSE file).
+ *
+ * Authors: Gil Hoben
+ */
+
+#ifndef SHOGUN_OPENMLFLOW_H
+#define SHOGUN_OPENMLFLOW_H
+
+#ifdef HAVE_CURL
+
+#include <shogun/io/SGIO.h>
+
+#include <curl/curl.h>
+#include <numeric>
+#include <string>
+#include <unordered_map>
+#include <vector>
+#include <iostream>
+
+namespace shogun
+{
+	class OpenMLReader
+	{
+
+	public:
+		explicit OpenMLReader(const std::string& api_key);
+
+		/**
+		 * Returns a string returned by the server given a request.
+		 * Raises an error if the returned code is not 200.
+		 * Additional arguments can be passed to the request,
+		 * which are then concatenated with a "/" character.
+		 *
+		 * @tparam Args argument type pack, should all be std::string
+		 * @param request the request name, see m_request_options
+		 * @param format the format to return the data in, see m_format_options
+		 * @param args the additional arguments to be passed to request
+		 * @return the returned stream from the server if the return code is 200
+		 */
+		template <typename... Args>
+		std::string
+		get(const std::string& request, const std::string& format, Args... args)
+		{
+			auto find_format = m_format_options.find(format);
+			if (find_format == m_format_options.end())
+			{
+				SG_SERROR(
+				    "The provided format \"%s\" is not available\n",
+				    format.c_str())
+			}
+			auto find_request = m_request_options.find(request);
+			if (find_request == m_request_options.end())
+			{
+				SG_SERROR(
+				    "Could not find a way to solve the request \"%s\"\n",
+				    request.c_str())
+			}
+			std::string request_format = find_format->second;
+			std::string request_path = find_request->second;
+
+			// get additional args and concatenate them with "/"
+			if (sizeof...(Args) > 0)
+			{
+				if (request_path.substr(request_path.size() - 2) == "{}")
+				{
+					request_path =
+					    request_path.substr(0, request_path.size() - 2);
+				}
+				else
+				{
+					SG_SERROR(
+					    "The provided request \"%s\" cannot handle additional "
+					    "args.\n",
+					    request.c_str())
+				}
+				std::vector<std::string> args_vec = {args...};
+				std::string args_string = std::accumulate(
+				    args_vec.begin() + 1, args_vec.end(), args_vec.front(),
+				    [](std::string s0, std::string& s1) {
+					    return s0 += "/" + s1;
+				    });
+				request_path += args_string;
+			}
+
+			std::string url = request_format + request_path + "?" + m_api_key;
+
+			openml_curl_request_helper(url);
+
+			return m_curl_response_buffer;
+		}
+
+		void post(const std::string& request, const std::string& data);
+
+	private:
+
+		std::string m_curl_response_buffer;
+
+		/**
+		 * Initialises CURL session and gets the data.
+		 * This function also handles the response code from the server.
+		 *
+		 * @param url the url to query
+		 */
+		void openml_curl_request_helper(const std::string& url);
+
+		/**
+		 * Handles all possible codes
+		 *
+		 * @param code the code returned by the query
+		 */
+		void openml_curl_error_helper(CURLcode code);
+
+		std::string m_api_key;
+
+		static const char* xml_server;
+		static const char* json_server;
+
+		static const std::unordered_map<std::string, std::string>
+		    m_format_options;
+		static const std::unordered_map<std::string, std::string>
+		    m_request_options;
+
+		/* DATA API */
+		static const char* dataset_description;
+		static const char* list_data_qualities;
+		static const char* data_features;
+		static const char* list_dataset_qualities;
+		static const char* list_dataset_filter;
+
+		/* FLOW API */
+		static const char* flow_file;
+	};
+
+	class OpenMLFlow
+	{
+
+	public:
+		explicit OpenMLFlow(
+		    const std::string& api_key, const std::string& flow_id)
+		    : m_api_key(api_key), m_flow_id(flow_id){};
+
+		void download_flow();
+
+		static void upload_flow(const OpenMLFlow& flow);
+
+	private:
+		std::string m_api_key;
+		std::string m_flow_id;
+	};
+} // namespace shogun
+#endif // HAVE_CURL
+
+#endif // SHOGUN_OPENMLFLOW_H


src/shogun/io/OpenmlFlow.cpp
```diff
@@ -0,0 +1,100 @@
...
+		// Append the data to the buffer
+		buffer_in->append(data, size * nmemb);
+
+		return size * nmemb;
+	}
+
+	return 0;
+}
+
+const char* OpenMLReader::xml_server = "https://www.openml.org/api/v1/xml";
```
karlnapf: should these things be hard-coded here?

src/shogun/io/OpenmlFlow.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+void OpenMLReader::openml_curl_request_helper(const std::string& url)
+{
+	CURL* curl_handle = nullptr;
+
+	curl_handle = curl_easy_init();
+
+	if (!curl_handle)
+	{
+		SG_SERROR("Failed to initialise curl handle.")
```
karlnapf: maybe some infos on what happened?

karlnapf:
sweet, looking forward to see this once it is working :)

src/shogun/io/OpenmlFlow.cpp
vigsterkr: yes

src/shogun/io/OpenmlFlow.cpp
gf712: I am not sure what the exact error would be... From the docs: "If this function [curl_easy_init] returns NULL, something went wrong and you cannot use the other curl functions."

gf712:
@vigsterkr I copied a few things from deadbeef to add RapidJSON as a dependency

vigsterkr:
saw it 😼

src/shogun/CMakeLists.txt
```diff
@@ -407,55 +400,58 @@ IF(BUNDLE_JSON)
...
 SHOGUN_DEPENDENCIES(
-  LIBRARY LibXml2
-  SCOPE PUBLIC
-  CONFIG_FLAG HAVE_XML)
+        LIBRARY LibXml2
+        SCOPE PUBLIC
+        CONFIG_FLAG HAVE_XML)
+# RapidJSON
+include(external/RapidJSON)
+SHOGUN_INCLUDE_DIRS(SCOPE PUBLIC ${RAPIDJSON_INCLUDE_DIR})
```
gf712: @vigsterkr is it ok to have RapidJSON as PUBLIC?

vigsterkr: i guess it was like this in the other PR, right? i mean currently i include the RapidJSON only in .cpp so private should be fine.... but maybe/probably that caused me troubles with unit tests ...?

gf712: it was private in the other PR, but swig needs the RapidJSON header files

vigsterkr: ~why?~ now i understand why :)

src/shogun/io/OpenMLFlow.h
```diff
@@ -0,0 +1,222 @@
...
+
+#include <shogun/lib/config.h>
+
+#ifdef HAVE_CURL
+
+#include <shogun/base/SGObject.h>
+#include <shogun/io/SGIO.h>
+
+#include <curl/curl.h>
+#include <rapidjson/document.h>
```
vigsterkr: i would avoid this by all means :)

src/shogun/io/OpenMLFlow.h
```diff
@@ -0,0 +1,222 @@
...
+
+	private:
+		std::string m_name;
+		std::string m_description;
+		std::string m_class_name;
+		parameters_type m_parameters;
+		components_type m_components;
+
+#ifndef SWIG
+		static void check_flow_response(rapidjson::Document& doc);
```
vigsterkr: does it really need to be part of the class, meaning can't you just have this function in the implementation?

src/shogun/io/OpenMLFlow.h
```diff
@@ -0,0 +1,222 @@
...
+
+		static SG_FORCED_INLINE void emplace_string_to_map(
```
vigsterkr: same as above...

src/shogun/io/OpenMLFlow.h
```diff
@@ -0,0 +1,222 @@
...
+		    std::unordered_map<std::string, std::string>& param_dict,
+		    const std::string& name)
+		{
+			if (v[name.c_str()].GetType() == rapidjson::Type::kStringType)
+				param_dict.emplace(name, v[name.c_str()].GetString());
+			else
+				param_dict.emplace(name, "");
+		}
+
+		static SG_FORCED_INLINE void emplace_string_to_map(
```
vigsterkr: same as above

src/shogun/io/OpenMLFlow.h
gf712: no, I guess I can move to free function and avoid exposing RapidJSON to swig :)

gf712:
I have a local version mostly working (with binary classification tasks at least). There are some memory issues but they will be fixed once we switch to smart pointers.

src/shogun/io/openml/OpenMLRun.cpp
```diff
@@ -0,0 +1,183 @@
...
+		{
+			SG_SERROR("INTERNAL ERROR: failed to cast model to machine!\n")
+		}
+		auto train_idx = task->get_train_indices();
+		auto test_idx = task->get_test_indices();
+
+		auto xval_storage = std::make_shared<CrossValidationStorage>();
+		xval_storage->set_num_folds(task->get_num_fold());
+		xval_storage->set_num_runs(task->get_num_repeats());
+		machine->set_store_model_features(true);
```
gf712, line 59: @karlnapf I put this line because it's in crossvalidation, but I am not sure I understand what it does? The training does fail if it isn't set because the kernel machine tries to access a support vector element that is out of range...

gf712:
@karlnapf so this works now, a minimal example looks like this:
```cpp
#include <shogun/base/init.h>
#include <shogun/io/openml/OpenMLFlow.h>
#include <shogun/io/openml/OpenMLTask.h>
#include <shogun/io/openml/OpenMLRun.h>
#include <shogun/io/openml/OpenMLSplit.h>

using namespace shogun;

int main()
{
	init_shogun_with_defaults();
	sg_io->set_loglevel(MSG_GCDEBUG);
	auto flow = OpenMLFlow::download_flow("9602", "");
	auto task = OpenMLTask::get_task("4423", "");
	auto run = OpenMLRun::run_flow_on_task(flow, task);
        // TODO: run.publish()
	exit_shogun();
}
```
Once the smart pointer PR is merged it'll work with interfaces

src/shogun/io/openml/OpenMLRun.cpp
karlnapf: kernel machines are models of the form `$f(x)=\sum_i \alpha_i k(x_i,x)$`, so they can only be evaluated together with the training data `${x_i}_{i=1}^n$`. This means you need to keep track of that data. Right now, subset views on features modify the state of the feature object (which is stupid), so when the xvalidation validation set is evaluated, the code removes/adds another subset. This means that the data stored in the kernel machine is now different (as its state changed). The call is there to make the machine store a copy of the training data with non-zero `$\alpha_i$` coefficients. We don't clone the features as e.g. SVMs have sparse `$\alpha_i$` coefficients ...

All this should be removed. Features should be immutable, subsets should be new instances that share memory with the original feature, then `store_model_features` is not needed anymore. I will have a go at this at some point soon, it has been on my list for a while

karlnapf: so for now, you need this call

karlnapf:
> @karlnapf so this works now, a minimal example looks like this:

great! this is awesome!

> Once the smart pointer PR is merged it'll work with interfaces

Ah yes, a meta example would be good for that indeed. Is it reasonable that the smart pointer PR will be merged soon?

src/shogun/features/CombinedFeatures.cpp
```diff
@@ -34,15 +34,14 @@ CFeatures* CCombinedFeatures::duplicate() const
 
 CCombinedFeatures::~CCombinedFeatures()
 {
-	SG_UNREF(feature_array);
```
karlnapf: ?

src/shogun/features/CombinedFeatures.h
```diff
@@ -185,7 +185,7 @@ class CCombinedFeatures : public CFeatures
 
 	protected:
 		/** feature array */
-		CDynamicObjectArray* feature_array;
+		std::vector<CFeatures*> feature_array;
```
karlnapf: ah I see. nice!

karlnapf: ah wait. Aren't there reference counting issues with this? CDynamicObjectArray increases/decreases ref counters of the objects inside, which isnt happening here anymore as far as I can see?

src/shogun/features/CombinedFeatures.cpp
```diff
@@ -146,27 +146,25 @@ bool CCombinedFeatures::append_feature_obj(CFeatures* obj)
 	num_vec=n;
 
 	int num_feature_obj = get_num_feature_obj();
-	feature_array->push_back(obj);
```
karlnapf: this used to `SG_REF`

src/shogun/features/CombinedFeatures.cpp
```diff
@@ -146,27 +146,25 @@ bool CCombinedFeatures::append_feature_obj(CFeatures* obj)
...
-	return num_feature_obj+1 == feature_array->get_num_elements();
+	feature_array.push_back(obj);
+	return num_feature_obj+1 == feature_array.size();
 }
 
 bool CCombinedFeatures::delete_feature_obj(int32_t idx)
 {
-	return feature_array->delete_element(idx);
```
karlnapf: and this used to `SG_UNREF`

src/shogun/io/openml/OpenMLData.cpp
```diff
@@ -0,0 +1,271 @@
...
+	    std::find(m_feature_names.begin(), m_feature_names.end(), label);
+	if (find_label == m_feature_names.end())
+		SG_SERROR("Requested label \"%s\" not in the dataset!\n", label.c_str())
+	auto col_idx = std::distance(m_feature_names.begin(), find_label);
+	auto feat_type_copy = m_feature_types;
+	feat_type_copy.erase(feat_type_copy.begin() + col_idx);
+	for (const auto type : feat_type_copy)
+	{
+		if (type == ARFFDeserializer::Attribute::STRING)
+			SG_SERROR("Currently cannot process string features!\n")
```
karlnapf: SG_NOTIMPLEMENTED

src/shogun/io/openml/OpenMLData.cpp
```diff
@@ -0,0 +1,271 @@
...
+	}
+	//	auto result = std::make_shared<CDenseFeatures>();
+	std::shared_ptr<CDenseFeatures<float64_t>> result;
```
karlnapf, line 159: there is again the question of the datatype inside the feautres.....

src/shogun/features/CombinedFeatures.h
gf712: Yes, there are issues with this, but I think once this is all smart pointer it'll be fine. I was getting issues before so I changed this a bit. I'll rebase when smart pointers are merged, pretty sure @vigsterkr changed all CDynamicObjectArray with std::vector

gf712:
>Ah yes, a meta example would be good for that indeed. Is it reasonable that the smart pointer PR will be merged soon?
Need to ask @vigsterkr, last week he said he was mostly there

src/shogun/io/openml/OpenMLData.cpp
```diff
@@ -0,0 +1,271 @@
...
+
+	switch (m_feature_types[col_idx])
+	{
+	// real features
+	case ARFFDeserializer::Attribute::REAL:
+	case ARFFDeserializer::Attribute::NUMERIC:
+	case ARFFDeserializer::Attribute::INTEGER:
+	case ARFFDeserializer::Attribute::DATE:
+	{
+		auto casted_feat = std::dynamic_pointer_cast<CDenseFeatures<float64_t>>(
```
karlnapf, line 227: why are labels stores as features?

src/shogun/io/openml/OpenMLData.cpp
```diff
@@ -0,0 +1,271 @@
...
+	break;
+		// nominal features
+	case ARFFDeserializer::Attribute::NOMINAL:
+	{
+		auto casted_feat = std::dynamic_pointer_cast<CDenseFeatures<float64_t>>(
+		    target_label_as_feat);
+		auto labels_vec = casted_feat->get_feature_matrix().get_row_vector(0);
+		for (auto& val : labels_vec)
+		{
+			if (val == 0)
```
karlnapf, line 244: maybe it should be an option how those are encoded?

src/shogun/io/openml/OpenMLData.h
```diff
@@ -0,0 +1,165 @@
...
+		    const std::string& default_target_attribute,
+		    const std::string& row_id_attribute,
+		    const std::string& ignore_attribute,
+		    const std::string& version_label, const std::string& citation,
+		    std::vector<std::string> tag, const std::string& visibility,
+		    const std::string& original_data_url, const std::string& paper_url,
+		    const std::string& update_comment, const std::string& md5_checksum,
+		    std::vector<
+		        std::unordered_map<std::string, std::vector<std::string>>
+
```
karlnapf, line 41: whitespace issue

src/shogun/io/openml/OpenMLData.cpp
gf712: Yea, I am thinking if it would be giving the user the option of how to add a preprocessing step for this?

src/shogun/io/openml/OpenMLData.cpp
gf712: Because in ARFF any value can be a feature or label, so I store everything inside CombinedFeatures

src/shogun/io/openml/OpenMLReader.cpp
```diff
@@ -0,0 +1,97 @@
...
+/* FLOW API */
+const char* OpenMLReader::flow_file = "/flow/{}";
+/* TASK API */
+const char* OpenMLReader::task_file = "/task/{}";
+/* SPLIT API */
+const char* OpenMLReader::get_split = "";
+
+const std::unordered_map<std::string, std::string>
+		OpenMLReader::m_format_options = {{"xml", xml_server},
+										  {"json", json_server},
```
karlnapf: some whitespace issues

src/shogun/io/openml/OpenMLReader.cpp
```diff
@@ -0,0 +1,97 @@
...
+	curl_easy_setopt(curl_handle, CURLOPT_WRITEFUNCTION, writer);
+	curl_easy_setopt(curl_handle, CURLOPT_WRITEDATA, &m_curl_response_buffer);
+
+	CURLcode res = curl_easy_perform(curl_handle);
+
+	if (res != CURLE_OK)
+		SG_SERROR("Connection error: %s.\n", curl_easy_strerror(res))
+
+	curl_easy_cleanup(curl_handle);
+#endif // HAVE_CURL
```
karlnapf: maybe an error message with instructions how to compile shogun with curl?

src/shogun/io/openml/OpenMLData.cpp
gf712: yea, this isn't great, it was just to make things work... It would be good to implement more label/feature conversions inside the machine classes

src/shogun/io/openml/OpenMLRun.cpp
```diff
@@ -0,0 +1,183 @@
...
+	features = data->get_features(data->get_default_target_attribute());
+
+	auto storage = std::make_shared<CrossValidationStorage>();
+
+	if (task->get_split()->contains_splits())
+	{
+		auto machine = std::dynamic_pointer_cast<CMachine>(model);
+		if (!machine)
+		{
+			SG_SERROR("INTERNAL ERROR: failed to cast model to machine!\n")
```
karlnapf: please give the user more infos here. you could use `get_name()` to print the class name. No need to write INTERNAL_ERROR

gf712: yup, this is just a debugging message for myself. this will be replaced with `->as<CMachine>` once the smart pointers are ready

src/shogun/io/openml/OpenMLRun.cpp
```diff
@@ -0,0 +1,183 @@
...
+			{
+				auto* fold = new CrossValidationFoldStorage();
+				SG_REF(fold)
+
+				auto cloned_machine = (CMachine*)machine->clone();
+
+				// TODO while these are not used through const interfaces,
+				//  we unfortunately have to clone, even though these could be
+				//  shared
+				auto features_clone = (CFeatures*)features->clone();
```
karlnapf: I see the TODO is taken from the other xval. But we shouldn't merge this.
What about you create a new features instance with the same data instead of cloning it. Same with the labels

src/shogun/io/openml/OpenMLReader.cpp
gf712: yes that would be good. Is the shogun release with conda compiled with curl.h btw?

src/shogun/io/openml/OpenMLRun.cpp
```diff
@@ -0,0 +1,183 @@
...
+
+		// copied/adapted from crossvalidation
+		for (auto repeat_idx : range(task->get_num_repeats()))
+		{
+			for (auto fold_idx : range(task->get_num_fold()))
+			{
+				auto* fold = new CrossValidationFoldStorage();
+				SG_REF(fold)
+
+				auto cloned_machine = (CMachine*)machine->clone();
```
karlnapf: this might clone the attached data of the machine. Could we have a method (in develop actually) that clones all model and hyper-parameters of a machine? We somehow need to make a clone without copying the features/labels

gf712: something like a deep vs shallow clone?

src/shogun/features/CombinedFeatures.h
karlnapf: ah ok then it will be `std::vector<shared_ptr<CFeautres>>`?

src/shogun/io/openml/OpenMLData.cpp
karlnapf: sounds good!

src/shogun/io/openml/OpenMLData.cpp
karlnapf: I see ...

src/shogun/io/openml/OpenMLReader.cpp
karlnapf: idk actually
@dougalsutherland you know?

gf712:
> Really good :)
> I made a round of comments ... no rush to address them, but would be good to do that before we merge this. What else is missing?

I need to add the POST methods to publish and then it'll be minimally working, i.e. can do a basic workflow. I need to write the xml version of this for that. I'll write some utility functions for that, so shouldn't be too hard. 
Then it would conform with openml so we could merge it, and then later I can add more utility functions used to query the server, i.e. list all tasks

src/shogun/io/openml/OpenMLRun.cpp
gf712: yes, I did some shameless copy/paste here :D

src/shogun/io/openml/OpenMLRun.cpp
karlnapf: kind of, yes. Something temporary is fine (at least better than cloning!)
Eventually, the `view` or `subset` method will return a new instance

I remember @vinx13 was working on this at some point. What is the state of that @vinx13 
Might be something to tackle in Berlin :)

src/shogun/io/openml/OpenMLReader.cpp
djsutherland: Yeah it should be: https://github.com/conda-forge/shogun-cpp-feedstock/blob/25017238b169cc8c8d28f12609a7bb9c379997ab/recipe/meta.yaml#L68

I need to update the recipe though, not sure whether it really works with current conda-forge state anymore :|

src/shogun/io/openml/OpenMLRun.cpp
karlnapf: Actually that makes me think. Can't we re-use the existing code? Could refactor it to allow that. Because having it twice is meh

src/shogun/io/openml/OpenMLReader.cpp
karlnapf: Ok thanks for the quick reply. BTW we have a meeting in Berlin June 7-10 in case you happen to be available and keen. :)

src/shogun/io/openml/OpenMLRun.cpp
gf712: sure thing, I mean what I wrote is the boilerplate, so can just add a utility function in crossvalidation that does a fold train/test/eval and takes subset as an argument

karlnapf: for example, yes :)
Maybe we can do that in Berlin ...

src/shogun/features/CombinedFeatures.h
gf712: Yup, I just didn't mess with that because then I would have to rewrite more things, and this just made it work!

src/shogun/io/openml/OpenMLData.cpp
gf712: btw see my last comment on https://github.com/shogun-toolbox/shogun/pull/4627 about this (tldr we probably should use CList instead, no?)

src/shogun/io/openml/OpenMLRun.cpp
vinx13: We have only features or labels view/subset, but not for machine

src/shogun/features/CombinedFeatures.h
karlnapf: Sure thing! We can tune these things in a second round

src/shogun/io/openml/OpenMLData.cpp
karlnapf: Yes then you can mix and match objects

src/shogun/io/openml/OpenMLRun.cpp
karlnapf: That’s fine. So view creates a shallow clone and then adds subset? Is that in a branch or in develop? If I’m branch maybe it’s time to merge/finalise it?

vinx13: It creates a shallow clone and adds subset. We already have it in develop branch

src/shogun/io/openml/OpenMLReader.cpp
djsutherland: @karlnapf Yep, I'm going to be at ICML though. :|

karlnapf: Ah yes of course. Deep kexp :) enjoy then!

src/shogun/io/OpenmlFlow.cpp
iglesias: Any value in using this other global init instead?
https://stackoverflow.com/questions/50987505/how-to-get-error-reason-when-curl-easy-init-fails

iglesias:
> @karlnapf so this works now, a minimal example looks like this:
> 
> ```c++
> #include <shogun/base/init.h>
> #include <shogun/io/openml/OpenMLFlow.h>
> #include <shogun/io/openml/OpenMLTask.h>
> #include <shogun/io/openml/OpenMLRun.h>
> #include <shogun/io/openml/OpenMLSplit.h>
> 
> using namespace shogun;
> 
> int main()
> {
> 	init_shogun_with_defaults();
> 	sg_io->set_loglevel(MSG_GCDEBUG);
> 	auto flow = OpenMLFlow::download_flow("9602", "");
> 	auto task = OpenMLTask::get_task("4423", "");
> 	auto run = OpenMLRun::run_flow_on_task(flow, task);
>         // TODO: run.publish()
> 	exit_shogun();
> }
> ```
> 
> Once the smart pointer PR is merged it'll work with interfaces

What are 9602 and 4423?

src/shogun/io/OpenmlFlow.cpp
gf712: ah thanks, yes, that seems helpful!

karlnapf:
yes, let's keep it :)

PR #4632, opened 2019-05-12
Fix bug to make macOS build in observable framework 

theartful:
@geektoni not sure if you need this


geektoni:
Thank you @theartful! I surely needed this! :) 

Merging (the python error will be fixed by #4626)!

PR #4633, opened 2019-05-13
Refactor LDA classifier to make it more "observable".

geektoni:
* Removed getter and setters in favour of the `get`/`put` architecture;
* Use `put()` to set weights and bias;
* Correct LDA unit tests;

### DIFF
--- a/src/shogun/classifier/LDA.cpp
+++ b/src/shogun/classifier/LDA.cpp
@@ -34,8 +34,13 @@ CLDA::CLDA(
     : CDenseRealDispatch<CLDA, CLinearMachine>(), m_gamma(gamma)
 {
 	init();
-	set_features(traindat);
-	set_labels(trainlab);
+
+	features = traindat;
+	SG_REF(features)
+
+	m_labels = trainlab;
+	SG_REF(trainlab)
+
 	m_method = method;
 	m_gamma = gamma;
 	m_bdc_svd = bdc_svd;
@@ -66,7 +71,6 @@ bool CLDA::train_machine_templated(CDenseFeatures<ST>* data)
 {
 	index_t num_feat = data->get_num_features();
 	index_t num_vec = data->get_num_vectors();
-	;
 
 	bool lda_more_efficient = (m_method == AUTO_LDA && num_vec <= num_feat);
 
@@ -103,9 +107,12 @@ bool CLDA::solver_svd(CDenseFeatures<ST>* data)
 	// copy w_st into w
 	for (index_t i = 0; i < w.size(); ++i)
 		w[i] = sign * w_st[i];
-	set_w(w);
 
+	set_w(w);
 	set_bias(-0.5 * sign * (m_neg + m_pos));
+
+	observe<SGVector<float64_t >>(0,"w");
+	observe<float64_t>(1,"bias");
 
 	return true;
 }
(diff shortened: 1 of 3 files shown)


src/shogun/classifier/LDA.cpp
```diff
@@ -34,8 +34,8 @@ CLDA::CLDA(
     : CDenseRealDispatch<CLDA, CLinearMachine>(), m_gamma(gamma)
 {
 	init();
-	set_features(traindat);
-	set_labels(trainlab);
+	put("features", dynamic_cast<CFeatures*>(traindat));
```
karlnapf: nooooo
I would really try to avoid using this string based setting of parameters inside c++ code. It is more for swig

src/shogun/classifier/LDA.h
```diff
@@ -158,19 +140,6 @@ class CLDA : public CDenseRealDispatch<CLDA, CLinearMachine>
...
 		}
 
-		/** set features
-		 *
-		 * @param feat features to set
-		 */
-		virtual void set_features(CDotFeatures* feat)
-		{
-			if (feat->get_feature_class() != C_DENSE ||
-				feat->get_feature_type() != F_DREAL)
```
karlnapf: these checks are important, need to move them into train if this is removed

src/shogun/classifier/LDA.cpp
karlnapf: Sorry those are within class methods.... I didnt realise

karlnapf: the dynamic cast is unchecked here

karlnapf: if it was for me I would make both labels and features read-only (not puttable), but those are only received through a "train" method. But no need to do that here now

tests/unit/classifier/LDA_unittest.cc
```diff
@@ -200,10 +200,10 @@ TEST(LDA, num_classes_in_labels_exception)
 	SGVector<float64_t> lab{1, -1, 2};
 	SGMatrix<float64_t> feat(1, 3);
 	auto labels = some<CMulticlassLabels>();
-	labels->set_labels(lab);
+	labels->put("labels", lab);
```
karlnapf: why not use the ctor above?

tests/unit/classifier/LDA_unittest.cc
```diff
@@ -200,10 +200,10 @@ TEST(LDA, num_classes_in_labels_exception)
...
 	auto features = some<CDenseFeatures<float64_t>>(feat);
 	auto lda = some<CLDA>(0, SVD_LDA);
-	lda->set_labels(labels);
+	lda->put("labels", dynamic_cast<CLabels*>(labels.get()));
```
karlnapf: is there the need to use .get? I think there is a put for some, or?

src/shogun/classifier/LDA.h
geektoni: LDA uses features dispatching, so there is no need for these checks anymore since we are sure that we will get that type of features.

src/shogun/classifier/LDA.cpp
geektoni: err I'm a bit confused. Anyway, I just assign them directly now.

tests/unit/classifier/LDA_unittest.cc
geektoni: Yes, there is a `put` for `Some` indeed.

src/shogun/classifier/LDA.cpp
```diff
@@ -66,7 +71,6 @@ bool CLDA::train_machine_templated(CDenseFeatures<ST>* data)
 {
 	index_t num_feat = data->get_num_features();
 	index_t num_vec = data->get_num_vectors();
-	;
```
karlnapf: :dancer:

tests/unit/classifier/LDA_unittest.cc
```diff
@@ -105,15 +105,15 @@ void test_with_method(
 	generate_test_data<float64_t>(lab, feat);
 
 	auto features = some<CDenseFeatures<float64_t>>(feat);
-	auto labels = some<CBinaryLabels>(lab);
+	auto labels = Some<CLabels>::from_raw(new CBinaryLabels(lab));
```
karlnapf: I don't understand why this change is needed. could you explain?

karlnapf: ah because of the downcast

geektoni: That's because of `put(std::string, Some<>)`. It works only for types which are `is_sg_base`, therefore I need to pass `CLabels`.

karlnapf: yep, all good.

geektoni:
This will go again to the `feature/observable_framework` branch instead of `develop`, since I am using the new `observe()` method instead of `put()`.

geektoni:
Merging since MacOS MKL failures are unrelated.

PR #4634, opened 2019-05-14
Change some feature functions to be constant

theartful:


src/shogun/features/hashed/HashedSparseFeatures.cpp
```diff
@@ -275,7 +275,7 @@ void CHashedSparseFeatures<ST>::add_to_dense_vec(float64_t alpha, int32_t vec_id
 }
 
 template <class ST>
-int32_t CHashedSparseFeatures<ST>::get_nnz_features_for_vector(int32_t num)
+int32_t CHashedSparseFeatures<ST>::get_nnz_features_for_vector(int32_t num) const
```
gf712: can this be `const noexcept`?

theartful: I believe so

theartful: @gf712 `SG_NOTIMPLEMENTED` raises a `ShogunException` error. This would disqualify it
https://github.com/shogun-toolbox/shogun/blob/6182956e39d140943a55cc1113659a8062c0090c/src/shogun/features/ExplicitSpecFeatures.cpp#L171-L175

gf712: Ah ok, fair enough!

PR #4635, opened 2019-05-14
Use DotIterator in NearestCentroid classifier

theartful:


src/shogun/classifier/NearestCentroid.cpp
```diff
@@ -54,34 +54,33 @@ namespace shogun{
...
 			data = distance->get_lhs();
 		}
-		int32_t num_vectors = data->get_num_vectors();
-		int32_t num_classes = ((CMulticlassLabels*) m_labels)->get_num_classes();
-		int32_t num_feats = ((CDenseFeatures<float64_t>*) data)->get_num_features();
-		SGMatrix<float64_t> centroids(num_feats,num_classes);
+		ASSERT(data->get_feature_class() == C_DENSE)
+
+		auto* multiclass_labels = m_labels->as<CMulticlassLabels>();
+		auto* dense_data = data->as<CDenseFeatures<float64_t>>();
```
karlnapf: if this is done, then there needs to be no check above (different for the previous static cast of course)

src/shogun/classifier/NearestCentroid.cpp
```diff
@@ -54,34 +54,33 @@ namespace shogun{
...
+			current.add(1, target_vec);
 			num_per_class[current_class]++;
-			((CDenseFeatures<float64_t>*)data)->free_feature_vector(current, current_len, current_free);
 		}
 
-
 		for (int32_t i=0 ; i<num_classes ; i++)
 		{
-			float64_t* target = centroids.matrix + num_feats*i;
+			auto* target = centroids.get_column_vector(i);
```
karlnapf: maybe SGVector this? instead of raw pointers?

src/shogun/classifier/NearestCentroid.cpp
```diff
@@ -54,34 +54,33 @@ namespace shogun{
...
+		for(const auto& current : DotIterator(dense_data))
 		{
-			int32_t current_len;
-			bool current_free;
-			int32_t current_class = ((CMulticlassLabels*) m_labels)->get_label(idx);
-			float64_t* target = centroids.matrix + num_feats*current_class;
-			float64_t* current = ((CDenseFeatures<float64_t>*)data)->get_feature_vector(idx,current_len,current_free);
-			SGVector<float64_t>::add(target,1.0,target,1.0,current,current_len);
+			const auto current_class = *(iter_labels++);
+			auto target = centroids.get_column_vector(current_class);
```
karlnapf: maybe those two lines can be combined via returning SGVector as column vector?

theartful: done

src/shogun/classifier/NearestCentroid.cpp
```diff
@@ -54,51 +53,46 @@ namespace shogun{
...
-		int32_t num_feats = ((CDenseFeatures<float64_t>*) data)->get_num_features();
-		SGMatrix<float64_t> centroids(num_feats,num_classes);
+
+		auto* multiclass_labels = m_labels->as<CMulticlassLabels>();
+		auto* dense_data = data->as<CDenseFeatures<float64_t>>();
+
+		int32_t num_classes = multiclass_labels->get_num_classes();
+		int32_t num_feats = dense_data->get_num_features();
+
+		SGMatrix<float64_t> centroids(num_feats, num_classes);
```
karlnapf: doesnt the ctor init with zeros these days?

src/shogun/classifier/NearestCentroid.cpp
```diff
@@ -54,51 +53,46 @@ namespace shogun{
...
 		centroids.zero();
 
-		int64_t* num_per_class = new int64_t[num_classes];
-		for (int32_t i=0 ; i<num_classes ; i++)
-		{
-			num_per_class[i]=0;
-		}
+		SGVector<int64_t> num_per_class(num_classes);
```
karlnapf: and this one? maybe I remember this wrong

src/shogun/classifier/NearestCentroid.cpp
```diff
@@ -54,51 +53,46 @@ namespace shogun{
...
+			auto target = centroids.get_column(i);
 			int32_t total = num_per_class[i];
 			float64_t scale = 0;
 			if(total>1)
 				scale = 1.0/((float64_t)(total-1));
 			else
 				scale = 1.0/(float64_t)total;
 
-			SGVector<float64_t>::scale_vector(scale,target,num_feats);
+			target.scale(scale);
```
karlnapf: I wonder whether we should offer a linalg method to do this.....Might be a bit faster if done in eigen matrix-wise
`scale_cols(SGMatrix<T> mat, SGVector<T> scalars)`

src/shogun/classifier/NearestCentroid.cpp
```diff
@@ -54,51 +53,46 @@ namespace shogun{
...
+		num_per_class.zero();
 
-		for (int32_t idx=0 ; idx<num_vectors ; idx++)
+		auto iter_labels = multiclass_labels->get_int_labels().begin();
+		for(const auto& current : DotIterator(dense_data))
 		{
-			int32_t current_len;
-			bool current_free;
-			int32_t current_class = ((CMulticlassLabels*) m_labels)->get_label(idx);
-			float64_t* target = centroids.matrix + num_feats*current_class;
```
karlnapf: lol this is definitely cleaner :)

src/shogun/classifier/NearestCentroid.cpp
theartful: I will check it

theartful: Both only use SG_CALLOC, which I don't think it does.

karlnapf: http://www.cplusplus.com/reference/cstdlib/calloc/

theartful: I didn't think any C function would be so kind to initialize allocated memory!

src/shogun/classifier/NearestCentroid.cpp
theartful: I think a column/row manipulation api would be better than getting the row/column first, then operating on it. I found it a bit confusing that `get_column` returns an SGVector that maps to the matrix, but `get_row` copies the values (which makes sense since it's in column-major).

src/shogun/classifier/NearestCentroid.cpp
karlnapf: hehe :)

src/shogun/classifier/NearestCentroid.cpp
karlnapf: It is confusing indeed. I think in fact that the `get_row` should be removed, or rather renamed to `copy_row`.
And then we can add a `linalg` call to operate on the columns

theartful: `scale_cols(SGMatrix<T> mat, SGVector<T> scalars)` function would make things cleaner
Does implementing it require doing it in the different backends?
Should I also change `get_row` to `copy_row`? I think that's a better name.

karlnapf: Just do it in the eigen3 backend for now. See how the other things are added. Btw the linalg thing is a big mess inside. Maybe we can think about simplifying some things too during gsoc.

If possible I would just delete the row getter as it encourages bad things. And then possibly refactor the code that uses it? How many places do call it?

karlnapf: Oh and maybe name it just ‘scale’ as the signature tells the story already.

theartful:
sorry for the delay. I implemented the scale function in linalg and used it.

src/shogun/classifier/NearestCentroid.cpp
```diff
@@ -72,18 +72,16 @@ namespace shogun{
...
 			int32_t total = num_per_class[i];
-			float64_t scale = 0;
 			if(total>1)
-				scale = 1.0/((float64_t)(total-1));
+				scale[i] = 1.0/((float64_t)(total-1));
 			else
-				scale = 1.0/(float64_t)total;
-
-			target.scale(scale);
+				scale[i] = 1.0/(float64_t)total;
```
karlnapf: no need to loop this if all values are the same. Just use some kind of set_const call ...

src/shogun/classifier/NearestCentroid.cpp
```diff
@@ -72,18 +72,16 @@ namespace shogun{
...
 		}
+		linalg::scale(centroids, centroids, scale);
```
karlnapf: sweet!

karlnapf: i think your signature doesnt match the one in linalg where result is the last ?

src/shogun/mathematics/linalg/LinalgNamespace.h
```diff
@@ -1644,6 +1644,42 @@ namespace shogun
...
+		 * User should pass an appropriately pre-allocated memory matrix
+		 * Or pass the operands argument A as a result
+		 *
+		 * @param A First matrix
+		 * @param alpha Scale factor
+		 * @param result The scaled matrix
+		 */
+		template <typename T>
+		void scale(
+		    const SGMatrix<T>& A, SGMatrix<T>& result,
```
karlnapf: no it doesnt match with this one here

theartful: All the scale functions have differing signatures. The one in the `linalg` namespace is different from the one in the `linalgbackend`, so I just followed the same convention.

src/shogun/classifier/NearestCentroid.cpp
theartful: the values are not the same. `total = num_per_class[i]`

src/shogun/classifier/NearestCentroid.cpp
theartful: the signature of `scale` in `LinalgBackendBase` has the result at the end, while the one in `linalg` namespace has the scaling factor in the end.

src/shogun/mathematics/linalg/LinalgNamespace.h
karlnapf: i see, sorry about the noise then.

src/shogun/classifier/NearestCentroid.cpp
karlnapf: sorry I should have read that better :D

src/shogun/classifier/NearestCentroid.cpp
```diff
@@ -54,51 +53,41 @@ namespace shogun{
...
+		auto* dense_data = data->as<CDenseFeatures<float64_t>>();
+
+		int32_t num_classes = multiclass_labels->get_num_classes();
+		int32_t num_feats = dense_data->get_num_features();
 
-		for (int32_t idx=0 ; idx<num_vectors ; idx++)
+		SGMatrix<float64_t> centroids(num_feats, num_classes);
+		SGVector<int64_t> num_per_class(num_classes);
+
+		auto iter_labels = multiclass_labels->get_int_labels().begin();
```
karlnapf: This generates a copy I think. Is that necessary? We can also cast the labels on the fly below.

Also. This reminds me of another thing. I don't like this method, or any of the other "casting" methods in `CDenseLabels`. A good clean up PR would be

* remove all those methods and only keep `get_labels()` and `get_label(index_t)`.
* Instead, add a casting method `SGVector<T>::as<S>()` which returns a new vector with the converted data type
* Replace calls `labels->get_int_labels()` with the more explicit `labels->get_labels()->as<int32_t>()`

theartful:
@karlnapf I think this is ready. CI failed when "checking out", don't know why.

karlnapf:
Could you do a nop push to re-trigger the CI? I think we can merge otherwise

theartful:
@karlnapf All is green

src/shogun/classifier/NearestCentroid.cpp
theartful: `SGVector<T>::as<S>` is already implemented. Should I use it here? It creates a copy as well.

src/shogun/classifier/NearestCentroid.cpp
```diff
@@ -54,51 +53,41 @@ namespace shogun{
...
+
+		int32_t num_classes = multiclass_labels->get_num_classes();
+		int32_t num_feats = dense_data->get_num_features();
 
-		for (int32_t idx=0 ; idx<num_vectors ; idx++)
+		SGMatrix<float64_t> centroids(num_feats, num_classes);
+		SGVector<int64_t> num_per_class(num_classes);
+
+		auto iter_labels = multiclass_labels->get_labels().begin();
+		for (const auto& current : DotIterator(dense_data))
```
karlnapf: we should implement a zip-like iterator for iterating through labels and features at the same time...at some point

PR #4636, opened 2019-05-15
Move feature weights to CombinedFeature class

theartful:
This is to reduce the state of DotFeatures, and for better abstraction.

### DIFF
--- a/src/shogun/features/CombinedDotFeatures.cpp
+++ b/src/shogun/features/CombinedDotFeatures.cpp
@@ -13,7 +13,7 @@
 
 using namespace shogun;
 
-const float64_t CCombinedDotFeatures::initial_weight = 1.0;
+constexpr float64_t CCombinedDotFeatures::initial_weight = 1.0;
 
 CCombinedDotFeatures::CCombinedDotFeatures() : CDotFeatures()
 {
@@ -23,9 +23,10 @@ CCombinedDotFeatures::CCombinedDotFeatures() : CDotFeatures()
 	update_dim_feature_space_and_num_vec();
 }
 
-CCombinedDotFeatures::CCombinedDotFeatures(const CCombinedDotFeatures & orig)
-: CDotFeatures(orig), num_vectors(orig.num_vectors),
-	num_dimensions(orig.num_dimensions)
+CCombinedDotFeatures::CCombinedDotFeatures(const CCombinedDotFeatures& orig)
+    : CDotFeatures(orig), feature_array(orig.feature_array),
+      feature_weights(orig.feature_weights), num_vectors(orig.num_vectors),
+      num_dimensions(orig.num_dimensions)
 {
 	init();
 }
--- a/src/shogun/features/CombinedDotFeatures.cpp
+++ b/src/shogun/features/CombinedDotFeatures.cpp
@@ -141,14 +141,14 @@ void CCombinedDotFeatures::dense_dot_range(float64_t* output, int32_t start, int
 	uint32_t offs=0;
 	int32_t num=stop-start;
 	SGVector<float64_t> tmp(num);
-	std::fill(output, output+num, 0);
+	std::fill(output, output+num, b);
 
 	for (index_t f_idx=0; f_idx<get_num_feature_obj(); f_idx++)
 	{
 		CDotFeatures* f = get_feature_obj(f_idx);
 		int32_t f_dim = f->get_dim_feature_space();
 
-		f->dense_dot_range(tmp.vector, start, stop, alphas, vec+offs, f_dim, b);
+		f->dense_dot_range(tmp.vector, start, stop, alphas, vec+offs, f_dim, 0);
 		for (int32_t i=0; i<num; i++)
 			output[i] += get_subfeature_weight(f_idx) * tmp[i];
 
--- a/src/shogun/features/CombinedDotFeatures.cpp
+++ b/src/shogun/features/CombinedDotFeatures.cpp
@@ -28,7 +28,8 @@ CCombinedDotFeatures::CCombinedDotFeatures(const CCombinedDotFeatures& orig)
       feature_weights(orig.feature_weights), num_vectors(orig.num_vectors),
       num_dimensions(orig.num_dimensions)
 {
-	init();
+	register_params();
+	update_dim_feature_space_and_num_vec();
 }
 
 CFeatures* CCombinedDotFeatures::duplicate() const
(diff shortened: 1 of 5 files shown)


src/shogun/features/CombinedDotFeatures.cpp
```diff
@@ -11,6 +11,8 @@
 
 using namespace shogun;
 
+const float64_t CCombinedDotFeatures::initial_weight = 1.0;
```
gf712: Maybe just make it `constexpr`? Or does that only work in C++17?

theartful: Yeah that would be best. It works for c++11 as well.

gf712: ah ok, I think in C++17 what changes is that you can actually inline variables, so you could just have it the class declaration!

src/shogun/features/CombinedDotFeatures.cpp
```diff
@@ -309,32 +315,35 @@ int32_t CCombinedDotFeatures::get_nnz_features_for_vector(int32_t num) const
...
-	for (index_t f_idx=0; f_idx<num_weights; f_idx++)
-	{
-		CDotFeatures* f = get_feature_obj(f_idx);
-		weights[f_idx] = f->get_combined_feature_weight();
-		SG_UNREF(f);
-	}
 	return SGVector<float64_t>(weights,num_weights);
 }
 
 void CCombinedDotFeatures::set_subfeature_weights(SGVector<float64_t> weights)
```
gf712: right now the weights are passed by value, maybe pass by const ref? otherwise you are calling constructor right?

gf712: sorry this is unrelated to your PR, just noticed it

theartful: SGVector is passed by value, but the copy constructor only copies the pointer to the data. I think constant reference would be better though.

gf712: yup!

gf712: there are probably lots of other things like this in shogun so can probably leave it for another PR

src/shogun/features/CombinedDotFeatures.cpp
```diff
@@ -309,32 +315,35 @@ int32_t CCombinedDotFeatures::get_nnz_features_for_vector(int32_t num) const
 	return result;
 }
 
-SGVector<float64_t> CCombinedDotFeatures::get_subfeature_weights()
+SGVector<float64_t> CCombinedDotFeatures::get_subfeature_weights() const
```
karlnapf: Very good change this is! :)

src/shogun/features/CombinedDotFeatures.cpp
```diff
@@ -309,32 +315,35 @@ int32_t CCombinedDotFeatures::get_nnz_features_for_vector(int32_t num) const
...
 {
 	int32_t num_weights = get_num_feature_obj();
 	ASSERT(num_weights > 0)
 
 	float64_t* weights=SG_MALLOC(float64_t, num_weights);
```
karlnapf: maybe refactor this old school thing here as well? Allocate via SGVector ctor

src/shogun/features/CombinedDotFeatures.cpp
karlnapf: if you copy inside, then a const ref would be better. We use the copy by value for SG* for the reference counting magic

src/shogun/features/CombinedDotFeatures.cpp
```diff
@@ -347,5 +356,7 @@ void CCombinedDotFeatures::init()
 	SG_ADD(
 	    &num_vectors, "num_vectors", "Total number of vectors.");
 	SG_ADD(&feature_array, "feature_array", "Feature array.");
+	// FIXME
+	// SG_ADD(&feature_weights, "feature_weights", "Subfeature weights.");
```
karlnapf: whats here?

src/shogun/features/CombinedDotFeatures.h
```diff
@@ -15,6 +15,8 @@
 #include <shogun/lib/DynamicObjectArray.h>
 #include <shogun/features/DotFeatures.h>
 
+#include <vector>
```
karlnapf: please never include things like this is header files.... forward declare and include in .cpp file

src/shogun/features/CombinedDotFeatures.cpp
theartful: The weights vector should be added as parameter, but std::vector isn't supported yet.

karlnapf: I think it is! (Not via SG_ADD though, you have to use `watch_param`

theartful: yup, you're right

theartful:
I have no idea why gcc fails in the setup stage.

src/shogun/features/CombinedDotFeatures.cpp
theartful: you're right, it would need to be inlined, which would work only in C++17
https://dev.azure.com/shogunml/shogun/_build/results?buildId=1382

gf712: but you can still replace the const with constexpr right?

theartful: I tried just changing const to constexpr but it gave me compilation error
`constexpr static data member ‘initial_weight’ must have an initializer`
I also tried initializing it in the header file (which is ugly to my taste), it compiled locally but not with CI.

theartful: I don't fully understand constexpr stuff, so I might be doing something wrong

gf712: where did it fail in the CI? I just tried it in compiler explorer and it worked fine in C++14 (which it shouldn't).
constexpr has to be part of your declaration because it is evaluated at compile time, it's basically preprocessor with type safety. so you aren't "initialising" a variable, you are telling the compiler that is the value of the variable at runtime.

theartful: makes sense. I messed up the times before, I didn't change `const` to `constexpr` in the declaration.

karlnapf: so let's do it :)

theartful: sorry, was waiting to write up the tests first

src/shogun/features/CombinedDotFeatures.cpp
```diff
@@ -129,67 +135,53 @@ float64_t CCombinedDotFeatures::dense_dot(int32_t vec_idx1, const float64_t* vec
 
 void CCombinedDotFeatures::dense_dot_range(float64_t* output, int32_t start, int32_t stop, float64_t* alphas, float64_t* vec, int32_t dim, float64_t b) const
```
theartful: `dense_dot_range` and `dense_dot_range_subset` ignored the feature_weights, while other dot functions used it, so I had to change them before writing tests

src/shogun/features/CombinedDotFeatures.cpp
```diff
@@ -19,9 +23,10 @@ CCombinedDotFeatures::CCombinedDotFeatures() : CDotFeatures()
...
 
-CCombinedDotFeatures::CCombinedDotFeatures(const CCombinedDotFeatures & orig)
-: CDotFeatures(orig), num_vectors(orig.num_vectors),
-	num_dimensions(orig.num_dimensions)
+CCombinedDotFeatures::CCombinedDotFeatures(const CCombinedDotFeatures& orig)
+    : CDotFeatures(orig), feature_array(orig.feature_array),
+      feature_weights(orig.feature_weights), num_vectors(orig.num_vectors),
+      num_dimensions(orig.num_dimensions)
 {
 	init();
```
karlnapf: please note that aything that is initialised in `init` over-writes the initialization fields used above

src/shogun/features/CombinedDotFeatures.cpp
```diff
@@ -98,8 +103,8 @@ float64_t CCombinedDotFeatures::dot(int32_t vec_idx1, CDotFeatures* df, int32_t
 		ASSERT(f2)
 
 		result += f1->dot(vec_idx1, f2,vec_idx2) *
-			f1->get_combined_feature_weight() *
-			f2->get_combined_feature_weight();
+			get_subfeature_weight(f_idx) *
```
karlnapf: this stuff can also be done in batch matrix operations if no subset is active...something for a future PR, as mentioned in #4643

src/shogun/features/CombinedDotFeatures.cpp
```diff
@@ -129,67 +134,53 @@ float64_t CCombinedDotFeatures::dense_dot(int32_t vec_idx1, const float64_t* vec
 
 void CCombinedDotFeatures::dense_dot_range(float64_t* output, int32_t start, int32_t stop, float64_t* alphas, float64_t* vec, int32_t dim, float64_t b) const
 {
-	if (stop<=start)
-		return;
+	ASSERT(stop > start)
```
karlnapf: yes, much better!

src/shogun/features/CombinedDotFeatures.cpp
```diff
@@ -129,67 +134,53 @@ float64_t CCombinedDotFeatures::dense_dot(int32_t vec_idx1, const float64_t* vec
...
 	ASSERT(dim==num_dimensions)
 
 	uint32_t offs=0;
-	bool first=true;
 	int32_t num=stop-start;
-	float64_t* tmp=SG_MALLOC(float64_t, num);
-
+	
+	SGVector<float64_t> tmp(num);
```
karlnapf: `tmp.zero()` reads a bit easier
or you can also use linalg.
But I guess std::fill is also fine ... just saying :)

src/shogun/features/CombinedDotFeatures.cpp
```diff
@@ -309,32 +304,35 @@ int32_t CCombinedDotFeatures::get_nnz_features_for_vector(int32_t num) const
...
 	ASSERT(weights.vlen==get_num_feature_obj())
 
-	for (index_t f_idx=0; f_idx<get_num_feature_obj(); f_idx++)
-	{
-		CDotFeatures* f = get_feature_obj(f_idx);
-		f->set_combined_feature_weight(weights[f_idx]);
-		SG_UNREF(f);
-	}
+	std::copy(
+	    weights.vector, weights.vector + weights.vlen, feature_weights.begin());
```
karlnapf: why not do an assign -copy using `=` to reuse the memory here?

src/shogun/features/CombinedDotFeatures.cpp
```diff
@@ -309,32 +304,35 @@ int32_t CCombinedDotFeatures::get_nnz_features_for_vector(int32_t num) const
...
+
+float64_t CCombinedDotFeatures::get_subfeature_weight(index_t idx) const
+{
+	ASSERT(idx >= 0 && (size_t)idx < feature_weights.size())
+	return feature_weights[idx];
+}
+
+void CCombinedDotFeatures::set_subfeature_weight(index_t idx, float64_t weight)
+{
+	ASSERT(idx >= 0 && (size_t)idx < feature_weights.size())
```
karlnapf: if this is user-facing, it should have an error message that moans how the provided differs from the expected

src/shogun/features/CombinedDotFeatures.cpp
theartful: I used fill with `output`, which is a raw pointer. can't use `linalg` or `zero`

src/shogun/features/CombinedDotFeatures.cpp
theartful: sorry, not sure what you mean

karlnapf: I see, sorry I didnt get that. What about adding a ctor for that?

theartful:
@karlnapf sorry about polluting the push history with ci stuff. I added the unit tests.

PR #4640, opened 2019-05-16
Refactor KMeans/KMeansMiniBatch to make them observable.

geektoni:


src/shogun/clustering/KMeansBase.cpp
```diff
@@ -362,8 +314,10 @@ void CKMeansBase::init()
 	SG_ADD(&max_iter, "max_iter", "Maximum number of iterations", ParameterProperties::HYPER);
 	SG_ADD(&k, "k", "k, the number of clusters", ParameterProperties::HYPER);
 	SG_ADD(&dimensions, "dimensions", "Dimensions of data");
+	SG_ADD(&fixed_centers, "fixed_centers", "Use fixed centers");
 	SG_ADD(&R, "radiuses", "Cluster radiuses");
 	SG_ADD(&use_kmeanspp, "kmeanspp", "Whether use kmeans++", ParameterProperties::HYPER);
+	SG_ADD(&mus, "mus", "Cluster centers")
```
geektoni: @karlnapf I am a bit confused by this variable here. From what I could understand from the code, it should contain the actual centroids. However, I noticed that there is also a method called `get_cluster_centers`, which computes the centroids from the `distance`. Why is it that? Aren't they the same thing?

src/shogun/clustering/KMeansBase.cpp
```diff
@@ -134,6 +136,8 @@ void CKMeansBase::initialize_training(CFeatures* data)
 {
 	REQUIRE(distance, "Distance is not provided")
 	REQUIRE(distance->get_feature_type()==F_DREAL, "Distance's features type (%d) should be of type REAL (%d)")
+	REQUIRE(max_iter > 0, "The number of iterations must be greater than 0")
+	REQUIRE(k > 0, "The number of clusters must be greater than 0")
```
karlnapf: in those messages, always print what was provided by the user. also missing newline

src/shogun/clustering/KMeansBase.cpp
karlnapf: the cluster centers are stored in the lhs of some feature matrix. I think that is bad. Should be refactored at some point.
I think you can sidestep this via doing an explicit observe call on the result of that method. Once this is refactored and the cluster centers are just stored in a variable, I think it is better to observe them with a simple put, or?

src/shogun/clustering/KMeans.cpp
```diff
@@ -167,6 +167,9 @@ void CKMeans::Lloyd_KMeans(SGMatrix<float64_t> centers, int32_t num_centers)
 				}
 			}
 		}
+
+		observe<SGMatrix<float64_t>>(iter, "mus");
```
karlnapf: sorry, confused here. Why is this explicit observe call needed? Can't you just do "put". OR is this the same but more efficient?

src/shogun/clustering/KMeansBase.cpp
geektoni: :+1:

src/shogun/clustering/KMeans.cpp
geektoni: I used `observe()` here because it does not make sense to use `put()`. The method `Lloyd_KMeans` works directly on the `mus` variable, therefore I do not need to "put" it again (since it will be slower).

geektoni:
> Good!
> 
> With all those refactoring, could you always also paste a gist to a notebook that shows that it actually works? :) As we dont have any tests for that....

Okay, let me see what I can do.

src/shogun/clustering/KMeans.cpp
karlnapf: for now, let's keep the `observe`. For the future: refactor this as `CIterativeMachine` which automatically shall emit model parameters

examples/meta/src/observers/kmeans.sg
```diff
@@ -0,0 +1,28 @@
+Math:init_random(1)
+File f_feats_train = csv_file("../../data/classifier_binary_2d_linear_features_train.dat")
```
karlnapf: use nonlinear as that is clustered

examples/meta/src/observers/kmeans.sg
```diff
@@ -0,0 +1,28 @@
...
+
+#![create_features]
+Features features_train = features(f_feats_train)
+#![create_features]
+
+#![choose_distance]
+Distance d = distance("EuclideanDistance", lhs=features_train, rhs=features_train)
```
karlnapf: lhs and rhs shall not be modified like this. Instead use `.init`

geektoni:
This should be okay now. Once CIs become green it can be merged. @karlnapf any other thoughts?

tests/unit/clustering/kmeans_unittest.cc
```diff
@@ -11,9 +11,20 @@
 #include <shogun/clustering/KMeans.h>
 #include <shogun/clustering/KMeansMiniBatch.h>
 #include <shogun/distance/EuclideanDistance.h>
+#include <shogun/lib/observers/ParameterObserver.h>
+#include <shogun/lib/observers/ParameterObserverLogger.h>
 
 using namespace shogun;
 
+void check_consistency_observable(const CKMeans * kmeans, ParameterObserver * observer)
```
geektoni: @karlnapf I've updated one KMean unit test to check also if the observed values received are actually okay (such to avoid also to add a new copy-pasted test). Not sure if it is worth to update the other unit tests though...

tests/unit/clustering/kmeans_unittest.cc
```diff
@@ -11,9 +11,20 @@
...
+{
+	auto total_observations = observer->get<int32_t>("num_observations");
+	auto observation = observer->get_observation(total_observations-1);
+	auto centers = observation->get<SGMatrix<float64_t>>("mus");
+
+	EXPECT_TRUE(centers.equals(kmeans->get<SGMatrix<float64_t>>("cluster_centers")));
```
karlnapf: This only checks the last one?
Could you do a brief (internal, via print or so) check that the internal state matches the obs in every iteration and confirm?

This test is good though

geektoni: Yes, I'm checking the last one just to be sure that both the observer and model reach the same final values. Anyway, the internal state and the one observed by the observer match. We may need a better way to do these tests, since it would be really nice to have unit-tests that check the correctness of the observed values :)

karlnapf: it is hard to do this in tests without modifying the algorithm code

BUT what we can do is this: you write a little MockMachine that has some very minimal state changes in the training loop, and then you observe those. Then we can assert the observations in a unit test. We should do this for primitive, SG*, and objects (where object observations are made of a single objects whose state is changed in the training loop in the mock machine)

karlnapf:
ok to merge

PR #4641, opened 2019-05-16
Make observe() no op when there are no observers attached.

geektoni:


src/shogun/base/SGObject.h
```diff
@@ -1015,6 +1015,10 @@ class CSGObject
 	template <class T>
 	void observe(const int64_t step, const std::string& name) const;
 
+	template <class T>
+	void emit_observation(const int64_t step, const std::string& name, const T& value,
```
karlnapf: the name is a bit weird.... why not "observe"? Also this needs docs. Also shouldnt it be private?

geektoni: Yeah,  `observe` is a better name. Moreover, this cannot be private otherwise we won't be able to call it from derived classes.

karlnapf:
CI is not happy

geektoni:
@karlnapf  This should also be okay now. CIs keeps failing on MacOS MKL but it is still unrelated to this patch.

PR #4642, opened 2019-05-17
Fix bug in dense_dot_range

theartful:
This fixes a buffer underflow bug in `dense_dot_range`.

### DIFF
--- a/src/shogun/features/DotFeatures.cpp
+++ b/src/shogun/features/DotFeatures.cpp
@@ -51,9 +51,6 @@ float64_t CDotFeatures::dense_dot_sgvec(int32_t vec_idx1, SGVector<float64_t> ve
 void CDotFeatures::dense_dot_range(float64_t* output, int32_t start, int32_t stop, float64_t* alphas, float64_t* vec, int32_t dim, float64_t b) const
 {
 	ASSERT(output)
-	// write access is internally between output[start..stop] so the following
-	// line is necessary to write to output[0...(stop-start-1)]
-	output-=start;
 	ASSERT(start>=0)
 	ASSERT(start<stop)
 	ASSERT(stop<=get_num_vectors())
@@ -81,7 +78,7 @@ void CDotFeatures::dense_dot_range(float64_t* output, int32_t start, int32_t sto
 #endif
 
 		int32_t t_start=thread_num*step;
-		int32_t t_stop=(thread_num==num_threads) ? stop : (thread_num+1)*step;
+		int32_t t_stop=(thread_num==num_threads) ? num_vectors : (thread_num+1)*step;
 
 #ifdef WIN32
 		for (int32_t i=t_start; i<t_stop; i++)
@@ -93,9 +90,9 @@ void CDotFeatures::dense_dot_range(float64_t* output, int32_t start, int32_t sto
 #endif
 		{
 			if (alphas)
-				output[i]=alphas[i]*this->dense_dot(i, vec, dim)+b;
+				output[i]=alphas[i]*this->dense_dot(i + start, vec, dim)+b;
 			else
-				output[i]=this->dense_dot(i, vec, dim)+b;
+				output[i]=this->dense_dot(i + start, vec, dim)+b;
 			pb.print_progress();
 		}
 	}
--- a/tests/unit/features/DotFeatures_unittest.cc
+++ b/tests/unit/features/DotFeatures_unittest.cc
@@ -121,3 +121,47 @@ TEST_F(DotFeaturesTest, compute_cov_nocopy)
 	for (index_t i = 0; i < (index_t)cov.size(); ++i)
 		EXPECT_NEAR(cov[i], ref_cov_ab[i], eps);
 }
+
+TEST_F(DotFeaturesTest, dense_dot_range)
+{
+	index_t num_feats = 2;
+	index_t num_vectors = 4;
+	float64_t bias = 25;
+
+	SGMatrix<float64_t> data(num_feats, num_vectors);
+	for (index_t i = 0; i < num_vectors; i++)
+		data.get_column(i).set_const(i);
+	auto feats = some<CDenseFeatures<float64_t>>(data);
+
+	SGVector<float64_t> vec(num_feats);
+	vec.range_fill(1);
+
+	index_t start1 = 0;
+	index_t stop1 = num_vectors - 1;
+	index_t start2 = 1;
+	index_t stop2 = num_vectors;
+
+	SGVector<float64_t> alphas(num_vectors - 1);
+	alphas.range_fill(0);
+
+	SGVector<float64_t> output1(num_vectors - 1);
+	SGVector<float64_t> output2(num_vectors - 1);
+
+	feats->dense_dot_range(
+	    output1.vector, start1, stop1, alphas.vector, vec.vector, num_feats,
+	    bias);
+	feats->dense_dot_range(
+	    output2.vector, start2, stop2, alphas.vector, vec.vector, num_feats,
+	    bias);
+
+	for (index_t i = 0; i < num_vectors - 1; i++)
+	{
+		// output1[i] = alpha[i] * (data[i] dot vec) + bias
+		ASSERT_EQ(
+		    output1[i], i * i * num_feats * ((1 + num_feats) / 2.0) + bias);
+		// output2[i] = alpha[i] * (data[i+1] dot vec) + bias
+		ASSERT_EQ(
+		    output2[i],
+		    i * (i + 1) * num_feats * ((1 + num_feats) / 2.0) + bias);
+	}
+}


src/shogun/features/DotFeatures.cpp
```diff
@@ -51,9 +51,6 @@ float64_t CDotFeatures::dense_dot_sgvec(int32_t vec_idx1, SGVector<float64_t> ve
 void CDotFeatures::dense_dot_range(float64_t* output, int32_t start, int32_t stop, float64_t* alphas, float64_t* vec, int32_t dim, float64_t b) const
 {
 	ASSERT(output)
-	// write access is internally between output[start..stop] so the following
-	// line is necessary to write to output[0...(stop-start-1)]
-	output-=start;
```
theartful: output was assumed to be accessed from `start` to `stop`, which wasn't the case.

src/shogun/features/DotFeatures.cpp
```diff
@@ -81,7 +78,7 @@ void CDotFeatures::dense_dot_range(float64_t* output, int32_t start, int32_t sto
 #endif
 
 		int32_t t_start=thread_num*step;
```
theartful: `t_start` starts from 0

src/shogun/features/DotFeatures.cpp
```diff
@@ -93,9 +90,9 @@ void CDotFeatures::dense_dot_range(float64_t* output, int32_t start, int32_t sto
 #endif
 		{
 			if (alphas)
-				output[i]=alphas[i]*this->dense_dot(i, vec, dim)+b;
```
theartful: and here happens buffer underflow if `t_start != 0`

src/shogun/features/DotFeatures.cpp
```diff
@@ -81,7 +78,7 @@ void CDotFeatures::dense_dot_range(float64_t* output, int32_t start, int32_t sto
 #endif
 
 		int32_t t_start=thread_num*step;
-		int32_t t_stop=(thread_num==num_threads) ? stop : (thread_num+1)*step;
+		int32_t t_stop=(thread_num==num_threads) ? num_vectors : (thread_num+1)*step;
 
 #ifdef WIN32
 		for (int32_t i=t_start; i<t_stop; i++)
```
iglesias: Can the TODO below be removed?

src/shogun/features/DotFeatures.cpp
```diff
@@ -93,9 +90,9 @@ void CDotFeatures::dense_dot_range(float64_t* output, int32_t start, int32_t sto
 #endif
 		{
 			if (alphas)
-				output[i]=alphas[i]*this->dense_dot(i, vec, dim)+b;
+				output[i]=alphas[i]*this->dense_dot(i + start, vec, dim)+b;
```
iglesias: I think `start` is included earlier.

To make sure everything is fine, have you tested if with different number of threads?

tests/unit/features/DotFeatures_unittest.cc
```diff
@@ -121,3 +121,47 @@ TEST_F(DotFeaturesTest, compute_cov_nocopy)
 	for (index_t i = 0; i < (index_t)cov.size(); ++i)
 		EXPECT_NEAR(cov[i], ref_cov_ab[i], eps);
 }
+
+TEST_F(DotFeaturesTest, dense_dot_range)
+{
```
iglesias: Thanks for including the test!
What about including also a much more trivial one, whose expected result could be seen just from reading the test code in a second?
Also, I think it would be good to ensure test coverage with different number of threads. Is that currently happening?

src/shogun/features/DotFeatures.cpp
theartful: `i` ranges from `t_start` to `t_stop`. But `t_start` begins from `0` and `t_stop` goes up to `stop`.
It should be that `t_start` begins from `0` and `t_stop` goes up to `stop - start`, or `t_start` begins from `start` and `t_stop` goes up to `stop`.

theartful: Judging from this comment
``` C++
// write access is internally between output[start..stop] so the following
// line is necessary to write to output[0...(stop-start-1)]
output-=start;
```
This was initially written so that we go from `start` to  `end`, so we subtract `start` from the output, so it transforms to `0` to `end - start`.

theartful: This assumption didn't hold, as `i` ranges from `0` to `t_stop`
```C++
int32_t t_start=thread_num*step;
int32_t t_stop=(thread_num==num_threads) ? stop : (thread_num+1)*step;
```
And even if it did, the indexing of `alpha` would be wrong.

tests/unit/features/DotFeatures_unittest.cc
theartful: I believe it should run concurrently on any machine with `openmp`

src/shogun/features/DotFeatures.cpp
theartful: Should we also remove the progress bar? I don't think such an operation needs it.

src/shogun/features/DotFeatures.cpp
theartful: Should I revert the changes and write a simple test that fails?

src/shogun/features/DotFeatures.cpp
iglesias: Why not? It could be useful if many vectors dotted.

src/shogun/features/DotFeatures.cpp
iglesias: That sounds like a good idea. You can add it to another branch if that's more efficient than reverting. Link to test results or copy+paste suffices if that is faster as well.

`t_start` starts on a different value depending on the thread.

tests/unit/features/DotFeatures_unittest.cc
iglesias: Let me rephrase, is there some test exercising this code with more than one thread?
Alternatively, have you tested it locally with different number of threads (at least one case with more than one).

src/shogun/features/DotFeatures.cpp
theartful: I ran this test on the old version, which tests the underflow:
```C++
TEST_F(DotFeaturesTest, dense_dot_range)
{
	index_t num_feats = 2;
	index_t num_vectors = 4;

	SGMatrix<float64_t> data(num_feats, num_vectors);
	for (index_t i = 0; i < num_vectors; i++)
		data.get_column(i).set_const(5);
	auto feats = some<CDenseFeatures<float64_t>>(data);

	SGVector<float64_t> vec(num_feats);
	vec.set_const(3);

	index_t start = 1;
	index_t stop = num_vectors;
	
	float64_t* buffer = new float64_t[num_vectors];
	float64_t& gaurd = buffer[0];
	gaurd = 0;
	float64_t* output = buffer  + 1;

	feats->dense_dot_range(
	    output, start, stop, nullptr, vec.vector, num_feats, 0);

	ASSERT_EQ(gaurd, 0);
}
```
Here `gaurd` should not be changed at all.
These are the results:
```
$ ./shogun-unit-test --gtest_filter=DotFeaturesTest.dense_dot_range
Note: Google Test filter = DotFeaturesTest.dense_dot_range
[==========] Running 1 test from 1 test case.
[----------] 1 test from DotFeaturesTest
[ RUN      ] DotFeaturesTest.dense_dot_range
/home/theartful/Programming/shogun/tests/unit/features/DotFeatures_unittest.cc:149: Failure
      Expected: gaurd
      Which is: 30
To be equal to: 0
[  FAILED  ] DotFeaturesTest.dense_dot_range (12 ms)
[----------] 1 test from DotFeaturesTest (12 ms total)

[==========] 1 test from 1 test case ran. (15 ms total)
[  PASSED  ] 0 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] DotFeaturesTest.dense_dot_range

 1 FAILED TEST
```

tests/unit/features/DotFeatures_unittest.cc
theartful: Locally `openmp` spawns 4 threads on my machine to do the operation.

src/shogun/features/DotFeatures.cpp
theartful: But does the user need to see them? I think the progress bar would make sense if it indicated the state of the algorithm, instead of an operation.

theartful: About the TODO. Do you mean remove it with the signal stuff, or implement it? If so, I would need to first check out what are these signals about :"D

src/shogun/features/DotFeatures.cpp
iglesias: Memory leak with `buffer`?
In any case, use std::vector instead.

theartful: The test leaks indeed. Freeing the memory doesn't change the test result. The problem is that the old `dense_dot_range` changes `output[-1]`

iglesias: All right, the point is clear now. Thanks for writing this test, good illustration.

PR #4643, opened 2019-05-20
Make kernel normalizer functions const

theartful:
This makes the kernel normalizer interface constant, which will be useful when attempting to make the kernel interface stateless. It also changes accessibility of `alloc_and_compute_diag` in some normalizers from public to protected, since it's only a helper function used in initialization.

### DIFF
--- a/src/shogun/kernel/normalizer/KernelNormalizer.h
+++ b/src/shogun/kernel/normalizer/KernelNormalizer.h
@@ -71,20 +71,28 @@ class CKernelNormalizer : public CSGObject
 		 * @param idx_rhs index of right hand side vector
 		 */
 		virtual float64_t normalize(
-			float64_t value, int32_t idx_lhs, int32_t idx_rhs)=0;
+			float64_t value, int32_t idx_lhs, int32_t idx_rhs) const=0;
 
 		/** normalize only the left hand side vector
 		 * @param value value of a component of the left hand side feature vector
 		 * @param idx_lhs index of left hand side vector
 		 */
-		virtual float64_t normalize_lhs(float64_t value, int32_t idx_lhs)=0;
+		virtual float64_t normalize_lhs(float64_t value, int32_t idx_lhs) const=0;
 
 		/** normalize only the right hand side vector
 		 * @param value value of a component of the right hand side feature vector
 		 * @param idx_rhs index of right hand side vector
 		 */
-		virtual float64_t normalize_rhs(float64_t value, int32_t idx_rhs)=0;
+		virtual float64_t normalize_rhs(float64_t value, int32_t idx_rhs) const=0;
 
+		/** getter for normalizer type
+		 */
+		ENormalizerType get_normalizer_type() const noexcept
+		{
+			return m_type;
+		}
+
+	protected:
 		/** register the parameters
 		 */
 		virtual void register_params()
@@ -92,13 +100,6 @@ class CKernelNormalizer : public CSGObject
 			SG_ADD_OPTIONS(
 			    (machine_int_t*)&m_type, "m_type", "Normalizer type.",
 			    ParameterProperties::NONE, SG_OPTIONS(N_REGULAR, N_MULTITASK));
-		}
-
-		/** getter for normalizer type
-		 */
-		ENormalizerType get_normalizer_type()
-		{
-			return m_type;
 		}
 
 		/** setter for normalizer type
@@ -109,7 +110,6 @@ class CKernelNormalizer : public CSGObject
 			m_type = type;
 		}
 
-	protected:
 		/** normalizer type */
 		ENormalizerType m_type;
 };
--- a/src/shogun/kernel/normalizer/ScatterKernelNormalizer.h
+++ b/src/shogun/kernel/normalizer/ScatterKernelNormalizer.h
@@ -77,7 +77,7 @@ public:
 	 *
 	 * @return testing class (-1 disabled, 0...class otherwise)
 	 */
-	int32_t get_testing_class()
+	int32_t get_testing_class() const noexcept
 	{
 		return m_testing_class;
 	}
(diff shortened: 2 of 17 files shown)


src/shogun/kernel/normalizer/KernelNormalizer.h
```diff
@@ -71,20 +71,28 @@ class CKernelNormalizer : public CSGObject
...
 		/** normalize only the right hand side vector
 		 * @param value value of a component of the right hand side feature vector
 		 * @param idx_rhs index of right hand side vector
 		 */
-		virtual float64_t normalize_rhs(float64_t value, int32_t idx_rhs)=0;
+		virtual float64_t normalize_rhs(float64_t value, int32_t idx_rhs) const=0;
 
+		/** getter for normalizer type
+		 */
+		ENormalizerType get_normalizer_type() const
```
gf712: noexcept?

theartful: done

src/shogun/kernel/normalizer/ScatterKernelNormalizer.h
```diff
@@ -77,7 +77,7 @@ class CScatterKernelNormalizer: public CKernelNormalizer
 	 *
 	 * @return testing class (-1 disabled, 0...class otherwise)
 	 */
-	int32_t get_testing_class()
+	int32_t get_testing_class() const
```
gf712: noexcept

src/shogun/transfer/multitask/MultitaskKernelMaskNormalizer.h
```diff
@@ -222,7 +222,7 @@ class CMultitaskKernelMaskNormalizer: public CKernelNormalizer
 	/**
 	 * @return list of active task ids
 	 */
-	std::vector<int32_t> get_active_tasks()
+	std::vector<int32_t> get_active_tasks() const
```
gf712: noexcept

src/shogun/transfer/multitask/MultitaskKernelMaskNormalizer.h
```diff
@@ -202,7 +202,7 @@ class CMultitaskKernelMaskNormalizer: public CKernelNormalizer
 	 * @param task_rhs task_id on right hand side
 	 * @return similarity between tasks
 	 */
-	float64_t get_similarity(int32_t task_lhs, int32_t task_rhs)
+	float64_t get_similarity(int32_t task_lhs, int32_t task_rhs) const
```
gf712: noexcept

src/shogun/transfer/multitask/MultitaskKernelNormalizer.h
```diff
@@ -99,7 +99,7 @@ class CMultitaskKernelNormalizer: public CKernelNormalizer
 	 * @param vec vector with containing task_id for each example
 	 * @return number of unique task ids
 	 */
-	int32_t get_num_unique_tasks(std::vector<int32_t> vec) {
+	int32_t get_num_unique_tasks(std::vector<int32_t> vec) const {
```
gf712: noexcept

src/shogun/transfer/multitask/MultitaskKernelMaskNormalizer.h
theartful: this function creates an `std::vector` and uses `push_back` which can throw exception

karlnapf:
> sorry about the noexcept spam, it might be worth adding it wherever you can. I am not sure if you can with the ones I commented on, it just makes sense to have const noexcept getters

I am not so sure about this, see e.g. [here](https://stackoverflow.com/questions/21330807/should-i-use-noexcept-for-getters-always).

karlnapf:
I mean in case of the returning int it is probably ok. But the question is about the benefit is still open ...

src/shogun/transfer/multitask/MultitaskKernelMaskPairNormalizer.h
```diff
@@ -188,7 +188,7 @@ class CMultitaskKernelMaskPairNormalizer: public CKernelNormalizer
 	 * @param task_rhs task_id on right hand side
 	 * @return similarity between tasks
 	 */
-	float64_t get_similarity(int32_t task_lhs, int32_t task_rhs)
+	float64_t get_similarity(int32_t task_lhs, int32_t task_rhs) const noexcept
 	{
```
karlnapf: e.g. is the noexcept a good idea here?

theartful: The function doesn't seem to do anything that would throw an exception, but it's probably a bad decision. Should've stuck with simple getters :"D

karlnapf: returning ints is fine, but the problem is that "noexept is a promise that is very hard to take back" as mentioned in the stackoverflow post. So while no exception is thrown here, maybe that changes at some point, since it is a nontrivial function

gf712:
> I mean in case of the returning int it is probably ok. But the question is about the benefit is still open 

If something is noexcept i would assume it gets inlined and is in the hot path and there are no jumps. For primitive types it is definitely worth it. For containers, i.e. std::string it could sometimes cause issues when you're in a memory limited environment, but it's not like anyone catches errors from getters, no? 

I think in general make const noexcept unless you expect it to throw, the benefit is significant mostly if it's something that runs in a loop and the compiler is not doing a great job optimising on its own.

src/shogun/transfer/multitask/MultitaskKernelMaskPairNormalizer.h
gf712: this might be a bit too far, but still unless you get  bad allocation error I don't see what else would throw?

iglesias:
In addition to the modern syntax 😉 the benefit of `noexcept` I am familiar with has to do with sprinkling it over constructors of custom types to enable move semantics in standard containers.

https://en.cppreference.com/w/cpp/language/noexcept_spec (notes section)

iglesias:
Ready to merge? @gf712

gf712:
Yup, it’s fine with me https://isocpp.org/blog/2014/01/quick-q-should-property-get-functions-be-noexcept-stackoverflow

iglesias:
Do you guys know what's up with the tests in MacOS MKL?

gf712:
Not yet... the only difference is the mkl dependency and that hasn’t changed for a couple of months..

PR #4645, opened 2019-05-23
[WIP] Refactor again ObservedValue/ObservedValueTemplated and put()/observe().

geektoni:
1. Move `ObservedValue`/`ObservedValueTemplated` out from `SGObject.h`;
2. Remove `observe()` from `put()` call;
3. Replace all the occurrencies of `put()` with `observe()` inside observable algorithms;

### DIFF
--- a/src/shogun/base/SGObject.h
+++ b/src/shogun/base/SGObject.h
@@ -46,7 +46,8 @@ class ObservedValue;
 class ParameterObserver;
 class CDynamicObjectArray;
 
-template <class T> class ObservedValueTemplated;
+template <class T>
+class ObservedValueTemplated;
 
 #ifndef SWIG
 #ifndef DOXYGEN_SHOULD_SKIP_THIS
@@ -141,13 +142,13 @@ public:
 	typedef rxcpp::subjects::subject<Some<ObservedValue>> SGSubject;
 	/** Definition of observable */
 	typedef rxcpp::observable<Some<ObservedValue>,
-			rxcpp::dynamic_observable<Some<ObservedValue>>>
-			SGObservable;
+		                      rxcpp::dynamic_observable<Some<ObservedValue>>>
+		SGObservable;
 	/** Definition of subscriber */
 	typedef rxcpp::subscriber<
-			Some<ObservedValue>,
-			rxcpp::observer<Some<ObservedValue>, void, void, void, void>>
-			SGSubscriber;
+		Some<ObservedValue>,
+		rxcpp::observer<Some<ObservedValue>, void, void, void, void>>
+		SGSubscriber;
 
 	/** default constructor */
 	CSGObject();
@@ -378,8 +379,8 @@ public:
 			if (!parameter_value.cloneable())
 			{
 				SG_ERROR(
-						"Cannot put parameter %s::%s.\n", get_name(),
-						_tag.name().c_str());
+					"Cannot put parameter %s::%s.\n", get_name(),
+					_tag.name().c_str());
 			}
 			try
 			{
@@ -388,10 +389,10 @@ public:
 			catch (const TypeMismatchException& exc)
 			{
 				SG_ERROR(
-						"Cannot put parameter %s::%s of type %s, incompatible "
-								"provided type %s.\n",
-						get_name(), _tag.name().c_str(), exc.actual().c_str(),
-						exc.expected().c_str());
+					"Cannot put parameter %s::%s of type %s, incompatible "
+					"provided type %s.\n",
+					get_name(), _tag.name().c_str(), exc.actual().c_str(),
+					exc.expected().c_str());
 			}
 			ref_value(value);
 			update_parameter(_tag, make_any(value));
@@ -399,11 +400,10 @@ public:
 		else
 		{
 			SG_ERROR(
-					"Parameter %s::%s does not exist.\n", get_name(),
-					_tag.name().c_str());
+				"Parameter %s::%s does not exist.\n", get_name(),
+				_tag.name().c_str());
 		}
 	}
-
 
 	/** Setter for a class parameter that has values of type string,
 	 * identified by a Tag.
@@ -1072,7 +1072,7 @@ protected:
 	 * Observe a parameter value, given a pointer.
 	 * @param value Observed parameter's value
 	 */
-	void observe(ObservedValue * value) const;
+	void observe(ObservedValue* value) const;
 
 	/**
 	 * Observe a parameter value given custom properties for the Any.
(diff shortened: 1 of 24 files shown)


src/shogun/base/SGObject.h
```diff
@@ -1044,7 +1085,14 @@ class CSGObject
 	 */
 	template <class T>
 	void observe(const int64_t step, const std::string& name, const T& value,
-						  const AnyParameterProperties properties) const;
+						  const AnyParameterProperties properties) const
+	{
+		// If there are no observers attached, do not create/emit anything.
+		if (get_num_subscriptions() == 0) return;
+
+		auto obs = new ObservedValueTemplated<T>(step, name, static_cast<T>(clone_utils::clone(value)), properties);
```
theartful: First off, very cool :"D
I have a question, how does this work? I thought that you can't call the constructor of a class that wasn't included.

geektoni: It is using a forward declaration (https://en.wikipedia.org/wiki/Forward_declaration). Basically, I can instantiate a pointer to that object here in the `.h` file and everything work. However, when I'll need to use the `observe()` function (inside a `.cpp` file), I will have to include the `ObservedValueTemplated.h` file to let the compiler know how that class is structured.

theartful: Sorry this is not related to the PR, I'm just curious since I had a similar problem. 
But I believe a forward declared class is an incomplete type, and it shouldn't be possible to use any function of it including the constructor (it's obviously possible, hence the code compiles correctly) unless the header is included in the same place it is used. Like in this link: https://stackoverflow.com/questions/35263459/class-has-no-constructor-forward-declaration

geektoni: This is exactly what happens :) e.g. if you look at `SGObject.h`, there are some functions which uses `new ObservedValueTemplated()`. However, in each `.cpp` file in which those functions are used, you can see that there we include both the `SGObject.h` header (it can be included indirectly sometimes) and the `ObservedValueTemplated.h` header. Therefore, the compiler will have both the method and the class definition, and thus it will compile correctly.

theartful: Oh, right, that makes sense. Thanks :")

src/shogun/base/SGObject.h
```diff
@@ -46,6 +46,8 @@ class ObservedValue;
 class ParameterObserver;
 class CDynamicObjectArray;
 
+template <class T> class ObservedValueTemplated;
```
karlnapf: yippie! :)

src/shogun/lib/observers/ObservedValue.cpp
```diff
@@ -0,0 +1,18 @@
...
+#include <shogun/base/Parameter.h>
+#include <shogun/lib/observers/ObservedValue.h>
+
+using namespace shogun;
+
+ObservedValue::ObservedValue(const int64_t step, const std::string& name)
+		: CSGObject(), m_step(step), m_name(name), m_any_value(Any())
+{
+	SG_ADD(&m_step, "step", "Step");
+	this->watch_param(
```
karlnapf: shall we make those read-only?

src/shogun/lib/observers/ObservedValue.h
```diff
@@ -0,0 +1,61 @@
...
+
+namespace shogun {
+
+	template<class T>
+	class ObservedValueTemplated;
+
+	/**
+ 	* Observed value which is emitted by algorithms.
+ 	*/
+	class ObservedValue : public CSGObject {
```
karlnapf: cool, after this PR is in, @theartful can try to split the params as a mixin and we can try to remove the base class here

src/shogun/classifier/Perceptron.cpp
```diff
@@ -73,9 +74,11 @@ void CPerceptron::iteration()
 		{
 			converged = false;
 			const auto gradient = learn_rate * true_label;
-			put("bias", bias + gradient);
+			bias += gradient;
 			v.add(gradient, w);
-			put("w", w);
+
+			observe<SGVector<float64_t>>(m_current_iteration, "w");
+			observe<float64_t>(m_current_iteration, "bias");
```
karlnapf: question. If the step is passed here, then we don't need it inside SGObject anymore right? So that problem is also solved?

src/shogun/base/SGObject.h
geektoni: :dancing_women:

src/shogun/classifier/Perceptron.cpp
geektoni: Yes, the problem is solved since we will need to specify the `step` param directly from now on (and we won't need to infer it).

src/shogun/lib/observers/ObservedValue.cpp
geektoni: It seems reasonable to me :)

geektoni:
No, I've never tried to merge it inside develop. I guess it should be fine, though. It could be slightly more difficult if, let's say, @vigsterkr's PR about smart pointers goes in since there may be merge conflicts, but it should be pretty straightforward :)

karlnapf:
Let’s keep the other pr updated once this one here is merged into the feature branch and then discuss there.
I don’t have any objections with this one now

geektoni:
Any comments on this one? @vigsterkr @gf712 

Otherwise, I'll merge it straight away and we can start the review in the "final" PR #4653.

src/shogun/lib/observers/ObservedValue.h
```diff
@@ -0,0 +1,61 @@
...
+		 */
+		~ObservedValue() {};
+
+#ifndef SWIG
+
+		/**
+		* Return a any version of the stored type.
+		* @return the any value.
+		*/
+		virtual Any get_any() const {
```
gf712: maybe `const Any&` instead to avoid a copy?

src/shogun/lib/observers/ObservedValueTemplated.h
```diff
@@ -0,0 +1,66 @@
...
+#include <shogun/lib/observers/ObservedValue.h>
+
+namespace shogun {
+
+	/**
+ 	* Templated specialisation of ObservedValue that stores the actual data.
+ 	* @tparam T the type of the observed value
+ 	*/
+	template<class T>
+	class ObservedValueTemplated : public ObservedValue {
```
gf712: I am not sure I understand why you have a templated class though? Why do you need to have a template and the Any value? I think I am missing something :)

geektoni: The templated version is needed to be able to register the observed value within the parameter framework. The `Any` value is needed because there are situations in which we need to use the `sg_any_dispatch` method to retrieve the observed value. Therefore I had to add both functionalities to the classes :)

gf712: ah ok yes, I see!

src/shogun/lib/observers/ObservedValue.h
karlnapf: +1

PR #4647, opened 2019-05-24
Constrain parameter values

gf712:
Here is a design I came up with to constrain parameter values. At compile time we know the constraints so it can all be wrapped into a tuple that is "evaluated" and should be inlined properly at O3. The issue is the casting from any to the respective type, which is expensive, but then we can "ignore" the parameter type and have something like this:
```cpp
SG_ADD(&k, "k", "k, the number of clusters",
        ParameterProperties::HYPER | ParameterProperties::CONSTRAIN,
	SG_CONSTRAINT(positive<>()));
```
and then python
```python
In [1]: import shogun as sg                                                                                                                                                                                 

In [2]: kmeans = sg.machine("KMeans")                                                                                                                                                                       

In [3]: kmeans.put("k", -1)                                                                                                                                                                                 
<ipython-input-3-b598556376d3> in <module>
----> 1 kmeans.put("k", -1)

SystemError: [ERROR] In file /Users/ghoben/shogun/src/shogun/base/SGObject.cpp line 753: KMeans::k cannot be updated because of its constraint!
```

### DIFF
--- a/src/shogun/base/AnyParameter.h
+++ b/src/shogun/base/AnyParameter.h
@@ -37,7 +37,8 @@ namespace shogun
 	        {ParameterProperties::MODEL, "MODEL"},
 	        {ParameterProperties::AUTO, "AUTO"},
 	        {ParameterProperties::READONLY, "READONLY"},
-	        {ParameterProperties::RUNFUNCTION, "RUNFUNCTION"}};
+	        {ParameterProperties::RUNFUNCTION, "RUNFUNCTION"},
+	        {ParameterProperties::CONSTRAIN, "CONSTRAIN"}};
 
 	enableEnumClassBitmask(ParameterProperties);
 
@@ -87,11 +88,11 @@ namespace shogun
 		{
 			return static_cast<bool>(m_attribute_mask & other);
 		}
-		bool compare_mask(const ParameterProperties other) const
+		bool compare_mask(ParameterProperties other) const
 		{
 			return m_attribute_mask == other;
 		}
-		void remove_property(const ParameterProperties other)
+		void remove_property(ParameterProperties other)
 		{
 			m_attribute_mask &= ~other;
 		}
@@ -139,9 +140,9 @@ namespace shogun
 		{
 		}
 		AnyParameter(
-		    const Any& value, AnyParameterProperties properties,
-		    std::function<bool(Any)> constrain_function)
-		    : m_value(value), m_properties(properties),
+		    Any&& value, const AnyParameterProperties& properties,
+		    std::function<std::string(Any)> constrain_function)
+		    : m_value(std::move(value)), m_properties(properties),
 		      m_constrain_function(std::move(constrain_function))
 		{
 		}
@@ -172,12 +173,13 @@ namespace shogun
 			return m_properties;
 		}
 
-		std::shared_ptr<params::AutoInit> get_init_function() const
+		const std::shared_ptr<params::AutoInit>& get_init_function() const
 		{
 			return m_init_function;
 		}
 
-		const std::function<bool(Any)>& get_constrain_function() const noexcept
+		const std::function<std::string(Any)>& get_constrain_function() const
+		    noexcept
 		{
 			return m_constrain_function;
 		}
@@ -199,7 +201,7 @@ namespace shogun
 		Any m_value;
 		AnyParameterProperties m_properties;
 		std::shared_ptr<params::AutoInit> m_init_function;
-		std::function<bool(Any)> m_constrain_function;
+		std::function<std::string(Any)> m_constrain_function;
 	};
 } // namespace shogun
 
--- a/src/shogun/base/AnyParameter.h
+++ b/src/shogun/base/AnyParameter.h
@@ -25,7 +25,8 @@ namespace shogun
 		MODEL = 1u << 2,
 		AUTO = 1u << 10,
 		READONLY = 1u << 11,
-		RUNFUNCTION = 1u << 12
+		RUNFUNCTION = 1u << 12,
+		CONSTRAIN = 1u << 13
 	};
 
 	static const std::list<std::pair<ParameterProperties, std::string>>
@@ -137,9 +138,17 @@ namespace shogun
 		      m_init_function(std::move(auto_init))
 		{
 		}
+		AnyParameter(
+		    const Any& value, AnyParameterProperties properties,
+		    std::function<bool(Any)> constrain_function)
+		    : m_value(value), m_properties(properties),
+		      m_constrain_function(std::move(constrain_function))
+		{
+		}
 		AnyParameter(const AnyParameter& other)
 		    : m_value(other.m_value), m_properties(other.m_properties),
-		      m_init_function(other.m_init_function)
+		      m_init_function(other.m_init_function),
+		      m_constrain_function(other.m_constrain_function)
 		{
 		}
 
@@ -168,6 +177,11 @@ namespace shogun
 			return m_init_function;
 		}
 
+		const std::function<bool(Any)>& get_constrain_function() const noexcept
+		{
+			return m_constrain_function;
+		}
+
 		/** Equality operator which compares value but not properties.
 		 * @return true if value of other parameter equals own */
 		inline bool operator==(const AnyParameter& other) const
@@ -185,6 +199,7 @@ namespace shogun
 		Any m_value;
 		AnyParameterProperties m_properties;
 		std::shared_ptr<params::AutoInit> m_init_function;
+		std::function<bool(Any)> m_constrain_function;
 	};
 } // namespace shogun
 
--- a/src/shogun/base/SGObject.cpp
+++ b/src/shogun/base/SGObject.cpp
@@ -650,18 +650,20 @@ void CSGObject::create_parameter(
 
 void CSGObject::update_parameter(const BaseTag& _tag, const Any& value)
 {
-	auto any_param = self->map[_tag];
-	auto pprop = any_param.get_properties();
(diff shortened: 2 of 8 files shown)


src/shogun/base/composition.h
```diff
@@ -0,0 +1,102 @@
...
+#define __COMPOSITION_H__
+
+#include <tuple>
+
+namespace shogun {
+
+	template <typename T, typename ...Args, std::size_t... Idx>
+	bool apply_helper(T&& val, std::tuple<Args...> funcs, std::index_sequence<Idx...>)
+	{
+		return (std::get<Idx>(funcs)(std::forward<T>(val)) && ...);
```
gf712: this is c++17 but can be easily written in c++14

src/shogun/base/SGObject.cpp
```diff
@@ -733,15 +733,25 @@ void CSGObject::create_parameter(
...
-		    "%s::%s is marked as read-only and cannot be modified", get_name(),
+		    "%s::%s is marked as read-only and cannot be modified!\n", get_name(),
 		    _tag.name().c_str());
 	}
+	if (pprop.has_property(ParameterProperties::CONSTRAIN))
+	{
+		if (any_param.get_constrain_function()(value))
+			self->update(_tag, value);
+		else
+			SG_ERROR("%s::%s cannot be updated because of its constraint!\n", get_name(),
```
gf712: the error message could be obtained from the constrain_functions which would have some virtual function with a message, e.g. positive would have "requires value to be positive"

src/shogun/base/SGObject.h
```diff
@@ -113,6 +113,32 @@ using stringToEnumMapType = std::unordered_map<std::string, std::unordered_map<s
...
+		this->watch_param(                                                     \
+		    name, param,                                                       \
+		    [](auto val) {                                                     \
+			    bool state;                                                    \
+			    auto lambda = [&state](auto casted_val) {                      \
+				    state = make_composer(__VA_ARGS__).run(casted_val);        \
+			    };                                                             \
+			    sg_any_dispatch(val, sg_all_typemap, lambda);                  \
+			    return state;                                                  \
+		    },                                                                 \
```
gf712: this is the best I could come up with without using a stateful lambda (which can't be casted to a function pointer)

gf712: state stores the result of the evaluation of the constraint functions

src/shogun/clustering/KMeansBase.cpp
```diff
@@ -354,17 +355,23 @@ SGMatrix<float64_t> CKMeansBase::kmeanspp()
...
+	dimensions = 0;
+	fixed_centers = false;
+	use_kmeanspp = false;
+	SG_ADD(
+	    &max_iter, "max_iter", "Maximum number of iterations",
+	    ParameterProperties::HYPER);
+	SG_ADD_CONSTRAINT(
+	    &k, "k", "k, the number of clusters",
+	    ParameterProperties::HYPER | ParameterProperties::CONSTRAIN,
+	    positive<>());
```
gf712: the `<>` can only be dropped in c++17. the template is by default double (it can't be inferred here)

src/shogun/base/AnyParameter.h
```diff
@@ -220,6 +229,11 @@ namespace shogun
 			return m_init_function;
 		}
 
+		std::function<bool(Any)> get_constrain_function() const
```
gf712: should this return `std::function<bool(Any)>&`? I am not sure if this is making a copy of the function each time?

gf712: quick check on compiler explorer and it seems like in O1 returning by reference produces fewer instructions!

src/shogun/base/SGObject.cpp
```diff
@@ -733,25 +733,33 @@ void CSGObject::create_parameter(
...
 		SG_ERROR(
-		    "%s::%s is marked as read-only and cannot be modified", get_name(),
-		    _tag.name().c_str());
+		    "%s::%s is marked as read-only and cannot be modified!\n",
+		    get_name(), _tag.name().c_str());
+	if (pprop.has_property(ParameterProperties::CONSTRAIN))
+	{
+		if (!any_param.get_constrain_function()(value))
+			SG_ERROR(
+			    "%s::%s cannot be updated because of its constraint!\n",
```
karlnapf: should print a to-string of the constraint here, as this is user facing

src/shogun/clustering/KMeansBase.cpp
```diff
@@ -354,17 +355,23 @@ SGMatrix<float64_t> CKMeansBase::kmeanspp()
...
-	SG_ADD(&k, "k", "k, the number of clusters", ParameterProperties::HYPER);
+	max_iter = 300;
+	k = 8;
+	dimensions = 0;
+	fixed_centers = false;
+	use_kmeanspp = false;
+	SG_ADD(
+	    &max_iter, "max_iter", "Maximum number of iterations",
+	    ParameterProperties::HYPER);
+	SG_ADD_CONSTRAINT(
```
karlnapf: One thing I don't like so much is that we need a new macro here.
The options were added as a part of the `SG_ADD` with this nice `SG_OPTION`.
Is it not possible to do a similar thing here?
`SG_ADD(&my_aram, "my_param", "descr", SG_CONSTRAINT(positive<>)`

gf712: well I was thinking about this, and in an ideal world we would drop these macros. Instead there would be a `sg_add` method that has these overloads with variadic template arguments, and it would be type safe and easier to debug

gf712: but yes, I think it should be possible to do that

karlnapf: i agree with you actually. I think we wanted to keep it this way to keep the old framework working until we decide to kill it. But we should change to a method once the old param stuff goes (hopefully workshop? :D)

gf712:
> Much appreciated!
> Two things:
> 
> * constraint error msgs are user facing so they need to be more elaborate

How elaborate? I just added something that says `KMeans::k cannot be updated because it must be: positive!` and then it could do stuff like `positive and less than 10`, depending on the constraint in sg_add

> * need a user-facing way to look at all constraints (as strings?, see above)

I am not sure I understand what you mean. Is this to display the error? Or you mean something you can query from the SGObject?

> * I would prefer to re-use the existing macro rather than adding a new one (see inline comment)

This will be a bit tougher with macros (but still doable). However at some point things will have to be mixed together, i.e. constrain a value and add a initialisation heuristic, and with a macro this will get out of hand. See #4651 how I handle these situations.

src/shogun/clustering/KMeansBase.cpp
lisitsyn: I think constraints are optional so something like:
`add(...).constraint(positive())`
might be good.

gf712: @lisitsyn if/when we move away from the SG_ADD macros?

lisitsyn: yeah seems so

karlnapf: Maybe it is a good time to bury that macro then... we could also call the old param registrations in the new method?

gf712: yup I don't see why not, I started working a bit on it, and it shouldn't be too hard even with all the new parameter properties.

src/shogun/base/SGObject.h
```diff
@@ -98,15 +99,12 @@ using stringToEnumMapType = std::unordered_map<std::string, std::unordered_map<s
 			this->m_gradient_parameters->add(param, name, description);        \
 	}
 
-#define SG_ADD5(param, name, description, param_properties, auto_init)         \
+#define SG_ADD5(param, name, description, param_properties, auto_or_constraint)\
```
gf712: @karlnapf the issue with this is that now auto_init and constraint are mutually exclusive. This is fine for now, but will have to be taken into consideration when switching to a function, instead of macro.

src/shogun/clustering/KMeansBase.cpp
karlnapf: ++++

gf712:
Not sure what the C# error is about. I get the same error on my machine in both the develop branch and this branch, but only sometimes..

src/shogun/clustering/KMeansBase.cpp
```diff
@@ -317,19 +317,25 @@ SGMatrix<float64_t> CKMeansBase::kmeanspp()
...
+	dimensions = 0;
+	fixed_centers = false;
+	use_kmeanspp = false;
+	SG_ADD(
+	    &max_iter, "max_iter", "Maximum number of iterations",
+	    ParameterProperties::HYPER);
+	SG_ADD(
+	    &k, "k", "k, the number of clusters",
+	    ParameterProperties::HYPER | ParameterProperties::CONSTRAIN,
+	    SG_CONSTRAINT(positive<>()));
```
gf712: the `<>` has to stay because gcc cannot do the deduction without it until gcc-9...

gf712:
@karlnapf this one is good to go now

tests/unit/base/SGObject_unittest.cc
```diff
@@ -616,3 +616,13 @@ TEST(SGObject, unsubscribe_observer_failure)
 
 	EXPECT_THROW(obj->unsubscribe(param_obs_not_in), ShogunException);
 }
+
+TEST(SGObject, constrained_parameter)
```
karlnapf: exactly! I think this is good!

PR #4650, opened 2019-05-28
mapped labels

karlnapf:
Enables mapping user-provided labels into the required internal algorithm representation.

This roughly resembles what we discussed @lisitsyn @gf712 I created a drop-in replacement for `CBinaryLabels`, which replaces the provided labels inside `CMachine::train`. These perform shared-memory (if possible) on-the-fly mappings to the required internal label space when the usual access functions of `CDenseLabels` are used. The mapping code is self-contained and the drop-in functionality is achieved via adding a mixing to the desired label classes.

The user gets back prediction labels in the same space/class as she provided in `train`.

I am looking for high level feedback for now, will tune this stuff later on.

### DIFF
--- a/src/shogun/classifier/Perceptron.cpp
+++ b/src/shogun/classifier/Perceptron.cpp
@@ -60,7 +60,7 @@ void CPerceptron::iteration()
 	bool converged = true;
 	SGVector<float64_t> w = get_w();
 
-	auto labels = binary_labels(m_labels)->get_int_labels();
+	auto labels = m_labels->as<CBinaryLabels>()->get_int_labels();
 	auto iter_train_labels = labels.begin();
 
 	for (const auto& v : DotIterator(features))
--- a/src/shogun/labels/MappedBinaryLabels.cpp
+++ b/src/shogun/labels/MappedBinaryLabels.cpp
@@ -1,0 +1,43 @@
+/*
+ * This software is distributed under BSD 3-clause license (see LICENSE file).
+ *
+ * Authors: Heiko Strathmann
+ */
+
+#include <shogun/labels/MappedBinaryLabels.h>
+#include <shogun/labels/MulticlassLabels.h>
+
+
+using namespace shogun;
+
+std::pair<LabelMap, LabelMap>  CMappedBinaryLabels::create_mapping(const CLabels* orig) const
+{
+	LabelMap to_internal;
+	LabelMap from_internal;
+
+	switch (orig->get_label_type())
+	{
+		case LT_BINARY:
+			break;
+		case LT_MULTICLASS:
+		case LT_REGRESSION:
+		{
+			auto dense = static_cast<const CDenseLabels*>(orig);
+			auto unique = dense->get_labels().unique();
+			REQUIRE(unique.size()<=2, "Cannot use %d label values as binary labels.\n", unique.size());
+			to_internal[unique[0]] = -1;
+			from_internal[-1] = unique[0];
+
+			if (unique.size()==2)
+			{
+				to_internal[unique[1]] = 1;
+				from_internal[+1] = unique[1];
+			}
+			break;
+		}
+		default:
+			SG_ERROR("Cannot use %s as %s.", orig->get_name(), orig->get_name());
+	}
+
+	return std::make_pair(to_internal, from_internal);
+}
--- a/src/shogun/labels/MappedBinaryLabels.h
+++ b/src/shogun/labels/MappedBinaryLabels.h
@@ -1,0 +1,28 @@
+/*
+ * This software is distributed under BSD 3-clause license (see LICENSE file).
+ *
+ * Authors: Heiko Strathmann
+ */
+
+#ifndef MAPPED_BINARY_LABELS_H
+#define MAPPED_BINARY_LABELS_H
+
+#include <shogun/labels/MappedLabels.h>
+#include <shogun/labels/BinaryLabels.h>
+
+namespace shogun
+{
+
+class CMappedBinaryLabels : public MappedLabels<CBinaryLabels>
+{
+public:
+	CMappedBinaryLabels() {} // for class_list.h
+	CMappedBinaryLabels(CLabels* l) : MappedLabels(l) {}
+	virtual const char* get_name() const { return "MappedBinaryLabels"; }
+protected:
+	virtual std::pair<LabelMap, LabelMap> create_mapping(const CLabels* orig) const;
+};
+
+} // namespace shogun
+
+#endif // MAPPED_BINARY_LABELS_H
--- a/src/shogun/labels/MappedLabels.h
+++ b/src/shogun/labels/MappedLabels.h
@@ -1,0 +1,150 @@
+/*
+ * This software is distributed under BSD 3-clause license (see LICENSE file).
+ *
+ * Authors: Heiko Strathmann
+ */
+
+#ifndef MAPPED_LABELS_H
+#define MAPPED_LABELS_H
+
+#ifndef SWIG
+#include <shogun/lib/config.h>
+
+#include <shogun/lib/common.h>
+
+#include <shogun/labels/DenseLabels.h>
+#include <shogun/base/class_list.h>
+#include <shogun/base/range.h>
+
+
+namespace shogun
+{
+
+typedef std::map<float64_t, float64_t> LabelMap;
+
+template <typename T,
+	      std::enable_if_t<
+	          std::is_base_of<
+	              CDenseLabels, typename std::remove_pointer<T>::type>::value,
+	          T>* = nullptr>
+class MappedLabels : public T
+{
+public:
+	MappedLabels() : T() { init(); }
+	MappedLabels(CLabels* orig) : T()
+	{
+		init();
+
+		auto dense = orig->as<CDenseLabels>();
+		ASSERT(dense);
+		m_orig = dense;
+		SG_REF(m_orig);
+	}
+
+	virtual ~MappedLabels()
+	{
+		SG_UNREF(m_orig);
+	}
+
+	virtual SGVector<float64_t> get_labels() const
+	{
+		if (m_to_internal.empty())
+			return m_orig->get_labels();
+
+		auto num_labels = m_orig->get_num_labels();
+		SGVector<float64_t> converted(num_labels);
+		for (auto i : range(num_labels))
+			converted[i] = m_to_internal.at(m_orig->get_label(i));
+
+		return converted;
+	}
+
+	virtual float64_t get_label(index_t i) const
+	{
+		auto orig = m_orig->get_label(i);
+		return m_to_internal.size() ? m_to_internal.at(orig) : orig;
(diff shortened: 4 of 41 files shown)


src/shogun/labels/MappedBinaryLabels.cpp
```diff
@@ -0,0 +1,50 @@
...
+ * Authors: Heiko Strathmann
+ */
+
+#include <shogun/labels/MappedBinaryLabels.h>
+#include <shogun/labels/MulticlassLabels.h>
+
+
+using namespace shogun;
+
+std::tuple<LabelMap, LabelMap> CMappedBinaryLabels::create_mapping(const CLabels* orig) const
```
gf712: std::pair? not sure if it makes any performance difference, but then you can use .first and .second instead of std::get<>()

src/shogun/labels/MappedBinaryLabels.cpp
```diff
@@ -0,0 +1,50 @@
...
+	LabelMap to_internal;
+	LabelMap from_internal;
+
+	switch (orig->get_label_type())
+	{
+		case LT_BINARY:
+			break;
+		case LT_MULTICLASS:
+		{
+			auto mc = (const CMulticlassLabels*) orig; // TODO make const version of .as<>
```
gf712: what happens when you use .as<> with const? Does it remove qualifiers?

tests/unit/labels/MappedLabels_unittest.cc
```diff
@@ -0,0 +1,117 @@
...
+	SGMatrix<float64_t> X;
+	CDenseFeatures<float64_t>* feats;
+	CDenseLabels* labs;
+	CMulticlassLabels* labels_mc;
+	CLibSVM* svm;
+	CGaussianKernel* kernel;
+	CLabels* result;
+};
+
+TEST_F(MappedLabelsFixture, mc_as_bin)
```
karlnapf: see here for how it works for the user providing multiclass labels with only two (arbitrary!) values as binary labels

src/shogun/labels/MappedBinaryLabels.cpp
karlnapf: ah yes thx, I had another thing in there previously.

src/shogun/labels/MappedBinaryLabels.cpp
karlnapf: doesnt work, would need to be added. but also I guess the static cast does the job here in the switch?

src/shogun/labels/MappedBinaryLabels.h
```diff
@@ -0,0 +1,23 @@
...
+#include <shogun/labels/MappedLabels.h>
+#include <shogun/labels/BinaryLabels.h>
+
+namespace shogun
+{
+
+class CMappedBinaryLabels : public MappedLabels<CBinaryLabels>
+{
+public:
+	CMappedBinaryLabels() {} // for class_list.h
```
gf712: can it be `=default`?

src/shogun/labels/MappedBinaryLabels.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+	SG_SDEBUG("to binary label mapping:\n");
+	for (auto it : to_internal)
+		SG_SDEBUG("\t%f->%f\n", it.first, it.second)
+
+	SG_SDEBUG("from binary label mapping:\n");
+	for (auto it : from_internal)
+		SG_SDEBUG("\t%f->%f\n", it.first, it.second)
+
+	return std::pair<LabelMap, LabelMap>(to_internal, from_internal);
```
gf712: you could use std::make_pair instead btw

src/shogun/labels/MappedLabels.h
```diff
@@ -0,0 +1,127 @@
...
+		result->set_labels(inverted);
+		return result;
+	}
+
+	// trait to only define for mixin instantiations
+	template<class MAPPED_CLASS, std::enable_if_t<std::is_base_of<T, MAPPED_CLASS>::value, MAPPED_CLASS>* = nullptr>
+	static T* wrap_if_necessary(CLabels* orig)
+	{
+		auto casted = dynamic_cast<T*>(orig);
+		if (casted)
```
gf712: you can move the line above into the `if` so it stays in its own scope, i.e. `if (auto* casted = dynamic_cast<T*>(orig)) return casted`

src/shogun/labels/MappedLabels.h
```diff
@@ -0,0 +1,127 @@
...
+	virtual ~MappedLabels()
+	{
+		SG_UNREF(m_orig);
+	}
+
+	// access now needs to be virtual in CDenseLabels
+	virtual SGVector<float64_t> get_labels() const
+	{
+		auto orig = m_orig->get_labels();
+		if (!m_to_internal.size())
```
gf712: you should use .empty(), it's more explicit imo

src/shogun/machine/Machine.cpp
```diff
@@ -53,6 +55,26 @@ bool CMachine::train(CFeatures* data)
 		if (m_labels == NULL)
 			SG_ERROR("%s@%p: No labels given", get_name(), this)
 
+		switch (get_machine_problem_type())
+		{
+			case PT_BINARY:
+			{
+				auto old = m_labels;
+				m_labels = MappedLabels<CBinaryLabels>::wrap_if_necessary<CMappedBinaryLabels>(m_labels);
```
karlnapf: this and the apply below are a bit copy paste... will fix it

src/shogun/labels/MappedLabels.h
```diff
@@ -0,0 +1,140 @@
...
+		}
+
+		SG_SDEBUG("Mapping %s as %s.\n", orig->get_name(), demangled_type<MAPPED_CLASS>().c_str());
+		MAPPED_CLASS* result = new MAPPED_CLASS(orig);
+		SG_REF(result);
+
+		// cast is safe since this is a mixin
+		auto mapping = (static_cast<MappedLabels<T>*>(result))->create_mapping(orig);
+
+		result->m_to_internal = mapping.first;
```
karlnapf: @gf712 maybe you know some trick here how to avoid copying the maps. I'd like to avoid passing them as references to `create_mapping` as I want that method `const`. But somehow move-construct the elements of the `std::pair` is what I am looking for

karlnapf: Note: Just changing the return type of `create_mapping` to `std::pair<LabelMap&&, LabelMap&&>` gives a stack overflow in the unit tests :scream

karlnapf:
Now mem-error and leak free, and less verbose/copy-pasted code

src/shogun/machine/Machine.cpp
```diff
@@ -53,6 +55,20 @@ bool CMachine::train(CFeatures* data)
 		if (m_labels == NULL)
 			SG_ERROR("%s@%p: No labels given", get_name(), this)
 
+		CLabels* potentially_mapped = m_labels;
```
karlnapf: changes to CMachine are minimal

src/shogun/machine/Machine.cpp
```diff
@@ -201,12 +213,14 @@ CLabels* CMachine::apply(CFeatures* data)
 	{
 		case PT_BINARY:
 			result=apply_binary(data);
+			result=invert_labels_if_possible<CMappedBinaryLabels>(m_labels, result);
```
karlnapf: changes to CMachine are minimal

tests/unit/labels/MappedLabels_unittest.cc
```diff
@@ -0,0 +1,211 @@
...
+{
+	test_mapped_labels<CMulticlassLabels, CMappedBinaryLabels, CBinaryLabels>({0,2,0,2}, {-1,1,-1,1}, PT_BINARY);
+}
+
+TEST_F(MappedLabelsFixture, bin_as_bin)
+{
+	test_mapped_labels<CBinaryLabels>({-1,1,-1,1}, PT_BINARY);
```
karlnapf: this ensures that no new instances are created when labels do not need to be mapped (so far this was the case in almost all machines where we asserted the necessary label type)

tests/unit/labels/MappedLabels_unittest.cc
```diff
@@ -0,0 +1,211 @@
...
+}
+
+TEST_F(MappedLabelsFixture, mc_as_mc)
+{
+	test_mapped_labels<CMulticlassLabels>({0,1,2,3}, PT_MULTICLASS);
+}
+
+TEST_F(MappedLabelsFixture, non_contiguous_mc_as_mc)
+{
+	test_mapped_labels<CMulticlassLabels, CMappedMulticlassLabels, CMulticlassLabels>({0,2,4,6}, {0,1,2,3}, PT_MULTICLASS);
```
karlnapf: we can now use arbitrary integers as multiclass labels, they are mapped to contiguous ones internally

src/shogun/labels/MappedLabels.h
```diff
@@ -0,0 +1,151 @@
...
+		SG_REF(m_orig);
+	}
+
+	virtual ~MappedLabels()
+	{
+		SG_UNREF(m_orig);
+	}
+
+	// access now needs to be virtual in CDenseLabels
+	virtual SGVector<float64_t> get_labels() const
```
karlnapf: cannot avoid copies when accessing label memory in vectorized form

src/shogun/labels/MappedLabels.h
```diff
@@ -0,0 +1,151 @@
...
+		auto lab = m_orig->get_labels(); // TODO use iterator on label directly
+		std::transform(lab.begin(), lab.end(), converted.begin(),
+				[this](float64_t l) {
+					return this->m_to_internal.at(l);
+				});
+		return converted;
+	}
+
+	// access now needs to be virtual in CDenseLabels
+	virtual float64_t get_label(index_t i) const
```
karlnapf: this maps labels on the fly

src/shogun/classifier/Perceptron.cpp
```diff
@@ -60,7 +60,7 @@ void CPerceptron::iteration()
 	bool converged = true;
 	SGVector<float64_t> w = get_w();
 
-	auto labels = binary_labels(m_labels)->get_int_labels();
+	auto labels = m_labels->as<CBinaryLabels>()->get_int_labels();
```
karlnapf: reverting the old mapping approach which was invasive. Now algorithms simply assume they are given the right labels (as the case before any mapping was done)

lisitsyn: looks good!

karlnapf:
Cool Ill polish it / make it work with CI / update the old mess, and then get it in

gf712:
@karlnapf does this only require a rebase?

karlnapf:
I think we can close this in favour of @LiuYuHui s new label encoder (should be added to Machine::train)

PR #4651, opened 2019-05-28
[WIP] first draft of sg_add method

gf712:
Here is a simple first draft of how the SG_ADD macros could be replaced with methods. I try to make the variadic argument pack somewhat type safe by adding some checks. This could probably be done inside a macro, but this way we can remove all the macro "overloads" of SG_ADD which are annoying to deal with and borderline portable imo because the preprocessor behaviour is not specified in the standard...

The variadic pack is required because a parameter could require both a constraint function #4647 and an automatic initialiser:
```cpp
sg_add(&k, "k", "k, the number of clusters", 
ParameterProperties::HYPER | ParameterProperties::AUTO | ParameterProperties::CONSTRAIN,
std::make_shared<my_kmeans_heuristic>(this), positive<>);
``` 

This also brings into question whether ParameterProperties::AUTO and ParameterProperties::CONSTRAIN have to be explicitly passed to the method

### DIFF
--- a/src/shogun/base/SGObject.h
+++ b/src/shogun/base/SGObject.h
@@ -790,7 +790,7 @@ protected:
 	}
 
 	template <ParameterProperties PPVal, typename ...Args, std::enable_if_t<static_cast<bool>(PPVal&ParameterProperties::AUTO)>* = nullptr>
-	void sg_add_auto_func(AnyParameter& any_pprop, Args&&... args)
+	void declare_auto_func(AnyParameter& any_pprop, Args&&... args)
 	{
 		static_assert(sg_is_any_base_of<params::AutoInit, Args...>::value,
 				"Expected a params::AutoInit function when passing param with"
@@ -800,11 +800,12 @@ protected:
 	}
 
 	template <ParameterProperties pprop, typename T, typename ...Args>
-	void sg_add(T* value, const std::string& name, const std::string& description, Args&& ...args)
+	void declare(T* value, const std::string& name, const std::string& description, Args&& ...args)
 	{
 		auto any_pprop = AnyParameterProperties(description, pprop);
 		auto anyp = AnyParameter(make_any_ref(value), any_pprop);
-		sg_add_auto_func<pprop>(anyp, args...);
+		if constexpr (static_cast<bool>(pprop&ParameterProperties::AUTO))
+            declare_auto_func<pprop>(anyp, args...);
 
 		BaseTag tag(name);
 		create_parameter(tag, anyp);
--- a/src/shogun/base/SGObject.h
+++ b/src/shogun/base/SGObject.h
@@ -784,31 +784,27 @@ public:
 	}
 
 protected:
-	void sg_add_auto_func(AnyParameter& any_pprop, std::shared_ptr<params::AutoInit> func)
+	template <ParameterProperties PPVal, typename ...Args, std::enable_if_t<!static_cast<bool>(PPVal&ParameterProperties::AUTO)>* = nullptr>
+	void sg_add_auto_func(AnyParameter& any_pprop, Args&&... args)
 	{
-		any_pprop.set_init_function(std::move(func));
 	}
 
-	template <typename ...Args>
+	template <ParameterProperties PPVal, typename ...Args, std::enable_if_t<static_cast<bool>(PPVal&ParameterProperties::AUTO)>* = nullptr>
 	void sg_add_auto_func(AnyParameter& any_pprop, Args&&... args)
 	{
 		static_assert(sg_is_any_base_of<params::AutoInit, Args...>::value,
 				"Expected a params::AutoInit function when passing param with"
 				" ParameterProperty::AUTO!");
 		constexpr size_t idx = get_idx_from_pack<params::AutoInit, Args...>::idx;
-		any_pprop.set_init_function(std::get<idx>(std::forward_as_tuple(args...)));
+		any_pprop.set_init_function(std::move(std::get<idx>(std::forward_as_tuple(args...))));
 	}
 
-	template <typename T, typename ...Args>
-	void sg_add(T* value, const std::string& name, const std::string& description,
-			ParameterProperties pprop, Args&& ...args)
+	template <ParameterProperties pprop, typename T, typename ...Args>
+	void sg_add(T* value, const std::string& name, const std::string& description, Args&& ...args)
 	{
 		auto any_pprop = AnyParameterProperties(description, pprop);
 		auto anyp = AnyParameter(make_any_ref(value), any_pprop);
-		if (any_pprop.has_property(ParameterProperties::AUTO))
-		{
-			sg_add_auto_func(anyp, args...);
-		}
+		sg_add_auto_func<pprop>(anyp, args...);
 
 		BaseTag tag(name);
 		create_parameter(tag, anyp);
--- a/src/shogun/base/SGObject.h
+++ b/src/shogun/base/SGObject.h
@@ -16,6 +16,7 @@
 #include <shogun/base/base_types.h>
 #include <shogun/base/constraint.h>
 #include <shogun/base/macros.h>
+#include <shogun/base/sg_type_traits.h>
 #include <shogun/base/unique.h>
 #include <shogun/io/SGIO.h>
 #include <shogun/lib/DataType.h>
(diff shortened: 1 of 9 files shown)


src/shogun/kernel/SigmoidKernel.cpp
```diff
@@ -57,9 +57,13 @@ void CSigmoidKernel::init()
 	gamma = 0.0;
 	coef0 = 0.0;
 
-	SG_ADD(
-	    &gamma, "gamma", "Scaler for the dot product.",
-	    ParameterProperties::HYPER | ParameterProperties::AUTO,
-	    std::make_shared<params::GammaFeatureNumberInit>(this));
+	sg_add(&gamma, "gamma", "Scaler for the dot product.",
+		   ParameterProperties::HYPER | ParameterProperties::AUTO,
+		   std::make_shared<params::GammaFeatureNumberInit>(this));
```
karlnapf: would it be possible to move the `make_shared` into the method rather than having to write it everytime a parameter is registered?

gf712: I am not sure how that would work though. We could pass the type `GammaFeatureNumberInit`, i.e. `sg_add<params::GammaFeatureNumberInit>(&gamma, "gamma", "Scaler for the dot product.", ParameterProperties::HYPER | ParameterProperties::AUTO);`, but I am not sure that is nicer?

lisitsyn: Wouldn't that require too many classes?

gf712: in what sense? you mean for template instantiations?

lisitsyn: Will we have `params::GammaFeatureNumberInit`, `params::AlphaFeatureNumberInit`, etc?

gf712: yes, but the templated method can check if the passed type is derived from the abstract class (I think its `AbstractInit`). not sure if that's what you mean?

src/shogun/kernel/SigmoidKernel.cpp
```diff
@@ -57,9 +57,8 @@ void CSigmoidKernel::init()
 	gamma = 0.0;
 	coef0 = 0.0;
 
-	SG_ADD(
+	sg_add<ParameterProperties::HYPER | ParameterProperties::AUTO>(
```
gf712: what about moving the flags to the template deduction so that we can use a static_assert like in the macro?

karlnapf: +

gf712: is this a typo or is it not rendering in my browser?

src/shogun/kernel/SigmoidKernel.cpp
```diff
@@ -57,9 +57,8 @@ void CSigmoidKernel::init()
...
 	    &gamma, "gamma", "Scaler for the dot product.",
-	    ParameterProperties::HYPER | ParameterProperties::AUTO,
 	    std::make_shared<params::GammaFeatureNumberInit>(this));
-	SG_ADD(&coef0, "coef0", "Coefficient 0.", ParameterProperties::HYPER);
+	sg_add<ParameterProperties::HYPER>(&coef0, "coef0", "Coefficient 0.");
```
karlnapf: @lisitsyn demands declare

gf712: Should’ve done a poll :D

gf712: alright renamed it to declare

src/shogun/base/SGObject.h
```diff
@@ -780,6 +781,28 @@ class CSGObject
...
+		any_pprop.set_init_function(std::move(std::get<idx>(std::forward_as_tuple(args...))));
+	}
+
+	template <ParameterProperties pprop, typename T, typename ...Args>
+	void declare(T* value, const std::string& name, const std::string& description, Args&& ...args)
+	{
+		auto any_pprop = AnyParameterProperties(description, pprop);
+		auto anyp = AnyParameter(make_any_ref(value), any_pprop);
+		if constexpr (static_cast<bool>(pprop&ParameterProperties::AUTO))
+            declare_auto_func<pprop>(anyp, args...);
```
karlnapf: just curious.
Is it possible at this point to still register the old parameters?
Because then we could actually merge this pretty soon

gf712: no because shogun::Parameter is an incomplete type at this point :/

karlnapf: Ah yes of course... well that’s fine....

src/shogun/base/SGObject.h
```diff
@@ -795,6 +796,64 @@ class CSGObject
...
+			Args&&... args)
+	{
+		auto any_pprop = AnyParameterProperties(description, pprop);
+		auto anyp = AnyParameter(make_any_ref(value), any_pprop);
+		if constexpr (static_cast<bool>(pprop & ParameterProperties::AUTO))
+			declare_auto_func<pprop>(anyp, std::forward<Args&&>(args)...);
+		if constexpr (static_cast<bool>(
+				pprop & ParameterProperties::CONSTRAIN))
+			declare_constraint_func<pprop, T>(
+					anyp, std::forward<Args&&>(args)...);
```
gf712: so these two calls actually work well together, and now we can add a constraint and auto initialiser in whichever order, to the same parameter. We can keep on adding more callbacks with parameter registration this way if we want in a type safe way

karlnapf: very nice! :)

gf712: btw I have been experimenting with making AnyParameterProperties constexpr, but the default parameter mask that I added in the other PR won't allow it. long story short the default mask is non constant evaluated unless we make it a non type template, which would make SGObject templated.. so that is a no go

gf712: This would have been neat:

```cpp
template <ParameterProperties pprop = ParameterProperties::NONE, typename T, typename ConstexprLambda, typename... Args>
void declare(T* value, const std::string& name, ConstexprLambda&& description, Args&&... args)
{
//	    constexpr auto mask = pprop | m_const_default_mask; <- breaks everything
	constexpr auto any_pprop = AnyParameterProperties(description(), pprop);
	auto anyp = AnyParameter(make_any_ref(value), any_pprop);
	if constexpr (any_pprop.has_property(ParameterProperties::AUTO))
		declare_auto_func<pprop>(anyp, std::forward<Args&&>(args)...);
	if constexpr (any_pprop.has_property(ParameterProperties::CONSTRAIN))
		declare_constraint_func<pprop, T>(
				anyp, std::forward<Args&&>(args)...);
	BaseTag tag(name);
	create_parameter(tag, anyp);
}
```
However only in C++20 will we be able to pass string literals as non type templates, that's why the description argument is actually a lambda that returns the description (lambdas are const evaluated in C++17). It makes the code a bit ugly:
```cpp
#define MAKE_CONSTEXPR_LAMBDA(x) [](){return x;}

declare<ParameterProperties::HYPER>(&coef0, "coef0", MAKE_CONSTEXPR_LAMBDA("Coefficient 0."));
```
Anyway, I guess this is a no go for now...

karlnapf: > btw I have been experimenting with making AnyParameterProperties constexpr, but the default parameter mask that I added in the other PR won't allow it. long story short the default mask is non constant evaluated unless we make it a non type template, which would make SGObject templated.. so that is a no go

I think the default mask is something less important. Maybe we do this differently in favour of having the map itself compile time?

gf712: oh this is unrelated to the map being compile time. I am also working on that now, just getting it to complain when a key already exists. The thing I am talking about her is to make AnyParameterProperties compile time because all the information is there at compile time, but the performance gain of this is minimal..

karlnapf: Is constexpr and end in itself? :D

gf712: well it's just annoying to have to execute code at runtime that is known at compile time :p

karlnapf:
this would be great to have dont you think?

gf712:
Yup, when I hand in my thesis I'll have some free time again to write code and can start by finishing this up :)

PR #4652, opened 2019-05-28
Make GMM model observable.

geektoni:
* Emit log-likelihood, alphas and gaussians when using the `train_em()` method;
* Emit log-likelihood, alphas and gaussians when using the `train_smem()` method;
* Change m_coefficients registered name to "coefficients" and fix meta example;

### DIFF
--- a/examples/meta/src/observers/gaussian_mixture_models.sg
+++ b/examples/meta/src/observers/gaussian_mixture_models.sg
@@ -19,6 +19,7 @@ gmm.subscribe(obs)
 #![train_sample]
 gmm.put("features", features_train)
 gmm.train_em()
+RealVector output = gmm.sample()
 #![train_sample]
 
 #![training_smem]
@@ -26,10 +27,14 @@ obs.clear()
 gmm.train_smem()
 #![training_smem]
 
-#![extract_params]
+#![extract_coeff]
 int num_observations = obs.get_int("num_observations")
-ObservedValue obs_val = obs.get_observation(num_observations-1)
-RealVector coeff =  obs_val.get_real_vector("coefficients")
-real log_likelihood = obs_val.get_real("log_likelihood")
-RealVector coef = gmm.get_real_vector("coefficients")
-#![extract_params]
+ObservedValue obs_val = obs.get_observation(num_observations-2)
+RealVector observed_coeff =  obs_val.get_real_vector("coefficients")
+RealVector coeff = gmm.get_real_vector("coefficients")
+#![extract_coeff]
+
+#![extract_likelihood]
+ObservedValue obs_val_2 = obs.get_observation(num_observations-3)
+real log_likelihood = obs_val_2.get_real("log_likelihood")
+#![extract_likelihood]
--- a/examples/meta/src/observers/gaussian_mixture_models.sg
+++ b/examples/meta/src/observers/gaussian_mixture_models.sg
@@ -23,7 +23,9 @@ RealVector output = gmm.sample()
 #![train_sample]
 
 #![training_smem]
-obs.clear()
+gmm.unsubscribe(obs)
+ParameterObserver obs_smem = parameter_observer("ParameterObserverLogger")
+gmm.subscribe(obs_smem)
 gmm.train_smem()
 #![training_smem]
 
--- a/examples/meta/src/observers/gaussian_mixture_models.sg
+++ b/examples/meta/src/observers/gaussian_mixture_models.sg
@@ -1,0 +1,35 @@
+File f_feats_train = csv_file("../../data/classifier_4class_2d_linear_features_train.dat")
+
+Math:init_random(1)
+
+#![create_features]
+Features features_train = features(f_feats_train)
+#![create_features]
+
+#![create_gmm_instance]
+int num_components = 3
+GMM gmm(num_components)
+#![create_gmm_instance]
+
+#![add_observer]
+ParameterObserver obs = parameter_observer("ParameterObserverLogger")
+gmm.subscribe(obs)
+#![add_observer]
+
+#![train_sample]
+gmm.put("features", features_train)
+gmm.train_em()
+#![train_sample]
+
+#![training_smem]
+obs.clear()
+gmm.train_smem()
+#![training_smem]
+
+#![extract_params]
+int num_observations = obs.get_int("num_observations")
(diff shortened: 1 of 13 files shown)


examples/meta/src/observers/gaussian_mixture_models.sg
```diff
@@ -0,0 +1,35 @@
...
+gmm.subscribe(obs)
+#![add_observer]
+
+#![train_sample]
+gmm.put("features", features_train)
+gmm.train_em()
+#![train_sample]
+
+#![training_smem]
+obs.clear()
```
karlnapf: Why throw away the old obs? Maybe just a new observer/model?

src/shogun/clustering/GMM.cpp
```diff
@@ -335,6 +339,11 @@ float64_t CGMM::train_smem(int32_t max_iter, int32_t max_cand, float64_t min_cov
 		}
 		if (!better_found)
 			break;
+
+		this->observe<float64_t>(iter, "log_likelihood", "Log Likelihood", cur_likelihood);
+		this->observe<SGVector<float64_t>>(iter, "coefficients");
+		this->observe<std::vector<CGaussian*>>(iter, "gaussians", "Gaussians Components",m_components);
```
karlnapf: Isn’t this a registered parameter? I think it’s not that nice that he description is here even though this belongs to the class.
It should be like he coefficients no?

src/shogun/clustering/GMM.cpp
```diff
@@ -335,6 +339,11 @@ float64_t CGMM::train_smem(int32_t max_iter, int32_t max_cand, float64_t min_cov
 		}
 		if (!better_found)
 			break;
+
+		this->observe<float64_t>(iter, "log_likelihood", "Log Likelihood", cur_likelihood);
```
karlnapf: This is ok as it’s not part of the model

src/shogun/clustering/GMM.cpp
```diff
@@ -826,5 +835,5 @@ void CGMM::register_params()
 	//TODO serialization broken
 	//m_parameters->add((SGVector<CSGObject*>*) &m_components, "m_components", "Mixture components");
```
karlnapf: Could you try to enable this? Should work with the new framework at least (not sg add though, but watch param)

examples/meta/src/observers/gaussian_mixture_models.sg
geektoni: It was just to also show that feature. Since it is training the GMM with two different methods, maybe the user is interested just into the final observations and therefore he will get rid of the previous ones :)

src/shogun/clustering/GMM.cpp
geektoni: It seems to work :) however, GMM needs some unit-tests to ensure proper validation. I may try to add some in a new PR.

src/shogun/clustering/GMM.cpp
geektoni: Yes, indeed. It is fixed now.

examples/meta/src/observers/gaussian_mixture_models.sg
```diff
@@ -0,0 +1,40 @@
...
+Features features_train = features(f_feats_train)
+#![create_features]
+
+#![create_gmm_instance]
+int num_components = 3
+GMM gmm(num_components)
+#![create_gmm_instance]
+
+#![add_observer]
+ParameterObserver obs = parameter_observer("ParameterObserverLogger")
```
karlnapf: as said in my other unrelated comment, it would be good/better if this guy used the mechanics as used in `SGObject::to_string()`

examples/meta/src/observers/gaussian_mixture_models.sg
karlnapf: I would always create a new instance to avoid state-changing confusions... we had done this before and it was causing trouble

src/shogun/clustering/GMM.cpp
```diff
@@ -185,6 +185,10 @@ float64_t CGMM::train_em(float64_t min_cov, int32_t max_iter, float64_t min_chan
 		pb.print_progress();
 		max_likelihood(alpha, min_cov);
 
+		this->observe<float64_t>(iter, "log_likelihood", "Log Likelihood", log_likelihood_cur);
+		this->observe<SGVector<float64_t>>(iter, "coefficients", "Mixture Coefficients", alpha);
```
karlnapf: question (related to my "object" question earlier): is this vector copied? Because it will be modified in the next iteration

src/shogun/clustering/GMM.cpp
```diff
@@ -185,6 +185,10 @@ float64_t CGMM::train_em(float64_t min_cov, int32_t max_iter, float64_t min_chan
...
+		this->observe<std::vector<CGaussian*>>(iter, "components");
```
karlnapf: and these would have to be cloned actually (unless they are just printed)

src/shogun/clustering/GMM.cpp
```diff
@@ -823,8 +832,7 @@ SGVector<float64_t> CGMM::cluster(SGVector<float64_t> point)
 
 void CGMM::register_params()
 {
-	//TODO serialization broken
-	//m_parameters->add((SGVector<CSGObject*>*) &m_components, "m_components", "Mixture components");
+	this->watch_param("components", &m_components, AnyParameterProperties("Mixture components"));
```
karlnapf: nice!

examples/meta/src/observers/gaussian_mixture_models.sg
geektoni: Okay! I will add a new observer then :)

src/shogun/clustering/GMM.cpp
geektoni: Yes, it is copied.

src/shogun/clustering/GMM.cpp
geektoni: Same here. They are all copied/cloned :)

examples/meta/src/observers/gaussian_mixture_models.sg
geektoni: I may update it in another PR since I do not want to bloat this one :)

PR #4653, opened 2019-05-28
[WIP] Feature/observable_framework

geektoni:
@karlnapf This PR was opened such to enable better review of the feature branch before merging it inside `develop`.


karlnapf:
I think we should only look at this once the refactoring pr is in the feature branch as this is still the old design that bloats sgobject

karlnapf:
uuuuh yeah :)

PR #4655, opened 2019-05-30
[WIP] Make Random Forest observable.

geektoni:


examples/meta/src/multiclass/random_forest.sg
```diff
@@ -19,7 +19,7 @@ MajorityVote m_vote()
 
 #![create_instance]
 RandomForest rand_forest(features_train, labels_train, 100)
-rand_forest.set_combination_rule(m_vote)
+rand_forest.put("combination_rule", as_combination_rule(m_vote))
```
karlnapf: could you use a factory above rather than this casting?

src/shogun/base/base_types.h
```diff
@@ -54,7 +55,8 @@ namespace shogun
 	                    std::is_same<CMeanFunction, T>::value ||
 	                    std::is_same<CLossFunction, T>::value ||
 	                    std::is_same<CTokenizer, T>::value ||
-	                    std::is_same<CEvaluationResult, T>::value>
+	                    std::is_same<CEvaluationResult, T>::value ||
+						std::is_same<CCombinationRule, T>::value>
```
karlnapf: some whitespace

examples/meta/src/multiclass/random_forest.sg
karlnapf: see `factory.h`

src/shogun/machine/BaggingMachine.cpp
```diff
@@ -226,12 +230,17 @@ bool CBaggingMachine::train_machine(CFeatures* data)
...
+			// get out of bag indexes
+			CDynamicArray<index_t>* oob = get_oob_indices(idx);
+			m_oob_indices->push_back(oob);
 
-		// add trained machine to bag array
-		m_bags->push_back(c);
+			// add trained machine to bag array
+			m_bags->push_back(c);
+
+			// observe some variables
```
karlnapf: superflous comment! :)

src/shogun/machine/BaggingMachine.cpp
```diff
@@ -226,12 +230,17 @@ bool CBaggingMachine::train_machine(CFeatures* data)
...
+			auto oob_error = this->get_oob_error();
+			this->observe<CMachine*>(i, "trained_machine", "Trained machine for this bag", c);
```
karlnapf: this reminds me of an unrelated question I had, two actually:

* How are the machines (or general: objects) stored in the observations? Because if just the pointer is stored and the objects are modified further downstream of the algorithm, then this breaks stuff. Don't they need to be cloned?
* For the printing (the logger observer), could we pls use the visitor pattern of Any that is used in `CSGObject::to_string()`? See that method for how to do that

src/shogun/machine/BaggingMachine.cpp
```diff
@@ -381,29 +345,45 @@ float64_t CBaggingMachine::get_oob_error(CEvaluation* eval) const
...
 			break;
 
 		default:
 			SG_ERROR("Unsupported label type\n");
 	}
+	SG_REF(evaluation);
 	SG_REF(predicted);
 
+	if (eval != NULL)
+	{
```
karlnapf: this with the above is a bit weird

src/shogun/machine/BaggingMachine.cpp
```diff
@@ -381,29 +345,45 @@ float64_t CBaggingMachine::get_oob_error(CEvaluation* eval) const
...
+		SG_UNREF(evaluation)
+		evaluation = eval;
+	}
+
 	m_labels->add_subset(SGVector<index_t>(idx.data(), idx.size(), false));
-	float64_t res = eval->evaluate(predicted, m_labels);
+	float64_t res = evaluation->evaluate(predicted, m_labels);
```
karlnapf: could you explain this whole introduction of a new variable here?

src/shogun/machine/BaggingMachine.h
```diff
@@ -110,10 +51,13 @@ namespace shogun
 			/** get out-of-bag error
 			 * CombinationRule is used for combining the predictions.
 			 *
-			 * @param eval Evaluation method to use for calculating the error
+			 * @param eval Evaluation method to use for calculating the error. It can be
+			 * null. In such a case, default evaluation criteria will be used:
+			 * - Binary/Multiclass labels: Binary/Multiclass Accuracy;
```
karlnapf: I see! I am not a big fan of this. Obscures the code, what about new cases, etc....
I'd rather have a factory or convenience method that instantiates the evaluation with the choices you made here, but not have that in the class itself

tests/unit/labels/MockLabels.h
```diff
@@ -12,6 +12,14 @@ namespace shogun {
 			MOCK_METHOD0(get_values, SGVector<float64_t>());
 
 			virtual const char* get_name() const { return "MockCLabels"; }
+
+		protected:
+			virtual CSGObject *create_empty() const
```
karlnapf: i dont understand why you need this as there is never an empty instance created ... or am I wrong?

tests/unit/multiclass/BaggingMachine_unittest.cc
```diff
@@ -91,10 +91,10 @@ TEST_F(BaggingMachine, mock_train)
...
 
 	bm->parallel->set_num_threads(1);
-	bm->set_machine(&mm);
-	bm->set_bag_size(bag_size);
-	bm->set_num_bags(num_bags);
-	bm->set_combination_rule(mv);
+	bm->put<CMachine>("machine", bm);
+	bm->put<int32_t>("bag_size", bag_size);
+	bm->put<int32_t>("num_bags", num_bags);
+	bm->put<CCombinationRule>("combination_rule", dynamic_cast<CCombinationRule*>(mv));
```
karlnapf: is the cast necessary?

src/shogun/machine/BaggingMachine.cpp
geektoni: > this reminds me of an unrelated question I had, two actually:
> 
> * How are the machines (or general: objects) stored in the observations? Because if just the pointer is stored and the objects are modified further downstream of the algorithm, then this breaks stuff. Don't they need to be cloned?

Objects are copied by the `observe()` and then the copies are sent to the observers.

tests/unit/labels/MockLabels.h
geektoni: There were situations in which we were emitting `MockLabels` from a machine and therefore I needed to implement the `create_empty()` method to make `clone()` work.

examples/meta/src/multiclass/random_forest.sg
geektoni: Updated.

src/shogun/machine/BaggingMachine.cpp
```diff
@@ -381,29 +346,45 @@ float64_t CBaggingMachine::get_oob_error(CEvaluation* eval) const
...
-			break;
+		    evaluation = new CAccuracyMeasure();
+		    break;
 
-		case LT_MULTICLASS:
-			predicted = new CMulticlassLabels(lab);
-			break;
+	    case LT_MULTICLASS:
+		    predicted = new CMulticlassLabels(lab);
+		    evaluation = new CMulticlassAccuracy();
```
karlnapf: still think this eval thing shouldnt be done automatically

gf712:
@geektoni is this still needed?

geektoni:
> @geektoni is this still needed?

I'm picking it up again these days. Let's see after I fix all these merge conflicts.

src/shogun/machine/BaggingMachine.cpp
```diff
@@ -203,12 +208,26 @@ bool BaggingMachine::train_machine(std::shared_ptr<Features> data)
...
+
+			// observe some variables. The oob error is computed only when when
+			// we have all the variables set and there are observers attached.
+			this->observe<std::shared_ptr<Machine>>(
+			    i, "trained_machine", "Trained machine for this bag", c);
+
+			if (this->m_combination_rule && this->m_oob_evaluation_metric &&
+			    this->get_num_subscriptions() != 0)
+			{
+				auto oob_error = this->get_oob_error();
```
karlnapf, line 227: question: you now always compute the oob error whenever there is an observer ...Is that good? I forgot how the specifying to be observed variables works, but can't the user somehow pick what she wants to observe? In that case this should only be computed if the user requested it

geektoni: So, the idea is that as long as the model has some observer attached, then it will output all the available observations. This means that the OOB error will be computed regardless of the user preferences.

It will be the observer job to filter the observations and keep just the ones that the user wants.

geektoni: At the moment, there is no way to tell an observable model to emit only certain observations.

karlnapf: I don’t think that’s good and we should probably think about a way to avoid this. Passing (and ignoring) pointers to existing data is one thing, but expensive operations should only be executed if desired.
I could imagine that this could be done with a lambda function that is only executed if the observer actually wants (as in user instructed) to store the data. Maybe @gf712 has an idea he is the lambda master?

geektoni: I see the problem. One small issue here is that the model does not have any idea about which things the observers want to observe. I guess the solution could be emitting functions (lambdas) instead which can then be executed by the observers if they want to store that kind of data. Is this what you meant @karlnapf?

karlnapf: yes, the model doesnt know. In general, it should *offer* the observer to observe something (which the observer then can or cannot do, depending on its settings). Passing a pointer is a good example for that. The observer can clone the underlying structure for example.
But for costly things, passing a lambda that would compute something expensive is a better choice imo, so yet.

geektoni: Okay, that makes sense! I will do some experimenting to try adding this lambda feature. I'm leaving this one as it is (or maybe we could merge it in a `feature` branch) and then I will open a new PR with the lambda observers.

karlnapf: I say let's just keep this PR open, and then you could rebase it once the the lambda thing works. Alternatively, just remove the oob error stuff and we can merge everything else?

PR #4656, opened 2019-05-30
Fix MKL macOS bug

theartful:
The problem is that mkl tries to run in parallel mode without proper openmp available. I found that these options solve the problem:
1. Install openmp and link to it (which is what I did here)
2. Tell mkl programatically to run in sequential mode if openmp is not available (not the best solution, since it wasn't designed to solve linking problems. `mkl_rt` should solve those).
3. Don't use the dynamic signle library `mkl_rt` and link to each single library needed (`mkl_core`, etc.), and use the sequential library `mkl_sequential`.

### DIFF
--- a/.ci/ci.yml
+++ b/.ci/ci.yml
@@ -133,6 +133,11 @@ jobs:
     ccacheArtifactName: 'ccache-macos'
     ccacheArchive: '$(Build.ArtifactStagingDirectory)/ccache-$(Build.BuildId).tar'
     commonMacOSCMakeFlags: '-DBUILD_EXAMPLES=OFF -DBUNDLE_JSON=OFF -DBUNDLE_NLOPT=OFF -DENABLE_TESTING=ON -DENABLE_COVERAGE=OFF -DBUILD_META_EXAMPLES=OFF'
+    openMPMacOSCMakeFlags:
+      '-DOpenMP_CXX_FLAGS="-Xpreprocessor -fopenmp"
+      -DOpenMP_C_FLAGS="-Xpreprocessor -fopenmp"
+      -DOpenMP_CXX_LIB_NAMES="omp" -DOpenMP_C_LIB_NAMES="omp"
+      -DOpenMP_omp_LIBRARY="/usr/local/opt/libomp/lib/libomp.dylib"'
 
   pool:
     vmImage: xcode9-macos10.13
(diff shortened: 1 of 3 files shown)


geektoni:
Impressive! :)

src/shogun/CMakeLists.txt
```diff
@@ -200,6 +200,8 @@ if (OPENMP_FOUND)
   SHOGUN_COMPILE_OPTS(${OpenMP_CXX_FLAGS})
   if (CMAKE_COMPILER_IS_GNUCC)
     SHOGUN_LINK_LIBS(gomp)
+  else ()
+    SHOGUN_LINK_LIBS(OpenMP::OpenMP_CXX)
```
gf712: btw what happens if OpenML hasn't been found?

theartful: you only reach this line if it has been found. this extends it to compilers other than gcc
https://github.com/shogun-toolbox/shogun/blob/f7255cf2cc6b5116e50840816d70d21e7cc039bb/src/shogun/CMakeLists.txt#L197-L206

gf712: ah yes, sorry! missed that!

geektoni:
If no one has objections I guess this could be merged such to bring back to like MacOS MKL tests :)

gf712:
yup, go for it!

.ci/ci.yml
```diff
@@ -133,6 +133,11 @@ jobs:
     ccacheArtifactName: 'ccache-macos'
     ccacheArchive: '$(Build.ArtifactStagingDirectory)/ccache-$(Build.BuildId).tar'
     commonMacOSCMakeFlags: '-DBUILD_EXAMPLES=OFF -DBUNDLE_JSON=OFF -DBUNDLE_NLOPT=OFF -DENABLE_TESTING=ON -DENABLE_COVERAGE=OFF -DBUILD_META_EXAMPLES=OFF'
+    openMPMacOSCMakeFlags:
+      '-DOpenMP_CXX_FLAGS="-Xpreprocessor -fopenmp"
+      -DOpenMP_C_FLAGS="-Xpreprocessor -fopenmp"
+      -DOpenMP_CXX_LIB_NAMES="omp" -DOpenMP_C_LIB_NAMES="omp"
+      -DOpenMP_omp_LIBRARY="/usr/local/opt/libomp/lib/libomp.dylib"'
```
karlnapf: i wonder whether we can detect this rather than using this fixed path?

.ci/macos-steps.yml
```diff
@@ -17,7 +17,7 @@ steps:
 
 - bash: |
     brew update
-    brew install ccache pkg-config arpack eigen glpk hdf5 json-c lapack lzo nlopt snappy xz
+    brew install ccache pkg-config arpack eigen glpk hdf5 json-c lapack lzo nlopt snappy xz libomp
```
karlnapf: even though this is going to /usr/local ...

but I am not sure, just a thought,  I am not a mac user

karlnapf:
kudos for bringing the CI back :)

.ci/ci.yml
theartful: We could use `$(brew --prefix libomp)/lib/libomp.dylib`.

theartful: Another option would be to use llvm clang instead of xcode clang. New versions of llvm clang come with openmp built in, so I believe in that case we wouldn't need the openmp flags at all, and cmake would detect it.

karlnapf: imo that would be much better

theartful: the brew stuff or clang stuff?
+ sorry, my mistake, I actually used `$(brew --prefix libomp)` when testing.

PR #4658, opened 2019-06-04
Use SGVector instead of raw pointers in dense_dot

theartful:
This is part of getting rid of raw pointers in Features. I renamed `dense_dot` to `dot`, and now it uses `SGVector`. Currently `dense_dot` just calls `dot` so as not to break current algorithms. If this is ok, I will change the rest of the code to use this instead of the one with raw pointers.

### DIFF
--- a/src/shogun/classifier/svm/LibLinear.cpp
+++ b/src/shogun/classifier/svm/LibLinear.cpp
@@ -344,7 +344,7 @@ void CLibLinear::solve_l2r_l1l2_svc(
 			i = index[s];
 			int32_t yi = y[i];
 
-			G = prob->x->dense_dot(i, w.vector, n);
+			G = prob->x->dot(i, w.slice(0, n));
 			if (prob->use_bias)
 				G += w.vector[n];
 
@@ -1272,7 +1272,7 @@ void CLibLinear::solve_l2r_lr_dual(
 			double C = upper_bound[GETI(i)];
 			double ywTx = 0, xisq = xTx[i];
 
-			ywTx = prob->x->dense_dot(i, w.vector, w_size);
+			ywTx = prob->x->dot(i, w.slice(0, w_size));
 			if (prob->use_bias)
 				ywTx += w.vector[w_size];
 
--- a/src/shogun/features/BinnedDotFeatures.cpp
+++ b/src/shogun/features/BinnedDotFeatures.cpp
@@ -115,7 +115,8 @@ float64_t CBinnedDotFeatures::dot(int32_t vec_idx1, CDotFeatures* df, int32_t ve
 
 }
 
-float64_t CBinnedDotFeatures::dot(int32_t vec_idx1, const SGVector<float64_t> vec2) const
+float64_t
+CBinnedDotFeatures::dot(int32_t vec_idx1, const SGVector<float64_t> vec2) const
 {
 	assert_shape(vec2.vlen);
 
--- a/src/shogun/features/BinnedDotFeatures.cpp
+++ b/src/shogun/features/BinnedDotFeatures.cpp
@@ -115,9 +115,9 @@ float64_t CBinnedDotFeatures::dot(int32_t vec_idx1, CDotFeatures* df, int32_t ve
 
 }
 
-float64_t CBinnedDotFeatures::dense_dot(int32_t vec_idx1, const float64_t* vec2, int32_t vec2_len) const
+float64_t CBinnedDotFeatures::dot(int32_t vec_idx1, const SGVector<float64_t> vec2) const
 {
-	assert_shape(vec2_len);
+	assert_shape(vec2.vlen);
 
 	float64_t result=0;
 	double sum=0;
--- a/src/shogun/features/BinnedDotFeatures.cpp
+++ b/src/shogun/features/BinnedDotFeatures.cpp
@@ -118,7 +118,7 @@ float64_t CBinnedDotFeatures::dot(int32_t vec_idx1, CDotFeatures* df, int32_t ve
 float64_t
 CBinnedDotFeatures::dot(int32_t vec_idx1, const SGVector<float64_t> vec2) const
 {
-	assert_shape(vec2.vlen);
+	assert_shape(vec2.size());
 
 	float64_t result=0;
 	double sum=0;
--- a/src/shogun/features/BinnedDotFeatures.cpp
+++ b/src/shogun/features/BinnedDotFeatures.cpp
@@ -116,7 +116,7 @@ float64_t CBinnedDotFeatures::dot(int32_t vec_idx1, CDotFeatures* df, int32_t ve
 }
 
 float64_t
-CBinnedDotFeatures::dot(int32_t vec_idx1, const SGVector<float64_t> vec2) const
+CBinnedDotFeatures::dot(int32_t vec_idx1, const SGVector<float64_t>& vec2) const
 {
 	assert_shape(vec2.size());
 
--- a/src/shogun/features/CombinedDotFeatures.cpp
+++ b/src/shogun/features/CombinedDotFeatures.cpp
@@ -115,7 +115,7 @@ float64_t CCombinedDotFeatures::dot(int32_t vec_idx1, CDotFeatures* df, int32_t 
 }
 
 float64_t CCombinedDotFeatures::dot(
-    int32_t vec_idx1, const SGVector<float64_t> vec2) const
+    int32_t vec_idx1, const SGVector<float64_t>& vec2) const
 {
 	float64_t result=0;
 
--- a/src/shogun/features/CombinedDotFeatures.cpp
+++ b/src/shogun/features/CombinedDotFeatures.cpp
@@ -125,8 +125,8 @@ float64_t CCombinedDotFeatures::dot(
 	{
 		CDotFeatures* f = get_feature_obj(f_idx);
 		int32_t dim = f->get_dim_feature_space();
-		result += f->dense_dot(vec_idx1, &vec2[offs], dim) *
-		          f->get_combined_feature_weight();
+		result += f->dot(vec_idx1, vec2.slice(offs, offs+dim)) *
+		          get_subfeature_weight(f_idx);
 		offs += dim;
 
 		SG_UNREF(f);
--- a/src/shogun/features/CombinedDotFeatures.cpp
+++ b/src/shogun/features/CombinedDotFeatures.cpp
@@ -114,7 +114,8 @@ float64_t CCombinedDotFeatures::dot(int32_t vec_idx1, CDotFeatures* df, int32_t 
 	return result;
 }
 
-float64_t CCombinedDotFeatures::dot(int32_t vec_idx1, const SGVector<float64_t> vec2) const
+float64_t CCombinedDotFeatures::dot(
+    int32_t vec_idx1, const SGVector<float64_t> vec2) const
 {
 	float64_t result=0;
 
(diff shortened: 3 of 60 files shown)


src/shogun/features/LBPPyrDotFeatures.cpp
```diff
@@ -166,10 +166,13 @@ uint32_t* CLBPPyrDotFeatures::get_image(int32_t index, int32_t& width, int32_t&
 	return img;
 }
 
-float64_t CLBPPyrDotFeatures::dense_dot(int32_t vec_idx1, const float64_t* vec2, int32_t vec2_len) const
+float64_t
+CLBPPyrDotFeatures::dot(int32_t vec_idx1, const SGVector<float64_t> vec2) const
 {
-	if (vec2_len != vec_nDim)
-		SG_ERROR("Dimensions don't match, vec2_dim=%d, vec_nDim=%d\n", vec2_len, vec_nDim)
+	if (vec2.vlen != vec_nDim)
```
gf712: you could replace this with a `REQUIRE(vec2.vlen == vec_nDim, ...)`

src/shogun/features/PolyFeatures.cpp
```diff
@@ -136,10 +136,13 @@ float64_t CPolyFeatures::dot(int32_t vec_idx1, CDotFeatures* df, int32_t vec_idx
 	return sum;
 }
 
-float64_t CPolyFeatures::dense_dot(int32_t vec_idx1, const float64_t* vec2, int32_t vec2_len) const
+float64_t
+CPolyFeatures::dot(int32_t vec_idx1, const SGVector<float64_t> vec2) const
 {
-	if (vec2_len != m_output_dimensions)
-		SG_ERROR("Dimensions don't match, vec2_dim=%d, m_output_dimensions=%d\n", vec2_len, m_output_dimensions)
+	if (vec2.vlen != m_output_dimensions)
```
gf712: same here

src/shogun/features/SparseFeatures.cpp
```diff
@@ -457,13 +457,15 @@ template<> float64_t CSparseFeatures<complex128_t>::dot(int32_t vec_idx1,
...
+template <class ST>
+float64_t
+CSparseFeatures<ST>::dot(int32_t vec_idx1, const SGVector<float64_t> vec2) const
 {
-	REQUIRE(vec2, "dense_dot(vec_idx1=%d,vec2_len=%d): vec2 must not be NULL\n",
-		vec_idx1, vec2_len);
-	REQUIRE(vec2_len>=get_num_features(),
-		"dense_dot(vec_idx1=%d,vec2_len=%d): vec2_len should contain number of features %d %d\n",
-		vec_idx1, vec2_len, get_num_features());
+	REQUIRE(
```
gf712: basically have this instead (this isn't related to the PR but would be good to clean this up)

theartful: done

gf712: cool! thanks!

src/shogun/features/SparseFeatures.h
```diff
@@ -48,6 +48,8 @@ template <class T> class CCache;
  */
 template <class ST> class CSparseFeatures : public CDotFeatures
 {
+	using CDotFeatures::dense_dot;
```
theartful: SparseFeatures has its own `dense_dot` function (with different signature), which masks over `CDotFeatures::dense_dot`, so this is to enable both of them.

src/shogun/features/BinnedDotFeatures.cpp
```diff
@@ -115,9 +115,10 @@ float64_t CBinnedDotFeatures::dot(int32_t vec_idx1, CDotFeatures* df, int32_t ve
 
 }
 
-float64_t CBinnedDotFeatures::dense_dot(int32_t vec_idx1, const float64_t* vec2, int32_t vec2_len) const
+float64_t
+CBinnedDotFeatures::dot(int32_t vec_idx1, const SGVector<float64_t> vec2) const
 {
-	assert_shape(vec2_len);
+	assert_shape(vec2.vlen);
```
karlnapf: maybe use `.size()` as that makes the code more uniform

src/shogun/features/LBPPyrDotFeatures.cpp
```diff
@@ -166,10 +166,13 @@ uint32_t* CLBPPyrDotFeatures::get_image(int32_t index, int32_t& width, int32_t&
 	return img;
 }
 
-float64_t CLBPPyrDotFeatures::dense_dot(int32_t vec_idx1, const float64_t* vec2, int32_t vec2_len) const
+float64_t
+CLBPPyrDotFeatures::dot(int32_t vec_idx1, const SGVector<float64_t> vec2) const
 {
-	if (vec2_len != vec_nDim)
-		SG_ERROR("Dimensions don't match, vec2_dim=%d, vec_nDim=%d\n", vec2_len, vec_nDim)
+	REQUIRE(
```
karlnapf: could we maybe pass this type of check/warning on to `linalg`?

src/shogun/mathematics/linalg/LinalgBackendBase.h
```diff
@@ -54,47 +55,6 @@ namespace shogun
...
-#define DEFINE_FOR_REAL_PTYPE(METHODNAME, Container)                           \
-	METHODNAME(float32_t, Container);                                          \
-	METHODNAME(float64_t, Container);                                          \
-	METHODNAME(floatmax_t, Container);
-
-#define DEFINE_FOR_NON_INTEGER_PTYPE(METHODNAME, Container)                    \
-	METHODNAME(float32_t, Container);                                          \
-	METHODNAME(float64_t, Container);                                          \
-	METHODNAME(floatmax_t, Container);                                         \
-	METHODNAME(complex128_t, Container);
```
theartful: These were already define in LinalgMacro

src/shogun/mathematics/linalg/LinalgEnums.h
```diff
@@ -32,11 +32,34 @@
...
+
+			using complex_type =
+				std::conditional_t<std::is_same<T, complex128_t>::value, T, U>;
+			using floating_type =
+				std::conditional_t<std::is_floating_point<T>::value, T, U>;
+			using bigger_type = std::conditional_t<(sizeof(T) > sizeof(U)), T, U>;
+
+			using type = std::conditional_t < is_one_complex, complex_type,
+				std::conditional_t<is_one_floating, floating_type, bigger_type>>;
+		};
```
theartful: probably not the best place to define this. can't be defined in LinalgNamespace, so i defined it here

src/shogun/mathematics/linalg/LinalgNamespace.h
```diff
@@ -845,8 +846,8 @@ namespace shogun
 		 * represented
 		 * as \f$\sum_i a_i b_i\f$
 		 */
-		template <typename T>
-		T dot(const SGVector<T>& a, const SGVector<T>& b)
+		template <typename T, typename U, typename TU = typename promote<T, U>::type>
+		TU dot(const SGVector<T>& a, const SGVector<U>& b)
```
theartful: a more general purpose dot product interface. but the backend only implements all combinations with T=U and with T=float64_t, U != T

src/shogun/features/LBPPyrDotFeatures.cpp
theartful: yea, it's possible at some places

theartful:
This is getting ugly, I will split stuff to other PRs

src/shogun/classifier/svm/LibLinear.cpp
```diff
@@ -345,7 +345,7 @@ void CLibLinear::solve_l2r_l1l2_svc(
 			i = index[s];
 			int32_t yi = y[i];
 
-			G = prob->x->dense_dot(i, w.vector, n);
+			G = prob->x->dot(i, w);
```
karlnapf: So much nicer! :)

src/shogun/features/CombinedDotFeatures.cpp
```diff
@@ -114,7 +114,8 @@ float64_t CCombinedDotFeatures::dot(int32_t vec_idx1, CDotFeatures* df, int32_t
 	return result;
 }
 
-float64_t CCombinedDotFeatures::dense_dot(int32_t vec_idx1, const float64_t* vec2, int32_t vec2_len) const
+float64_t CCombinedDotFeatures::dot(
+    int32_t vec_idx1, const SGVector<float64_t> vec2) const
```
karlnapf: The second argument should be a const ref rather than passing by value imo

src/shogun/features/DenseFeatures.cpp
```diff
@@ -631,63 +631,13 @@ GET_FEATURE_TYPE(F_DREAL, float64_t)
...
-	free_feature_vector(vec1, vec_idx1, vfree);
-
-	return result;
-}
-
-template<> float64_t CDenseFeatures<float64_t>::dense_dot(
-		int32_t vec_idx1, const float64_t* vec2, int32_t vec2_len) const
+template <typename ST>
+float64_t
+CDenseFeatures<ST>::dot(int32_t vec_idx1, const SGVector<float64_t> vec2) const
```
karlnapf: In all those places actually

src/shogun/features/SparseFeatures.cpp
```diff
@@ -474,9 +476,11 @@ template<class ST> float64_t CSparseFeatures<ST>::dense_dot(int32_t vec_idx1, co
 			"sparse_matrix[%d] check failed (matrix features %d >= vector dimension %d)\n",
 			vec_idx1, get_num_features(), sv.get_num_dimensions());
 
-		REQUIRE(vec2_len >= sv.get_num_dimensions(),
-			"sparse_matrix[%d] check failed (dense vector dimension %d >= vector dimension %d)\n",
-			vec_idx1, vec2_len, sv.get_num_dimensions());
+		REQUIRE(
```
karlnapf: Putting these check so low level is costly. Why not check once outside the loop instead? Not sure.....

src/shogun/features/hashed/HashedDocDotFeatures.cpp
```diff
@@ -96,14 +96,10 @@ float64_t CHashedDocDotFeatures::dot(int32_t vec_idx1, CDotFeatures* df, int32_t
...
+float64_t CHashedDocDotFeatures::dot(
+	int32_t vec_idx1, const SGVector<float64_t> vec2) const
 {
-	return dense_dot(vec_idx1, vec2.vector, vec2.vlen);
-}
-
-float64_t CHashedDocDotFeatures::dense_dot(int32_t vec_idx1, const float64_t* vec2, int32_t vec2_len) const
-{
-	ASSERT(vec2_len == CMath::pow(2,num_bits))
+	ASSERT(vec2.size() == CMath::pow(2,num_bits))
```
karlnapf: Pls use std::pow

src/shogun/features/SparseFeatures.cpp
theartful: I agree that they're more costly, but I think putting checks in slightly lower level functions is cleaner and less error prone (like Linalg does).

src/shogun/features/CombinedDotFeatures.cpp
theartful: @karlnapf Doing stuff like `dot(index, vector.slice(a, b))` wouldn't work if we used an lvalue reference.

theartful: I'm wrong, nvm :")

theartful:
All uses of this variant of dense_dot is replaced, except for some cases which will be uglier in the gpl code.

karlnapf:
Let's rebase (fix the submodule conflict) and then merge ....

karlnapf:
One good extension of this topic would be (not in this PR though).

Quite often, we do iterative dot products in loops over all feature vectors. This could be done via a matrix multiplication for dense features. And that would be faster. So I think it would be cool to identify such opportunities and implement and API for that case. In a first step, the API would just call the existing one-by-one dot products, a second step would be to check whether a matrix-multiple batched version is possible (e.g. dense feats, no subsets), and then check whether we get speed improvements

theartful:
this is ready

PR #4659, opened 2019-06-04
Move Any in AnyParameter ctor

gf712:
We always call the `AnyParameter` ctor with a temporary `Any`, so we might as well just move it in the ctor instead of copying.

### DIFF
--- a/src/shogun/base/AnyParameter.h
+++ b/src/shogun/base/AnyParameter.h
@@ -11,8 +11,8 @@
 #include <shogun/lib/any.h>
 #include <shogun/lib/bitmask_operators.h>
 
-#include <memory>
 #include <list>
+#include <memory>
 
 namespace shogun
 {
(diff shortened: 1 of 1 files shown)


src/shogun/base/AnyParameter.h
```diff
@@ -175,17 +177,18 @@ namespace shogun
 		AnyParameter() : m_value(), m_properties()
 		{
 		}
-		explicit AnyParameter(const Any& value) : m_value(value), m_properties()
+		explicit AnyParameter(Any&& value)
```
gf712: Changes are here, the rest is just clang formatting

src/shogun/base/AnyParameter.h
```diff
@@ -43,13 +43,15 @@ namespace shogun
...
-		{ParameterProperties::READONLY, "READONLY"}};
+	static const std::list<std::pair<ParameterProperties, std::string>>
+	    kParameterPropNames = {
+	        {ParameterProperties::NONE, "NONE"},
+	        {ParameterProperties::HYPER, "HYPER"},
+	        {ParameterProperties::GRADIENT, "GRADIENT"},
+	        {ParameterProperties::MODEL, "MODEL"},
+	        {ParameterProperties::AUTO, "AUTO"},
+	        {ParameterProperties::READONLY, "READONLY"},
+	        {ParameterProperties::RUNFUNCTION, "RUNFUNCTION"}};
```
gf712: also added the latest enum

PR #4661, opened 2019-06-05
Refactor the ParameterObserverLogger to use ToStringVisitor.

geektoni:


src/shogun/io/visitors/ToStringVisitor.h
```diff
@@ -0,0 +1,100 @@
...
+				for (auto i : range(v->vlen)) {
+					stream() << (*v)[i];
+					if (i < v->vlen - 1)
+						stream() << ",";
+				}
+				stream() << "]";
+			}
+		}
+
+	private:
```
karlnapf: so this is just moving the code right?

src/shogun/lib/observers/ParameterObserverLogger.cpp
```diff
@@ -38,37 +39,18 @@ void CParameterObserverLogger::on_next_impl(const TimedObservedValue& value)
...
-		    "[%lu] Received a value called \"%s\" which contains: %s\n",
-		    convert_to_millis(value.second), name.c_str(),
-		    value.first->to_string().c_str())
+		any_val.visit(&visitor);
+	} else {
+		stream << "{function}";
 	}
+
+	SG_PRINT(
+			"[%lu] Received a value called \"%s\" which contains: %s\n",
```
karlnapf: What about just `"\"%s\"=%s\n", name.c_str(), stream.str().c_str()`

src/shogun/io/visitors/ToStringVisitor.h
geektoni: yes, indeed.

src/shogun/lib/any.h
```diff
@@ -318,8 +340,10 @@ namespace shogun {
 			enter_matrix(std::addressof(rows), std::addressof(cols));
 			if ((rows != _matrix->num_rows) || (cols != _matrix->num_cols))
 				*_matrix = SGMatrix<T>(rows, cols);
-			for (auto& _value: *_matrix)
-				on(std::addressof(_value));
+			for (auto index=0; index < cols; index++) {
+				on_matrix_row(std::addressof(rows), std::addressof(index), _matrix);
```
geektoni: I've changed a bit how the visitor behaves when dealing with SGMatrix (this way is much more customizable). @karlnapf @vigsterkr

PR #4662, opened 2019-06-05
Implement SGVector slicing

theartful:
This will be useful for replacing raw pointers. A slice can be reference counted, and we keep track of an offset, which is needed in deallocation.

Why this is useful?
When we want to do an operation with only a part of a vector. For example
```C++
float64_t CCombinedDotFeatures::dense_dot(int32_t vec_idx1, const float64_t* vec2, int32_t vec2_len) const
{
	float64_t result=0;

	uint32_t offs=0;

	for (index_t f_idx=0; f_idx<get_num_feature_obj(); f_idx++)
	{
		CDotFeatures* f = get_feature_obj(f_idx);
		int32_t dim = f->get_dim_feature_space();
		result += f->dense_dot(vec_idx1, vec2+offs, dim)*f->get_combined_feature_weight();
		offs += dim;

		SG_UNREF(f);
	}

	return result;
}
```
Where we do a dot product with `vec2 + offs`. Without slicing, we would have to create a new SGVector that points to `vec2 + offs` then call `dot` which accept SGVector.

### DIFF
--- a/src/shogun/lib/SGVector.cpp
+++ b/src/shogun/lib/SGVector.cpp
@@ -309,6 +309,15 @@ void SGVector<T>::resize_vector(int32_t n)
 	vlen=n;
 }
 
+template <class T>
+SGVector<T> SGVector<T>::slice(index_t l, index_t h) const
+{
+	assert_on_cpu();
+	ASSERT(l >= 0 && h <= vlen);
+
+	return SGVector<T>(vector, h - l, l);
+}
+
 /** addition operator */
 template<class T>
 SGVector<T> SGVector<T>::operator+ (SGVector<T> x)
--- a/src/shogun/lib/SGVector.h
+++ b/src/shogun/lib/SGVector.h
@@ -270,6 +270,13 @@ template<class T> class SGVector : public SGReferencedData
 		 */
 		void resize_vector(int32_t n);
 
+		/** Returns a view of the vector from l inclusive to h exclusive
+		 *
+		 * @param l slice start index (inclusive)
+		 * @param h slice end index (exclusive)
+		 */
+		SGVector<T> slice(index_t l, index_t h) const;
+
 		/** Operator overload for vector read only access
 		 *
 		 * @param index dimension to access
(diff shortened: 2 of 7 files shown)


src/shogun/lib/SGVector.h
```diff
@@ -270,6 +269,37 @@ template<class T> class SGVector : public SGReferencedData
...
+		 * @param h slice end index (exclusive)
+		 */
+		SGVector<T> slice(index_t l, index_t h) const;
+
+		/** Operator overloading for vector slicing in read-only mode
+		 *
+		 * @param l slice start index (inclusive)
+		 * @param h slice end index (exclusive)
+		 */
+		inline const SGVector<T> operator()(index_t l, index_t h) const
```
karlnapf: really nice!

theartful:
@karlnapf It's already implemented in `SGMatrix` with the name `submatrix`, but it's not owning, so no reference counting. @vigsterkr also found that reference counting with views isn't necessary, so I think I should change this to that as well.

karlnapf:
Yes subslices of things usually are not stored. However, things should be unified (no different names/behaviour among SGMatrix and SGVector ....)

theartful:
Now they're same name (slice) and same behaviour (not reference counted). I also removed slice operator overloading in SGVector, since in SGMatrix it is used to access a specific element, so it might cause confusion.

src/shogun/lib/SGVector.cpp
```diff
@@ -309,6 +309,15 @@ void SGVector<T>::resize_vector(int32_t n)
 	vlen=n;
 }
 
+template <class T>
+SGVector<T> SGVector<T>::slice(index_t l, index_t h) const
+{
+	assert_on_cpu();
+	ASSERT(l >= 0 && h <= vlen);
+
+	return SGVector<T>(vector, h - l, l);
```
gf712: cool! much simpler this way

theartful:
@karlnapf I think this is ready

src/shogun/lib/SGVector.cpp
karlnapf: so you do reference counting for the returned vector?

theartful: no reference counting

PR #4665, opened 2019-06-06
[WIP] Separate kernel caching from the interface

theartful:
This is done by creating another class `CachedKernel` that is composed of the original kernel, in addition to the caching data/methods.


src/shogun/kernel/CachedKernel.h
```diff
@@ -0,0 +1,241 @@
...
+
+#include <shogun/io/SGIO.h>
+#include <shogun/kernel/Kernel.h>
+#include <shogun/lib/common.h>
+#include <shogun/lib/config.h>
+
+namespace shogun
+{
+
+#ifdef USE_SVMLIGHT
```
karlnapf, line 11: does it use SVMLight?

src/shogun/kernel/CachedKernel.h
```diff
@@ -0,0 +1,241 @@
...
+			/** start */
+			int32_t start;
+			/** end */
+			int32_t end;
+			/** of vectors */
+			int32_t num_vectors;
+		};
+#endif // DOXYGEN_SHOULD_SKIP_THIS
+
+		static void* cache_multiple_kernel_row_helper(void* p);
```
karlnapf, line 228: I assume all this is basically cut/pasted from somewhere else?

ghost: yes, from Kernel.h

src/shogun/kernel/CachedKernel.h
ghost: It's only used by SVMLight

iglesias:
What's the status here, @theartful?

ghost:
Almost all SVMLight related examples are ported, and I added a unit-test that checks that cache-size doesn't affect results.

tests/unit/classifier/svm/SVMLight_unittest.cc
```diff
@@ -0,0 +1,53 @@
...
+#include <shogun/kernel/GaussianKernel.h>
+#include <shogun/lib/SGVector.h>
+#include <shogun/lib/config.h>
+
+using namespace shogun;
+
+#ifdef USE_SVMLIGHT
+TEST(SVMLight, CacheSizeInvariance)
+{
+	const int32_t num_samples = 50;
```
iglesias, line 15: Minor on style, uniform initialization is preferred. No need to update it here; I find it a convenient rule.

karlnapf:
I think this can be merged soon?

ghost:
While we're at it, one last thing. I don't like this line here:
```
gaussian_kernel->set_cache_size(cache_size);
```
There are multiple algorithms that use this cache size property, but shouldn't it be a parameter of these algorithms?

iglesias:
> While we're at it, one last thing. I don't like this line here:
> 
> ```
> gaussian_kernel->set_cache_size(cache_size);
> ```
> 
> There are multiple algorithms that use this cache size property, but shouldn't it be a parameter of these algorithms?

It could make better sense, indeed. Why do you think it'd better as parameter?

ghost:
@iglesias It's currently a parameter of the Kernel interface, but if we're going stateless, it shouldn't. Also I think this doesn't concern/affect the behaviour of the kernel as it does the algorithms.

gf712:
@theartful I'll keep this alive with a comment and then can see

ghost:
@gf712 Thanks! I will tie up some of these PRs this weekend.

karlnapf:
@theartful would be great to get this in! :)

theartful:
September first I will be free :") Will look into it then along side my other PRs.

karlnapf:
That'd be ace!

PR #4670, opened 2019-06-08
Set both CXX and C flags for compilation

lisitsyn:
This helps to avoid compilation error `unrecognized option -fopenmp=libomp` with clang 6 on bionic.


lisitsyn:
Something is wrong here. It basically disables openmp for some reason

PR #4674, opened 2019-06-08
Thread safe x-validation

karlnapf:


src/shogun/evaluation/CrossValidation.cpp
```diff
@@ -159,186 +120,74 @@ float64_t CCrossValidation::evaluate_one_run(
...
+			m_splitting_strategy->generate_subset_inverse(i);
+
+		SGVector<index_t> idx_test =
+			m_splitting_strategy->generate_subset_indices(i);
+
+		auto features_train = view(m_features, idx_train);
+		auto labels_train = view(m_labels, idx_train);
+		auto features_test = view(m_features, idx_test);
+		auto labels_test = view(m_labels, idx_test);
+		SG_REF(features_train);
```
vinx13, line 117: why not use `Some`?

examples/undocumented/libshogun/classifier_svmlight_string_features_precomputed_kernel.cpp
```diff
@@ -93,7 +93,6 @@ void test_svmlight()
 	CSVM* svm=new CSVMLight(C, kernel, labels_train);
 //	CSVM* svm=new CLibSVM(C, kernel, labels_train);
 	svm->parallel->set_num_threads(1);
-	svm->set_store_model_features(false);
```
karlnapf, line 96: disabled as not needed any more

src/shogun/base/AnyParameter.h
```diff
@@ -17,15 +17,16 @@
 namespace shogun
 {
 	/** parameter properties */
-	enum class ParameterProperties
+	enum class ParameterProperties : uint32_t
```
karlnapf, line 20: needs to be int so i can have a max

src/shogun/base/SGObject.cpp
```diff
@@ -702,7 +702,7 @@ CSGObject* CSGObject::clone() const
 	          "CSGObject::create_empty() overridden.\n",
 	    get_name());
 
-	for (const auto &it : self->map)
+	for (const auto &it : self->filter(pp))
```
karlnapf, line 586: need this in xvalidation to only clone hyperparameters

src/shogun/evaluation/CrossValidation.cpp
```diff
@@ -56,66 +55,17 @@ void CCrossValidation::init()
...
-		storage->post_init();
-		SG_DEBUG("Ending CrossValidationStorage initilization.\n")
-
 		SG_DEBUG("entering cross-validation run %d \n", i)
-		results[i] = evaluate_one_run(i, storage);
+		results[i] = evaluate_one_run(i);
 		SG_DEBUG("result of cross-validation run %d is %f\n", i, results[i])
-
-		/* Emit the value*/
-		observe(
```
karlnapf, line 115: @geektoni I removed this for now as it wasnt thread safe and not const

src/shogun/evaluation/CrossValidation.cpp
```diff
@@ -145,200 +88,55 @@ void CCrossValidation::set_num_runs(int32_t num_runs)
...
-			    m_splitting_strategy->generate_subset_indices(i);
-
-			/* evtl. update xvalidation output class */
-			fold->set_train_indices(inverse_subset_indices);
-			auto fold_machine = (CMachine*)m_machine->clone();
-			fold->set_trained_machine(fold_machine);
-			SG_UNREF(fold_machine)
-
-			/* produce output for desired indices */
-			CLabels* result_labels = m_machine->apply_locked(subset_indices);
```
karlnapf, line 198: I removed the whole locked story....should be done properly or not at all ...this was just copy pasted code

src/shogun/evaluation/CrossValidation.cpp
karlnapf: because of viktors PR :)

src/shogun/evaluation/CrossValidationSplitting.cpp
```diff
@@ -25,6 +25,7 @@ CCrossValidationSplitting::CCrossValidationSplitting(
 
 void CCrossValidationSplitting::build_subsets()
 {
+	REQUIRE(m_labels, "No labels provided.\n");
```
karlnapf, line 27: used to segfault :)

src/shogun/evaluation/GradientEvaluation.cpp
```diff
@@ -69,7 +71,7 @@ CEvaluationResult* CGradientEvaluation::evaluate_impl()
 
 	SG_UNREF(gradient);
 
-	update_parameter_hash();
+//	update_parameter_hash();
```
karlnapf: this kills gradient model selection, but this is not const so bad

karlnapf: @gf712 would be good to get gradients that are behind const methods

src/shogun/features/Features.cpp
```diff
@@ -29,9 +29,9 @@ CFeatures::CFeatures(const CFeatures& orig)
 	init();
 
 	// Call to init creates new preproc arrays.
-	SG_UNREF(preproc);
-	preproc = orig.preproc;
-	SG_REF(preproc);
+	auto old = preproc;
+	preproc = make_clone(preproc);
```
karlnapf, line 33: these need to be cloned for `View` to be thread safe @vinx13

karlnapf: (called from duplicate method)

src/shogun/labels/ClassificationLabels.h
```diff
@@ -0,0 +1,44 @@
...
+
+#ifndef CLASSIFICATION_LABELS_H
+#define CLASSIFICATION_LABELS_H
+
+#include <shogun/lib/config.h>
+
+#include <shogun/labels/Labels.h>
+#include <shogun/lib/common.h>
+
+namespace shogun
```
karlnapf: accidental commit, will remove

src/shogun/base/SGObject.h
```diff
@@ -1059,7 +1056,7 @@ class CSGObject
 	Parameter* m_gradient_parameters;
 
 	/** Hash of parameter values*/
-	uint32_t m_hash;
+	mutable uint32_t m_hash;
```
karlnapf: to allow to make `evaluate` of `CGradientEvaluation` const

src/shogun/base/SGObject.h
```diff
@@ -1081,6 +1078,20 @@ class CSGObject
...
 
+template <class T>
+T* make_clone(T* orig, ParameterProperties pp = ParameterProperties::ALL)
+{
+	REQUIRE(orig, "No object provided.\n");
+	return orig->clone(pp)->template as<T>();
+}
+
+template <class T>
+const T* make_clone(const T* orig, ParameterProperties pp = ParameterProperties::ALL)
```
karlnapf, line 1296: so I can clone and get back the same type as I provided, it is annoying to always have to cast the result

karlnapf: const verison is required from xvalidation

src/shogun/evaluation/CrossValidation.cpp
karlnapf: because soon smart pointer :)

karlnapf:
`Could not find compiler set in environment variable CC:`
https://dev.azure.com/shogunml/shogun/_build/results?buildId=1700&view=logs&jobId=089c709a-44eb-5f6e-96e7-15e9ee1ff5bf&taskId=f13b83c9-aa1a-5ed1-151f-7b757052fcaa&lineStart=100&lineEnd=101&colStart=1&colEnd=1

Any comments on that @vigsterkr @gf712 @lisitsyn ?

gf712:
> `Could not find compiler set in environment variable CC:`
> https://dev.azure.com/shogunml/shogun/_build/results?buildId=1700&view=logs&jobId=089c709a-44eb-5f6e-96e7-15e9ee1ff5bf&taskId=f13b83c9-aa1a-5ed1-151f-7b757052fcaa&lineStart=100&lineEnd=101&colStart=1&colEnd=1

Not sure why that is happening. That means that no c compiler was found on that machine... might be an azure bug? trigger CI again and see if it still happens

vigsterkr:
rebase.... i was updating the docker image to use clang-4.0 but haven't updated the env var for the compiler for swig interfaces

vigsterkr:
plz dont push stuff until further notice...

src/shogun/base/SGObject.h
```diff
@@ -1207,6 +1204,20 @@ class CSGObject
 		int64_t m_next_subscription_index;
 	};
 
+template <class T>
+T* make_clone(T* orig, ParameterProperties pp = ParameterProperties::ALL)
+{
+	REQUIRE(orig, "No object provided.\n");
+	return orig->clone(pp)->template as<T>();
```
gf712: I guess you could actually just have a static_cast here no? you know it must be of type T so it shouldn't require a runtime check?

karlnapf: yes

karlnapf:
Ok I am now at a point where the seed setting in the tests doesn't work when running multithreaded code (e.g. LibLinear::train uses CMath::random(), which screws up reproducibility when called from xval), so I will park this until we have a thread-safe way of generating random numbers @vigsterkr @theartful

src/shogun/base/SGObject.h
```diff
@@ -1244,7 +1241,7 @@ class CSGObject
 		Parameter* m_gradient_parameters;
 
 		/** Hash of parameter values*/
-		size_t m_hash;
+		mutable size_t m_hash;
```
karlnapf, line 1261: @vigsterkr as discussed at workshop. still ok with the new hash?

src/shogun/evaluation/CrossValidation.cpp
```diff
@@ -16,31 +16,29 @@
 #include <shogun/lib/observers/ObservedValueTemplated.h>
 #include <shogun/machine/Machine.h>
 #include <shogun/mathematics/Statistics.h>
-#include <shogun/util/converters.h>
+#include <shogun/lib/View.h>
 
 using namespace shogun;
 
-CCrossValidation::CCrossValidation() : CMachineEvaluation()
+CCrossValidation::CCrossValidation() : RandomMixin<CMachineEvaluation>()
```
karlnapf, line 23: @theartful @vigsterkr question (also see email): is there a better way to allow to set the seed on the `CrossValidation` object? It doesnt use/generate random numbers itself, but the recursive setting is really helpful here (otherwise user needs to set multiple seeds in the parameters)

karlnapf:
Have some tested thread safety (same results independent of num threads) on a few selected algorithms. @theartful the random mixin does quite nicely here ...

src/shogun/features/Features.cpp
```diff
@@ -28,10 +28,15 @@ CFeatures::CFeatures(const CFeatures& orig)
...
-	// Call to init creates new preproc arrays.
-	SG_UNREF(preproc);
-	preproc = orig.preproc;
-	SG_REF(preproc);
+	// TODO this should be a shallow copy
+	auto old_preproc = preproc;
+	preproc = make_clone(preproc);
+	SG_UNREF(old_preproc);
+
+	// TODO this should be a shallow copy
```
karlnapf: @lisitsyn this is what I need the shallow clone method for (just recursively create new instances, but always share the data (vectors/matrices, features, etc)

karlnapf:
Ok this finally passes CI.
Please have a look @vigsterkr @gf712

tests/unit/base/CrossValidation_unittest.cc
```diff
@@ -0,0 +1,204 @@
...
+	{
+		init();
+		this->cv->put("seed", 1);
+		get_global_parallel()->set_num_threads(4);
+		auto result = cv->evaluate()->get<float64_t>("mean");
+		clean();
+		return result;
+	}
+
+	void generate_data(EProblemType pt)
```
karlnapf: I was using the test environments at first but most algorithms get accuracy of 1.0 or similar in there, so it is not that meaningful to compare the xvalidation results with and without threads.... so now some custom problems. Once this is merged, I will try to write a test framework that tests all machines (or sub-sets of those) on particular environments. But for now I think this is good

src/shogun/base/SGObject.h
```diff
@@ -1286,6 +1283,22 @@ class CSGObject
...
+	auto clone = orig->clone(pp);
+	ASSERT(clone);
+	return static_cast<T*>(clone);
+}
+
+template <class T>
+const T* make_clone(const T* orig, ParameterProperties pp = ParameterProperties::ALL)
+{
+	REQUIRE(orig, "No object provided.\n");
+	return orig->clone(pp)->template as<T>();
```
gf712: Shouldn't you also do a static_cast here?

karlnapf: yes! updated.

karlnapf:
@theartful the liblinear error was not coming from random :)

PR #4675, opened 2019-06-09
Remove Maybe to be replaced with std::optional

lisitsyn:


lisitsyn:
We don't use it anywhere so just removing for now.

PR #4678, opened 2019-06-09
Filter observed values by ParameterProperties.

geektoni:


karlnapf:
hard to see the diff now with all these commits

geektoni:
It should be much more readable now ;)

src/shogun/lib/observers/ParameterObserver.cpp
```diff
@@ -50,15 +50,30 @@ ParameterObserver::ParameterObserver()
...
+									 std::vector<ParameterProperties>& properties)
     : ParameterObserver()
 {
 	m_observed_parameters = parameters;
+	m_observed_properties = properties;
+}
+
+ParameterObserver::ParameterObserver(std::vector<std::string>& parameters)
+{
+
```
karlnapf: what is this?

src/shogun/lib/observers/ParameterObserver.cpp
```diff
@@ -72,10 +87,24 @@ bool ParameterObserver::filter(const std::string& param)
...
-			return true;
-	return false;
+	auto res = std::find(m_observed_parameters.begin(),
+						 m_observed_parameters.end(),
+						 param);
+
+	return res != m_observed_parameters.end();
+}
+
+bool ParameterObserver::filter(const AnyParameterProperties &property) {
```
karlnapf: filter is not a good name, that would imply it returns a (sub)set of parameters, if it returns a bool "has" would be more appropriate or?

src/shogun/lib/observers/ParameterObserverCV.h
```diff
@@ -51,6 +51,22 @@ namespace shogun
 
 	public:
 		CParameterObserverCV(bool verbose = false);
+
+        CParameterObserverCV(std::vector<std::string> &parameters, bool verbose = false);
```
karlnapf: maybe remove the "verbose" as we have the logger now

src/shogun/lib/observers/ParameterObserverScalar.h
```diff
@@ -52,10 +52,21 @@ namespace shogun
 
 	public:
 		CParameterObserverScalar();
-		CParameterObserverScalar(std::vector<std::string>& parameters);
-		CParameterObserverScalar(
-		    const std::string& filename, std::vector<std::string>& parameters);
-		~CParameterObserverScalar();
+
+        CParameterObserverScalar(std::vector<std::string> &parameters);
```
karlnapf: do we even still need this class now?

tests/unit/lib/observers/ParameterObserverTest_unittest.cc
```diff
@@ -0,0 +1,141 @@
+/*
+* BSD 3-Clause License
+*
+* Copyright (c) 2017, Shogun-Toolbox e.V. <shogun-team@shogun-toolbox.org>
+* All rights reserved.
+*
+* Redistribution and use in source and binary forms, with or without
+* modification, are permitted provided that the following conditions are met:
```
karlnapf: pls use the new style short license

tests/unit/lib/observers/ParameterObserverTest_unittest.cc
```diff
@@ -0,0 +1,141 @@
...
+	virtual ~MockEmitter() {}
+
+	virtual const char *get_name() const {
+		return "MockEmitter";
+	}
+
+	void emit_value()
+	{
+		AnyParameterProperties p("Name of the observed value", ParameterProperties::READONLY);
+        AnyParameterProperties p2("Name of the observed value", ParameterProperties::AUTO);
```
karlnapf: needs some formating

karlnapf:
Please check this: if we filter and use `ALL`, then I explicitly include parameters with no properties (`NONE`) in `clone`. Might be good to do that here as well: https://github.com/shogun-toolbox/shogun/pull/4674/files#diff-9c3599c0d2090e493be261b079e9b63eR104

karlnapf:
updates here? :)

geektoni:
> updates here? :)

Currently, I'm trying to rebase it and to finish it up! :)

vigsterkr:
@geektoni goodluckhavefun... sorry man but i didn't want to do a 4th rebase of the shared_ptr stuff :P

geektoni:
> @geektoni goodluckhavefun... sorry man but i didn't want to do a 4th rebase of the shared_ptr stuff :P

:dancer: :dancer: :dancer:

src/shogun/lib/observers/ParameterObserverScalar.h
geektoni: This class is still needed atm since it is used to serialize an observation into the Tensorboard format (which can be read by `tensorboard --logdir lala`).

geektoni:
> Please check this: if we filter and use `ALL`, then I explicitly include parameters with no properties (`NONE`) in `clone`. Might be good to do that here as well: https://github.com/shogun-toolbox/shogun/pull/4674/files#diff-9c3599c0d2090e493be261b079e9b63eR104

This behaviour is also respected by the ParameterObservers now.

src/shogun/lib/observers/ParameterObserver.cpp
geektoni: `has` is already used by SGObject and I would like to avoid confusion. I don't mind the `filter` name since basically it is returning a sub-set of observations (the one we care about).

geektoni:
@karlnapf This should be mostly done now.

src/shogun/lib/observers/ParameterObserver.cpp
karlnapf: but it returns a bool, or am I wrong?

karlnapf: Actually why not `has` again? The class is separate or?

geektoni: yes yes, it returns a boolean in the end. `ParameterObserver` inherits from SGOject and thus it has already an `has` method.

karlnapf: I'd suggest `observes` then, as again, `filter` implies that it returns a subset, not a bool

karlnapf:
Cool! What's next with this? Examples :) ?

src/shogun/lib/observers/ParameterObserver.cpp
geektoni: `observes` sounds good.

geektoni:
> Cool! What's next with this? Examples :) ?

Yup, at least one meta example to show how these things can be used in practice.

gf712:
> > Cool! What's next with this? Examples :) ?
> 
> Yup, at least one meta example to show how these things can be used in practice.

Would it be possible to extend a notebook with this? I think there is a tensorboard extension for notebooks.

geektoni:
> > > Cool! What's next with this? Examples :) ?
> > 
> > 
> > Yup, at least one meta example to show how these things can be used in practice.
> 
> Would it be possible to extend a notebook with this? I think there is a tensorboard extension for notebooks.

Yes, it could be possible to extend a notebook with visualizations taken from the observers. As long as the underlying model used is observable of course.

geektoni:
This is ready to be merged once the CI approves.

src/shogun/lib/observers/ParameterObserver.h
```diff
@@ -64,13 +36,30 @@ namespace shogun
 		 */
 		ParameterObserver(std::vector<std::string>& parameters);
 
+		/**
+		 * Constructor
+		 * @param parameters list of parameters which we want to watch over
+		 */
+		ParameterObserver(std::vector<ParameterProperties>& properties);
```
geektoni, line 43: @karlnapf @vigsterkr since these observers will be mainly used by the users, do we still need these constructors or would it be better to populate the member variables by using only `put` calls?

karlnapf: fine from my side, but no hard feelings. I think Viktor might prefer c++ ctors to remain in there (reason the lib might be used natively)

geektoni: Okay, I will keep them in case they need to be used natively.

PR #4679, opened 2019-06-09
linalg header only draft

theartful:
This shows how a header-only linalg might be like. It would be useful since it will provide greater flexibility, since no explicit instantiation is needed. Also it's very hard currently to create a linalg function with two template parameters, since there are a lot of combinations.
My first thought was to do it using CRTP, and get rid of virtual functions:
```C++
template <typename Derived>
class LinalgBackendBase
{
public:
    template <typename T>
    void foo()
    {
        static_cast<const Derived*>(this)->foo();
    }
}
```
But that would make it harder to use different types of backends at the same time. So I used variant instead:
```C++
class LinalgBackendBase
{
public:
    template <typename Derived>
    void set_derived(Derived* derived)
    {
        m_derived = derived;
    }
    template <typename T>
    void foo()
    {
        visit([&](auto&& backend) { backend->foo(); }, derived);
    }
private:
    variant</* list of all backends */> m_derived;
}
```
And the derived backend should register itself to the base class. So it's as if we're mimicking the virtual table using variant, but without using virtual, and with templates. This results in less bloat, and shouldn't affect build time that much, since only needed functions are built.
Also, with this design, we might mix function definitions in .cpp and in .h (to minimize compile time if needed), so it doesn't have to be header-only but still gives the benefit of flexibility.


karlnapf:
A few things that we had in mind when designing the current linagl (and some history)

The old linalg (you never saw that) had the downside that one had to re-compile shogun when changing the backend, which causes a combinatorial explosion of the builds in our CI. This is why we came up with the idea of making linalg a plugin-style things that serves a previously defined API, which can then be exchanged at runtime. Also, people can write their own linalg implementations and just chuck them in.

Going header only kills that feature, i.e. we cannot have drop-in replacements of the backend at runtime (correct me if I am wrong). I think that with the variant version, we would need to compile/define all backends at compile time and then can change among those... which has similarities with what we want, but it is also different. (I haven't looked at the code in detail yet).

Going back and forth between design principles is maybe not the best idea, so before we go into this, there needs to be a discussion of the pros/cons of each approach, taking into account the lessons learned.

One thing is for sure: the current linalg is not very scalable/nice after having added all these methods. There are a lot of problems. So good to get involved here. Let's keep this in planning mode while you are working on the other things, and then come back to it once those are finished

theartful:
@karlnapf  As compared to the previous design, this has the downside that you have to forward declare all backends in `LinalgBackendBase` as you said (I don't think it makes a big difference), but then they can be changed at runtime.
You are definitely right about recompiling shogun if any of the backends changed, but this can be easily fixed, since this is not restricted to header only. Here are possible solutions:
1. Define most of the functions in a cpp file, and specialize them, like the previous design. But leave highly generic functions in the header, which wasn't possible in the previous design, since templates and virtual don't mix well. This will be nice to define a dot product that accepts two different types of vectors for example (which would fix the ugly for loops in features).
2. Instead of one big `LinalgNamespace` divide it into segments `BasicOps`, `Decomposition`, etc. So instead of including all, you just include what you want. And if you want all, include `LinalgNamespace`. I like this solution better, since header only libraries are very flexible.

Would like to hear what you think.

iglesias:
Thanks for the discussion, guys. I like the separation described in point 2, it should help with scalability. It may also be possible to mix both strategies as needed (separation into name spaces and relying on non-header files).

karlnapf:
I think what you describe is a good compromise that makes things better than they are.

Another thing we should do is to get rid of the macros

Open issues remain: 
* old-school immediate computation vs more modern graph building/delayed/lazy evaluation
* autodiff (can be done at runtime or compile time)
* an untemplated api that dispatches inner types at runtime efficiently, to be used inside algos? Or do we dispatch before, see shubhams work
* what should be in CFeatures vs in linalg and how should those apis conceptually differ?

iglesias:
This looks neat (the BODY macro with base/derived tags, std::visit with the variant member for the specific backend). Exciting tricks!

Nitpicking, an artifact is the last argument derived_tag in the implementations. I wonder whether a template and specialization could lead to more elegant syntax.

How much does this approach hurt compilation with respect to the current one?

Can we have a minimal example of changing the backend at runtime, @theartful?
I doubt whether this covers the drop-in replacements. @karlnapf, how would one of those drop-ins look like? Can you point me to some code doing the exchange at runtime?

theartful:
@iglesias 
Compilation is currently affected since everything in this draft is header only. In fact, MSVC compiler couldn't complete compilation because of a "compiler is out of heap space" error. But this can be 
mitigated by mixing between header only and cpp files & dividing into segments. In theory, this should be just as fast as the other implementation if all were migrated to cpp files.

Locally this builds and works fine, and passes all current tests, and it did so also in CI (except for windows), but I didn't test changing backend at runtime yet, I would be happy to write a test for it.

The tag artifact is there to allow default behaviour if the backend doesn't implement a specific function. It can be removed if we force backends to implement all of the interface. There might be a more elegant way to do this though.

iglesias:
Hey @theartful, thanks a lot for the very swift and clear answers.

It all sounds good and reasonable. It is fine if the compilation time is still too large with MSVC. I asked to get an idea of the current state (I wondered even if this could be merge-able). It is even better that we have ideas how to proceed to improve this 👍  

Good also to know that the current tests exercise this new header-only linalg variant (🥁).

Regarding the test changing the back-end at runtime and how to proceed with this PR. Before anything else, let's get clarification on the runtime changing back-ends and drop-in replacements. I think also keeping it in planning/discussing mode for the moment is just fine - I have my own backlog of your PRs to check so my reactions sometimes come a bit late :-/

PR #4681, opened 2019-06-10
Feature iterator

gf712:
* Adds an iterator to iterable feature and label classes. ~~If a feature has a subset it will sort the subset in order to reduce cache misses.~~ No sorting is performed as it slows down performance significantly. 
According to micro benchmarks done on my laptop using an iterator is consistently faster than using a for loop (~10%).
```cpp
for (const auto& iter: *feat)
   do stuff...
```
* adds a zip_iterator, which comes at a performance cost. This cost is minimised as much as possible with move constructors. (The issue is that there is no RVO here since the iterator is a tuple from where the range loop gets a reference from each index, but there is no way around it.)
```cpp
for (auto [label_i, data_i]: zip_iterator(feat, labels))
   do stuff...
```
* added move constructors for SGVector and SGMatrix and started replacing/adding functions that accept rvalue references


src/shogun/base/subset_iterators.h
```diff
@@ -0,0 +1,86 @@
...
+				m_subset = pointer->get_subset_stack()
+				               ->get_last_subset()
+				               ->get_subset_idx();
+			m_idx = idx;
+			// need to keep this variable so that the vector destructor isn't
+			// called
+			m_argsorted_subset = CMath::argsort(m_subset);
+			m_argsorted_subset_iter = m_argsorted_subset.begin();
+			m_ptr = pointer;
+//			SG_REF(m_ptr)
```
gf712: @karlnapf not sure about this referencing because what happens if the feature is on the stack?

src/shogun/base/subset_iterators.h
```diff
@@ -0,0 +1,86 @@
...
+		}
+
+		~SubsetIterator()
+		{
+//		    SG_UNREF(m_ptr)
```
gf712: @karlnapf if feature is on the stack this will cause issues

src/shogun/features/DenseFeatures.h
```diff
@@ -134,6 +137,29 @@ template<class ST> class CDenseFeatures: public CDotFeatures
...
+		return SubsetIterator(this);
+	}
+
+	/**
+	 * Returns an iterator to the element following the end of features.
+	 */
+	iterator end() noexcept
+	{
+		if (auto stack = get_subset_stack()->get_last_subset();
+			stack == nullptr)
```
gf712: c++ 17 in its glory with init statements in if

gf712: i think clang format doesn't know what to do in this case btw

src/shogun/base/subset_iterators.h
gf712: ideally this would use the actual feature matrix no?

src/shogun/base/subset_iterators.h
```diff
@@ -0,0 +1,86 @@
...
+		SubsetIterator(T* pointer, index_t idx = 0)
+		{
+			if (pointer->get_subset_stack()->get_last_subset() != nullptr)
+				m_subset = pointer->get_subset_stack()
+				               ->get_last_subset()
+				               ->get_subset_idx();
+			m_idx = idx;
+			// need to keep this variable so that the vector destructor isn't
+			// called
+			m_argsorted_subset = CMath::argsort(m_subset);
```
ghost: will sorting be called multiple times in one range loop?

gf712: no, this would only be at the start and end when the iterator is instantiated. still need to clean this up more though

src/shogun/base/subset_iterators.h
karlnapf: Objects never are on stack so should be ok

src/shogun/base/subset_iterators.h
```diff
@@ -0,0 +1,86 @@
...
+		}
+
+		const auto operator*()
+		{
+			index_t idx;
+			if (m_ptr->get_subset_stack()->get_last_subset() != nullptr)
+				idx = *m_argsorted_subset_iter;
+			else
+				idx = m_idx;
+			if constexpr (std::is_base_of<CFeatures, T>::value)
```
karlnapf: Yessss

src/shogun/base/subset_iterators.h
```diff
@@ -0,0 +1,86 @@
...
+#include <shogun/labels/Labels.h>
+#include <shogun/features/Features.h>
+
+namespace shogun
+{
+	template <
+	    typename T, std::enable_if<
+	                    (std::is_base_of<T, CFeatures>::value) ||
+	                    (std::is_base_of<T, CLabels>::value)>* = nullptr>
+	class SubsetIterator
```
karlnapf: Find the name a bit Strang but I don’t know a better one :)

src/shogun/features/DenseFeatures.h
```diff
@@ -134,6 +137,29 @@ template<class ST> class CDenseFeatures: public CDotFeatures
...
+			return SubsetIterator(this, get_num_vectors());
+		else
+			return SubsetIterator(
+				this, *std::max_element(
```
karlnapf: Maybe this for can be a mixin as it’ll be the same for labels and other features

src/shogun/base/subset_iterators.h
```diff
@@ -0,0 +1,86 @@
...
+	{
+	public:
+		SubsetIterator(T* pointer, index_t idx = 0)
```
karlnapf: Could the pointer be const?

gf712: I was just thinking that instead we should pass a pointer to the matrix/vector and the subset pointer. that would make the iterator more conventional. I don't like that it holds a reference to the class...

src/shogun/base/subset_iterators.h
gf712: hmm i agree, i just thought it was easier to distinguish from the vector iterator

src/shogun/base/subset_iterators.h
gf712: and that we to pass `this`

src/shogun/base/subset_iterators.h
```diff
@@ -0,0 +1,120 @@
...
+		}
+
+		bool operator==(const SubsetIterator<T>& other)
+		{
+			if (m_ptr->get_subset_stack()->get_last_subset() == nullptr)
+				return m_ptr == other.m_ptr && m_idx == other.m_idx;
+			else
+				return m_ptr == other.m_ptr &&
+                       (*m_argsorted_subset_iter) == *(const_cast<SubsetIterator<T>&>(other).m_argsorted_subset_iter) &&
+				       m_idx == other.m_idx;
```
gf712: @theartful is this what you were talking about on irc?

src/shogun/base/subset_iterators.h
karlnapf: No, some features don’t have matrices but a get feature vector method

gf712: @karlnapf i removed the reference counting because it doesn't make sense for the iterator to have any ownership over the class. This should be a weak reference. If the feature object is deleted the user should not be using the iterator anymore anyway

src/shogun/base/subset_iterators.h
```diff
@@ -0,0 +1,112 @@
...
+		operator++()
+		{
+			m_argsorted_subset_iter++;
+			m_idx++;
+			return *this;
+		}
+
+		bool operator==(const SubsetIterator<T>& other)
+		{
+            return m_ptr == other.m_ptr && m_idx == other.m_idx;
```
ghost: cool!

src/shogun/base/subset_iterators.h
karlnapf: exactly! I cannot imagine a case where this would not be the case.
But the const would also be good for potential speed no?

gf712: well we return by value so i am not sure it needs a const? ideally we would return a const ref..

karlnapf: True It’s a new vector.
Just we want the make the vector access methods const later on and we want to avoid that people write into the vectors they retrieve

gf712: ah yes true!

src/shogun/base/subset_iterators.h
```diff
@@ -0,0 +1,128 @@
...
+		{
+			return m_ptr == other.m_ptr && m_idx == other.m_idx;
+		}
+
+		bool operator!=(const SubsetIterator<T>& other)
+		{
+			return !(*this == other);
+		}
+
+		const reference operator*()
```
gf712: from what i understand this is legal c++ (at least for now and the compilers we support). dereferencing an iterator doesn't have to return a `const obj&`.. but this can cause issues (see http://ericniebler.com/2015/01/28/to-be-or-not-to-be-an-iterator/)

src/shogun/base/subset_iterators.h
gf712: btw back to the original question of passing const pointer, I don't think it is worth it. const pointer class members don't bring any (or at least negligible)  performance gains

src/shogun/base/subset_iterators.h
gf712: but then this won't compile: `const auto& val = *it;`

src/shogun/base/subset_iterators.h
karlnapf: can this iterator be used when I only have const a pointer say `const CDenseFeatures<...>*`?

src/shogun/base/subset_iterators.h
karlnapf: lol, all new stuff  for me :)

src/shogun/features/DenseFeatures.h
```diff
@@ -134,6 +137,27 @@ template<class ST> class CDenseFeatures: public CDotFeatures
...
+	}
+
+	/**
+	 * Returns an iterator to the element following the end of features.
+	 */
+	iterator end() noexcept
+	{
+		if (auto stack = get_subset_stack()->get_last_subset();
+			stack == nullptr)
+			return SubsetIterator(this, get_num_vectors());
```
karlnapf: any chance this can be shortened/moved inside the iterator?

src/shogun/features/DenseFeatures.h
```diff
@@ -134,6 +137,27 @@ template<class ST> class CDenseFeatures: public CDotFeatures
 	 */
 	SGVector<ST> get_feature_vector(int32_t num) const;
 
+	/**
+	 * Returns an iterator to the first vector of features.
+	 */
+	iterator begin() noexcept
```
karlnapf: This code is the same for all feature classes right?
Maybe we can move it somewhere (mixin?)

gf712: done!

src/shogun/features/DenseFeatures.h
gf712: yup, put this in the crtp base class

src/shogun/base/subset_iterators.h
gf712: but if it is const you wouldn't have subset? In that case need to add a const_iterator, should be pretty much the same

gf712: just need to make get_subset_stack() const and then it should work!

src/shogun/base/subset_iterators.h
```diff
@@ -0,0 +1,191 @@
...
+			        remove_cvptr_t<T>>::type,
+			    typename return_type_subset_iterator<remove_cvptr_t<T>>::type>;
+			using pointer = typename std::conditional_t<
+			    is_const,
+			    const typename return_type_subset_iterator<
+			        remove_cvptr_t<T>>::type*,
+			    typename return_type_subset_iterator<remove_cvptr_t<T>>::type*>;
+
+			using internal_pointer = typename std::conditional_t<
+			    is_const, const remove_cvptr_t<T>*, remove_cvptr_t<T>*>;
```
gf712: the logic is a bit convoluted in the whole class but i wanted to avoid writing two class: one for iterator and one for const_iterator

src/shogun/base/subset_iterators.h
```diff
@@ -0,0 +1,191 @@
...
+				    this_casted, this_casted->get_num_vectors());
+			else
+				return subset_iterator<decltype(this_casted)>(
+				    this_casted, stack->get_subset_idx().size());
+		}
+
+		/**
+		 * Returns a const iterator to the first vector of features/labels.
+		 */
+		auto begin() const
```
gf712: @karlnapf the range loop calls the const version if it's a const pointer

src/shogun/features/Features.h
```diff
@@ -282,7 +282,7 @@ class CFeatures : public CSGObject
 		 *
 		 * @return subset stack
 		 */
-		virtual CSubsetStack* get_subset_stack();
+		virtual CSubsetStack* get_subset_stack() const;
```
gf712: has to be const for const iterator to work

src/shogun/base/subset_iterators.h
```diff
@@ -0,0 +1,191 @@
...
+	{
+	};
+
+	template <typename T>
+	struct return_type_subset_iterator<CDenseFeatures<T>>
+	{
+		using type = SGVector<T>;
+	};
+
+	template <template <typename> class IterableSubsetContainer, typename ST>
```
karlnapf: good names!

src/shogun/base/subset_iterators.h
karlnapf: yep, looks ok to me for now

src/shogun/base/subset_iterators.h
```diff
@@ -0,0 +1,191 @@
...
+			subset_iterator(internal_pointer ptr, index_t idx = 0)
+			{
+				if (ptr->get_subset_stack()->get_last_subset() != nullptr)
+					m_subset = ptr->get_subset_stack()
+					               ->get_last_subset()
+					               ->get_subset_idx();
+				m_idx = idx;
+				// need to keep this variable so that the vector destructor
+				// isn't called
+				m_argsorted_subset = CMath::argsort(m_subset);
```
karlnapf: maybe we could move this method to SGVector in a separate PR...for later

src/shogun/base/subset_iterators.h
karlnapf: sweet, very nice template magic!

src/shogun/features/Features.cpp
```diff
@@ -287,7 +287,7 @@ void CFeatures::remove_all_subsets()
 	subset_changed_post();
 }
 
-CSubsetStack* CFeatures::get_subset_stack()
+CSubsetStack* CFeatures::get_subset_stack() const
```
karlnapf: is better imo

karlnapf:
Alright, what is missing? :)
Use case?

gf712:
yup, would be good to replace a loop somewhere.. What was the code you showed me in the ws?

karlnapf:
`void LDASolver<T>::compute_means()` I think

gf712:
ah but this is with labels, which aren't templated... so need to do that first I think

karlnapf:
I think one loop is just feature no?

gf712:
you mean this snippet no?
https://github.com/shogun-toolbox/shogun/blob/9d7191e0fe9566d3493be26b802548a164a9e353/src/shogun/solver/LDASolver.h#L126-L131

karlnapf:
Yes, ah it has labels indeed

This brings up a thought: could we have an iterator that takes labels AND features, like a composition of the one you wrote here?

gf712:
I guess so. Can have either a make_iterator(features, labels) or zip(feature_iterator, labels_iterator)

karlnapf:
I like the idea of a generic zip that composes iterators

src/shogun/solver/LDASolver.h
```diff
@@ -110,32 +111,31 @@ namespace shogun
...
+			m_class_mean[i] = SGVector<T>(m_features->get_num_features());
 			linalg::zero(m_class_mean[i]);
 		}
-		m_mean = SGVector<T>(data.num_rows);
+		m_mean = SGVector<T>(m_features->get_num_features());
 		linalg::zero(m_mean);
 
-		// calculate the total mean and the classes' mean.
-		for (index_t i = 0; i < data.num_cols; ++i)
+		for (const auto& [data_i, label_i]: zip_iterator(m_features, m_labels))
```
karlnapf: very nice! I like this a lot!

src/shogun/base/zip_iterator.h
```diff
@@ -0,0 +1,107 @@
...
+
+#ifndef SHOGUN_ZIP_ITERATOR_H
+#define SHOGUN_ZIP_ITERATOR_H
+
+#include <tuple>
+#include <utility>
+
+namespace shogun
+{
+	template <typename... Args1, typename... Args2, size_t... Idx>
```
karlnapf: I can't figure out: does this thing check that the two iterators have the same "size" somehow? Or does that even make sense?

src/shogun/solver/LDASolver.h
```diff
@@ -171,6 +171,13 @@ namespace shogun
 			    centered_class_col[c], data.get_column(i));
 			++centered_class_col[c];
 		}
+//        for (const auto& [data_i, label_i]: zip_iterator(m_features, m_labels))
```
karlnapf: forgotten?

src/shogun/solver/LDASolver.h
```diff
@@ -110,32 +111,31 @@ namespace shogun
 	void LDASolver<T>::compute_means()
 	{
 		index_t num_class = m_labels->get_num_classes();
-		auto data = m_features->get_feature_matrix();
```
karlnapf, line 109: We should benchmark memory use xvalidation before and after this PR.... or maybe at a later point at least! :)

src/shogun/base/zip_iterator.h
gf712: no, but i can add a runtime check that could do that (would need know what method to call). this checks that all the values of the current iteration are equals to another iteration, i.e. `iter == container.end()`, where `iter` would be a tuple of pointers

src/shogun/solver/LDASolver.h
gf712: was causing issues, so need to figure out what is happening here

src/shogun/solver/LDASolver.h
gf712: sure! i would imagine this is only faster when the matrix being used is larger than the L3 cache... i wonder if there is a noticeable impact on smaller examples. there might also be some unwanted copies, i need to check the code a few mote times...

gf712: i did some micro benchmarking with google benchmark and the iterator without subset is as fast as using the getter in the loop. However, so far the subset is always slower. The fact that we have to do two argsort operations (in begin and end) cannot really be compensated for..

src/shogun/solver/LDASolver.h
```diff
@@ -110,48 +111,44 @@ namespace shogun
...
-		auto data = m_features->get_feature_matrix().clone();
+		auto data = static_cast<CDenseFeatures<T>*>(m_features->clone());
 
 		// Center data with respect to each data point's class
-		for (index_t i = 0; i < data.num_cols; ++i)
-			linalg::add_col_vec(
-			    data, i, m_class_mean[m_labels->get_label(i)], data, (T)1.0,
-			    (T)-1.0);
+		for (auto [data_i, label_i]: zip_iterator(data, m_labels))
+			linalg::add(data_i, m_class_mean[label_i], data_i, static_cast<T>(1), static_cast<T>(-1));
```
gf712: @karlnapf I am not sure if this is any nicer than before? I guess it's a matter of taste! In any case, speed wise it'll be the same

gf712:
this is as fast as it is going to get I think. You can get some speed gains (sometimes), but right now it's a bit slower using it with features because there is no move constructor in SGVector. With labels this is faster than using a getter in a loop

src/shogun/solver/LDASolver.h
karlnapf: I like it

karlnapf:
I think the case where we gain speed is a big matrix and xvalidation (I.e. largish Training Sets with random subset indices) here the cache friendliness should kick in.
For the cases where this doesn’t make a difference, I think the additional sort doesn’t matter

src/shogun/solver/LDASolver.h
```diff
@@ -110,48 +111,44 @@ namespace shogun
...
 
 	template <typename T>
 	void LDASolver<T>::compute_within_cov()
 	{
 		index_t num_features = m_features->get_num_features();
-		index_t num_vectors = m_features->get_num_vectors();
 		index_t num_class = m_labels->get_num_classes();
 
-		auto data = m_features->get_feature_matrix().clone();
+		auto data = static_cast<CDenseFeatures<T>*>(m_features->clone());
```
karlnapf: why clone that matrix?

src/shogun/solver/LDASolver.h
karlnapf: Does this write into the feature vector? Because if so, no good! Shouldn't be allowed
If that is the case, we need to clone/copy the matrix above explicitly (instead of cloning the features)

gf712: To not be able to write to the original data you need to do `const auto&`. Ah i see, but doesn't the clone of features clone the matrix?

karlnapf:
merge soon?

karlnapf:
ah and did you see my comment re the cloning of the features?

src/shogun/solver/LDASolver.h
karlnapf: we want to avoid writing to the output of the iteratior definitely imo. that is by design of the iterator...otherwise people will use it.

and i am not sure why clone the features? not necessary

src/shogun/base/subset_iterators.h
```diff
@@ -0,0 +1,185 @@
...
+
+		template <typename T>
+		struct return_type_subset_iterator<CDenseLabels, T>
+		{
+			using type = float64_t;
+		};
+	} // namespace subset_iterator_detail
+
+	template <class IterableSubsetContainer, typename ST>
+	class SubsetIteratorBase
```
gf712: need to come up with a better name now that the subset doesn't change the behaviour

src/shogun/features/DenseFeatures.h
```diff
@@ -541,7 +542,7 @@ template<class ST> class CDenseFeatures: public CDotFeatures
 	 * offset, it throws an error. It then copies into the target matrix, starting
 	 * from base + (colum_offset * num_features) location.
 	 */
-	void copy_feature_matrix(SGMatrix<ST> target, index_t column_offset=0) const;
+	void copy_feature_matrix(SGMatrix<ST>& target, index_t column_offset=0) const;
```
gf712, line 544: avoid copying stuff around

src/shogun/labels/Labels.cpp
```diff
@@ -68,9 +68,9 @@ void CLabels::remove_all_subsets()
 	m_subset_stack->remove_all_subsets();
 }
 
-CSubsetStack* CLabels::get_subset_stack()
+CSubsetStack* CLabels::get_subset_stack() const
```
gf712: has to be const for const only labels iterator

src/shogun/lib/SGMatrix.cpp
```diff
@@ -100,6 +100,19 @@ SGMatrix<T>::SGMatrix(const SGMatrix &orig) : SGReferencedData(orig)
...
+	: SGReferencedData(std::move(orig)), matrix(orig.matrix),
+	  num_rows(orig.num_rows), num_cols(orig.num_cols),
+	  gpu_ptr(std::move(orig.gpu_ptr))
+{
+    orig.matrix = nullptr;
+    orig.num_rows = 0;
+    orig.num_cols = 0;
+	m_on_gpu.store(orig.m_on_gpu.load(
+			std::memory_order_acquire), std::memory_order_release);
+}
```
gf712, line 115: move constructor for sgmatrix works fine. I am just not sure what happens when using GPU backend... In theory this is how it should be done!

src/shogun/lib/SGMatrix.cpp
```diff
@@ -409,14 +422,23 @@ SGVector<T> SGMatrix<T>::get_column(index_t col) const
 }
 
 template <class T>
-void SGMatrix<T>::set_column(index_t col, const SGVector<T> vec)
+void SGMatrix<T>::set_column(index_t col, const SGVector<T>& vec)
```
gf712, line 424: this should be const ref, no need for copy

src/shogun/lib/SGMatrix.cpp
```diff
@@ -409,14 +422,23 @@ SGVector<T> SGMatrix<T>::get_column(index_t col) const
...
 {
 	assert_on_cpu();
 	ASSERT(!vec.on_gpu())
 	ASSERT(vec.vlen == num_rows)
 	sg_memcpy(&matrix[num_rows * col], vec.vector, sizeof(T) * num_rows);
 }
 
+template <class T>
+void SGMatrix<T>::set_column(index_t col, SGVector<T>&& vec)
```
gf712, line 433: does exactly the same thing as above, but users can still pass xvalues (or is it prvalues)? example:
```cpp
mymatrix.set_column(idx, linalg::some_op(...))
```

src/shogun/lib/SGMatrix.h
```diff
@@ -160,11 +163,17 @@ template<class T> class SGMatrix : public SGReferencedData
 		 */
 		SGVector<T> get_column(index_t col) const;
 
-		/** Copy the content of a vector into a column
+		/** Copy the content of a vector lvalue reference into a column
```
gf712, line 180: not sure if it makes sense to says this explicitly in the docs? Docs should probably not mention these things

src/shogun/lib/SGReferencedData.cpp
```diff
@@ -21,6 +21,11 @@ SGReferencedData::SGReferencedData(const SGReferencedData &orig)
 	ref();
 }
 
+SGReferencedData::SGReferencedData(SGReferencedData&& orig) noexcept: m_refcount(orig.m_refcount)
```
gf712: this is the move constructor for refcounting. basically when you move a container it passes on the refcounting from the original to the new object and avoids double free

src/shogun/lib/SGVector.h
```diff
@@ -178,6 +182,12 @@ template<class T> class SGVector : public SGReferencedData
 		/** Returns an iterator to the element following the last element of the container. */
 		iterator end() noexcept { return iterator(vector + vlen); }
 
+		/** Returns a const iterator to the first element of the container. */
+		const_iterator begin() const noexcept { return const_iterator(vector); }
+
+		/** Returns a const iterator to the element following the last element of the container. */
+		const_iterator end() const noexcept { return const_iterator(vector + vlen); }
```
gf712, line 188: these are useful to have. I use them in the tests

src/shogun/solver/LDASolver.h
```diff
@@ -164,12 +165,14 @@ namespace shogun
 			centered_class[i] = SGMatrix<T>(num_features, m_class_count[i]);
 			linalg::zero(centered_class[i]);
 		}
-		for (index_t i = 0; i < num_vectors; ++i)
+		counter=0;
```
gf712: need to track the row number because the matrix iterator is along elements rather than the column vectors

gf712:
@karlnapf i think this is almost fine to go I added some comments/ Let me know if you agree!

gf712:
also should wait for the shared_ptr PR before merging, otherwise it'll be another thing for @vigsterkr to rebase against :D

src/shogun/base/subset_iterators.h
karlnapf: We can also do that later ;)

src/shogun/features/DenseFeatures.h
karlnapf: I think these things should be removed eventually

src/shogun/lib/SGMatrix.h
karlnapf: I think it does
Is not user facing anyways

karlnapf:
Agree on the rebase ;)

src/shogun/features/DenseFeatures.h
karlnapf: Actually might be good
But why not return the matrix?

src/shogun/solver/LDASolver.h
```diff
@@ -110,49 +111,49 @@ namespace shogun
...
 	template <typename T>
 	void LDASolver<T>::compute_within_cov()
 	{
 		index_t num_features = m_features->get_num_features();
-		index_t num_vectors = m_features->get_num_vectors();
 		index_t num_class = m_labels->get_num_classes();
 
-		auto data = m_features->get_feature_matrix().clone();
+		auto mean_matrix = SGMatrix<T>(num_features, m_features->get_num_vectors());
+		linalg::zero(mean_matrix);
```
karlnapf: No need th matrix ctor does this (although I don’t like that)

src/shogun/features/DenseFeatures.h
gf712: i guess because RVO is not guaranteed by the standard until c++17

karlnapf:
This was just a rebase while we’re waiting for smart pointers? :)

gf712:
> This was just a rebase while we’re waiting for smart pointers? :)

Yup, and some cleaning up and checking if it still works... tbh can merge it and worry about smart pointers later... :D

src/shogun/base/range.h
```diff
@@ -64,9 +64,16 @@ namespace shogun
 		}
 
 		/** @class Iterator spawned by @ref Range. */
-		class Iterator : public std::iterator<std::input_iterator_tag, T>
+		class Iterator
```
vigsterkr, line 67: 👍

src/shogun/features/DenseFeatures.h
```diff
@@ -60,7 +61,7 @@ class DotFeatures;
  * See comments to find out whether it is supported for that method.
  * See also Features class documentation
  */
-template<class ST> class DenseFeatures: public DotFeatures
+template<class ST> class DenseFeatures: public DotFeatures, public SGContainerIteratorBase<DenseFeatures<ST>, ST>
```
vigsterkr, line 64: mmmm what happens in this case for langs where multi inheritance is not supported (like java)?

vigsterkr: see swig docs: `Note that Java does not support multiple inheritance so any multiple inheritance in the C++ code is not going to work`

src/shogun/labels/DenseLabels.h
```diff
@@ -31,7 +32,7 @@ namespace shogun
 	 * label
 	 * (vectors).
 	 */
-	class DenseLabels : public Labels
+	class DenseLabels : public Labels, public SGContainerIteratorBase<DenseLabels, float64_t>
```
vigsterkr, line 35: same question about multi inheritance, although this should be hidden somehow (if i get it right) that we never actually expose DenseLabels directly to swig

src/shogun/features/DenseFeatures.h
gf712: I'm not sure actually.... I didn't get any errors/warnings in the tests. The files are exposed in the interfaces via swig. Can hide it from swig?

gf712: in any case, it says the second class will be ignored, which is fine, the interface shouldn't have access to this functionality right?

PR #4682, opened 2019-06-10
check for c++17 features used in shogun

gf712:
added checks for c++17 features:
 - if constexpr
 - folding expressions
 - init-statements for if

let me know what else you need

### DIFF
--- a/.ci/ci.yml
+++ b/.ci/ci.yml
@@ -39,7 +39,7 @@ jobs:
       clang:
         CC: "clang-4.0"
         CXX: "clang++-4.0"
-        cmakeOptions: '-DCLANG_DEFAULT_CXX_STDLIB=libc++'
+        cmakeOptions: ''
       gcc:
         CC: "gcc"
         CXX: "g++"
--- a/.ci/ci.yml
+++ b/.ci/ci.yml
@@ -39,7 +39,7 @@ jobs:
       clang:
         CC: "clang-4.0"
         CXX: "clang++-4.0"
-        cmakeOptions: ''
+        cmakeOptions: '-DCLANG_DEFAULT_CXX_STDLIB=libc++'
       gcc:
         CC: "gcc"
         CXX: "g++"
--- a/src/shogun/CMakeLists.txt
+++ b/src/shogun/CMakeLists.txt
@@ -351,6 +351,48 @@ IF (NOT HAVE_STD_VARIANT)
   )
 ENDIF()
 
+CHECK_CXX_SOURCE_COMPILES("
+template<typename... Args>
+auto func(Args... args) {
+    return (args + ...);
+}
+int main()
+{return func(1,2,3);}
+" HAVE_FOLDING_EXPRESSIONS)
+
+IF(NOT HAVE_FOLDING_EXPRESSIONS)
+  MESSAGE(FATAL_ERROR "Compiling shogun requires a compiler with folding expression support!")
+ENDIF()
+
+CHECK_CXX_SOURCE_COMPILES("
+int main()
+{
+  if constexpr(true)
+    return 1;
+}
+" HAVE_IF_CONSTEXPR)
+
+IF(NOT HAVE_IF_CONSTEXPR)
+  MESSAGE(FATAL_ERROR "Compiling shogun requires a compiler with if constexpr support!")
+ENDIF()
+
+CHECK_CXX_SOURCE_COMPILES("
+int main()
+{
+  if (int value = 1; value==1)
+  {
+    return value;
+  }
+  else
+  {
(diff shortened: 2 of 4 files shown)


src/shogun/CMakeLists.txt
```diff
@@ -351,6 +351,50 @@ IF (NOT HAVE_STD_VARIANT)
...
+auto func(Args... args) {
+    return (args + ...);
+}
+int main()
+{return func(1,2,3);}
+" HAVE_FOLDING_EXPRESSIONS)
+
+IF(NOT HAVE_FOLDING_EXPRESSIONS)
+  MESSAGE(FATAL_ERROR "Compiling shogun requires a compiler with folding expression support!")
+ENDIF()
```
gf712: right now this doesn't have to be fatal, but would be good to be able to rely on this

src/shogun/CMakeLists.txt
```diff
@@ -351,6 +351,50 @@ IF (NOT HAVE_STD_VARIANT)
...
+    return value;
+  }
+  else
+  {
+    return value - 1;
+  }
+}
+" HAVE_IF_INIT)
+IF(NOT HAVE_IF_INIT)
+  MESSAGE(FATAL_ERROR "Compiling shogun requires a compiler with if init-statements support!")
```
gf712: same here

lisitsyn:
Yes!

gf712:
Yes, need clang 4 and gcc 7 in the dockerfile. But there is no stable support for gcc 7 in Debian yet..

lisitsyn:
> Yes, need clang 4 and gcc 7 in the dockerfile. But there is no stable support for gcc 7 in Debian yet..

Why Debian? I think we can install them from PPA in Ubuntu 16

gf712:
> Why Debian? I think we can install them from PPA in Ubuntu 16
I think the dockerfile uses debian. And the CI in linux uses the docker image no? need to check..

.ci/ci.yml
```diff
@@ -39,7 +39,7 @@ jobs:
       clang:
         CC: "clang-4.0"
         CXX: "clang++-4.0"
-        cmakeOptions: ''
+        cmakeOptions: '-DCLANG_DEFAULT_CXX_STDLIB=libc++'
```
vigsterkr: this is just a hack... just saying... plz dont ever merge it.

gf712: yup just checking if this is the issue

vigsterkr:
> > Yes, need clang 4 and gcc 7 in the dockerfile. But there is no stable support for gcc 7 in Debian yet..
> 
> Why Debian? I think we can install them from PPA in Ubuntu 16

if you upgrade donbot to ubuntu 16 etc then go ahead... see our buildbot infra

lisitsyn:
Oh I see what you mean :)

vigsterkr:
FYI
```/usr/include/python2.7/unicodeobject.h:534:5: error: ISO C++17 does not allow 'register' storage class specifier [-Wregister]
    register PyObject *obj,     /* Object */
    ^~~~~~~~~
/usr/include/python2.7/unicodeobject.h:553:5: error: ISO C++17 does not allow 'register' storage class specifier [-Wregister]
    register PyObject *obj      /* Object */
    ^~~~~~~~~
/usr/include/python2.7/unicodeobject.h:575:5: error: ISO C++17 does not allow 'register' storage class specifier [-Wregister]
    register const wchar_t *w,  /* wchar_t buffer */
    ^~~~~~~~~
/usr/include/python2.7/unicodeobject.h:593:5: error: ISO C++17 does not allow 'register' storage class specifier [-Wregister]
    register wchar_t *w,        /* wchar_t buffer */
    ^~~~~~~~~
In file included from /home/wiking/shogun/build/src/interfaces/python/shogunPYTHON_wrap.cxx:175:
In file included from /usr/include/python2.7/Python.h:94:
/usr/include/python2.7/stringobject.h:173:5: error: ISO C++17 does not allow 'register' storage class specifier [-Wregister]
    register PyObject *obj,     /* string or Unicode object */
    ^~~~~~~~~
/usr/include/python2.7/stringobject.h:174:5: error: ISO C++17 does not allow 'register' storage class specifier [-Wregister]
    register char **s,          /* pointer to buffer variable */
    ^~~~~~~~~
/usr/include/python2.7/stringobject.h:175:5: error: ISO C++17 does not allow 'register' storage class specifier [-Wregister]
    register Py_ssize_t *len    /* pointer to length variable or NULL
    ^~~~~~~~~```

so basically python2.7 is bye bye bye....

karlnapf:
Kill it, think official support stops soon as well

PR #4683, opened 2019-06-12
 register doesn't do anything anyway in modern compilers

gf712:
@vigsterkr can we just do this?
the `#define` is inserted before the `#include <Python/ruby/... header>`


vigsterkr:
closing...

PR #4684, opened 2019-06-12
Generic linalg::dot

theartful:
Enable dot product between float64_t and any other type.


src/shogun/mathematics/linalg/LinalgEnums.h
```diff
@@ -32,12 +32,20 @@
...
 namespace shogun
 {
 
 	namespace linalg
 	{
-
+		template <typename T, typename U>
+		struct promote
+		{
+			using type = decltype(T(0) + U(0));
```
gf712: can you not just do decltype(T + U);?

gf712: nvm!

gf712: actually to generalise this you could do `decltype(T{}+U{})`

theartful: this is much cleaner

src/shogun/mathematics/linalg/backend/eigen/BasicOps.cpp
```diff
@@ -348,11 +358,21 @@ void LinalgBackendEigen::add_scalar_impl(SGMatrix<T>& a, T b) const
...
+	    	.dot(typename SGVector<U>::EigenVectorXtMap(b));
+	else if constexpr (std::is_same<T, TU>::value)
+		return (typename SGVector<T>::EigenVectorXtMap(a))
+			.dot(typename SGVector<U>::EigenVectorXtMap(b).template cast<T>());
+	else if constexpr (std::is_same<U, TU>::value)
+		return (typename SGVector<T>::EigenVectorXtMap(a).template cast<U>())
+			.dot(typename SGVector<U>::EigenVectorXtMap(b));
+	else
+		return (typename SGVector<T>::EigenVectorXtMap(a).template cast<TU>())
+			.dot(typename SGVector<U>::EigenVectorXtMap(b).template cast<TU>());
```
theartful: I benchmarked this. Eigen is clever enough to ignore the cast of no cast is needed, so I might as well cast in all casses and get rid of the if statements.

src/shogun/mathematics/linalg/LinalgEnums.h
```diff
@@ -32,12 +32,20 @@
...
+#include <shogun/lib/common.h>
+
 namespace shogun
 {
 
 	namespace linalg
 	{
-
+		template <typename T, typename U>
+		struct promote
```
karlnapf: this is a nice idea!

src/shogun/mathematics/linalg/LinalgBackendEigen.h
```diff
@@ -544,8 +553,8 @@ namespace shogun
 		T cross_entropy_impl(const SGMatrix<T>& p, const SGMatrix<T>& q) const;
 
 		/** Eigen3 vector dot-product method */
-		template <typename T>
-		T dot_impl(const SGVector<T>& a, const SGVector<T>& b) const;
+		template <typename T, typename U, typename TU = typename linalg::promote<T, U>::type>
```
karlnapf: I wonder whether it would be a point to let the caller decide the return type? But maybe that is overkill ...

src/shogun/mathematics/linalg/backend/eigen/BasicOps.cpp
```diff
@@ -348,11 +358,13 @@ void LinalgBackendEigen::add_scalar_impl(SGMatrix<T>& a, T b) const
...
-T LinalgBackendEigen::dot_impl(const SGVector<T>& a, const SGVector<T>& b) const
+template <typename T, typename U, typename TU>
+TU LinalgBackendEigen::dot_impl(const SGVector<T>& a, const SGVector<U>& b) const
 {
-	return (typename SGVector<T>::EigenVectorXtMap(a))
-	    .dot(typename SGVector<T>::EigenVectorXtMap(b));
+	typename SGVector<T>::EigenMatrixXtMap a_eig = a;
+	typename SGVector<U>::EigenMatrixXtMap b_eig = b;
+
+	return a_eig.template cast<TU>().dot(b_eig.template cast<TU>());
```
karlnapf: Could you check whether eigen SIMDs this if the types are the same?
I know things are casted on the fly, but that makes the whole thing slower if done when the types are the same

theartful: yeah I checked, same performance with O3 optimization. It gets much slower when using different types
```
2019-06-13 23:32:58
Running ./out
Run on (4 X 3000 MHz CPU s)
CPU Caches:
  L1 Data 32K (x2)
  L1 Instruction 32K (x2)
  L2 Unified 256K (x2)
  L3 Unified 4096K (x1)
Load Average: 2.77, 4.82, 5.70
***WARNING*** CPU scaling is enabled, the benchmark real time measurements may be noisy and will incur extra overhead.
-----------------------------------------------------------------
Benchmark                       Time             CPU   Iterations
-----------------------------------------------------------------
BM_std_dot                  10237 ns        10233 ns        67238
BM_dot                       3654 ns         3654 ns       190775
BM_dot_without_casting       3644 ns         3644 ns       193444
BM_loop_dot                 30550 ns        30549 ns        22922
```
BM_dot is eigen without casting. std_dot is std::inner_product. This is the test case
```
static void BM_dot(benchmark::State& state)
{
	SGVector<float64_t> x(10000);
	x.range_fill(1);

	SGVector<float64_t> y(10000);
	y.set_const(1);

	for (auto _ : state)
	{
		benchmark::DoNotOptimize(dot(x, y));
	}
}
```

src/shogun/mathematics/linalg/LinalgBackendEigen.h
theartful: it should be possible if you provided `TU`. However, it would throw an error since these casses aren't specialized.

src/shogun/mathematics/linalg/backend/eigen/BasicOps.cpp
karlnapf: Ok thanks!
I wonder: If we take such a performance penalty, then why offer this?

theartful: Eigen is still faster than using for loops with different types. Also it will make things more elegant (as in DenseFeatures where loops are used when different types are multiplied).

karlnapf: ok cool. Just saying though, people tend to use code without thinking about it. So sometimes it might be better to not allow certain things, and force people to be more explicit in their code. But here with the features, ok makes sense, just looking at the other Pr

theartful: should we make it harder to use this? like by passing a "slow_dot" tag or something

karlnapf: I think that might be an idea. Only question is whether this is overkill ;)
But let’s try

theartful: @karlnapf It looks ugly. I have to ask about the type in the tag in LinalgNamespace, and then dispatch accordingly since the backend can't do that

karlnapf: Maybe a decorator that one hs to pass to avoid compiler warning? Not sure that is even possible....

gf712: why don't you have an allow_cast templated bool that is by default false and if you have to cast it is a compiler error and then a user has to set it as true if he wants to cast/knows he has to cast?

theartful: @gf712  That is more elegant, but it's runtime. I found the compile time solution ugly since LinalgNamespace has to make the decision not the backend

gf712: you mean you need something for runtime?

src/shogun/mathematics/linalg/LinalgNamespace.h
```diff
@@ -845,13 +845,21 @@ namespace shogun
 		 * represented
 		 * as \f$\sum_i a_i b_i\f$
 		 */
-		template <typename T, typename U, typename TU = typename promote<T, U>::type>
-		TU dot(const SGVector<T>& a, const SGVector<U>& b)
+		template <
+		    typename T, typename U, typename TU = typename promote<T, U>::type,
+		    typename Tag = void*>
```
gf712: what i mean is you can have here instead `template<..., bool allow_cast = false>` and that is compile time evaluated no?

theartful: Ah yes i misunderstood you. But what about type deduction? We would have to specify T, U and TU before we get to the boolean.

gf712: Ah ok, could you put the Boolean first?

theartful: That would work :)

gf712: It's just a suggestion btw!

theartful: I find tags more expressive, so I will use the name allow_cast, since likely_slow is horrible

gf712: sounds good!

PR #4685, opened 2019-06-12
Dense dot range cleanup

theartful:


src/shogun/features/DenseFeatures.cpp
```diff
@@ -865,6 +865,33 @@ SGVector<ST> CDenseFeatures<ST>::dot(const SGVector<ST>& other) const
 	return linalg::matrix_prod(get_feature_matrix(), other, false);
 }
 
+template <typename ST>
+void CDenseFeatures<ST>::dot(
+	SGVector<float64_t> output, IntRange feat_range,
```
karlnapf: I wonder: isn't "range" something that could be achieved with subsets? Instead of passing the parameter around?

src/shogun/features/DenseFeatures.cpp
```diff
@@ -865,6 +865,33 @@ SGVector<ST> CDenseFeatures<ST>::dot(const SGVector<ST>& other) const
...
+	if (m_subset_stack->has_subsets() || !feature_matrix.matrix)
+		return CDotFeatures::dot(output, feat_range, vec, alphas, b);
+
+	ASSERT(feat_range.m_begin >= 0)
+	ASSERT(feat_range.m_end <= get_num_vectors())
+
+	auto sliced_matrix =
+		feature_matrix.submatrix(feat_range.m_begin, feat_range.m_end);
+
+	if constexpr (std::is_same<ST, float64_t>::value)
```
karlnapf: yessss

src/shogun/features/DenseFeatures.cpp
```diff
@@ -865,6 +865,33 @@ SGVector<ST> CDenseFeatures<ST>::dot(const SGVector<ST>& other) const
...
+		linalg::matrix_prod(sliced_matrix, vec, output, true);
+	else
+		linalg::matrix_prod(
+			sliced_matrix.template as<float64_t>(), vec, output, true);
+
+	if (alphas.size() != 0)
+		linalg::element_prod(output, alphas, output);
+	if (b != 0)
+		linalg::add_scalar(output, b);
+}
```
karlnapf: really like this now. Doing the batch thing if possible, otherwise the normal thing...quite cute!

src/shogun/features/DenseFeatures.cpp
theartful: I thought about this. subsets are not ranges, we only provide functionality for subsets. To represent a range, we would have to create an array with every single index in the range, so treating ranges as subsets isn't the best way I think.

src/shogun/lib/SGMatrix.h
```diff
@@ -133,6 +134,15 @@ template<class T> class SGMatrix : public SGReferencedData
 		virtual ~SGMatrix();
 
 #ifndef SWIG // SWIG should skip this parts
+		/** @return a (copied) typed matrix with same content */
+		template <class X>
+		SGMatrix<X> as() const
```
karlnapf: ++

src/shogun/lib/SGMatrix.h
```diff
@@ -133,6 +134,15 @@ template<class T> class SGMatrix : public SGReferencedData
...
+		{
+			SGMatrix<X> mat(num_rows, num_cols, true);
```
karlnapf: the true here technically is not needed....do you prefer it?

karlnapf:
This is nice. Maybe a unit test that ensures the same results of the multiple versions?

src/shogun/features/DenseFeatures.cpp
theartful: It would be better imo if we provided functionality to "slice" a feature object using a range, instead of individual indices. We can start with DenseFeatures and other features where this clear how to do efficiently, and the other features make them reduce to the subset behaviour.

karlnapf: yeah I agree the implementation would have to account for that. But it could right?
I am more talking about the conceptual interface. 
Subset is a class that could be overloaded btw ....

theartful:
@karlnapf I added a unit test for dense_dot_range in a previous PR, and it passes. It uses the old interface, which now only calls the new one. I will add a unit test for the matrix multiplication thing.

src/shogun/lib/SGMatrix.h
theartful: nope, I will change it

karlnapf:
> @karlnapf I added a unit test for dense_dot_range in a previous PR, and it passes. It uses the old interface, which now only calls the new one. I will add a unit test for the matrix multiplication thing.

sure I was aware of that. the new one would just check that they produce the same result, so a bit simpler than checking for correct results. Nice one :)

src/shogun/features/DenseFeatures.cpp
theartful: alright, so a subset can either have an array of indices, or a range. if it's a range, it should be able to do the same functionality as the one with indices so not to break things.
we add a view function that accepts a range instead of SGVector of indices
we remove the IntRange thing, and also remove the other "dot" method that multiplies with the whole matrix, and leave this with the "alphas" and "bias"

theartful: @karlnapf sounds reasonable?

karlnapf: Yes polymorphic subsets/views
And the we get a cleaner api

I wonder whether there are downsides to this? You recently looked at all the dot range methods, something came up?
Only thing I can think of are the combined features where a view would shallow copy all sub features...

karlnapf:
rebase and also try to merge this soon?

karlnapf:
btw there are still quite a few `dense_dot` and `dense_dot_range` in the codebase.... these should be replaced I guess?

theartful:
Yeah they should, and ideally by iterators. The `dense_dot` left is the other version that takes anot her Features object and two indices.

gf712:
is this still going to be worked on? (doing some spring cleaning)

karlnapf:
bump

PR #4687, opened 2019-06-13
Use aligned malloc for SGVector and SGMatrix allocation

theartful:
@vigsterkr I made SG_ALIGNED_MALLOC reduce to SG_MALLOC if aligned malloc is not available, so as to not repeat the macroed if statement. But now SG_CALLOC is turned to aligned malloc followed by std::fill, so I'm not sure.

### DIFF
--- a/doc/OpenCV_docs/OpenCV_SVM_vs_Shogun_SVM.md
+++ b/doc/OpenCV_docs/OpenCV_SVM_vs_Shogun_SVM.md
@@ -22,6 +22,7 @@ Let's start with the includes!
 #include <shogun/multiclass/MulticlassLibSVM.h>
 #include <shogun/labels/MulticlassLabels.h>
 #include <shogun/kernel/LinearKernel.h>
+#include <shogun/mathematics/linalg/LinalgNamespace>
 
 // for measuring time
 #include <omp.h>
@@ -191,7 +192,7 @@ We start with creating the training data as the ```DenseFeatures```.
 
 ```CPP
     SGMatrix<float64_t> shogun_traindata = CV2SGFactory::get_sgmatrix<float64_t>(traindata);
-    SGMatrix<float64_t>::transpose_matrix(shogun_traindata.matrix, shogun_traindata.num_rows, shogun_traindata.num_cols);
+    shogun_traindata = linalg::transpose_matrix(shogun_traindata);
     CDenseFeatures<float64_t>* shogun_trainfeatures = new CDenseFeatures<float64_t>(shogun_traindata);
 ```
 
@@ -220,7 +221,7 @@ Now we are ready to initialize the SVM for Shogun. We train it here!
 Prepare the testing data.
 ```CPP
     SGMatrix<float64_t> shogun_testdata=CV2SGFactory::get_sgmatrix<float64_t>(testdata);
-    SGMatrix<float64_t>::transpose_matrix(shogun_testdata.matrix, shogun_testdata.num_rows, shogun_testdata.num_cols);
+    shogun_testdata = linalg::transpose_matrix(shogun_testdata);
     CDenseFeatures<float64_t>* testfeatures=new CDenseFeatures<float64_t>(shogun_testdata);
 ```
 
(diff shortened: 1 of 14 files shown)


vigsterkr:
Perfect! Only maybe we should add a compile time warning that it is not actually aligned if aligned malloc is not available. #warning macro? I know #error exists one would hope that warning as well

theartful:
yup, warning exists  https://gcc.gnu.org/onlinedocs/gcc-4.5.0/cpp/Diagnostics.html

src/shogun/lib/memory.h
```diff
@@ -55,6 +55,10 @@ void operator delete[](void *p, std::align_val_t al);
 #else
 #define SG_ALIGNED_MALLOC(type, len, al) sg_aligned_malloc<type>(size_t(len), al)
 #endif //TRACE_MEMORY_ALLOCS
+
+#else
+#warning "Aligned malloc is not available"
```
gf712: `#pragma message("Aligned malloc is not available")` for msvc

theartful: right, thanks

src/shogun/lib/SGMatrix.cpp
```diff
@@ -364,7 +366,8 @@ template <class T>
 void SGMatrix<T>::transpose_matrix(T*& matrix, int32_t& num_feat, int32_t& num_vec)
```
karlnapf: IMO, we should remove this method. Not necessary and slow compared to passing flags to linalg/eigen

theartful: We can get rid of this, and replace its uses with linalg::transpose_matrix, and in another PR refactor them.

karlnapf: As said, I think this is an operation that shouldn’t be offered. People we it mindlessly. With the current linalg, we use flags in the operations. In some expression template approach this would be stored as a flag(like in eigen)

vigsterkr:
@theartful sorry just merged something that causes conflict with this one... could you plz rebase and push so we can merge this? thnx and sorry

src/shogun/lib/SGMatrix.cpp
theartful: I removed `transpose_matrix` operation. It was mainly used in unit-tests and documentation, so I changed those to linalg::transpose_matrix. NewtonSVM used it, but it was easy to refactor.

theartful:
@vigsterkr done

examples/undocumented/libshogun/mathematics_lapack.cpp
```diff
@@ -68,9 +69,9 @@ void test_matrix_multiply()
 	SGMatrix<float64_t> A(n, m);
 	SGVector<float64_t>::range_fill_vector(A.matrix, m*n);
 	SGMatrix<float64_t>::display_matrix(I, "I");
-	SGMatrix<float64_t>::transpose_matrix(A.matrix, A.num_rows, A.num_cols);
```
karlnapf: this test should be done without transposes, because the multiply method accepts flags for that

doc/OpenCV_docs/OpenCV_SVM_vs_Shogun_SVM.md
```diff
@@ -191,7 +192,7 @@ We start with creating the training data as the ```DenseFeatures```.
 
 ```CPP
     SGMatrix<float64_t> shogun_traindata = CV2SGFactory::get_sgmatrix<float64_t>(traindata);
```
karlnapf: maybe we can change this method to transpose the matrix upon creation? rather than doing it afterwards?

src/shogun/classifier/svm/NewtonSVM.cpp
```diff
@@ -91,7 +91,6 @@ void CNewtonSVM::iteration()
 		sgv = features->get_computed_dot_feature_vector(sv[k]);
 		sg_memcpy(&Xsv.matrix[k * x_d], sgv.data(), sizeof(float64_t) * (x_d));
 	}
-	Xsv = linalg::transpose_matrix(Xsv);
```
karlnapf: yes!
more efficient!

src/shogun/mathematics/Statistics.cpp
```diff
@@ -731,7 +732,7 @@ SGMatrix<float64_t> CStatistics::sample_from_gaussian(SGVector<float64_t> mean,
 		s=ldlt.solve(s);
 	}
 
-	SGMatrix<float64_t>::transpose_matrix(S.matrix, S.num_rows, S.num_cols);
+	S = linalg::transpose_matrix(S);
```
karlnapf: right, here we actually to need this :(

PR #4690, opened 2019-06-14
Port SVM examples

theartful:


src/shogun/kernel/string/WeightedDegreeStringKernel.cpp
```diff
@@ -978,6 +976,8 @@ bool CWeightedDegreeStringKernel::set_max_mismatch(int32_t max)
 
 void CWeightedDegreeStringKernel::init()
 {
+	type = E_WD;
```
ghost, line 979: type was uninitialized if default constructor is used

src/shogun/kernel/string/WeightedDegreeStringKernel.cpp
```diff
@@ -137,6 +132,9 @@ void CWeightedDegreeStringKernel::create_empty_tries()
 
 bool CWeightedDegreeStringKernel::init(CFeatures* l, CFeatures* r)
 {
+	if (type != E_EXTERNAL)
+		set_wd_weights_by_type(type);
```
ghost, line 136: to initialize weights based on type, since "put" changes the type quietly

vinx13: I think rather than checking which overloading of `init` is called, you can check whether features have been initialized via `put`. This will allow us to initialize kernels in this style, which has been already used in a few examples
```
k = kernel('WeightedDegreeStringKernel', lhs=..., rhs=...)
k.init()
```

examples/meta/src/binary/svmlight.sg
```diff
@@ -0,0 +1,24 @@
+File f_feats_train = csv_file("../../data/fm_train_dna.dat")
+File f_feats_test = csv_file("../../data/fm_test_dna.dat")
+File f_labels_train = csv_file("../../data/label_train_dna.dat")
+
+ #![create_features]
+Features feats_train = string_features(f_feats_train, enum EAlphabet.DNA)
```
vinx13, line 6: nice work, now we have an example using `string_features`

src/shogun/kernel/string/WeightedDegreeStringKernel.cpp
ghost: the Kernel interface has only one init, which accepts lhs and rhs, do you mean Distance?
`‘void shogun::CKernel::init()’ is private within this context`

src/shogun/kernel/Kernel.cpp
```diff
@@ -109,15 +109,15 @@ bool CKernel::init(CFeatures* l, CFeatures* r)
...
-
 	SG_REF(l);
 	if (l==r)
 		lhs_equals_rhs=true;
 	else // l!=r
 		SG_REF(r);
 
+	//remove references to previous features
+	remove_lhs_and_rhs();
+
```
ghost: Factories create objects with refcount = 0. When init is called with two features, and then is called again with a feature in common, calling "remove_lhs_and_rhs" first  results in the feature object being deleted, causing segfaults.

ghost: why don't factories reference the objects before returning them?

karlnapf: Was recently rly changed, Viktor can comment
I would just add a workaround as soon we will use shared ptr

src/shogun/kernel/string/WeightedDegreeStringKernel.cpp
karlnapf: Wuwei we marked lhs ans rhs as read only to force users to call kernel.init recently 
Avoids all sorts of problems
So this change is good

vinx13: Didn't know that, thanks for the clarification

src/shogun/kernel/Kernel.cpp
gf712: That’s because swig increases the refcount instead. So we ended with refcount of 2, and that leaked memory

ghost: @gf712 I checked it on my machine. The kernel I created had a refcount of zero.

ghost: the feature object i mean, sorry

ghost: @gf712 I checked other objects. You're right, swig does increase the reference count for all except string_features

vinx13: @theartful is there something missing in https://github.com/shogun-toolbox/shogun/blob/6fb8b6f613a6aada95cd9f53b588111ccff6279b/src/interfaces/swig/factory.i#L19

ghost: signature is different, i will check

theartful:
I added the integration testing. still need to figure out that bug

karlnapf:
> I added the integration testing. still need to figure out that bug

you need to update the `data` submodule to the commit in your data PR. Then the CI will execute the integration test here, and we can have it checked before merging both (github knows about the `data` commit even if the other PR is not yet merged)

examples/meta/src/binary/svmocas.sg
```diff
@@ -0,0 +1,23 @@
...
+Labels labels_train = labels(f_labels_train)
+#![create_features]
+
+#![create_classifier]
+Machine svm = machine("SVMOcas", features=feats_train, labels=labels_train, use_bias=False, C1=0.9, C2=0.9, epsilon=0.00001)
+#![create_classifier]
+
+#![train_and_apply]
+svm.train()
+Labels labels_predict = svm.apply(feats_test)
```
karlnapf: could you extract the label vector in both examples, then this will also be added to the integration test, which is useful

theartful:
I tried this multiple times with different tweaks, and I never get the same output as the one in CI. I will freeze this until I figure out the problem, and continue with different examples.

karlnapf:
we can first add this one without an integration test in that case?

theartful:
ok. svmocas works fine, so I will remove only svmlight integration test

karlnapf:
update and merge? :)

theartful:
sure will do that, and the problem with `string_features` should be fixed anyways once viktor's `shared_ptr` PR gets merged

ghost:
This is ready. CI failed for unrelated reason.

karlnapf:
Great, thanks for porting those :)

PR #4692, opened 2019-06-17
RandomMixin

theartful:
I tried a bunch of stuff, and I think this is neat. Here are the changes:
1. Add the option for callback functions which is called after "put". The current implementation is simple: if there is a parameter called "seed", the callback function will be "seed_callback".
2. Add "put_quietly" which will not invoke the callback function.
edit: removed. not needed anymore.
2. Add the option to iterate over all the parameter of an object filtered by type. For example
```C++
sg_object->for_each_param_of_type<CSGObject*>(
	[&](const std::string& name, CSGObject** param) {
		SG_DEBUG(name);
	});
```
3. Random will be a mixin added by inheritance, and the inheriting class has to be of type SGObject, and will pass itself to the constructor of Random so that it can do its magic.
4. Random will add a recursive callback function so that when put is called it iterates over all SGObject parameters, check if it has seed and puts it, then calls itself but on the child SGObject.
```C++
std::function<void(CSGObject*, uint32_t&)> callback;
callback = [&](CSGObject* object, uint32_t& m_seed) {
	object->for_each_param_of_type<CSGObject*>(
	    [&](const std::string& name, CSGObject** param) {
		    if ((*param)->has("seed"))
			    (*param)->put_quietly("seed", m_seed);

		    callback(param, m_seed);
	    });
};
```


src/shogun/base/SGObject.cpp
```diff
@@ -986,6 +988,23 @@ bool CSGObject::equals(const CSGObject* other) const
 	return true;
 }
 
+template <typename T>
+void CSGObject::for_each_param_of_type(
+	std::function<void(const std::string&, T*)> operation)
+{
+	std::unique_ptr<FilterVisitor<T>> visitor(new FilterVisitor<T>(operation));
```
vigsterkr: std::make_unique

vigsterkr:
great start but there is a problem with multi-class inheritance in swig. my initial idea was to have a RandomMixin, something like this (not everything works here afaik):

```
	template <class T, class PRNGType = std::mt19937_64>
	class RandomMixin: public T
	{
	/** parameter key for the seed */
	static const char* const kSeedKey;
	static Tag<Maybe<int32_t>> kSeedTag;
	public:
		/** PRNG type */
		typedef PRNGType prng_type;

		/** Default constructor */
		RandomMixin() : T()
		{
			Maybe<int32_t> seed = Nothing();
			T::register_param(kSeedTag, seed);
			init_prng();
		}

		~RandomMixin() override = default;

		RandomMixin(int32_t seed) : T()
		{
			T::register_param(kSeedTag, Just(seed));
			//m_seed = Just(seed);
			//T::register_parameter(kSeedTag, m_seed);
			//T::register_param(kSeedTag, Just(seed));
			init_prng();
		}


		/**
		 * Set seed for PRNG.
		 * Sets the seed for this object's PRNG as well as all the objects that
		 * this object has in it's paramter map.
		 *
		 * @param seed the seed for the prng.
		 */
		void set_seed(const int32_t seed)
		{
			auto new_seed = Just(seed);
			T::update_parameter(kSeedTag, make_any(new_seed));
			m_prng.reset(new PRNGType(seed));

			/*
			Maybe<int32_t> new_seed = Just(seed);
			update_parameter(kSeedTag, make_any(new_seed));
			self->prng.reset(new PRNGType(seed));

			// traverse the tree
			auto children = get_children();
			if (children.size() == 0)
				return;

			std::deque<CSGObject*> to_visit(children.begin(), children.end());
			do
			{
				auto c = to_visit.back();
				to_visit.pop_back();
				c->update_parameter(kSeedTag, make_any(new_seed));
				for (auto cc: c->get_children())
				{
					if (children.find(cc) == children.end())
					{
						to_visit.push_front(cc);
						children.emplace(cc);
					}
				}
			} while (!to_visit.empty());
			*/
		}

	protected:
		void init_prng()
		{
			if (!m_prng)
			{
				auto seed = T::get(kSeedTag);
				if (seed)
				{
					m_prng = std::make_shared<PRNGType>(seed.value());
				}
				else
				{
					int32_t random_data[PRNGType::state_size];
					std::random_device source;
					std::generate(std::begin(random_data), std::end(random_data), std::ref(source));
					std::seed_seq seeds(std::begin(random_data), std::end(random_data));
					// save seed to the obj
					m_prng = std::make_shared<PRNGType>(seeds);
				}
			}
		}

		std::shared_ptr<PRNGType> m_prng;
	};
	template <class T, class PRNGType>
	Tag<Maybe<int32_t>> RandomMixin<T, PRNGType>::kSeedTag = Tag<Maybe<int32_t>>(kSeedKey);

	template <class T, class PRNGType>
		const char* const RandomMixin<T, PRNGType>::kSeedKey = "seed";


}
```

src/shogun/base/SGObject.h
```diff
@@ -448,6 +450,9 @@ class CSGObject
 	void put(const std::string& name, T* value)
 	{
 		put(Tag<T*>(name), value);
+		std::string callback_fun = name + "_callback";
+		if(has(callback_fun))
```
karlnapf: one thing I am skeptical about is that this is a bit intrusive for the existing ::put methods.
Also, would there be another thing using these callbacks? Because if not this seems like a lot of architecture for a simple thing like recursively applying seeds.

I have the feeling maybe we should add an `init_seed()` method in SGObject, which does the recursive seed setting (if the parameter exists). Then everything would be in one place at least and it would be a NOP for non-random classes (could even warn/error in that case). This would not change "put" at all, but on the other hand would pollute CSGObject. Although that could be solved with a function to init the seed, e.g. `init_seed(my_object)`, which would behave as described above.

There are obviously pro/con arguments for each approach

src/shogun/mathematics/Random.cpp
```diff
@@ -120,7 +121,36 @@ void CRandom::init()
...
+	obj->watch_param("seed", &m_seed);
+
+	std::function<void(CSGObject*, uint32_t&)> callback;
+	callback = [&](CSGObject* object, uint32_t& seed) {
+		object->for_each_param_of_type<CSGObject*>(
+		    [&](const std::string& name, CSGObject** param) {
+			    if ((*param)->has("seed"))
+				{
+				    (*param)->put_quietly("seed", seed);
+				    (*param)->run("reinit");
```
karlnapf: why can't we call the method directly here?

src/shogun/mathematics/Random.cpp
```diff
@@ -120,7 +121,36 @@ void CRandom::init()
 	m_sfmt_32 = SG_MALLOC(sfmt_t, 1);
 	m_sfmt_64 = SG_MALLOC(sfmt_t, 1);
 	m_dsfmt = SG_MALLOC(dsfmt_t, 1);
-	reinit(m_seed);
+	reinit();
+
```
karlnapf: I wonder: do we actually need the object as parameter in all this? If Random is a mixin, it should have access, no?

src/shogun/mathematics/linalg/ratapprox/tracesampler/ProbingSampler.cpp
```diff
@@ -197,7 +197,7 @@ SGVector<float64_t> CProbingSampler::sample(index_t idx) const
 	{
 		if (m_coloring_vector[i]==idx)
 		{
-			float64_t x=sg_rand->std_normal_distrib();
+			float64_t x=std_normal_distrib();
```
karlnapf: I wonder whether `m_rand->std_random()` wouldnt be cleaner? I.e. make clear that this is a method from the random.....? Then the mixin would add that member instead of all the methods ... but not sure whether this is better than what you did here.....

src/shogun/mathematics/linalg/ratapprox/tracesampler/TraceSampler.h
```diff
@@ -17,12 +18,12 @@ template<class T> class SGVector;
...
  * the trace of a linear operator using an abstract sample method
  */
-class CTraceSampler : public CSGObject
+class CTraceSampler : public CSGObject, protected CRandom
 {
 public:
 	/** default constructor */
 	CTraceSampler()
-	: CSGObject()
+	: CSGObject(), CRandom(this)
```
karlnapf: Could we somehow create something that doesnt rely on this ctor (like calling default ctor)? Likely to be forgotten :D
I am not sure it is possible though

src/shogun/mathematics/Random.cpp
```diff
@@ -330,10 +360,9 @@ float64_t CRandom::GaussianPdfDenormInv(float64_t y) const
 	return std::sqrt(-2.0 * std::log(y));
 }
 
-void CRandom::reinit(uint32_t seed)
+void CRandom::reinit()
```
karlnapf: yep, this is good

src/shogun/io/visitors/FilterVisitor.h
```diff
@@ -0,0 +1,174 @@
+/*
+ * This software is distributed under BSD 3-clause license (see LICENSE file).
+ *
+ * Authors: Viktor Gal, Giovanni De Toni
+ */
+
+#ifndef SHOGUN_FILTERVISITOR_H
+#define SHOGUN_FILTERVISITOR_H
```
karlnapf: not sure what this is about ....

karlnapf: forgot to rebase?

src/shogun/mathematics/Random.cpp
theartful: SGObject doesn't have it, we're sure it's there since "seed" exists, which implies there is a Random mixin.

src/shogun/evaluation/TimeSeriesSplitting.cpp
```diff
@@ -84,7 +83,7 @@ void CTimeSeriesSplitting::build_subsets()
 		SG_UNREF(current);
 	}
 
-	m_subset_indices->shuffle(m_rng);
+	m_subset_indices->shuffle(this);
```
karlnapf: the array should have its own rng with this new design imo.
But this is minor  ...

src/shogun/base/init.cpp
```diff
@@ -77,7 +77,6 @@ namespace shogun
 		SG_REF(sg_io);
 		SG_REF(sg_parallel);
 		SG_REF(sg_version);
-		SG_REF(sg_rand);
```
karlnapf: nice to see this removed!

src/shogun/mathematics/linalg/ratapprox/tracesampler/ProbingSampler.cpp
theartful: yup it's actually better imo. don't know why i didn't do that :")

src/shogun/mathematics/Random.cpp
karlnapf: but you know that the object implements the random mixin or?

src/shogun/mathematics/linalg/ratapprox/tracesampler/TraceSampler.h
theartful: using a mixin that has an CRandom object (as you suggested) will solve this.

src/shogun/io/visitors/FilterVisitor.h
theartful: I need a way to iterate over all parameters of certain type, so this is to help with that.

src/shogun/mathematics/Random.cpp
theartful: yes

src/shogun/base/SGObject.h
theartful: I thought this is less ad-hoc, since we could benefit from callbacks in other classes, and also to unify the way parameters are set. not sure though

src/shogun/mathematics/Random.cpp
theartful: we know by implication: "seed" exists, since it must implement the Random mixin. maybe there is a better way.

vigsterkr:
As i’ve mentioned plz dont use CRandom obj... that should b dropped...

src/shogun/mathematics/linalg/ratapprox/tracesampler/ProbingSampler.cpp
vigsterkr: nope ... i disagree with it. rather follow PRNG concept requirements for prng in standard, namely have a class for each distribution and overload `operator()` that way we'll be able to use those in STL containers and algos...

theartful: @vigsterkr explaned this to me. the random mixin will have the sole job of hosting a prng object, and providing a functionality for setting seed, so CRandom class is not needed.

theartful:
@vigsterkr Please check now if that what you mean. I wrote the RandomMixin, and wrote NormalDistribution based on the one in CRandom. I played with them locally, and they seem to work.

src/shogun/mathematics/NormalDistribution.cpp
```diff
@@ -0,0 +1,97 @@
...
+
+using namespace shogun;
+
+NormalDistribution::NormalDistribution(float64_t mean, float64_t stddev)
+{
+	m_mean = mean;
+	m_stddev = stddev;
+
+	/** init ziggurat variables */
+	m_blockCount = 128;
```
vigsterkr: all of these could be constexpr no?

src/shogun/mathematics/NormalDistribution.h
```diff
@@ -0,0 +1,152 @@
+#ifndef __NORMALDISTRIBUTION_H__
+#define __NORMALDISTRIBUTION_H__
+
+#include <shogun/lib/Lock.h>
+#include <shogun/lib/common.h>
+#include <shogun/lib/config.h>
+
+namespace shogun
+{
+	class NormalDistribution
```
vigsterkr: template by type 
```
template<typename T>
class NormalDistribution { 
NormalDistribution(T mean = 0, T stddev = 1);
...
};
```

src/shogun/mathematics/RandomMixin.h
```diff
@@ -0,0 +1,74 @@
...
+		}
+
+		void set_seed(const int32_t seed)
+		{
+			m_seed = seed;
+			reinit();
+			set_seed_callback(this, seed);
+		}
+
+		void reinit()
```
vigsterkr: use-case?

src/shogun/mathematics/NormalDistribution.h
theartful: I'm not sure whether the algorithm would work with other types. I will check it.

src/shogun/mathematics/RandomMixin.h
theartful: when seed is set, we create a new PRNG object with the new seed, since the PRNG concept doesn't provide any other way.

vigsterkr: what i meant is: wouldn't having `set_seed` be enough? that reinits the PRNG. do we need a separate reinit (that is for the same seed)?

theartful: yes we need it. in the callback function we tell the object to `reinit` once seed is changed via `put`. there is a bug here, I forgot to register `reinit`

src/shogun/mathematics/NormalDistribution.cpp
```diff
@@ -0,0 +1,97 @@
...
+	m_R = 3.442619855899;
+	m_A = 9.91256303526217e-3;
+	m_uint32ToU = 1.0 / (float64_t)std::numeric_limits<uint32_t>::max();
+
+	m_x = SG_MALLOC(float64_t, m_blockCount + 1);
```
vigsterkr: these could be `std::array`?

src/shogun/mathematics/NormalDistribution.cpp
```diff
@@ -0,0 +1,97 @@
+#include <shogun/io/SGIO.h>
+#include <shogun/mathematics/NormalDistribution.h>
+
+#include <cmath>
+#include <limits>
+
+using namespace shogun;
+
+NormalDistribution::NormalDistribution(float64_t mean, float64_t stddev)
```
vigsterkr: the ctor could be constexpr, or?

src/shogun/mathematics/NormalDistribution.h
vigsterkr: in that case let this be the specialization for T=float64_t

src/shogun/mathematics/NormalDistribution.cpp
theartful: it should be possible

src/shogun/mathematics/NormalDistribution.cpp
theartful: constexpr std::array is very hard to initialize

src/shogun/mathematics/NormalDistribution.cpp
theartful: nope, not possible. constexpr constructors must be default/deleted, or only contain typedefs

src/shogun/mathematics/RandomMixin.h
theartful: I got rid of `reinit` and `put_quietly`

src/shogun/base/SGObject.h
```diff
@@ -448,6 +448,10 @@ class CSGObject
 	void put(const std::string& name, T* value)
 	{
 		put(Tag<T*>(name), value);
+
+		std::string callback_fun = name + "_callback";
+		if (has(callback_fun))
```
karlnapf: @lisitsyn @vigsterkr What are your thoughts on this callback stuff? I am not sure I like it.. One alternative is to add `CSGObject::init_seed` which is NOP if there is no seed parameter, but I am not sure I like that either

vigsterkr:
@theartful the callback function should be identified by an AnyProperty instead of a fixed string... if you have questions how to do it, ask it from @gf712 or me

theartful:
@vigsterkr I added CALLBACKFUNCTION property. Now `run` function doesn't work with a callback function. If this is ok, I will start writing tests.

src/shogun/mathematics/RandomMixin.h
```diff
@@ -0,0 +1,64 @@
...
+
+	template <typename Parent, typename PRNG = std::mt19937_64>
+	class RandomMixin : public Parent
+	{
+	public:
+		template <typename... T>
+		RandomMixin(T... args) : Parent(args...)
+		{
+			init();
+			m_seed = 0;
```
vigsterkr: this is wrong... this should use as in case of the example random_device... so that the PRNG is actually really RNG...

theartful: An earlier implementation had a random_device but I removed it. I will bring it back.

vigsterkr: and actually the problem is that we should support array of seeds as well :DDD
because the PRNG to be really a RNG should be init-ed like this (with an array):
```
std::random_device source;
std::generate(std::begin(random_data), std::end(random_data), std::ref(source));
std::seed_seq seeds(std::begin(random_data), std::end(random_data));
PRNG(seeds);
```

vigsterkr: but for the time being that we figure out stuff, let's just use a single int32_t... but that should come from random_device

vigsterkr: ok good, basically if seed is not set it should be random

src/shogun/base/SGObject.h
```diff
@@ -448,6 +448,10 @@ class CSGObject
 	void put(const std::string& name, T* value)
 	{
 		put(Tag<T*>(name), value);
+
+		std::string callback_fun = name + "_callback";
```
vigsterkr: @gf712 do you think we could somehow define in AnyParameters that a given parameter has dependent parameters? :D

src/shogun/mathematics/RandomMixin.h
```diff
@@ -28,16 +29,17 @@ namespace shogun
...
-				seed_callback(this, m_seed);
-				return true;
-			});
+			Parent::template watch_method<bool>(
+			    "seed_callback",
+			    [&]() {
+				    m_prng = PRNG(m_seed);
+				    seed_callback(this, m_seed);
+				    return true;
+			    },
```
gf712: You can try and make this void instead, I think Any supports void functions now. Does this compile though? It's a stateful lambda that is then converted to a function pointer..?

src/shogun/base/SGObject.h
gf712: hmmm I am not sure. We could add a vector of reference wrappers to other parameters? But why do you need it?

src/shogun/base/SGObject.cpp
```diff
@@ -986,6 +988,27 @@ bool CSGObject::equals(const CSGObject* other) const
...
 
+template <typename T>
+void CSGObject::for_each_param_of_type(
+    std::function<void(const std::string&, T*)> operation)
+{
+	auto visitor = std::make_unique<FilterVisitor<T>>(operation);
+	const auto& param_map = self->map;
+
+	std::for_each(param_map.begin(), param_map.end(), [&](auto& pair) {
+		Any any_param = pair.second.get_value();
```
gf712: Any& ?

src/shogun/util/visitors/FilterVisitor.h
```diff
@@ -0,0 +1,168 @@
...
+					return;
+
+			if constexpr (std::is_same<T, U>::value)
+				m_operation(m_name, v);
+
+			if constexpr (std::is_base_of<
+			                  typename std::remove_pointer<U>::type,
+			                  typename std::remove_pointer<T>::type>::value)
+			{
+				auto obj = dynamic_cast<T>(*v);
```
gf712: isn't it T*?

theartful: T* would be SGObject** or Kernel**. I'm not sure we can cast to a pointer pointer.

src/shogun/base/SGObject.cpp
theartful: Any& doesn't work
get_value returns Any

tests/unit/mathematics/RandomMixin_unittest.cc
```diff
@@ -0,0 +1,176 @@
...
+#include <shogun/base/SGObject.h>
+#include <shogun/mathematics/RandomMixin.h>
+#include <gtest/gtest.h>
+
+#include <memory>
+#include <array>
+
+using namespace shogun;
+
+class MockObject : public CSGObject
```
vigsterkr: there's already a mockobject within unit tests fw... would be great if we would have one if possible in the repo :)

tests/unit/mathematics/RandomMixin_unittest.cc
```diff
@@ -0,0 +1,176 @@
...
+};
+
+class MockNonRandom : public MockObject
+{
+};
+
+class MockRandom : public RandomMixin<MockObject>
+{
+public:
+	std::mt19937_64& m_prng = RandomMixin<MockObject>::m_prng;
```
vigsterkr: not sure if i get this one... could you elaborate?

theartful: m_prng is protected in RandomMixin, so I'm making it public

vigsterkr: this one i get, but why... i mean a unit test should just test publicly exposed API no need to test internals.... as those changes quite frequently, the important thing that it acts the same way after internal refactor... that way you dont need to rewrite tests

theartful: I get what you mean. I did this to test reproduciblity when seed is changed, since m_prng is never part of the public interface, however it's part of what defines a random mixin.
would something like this be better?
```
class MockRandom : public RandomMixin<CMockObject>
{
public:
	auto sample() -> decltype(m_prng())
	{
		return m_prng();
	}
};
```

src/shogun/util/visitors/FilterVisitor.h
gf712: ah ok, sorry!

src/shogun/mathematics/SFMT.cpp
```diff
@@ -0,0 +1,158 @@
+#include <shogun/lib/external/SFMT/SFMT.h>
+#include <shogun/lib/external/dSFMT/dSFMT.h>
+#include <shogun/mathematics/SFMT.h>
+
+using namespace shogun;
+
+template <typename T>
+SFMT<T>::SFMT(uint32_t seed) : m_seed(seed)
```
vigsterkr: why??? we dont want this... no need for this lib

theartful: I want to get rid of CRandom
porting this well allow us to use the SIMD instructions it provides, and get rid of CRandom until there is an alternative

vigsterkr: no you want to get rid of CRandom and *SFMT... meaning leave them there as they are, dont add more stuff to it. but start porting stuff that is using CRandom and once all CRandom use-cases are ported we can drop the whole thing

vigsterkr: and i thought we've agreed that even though *SFMT is good, it is unmaintained and we are certainly not going to maintain it... so we wanna get rid of it, even if it gives you SIMD random.

theartful: I misunderstood you on that point. I thought we couldn't get rid of it yet because of the SIMD stuff.

src/shogun/mathematics/RandomNamespace.h
```diff
@@ -0,0 +1,119 @@
...
+		 *
+		 * @param min_value minimum value
+		 * @param max_value maximum value
+		 * @return random number
+		 */
+		template <
+		    typename PRNG, typename U,
+		    typename std::enable_if_t<
+		        std::is_same<U, typename PRNG::result_type>::value>* = nullptr>
+		static inline U random(U min_value, U max_value, PRNG& prng)
```
theartful: random::random looks weird, should I change its name?

karlnapf: maybe call the namespace rng? shorter / less typing. I like "std::" :)
But just an opinion, random::random is fine as well

src/shogun/base/SGObject.cpp
```diff
@@ -80,6 +81,11 @@ namespace shogun
 			map.at(tag).set_value(value);
 		}
 
+		AnyParameter& at(const BaseTag& tag)
```
karlnapf: might also need a const version of this?

theartful: sure, might be useful & for completion sake

src/shogun/mathematics/RandomNamespace.h
theartful: rng is better imo

gf712: rng will conflict with the ranges namespace alias in c++20 :D

karlnapf: :D

src/shogun/base/DynArray.h
```diff
@@ -441,22 +441,24 @@ template <class T> class DynArray
...
-		/** randomizes the array (not thread safe!) */
-		void shuffle()
+		T* begin()
 		{
-			for (index_t i=0; i<=current_num_elements-1; ++i)
-				std::swap(
-					array[i], 
-					array[CMath::random(i, current_num_elements - 1)]);
+			return array;
+		}
```
theartful: I added iterators to ease the shuffling of DynArrays and DynamicObjectArrays. Not sure if this is ok

src/shogun/mathematics/RandomNamespace.h
```diff
@@ -22,9 +26,8 @@ namespace shogun
 		 *
 		 * @param container the container holding the elements
 		 */
-		template <
-		    template <typename> class Container, typename T, typename PRNG>
-		static inline void shuffle(Container<T>& container, PRNG& prng)
+		template <typename Container, typename PRNG>
+		static inline void shuffle(Container& container, PRNG& prng)
```
vigsterkr: do we really need this? i mean `std::shuffle` is basically doing this for you... or you just wanna have a syntactic sugar?

src/shogun/mathematics/RandomNamespace.h
```diff
@@ -0,0 +1,122 @@
...
+{
+	namespace random
+	{
+		/** Reorders the elements in [first,  last) randomly
+		 *
+		 * @param first an iterator to the first element in the range
+		 * @param last an iterator to the last element in the range
+		 */
+		template <typename RandomIt, typename PRNG>
+		static inline void shuffle(RandomIt first, RandomIt last, PRNG& prng)
```
vigsterkr: why do we need this wrapper for std::shuffle?

src/shogun/mathematics/RandomNamespace.h
```diff
@@ -0,0 +1,122 @@
...
+		 * range [0, PRNG::max() - PRNG::min()] if
+		 * signed(PRNG::max() - PRNG::min()) > 0, otherwise
+		 * the range is [0, std::limits<I>::max()]
+		 *
+		 * @return the random positive signed integer
+		 */
+		template <
+		    typename PRNG, typename I = typename std::make_signed<
+		                       typename PRNG::result_type>::type>
+		static inline I random_pos(PRNG& prng)
```
vigsterkr: this should be done via a UniformDistribution...

src/shogun/mathematics/RandomNamespace.h
```diff
@@ -0,0 +1,122 @@
...
+			return (prng() - PRNG::min()) &
+			       ((typename PRNG::result_type(-1) << 1) >> 1);
+		}
+
+		/**
+		 * Get random
+		 * @return a float64_t random from [0,1] interval
+		 */
+		template <typename PRNG>
+		static inline float64_t random_close(PRNG& prng)
```
vigsterkr: why? i mean why not a UniformDistribution?

src/shogun/mathematics/RandomNamespace.h
```diff
@@ -0,0 +1,122 @@
...
+			       float64_t(PRNG::max() - PRNG::min());
+		}
+
+		/**
+		 * Get random
+		 *
+		 * @return a float64_t random from [0,1) interval
+		 */
+		template <typename PRNG>
+		static inline float64_t random_half_open(PRNG& prng)
```
vigsterkr: same as above

src/shogun/mathematics/RandomNamespace.h
vigsterkr: this should be a distribution...UniformDistribution

src/shogun/mathematics/RandomNamespace.h
```diff
@@ -0,0 +1,122 @@
...
+		 * [min_value, max_value] (closed interval!)
+		 *
+		 * @param min_value minimum value
+		 * @param max_value maximum value
+		 * @return random number
+		 */
+		template <
+		    typename PRNG, typename S,
+		    typename std::enable_if_t<std::is_signed<S>::value>* = nullptr>
+		static inline S random(S min_value, S max_value, PRNG& prng)
```
vigsterkr: same as above... UniformDistribution

src/shogun/mathematics/RandomNamespace.h
```diff
@@ -0,0 +1,122 @@
...
+		/** generate a floating point value in the range
+		 * [min_value, max_value] (closed interval!)
+		 *
+		 * @param min_value minimum value
+		 * @param max_value maximum value
+		 * @return random number
+		 */
+		template <typename PRNG>
+		static inline float64_t
+		random(float64_t min_value, float64_t max_value, PRNG& prng)
```
vigsterkr: UniformDistribution

src/shogun/mathematics/RandomNamespace.h
theartful: I had the same thought. The previous implementation was this way, so I benchmarked uniform distribution against it, and uniform distribution is 3 much slower on my machine. Probably because it produces better results.

src/shogun/mathematics/RandomNamespace.h
theartful: In the case this grows and checks for gpu backend. It's not really useful otherwise

src/shogun/mathematics/RandomNamespace.h
vigsterkr: i'm not so sure what you mean by being slower... which implementation? i mean it would be good that all this is actually factored out to class that gives you samples from uniform distribution in a given range... just like NormalDistribution

theartful: my bad, it's much slower only without O3 flag, otherwise they're comparable.

src/shogun/mathematics/UniformIntDistribution.h
```diff
@@ -0,0 +1,93 @@
...
+			const uint64_t required_range = max - min + 1;
+			const auto max_val = prng_range - (prng_range % required_range);
+
+			typename PRNG::result_type result;
+			do
+			{
+				result = prng() - PRNG::min();
+			} while (result >= max_val);
+
+			return (result % required_range) + m_min;
```
theartful: the normal modulo method has bias towards small numbers if prng range is not divisible by the required range (more numbers map to small numbers than the bigger ones). this fixes it by only accepting numbers falling into a new modified range that is at least half as big as the original range.
adapted from
https://stackoverflow.com/questions/10984974/why-do-people-say-there-is-modulo-bias-when-using-a-random-number-generator

karlnapf:
What’s the expected number of iterations of that loop?
I find it a bit concerning to have a while loop in a rng... ;) but I didn’t read the details yet

theartful:
In the worst case scenario, which should rarely happen, the expected number is 2.

karlnapf:
Maybe try to run it a few million times and count? Just to be safe....

src/shogun/mathematics/RandomNamespace.h
vigsterkr: i've just done a git grep for this function and i dont see it anywhere being used in the main source... can we just drop all these functions, especially if not used

src/shogun/mathematics/RandomNamespace.h
vigsterkr: drop it.

src/shogun/mathematics/NormalDistribution.h
```diff
@@ -0,0 +1,182 @@
...
+		 * @return
+		 */
+		template <typename PRNG>
+		float64_t sample_tail(PRNG& prng) const
+		{
+			float64_t x, y;
+			float64_t m_R_reciprocal = 1.0 / m_R;
+			do
+			{
+				x = -std::log(random::random_half_open(prng)) * m_R_reciprocal;
```
vigsterkr: this is not really efficient as you create a UniformDistr at every call within this loop... you could just simply create a UniformDistr outside of this loop and sample it within this loop

src/shogun/mathematics/NormalDistribution.h
```diff
@@ -0,0 +1,182 @@
...
+				y = -std::log(random::random_half_open(prng));
```
vigsterkr: same as above

src/shogun/mathematics/NormalDistribution.h
```diff
@@ -0,0 +1,182 @@
...
+
+				// Generated x is outside of the rectangle.
+				// Generate a random y coordinate and test if our (x,y) is
+				// within the distribution curve. This execution path is
+				// relatively slow/expensive (makes a call to Math.Exp()) but
+				// relatively rarely executed, although more often than the
+				// 'tail' path (above).
+				float64_t x = u2 * m_uint32ToU * m_x[i];
+				if (m_y[i - 1] +
+				        ((m_y[i] - m_y[i - 1]) * random::random_half_open(prng)) <
```
vigsterkr: again... here you could just directly sample a UniformDistr that is created outside of this loop

src/shogun/multiclass/ecoc/ECOCRandomDenseEncoder.cpp
```diff
@@ -67,7 +68,7 @@ SGMatrix<int32_t> CECOCRandomDenseEncoder::create_codebook(int32_t num_classes)
         {
             for (int32_t j=0; j < num_classes; ++j)
             {
-                float64_t randval = CMath::random(0.0, 1.0);
+                float64_t randval = random::random(0.0, 1.0, m_prng);
```
vigsterkr: this is my problem with these utility functions... again here you can just create the uniformdistr outside of this loop and sample it here..

src/shogun/multiclass/ecoc/ECOCRandomSparseEncoder.cpp
```diff
@@ -83,7 +82,7 @@ SGMatrix<int32_t> CECOCRandomSparseEncoder::create_codebook(int32_t num_classes)
             // assign the remaining positions
             for (int32_t j=2; j < num_classes; ++j)
             {
-                float64_t randval = CMath::random(0.0, 1.0);
+                float64_t randval = random::random(0.0, 1.0, m_prng);
```
vigsterkr: same problem with inefficiency...

src/shogun/neuralnets/DeepBeliefNetwork.cpp
```diff
@@ -353,7 +356,7 @@ void CDeepBeliefNetwork::reset_chain()
 	SGMatrix<float64_t> s = m_states[m_num_layers-2];
 
 	for (int32_t i=0; i<s.num_rows*s.num_cols; i++)
-		s[i] = CMath::random(0.0,1.0) > 0.5;
+		s[i] = random::random(0.0,1.0, m_prng) > 0.5;
```
vigsterkr: again same story

src/shogun/neuralnets/DeepBeliefNetwork.cpp
```diff
@@ -435,7 +438,7 @@ void CDeepBeliefNetwork::down_step(int32_t index, SGVector< float64_t > params,
 	{
 		int32_t len = m_layer_sizes->element(index)*m_batch_size;
 		for (int32_t i=0; i<len; i++)
-			result[i] = CMath::random(0.0,1.0) < result[i];
+			result[i] = random::random(0.0,1.0, m_prng) < result[i];
```
vigsterkr: again, inefficient

src/shogun/neuralnets/DeepBeliefNetwork.cpp
```diff
@@ -465,7 +468,7 @@ void CDeepBeliefNetwork::up_step(int32_t index, SGVector< float64_t > params,
 	if (sample_states && index>0)
 	{
 		for (int32_t i=0; i<len; i++)
-			result[i] = CMath::random(0.0,1.0) < result[i];
+			result[i] = random::random(0.0,1.0, m_prng) < result[i];
```
vigsterkr: yet again...

src/shogun/neuralnets/NeuralLayer.cpp
```diff
@@ -96,7 +97,7 @@ void CNeuralLayer::dropout_activations()
 		int32_t len = m_num_neurons*m_batch_size;
 		for (int32_t i=0; i<len; i++)
 		{
-			m_dropout_mask[i] = CMath::random(0.0,1.0) >= dropout_prop;
+			m_dropout_mask[i] = random::random(0.0,1.0,m_prng) >= dropout_prop;
```
vigsterkr: and again

src/shogun/neuralnets/NeuralNetwork.cpp
```diff
@@ -572,12 +573,12 @@ float64_t CNeuralNetwork::check_gradients(float64_t approx_epsilon, float64_t s)
 	SGMatrix<float64_t> y(get_num_outputs(),1);
 
 	for (int32_t i=0; i<x.num_rows; i++)
-		x[i] = CMath::random(0.0,1.0);
+		x[i] = random::random(0.0,1.0, m_prng);
```
vigsterkr: inefficient

src/shogun/neuralnets/NeuralNetwork.cpp
```diff
@@ -572,12 +573,12 @@ float64_t CNeuralNetwork::check_gradients(float64_t approx_epsilon, float64_t s)
...
 
 	// the outputs are set up in the form of a probability distribution (in case
 	// that is required by the output layer, i.e softmax)
 	for (int32_t i=0; i<y.num_rows; i++)
-		y[i] = CMath::random(0.0,1.0);
+		y[i] = random::random(0.0,1.0, m_prng);
```
vigsterkr: inefficient... and actually fill_array would be better...

theartful:
@karlnapf results depend on the range. It only makes a difference when the range reaches 2^63, otherwise the extra invocations is very small. Btw boost does something similar using buckets https://www.boost.org/doc/libs/1_65_0/boost/random/uniform_int_distribution.hpp (the last few lines in generate_uniform_int), and this implementation in llvm is pretty much the same https://reviews.llvm.org/D7745 (https://reviews.llvm.org/differential/changeset/?ref=146467&whitespace=ignore-most).
I'm okay with dropping it though

src/shogun/statistical_testing/kernelselection/internals/MaxCrossValidation.h
```diff
@@ -45,11 +45,12 @@ template <typename T> class SGVector;
 namespace internal
 {
 #ifndef DOXYGEN_SHOULD_SKIP_THIS
+template <typename PRNG>
 class MaxCrossValidation : public KernelSelection
```
theartful: Making this class templated for PRNG is ugly, but I don't know what else to do (it's not CSGObject, and it implements the KernelSelection interface, in a method of which it uses PRNG)

theartful: overall the internals of statistical_testing are uglier this way. they are not SGObjects, so can't use RandomMixin, and either the functions (if possible) if the class itself (as in this case) are templated to accept a reference of the PRNG

src/shogun/neuralnets/NeuralNetwork.cpp
theartful: @vigsterkr  no more random::random stuff.
All the code base is ported now, I will start porting unit tests, and will come back to the code base afterwards for revision

src/shogun/mathematics/RandomNamespace.h
vigsterkr: btw do you think it would be worth to set PRNG as a forwarding reference i.e. just in case of std::shuffle? `PRNG&&`

src/shogun/statistical_testing/internals/mmd/CrossValidationMMD.h
```diff
@@ -67,10 +67,11 @@ struct CrossValidationMMD : PermutationMMD
 		m_num_runs=DEFAULT_NUM_RUNS;
 		m_alpha=DEFAULT_ALPHA;
 
-		init();
+		init(seed);
```
vigsterkr: if seed = -1, then it's -1 and not a random seed, or?

vigsterkr:
@theartful really great stuff so far! just need to fix the integration test data and some minor stuff i've just mentioned worth considering?

src/shogun/mathematics/RandomNamespace.h
theartful: PRNG should always be an l-value, so std::shuffle wouldn't use move semantics, but would accept it as an lvalue reference

theartful: but using `PRNG&&` covers both cases (lvalue reference and forwarding reference), so it's better

src/shogun/statistical_testing/internals/mmd/CrossValidationMMD.h
theartful: seed is not passed anymore. it's always a prng object that is passed, and for seeding created object, I added a function RandomNamespace called 'seed'

src/shogun/base/SGObject.cpp
```diff
@@ -701,6 +712,19 @@ bool CSGObject::has_parameter(const BaseTag& _tag) const
...
+    const std::string& name, std::function<void()> function)
+{
+	REQUIRE(function, "Function object is not callable");
+	BaseTag tag(name);
+	REQUIRE(
+	    has_parameter(tag), "There is no parameter called \"%s\" in %s\n",
+	    tag.name().c_str(), get_name());
+
+	auto& param = self->at(tag);
+	param.add_callback_function(function);
```
gf712: shouldn't you move this to avoid copying the callback function? I think function pointer copies are quite expensive..

src/shogun/base/AnyParameter.h
```diff
@@ -151,6 +151,9 @@ namespace shogun
 		void set_value(const Any& value)
 		{
 			m_value = value;
+
+			for (auto& method : m_callback_functions)
+				method();
```
gf712: wouldn't this be better in `CSGObject::update_parameter`? I think we either put all those things in SGObject or put all in AnyParameter, mixing it starts getting confusing.. Putting it here might actually be better in terms of performance? @vigsterkr @karlnapf

src/shogun/base/SGObject.cpp
theartful: good point

src/shogun/base/AnyParameter.h
theartful: do you think doing it in `Self::update` would be a good place?

gf712: I am not sure actually, seems a bit heavy for something that just contains a map?

theartful: `CSGObject::update_parameter` uses `Self::update`. If we to perform callback functions there, we would get the AnyParameter ourselves, and not use `Self::update`. Is this what you mean?

gf712: yea, you'd need a getter to return a reference to the vector of callback functions, e.g. `self->map[_tag].get_callbacks()`. But I guess that makes it a bit ugly?

theartful: nope, I think this type of separation of concerns is better

src/shogun/base/AnyParameter.h
```diff
@@ -168,6 +168,16 @@ namespace shogun
 			return m_init_function;
 		}
 
+		void add_callback_function(std::function<void()>&& method)
+		{
+			m_callback_functions.push_back(std::move(method));
+		}
+
+		auto& get_callbacks()
```
gf712: is it worth returning `const auto&`? And make the function const?

src/shogun/base/SGObject.cpp
```diff
@@ -650,16 +661,20 @@ void CSGObject::create_parameter(
...
 void CSGObject::update_parameter(const BaseTag& _tag, const Any& value)
 {
-	if (!self->map[_tag].get_properties().has_property(
-	        ParameterProperties::READONLY))
+	auto& param = self->at(_tag);
+	if (!param.get_properties().has_property(ParameterProperties::READONLY))
+	{
 		self->update(_tag, value);
+		for (auto& method : param.get_callbacks())
+			method();
```
gf712: imo this is much nicer and easier to keep track of

src/shogun/lib/DynamicArray.h
```diff
@@ -537,11 +537,15 @@ template <class T> class CDynamicArray :public CSGObject
 			return *this;
 		}
 
-		/** shuffles the array (not thread safe!) */
-		inline void shuffle() { m_array.shuffle(); }
+		inline auto begin()
```
gf712: For this to work with swig you need to add `->decltype(m_array.end())` as trailing return type. Swig is C++11 compatible and just writing auto is C++14 (I think)

gf712: And if that doesn't work you can just explicitly write RandomIterator<T>, or even hide it from SWIG, since this shouldn't be available from the interface anyway

theartful: I looked up iterators at SGVector and found that they are removed from swig "#ifndef SWIG", so I did the same.

gf712: yup that makes sense!

src/interfaces/swig/Classifier.i
```diff
@@ -73,10 +73,22 @@
 
 /* Include Class Headers to make them visible from within the target language */
 %include <shogun/machine/Machine.h>
+
+/** Instantiate RandomMixin */
+%template(RandomMixinMachine) shogun::RandomMixin<shogun::CMachine, std::mt19937_64>;
+
```
theartful: these are extremely ugly, but I have to specialize every RandomMixin right after the definition of the base class, so that derived classes can recognize it

src/shogun/lib/tapkee/tapkee_shogun.cpp
```diff
@@ -5,10 +5,10 @@
...
 #include <shogun/lib/tapkee/tapkee_shogun.hpp>
 
-
-#define CUSTOM_UNIFORM_RANDOM_INDEX_FUNCTION shogun::CMath::random()
-#define CUSTOM_UNIFORM_RANDOM_FUNCTION shogun::CMath::random(static_cast<tapkee::ScalarType>(0),static_cast<tapkee::ScalarType>(1))
-#define CUSTOM_GAUSSIAN_RANDOM_FUNCTION shogun::CMath::normal_random(static_cast<tapkee::ScalarType>(0),static_cast<tapkee::ScalarType>(1))
+// FIXME: Random Refactor PR
+// #define CUSTOM_UNIFORM_RANDOM_INDEX_FUNCTION shogun::CMath::random()
+// #define CUSTOM_UNIFORM_RANDOM_FUNCTION shogun::CMath::random(static_cast<tapkee::ScalarType>(0),static_cast<tapkee::ScalarType>(1))
+// #define CUSTOM_GAUSSIAN_RANDOM_FUNCTION shogun::CMath::normal_random(static_cast<tapkee::ScalarType>(0),static_cast<tapkee::ScalarType>(1))
```
theartful: tapkee uses CMath random
luckily these are never relied on  in shogun code

src/shogun/statistical_testing/internals/DataFetcher.cpp
```diff
@@ -118,7 +118,8 @@ void DataFetcher::shuffle_features()
 			shuffle_subset=SGVector<index_t>(size);
 		}
 		std::iota(shuffle_subset.data(), shuffle_subset.data()+shuffle_subset.size(), 0);
-		CMath::permute(shuffle_subset);
+		// FIXME: Random Refactor PR
+		// CMath::permute(shuffle_subset);
```
theartful: DataFetcher is not an SGObject, so it can't get a prng object. And this function is virtual, so it can't be templated to accept a PRNG object. The solution I have in mind is to template DataFetcher based on the PRNG type, which I also did with `MaxCrossValidation` (same case)

tests/unit/statistical_testing/internals/CrossValidationMMD_unittest.cc
```diff
@@ -63,8 +64,14 @@ TEST(CrossValidationMMD, biased_full)
...
 	const auto stype=ST_BIASED_FULL;
 
+	std::mt19937_64 prng(seed);
+
 	auto gen_p=some<CMeanShiftDataGenerator>(0, dim, 0);
 	auto gen_q=some<CMeanShiftDataGenerator>(difference, dim, 0);
+	// FIXME: Random Refactor PR
+	// FIXME: CMeanShiftDataGenerator is of Feature type, which is immutable
+	gen_p->CSGObject::put("seed", seed);
+	gen_q->CSGObject::put("seed", seed);
```
theartful: CMeanShiftDataGenerator uses the RandomMixin, but it's immutable (derived from CFeatures), so can't use it's version of "put"

theartful:
yay, tests passed, It's nit-picking time! I will go through the changes and look for stuff that can be made better.
some notes:
1. there are classes that use RandomMixin just because they have member that use it (so that putting seed works). these include: MMD, MachineEvaluation, SoftMaxLiklihood
2. putting parameters after setting the seed has no effect on these parameters seed. so the seed should be put last (I have some ideas to fix this, if you think it's a problem, using callbacks).
3. any method that uses PRNG can't be const
4. If you create an SGObject that uses random within a class, you have to seed it. for example (in GMM.cpp):
```C++
if (m_components[0]->get_mean().vector==NULL)
{
	CKMeans* init_k_means=new CKMeans(int32_t(m_components.size()), new CEuclideanDistance());
	seed(init_k_means);
	init_k_means->train(dotdata);
	SGMatrix<float64_t> init_means=init_k_means->get_cluster_centers();
	alpha=alpha_init(init_means);
	SG_UNREF(init_k_means);
	max_likelihood(alpha, min_cov);
}
```
5. PRNG should not be used in the initialization/constructor of an object, otherwise you don't get deterministic behaviour, since seed is initialized by a random value. A similar problem was with `Autoencoder` which if given layers on initialization, it initialized their weights using their prng objects. My solution was to set the weights of these layers before passing them.

examples/meta/src/converter/independent_component_analysis_ff_sep.sg
```diff
@@ -1,7 +1,5 @@
 File f_feats = csv_file("../../data/ica_2_sources.dat")
 
-Math:init_random(1)
```
karlnapf: these ica algos dont use any rng?

examples/meta/src/gaussian_process/classifier.sg
```diff
@@ -18,6 +17,7 @@ ConstMean mean_function()
 
 #![create_instance]
 SoftMaxLikelihood gauss_likelihood()
+gauss_likelihood.put("seed", 1)
```
karlnapf: why would a likelihood have the need for a rng?

examples/meta/src/meta_api/calls.sg
```diff
@@ -1,6 +1,3 @@
-# static call
-Math:init_random(1)
-
 # global function call
```
karlnapf: there is a great side-effect of removing the only source of static calls in the examples: R interfaces did not work with static calls, but that was the only thing, so after this is merged, we can try activating R again :)

examples/undocumented/python/classifier_multiclassocas.py
```diff
@@ -4,7 +4,6 @@
 
 def classifier_multiclassocas (num_vec=10,num_class=3,distance=15,width=2.1,C=1,epsilon=1e-5,seed=1):
 	from shogun import MulticlassLabels
-	from shogun import Math_init_random
```
karlnapf: you actually don't need to replace the seed calls in the undocumented examples -- those were only important for integration tests (the old ones) but we dont have them anymore ... just remove the init_seed calls everywhere and all good

karlnapf: although no need to revert the added .put(seed) calls

src/interfaces/swig/Classifier.i
karlnapf: mmmh

imo, there is no need for this. We only want to expose the base class (`Machine`) anyways. So all the other classes are OK to be hidden from SWIG (just need to port all python-undocumented examples as they of course will fail).

src/interfaces/swig/Kernel.i
```diff
@@ -74,6 +74,9 @@ PROTOCOLS_CUSTOMKERNEL(CustomKernel, float32_t, "f\0", NPY_FLOAT32)
 
 %include <shogun/kernel/DotKernel.h>
 
+/** Instantiate RandomMixin */
```
karlnapf: yeah definitely remove those

vigsterkr:
for 3. you could simply define the PRNG member variable `mutuable`. imo it's totally valid to use this trick here.

i'm not so sure if i understand why do we need the `seed(CSGObject*)` method... why not just sgo.set_seed(...) if u wanna set the seed of the given model? the setting of PRNG to have a random seed should be by default during ctor of the object

src/shogun/base/AnyParameter.h
```diff
@@ -168,6 +168,16 @@ namespace shogun
 			return m_init_function;
 		}
 
+		void add_callback_function(std::function<void()>&& method)
```
karlnapf: could I suggest you carve out this callback stuff into another PR for improved review :)

src/shogun/base/SGObject.cpp
karlnapf: shouldn't it return a reference as well?

src/shogun/mathematics/RandomMixin.h
```diff
@@ -0,0 +1,154 @@
...
+		    std::enable_if_t<std::is_base_of<
+		        CSGObject, typename std::remove_pointer<T>::type>::value>* =
+		        nullptr>
+		inline void seed(T* object) const
+		{
+			random::seed(object, m_seed);
+		}
+
+		int32_t m_seed;
+		PRNG m_prng;
```
vigsterkr: `mutable PRNG m_prng`

src/shogun/mathematics/RandomNamespace.h
vigsterkr: ok so then we change it? :)

src/shogun/clustering/GMM.cpp
```diff
@@ -133,6 +136,7 @@ float64_t CGMM::train_em(float64_t min_cov, int32_t max_iter, float64_t min_chan
 	if (m_components[0]->get_mean().vector==NULL)
 	{
 		CKMeans* init_k_means=new CKMeans(int32_t(m_components.size()), new CEuclideanDistance());
+		seed(init_k_means);
```
karlnapf: what is this?

examples/undocumented/python/modelselection_grid_search_kernel.py
```diff
@@ -75,8 +75,7 @@ def create_param_tree():
 
 def modelselection_grid_search_kernel (num_subsets, num_vectors, dim_vectors):
 	# init seed for reproducability
-	Math.init_random(1)
-	random.seed(1);
+	random.seed(1)
```
vigsterkr: what's the purpose of this? :)

src/shogun/converter/ica/FastICA.cpp
```diff
@@ -112,11 +113,11 @@ void CFastICA::fit_dense(CDenseFeatures<float64_t>* features)
 	if (m_mixing_matrix.num_rows != m || m_mixing_matrix.num_cols != m)
 	{
 		m_mixing_matrix = SGMatrix<float64_t>(m,m);
-
+		NormalDistribution<float64_t> normal_dist;
```
karlnapf: will there be an array filling routine for normal rng? That's quite useful to have
@vigsterkr @theartful I didn't follow the discussions

examples/undocumented/python/modelselection_grid_search_kernel.py
theartful: Math.init_random doesn't exist any more, and I just removed the semicolon :"D

src/shogun/distributions/Gaussian.cpp
```diff
@@ -404,9 +405,10 @@ SGVector<float64_t> CGaussian::sample()
 	}
 
 	SGVector<float64_t> random_vec(m_mean.vlen);
+	NormalDistribution<float64_t> normal_dist;
```
karlnapf: @vigsterkr @theartful question: wouldn't it make sense to pass the rng in the ctor argument?

src/shogun/distributions/HMM.cpp
```diff
@@ -7,19 +7,21 @@
...
 #include <shogun/features/Alphabet.h>
+#include <shogun/mathematics/UniformRealDistribution.h>
 
 #include <stdlib.h>
 #include <stdio.h>
 #include <time.h>
 #include <ctype.h>
 
-#define VAL_MACRO log((default_value == 0) ? (CMath::random(MIN_RAND, MAX_RAND)) : default_value)
+#define VAL_MACRO log((default_value == 0) ? (uniform_real_dist(m_prng)) : default_value)
```
karlnapf: this doesnt seem to be the same?

src/shogun/mathematics/RandomMixin.h
```diff
@@ -0,0 +1,154 @@
...
+	namespace random
+	{
+		/** Seeds an SGObject using a random number generator as a seed source
+		 */
+		template <
+		    typename T, typename PRNG,
+		    std::enable_if_t<std::is_base_of<
+		        CSGObject, typename std::remove_pointer<T>::type>::value>* =
+		        nullptr>
+		static inline void seed(T* object, PRNG& prng)
```
vigsterkr: i'm really not so sure to get it why is this good.
if an SGO is a RandomMixin, then
 * the default ctored object should have a random seed (and prng seeded with that seed).
 * if the user wants to set a seed then call sgo.put("seed", 1)

src/shogun/mathematics/RandomMixin.h
```diff
@@ -0,0 +1,154 @@
...
+		}
+
+		/** Seeds an SGObject using a specific seed
+		 */
+		template <
+		    typename T,
+		    std::enable_if_t<std::is_base_of<
+		        CSGObject, typename std::remove_pointer<T>::type>::value>* =
+		        nullptr>
+		static inline void seed(T* object, int32_t seed)
```
vigsterkr: see above

theartful:
@vigsterkr I have two versions of the seed method (const overloading), one uses a random seed, and the other uses the same seed as the original seed as the object.

examples/undocumented/python/modelselection_grid_search_kernel.py
vigsterkr: yeah but i dont see the point of this at all to have in any context, regardless whether that is Math::init_random or random.seed()

vigsterkr:
@karlnapf lets not cut it. you can has the patch and review it offline https://github.com/shogun-toolbox/shogun/pull/4692.

src/shogun/distributions/Gaussian.cpp
vigsterkr: @karlnapf as mentioned earlier check standard...

src/shogun/converter/ica/FastICA.cpp
vigsterkr: @karlnapf it can be implemented later.

vigsterkr: but the interface we try to keep as std

tests/unit/statistical_testing/internals/CrossValidationMMD_unittest.cc
karlnapf: can you explain this to me? :)

karlnapf:
> for 3. you could simply define the PRNG member variable `mutable`. imo it's totally valid to use this trick here.
> 
> i'm not so sure if i understand why do we need the `seed(CSGObject*)` method... why not just sgo.set_seed(...) if u wanna set the seed of the given model? the setting of PRNG to have a random seed should be by default during ctor of the object

+1 for the `mutable` rng var

karlnapf:
> @karlnapf lets not cut it. you can has the patch and review it offline https://github.com/shogun-toolbox/shogun/pull/4692.patch

I am not a fan of these big PRs, especially the callback stuff is almost unrelated to the random stuff.... but I won't insist...

src/shogun/distributions/Gaussian.cpp
karlnapf: ok thx

src/shogun/converter/ica/FastICA.cpp
theartful: @karnlapf yeah, I already added it, just need to push :")

src/shogun/distributions/Gaussian.cpp
theartful: @karlnapf we're following the way the standard library does it. pass prng to operator()

examples/meta/src/gaussian_process/classifier.sg
theartful: @karlnapf it has a  `mc_sampler` method that creates a `GaussianDistribution` object that needs to be seeded

src/shogun/distributions/HMM.cpp
theartful: @karlnapf they are, but the limits (MIN_RAND, MAX_RAND) are set at construction site of uniform_real_dist

src/shogun/base/SGObject.cpp
theartful: I think that's reasonable, should I change its implementation?

karlnapf: i think so, but pls another pr so it is visible and we can discuss it with the others

examples/meta/src/converter/independent_component_analysis_ff_sep.sg
theartful: this variant doesn't need it

theartful:
PRNG is mutable now, and I changed back methods to be constant. This is more or less ready.

src/shogun/mathematics/RandomNamespace.h
```diff
@@ -0,0 +1,111 @@
...
+		 */
+		template <typename RandomIt, typename PRNG>
+		static inline void shuffle(RandomIt first, RandomIt last, PRNG&& prng)
+		{
+			using diff_t =
+			    typename std::iterator_traits<RandomIt>::difference_type;
+			UniformIntDistribution<diff_t> dist;
+			diff_t n = last - first;
+			for (diff_t i = n - 1; i > 0; --i)
+				std::swap(first[i], first[dist(prng, {0, i})]);
```
karlnapf: does this generate a uniform shuffling?

Simple test: Generate an array [1,2,3,4,5,6,7,8,9]. Then shuffle. Repeat this many many times and compute a histogram for each position (all digits should appear in each position equally often)

vigsterkr: mmmm rather check the UniformDistribution... but since that's checked, i reckon this is good to go

vigsterkr:
@theartful merge?

src/shogun/mathematics/RandomNamespace.h
karlnapf: too me, it is not obvious that if you use the uniform distribution and shuffle like here, the generated permutations are uniform...but maybe I am missing something

vigsterkr: https://en.cppreference.com/w/cpp/algorithm/random_shuffle

read the specs + (3) implementation.

karlnapf: Ah cool thx

karlnapf:
> @theartful merge?

I am ok as well, the only thing is I don't like the `%template` instantiations in swig, especially since they will be removed again soon when all examples are ported ....

PR #4695, opened 2019-06-22
Serializers does not need to inherit from CSGObject

vigsterkr:


lisitsyn:
Why not? We could reuse the parameters framework so that we can set options and flags etc..

vigsterkr:
i really dont see the point to add this monster base class for this tiny little thingy :)

karlnapf:
Maybe once we have a mixin for the parameters ...

PR #4697, opened 2019-06-27
implement std::hash<CSGObject>

vigsterkr:


src/shogun/util/hash.h
```diff
@@ -0,0 +1,37 @@
...
+	{
+		typedef T argument_type;
+		typedef std::size_t result_type;
+
+		result_type operator()(const argument_type& o) const
+		{
+			auto params = o.get_params();
+			result_type seed = 0;
+			for (const auto& p: params)
+				seed ^= p.second->get_value().hash() + 0x9e3779b9 + (seed<<6) + (seed>>2);
```
karlnapf: Are the numbers here arbitrary  or something standard?

src/shogun/base/SGObject.cpp
```diff
@@ -344,12 +339,7 @@ bool CSGObject::parameter_hash_changed()
 {
 	SG_DEBUG("entering\n")
 
-	uint32_t hash=0;
-	uint32_t carry=0;
-	uint32_t length=0;
-
-	get_parameter_incremental_hash(hash, carry, length);
-	hash=CHash::FinalizeIncrementalMurmurHash3(hash, carry, length);
+	auto hash = std::hash<CSGObject>{}(*this);
```
karlnapf: Could this method now be const?

src/shogun/util/hash.h
vigsterkr: https://www.boost.org/doc/libs/1_55_0/doc/html/hash/combine.html

src/shogun/base/SGObject.cpp
vigsterkr: yeah as soon as the problem with clang is solved...

src/shogun/util/hash.h
karlnapf: I see

vigsterkr:
it works but imo there's a problem with the current implementation... there should be a:
```
			else if constexpr (std::is_base_of_v<CSGObject, T>)
			{
				return std::hash<CSGObject>{}(value);
			}
```

as if we register a CKernel than std::is_hashable<CKernel>::value will be false, since std::hash<CKernel> is not defined... but as soon as i add the above code, there's the problem of instantiating `std::hash<CSGObject>` as that's not defined in any.h :) and i cannot even do it as CSGObject is forward declared...:S

gf712:
@vigsterkr what about this https://stackoverflow.com/a/36775973 ? This seems to be similar to what boost does so that the hash function itself is found with ADL

vigsterkr:
@gf712 the std::hash specialization still will be:
```
template<> hash<CSGObject>
...
```

hence it will not match anything of the derived classes :(

vigsterkr:
@lisitsyn @gf712 had to go with this... i.e. add SGObject::hash() const function... this way things should work... i wanted to avoid it, but idk whether this is more ugly or the macro set for all the possible registered types and use what @gf712 suggested...

lisitsyn:
No tengo ningun problema con ->hash() :)

PR #4698, opened 2019-06-29
Port some more examples

theartful:


examples/meta/src/evaluation/clustering.sg
```diff
@@ -0,0 +1,43 @@
...
+MulticlassLabels labels(labels_vector)
+Features fea_centroids = features(centers)
+Distance d2 = distance("EuclideanDistance", lhs=fea_centroids, rhs=fea_centroids)
+KNN knn(1, d2, labels)
+knn.train()
+MulticlassLabels gnd_hat = knn.apply_multiclass(features_train)
+#![assign_labels]
+
+#![measure_accuracy]
+ClusteringAccuracy accuracy_eval()
```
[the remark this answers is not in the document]
theartful: can't call `best_map` if I used base class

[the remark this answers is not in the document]
karlnapf: That’s what we have “wartch_method” for
Check k means base and example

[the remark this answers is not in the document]
theartful: @karlnapf do I add the arguments of the method as parameters "watch_parameter"?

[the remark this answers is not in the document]
karlnapf: These cases need refactoring as the `watch_method` functions do not accept arguments.

One option is to require that a caller uses `.put` to set the labels before calling the method (otherwise runtime error with instructions). The other option here would be to use the `.evaluate` method of `CEvaluation` as the signature of the base class is already appropriate here.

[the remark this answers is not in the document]
karlnapf: `ClusteringAccuracy` does not have a method `best_map` btw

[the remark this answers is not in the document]
karlnapf: ah I found it, see discussion in IRC:
```
(14:54:32) Heiko: 1.) the method shouldnt modify the given labels, but rather operate on a copy
(14:54:58) Heiko: 2.) there should be a flag that can disable invocation of the method in ::evaluate
(14:55:07) essam: that's reasonable
(14:55:43) Heiko: 3.) to ensure this is not copy-pasted across all specializations, implement this flag logic in CClusteringEvaluation, and then inside there call a purely virtual method evaluate_impl, which is then overloaded by the sub-classes
```

examples/meta/CMakeLists.txt
```diff
@@ -121,6 +121,7 @@ INSTALL(DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/cpp
 IF (NOT EXISTS ${CMAKE_CURRENT_BINARY_DIR}/data)
     # copy data symlink to cmake binary dir so that we can execute meta examples in there
     CREATE_DATA_SYMLINK(${CMAKE_CURRENT_SOURCE_DIR}/data ${CMAKE_CURRENT_BINARY_DIR}/data)
+    CREATE_DATA_SYMLINK(${CMAKE_SOURCE_DIR}/data ${CMAKE_CURRENT_BINARY_DIR}/full-data)
```
theartful: I either do this, or move uci optdigits dataset to toy data

karlnapf: Nope can’t do that

Toy means toy, so pls don’t put the full dataset but rather a small version of it, or better use an existing one (these are just api examples they’re supposed to run fast)

theartful: I tried some of the classification datasets, and the results were bad, so I added a smaller version of this dataset.

karlnapf: Just checked, clustering the 2d nonlinear classification example should work, as this is just a 2x2 gaussian bumps grid. If not, something is wrong :)

theartful: I get around 0.5 accuracy :"(

karlnapf: which algorithm on what data?

theartful: KMeans clustering on `classifier_binary_2d_nonlinear_features_train.dat`

examples/meta/src/evaluation/clustering.sg
```diff
@@ -0,0 +1,33 @@
...
+Labels labels_centroids = labels(f_labels_centroids)
+Features fea_centroids = features(centers)
+Distance d2 = distance("EuclideanDistance", lhs=fea_centroids, rhs=fea_centroids)
+Machine knn = machine("KNN", k=1, distance=d2, labels=labels_centroids)
+knn.train()
+Labels gnd_hat = knn.apply(features_train)
+#![assign_labels]
+
+#![measure_accuracy]
+Evaluation accuracy_eval = evaluation("ClusteringAccuracy")
```
karlnapf: Just remove this and store the cluster centres instead

theartful: Isn't this unsupervised learning? You cluster using KMeans, and used KNN around these clusters to predict the labels, then match your made up labels with the original. Sorry for persisting, but I think it makes sense, but if you don't think so though I will remove the evaluation code.

theartful: @karlnapf My bad :"D After writing the previous comment I realized that `k` should be equal to the number of labels, I will try it now.

theartful: it didn't make a difference, I will drop accuracy

src/shogun/evaluation/ClusteringEvaluation.cpp
```diff
@@ -33,17 +52,18 @@ int32_t CClusteringEvaluation::find_mismatch_count(SGVector<int32_t> l1, int32_t
...
 
-void CClusteringEvaluation::best_map(CLabels* predicted, CLabels* ground_truth)
+CLabels* CClusteringEvaluation::best_map(CLabels* predicted, CLabels* ground_truth)
 {
 	ASSERT(predicted->get_num_labels() == ground_truth->get_num_labels())
-	ASSERT(predicted->get_label_type() == LT_MULTICLASS)
-	ASSERT(ground_truth->get_label_type() == LT_MULTICLASS)
+	
+	auto predicted_multiclass = multiclass_labels(predicted);
+	auto ground_truth_multiclass = multiclass_labels(ground_truth);
```
theartful: not completely related, but it caused me problems when I tried the binary data `classifier_binary_2d_nonlinear_features_train.dat`

karlnapf: you needed to add this or this change didnt work?

theartful: it worked, to interpret binary labels as multiclass, but I don't need it now. thought I leave it anyways since it's better.

karlnapf: Pls remove if not needed then We’ll  have to refactor less later

karlnapf:
and otherwise I think this is ready to be merged or?

src/shogun/evaluation/ClusteringEvaluation.cpp
theartful: @karlnapf when you say `feats = features(something)` it interprets feats as BinaryLabels if possible. If it did, ClusteringEvaluation wouldn't work. This solves this since ClusteringEvaluation should work with both, and since the user can't/shouldn't specify that features should be interpreted as multiclass.

karlnapf: ideally, all those `multiclass_labels` calls would be removed and be replaced by assertions (since this system is faulty). As said I have something better in a PR but it is not ready yet. No need to spread the old stuff if it is not really necessary (which isnt the case here).

theartful:
I reverted those changes. it's ready now

karlnapf:
ok good to merge once CI is happy

PR #4699, opened 2019-06-30
Remove Features put overload

theartful:


karlnapf:
@gf712 could we have a helper method that applied the read i my prkoert to all parameters for cases like these?

gf712:
sure! we can add something that is called in the base class constructor that adds the readonly mask to everything that is added via SG_ADD

karlnapf:
@theartful I think this would be nicer as then we will never forget to mark parameters as readonly....Something like "mark_class_as_readonly" or so ....

gf712:
@theartful can you also add the m_default_mask to the copy constructor? I forgot it!

theartful:
@gf712 No, you're okay, the copy constructor also calls `init`.

tests/unit/ctags
```diff
@@ -1 +0,0 @@
-/usr/bin/ctags
```
karlnapf: what is this?

tests/unit/statistical_testing/KernelSelection_unittest.cc
```diff
@@ -245,8 +245,8 @@ TEST(KernelSelectionMaxTestPower, quadratic_time_single_kernel)
 
 	auto gen_p=new CMeanShiftDataGenerator(0, dim, 0);
 	auto gen_q=new CMeanShiftDataGenerator(difference, dim, 0);
-	gen_p->CSGObject::put("seed", seed);
-	gen_q->CSGObject::put("seed", seed);
+	gen_p->put("seed", seed);
```
karlnapf: not sure I get why this is part of the PR?

tests/unit/ctags
theartful: a mistake I made in the RandomMixin PR, thought I drop it now

tests/unit/statistical_testing/KernelSelection_unittest.cc
theartful: an artifact from the random mixin pr because of the overloaded function

theartful:
let's merge

PR #4701, opened 2019-06-30
Simplify value clone in Any

lisitsyn:


src/shogun/lib/any.h
```diff
@@ -401,22 +401,6 @@ namespace shogun {
...
-
-		struct general : by_default
-		{
-		};
-
-		struct more_important : general
-		{
-		};
-
-		struct maybe_most_important : more_important
```
lisitsyn: These guys are not needed anymore

tests/unit/lib/Any_unittest.cc
```diff
@@ -346,8 +346,8 @@ TEST(Any, equals_null_pointer)
 
 TEST(Any, equals_value)
 {
-	Simple a;
-	Simple b;
+	SimpleValue a;
```
lisitsyn: The test was wrong as it operates with `Simple` as a value.

PR #4573, opened 2019-03-14
Add tag based serialization

vigsterkr:
@lisitsyn have fun :)

PR #4702, opened 2019-07-01
option to set default parameter bitmask value

gf712:
@theartful @karlnapf this is how you could set the bitmask values for all parameters added via SG_ADD.
The issue is that there are also the watch_param methods... I think we really need the replacement from SG_ADD macro to a variadic method in SGObject, so we can add all this logic in one place.

### DIFF
--- a/src/shogun/base/SGObject.h
+++ b/src/shogun/base/SGObject.h
@@ -88,8 +88,10 @@ using stringToEnumMapType = std::unordered_map<std::string, std::unordered_map<s
 
 #define SG_ADD3(param, name, description)                                      \
 	{                                                                          \
+		auto pprop =                                                           \
+		    AnyParameterProperties(description, this->m_default_mask);         \
 		this->m_parameters->add(param, name, description);                     \
-		this->watch_param(name, param, AnyParameterProperties(description));   \
+		this->watch_param(name, param, pprop);                                 \
 	}
 
 #define SG_ADD4(param, name, description, param_properties)                    \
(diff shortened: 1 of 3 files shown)


karlnapf:
Since #4573 was merged, the only thing that is really missing now are the parameter gradients ... and then we can drop SG_ADD and `Parameter.cpp`

karlnapf:
Since #4573 was merged, the only thing that is really missing now are the parameter gradients ... and then we can drop SG_ADD and `Parameter.cpp`

src/shogun/base/SGObject.h
```diff
@@ -1026,6 +1033,20 @@ class CSGObject
...
+	 * ParameterProperties::NONE.
+	 * @param mask the default mask
+	 */
+	void set_default_mask(ParameterProperties mask) noexcept
+	{
+		m_default_mask = mask;
+	}
+
+	/** The default mask for all registered parameters */
+	ParameterProperties m_default_mask = ParameterProperties::NONE;
```
karlnapf: i guess we could also avoid this member via applying the default mask to all yet registered parameters....downside would be that we need to always call this after registering parameters ....

gf712: yes, and for example in Features you want the behaviour in all derived classes, and this enforces it. Otherwise each subclass has to call such a method after registering all the parameters..

karlnapf: ah yes, very good point with the subclasses!

karlnapf:
@theartful pls use this in your other PR

PR #4706, opened 2019-07-04
Split RandomMixin into prng and seed classes

theartful:
This has the artifact that if a parent class has the Seedable mixin, a derived class can't have RandomMixin (seed will be registered twice). It can be solved using SFINAE
```
RandomMixin : public Parent (if parent has Seedable)
RandomMixin : public Seedable<Parent> (if parent doesn't have Seedable)
```
but I'm not sure it's worth it, as it would complicate swig.

### DIFF
--- a/src/shogun/evaluation/CrossValidation.h
+++ b/src/shogun/evaluation/CrossValidation.h
@@ -13,7 +13,7 @@
 
 #include <shogun/evaluation/EvaluationResult.h>
 #include <shogun/evaluation/MachineEvaluation.h>
-#include <shogun/mathematics/RandomMixin.h>
+#include <shogun/mathematics/Seedable.h>
 
 namespace shogun
 {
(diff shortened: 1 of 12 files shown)


karlnapf:
Is anyone against this, otherwise I'll merge it for use in xval

iglesias:
+1.

karlnapf:
I actually have one request here: Could you change CrossValidation to use the "seedable" mixin....since then we have a use-case straight away :)

ghost:
@karlnapf done

karlnapf:
Nice thanks!

src/shogun/evaluation/CrossValidation.h
```diff
@@ -123,7 +123,7 @@ namespace shogun
 	 * HP Laboratories.] for details on this subject.
 	 *
 	 */
-	class CCrossValidation : public RandomMixin<CMachineEvaluation>
+	class CCrossValidation : public Seedable<CMachineEvaluation>
```
karlnapf, line 126: nice!

PR #4708, opened 2019-07-05
[WIP] Drop DynamicObjectArray

theartful:


src/shogun/distributions/MixtureModel.h
```diff
@@ -175,7 +175,7 @@ class MixtureModel : public Distribution
 
 	private:
 		/** array of components */
-		std::shared_ptr<DynamicObjectArray> m_components;
+		std::vector<std::shared_ptr<SGObject>> m_components;
```
vigsterkr: shouldn't this be simply `std::vector<std::shared_ptr<Distribution>>` ? this way you dont need to cast all the time....

ghost: I was thinking of the problem with `put` and `get` and casting when used this way

vigsterkr: mmm put should be fine no? i mean it's a base interface class...

vigsterkr: apparently not... but i guess it should be

karlnapf: yep, `CDistribution` please

src/shogun/distributions/EMMixtureModel.cpp
```diff
@@ -51,7 +51,7 @@ float64_t EMMixtureModel::expectation_step()
 		// for each component
 		for (int32_t j=0;j<data.alpha.num_cols;j++)
 		{
-			auto jth_component=data.components->get_element<Distribution>(j);
+			auto jth_component=data.components[j]->as<Distribution>();
```
karlnapf: is this really a good idea?

src/shogun/distributions/MixtureModel.h
ghost: Okay, and I will try restricting std::vector to the most derived (or at least a base class) in the following cases, and see if any problems come up.

karlnapf: For now, we are using base classes in those situations (parameter arrays, parameters). Put/get will only work then. So everything that is accessible to users should be that way.

There will be issues in CGMM as there all elements are CGaussian so we need to cast every time. A put that would downcast would be cool here but we dont have that, so for now you will have to use ->as in those cases (or come up with a better solution)

src/shogun/features/CombinedDotFeatures.cpp
```diff
@@ -243,42 +243,44 @@ void CombinedDotFeatures::free_feature_iterator(void* iterator)
...
-	return result;
+	return true;
 }
 
 bool CombinedDotFeatures::append_feature_obj(std::shared_ptr<DotFeatures> obj)
 {
 	ASSERT(obj)
 	int n = get_num_feature_obj();
-	feature_array->push_back(obj);
+	feature_array.push_back(obj);
```
gf712: in these situations it might be worth doing std::move. You copy the shared_ptr anyway because you pass by value, so just move that copy in the push_back, no?

ghost: yup

ghost:
DynamicArrays provide functionality for 2D and 3D arrays, without it things get very ugly with DynProg. Should I keep them?

ghost:
I don't know how to test this and I'm not sure which tests fail because of this PR :(

karlnapf:
I think 3D can be dropped. If something like this is needed maybe it could be another structure?

Btw Is dynprog used anywhere?

ghost:
There is SGNDArray, but it provides less functionality + no, DynProg isn't used anywhere

ghost:
DynamicObjectArray is used by the meta examples as a container for the output to be serialized.  Should we have a container for this or serialize each object separately?

karlnapf:
> There is SGNDArray, but it provides less functionality + no, DynProg isn't used anywhere

drop? :D

karlnapf:
> DynamicObjectArray is used by the meta examples as a container for the output to be serialized. Should we have a container for this or serialize each object separately?

good question! Might need either a wrapper class that stores the vector (so we have a single SGObject that contains all things), or a facility to serialize `std::vector<CSGObject>` directly. @vigsterkr is the new serialization master he will know :)

karlnapf:
> I don't know how to test this and I'm not sure which tests fail because of this PR :(

I think the best you can do is to test before/after and see what tests start failing .... though ultimately we will just need to make all tests work to be more or less confident that this all works

vigsterkr:
@theartful for the time being leave DynamicObjectArray we'll just mark the whole thing deprecated and will do this integration checking some other way....

ghost:
The tests that fail before this PR are the same as the one after. I'm not sure why clang and gcc fail at the setup step, but I don't think it's related.

PR #4711, opened 2019-07-06
fix memory error in liblinear svr

karlnapf:
produced wrong results due to memory error :(

### DIFF
--- a/src/shogun/regression/svr/LibLinearRegression.cpp
+++ b/src/shogun/regression/svr/LibLinearRegression.cpp
@@ -109,7 +109,10 @@ bool CLibLinearRegression::train_machine(CFeatures* data)
 	prob.l=num_vec;
 	prob.x=features;
 	auto labels = regression_labels(m_labels);
-	prob.y = labels->get_labels();
+
+	// store reference to vector locally in order to prevent free-ing
+	auto lab = labels->get_labels();
+	prob.y = lab.data();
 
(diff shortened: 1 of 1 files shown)


src/shogun/regression/svr/LibLinearRegression.cpp
```diff
@@ -109,7 +109,10 @@ bool CLibLinearRegression::train_machine(CFeatures* data)
 	prob.l=num_vec;
 	prob.x=features;
 	auto labels = regression_labels(m_labels);
-	prob.y = labels->get_labels();
```
karlnapf, line 112: subset is set -> this returns a new SGVector -> only one refeference -> assigning to `prob.y` -> deleting the SGVector -> `kaboom`.
This didn't work with subsets/ x validation before

ghost: nice!

karlnapf: We should get rid of such methods, wrong name, behaviour (new vs existing SGVector) changes depending on context, etc.
Either we should have `const SGVector& get_labels` (which would throw a runtime error if a subset is active so inappropriate here) or `SGVector labels_copy()` which always returns newly allocated memory, or even better we access the elements via the new iterator (although in this case liblinear needs a pointer to the label vector )

PR #4712, opened 2019-07-09
Systematic framework tests for all machines

karlnapf:
## testing
* [x] use sgobject iterator to instantiate and test all machines from a given list
* [x] individual ignore lists for each test (runtime)
* [ ] automate header and machine name extraction from `trained_model_serialization.cc.py`
* [ ] separate data generating code into environment
* [ ] clean up

## tests
* [x] consistency of views/subset with subsampled(no-view, same data)
* [x] training/apply leaves data unchanged
* [x] training thread consistency
* [x] test cross-validation thread consistency
* [x] import trained_model_serialization tests
* [ ] delete old trained_model_serialization tests
* [x] training without initializing does throw an exception (and doesnt crash)
* [ ] document non-working machines in ignore lists


tests/unit/machine/all_machines_unittest.cc
```diff
@@ -0,0 +1,304 @@
...
+		machine_subsampled->train(features_subsampled);
+
+		auto result_subset = machine_subset->apply(data.first);
+		auto result_subsampled = machine_subsampled->apply(data.first);
+
+		EXPECT_TRUE(result_subset->equals(result_subsampled));
+	}
+}
+
+TEST(AllMachines, cv_thread_consistency)
```
karlnapf, line 304: @theartful these are the kind of tests I was talking about.

tests/unit/evaluation/CrossValidation_unittest.cc
```diff
@@ -1,206 +0,0 @@
-/*
```
karlnapf, line 1: these are subsumed by the new ones

tests/unit/machine/all_machines_unittest.cc
ghost: these are very nice!

ghost: do you think separating the data generating code to the public interface would be a good idea? examples would benefit from them for instance

karlnapf: yes I want to move all that into an environment similar to those used by `trained_model_serialization` so the data can be re-used for other instances. Will add that to the list, thanks

tests/unit/machine/all_machines_unittest.cc
```diff
@@ -0,0 +1,386 @@
...
+		init_machine(machine);
+		auto data = generate_data(machine);
+		machine->set_labels(data.second);
+		machine->train(data.first);
+	}
+}
+
+TEST(AllMachines, train_thread_consistency)
+{
+	std::set<std::string> ignores = {
```
vigsterkr, line 217: copy-paste all around the source... plz avoid redundancy

tests/unit/machine/all_machines_unittest.cc
```diff
@@ -0,0 +1,386 @@
...
+		init_machine(machine);
+		auto data = generate_data(machine);
+		auto cv = generate_cv(machine, data);
+		auto cv2 = make_clone(cv);
+
+		get_global_parallel()->set_num_threads(1);
+		cv->put("seed", 1);
+		auto result_single = cv->evaluate();
+
+		get_global_parallel()->set_num_threads(4);
```
vigsterkr, line 323: this is a by-chance checking of the consistency... depends on many aspects. for example on the machine that the code is being run on etc.

tests/unit/machine/all_machines_unittest.cc
```diff
@@ -0,0 +1,386 @@
...
+			};
+	for (auto obj : sg_object_iterator<untemplated_sgobject>(all_machines).ignore(ignores))
+	{
+		auto machine = obj->as<CMachine>();
+		SCOPED_TRACE(machine->get_name());
+
+		init_machine(machine);
+		auto data = generate_data(machine);
+
+		machine->set_labels(data.second);
```
vigsterkr, line 371: the triplet: set_labels, train, apply sprayed around this source again... while it could be a function that returns a tuple that you nicely tie and get it back

vigsterkr (changes requested):
redundant and inconsistent

tests/unit/machine/all_machines_unittest.cc
karlnapf: the ignores are test-specific.

tests/unit/machine/all_machines_unittest.cc
karlnapf: true, but this already revealed bugs. open for ways to improve

vigsterkr: that's one thing that you have a test locally to have this tested... its another thing that have this formalised in a test that is part of the codebase and the CI that is going to be run on many different arch/distro etc...

tests/unit/machine/all_machines_unittest.cc
vigsterkr: yeah 3x times the very same ignore

tests/unit/machine/all_machines_unittest.cc
```diff
@@ -0,0 +1,384 @@
...
+	deserializer->attach(fis);
+	auto deser_obj = deserializer->read_object();
+	bool delete_success = !fs->delete_file(fname);
+	EXPECT_TRUE(delete_success);
+
+	return dynamic_cast<CMachine*>(deser_obj.get());
+}
+
+// TODO, generate this automatically, like in trained_model_serialization
+std::set<string> all_machines = {"LibSVM", "Perceptron", "LibLinear",
```
vigsterkr, line 184: at least make a function that returns this set instead of having it here a global var....

tests/unit/machine/all_machines_unittest.cc
vigsterkr: for further info: https://en.cppreference.com/w/cpp/language/structured_binding

vigsterkr: but in this case you would actually just return the predictions... so you dont even need that... just pass the rvalue of the tuple (labels, data)

tests/unit/machine/all_machines_unittest.cc
vigsterkr: for example:
 * do not use global_parallel - that is going to be dropped eventually - the sooner the better
 * why 4, why not 8, or 3 or 11 or 42? you could actually use a function to get the runtime env supported number of max threads and use that number, and if it's 1 then say that foobar.
....

tests/unit/machine/all_machines_unittest.cc
karlnapf: this is because the number of tested machines is small and not yet automated. Although there are already differences here, see random forest. For more machines tested, there will be more diversity. If not, i can of course drop this.

tests/unit/machine/all_machines_unittest.cc
karlnapf: the number is of course arbitrary, good idea to use the available number of threads, i will do that

tests/unit/machine/all_machines_unittest.cc
vigsterkr: since you copy paste 3 times the value... you could create a value of it instead of copy-pasting the same string around. 1 copy-paste is one too many imo

karlnapf: sure, i will look into this once the coverage is bigger

tests/unit/machine/all_machines_unittest.cc
```diff
@@ -0,0 +1,384 @@
...
+		SCOPED_TRACE(machine->get_name());
+
+		init_machine(machine);
+		auto machine2 = make_clone(machine);
+
+		auto data = generate_data(machine);
+
+		get_global_parallel()->set_num_threads(1);
+		machine->set_labels(data.second);
+		if (machine->has("seed"))
```
vigsterkr, line 232: there's actually a function for this: seed(machine, 1);

tests/unit/machine/all_machines_unittest.cc
```diff
@@ -0,0 +1,384 @@
...
+			machine->put("seed", 1);
+		machine->train(data.first);
+		auto result_single = machine->apply(data.first);
+
+		init_machine(machine);
+		get_global_parallel()->set_num_threads(4);
```
vigsterkr, line 238: total random number....

tests/unit/machine/all_machines_unittest.cc
```diff
@@ -0,0 +1,384 @@
...
+		machine2->set_labels(data.second);
+		if (machine2->has("seed"))
+			machine2->put("seed", 1);
```
vigsterkr, line 241: i guess the seed should be the same as above, right? in that case again just copy-pasting the same number around... refactoring is a pain.. use a variable.

tests/unit/machine/all_machines_unittest.cc
karlnapf: will do once this extraction is automated

tests/unit/machine/all_machines_unittest.cc
karlnapf: yep

tests/unit/machine/all_machines_unittest.cc
```diff
@@ -0,0 +1,384 @@
...
+		machine2->train(data.first);
```
vigsterkr, line 242: same as below: set_labels, train, apply. could you use a function for this?
the seeding part could be fixed by having that function a 4th arg that has a default value of some special number that means no seeding...

tests/unit/machine/all_machines_unittest.cc
```diff
@@ -0,0 +1,384 @@
...
+
+		auto machine_subset = make_clone(machine);
+		auto machine_subsampled = make_clone(machine);
+		if (machine->has("seed"))
+		{
+			machine_subset->put("seed", 1);
+			machine_subsampled->put("seed", 1);
+		}
+
+		machine_subset->set_labels(labels_subset);
```
vigsterkr, line 292: set_labels, train, apply... all over again

PR #4713, opened 2019-07-10
Support casting in Any

lisitsyn:


gf712:
for the class hierarchy could you generate lots of template specialisations with ctags? and then recursively get the type from the overloaded structs?
```cpp
template <typename T>
struct get_hierarchy {};

template <>
struct get_hierarchy<GaussianKernel>
{
  using types = Types<Kernel, SGObject>;
}
...
```

lisitsyn:
@gf712 nooo not in this PR please 

:)

lisitsyn:
I believe this can't break anything thus merging

gf712:
@lisitsyn Yea, just wondering what you thought about something like that

lisitsyn:
@gf712 looks like a good solution indeed

`std::bases` did not happen :( Sad!

src/shogun/lib/any.h
```diff
@@ -1112,18 +1137,56 @@ namespace shogun {
...
+		static CastingRegistry casting_registry;
 	};
 
 	bool operator==(const Any& lhs, const Any& rhs);
 
 	bool operator!=(const Any& lhs, const Any& rhs);
 
+	template <typename T>
+	inline void register_casts() {
+		if constexpr (std::is_base_of<CSGObject, typename std::remove_pointer<T>::type>::value)
```
karlnapf, line 1178: Could this be done for all base classes automatically?

karlnapf (commented):
Nice! :)

src/shogun/lib/any.h
karlnapf: As we have them in is_sg_base already anyways

gf712:
Yea, doesn’t seem like it’ll be around until C++23...

vigsterkr:
for the base classes i think there's one nice crazy fold expression based solution for this.... i'll try later today

PR #4714, opened 2019-07-10
Support arbitrary visitors

lisitsyn:


tests/unit/lib/Any_unittest.cc
```diff
@@ -836,3 +836,31 @@ TEST(Any, cast_into_sgobject)
...
+	};
+	Any::register_visitor<int, ExtractVisitor>([&] (int value, ExtractVisitor* visitor) { extracted_value = value; });
+	auto any = make_any<int>(42);
+	any.visit<ExtractVisitor>();
+	EXPECT_EQ(any.as<int>(), extracted_value);
+}
+
+TEST(Any, stateful_visit)
+{
+	std::string value = "hello, world!";
```
iglesias: const?

tests/unit/lib/Any_unittest.cc
```diff
@@ -836,3 +836,31 @@ TEST(Any, cast_into_sgobject)
...
+	struct StringStreamVisitor
+	{
+		std::stringstream ss;
+	};
+	Any::register_visitor<std::string, StringStreamVisitor>([&] (std::string value, StringStreamVisitor* visitor) {
```
iglesias: const std::string&?

tests/unit/lib/Any_unittest.cc
```diff
@@ -836,3 +836,31 @@ TEST(Any, cast_into_sgobject)
...
+		(visitor->ss) << value;
+	});
+	auto any = make_any(value);
+	StringStreamVisitor visitor;
+	any.visit(&visitor);
```
iglesias: What about a couple of calls to visit to really show the stateful bit?
Perhaps I am misinterpreting the state reference.

tests/unit/lib/Any_unittest.cc
```diff
@@ -836,3 +836,31 @@ TEST(Any, cast_into_sgobject)
...
 	EXPECT_THROW(any.cast<int>(), std::logic_error);
 }
+
+TEST(Any, simple_visit)
+{
+	int extracted_value{};
+	struct ExtractVisitor
+	{
+	};
+	Any::register_visitor<int, ExtractVisitor>([&] (int value, ExtractVisitor* visitor) { extracted_value = value; });
```
iglesias: Just wild thining here. From api pov, I am wondering whether the template arguments could be somehow omitted and extracted from the std::function.

iglesias (approved):
¡Qué bueno!

I left just mostly nitpicking comments 😅

tests/unit/lib/Any_unittest.cc
lisitsyn: Si

tests/unit/lib/Any_unittest.cc
lisitsyn: Meh doesn't matter :)

tests/unit/lib/Any_unittest.cc
lisitsyn: Replaced by auto

tests/unit/lib/Any_unittest.cc
lisitsyn: Well it is already stateful but I can visit twice to make it more visible

src/shogun/lib/any.h
```diff
@@ -1164,26 +1170,67 @@ namespace shogun {
...
+	void Any::register_caster(std::function<To(From)> caster)
+	{
+		const auto key = std::make_pair(std::type_index{typeid(From)}, std::type_index{typeid(To)});
+
+		if (casting_registry.count(key))
+		{
+			return;
+		}
+
+		casting_registry[key] = [caster] (void* src, void* dst)
```
gf712, line 1197: you already pass the function pointer by value, you might as well std::move here no?

src/shogun/lib/any.h
```diff
@@ -1164,26 +1170,67 @@ namespace shogun {
...
+		{
+			auto typed_src = static_cast<From*>(src);
+			auto typed_dst = static_cast<To*>(dst);
```
gf712, line 1200: out of curiosity how come you switched from reinterpret_cast to static_cast?

lisitsyn: We don't actually need reinterpret so I used more narrow one.

src/shogun/lib/any.h
lisitsyn: You mean caster&&?

karlnapf:
:dancer:

src/shogun/lib/any.h
gf712: Yup, [caster=std::move(caster)]

PR #4715, opened 2019-07-11
Move global variables to a ShogunEnv static class

ghost:


examples/meta/generator/targets/cpp.json
```diff
@@ -119,7 +119,7 @@
             "get_option": "$object->get<std::string>($arguments)",
             "get_string": "$object->get<std::string>($arguments)"
         },
-        "StaticCall": "C$typeName::$method($arguments)",
+        "StaticCall": "$typeName::$method($arguments)",
```
ghost: C prefix will be dropped anyways, and there are no static calls now except for ShogunEnv, so I saw no harm

src/shogun/io/SGIO.h
```diff
@@ -544,6 +546,12 @@ class SGIO
 
 	private:
 		RefCount* m_refcount;
+		/// function called to print normal messages
+		std::function<void(FILE*, const char*)> print_message;
+		/// function called to print warning messages
+		std::function<void(FILE*, const char*)> print_warning;
+		/// function called to print error messages
+		std::function<void(FILE*, const char*)> print_error;
```
ghost, line 554: it made more sense that these should be here instead of ShogunEnv

src/shogun/lib/Signal.h
```diff
@@ -30,7 +30,7 @@ namespace shogun
 	 *  option bewteen: immediately exit the running method and fall back to
 	 *  the command line, prematurely stop the current algoritmh and do nothing.
 	 */
-	class CSignal : public CSGObject
+	class CSignal
```
ghost, line 33: doesn't need to inherit from SGObject + it caused problems (ShogunEnv tries to create CSignal, and CSignal tries to create ShogunEnv)

vigsterkr:
i'm a bit ambivalent to make all these functions static functions of the class.... as far as i remember we had problems with one interface implementation calling static functions (maybe R?).

anyhow, the FileSystemRegistry should be as well part of this

src/shogun/base/ShogunEnv.h
```diff
@@ -0,0 +1,131 @@
...
+	class Parallel;
+	class SGLinalg;
+	class CSignal;
+
+	class ShogunEnv
+	{
+	public:
+		/** Remove copy and move constructors and assignment operators
+		 */
+		ShogunEnv(const ShogunEnv&) = delete;
```
vigsterkr: `SG_DELETE_COPY_AND_ASSIGN` from `shogun/base/macros.h`

src/shogun/base/ShogunEnv.h
```diff
@@ -0,0 +1,131 @@
...
+
+		/** Checks environment variables and modifies global objects
+		 */
+		void init_from_env();
+
+		Parallel* sg_parallel;
+		SGIO* sg_io;
+		Version* sg_version;
+		CSignal* sg_signal;
+		SGLinalg* sg_linalg;
```
vigsterkr: why not `std::unique_ptr`?

src/shogun/base/ShogunEnv.h
```diff
@@ -0,0 +1,131 @@
...
+		ShogunEnv();
+
+		/** Checks environment variables and modifies global objects
+		 */
+		void init_from_env();
+
+		Parallel* sg_parallel;
+		SGIO* sg_io;
+		Version* sg_version;
+		CSignal* sg_signal;
```
vigsterkr: why not `std::unique_ptr`?

karlnapf:
Yes R and static calls doesn’t work. And it’s a Pitty because since the random stuff is sorted, we don’t need static calls in the examples anymore.....so we could actually enable the R examples now

examples/meta/generator/targets/cpp.json
karlnapf: I think This should rather be done when the prefix is dropped, not yet

ghost:
The other solution I see is to make only one static function to get the environment, and the other non-static. But this has the same problem with R, and I think it will be more verbose in code (get environment then get what you want).

examples/meta/generator/targets/cpp.json
theartful: sure, but CShogunEnv is ugly though :"D

ghost:
Does this class need to be exposed at all?

vigsterkr:
imo you dont need to expose this class in the interfaces... or whatever that needs to be used by this class you can wrap those within another class that is exposed to the interface and then call via that... anyhow no need for such things... 
and just to be clear: this class does not need the C prefix as this class is a) not inherited of SGO b) no need to expose it to class_list...

ghost:
This is an attempt using multiple inheritance, please check if you find this approach acceptable @vigsterkr

vigsterkr:
looking nice! :) need to fix some legacy code or?

ghost:
I think this ready. It's worth reminding this it can't be wrapped by swig. I will rebase and push to solve the conflict.

vigsterkr:
@theartful great! good job! as we agree we never wanna expose the env directly to swig... so it should be fine.... and if one day we need to, then we know what to do :)

PR #4716, opened 2019-07-12
add base type matcher and register caster

vigsterkr:


vigsterkr:
this actually could be simplified more... will do that, just that in the meanwhile i have to take care of some edge cases.... but the idea is this, and it should work in theory.

lisitsyn (approved):
Godspeed!

vigsterkr:
ok this is the simplified version + it filters the cases when the registered object is not a base interface class... like CListElement and other objects like that in the lib

src/shogun/base/base_types.h
```diff
@@ -95,6 +96,39 @@ namespace shogun
 	{
 	};
 
+	template <typename...>
+	struct type_list{};
+
+	template <typename T>
+	struct type_holder
```
vigsterkr, line 103: anybody has any better way to do this... @gf712 @lisitsyn

src/shogun/lib/any.h
```diff
@@ -1222,9 +1223,12 @@ namespace shogun {
 
 	template <typename T>
 	inline void register_casts() {
-		if constexpr (std::is_base_of<CSGObject, typename std::remove_pointer<T>::type>::value)
+		using Derived = std::remove_pointer_t<T>;
+		if constexpr (std::is_base_of_v<CSGObject, Derived>)
 		{
 			Any::register_caster<T, CSGObject*>([] (T value) { return dynamic_cast<CSGObject*>(value); });
+			if constexpr (!std::is_same_v<nullptr_t, base_type<Derived>>)
```
vigsterkr: not the nicest way....

src/shogun/base/base_types.h
gf712: A while ago I wrote a type list struct if you want to reuse that? ~~I think also did recursive searches like the one below but with structs~~never mind it doesn’t do the is base of. this looks good to me

gf712: Btw would it be useful to have the whole inheritance tree? Or are we just doing type erasure with these classes?

src/shogun/lib/any.h
gf712: You can also write a is_any_of_derived to make it more explicit if you think this isn’t nice

src/shogun/base/base_types.h
gf712: I mean what if you have `CAccuracyMeasure` which is `CAccuracyMeasure -> CContingencyTableEvaluation -> CBinaryClassEvaluation -> CEvaluation -> CSGObject`?

vigsterkr: mmm well i would say that the base interface classes should not overlap, or? i mean they should have one common class (CSGO) but other than that we should not have a base interface that is subset of another base interface...

vigsterkr: and actually this base class 'extractor' was written with this in mind... of course in that case we actually a tuple should be returned of types and those should all be exposed ;)

gf712: yea, that could be a pain. would probably have to generate the whole thing with ctags, i was testing that a bit. And then the map increases in size.
I am just thinking that I would expect the casting to work for all the types, like it works with static and dynamic cast.
But I can't really think of a use case for that tbh..

gf712: usually what we do is get the "base"  type, like CMachine, and use the common interface. The rest is more algorithm specific and would probably be bad practice to use?

karlnapf: We have overlapping base classes atm. But in the cases I know, this is actually only due to the inability to cast any, so no need for this anymore. Eg svm&machine

vigsterkr: that actually depends... as this is the method that allows the parameter that is being registered as CSVM to be cast-able to get it as CMachine... but CSVM is exposed in the first place because of putting an svm into mkl, or? coz in that case unfortunately we need to still expose CSVM... and then this method has a problem :)

vigsterkr: mmm and actually the same is true for pipeline :))) ok i'll look into things.... but yeah this actually atm has a random behaviour in some edge cases :S

src/shogun/lib/any.h
```diff
@@ -1222,9 +1223,12 @@ namespace shogun {
 
 	template <typename T>
 	inline void register_casts() {
-		if constexpr (std::is_base_of<CSGObject, typename std::remove_pointer<T>::type>::value)
+		using Derived = std::remove_pointer_t<T>;
+		if constexpr (std::is_base_of_v<CSGObject, Derived>)
 		{
 			Any::register_caster<T, CSGObject*>([] (T value) { return dynamic_cast<CSGObject*>(value); });
+			if constexpr (!std::is_same_v<std::nullptr_t, base_type<Derived>>)
+				Any::register_caster<T, base_type<Derived>*>([] (T value) { return dynamic_cast<base_type<Derived>*>(value); });
```
vigsterkr, line 1231: doh... this fails with msvc 19.16, but not with 19.20 or after:
https://godbolt.org/z/8DjP0R

vigsterkr: if i change the definition of base_type to
```
using base_type = typename std::decay_t<decltype(find_base<Derived>(sg_interface{}))>;
```

and then do `dynamic_cast<base_type<Derived>::type *>(...);` it works.... anybody any ideas? :D

vigsterkr: ```using base_type = typename decltype(find_base<Derived>(sg_interface{}))::type;```
solved the problem... lemme know if you feel that that could cause some troubles :S

src/shogun/base/base_types.h
vigsterkr: actually i've dropped from sg_interface CSVM and CPipeline as in this particular case i didn't see the point having them in that set....

karlnapf: I think if we can cast there is no need to have them part of the interface, so all good :)

PR #4719, opened 2019-07-15
fix windows dll creation

vigsterkr:
fix windows conda environment activation


vigsterkr:
somebody help me with windows and dll path so that when running tests it can find the `shogun.dll`. basically this is the last blocker for this fix....

gf712:
but that doesn't seem to be the issue? There are 3 tests that pass and at least 2 need the dll to run as they call the factory no?

gf712:
if you think it an issue with pathing can you not add the dll path to the system PATH and see what happens?

vigsterkr:
@gf712 i've had reproduced the env on a local windows machine and whenever i try to call a unittest or example it cries about `shogun.dll` not being on path. then if i add `#pragma comment(lib, 'shogun.dll'` then it cries about not finding `mkl_rt.dll`. and yes i've added the path where shogun.dll resides and the path for mkl_rt.dll should be as well in (see conda env settings in the yaml file of the job). 🤷‍♂

vigsterkr:
blocked by https://github.com/gabime/spdlog/issues/973

vigsterkr:
closing in favour for the development ongoing in feature/modular-graph branch

vigsterkr:
ok so there's the fold here:
 1) the relative data path is not working since msvc has an additional directory for placing the binaries (`Debug`, `Release`)
 2) the unit tests are failing yet for an unknown reason: managed to isolate it that as soon as i add a test that uses `DataGenerator` the error appears....but couldnt find out why...

vigsterkr:
azure tests... something is up with the CI hooks

vigsterkr:
arghhh... didn't work ...

vigsterkr:
mmmm so seems once closed and in the meanwhile force-pushed, the pr cannot be reopened 🤷‍♂

PR #4722, opened 2019-07-16
Use to_string when appropriate

lisitsyn:


tests/unit/lib/Any_unittest.cc
```diff
@@ -865,3 +865,16 @@ TEST(Any, stateful_visit)
 	any.visit_with(&visitor);
 	EXPECT_EQ(visitor.ss.str(), initial_value + initial_value);
 }
+
+TEST(Any, to_string_using_std)
+{
+	EXPECT_EQ(make_any<int>(0).cast<std::string>(), "0");
```
vigsterkr: for future reference, expect_eq and any other gtest assertation method's semantics is the following: `<expected value>, <actual value>`

vigsterkr (approved):
great!

tests/unit/lib/Any_unittest.cc
lisitsyn: Thanks, reordered them

PR #4729, opened 2019-07-17
Use spdlog for logging

ghost:
spdlog is a logging library that uses "fmt" for formatting (which is another great library that might get added to the standard!).
TODO:
- [x] Add spdlog to CMake
- [ ] Discuss what to do next!

Ideas:
1. Allow users to change 'sinks' (using spdlog terminology), so that loggers can write to different targets without exposing spdlog to them (?).
2. Enable each algorithm to have its own spd logger (=> SGIO object). Might require us to do something like RandomMixin so that it's children would write to the same logger.
3. Get rid of macros (would probably mean getting rid of line number and file info).


src/shogun/io/SGIO.cpp
```diff
@@ -41,19 +42,12 @@ const char* SGIO::message_strings_highlighted[NUM_LOG_LEVELS]={"[GCDEBUG] \0", "
...
-
 SGIO::SGIO()
     : target(stdout), show_progress(false), location_info(MSG_NONE),
       syntax_highlight(true), loglevel(MSG_WARN)
 {
 	m_refcount = new RefCount();
-	print_message = &print_default;
-	print_warning = &print_default;
-	print_error = &print_default;
+	logger = spdlog::stdout_color_mt("console");
```
vigsterkr: mmm maybe i would opt for the async logger....?

vigsterkr (commented):
i guess you'll need to wrap around this: https://github.com/gabime/spdlog/wiki/4.-Sinks#adding-sinks-to-the-logger-after-creation

to be able to add user defined sinks runtime...?

src/shogun/io/SGIO.h
```diff
@@ -42,13 +51,10 @@ enum EMessageType
...
-	MSG_WARN=4,
-	MSG_ERROR=5,
-	MSG_CRITICAL=6,
-	MSG_ALERT=7,
-	MSG_EMERGENCY=8,
-	MSG_MESSAGEONLY=9
+	MSG_WARN=3,
+	MSG_ERROR=4,
+	MSG_CRITICAL=5,
+	MSG_MESSAGEONLY=6
```
ghost: only 7 levels instead of 10

src/shogun/io/SGIO.h
```diff
@@ -220,6 +171,19 @@ struct substring
...
+	constexpr SourceLocation(
+		const char* file_="", int32_t line_=0, const char* function_="")
+	: file(file_), line(line_), function(function_)
+	{
+	}
+
+	const char* file;
+	int32_t line;
+	const char* function;
+};
```
ghost: could've used spdlog::source_loc, but I chose not to include spdlog

src/shogun/io/SGIO.h
```diff
@@ -328,89 +300,45 @@ class SGIO
...
-
-		/** set target to stderr */
-		inline void set_target_to_stderr() { set_target(stderr); }
+		/** redirects stdout to another sink */
+		void redirect_stdout(std::shared_ptr<spdlog::sinks::sink> sink);
 
-		/** set target to stdout */
-		inline void set_target_to_stdout() { set_target(stdout); }
+		/** redirects stderr to another sink */
+		void redirect_stderr(std::shared_ptr<spdlog::sinks::sink> sink);
```
ghost: there are only two sinks: one for stdout and the other for stderr. if the user wants to redirect to multiple sinks, he could use "dist_sink"

src/shogun/io/SGIO.h
```diff
@@ -493,40 +415,16 @@ class SGIO
...
-		 * @return message intro or NULL if message is not to be
-		 *         printed
-		 */
-		const char* get_msg_intro(EMessageType prio) const;
+	private:
+		/** Prints a formatted message */
+		void message_(EMessageType prio, const SourceLocation& loc, const fmt::string_view& msg) const;
+
+		/** Updates log pattern */
+		void update_pattern();
```
ghost: pattern is `[date time {source_location} loglevel] msg`
example:
`[07/27/19 23:39:33 debug] Comparing parameter DynamicObjectArray::dim1_size of type int.`

src/shogun/io/SGIO.h
```diff
@@ -535,39 +433,29 @@ class SGIO
...
-		/// function called to print normal messages
-		std::function<void(FILE*, const char*)> print_message;
-		/// function called to print warning messages
-		std::function<void(FILE*, const char*)> print_warning;
-		/// function called to print error messages
-		std::function<void(FILE*, const char*)> print_error;
+		class RedirectSink;
+		std::shared_ptr<RedirectSink> io_sink;
+		std::shared_ptr<spdlog::logger> io_logger;
+		std::shared_ptr<spdlog::details::thread_pool> thread_pool;
```
ghost: loggers hold `weak_ptr` of the thread_pool, so have to hold it in the class

src/shogun/io/SGIO.h
```diff
@@ -535,39 +433,29 @@ class SGIO
...
-void SGIO::message(EMessageType prio, Args&&... args) const
+void SGIO::message(EMessageType prio, const SourceLocation& loc, const char* fmt, const Args&... args) const
 {
-	const auto& msg = format(prio, std::forward<Args>(args)...);
-	print(prio, msg);
+	if(should_log(prio))
+	{
+		fmt::memory_buffer msg;
+		fmt::format_to(msg, fmt, args...);
+		message_(prio, loc, fmt::string_view(msg.data(), msg.size()));
```
ghost: do the formatting in the header, and the logging in the cpp file. slightly inefficient because `should_log` is called twice + the extra indirection of the function call

src/shogun/io/SGIO.cpp
```diff
@@ -29,228 +37,188 @@
...
-	"[NOTICE] \0", "[WARN] \0", "[ERROR] \0",
-	"[CRITICAL] \0", "[ALERT] \0", "[EMERGENCY] \0", "\0"};
-
-const char* SGIO::message_strings_highlighted[NUM_LOG_LEVELS]={"[GCDEBUG] \0", "[DEBUG] \0", "[INFO] \0",
-	"[NOTICE] \0", "\033[1;34m[WARN]\033[0m \0", "\033[1;31m[ERROR]\033[0m \0",
-	"[CRITICAL] \0", "[ALERT] \0", "[EMERGENCY] \0", "\0"};
-
-
-void print_default(FILE* target, const char* str)
+class Formatter : public spdlog::formatter
```
ghost, line 26: This formatter is there just to support the loglevel "MSG_MESSAGEONLY" to not have any patterns

src/shogun/io/SGIO.h
```diff
@@ -84,124 +87,72 @@ enum EMessageLocation
...
-	if (SG_UNLIKELY(io->loglevel_above(MSG_GCDEBUG)))				\
-		io->message(MSG_GCDEBUG, __PRETTY_FUNCTION__, __FILE__, __LINE__, __VA_ARGS__);	\
-}
+#define SG_IO env()->io()
+#define _SRC_LOC SourceLocation{__FILE__, __LINE__, __PRETTY_FUNCTION__}
 
-#define SG_DEBUG(...) {												\
-	if (SG_UNLIKELY(io->loglevel_above(MSG_DEBUG)))					\
-		io->message(MSG_DEBUG, __PRETTY_FUNCTION__, __FILE__, __LINE__, __VA_ARGS__);	\
+#define SG_GCDEBUG(...) {								\
```
ghost: I have no strong feelings about macros, so I left them

src/shogun/io/SGIO.h
```diff
@@ -84,124 +87,72 @@ enum EMessageLocation
...
-		SG_IO->message(MSG_INFO,__PRETTY_FUNCTION__, __FILE__, __LINE__, __VA_ARGS__);	\
-}
-
-#define SG_SWARNING(...) { SG_IO->message(MSG_WARN,__PRETTY_FUNCTION__, __FILE__, __LINE__, __VA_ARGS__); }
-#define SG_STHROW(Exception, ...)                                              \
-	{                                                                          \
-		SG_IO->template error<Exception>(                                      \
-		    MSG_ERROR, __PRETTY_FUNCTION__, __FILE__, __LINE__, __VA_ARGS__);  \
-	}
-#define SG_SERROR(...) SG_STHROW(ShogunException, __VA_ARGS__)
```
ghost, line 170: no need for the "SG_S{loglevel}" class of macros

ghost (commented):
@vigsterkr need your input on this stuff

src/shogun/io/SGIO.h
karlnapf: yeah I actually never saw most of those used :)

src/shogun/io/SGIO.h
karlnapf: changing them would require some pro/con list I guess?

src/shogun/io/SGIO.h
```diff
@@ -84,124 +87,72 @@ enum EMessageLocation
...
-		    MSG_ERROR, __PRETTY_FUNCTION__, __FILE__, __LINE__, __VA_ARGS__);  \
-	}
-#define SG_CLASS_ERROR(c, ...)                                                 \
-	{                                                                          \
-		c::io->template error<ShogunException>(                                \
-		    MSG_ERROR, __PRETTY_FUNCTION__, __FILE__, __LINE__, __VA_ARGS__);  \
-	}
-#define SG_UNSTABLE(func, ...) { io->message(MSG_WARN, __PRETTY_FUNCTION__, __FILE__, __LINE__, \
+
+#define SG_UNSTABLE(func, ...) { SG_IO->message(MSG_WARN, _SRC_LOC, \
```
karlnapf: I never liked this message .... maybe we can remove it and only merge tested/stable code? :D

src/shogun/io/SGIO.h
karlnapf: Can you briefly point me to why this is not needed anymore?

src/shogun/io/SGIO.h
```diff
@@ -231,34 +195,38 @@ class SGIO
...
 
 		/** destructor */
 		virtual ~SGIO();
 
+		/** (re)initializes the default (asynchronous) logger
+		 * 
+		 * @param queue_size size of the message queue
+		 * @param n_threads number of logging threads
+		 */
+		void init_default_logger(uint64_t queue_size=128, uint64_t n_threads=1);
```
karlnapf: why the n threads parameter? wouldnt we use the shogun number of threads?

src/shogun/io/SGIO.h
```diff
@@ -328,89 +300,45 @@ class SGIO
...
+		inline void gpl_only(const SourceLocation& loc={}) const
 		{
 			error<ShogunException>(
-			    MSG_ERROR, function, file, line, "This feature is only "
-			                                     "available if Shogun is built "
-			                                     "with GPL codes.\n");
+			    MSG_ERROR, loc,
+			    "This feature is only "
+			    "available if Shogun is built "
+			    "with GPL codes.\n");
```
karlnapf: the cmake error on this was recently updated by @sbrice .. maybe it should be updated here as well?

src/shogun/io/SGIO.h
```diff
@@ -493,40 +415,16 @@ class SGIO
...
-		 * @return reference count
-		 */
-		int32_t ref_count() const;
-
-		/** decrement reference counter and deallocate object if refcount is zero
-		 * before or after decrementing it
-		 *
-		 * @return reference count
-		 */
-		int32_t unref();
```
karlnapf, line 513: thank god this is gone :)

karlnapf (commented):
made some comments.

You mentioned some high level questions / ideas you had in the recent emails. Shall we discuss those here?

src/shogun/io/SGIO.h
ghost: SGObject's member variable "io" is removed, and only the global object is used "env()->io().

src/shogun/io/SGIO.h
ghost: to point out that these are different threads

karlnapf: could you give me some use cases for those threads?

ghost: In general, if the logging code is slow, the performance will take a great hit. And in multithreaded code, it will get even slower because of the mutex locks. Async loggers solve this by having multiple logging queues in different threads so that logging calls are nonblocking.

src/shogun/io/SGIO.h
ghost: they're mainly used with default constructors for classes that shouldn't support them

ghost:
I got rid of macros, but I'm not sure if that's a good thing, as macros are more efficient. Using functions means that all the arguments have to be evaluated, whether the argument is going to be used or not.
This is expensive if the arguments are function calls, and when the function is used a lot, like `require`.

src/shogun/io/SGIO.h
karlnapf: maybe we can kill them?
@vigsterkr @iglesias @lisitsyn

karlnapf:
Is there any way to quantify how much we are loosing when getting rid of macros?
I think before doing this for good, it would be good to have a pro/con list for macros/functions.
Thoughts?

src/shogun/io/SGIO.h
iglesias: > they're mainly used with default constructors for classes that shouldn't support them

Hmm. Why are they defined?

iglesias: > maybe we can kill them?
> @vigsterkr @iglesias @lisitsyn

It sounds good to me. Just stating the obvious here; let's get rid of the macro uses first and then let's remove the macro.

vigsterkr:
lets have the ci fixed, i.e. interfaces and go from there...

src/shogun/io/SGIO.h
karlnapf: or both at once? :)

ghost:
@karlnapf imo the only difference that would have an impact is the lazy evaluation of macros. Other than that, functions are heavily inlined by compilers, so they behave more or less like them. And macros have the disadvantage of polluting the symbol space, and might have unpredictable behaviour if not done correctly (actually there is such a case with the current implementation where you can't have them as one-line in if-else statement).
The performance would depend on the arguments being called, but I think the overall performance of any algorithm wouldn't get affected that much. Should I benchmark some algorithms before and after?

src/shogun/distance/MahalanobisDistance.cpp
```diff
@@ -37,7 +37,7 @@ bool CMahalanobisDistance::init(CFeatures* l, CFeatures* r)
 {
 	// FIXME: See comments in
 	// https://github.com/shogun-toolbox/shogun/pull/4085#discussion_r166254024
-	ASSERT(CRealDistance::init(l, r));
```
ghost, line 40: With ASSERT being removed on Release build, there should be a refactor to replace the user facing ones with require.

karlnapf: That’s good!
So just to make sure o understand: assert then will become something like in Java: it is ignored unless explicitly asked not to (ie debug mode build)?
That’s good as we can be more liberal with expensive checks.

vigsterkr:
@theartful mmm i'm not so sure what's happening with the windows build: https://dev.azure.com/shogunml/shogun/_build/results?buildId=2152&view=logs&jobId=9ffa40d6-0008-5f71-b4ad-2956145303e6&taskId=4c6a8d5c-d9f7-5d77-51f8-d43faa094742&lineStart=40569&lineEnd=40570&colStart=1&colEnd=1

vigsterkr:
or maybe it really runs out of time... and this is just the message of the cancellation. could you maybe change the maximum minutes for the windows ci job to 180 minutes?

ghost:
Yes, it runs out of time. I'm writing unit-tests, once finished, I will push with the increased maximum time for windows.

src/shogun/io/SGIO.h
ghost: @iglesias We support default constructors for serialization and deserialization. Also for testing cloning.

src/shogun/distance/MahalanobisDistance.cpp
ghost: yup

vigsterkr:
@theartful this is good to be merged, right?

ghost:
@vigsterkr One final issue: I noticed that spdlog adds a new line at the end of any message by default, however, this can be prevented using
```
#define SPDLOG_EOL ""
```
which I added in `SGIO.cpp`. If you prefer the default behaviour, I will remove this line and the new line character at the end of the current logs.

vigsterkr:
since the formatting of the logs are prefixed with time and other information i'd say let's keep the default behaviour of spdlog (append '\n' to the end of the log). this would basically mean that in case of log commands we would need to remove '\n' in order to not to have a double new line, right? this could be handled by a regex, or?

ghost:
I agree, and yes regex would do the job. I will leave `io::print` without the newline at the end, since it has no formatting.

vigsterkr:
sounds good! once you've pushed and CI passed i'll merge this! thnx for this patch!

vinx13:
hey @theartful nice work! can you update the doc on wiki https://github.com/shogun-toolbox/shogun/wiki/Assertions , and put a bit details on your final blog

ghost:
@vinx13 sure, will do that.

karlnapf:
> hey @theartful nice work! can you update the doc on wiki https://github.com/shogun-toolbox/shogun/wiki/Assertions , and put a bit details on your final blog

I second that actually, might be worth going through the docs a bit and see what needs update ... (or at least remove stale parts in wiki/docs)

PR #4730, opened 2019-07-18
Use static singleton in ShogunEnv

lisitsyn:


lisitsyn:
@theartful

ghost:
that's better to call the destructor to clean up when program ends (y)

PR #4731, opened 2019-07-20
Port cartree, svrlight and svmlight examples

ghost:


src/shogun/multiclass/tree/CARTree.cpp
```diff
@@ -1496,7 +1498,6 @@ void CCARTree::init()
 	SG_ADD(&m_nominal, "nominal", "feature types");
 	SG_ADD(&m_weights, "weights", "weights");
 	SG_ADD(&m_weights_set, "weights_set", "weights set");
-	SG_ADD(&m_types_set, "types_set", "feature types set");
```
ghost, line 1499: this flag that is set automatically when setting m_nominal, and shouldn't be a property. it also caused some warnings since putting "m_nominal" doesn't update this flag

karlnapf: But it should be registered or? Think serialisation

examples/meta/src/binary/svmlight_batch_linadd.sg
```diff
@@ -0,0 +1,35 @@
...
+
+#![create_classifier]
+Machine svm = machine("SVMLight", kernel=svm_kernel, labels=labels_train, C1=2.3, C2=2.3, epsilon=0.0000001)
+#![create_classifier]
+
+#![train_and_apply]
+svm.train()
+svm_kernel.init(feats_train, feats_test)
+svm.put("use_batch_computation", False)
+svm.put("use_linadd", False)
```
karlnapf: Why not true?

examples/meta/src/binary/svmlight_batch_linadd.sg
```diff
@@ -0,0 +1,35 @@
...
+Labels labels_predict_no_batch = svm.apply(feats_test)
+svm.put("use_batch_computation", False)
```
karlnapf: True?

src/shogun/multiclass/tree/CARTree.h
```diff
@@ -424,6 +424,10 @@ class CCARTree : public RandomMixin<CTreeMachine<CARTreeNodeData>>
 	static const float64_t EQ_DELTA;
 
 protected:
+	/** Returns whether the type of various feature dimensions are specified
+	 * using is_nominal_feature
+	 */
+	bool are_types_set();
```
karlnapf: Just types_set

examples/meta/src/binary/svmlight_batch_linadd.sg
```diff
@@ -0,0 +1,35 @@
+File f_feats_train = csv_file("../../data/fm_train_dna.dat")
+File f_feats_test = csv_file("../../data/fm_test_dna.dat")
+File f_labels_train = csv_file("../../data/label_train_dna.dat")
+
+#![create_features]
+Features feats_train = string_features(f_feats_train, enum EAlphabet.DNA)
+Features feats_test = string_features(f_feats_test, enum EAlphabet.DNA)
+Labels labels_train = labels(f_labels_train)
```
karlnapf, line 8: For all examples that use code that is guarded (gpl, svm light) you need to add them to ignore lists in FindMetaExamples.cmake

karlnapf (changes requested):

src/shogun/multiclass/tree/CARTree.cpp
ghost: yup, good point

examples/meta/src/binary/svmlight_batch_linadd.sg
ghost: the python example made it false, but I think true makes sense (example name :"D)

src/shogun/multiclass/tree/CARTree.cpp
```diff
@@ -340,7 +346,8 @@ CBinaryTreeMachineNode<CARTreeNodeData>* CCARTree::CARTtrain(CDenseFeatures<floa
 				float64_t sum = linalg::dot(labels_vec, weights);
 				// lsd*total_weight=sum_of_squared_deviation
 				float64_t tot=0;
-				node->data.weight_minus_node=tot*least_squares_deviation(labels_vec,weights,tot);
```
ghost, line 343: This is UB. GCC produces the intended result, while clang produces 0. Caused the regression integration test to fail.

examples/meta/src/binary/svmlight_batch_linadd.sg
ghost: done

PR #4732, opened 2019-07-20
build: Describe method for downloading GPL codes in build error message

samdbrice:


samdbrice:
Need to fix the

samdbrice:
Not sure what's causing the FOLDING_EXPRESSIONS failure

gf712:
That’s an error you get when the compiler doesn’t support c++17 fold expressions

karlnapf:
only the CI needs to pass, dont worry about the buildbot atm.

If you think this is a better error msg, happy to merge it. From my perspective, there is not more information and I am always sceptical of quoting things like git commands as that leads to redundancy and it is hard to maintain (things tend to change). However, you have a better "outside" perspective :) so let's do it

samdbrice:
Sounds good to me. Feel free to merge this.

PR #4733, opened 2019-07-22
Use 3 threads instead of 1 in linux ci

ghost:


iglesias (approved):

karlnapf:
I vaguely remember that the number of jobs was limited due to memory issues.

ghost:
This was newly introduced in the move to ninja instead of make (2 or 3 weaks ago). Previously it had 3 threads.

ghost:
Nope this was introduced before it. I made this change in the spdlog PR to speed up compile time, and there was no memory issues, but this might be just luck. We can close this if that's the case.

karlnapf:
Happy to try it :)
But maybe let’s wait whether Viktor has a comment for a few more days... he wrote the whole thing

vigsterkr:
@theartful i had troubles in the past (running out of memory) when running more jobs... but then again if it works (as it seems) lets merge it, maybe the agent's memory quotas has been changed in the meanwhile on azure?

vigsterkr:
@theartful i had troubles in the past (running out of memory) when running more jobs... but then again if it works (as it seems) lets merge it, maybe the agent's memory quotas has been changed in the meanwhile on azure?

iglesias:
I merge just to get this going as we are all good with it.

PR #4735, opened 2019-07-24
Drop SGString

ghost:
The main difference between SGString and SGVector is reference counting. I tried to keep reference counting on, and remove SG_FREE, and depend on destructors instead.


ghost:
azure is not triggered for some reason :(

tests/unit/utils/Utils.h
```diff
@@ -41,10 +42,7 @@ SGStringList<T> generateRandomStringData(
 		    min_string_length;
 		SGVector<T> current(len);
 		/* fill with random uppercase letters (ASCII) */
-		for (index_t j = 0; j < len; ++j)
-		{
-			current[j] = (T)(std::rand() % ('Z' - 'A' + 1) + 'A');
-		}
+		random::fill_array(current, 'A', 'Z', prng);
```
karlnapf, line 45: does this include 'z'? :)

src/interfaces/csharp/swig_typemaps.i
```diff
@@ -190,18 +190,17 @@ TYPEMAP_SGMATRIX(float64_t, double, double)
 		return $null;
 	}
 
-	shogun::SGString<SGTYPE>* strings=SG_MALLOC(shogun::SGString<SGTYPE>, rows_$1);
+	shogun::SGVector<SGTYPE>* strings=SG_MALLOC(shogun::SGVector<SGTYPE>, rows_$1);
```
karlnapf: these things can be tested with appropriate meta examples / integration tests... and should be!

src/shogun/base/SGObject.cpp
```diff
@@ -449,7 +449,7 @@ SGStringList<char> CSGObject::get_modelsel_names()
         char* name=m_model_selection_parameters->get_parameter(i)->m_name;
         index_t len=strlen(name);
 		// +1 to have a zero terminated string
-        result.strings[i]=SGString<char>(name, len+1);
+        result.strings[i]=SGVector<char>(name, len+1);
```
karlnapf: lets drop this method... @gf712 has written nice replacements for it

src/shogun/lib/SGStringList.cpp
```diff
@@ -14,7 +14,7 @@ SGStringList<T>::SGStringList() : SGReferencedData()
 }
 
 template <class T>
-SGStringList<T>::SGStringList(SGString<T>* s, index_t num_s, index_t max_length,
+SGStringList<T>::SGStringList(SGVector<T>* s, index_t num_s, index_t max_length,
```
karlnapf: long term...this structure should be replaced with std::vector .. but maybe not just yet to keep the PR smaller?

karlnapf (commented):
Great!

I think your strategy is good....does it work? (our CI doesnt really check mem errors)

tests/unit/utils/Utils.h
ghost: Yup. For some reason, "uniform_int_dist" accepts inclusive range [a,b], while "uniform_real_dist" is [a,b)

src/shogun/lib/SGStringList.cpp
ghost: yes. also we need to get rid of all of these SGString pointer arrays and replace them either with std::vector or SGStringList.
the problem with std::vector is that they're expensive to copy, and can't be shared. in Qt for example there is QStringList which does intrusive reference counting like we do here.

edit: qt uses copy-on-write data structures not intrusive reference counting

ghost: I noticed that pointers of SGString are passed around alot, so I assumed they're shared between classes, I will check.

ghost:
I made sure that I remove every SG_FREE(sgstring.string), and almost all SG_MALLOC(sgstring.string), except for some small cases where it didn't seem to matter. so now it should be all up to reference counting. but I will run valgrind on examples that use string features to make sure.

ghost:
Using std::vector<SGVector<T>> instead of SGStringList<T> has a couple of problems:
1. std::vector<SGVector<T>> is not serializable (needs to be wrapped by an SGObject)
2. also it's not wrappable by some (edit: it's not a problem)
I also spent quite some time trying to make swig typemap work for it, but I will try again tomorrow :")

tests/unit/utils/Utils.h
karlnapf: is that something we came up with, or someone else? :)
Because usually, I would generally expect the int version to be open as well, e.g.
https://docs.scipy.org/doc/numpy-1.15.1/reference/generated/numpy.random.randint.html

src/shogun/lib/SGStringList.cpp
karlnapf: I think they are.... updates here?
the std::vector caused problems in the end I assume?

karlnapf:
> Using std::vector<SGVector> instead of SGStringList has a couple of problems:
> 
>     1. std::vector<SGVector> is not serializable (needs to be wrapped by an SGObject)
> 

It should be serializable if registered as a parameter no? Was SGStringList serializable before? Because if not I dont understand why this would matter here?

>     2. also it's not wrappable by some (edit: it's not a problem)
>        I also spent quite some time trying to make swig typemap work for it, but I will try again tomorrow :")

Re the typemap, yes we had explicit typemaps for SGStringList, so I assume you want to overload the existing std::vector typemap of swig to deal with the cases here?

src/interfaces/csharp/swig_typemaps.i
```diff
@@ -309,7 +307,7 @@ TYPEMAP_STRINGFEATURES(float64_t, double, double)
 }
 
 %typemap(out) shogun::SGStringList<char> {
-	shogun::SGString<char>* str = $1.strings;
+	shogun::SGVector<char>* str = $1.strings;
```
karlnapf: I wonder: as an intermediate step (before using std::vector for SGStringList), why don't we rename SGStringList to SGVectorList?

Also, I guess we should rename `strings` to `vectors`

karlnapf (commented):
Reading through this I think it is a good idea to keep the list for now (rename it), and refactor it in a second step..... what do you think?

ghost:
I've already changed all the use cases for SGStringList, and dropped it locally. There is a couple of bugs that I will fix, but I don't see any more problems.
If it makes for better review, then sure, let's split them!

karlnapf:
Ah really. Great then! No need to split in that case, just suggested it as I thought maybe it will be too painful to change all those things at once :) But go ahead then!

src/shogun/io/Serializable.h
```diff
@@ -127,31 +127,51 @@ template<class T> class CMatrixSerializable: public CSerializable<SGMatrix<T>>
...
-		m_parameters->add_vector(&m_value.strings, &m_value.num_strings, "value",
-				"Serialized value");
+		if constexpr (type_internal::is_sg_primitive<T>::value)
+			set_generic<T>();
+		// FIXME(?): std::vector<bool> is not a container
+		// doesn't play well with the parameter framework
+		if constexpr (!std::is_same<T, bool>::value)
+			watch_param("value", &m_value);
+		else
+			SG_ERROR("std::vector<bool> is not currently serializable. "
```
gf712: shouldn't this be a compile time error?

ghost: Yes, but class_list instantiates it for all types, so if I did that it wouldn't compile

gf712: ah i see, nvm then

tests/unit/utils/Utils.h
ghost: Nope, I just followed how std::uniform_int_dist does it
https://en.cppreference.com/w/cpp/numeric/random/uniform_int_distribution

examples/meta/generator/translate.py
```diff
@@ -399,12 +399,15 @@ def translateDependencyElement(self, dependencyElement):
 
     def getIncludePathForClass(self, type_):
         translatedType = self.translateType({"ObjectType": type_})
+        stdParameterMatcher = 'std::[a-zA-Z_]*\<(.*)\>'
```
karlnapf, line 402: ++

src/interfaces/csharp/swig_typemaps.i
```diff
@@ -106,7 +105,7 @@ TYPEMAP_SGVECTOR(float64_t, double, double)
 		return $null;
 	}
 
-	for (i = 0; i < rows_$1 * cols_$1; i++) {
+	for (int32_t i = 0; i < rows_$1 * cols_$1; i++) {
```
karlnapf, line 108: ++

src/interfaces/csharp/swig_typemaps.i
```diff
@@ -190,34 +189,28 @@ TYPEMAP_SGMATRIX(float64_t, double, double)
...
+%typemap(freearg) std::vector<shogun::SGVector<SGTYPE>> {
+	typemap_utils::free_if_pointer($1);
 }
 
-%typemap(out) shogun::SGStringList<SGTYPE> {
-	shogun::SGString<SGTYPE>* str = $1.strings;
-	int32_t i, j;
-	int32_t rows = $1.num_strings;
-	int32_t cols = str[0].slen;
+%typemap(out) std::vector<shogun::SGVector<SGTYPE>> {
```
karlnapf: did you memcheck the examples that use this typemap?

src/interfaces/csharp/swig_typemaps.i
```diff
@@ -285,51 +281,45 @@ TYPEMAP_STRINGFEATURES(float64_t, double, double)
...
-		res[i + 1] = SG_MALLOC(char, str[i].slen + 1);
-		sg_memcpy(res[i + 1], str[i].string, str[i].slen * sizeof(char));
+	
+	for (int i = 0; i < size; i++) {
+		res[i + 1] = SG_MALLOC(char, strings[i].vlen + 1);
+		sg_memcpy(res[i + 1], strings[i].vector, strings[i].vlen * sizeof(char));
 		
 		// null terminate string as C#'s Marshal.PtrToStringAnsi expects that
-		res[i+1][str[i].slen] = '\0';
+		res[i+1][strings[i].vlen] = '\0';
```
karlnapf: this causes me some trouble.
Whereas previously, we had the semantic idea of string-lists, this conversion to strings made sense.
But not we have a vector of SGVectors, why would be convert this to continuous style strings -- in csharp, arrays of chars and strings are different types right?
I didn't think of this problem before.... what are your thoughts on this?

karlnapf: Maybe the solution would be to use `std::vector<std::string>` for collections of character strings?

src/interfaces/perl/swig_typemaps.i
```diff
@@ -279,30 +274,11 @@ fail:
                         el_len = strnlen(el_str, l_len_max);
                         if(i + el_len > it->nvals) {
                             warn("string_from_pdl::offset error in string conversion::bayling out");
-                            for(int32_t j = 0; j < lind; j++)
-                                l_ss[j].~SGString< type >();
-                            SG_FREE(l_ss);
                             //free(inds);
                             return false;
                         }
-                        new(&l_ss[lind]) SGString< type >();
```
karlnapf, line 288: wow what code!! :D

src/interfaces/swig/Library.i
```diff
@@ -122,80 +120,6 @@ namespace shogun
     %template(ComplexSparseMatrix) SGSparseMatrix<complex128_t>;
 #endif
 
-#ifdef USE_BOOL
-    %template(BoolStringList) SGStringList<bool>;
-#endif
-#ifdef USE_CHAR
-    %template(CharStringList) SGStringList<char>;
-#endif
```
karlnapf, line 130: good to see those gone!

src/interfaces/swig/typemaps_utilities.i
```diff
@@ -0,0 +1,33 @@
...
+		{
+			v = T{};
+			return v;
+		}
+	}
+
+	template <typename T>
+	void free_if_pointer(T& t)
+	{
+		if constexpr (std::is_pointer_v<T>)
```
karlnapf: ++ :)

karlnapf (approved):
This is fantastic work @theartful ... really impressive that you managed to edit all the typemaps and the whole codebase! Well done :)

I made a few inline comments. Otherwise, once this is tested/mechecked, happy to merge it as soon as CI is content

src/interfaces/perl/swig_typemaps.i
ghost: I had to look up the syntax :"D

src/interfaces/csharp/swig_typemaps.i
ghost: I'm not sure I understand. We can't in a simple manner replace `std::vector<SGVector<char>>` with `std::vector<std::string>` as they have different semantics. Most of the functions (and classes) that accept `std::vector<SGVector<T>>` will be specialized for strings.

ghost:
I memchecked the cpp examples, and at the beginning it leaked a lot, but now it leaks less than the original. I will fix these leaks, then check the typemaps. I checked that the `%freearg` directives work as intended (even if some error was thrown inside the typemap, the freearg is called). However I currently don't know how to test for memory leaks for other languages.

ghost:
I fixed the memory leaks, so now string_features related examples don't leak. Java typemaps also work now.

src/interfaces/csharp/swig_typemaps.i
ghost: Java also is the same. `int[][]` is used for example for string features of ints, while `String[]` for chars. Maybe be consistent with all and use `char[][]`? But then it's harder for the user.
I think we don't have to expose how the internals work. For all the user cares, it's still an array of strings.

ghost: I also think that's why we should somehow separate between `SGVector` and `SGString`. We can create an `SGString` class that inherits from `SGVector` and has the functionality of returning/accepting std::strings if `T` is `char`, so that this convenience maps to c++ users as well.

karlnapf: > I'm not sure I understand. We can't in a simple manner replace `std::vector<SGVector<char>>` with `std::vector<std::string>` as they have different semantics. Most of the functions (and classes) that accept `std::vector<SGVector<T>>` will be specialized for strings.

yes, that's exactly what I mean: here we convert SGVector<char> to String, which is kinda mixing sutff.

karlnapf: > I also think that's why we should somehow separate between `SGVector` and `SGString`. We can create an `SGString` class that inherits from `SGVector` and has the functionality of returning/accepting std::strings if `T` is `char`, so that this convenience maps to c++ users as well.

mmh that might be an indea.

Or, I also also the other comment you made: users dont have to see the internals.
From an internals perspective, there is no point is representing anything as std::string, but rather SGVector<char>.
So we can

* map `SGVector<char>` to the native string type in the typemaps (currently done). Maybe not the nicest though convenient for users
* map to the native array types instead (array of char), and then offer some interface to extract string representations. This could be done at the `StringFeatres` level, i.e. `get_strings` (or different name) returns `std::vector<std::string>` but this method only works for `SGVector<char>`. And then `get_vectors` would return `std::vector<SGVector<...>>`. This way we could have clean semantics in internals (and typemaps) and wouldnt need to append `\0` at the end. Instead, all this would be happening inside our new getter. Get my point?

karlnapf: Certainly we should have some conversion methods from SGVector<char> to std::string as well

karlnapf: use case (Python)
```
data = ["ACG", "GCA"]
feats = string_features(data) # this automatically maps the given strings to `SGVector<char>` internally
svm.train(feats)
feats.get_strings() # returns `["ACG", "GCA"]` # would throw a runtime error for other template parameters of string features
feats.get_vectors() # returns `np.array(["ACG", "GCA"], dtype=char)
```
in those lines

karlnapf:
great! the CI is now green as well.
What's  missing?

src/interfaces/csharp/swig_typemaps.i
ghost: @karlnapf To make sure I understand correctly, you meant
``` python
feats.get_vectors() # returns `np.array([['A', 'C', 'G'], ['G', 'C', 'A']], dtype=char)
```
right? I'm ok with doing that, but I somewhat don't like the asymmetry, it's a bit confusing, and don't completely get the point.
What's confusing is: if a function takes `std::vector<SGVector<char>>` as an argument, the user should provide `String[]`. But if a function returns it (might be `get` from `SGObject`), he should expect to get `char[][]`. If we are not to treat `SGVector<char>` differently, we should do it all the way.

ghost:
Addressing comments, and final touches. Maybe also renaming `std::vector<SGVector>` to `SGVectorList`, but i'm not sure if exuberant tags will be fine with it.

src/interfaces/csharp/swig_typemaps.i
karlnapf: I meant that, yes, sorry!

karlnapf: Sorry, I haven't been clear. The asymmetry is what I don't like either, and I also don't like special treatments (`SGVector<char>` as `String`).

So the two fixes would be: everything is a vector, no more strings internally. But then we can have user-facing methods where users can pass their strings directly to methods that expect arrays of chars. Maybe let's discuss this in irc?

src/interfaces/python/swig_typemaps.i
```diff
@@ -165,51 +165,9 @@ static int is_pysparse_matrix(PyObject* obj, int typecode)
          ) ? 1 : 0;
 }
 
-static int is_pystring_list(PyObject* obj, int typecode)
-{
```
ghost, line 169: no need for this method, we can depend on std::vector typemaps

src/interfaces/python/swig_typemaps.i
```diff
@@ -346,155 +304,6 @@ static bool array_to_numpy(PyObject* &obj, SGNDArray<type> sg_array, int typecod
 	return descr!=NULL;
 }
 
-template <class type>
-static bool string_from_strpy(SGStringList<type>& sg_strings, PyObject* obj, int typecode)
-{
```
ghost, line 351: no need for this method by depending on std::string in typemap

src/interfaces/python/swig_typemaps.i
```diff
@@ -346,155 +304,6 @@ static bool array_to_numpy(PyObject* &obj, SGNDArray<type> sg_array, int typecod
...
-    else
-    {
-        PyErr_SetString(PyExc_TypeError,"not a/empty list");
-        return false;
-    }
-}
-
-template <class type>
-static bool string_to_strpy(PyObject* &obj, SGStringList<type> sg_strings, int typecode)
-{
```
ghost, line 455: same, we can depend on std::string out typemap

src/interfaces/python/swig_typemaps.i
```diff
@@ -824,71 +633,129 @@ static bool spvector_to_numpy(PyObject* &obj, SGSparseVector<type> sg_vector, in
...
-/* One dimensional output arrays */
-%define TYPEMAP_OUT_SGVECTOR(type,typecode)
-%typemap(out) shogun::SGVector<type>
+// tell swig about the traits of the type to be used in stl containers
+%fragment(SWIG_Traits_frag(shogun::SGVector<type>), "header",
+          fragment="StdTraits",
+          fragment=SWIG_From_frag(shogun::SGVector<type>),
+          fragment=SWIG_AsVal_frag(shogun::SGVector<type>),
+          fragment=SWIG_From_frag(std::string),
+          fragment=SWIG_AsVal_frag(std::string))
```
ghost, line 668: this fragment makes std::vector aware of SGVector typemaps

src/interfaces/python/swig_typemaps.i
```diff
@@ -824,71 +633,129 @@ static bool spvector_to_numpy(PyObject* &obj, SGSparseVector<type> sg_vector, in
...
+        {
+            if constexpr (!std::is_same<type, char>::value)
+            {
+                return SWIG_AsVal_dec(shogun::SGVector<type>)(obj, val);
+            }
+            else
+            {
+                // treat SGVector<char> as a string
+                std::string str;
+                int result = SWIG_AsVal_dec(std::string)(obj, &str);
```
ghost, line 701: use `std::string` "in" typemap to do the conversion

src/interfaces/python/swig_typemaps.i
```diff
@@ -824,71 +633,129 @@ static bool spvector_to_numpy(PyObject* &obj, SGSparseVector<type> sg_vector, in
...
+            if constexpr (!std::is_same<type, char>::value)
+            {
+                return SWIG_From(shogun::SGVector<type>)(val);
+            }
+            else
+            {
+                // treat SGVector<char> as a string
+                std::stringstream ss;
+                ss.write((char *)val.vector, val.vlen);
+                return SWIG_From_dec(std::string)(ss.str());
```
ghost, line 728: use std::string "out" typemap to do the conversion

src/interfaces/python/swig_typemaps.i
```diff
@@ -824,71 +633,129 @@ static bool spvector_to_numpy(PyObject* &obj, SGSparseVector<type> sg_vector, in
...
+            }
+        }
+    };
+}
 }
+
+%typemaps_asvalfromn(SWIG_TYPECHECK_POINTER, shogun::SGVector<type>);
```
theartful: this macro defines the "in" and "out" typemaps using the fragment conventions I used above

src/interfaces/python/swig_typemaps.i
karlnapf: great!

src/interfaces/python/swig_typemaps.i
karlnapf: pretty cool! :)

karlnapf (commented):
I really like the swig refactorings!

karlnapf:
Ok, this seems to be good to be in?

examples/meta/generator/targets/cpp.json
```diff
@@ -59,8 +59,8 @@
         "RealFeatures": "DenseFeatures<float64_t>",
         "RealSubsetFeatures": "DenseSubsetFeatures<float64_t>",
         "StringCharFeatures": "StringFeatures<char>",
-        "StringCharList": "SGStringList<char>",
-        "StringWordList": "SGStringList<uint16_t>",
+        "StringCharList": "std::vector<SGVector<char>>",
```
karlnapf, line 62: i wonder whether we should rename "List" to "Vector"?
But this is minor

src/interfaces/swig/typemaps_utilities.i
```diff
@@ -0,0 +1,74 @@
+/**
```
karlnapf, line 1: maybe we should add a license here, as well as you as author

karlnapf: oh there is an issue here: swig is gpl licensed. So we cannot simply copy paste code into our codebase as it would turn our code into GPL as well....needs to be re-written
On the other hand, I wonder whether someone cares ;)

src/shogun/features/StringFeatures.h
```diff
@@ -380,7 +378,7 @@ template <class ST> class CStringFeatures : public CFeatures
 		 * not possible with subset
 		 *
 		 */
-        void set_features(SGStringList<ST> feats);
+        bool set_features(const std::vector<SGVector<ST>>& string_list);
```
karlnapf, line 381: nicer interface!

src/interfaces/swig/typemaps_utilities.i
ghost: @karlnapf Swig has this clause in their website
```
When SWIG is used as it is distributed by the SWIG developers, its output is not governed by SWIG's license (including the GPL). SWIG's output contains code from three sources:

code generated by SWIG, which is not governed by copyright;
code copied from the SWIG library which is permissively licensed to be redistributed without restriction;
code derived from the user's input, which may be governed by the license of the code supplied by the user.
```
The second source they mention seems to cover our case?

ghost: Nope, I don't think so, nvm :"D

ghost: @karlnapf Alright, I think this clears us http://www.swig.org/Release/LICENSE
```
The SWIG library and examples, under the Lib and Examples top level 
directories, are distributed under the following terms:

  You may copy, modify, distribute, and make derivative works based on
  this software, in source code or object code form, without
  restriction. If you distribute the software to others, you may do
  so according to the terms of your choice. This software is offered as
  is, without warranty of any kind.
```
These macros were taken from files under Lib directory.

examples/meta/generator/targets/cpp.json
ghost: I know this just for the meta language, but "List" is more language neutral and common than "Vector". The equivalent of vector in java is ArrayList, and in C# is List.

src/interfaces/swig/typemaps_utilities.i
karlnapf: ok great! thanks for sorting it out!

examples/meta/generator/targets/cpp.json
karlnapf: good points and I agree! Let's keep it!

ghost:
@karlnapf yes this should be good.

karlnapf (approved):
great work and very helpful! 3.5k lines to 1.3k, and more neat!

PR #4737, opened 2019-07-29
Support swig 4.0.0

### LINKED ISSUE #4727: force disable USE_SWIG_DIRECTORS in case of SWIG 4.0.0
swig 4.0 uses std::unexpected_handler when directors are enabled, which is removed in c++17  (https://en.cppreference.com/w/cpp/error/unexpected_handler). So if SWIG 4.0.0 is used we need to forcefully disable USE_SWIG_DIRECTORS cmake setting.

this will be fixed in 4.0.1 and later (see: https://github.com/swig/swig/blob/master/CHANGES.current#L37) so the check should be only for swig 4.0.0.

ghost:
Fixes #4727

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -196,6 +196,12 @@ IF (${ANY_INTERFACE_ENABLED})
 	ENDIF()
 
 	FIND_PACKAGE(SWIG ${SWIG_VERSION_MINIMUM} REQUIRED)
+
+	IF(SWIG_VERSION EQUAL 4.0.0 AND USE_SWIG_DIRECTORS)
+		MESSAGE(FATAL_ERROR "Due to a bug, Swig directors are not supported with Swig 4.0.0.\n"
+		                    "Please disable USE_SWIG_DIRECTORS or upgrade Swig.")
+	ENDIF()
+
 	# use our own UseSWIG.cmake in order to be able to enable ccache-swig
 	SET(SWIG_USE_FILE ${CMAKE_SOURCE_DIR}/cmake/UseSWIG.cmake)
 	IF(ENABLE_CCACHE AND CCACHE_SWIG)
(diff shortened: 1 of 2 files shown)


src/shogun/util/factory.h
```diff
@@ -276,14 +276,14 @@ namespace shogun
...
 	{
 		CDenseLabels* result = nullptr;
 		// try to interpret as all dense label types, from most restrictive to
 		// least restrictive
-		try_labels<CBinaryLabels>(result, labels);
-		try_labels<CMulticlassLabels>(result, labels);
-		try_labels<CRegressionLabels>(result, labels);
+		try_labels<CBinaryLabels>(result, labels_vector);
+		try_labels<CMulticlassLabels>(result, labels_vector);
+		try_labels<CRegressionLabels>(result, labels_vector);
```
ghost, line 286: swig 4.0.0 had a bug when naming a function template argument with the same name as the function

CMakeLists.txt
```diff
@@ -196,6 +196,12 @@ IF (${ANY_INTERFACE_ENABLED})
 	ENDIF()
 
 	FIND_PACKAGE(SWIG ${SWIG_VERSION_MINIMUM} REQUIRED)
+
+	IF(SWIG_VERSION EQUAL 4.0.0)
+		MESSAGE("Swig 4.0.0 doesn't support directors with c++17")
```
karlnapf: maybe we can add a "disabling" to the message?
Or, I would like an error even more as I don't like silent changes of settings....

karlnapf: This error was only around in exactly 4.0.0?

ghost: yes. it was introduced in 4.0.0, and will be fixed in 4.0.1.
do you mean something like
"Disabling swig directors: Not supported in swig 4.0.0"

karlnapf: yes.
Or "Due to a bug, Swig directors are not supported with Swig 4.0.0. Please disable or upgrade Swig"

ghost: ok, I changed it

karlnapf (approved):

PR #4738, opened 2019-07-30
Remove list feature obj

ktiefe:


karlnapf:
For the future: please open/close PRs if something isnt right. You can very easily update old PRs through a force push, mentioned here https://github.com/shogun-toolbox/docs/blob/master/DEVELOPING.md

Issue #4593, opened 2019-03-27
p and q dimension reversed in Traceback


mazzma12:
Hi,
I am reporting a minor confusion, my traceback rshowed a SystemError when I called `mmd.compute_p_value(statistic)` with 
`mmd.set_null_approximation_method(sg.NAM_MMD2_GAMMA)` :

```python
SystemError: [ERROR] In file /feedstock_root/build_artefacts/shogun-cpp_1512688880429/work/shogun-shogun_6.1.3/src/shogun/statistical_testing/QuadraticTimeMMD.cpp line 439: Number of samples from p (1204) and q (2654) must be equal.
```

But the dimensions are reversed, p is of shape 2654 and q 1204. I doubled check using `mmd.get_p()` method

karlnapf:
Thanks for the report!

MaxGreil:
I can pick this issue up.
@karlnapf can you please explain what needs to be done here?

karlnapf:
* Read the developer readme
* find the place where the error message is raised
* revert the numbers
* send a PR

karlnapf:
Fixed via #4745 

Many thanks!

PR #4745, opened 2019-08-05
Reverse dimensions

MaxGreil:
This fixes issue #4593.
@karlnapf


karlnapf (approved):
That should do it! Thanks

@lamday you agree?

MaxGreil:
> That should do it! Thanks
> 
> @lamday you agree?

I think you meant @lambday

lambday:
Hey @MaxGreil! Thanks for sending the PR and tagging me. This indeed was the issue.

PR #4746, opened 2019-08-12
Add get_name to Seedable

ghost:
Calling pure virtual methods in the constructor is dangerous. The problem here is that Seedable calls `add_callback_function`, and in turn it calls
```
REQUIRE(
	    has_parameter(tag), "There is no parameter called \"%s\" in %s\n",
	    tag.name().c_str(), get_name());
```
Here `get_name` is pure virtual, and it's not yet defined because the chain of construction didn't reach the derived class yet. The only reason this doesn't segfault, or produce the error `pure virtual method called terminate called without an active exception` is the lazy nature of macros.

### DIFF
--- a/src/shogun/mathematics/Seedable.h
+++ b/src/shogun/mathematics/Seedable.h
@@ -36,6 +36,11 @@ namespace shogun
 			init();
 		}
 
+		virtual const char* get_name() const override
+		{
+			return "Seedable";
+		}
+
 	private:
 		void init()
 		{


karlnapf:
What would be the output of this error message now with this PR?

ghost:
get_name would return "Seedable" instead of the derived class name. But that's the case with any other class if construction failed midway. A better name would be "Seedable\<Parent::get_name()\>", but that won't work when Parent is SGObject.

karlnapf:
Mmmh this is pretty annoying.... but OK at least these errors will only ever occur for devs

karlnapf (approved):

PR #4747, opened 2019-08-15
Add distribution to base classes & port some distribution examples

ghost:


examples/meta/src/distribution/histogram.sg
```diff
@@ -0,0 +1,18 @@
...
+#![create_features]
+
+#![create_distribution]
+Distribution histogram = distribution("Histogram")
+histogram.set_features(embedded_feats)
+#![create_distribution]
+
+#![train_and_extract_values]
+histogram.train()
+RealVector histogram_array = histogram.get_real_vector("histogram")
```
ghost, line 15: This histogram implementation calculates frequencies in log space, but it's not normalized, which is a bit weird?

ghost: nvm, it was intentional, so maybe not
```C++
/** @brief Class Histogram computes a histogram over all 16bit unsigned
 * integers in the features.
 *
 * Values in histogram are absolute counts (logarithmic)
 */
```

karlnapf: i guess this is as to avoid overflows?
But that was written well before my (or anyone else's apart from @sonney2k ) time

examples/meta/src/distribution/histogram.sg
```diff
@@ -0,0 +1,18 @@
...
+RealVector out_likelihood = histogram.get_log_likelihood()
+real out_sample = histogram.get_log_likelihood_sample()
```
karlnapf, line 17: this API is .... ouch ;)
maybe we can update that at some point in the near future

karlnapf:
Nice one. Also thanks for the cleanups!

karlnapf:
Would be good to have a cookbook page for HMM ...

PR #4748, opened 2019-08-15
Fix C# StringList typemaps

ghost:
string_features example was disabled for C#, so I didn't notice. I changed C# type of string lists from `T[,]` to `T[][]`, since the latter enables variable length elements. Unfortunately marshaling `T[][]` wasn't possible (or at least I couldn't make it work), so I first flatten it to `T[]`.

I also noticed that `SGMatrix` and these typemaps leak memory, and it's not obvious to me currently how to solve it.

### DIFF
--- a/examples/meta/csharp/CMakeLists.txt
+++ b/examples/meta/csharp/CMakeLists.txt
@@ -9,11 +9,6 @@ FOREACH(META_EXAMPLE ${META_EXAMPLES})
     # assume a structure <target_language>/<category>/listing.sg
     GET_META_EXAMPLE_VARS(${META_EXAMPLE} EXAMPLE_NAME EXAMPLE_REL_DIR EXAMPLE_NAME_WITH_DIR)
 
-    # FIXME: enable string features once typemap is fixed
-    if (EXAMPLE_NAME STREQUAL "string_features")
-		continue()
-    endif()
-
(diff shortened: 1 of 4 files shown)


examples/meta/csharp/CMakeLists.txt
```diff
@@ -9,11 +9,6 @@ FOREACH(META_EXAMPLE ${META_EXAMPLES})
     # assume a structure <target_language>/<category>/listing.sg
     GET_META_EXAMPLE_VARS(${META_EXAMPLE} EXAMPLE_NAME EXAMPLE_REL_DIR EXAMPLE_NAME_WITH_DIR)
 
-    # FIXME: enable string features once typemap is fixed
```
karlnapf, line 12: nice thanks for that!

karlnapf:
> string_features example was disabled for C#, so I didn't notice. I changed C# type of string lists from `T[,]` to `T[][]`, since the latter enables variable length elements. 

ah yes, I remember that this was causing problems.... there were padding solutions which I always disliked. nice!

Unfortunately marshaling `T[][]` wasn't possible (or at least I couldn't make it work), so I first flatten it to `T[]`.

Should be ok. Maybe add a "TODO/FIXME" for future keen typemap hackers ;)

> 
> I also noticed that `SGMatrix` and these typemaps leak memory, and it's not obvious to me currently how to solve it.

Future work, but maybe can you file an issue with a reproducing example?

karlnapf (approved):
LGTM

PR #4754, opened 2019-09-01
Ignore GPU backend when user doesn't compile with viennacl

gf712:
This PR removes GPU backend checks when users don't compile shogun with viennacl and therefore can't use the GPU at any point at runtime
You shouldn't pay for what you don't use, but right now:
- every linalg operation checks if memory of the containers is on the GPU
- each copy (and eventually move) of a container checks if memory is on the GPU. This is an atomic operation so is unnecessarily costly imo..


karlnapf (approved):
Lgtm
Although I would even just remove the whole thing (not used in a single algorithm) and instead redesign the whole thing with Ahmed using eigen and co... maybe we can discuss this at some point soon

vigsterkr:
just slowing things down... shogun-graph FTW :)

PR #4755, opened 2019-09-03
Fix Lua interface [WIP]

gf712:


gf712:
as discussed with @vigsterkr will add/look into scilua for matrix and vector manipulation

gf712:
scilua seems a bit too difficult to integrate, it was not designed to be used from a C API, but rather expose C functions via lua's ffi.. That might be easier to implement, but potentially quite a big project.

With some modifications of the SWIG code I could generate this for example:

```cpp
extern "C" char* get_version_main() {
  char *result = 0 ;
  result = (char *)shogun::Version::get_version_main();
  return result;
}
```
```lua
>  ffi = require 'ffi' 
> shogun = ffi.load("/home/gf712/ulua/libshogun.so")
> ffi.cdef[[
>> const char* get_version_main();
>> ]]
> shogun.get_version_main()
> print(shogun.get_version_main())
cdata<const char *>: 0x7f6ae5080b00
> print(ffi.string(shogun.get_version_main()))
6.2.0
```
Once we move everything to just factories, it would be quite easy to implement this.

gf712:
for future reference https://csl.name/post/luajit-cpp/

karlnapf:
Any solutions for overloaded functions in lua?

gf712:
with the factories approach it shouldn't be an issue anymore no? But I am not sure I understand the function overloading issue... From what I see swig does parameter checks and dispatches to the correct function? If we don't use swig and use ffi for LuaJIT instead I think there wouldn't be any issues.

karlnapf:
I was thinking of „get“
Aaaah but that is renamed anyways

What about put?

PR #4756, opened 2019-09-03
clustering parameter properties

gf712:


src/shogun/clustering/KMeansBase.cpp
```diff
@@ -313,13 +313,17 @@ void CKMeansBase::init()
 	    &k, "k", "k, the number of clusters",
 	    ParameterProperties::HYPER | ParameterProperties::CONSTRAIN,
 	    SG_CONSTRAINT(positive<>()));
-	SG_ADD(&dimensions, "dimensions", "Dimensions of data");
-	SG_ADD(&fixed_centers, "fixed_centers", "Use fixed centers");
-	SG_ADD(&R, "radiuses", "Cluster radiuses");
+	SG_ADD(
+	    &dimensions, "dimensions", "Dimensions of data",
+	    ParameterProperties::MODEL);
```
gf712: @karlnapf not sure if this is model?

src/shogun/clustering/KMeansMiniBatch.cpp
```diff
@@ -121,13 +121,14 @@ SGVector<int32_t> CKMeansMiniBatch::mbchoose_rand(int32_t b, int32_t num)
...
+	void CKMeansMiniBatch::init_mb_params()
+	{
+		batch_size = 100;
 
-	SG_ADD(
-		&batch_size, "batch_size", "batch size for mini-batch KMeans");
-}
+		SG_ADD(
+		    &batch_size, "batch_size", "batch size for mini-batch KMeans",
+		    ParameterProperties::HYPER);
```
gf712: @karlnapf  hyper or setting?

karlnapf: Both?

src/shogun/clustering/KMeansBase.cpp
karlnapf: This is inferred from data or? Then read only

gf712: yes, it is set with the number of features of the data. read only and model, or just read only?

src/shogun/clustering/KMeansMiniBatch.cpp
gf712: OK! there are quite a few batch_size parameters in different algos, so let's say always hyper and setting for those?

src/shogun/clustering/KMeansBase.cpp
karlnapf: Just read only
Model parameters are those that are learnt by the train method

src/shogun/clustering/KMeansMiniBatch.cpp
karlnapf: Maybe let’s think about the implications of a moment.

gf712: well if we write some automatic hyperparameter tuning algorithm it would probably be a waste of resources to optimise batch size no? I mean it has an impact but if you choose a reasonable value it shouldn't be that important in prediction performance no? That's why i was thinking just setting

karlnapf: but the batch size needs to be tuned indeed for certain methods (batch size often relates to regularization so it can have a big impact)..... and of course we will have methods to exclude parameters from optimizing..... After thinking about it I think it should be both

gf712: ok!

PR #4757, opened 2019-09-04
Shared_ptr fixes

gf712:


src/shogun/evaluation/CrossValidationStorage.h
```diff
@@ -155,7 +155,7 @@ namespace shogun
 		std::shared_ptr<Labels> m_original_labels;
 
 		/** Vector with all the folds results */
-		std::vector<std::shared_ptr<CrossValidationFoldStorage>> m_folds_results;
+		std::vector<std::shared_ptr<EvaluationResult>> m_folds_results;
```
gf712, line 158: @karlnapf is there a reason why this was CrossValidationFoldStorage instead of the base class EvaluationResult?

src/shogun/multiclass/MulticlassSVM.cpp
```diff
@@ -45,6 +45,8 @@ bool MulticlassSVM::create_multiclass_svm(int32_t num_classes)
 	if (num_classes>0)
 	{
 		m_machines.clear();
+		for (auto i: Range(0, m_multiclass_strategy->get_num_machines()))
+			m_machines.push_back(nullptr);
```
gf712, line 49: @vigsterkr this a bit of a nonsense, but this is what it was before and now it works...

src/shogun/base/SGObject.h
```diff
@@ -421,18 +417,9 @@ namespace shogun
...
-			if (sgo_details::dispatch_array_type<T>(
-			        shared_from_this(), name, std::move(push_back_lambda)))
-				return;
-			SG_ERROR(
-			    "Cannot add object %s to array parameter %s::%s of type %s.\n",
-			    value->get_name(), get_name(), name.c_str(),
-			    demangled_type<T>().c_str());
+			auto dispatched = get(tag_vector);
+			dispatched.push_back(value);
+			update_parameter(BaseTag(name), make_any(dispatched), false);
```
gf712, line 422: @karlnapf we can't get a any_cast value by reference, we just get a copy. So need to put the copy to replace the original after pushing the value. But some parameters are readonly, so I added a flag to skip checks... It's not great but is the simplest solution until we can come up with a better design to update Any objects...

gf712:
@vigsterkr I think the only things failing now are related to the old parameter framework?

src/shogun/lib/any.h
```diff
@@ -655,7 +677,10 @@ namespace shogun {
 			{
 				return 0;
 			}
-			SG_FREE(ptr);
+			if constexpr (is_shared_ptr<T>::value)
+				ptr->reset();
```
gf712: call the shared_ptr reset method, instead of deleting with free. Not sure what happens with unique_ptr...

gf712: nvm, the issue is that we should call `delete[]` because this (could) represent an array with multiple pointers

gf712: and each pointer managed by shared_ptr manages another ptr, so basically have a 2D array.. I think that's it!

gf712: so have to delete the shared_ptr pointer and call it's destructor to destroy the managed pointer.. so `delete[] ptr`;

src/shogun/base/SGObject.h
karlnapf: Ok

PR #4758, opened 2019-09-06
removed evaluation objects from swig

gf712:


src/interfaces/swig/Evaluation.i
```diff
@@ -14,56 +14,16 @@
...
-%rename(DifferentiableFunction) CDifferentiableFunction;
 
 /* Include Class Headers to make them visible from within the target language */
-%include <shogun/evaluation/EvaluationResult.h>
 %include <shogun/evaluation/Evaluation.h>
-%include <shogun/evaluation/BinaryClassEvaluation.h>
+%include <shogun/evaluation/EvaluationResult.h>
 %include <shogun/evaluation/ClusteringEvaluation.h>
 %include <shogun/evaluation/ClusteringAccuracy.h>
 %include <shogun/evaluation/ClusteringMutualInformation.h>
```
gf712, line 28: @karlnapf can't remove these because of :
https://github.com/shogun-toolbox/shogun/blob/c948611b1e075b731d92c109d22abe79279b0a77/src/shogun/evaluation/ClusteringEvaluation.h#L36
not sure how to handle this?

karlnapf: I remember I discussed this with someone 
Think we need to refactor this so that either the best map is computed automatically by the eval (once) or make the user do the call (eg using a “run(“compute_best_map”) and a runtime error in eval if the user hasn’t done that)

karlnapf (commented):
Progress!!
I wonder how many classes are left ?

gf712:
not that many anymore tbh, but the SWIG wrapper is still huge (~800k loc) for some reason..

src/interfaces/swig/Evaluation.i
gf712: hmm, i see, i will look into that!

karlnapf:
I think removing all true templates feature classes will reduce this heaps
I had it down to a few secs compile time when o tried that a while ago

PR #4760, opened 2019-09-06
remove bagging objects from swig

gf712:


examples/meta/src/regression/random_forest_regression.sg
```diff
@@ -15,20 +15,17 @@ MeanRule mean_rule()
...
 #![train_and_apply]
-rand_forest.train()
+rand_forest.train(features_train)
 RegressionLabels labels_predict = rand_forest.apply_regression(features_test)
 #![train_and_apply]
 
 #![evaluate_error]
 Evaluation mse = evaluation("MeanSquaredError")
-real oob = rand_forest.get_oob_error(mse)
+# real oob = rand_forest.get_oob_error(mse)
```
gf712: need to find a solution for this

karlnapf: A function call(lazy) using watch_method?

karlnapf (approved):

examples/meta/src/regression/random_forest_regression.sg
gf712: you mean .put("oob_error_metric") and then .get("oob_error")?

karlnapf: Yes with a meaningful error if forgotten

tests/unit/multiclass/BaggingMachine_unittest.cc
```diff
@@ -151,7 +151,8 @@ TEST_F(BaggingMachine, classify_CART)
 	EXPECT_EQ(1.0,res_vector[4]);
 
 	auto eval = some<CMulticlassAccuracy>();
-	EXPECT_NEAR(0.642857,c->get_oob_error(eval),1e-6);
+	c->put("oob_error_metric", eval);
+	EXPECT_NEAR(0.642857,c->get<float64_t>("oob_error"),1e-6);
```
gf712: @karlnapf @vigsterkr not sure I like the put/get here. Maybe should have get_oob_error, and have another method that is exposed to the interface and added to parameter framework, i.e.
```cpp
watch_method("oob_error", &CBagginMachine::get_oob_error_interface)
...
float64_t CBagginMachine::get_oob_error_interface()
{
   require(m_oob_error_metric, "error message");
   return get_oob_error(m_oob_error_metric);
}
```

gf712: would then also have the respective setter to be used in the C++ interface

tests/unit/multiclass/BaggingMachine_unittest.cc
```diff
@@ -151,7 +151,8 @@ TEST_F(BaggingMachine, classify_CART)
 	EXPECT_EQ(1.0,res_vector[4]);
 
 	auto eval = some<CMulticlassAccuracy>();
-	EXPECT_NEAR(0.642857,c->get_oob_error(eval),1e-6);
+	c->put("oob_error_metric", eval);
```
vigsterkr: lets start to use what we've discussed in the past: if one uses put/get in C++ make sure that there's a static constexpr string_view defined for the key, something like:
```
static constexpr string_view kOOBErrorMetric("oob_error_metric")
```
and this is used when one registers the parameter in the class as well as when you put or get

tests/unit/multiclass/BaggingMachine_unittest.cc
vigsterkr: use `static constexpr string_view` for defining the "oob_error"

src/shogun/machine/BaggingMachine.cpp
```diff
@@ -255,18 +255,22 @@ void CBaggingMachine::set_machine_parameters(CMachine* m, SGVector<index_t> idx)
...
 	SG_ADD(&m_bags, "bags", "Bags array");
 	SG_ADD(
 	    &m_combination_rule, "combination_rule",
 	    "Combination rule to use for aggregating", ParameterProperties::HYPER);
 	SG_ADD(&m_all_oob_idx, "all_oob_idx", "Indices of all oob vectors");
-	SG_ADD(
-	    &m_oob_indices, "oob_indices", "OOB indices for each machine");
+	SG_ADD(&m_oob_indices, "oob_indices", "OOB indices for each machine");
 	SG_ADD(&m_machine, "machine", "machine to use for bagging");
+	SG_ADD(&m_oob_error_metric, "oob_error_metric",
```
vigsterkr: use `static constexpr string_view` to define "oob_error_metric"

src/shogun/machine/BaggingMachine.cpp
```diff
@@ -255,18 +255,22 @@ void CBaggingMachine::set_machine_parameters(CMachine* m, SGVector<index_t> idx)
...
+	    "metric to calculate the oob error");
+	watch_method("oob_error", &CBaggingMachine::get_oob_error);
```
vigsterkr: use `static constexpr string_view` to define "oob_error"

vigsterkr (changes requested):
minor changes and rebase plz

tests/unit/multiclass/BaggingMachine_unittest.cc
karlnapf: Not sure I understand the suggestion/problem Gil
This pattern is what we used for other similar situations, but the error check for the metric for sure needs to be done somewhere

karlnapf: Btw related to the constexpr Viktor was asking about, any news in making the map compile time? :D

gf712: @karlnapf the issue is that a user of the C++ API has to do a runtime lookup of the parameter map and then a any_cast..
There is a bit of progress with the compile time map. I have been thinking maybe we can just use an array and do a linear search. It would be easier to implement and a map might not be worth it for fewer than 20 or so key/value pairs

karlnapf: We could do some stats on how many parameters we have on average ...

karlnapf: Is the cast a problem if the method that is called is expensive (because it computes something)?

gf712: I am not sure how much more expensive, but I would expect so. I mean you need to get rtti.. But I guess as long as you don't use it in a loop it's fine!

tests/unit/multiclass/BaggingMachine_unittest.cc
gf712: @vigsterkr i added string_view support in #4761 so I'll wait for that to be merged to do that here. I had to add put/get methods for string_views, as they are not implicitly convertible to std::string

PR #4761, opened 2019-09-07
Interface SGObject::get visitor pattern

gf712:
Visitor pattern to have a single SGObject::get for Python API without trial and error


src/interfaces/swig/shogun.i
```diff
@@ -35,15 +35,379 @@
...
+    };
+#else
+    template <>
+    struct sg_to_npy_type<char>
+    {
+        const static NPY_TYPES type = NPY_STRING;
+    };
+#endif
+
+    class PythonVisitor : public AnyVisitor {
```
gf712: @vigsterkr I don't know if you want to centralise these somewhere?

src/interfaces/swig/shogun.i
```diff
@@ -35,15 +35,379 @@
...
+        {
+            PyObject* result = nullptr;
+            auto* obj = $self->get(name, index);
+            result = SWIG_InternalNewPointerObj(SWIG_as_voidptr(obj), SWIGTYPE_p_shogun__CSGObject, 0 |  0 );
+            return result;
+        }
+    }
+}
+
+%ignore get;
```
gf712: %ignore shogun::CSGObject::get and %ignore CSGObject::get didn't work here...

gf712:
need to find a way around the non visitable parameters registered with watch_method

src/shogun/lib/any.h
```diff
@@ -349,9 +356,22 @@ namespace shogun {
...
 		}
 
+		template <
+		    typename T,
+		    std::enable_if_t<std::is_same_v<
+		        std::result_of<std::function<T()>>::type, void>>* = nullptr>
+		void on(std::function<T()>* v)
+		{
+			value = std::make_shared<T>((*v)());
+			return on(static_cast<T*>(value.get()));
```
gf712: @vigsterkr [it ain't much...](https://i.kym-cdn.com/entries/icons/mobile/000/028/021/work.jpg)

gf712: the rest in this file is just clang-format

karlnapf (approved):
Nice!

tests/unit/lib/Any_unittest.cc
```diff
@@ -693,10 +693,10 @@ TEST(Any, lazy_cloneable_visitable)
...
 	EXPECT_TRUE(any.cloneable());
 	EXPECT_TRUE(any.visitable());
-	any = make_any<int>([=]() { return 222; });
-	EXPECT_FALSE(any.cloneable());
-	EXPECT_FALSE(any.visitable());
-	EXPECT_THROW(any.visit(nullptr), std::logic_error);
+//	any = make_any<int>([=]() { return 222; });
+//	EXPECT_FALSE(any.cloneable());
+//	EXPECT_FALSE(any.visitable());
+//	EXPECT_THROW(any.visit(nullptr), std::logic_error);
```
gf712: @karlnapf I think now everything is visitable, except void functions

tests/unit/lib/Any_unittest.cc
```diff
@@ -693,10 +693,10 @@ TEST(Any, lazy_cloneable_visitable)
 	    any.clone_from(make_any<int>([=]() { return 111; })), std::logic_error);
 	EXPECT_TRUE(any.cloneable());
 	EXPECT_TRUE(any.visitable());
-	any = make_any<int>([=]() { return 222; });
-	EXPECT_FALSE(any.cloneable());
-	EXPECT_FALSE(any.visitable());
-	EXPECT_THROW(any.visit(nullptr), std::logic_error);
```
iglesias, line 699: 🤔 is this why is WIP?

src/interfaces/swig/shogun.i
```diff
@@ -34,16 +34,357 @@
...
+				}
+				current_i++;
+			}
+
+			template <typename T>
+			void handle_pylist(const T* v)
+			{
+				bool new_obj = m_nested_py_obj ? false : true;
+				auto nested_visitor = PythonVisitor(m_nested_py_obj, dims, nested_current_i, n_dims);
+				nested_visitor.on(const_cast<T*>(v));
```
gf712: @karlnapf I think the visitor should be always const pointer, i.e. `on(const T*)`. I dont think there is a situation where you would change the pointer in the visitor?

src/interfaces/swig/shogun.i
```diff
@@ -34,16 +34,357 @@
...
+			{
+			  handle_sg(v);
+			}
+
+			void on(uint64_t *v) final
+			{
+				handle_sg(v);
+			}
+
+			void on(complex128_t *v) final
```
iglesias: Would it make sense to make all these on's less verbose?

tests/unit/lib/Any_unittest.cc
gf712: yea, it was also failing until now. Now all tests pass at least in my machine.. 
Need to figure out what can be visitable.. With this patch it's everything except void functions

src/interfaces/swig/shogun.i
gf712: less verbose in what sense? compress it to one line?

iglesias: Not needing to define all overloads this explicitly.

gf712: not sure you can because those are virtual functions? i havent tried, but i don't think you can put it all in a templated function..

gf712:
seems to be working!

src/interfaces/swig/shogun.i
```diff
@@ -34,16 +34,357 @@
...
+	{                                                  \
+		const static NPY_TYPES type = NPY_TYPE;        \
+	};
+
+	namespace shogun
+	{
+		template <typename T>
+		struct sg_to_npy_type {};
+		SG_TO_NUMPY_TYPE_STRUCT(bool,          NPY_BOOL)
+#ifdef PYTHON3 // str -> unicode for python3
```
vigsterkr: we dont support python2.x anymore so you can remove this

src/interfaces/swig/shogun.i
vigsterkr: @gf712 const T* doesn't work because actually in some circumstances with the visitor you set the values (see deserialization)... in this case it works of course but the base interface AnyVisitor::on should be T*

vigsterkr (changes requested):
great start buuuut... so the problem here atm is that unfortunately AnyVisitor cannot handle CKernel* etc, hence the new way of visitors from @lisitsyn.
i'll try to come up with a small example how i thought for interfaces...

src/interfaces/swig/shogun.i
gf712: ah makes sense, i was wondering why they weren't const

gf712:
R and python working

vigsterkr:
Amazingoooooooo ❤️🕺

lisitsyn:
What's working? I am lost :(

gf712:
@lisitsyn using the AnyVisitor to cast C++ to interface objects, i.e. C++ to PyObject and C++ to SEXP. So we can just use ::get(name) for everything, rather than the templates for each type, e.g get_real, get_int...
So now Python and R work, and the interface I wrote should be enough to do the same with the other dynamically typed languages

lisitsyn:
Sounds cool! How does that change the interface? There are no changes in tests/examples in the PR so I was wondering what does it look like for user.

gf712:
we already had this with python, but it worked differently, it had just a try except in a loop. The R meta examples change, if you see the R.json file has changed, now all get calls are just get

examples/meta/generator/targets/r.json
```diff
@@ -59,7 +59,17 @@
...
+            "get_int": "$object$$get($arguments)",
+            "get_long": "$object$$get($arguments)",
+            "get_int_vector": "$object$$get($arguments)",
+            "get_real": "$object$$get($arguments)",
+            "get_real_vector": "$object$$get($arguments)",
+            "get_real_matrix": "$object$$get($arguments)",
+            "get_char_string_list": "$object$$get($arguments)",
+            "get_word_string_list": "$object$$get($arguments)",
+            "get_option": "$object$$get($arguments)",
+            "get_string": "$object$$get($arguments)"
```
gf712, line 72: @lisitsyn everything becomes 
```r
val <- object$get(name)
```

tests/unit/lib/Any_unittest.cc
vigsterkr: @gf712 shouldn't this be reverted, i.e. enable the tests again?

examples/meta/generator/targets/r.json
karlnapf: Pretty nice! :)

karlnapf: Maybe we can think about refactoring the Json files then.... rather than listing all those replacements explicitly, we could maybe wildcard them ?

gf712: yes, maybe when we get more dynamically typed languages done? I am thinking octave is another low hanging fruit

karlnapf: Yeah sure, having your thing deployed to other Langs is higher priority, but the change to wildcard replacements is als very simple and can be optional without having to change all jsons

tests/unit/lib/Any_unittest.cc
gf712: i need to do some more refactoring here and rewrite the test, now this would be visitable and wouldn't throw

src/shogun/base/SGObject.h
```diff
@@ -394,7 +394,11 @@ class CSGObject
 	template <class T,
 		      class X = typename std::enable_if<is_sg_base<T>::value>::type,
 		      class Z = void>
+#ifdef SWIG
 	void put(const std::string& name, T* value)
+#else
+	void put(std::string_view name, T* value)
```
gf712, line 406: Added string_view for sgobject. There are no typemaps for it yet for swig so left std string signature for swig

src/shogun/mathematics/RandomMixin.h
```diff
@@ -97,6 +97,12 @@ namespace shogun
 		}
 
 		mutable PRNG m_prng;
+
+#ifndef SWIG
+	public:
+		static constexpr std::string_view kSetRandomSeed = "set_random_seed";
+		static constexpr std::string_view kSeed = "seed";
```
gf712, line 104: @vigsterkr is this what you have in mind for constexpr std::string_view tags?

iglesias:
> @lisitsyn using the AnyVisitor to cast C++ to interface objects, i.e. C++ to PyObject and C++ to SEXP. So we can just use ::get(name) for everything, rather than the templates for each type, e.g get_real, get_int...
> So now Python and R work, and the interface I wrote should be enough to do the same with the other dynamically typed languages

It sounds nice!

PR #4763, opened 2019-09-11
add cmake support for precompiled headers

vigsterkr:
do not merge as needs to be tested with cmake that actually supports `target_precompile_headers` command (3.16 or later, which as of this point has not been released yet)


vigsterkr:
dontdie! :)

PR #4772, opened 2019-09-18
safe visitable

gf712:
this doesn't change anything, just introduces the concept of something being safely visitable. For example a function can throw, so it might not always be a good idea to visit it.

Currently C++ doesn't have std::function<T() noexcept> but one day it might, and then we could actually have safe visitable functions where no error can be thrown. Or maybe we can think of some hack around this.


lisitsyn:
LGTM!

Issue #4710, opened 2019-07-06
refactor SGObject hacks


vigsterkr:
SGObject::dispatch_array_type SGObject::get_dispatch_all_base_types and related methods are all hacks that should be implemented using visitor patterns.

ktiefe:
First draft
https://github.com/shogun-toolbox/shogun/pull/4775

please check if it is a right way to resolve it.

ktiefe:
probably i will need to use this:
https://www.etlcpp.com/blog/2017/11/12/templated-implementation-of-the-visitor-pattern/

for next step.

PR #4775, opened 2019-09-20
Issue #4710: visitor for array_type dispatch

ktiefe:


src/shogun/base/SGObject.h
```diff
@@ -409,13 +427,16 @@ class CSGObject
 		      class X = typename std::enable_if<is_sg_base<T>::value>::type>
 	void add(const std::string& name, T* value)
 	{
+		s_visitor* a_t = new array_type(name);
```
ktiefe: smart pointer needs to be used here instead of raw

vigsterkr:
@ktiefe ok so here is the idea how to fix this:
first need to add an Any custom visitor in case the watched parameter is an array, something like this:
```
struct ArrayVisitor
{
    CSGObject* v;
};
Any::register_visitor<CDynamicObjectArray*, ArrayVisitor>([&] (auto value, auto visitor) {
	value->push_back(visitor->v);
});
```

in case of std::vector<T> this is a bit more complex, but could be done in add part
```
Any::register_visitor<std::vector<T*>, ArrayVisitor>([&] (auto value, auto visitor) {
	value.push_back(visitor->v);
});
```

and then finally in the SGObject::add you do this for the specific value:
```
ArrayVisitor array_visitor;
array_visitor.v = value;
auto array = get_parameter(BaseTag(name));
array.get_value().visit_with(&array_visitor);
```

Issue #4601, opened 2019-04-06
Performance Improvement in mathematics/Cplex.cpp


neroangelo296:
I was wondering if we could change 
```
for (int32_t i=0; i<num_variables; i++)
	{
		obj[i]=0;

		if (i<num_dim) //xi part
		{
			lb[i]=-CPX_INFBOUND;
			ub[i]=+CPX_INFBOUND;
		}
		else //beta part
		{
			lb[i]=0.0;
			ub[i]=1.0;
		}
	}

```
To something like this 
```
for (int32_t i=0; i<num_dim; i++) //xi part
	{
		obj[i]=0;
		lb[i]=-CPX_INFBOUND;
		ub[i]=+CPX_INFBOUND;
	}
for (int32_t i=num_dim ; i<num_variables; i++) //beta part
	{
		obj[i]=0;
		lb[i]=0.0;
		ub[i]=1.0;
	}
```
I'm assuming that num_dim is a subset of num_variables, and by making two seperate for blocks we avoid making comparisons on the second time around which improves speed.

karlnapf:
Probably :) Seem reasonably what you are saying. But numbers are better....
The way to do this would be to write a little benchmark on a reasonable problem (see `benchmarks` folder how to use the framework for that), and then try before and after, make sure the results are the same and execution time is faster. Wanna go ahead? :)

voldemoriarty:
http://quick-bench.com/TqRobxA8AGtgG8bf6gZhSjbJXq4

karlnapf:
amazing, let's have a PR with this, seems very useful :)

voldemoriarty:
can also use `std::fill`, produces equivalent performance and can also be parallelized in C++17 if the compiler supports it (MSVC does afaik)

http://quick-bench.com/w0nEo51PIU3QEivWmUWvqyiLgPU

karlnapf:
Yes that’s preferred actually
Also we just moved to 17 yesterday

Arpit2601:
Is this issue still open ?
I would like to work on it.

PR #4779, opened 2019-10-06
Performance Improvement in mathematics/Cplex.cpp #4601

Arpit2601:


karlnapf:
It would be really nice if we could merge this, but we need to ensure we have some test coverage that would catch the case where this would change the results of the code. I can't oversee whether it is correct from just looking at it
@Arpit2601 you are keen to have a go at that?

PR #4780, opened 2019-10-21
matern kernel

gf712:
This kernel will be useful for Bayesian optimisation.

### DIFF
--- a/src/shogun/kernel/MaternKernel.cpp
+++ b/src/shogun/kernel/MaternKernel.cpp
@@ -6,61 +6,52 @@
 
 #include <shogun/kernel/MaternKernel.h>
 
-#include <cmath>
+using namespace shogun;
 
-CMaternKernel::CMaternKernel(): CDistanceKernel(), nu(0.0)
+MaternKernel::MaternKernel(): DistanceKernel(), nu(1.5)
 {
 	init();
 }
 
-CMaternKernel::CMaternKernel(int32_t size, float64_t w, float64_t order, CDistance* dist): 
-	CDistanceKernel(size, w, dist), nu(order)
+MaternKernel::MaternKernel(int32_t size, float64_t w, float64_t order, const std::shared_ptr<Distance>& dist): 
+	DistanceKernel(size, w, dist), nu(order)
 {
-	ASSERT(distance)
-	SG_REF(distance);
 	init();
 }
 
-CMaternKernel::~CMaternKernel()
-{
-	cleanup();
-	SG_UNREF(distance);
-}
-
-void CMaternKernel::cleanup()
+MaternKernel::~MaternKernel()
 {
 }
 
-bool CMaternKernel::init(CFeatures* l, CFeatures* r)
+bool MaternKernel::init(std::shared_ptr<Features> l, std::shared_ptr<Features> r)
 {
 	ASSERT(distance)
-	CDistanceKernel::init(l,r);
+	DistanceKernel::init(l,r);
 	distance->init(l,r);
 	return init_normalizer();
 }
 
-void CMaternKernel::init()
+void MaternKernel::init()
 {
-	nu = 1.5;
-
 	SG_ADD(&nu, "nu", "order of the Bessel function of the second kind", ParameterProperties::HYPER)
 }
 
-float64_t CMaternKernel::compute(int32_t idx_a, int32_t idx_b)
+float64_t MaternKernel::compute(int32_t idx_a, int32_t idx_b)
 {
 	float64_t result;
 
 	float64_t dist = distance->distance(idx_a, idx_b);
 
 	// first we can check if we should use one of the approximations which are cheaper to calculate
-	if (CMath::fequals(nu, 0.5, std::numeric_limits<float64_t>::epsilon()))
+	// these formulas were taken from Rasmussen's GPs for ML
+	if (Math::fequals(nu, 0.5, std::numeric_limits<float64_t>::epsilon()))
 		result = std::exp(-dist/width);
-	else if (CMath::fequals(nu, 1.5, std::numeric_limits<float64_t>::epsilon()))
+	else if (Math::fequals(nu, 1.5, std::numeric_limits<float64_t>::epsilon()))
 	{
 		float64_t ratio = (std::sqrt(3) * dist) / width;
 		result = (1 + ratio) * std::exp(-ratio);
 	}
-	else if (CMath::fequals(nu, 2.5, std::numeric_limits<float64_t>::epsilon()))
+	else if (Math::fequals(nu, 2.5, std::numeric_limits<float64_t>::epsilon()))
 	{
 		float64_t ratio = (std::sqrt(5) * dist) / width;
 		result = (1 + ratio + ((5 * std::pow(dist, 2)) / (3 * std::pow(width, 2)))) * std::exp(-ratio);
@@ -78,8 +69,81 @@ float64_t CMaternKernel::compute(int32_t idx_a, int32_t idx_b)
 #else
 		float64_t bessel = std::cyl_neumann(nu, ratio);
 #endif
-		result = std::pow(2, 1 - nu) / std::tgamma(nu) * std::pow(ratio, nu) * bessel;
+		result = (std::pow(2, 1 - nu) / std::tgamma(nu)) * std::pow(ratio, nu) * bessel;
 	}
 
 	return result;
+}
+
+
+SGMatrix<float64_t> MaternKernel::get_parameter_gradient(Parameters::const_reference param, index_t index)
+{
+	require(lhs, "Left hand side features must be set!");
+	require(rhs, "Right hand side features must be set!");
+
+	if (param.first == "width")
+	{
+		SGMatrix<float64_t> derivative=SGMatrix<float64_t>(num_lhs, num_rhs);
+		std::function<float64_t(float64_t)> gradient_func;
+		
+		// the gradients Matern wrt width were computed with WolframAlpha
+		if (Math::fequals(nu, 0.5, std::numeric_limits<float64_t>::epsilon()))
+		{
+			const float64_t width_squared = std::pow(width, 2);
+			gradient_func = [&](const float64_t& dist){ 
+				return (dist * std::exp(-dist / width)) / width_squared;
+			};
+		}
+		else if (Math::fequals(nu, 1.5, std::numeric_limits<float64_t>::epsilon()))
+		{
+			const float64_t width_cubed = std::pow(width, 3);
+			gradient_func = [&](const float64_t& dist){ 
+					return (3 * std::pow(dist, 2) * std::exp(-(std::sqrt(3) * dist) / width)) / width_cubed; 
+			};
+		}
+		else if (Math::fequals(nu, 2.5, std::numeric_limits<float64_t>::epsilon()))
+		{
+			const float64_t width_power_4 = std::pow(width, 4);
+			gradient_func = [&](const float64_t& dist){ 
+				return (5 * std::pow(dist, 2) * std::exp(-(std::sqrt(5) * dist) / width) * (width + std::sqrt(5) * dist)) 
+						/ (3 * width_power_4);
+			};
+		}
+		else
+		{
+			gradient_func = [&](const float64_t& dist){
+				const auto ratio = (std::sqrt(2) * dist * std::sqrt(nu)) / width;
+#ifdef __clang__
+				auto bessel_nu_minus_1 = yn(nu - 1, ratio);
+				auto bessel_nu_plus_1 = yn(nu + 1, ratio);
+#else
+				auto bessel_nu_minus_1 = std::cyl_neumann(nu - 1, ratio);
+				auto bessel_nu_plus_1 = std::cyl_neumann(nu + 1, ratio);
+#endif
+				auto gradient = std::pow(2, (3 * nu) / 2 - 3 / 2);
+				gradient *= std::sqrt(nu) * dist;
+				gradient *= std::pow((std::sqrt(nu) * dist) / width, nu);
+				gradient *= bessel_nu_minus_1 - bessel_nu_plus_1;
+				gradient /= std::pow(width, 2) * std::tgamma(nu);
+
+				return gradient;
+			};
+		}
+
+		for (int k=0; k<num_rhs; k++)
+		{
+#pragma omp parallel for
+			for (int j=0; j<num_lhs; j++)
+			{
+				const float64_t element = compute(j, k);
+				derivative(j, k) = gradient_func(element);
+			}
+		}
+		return derivative;
+	}
+	else
+	{
+		error("Can't compute derivative wrt {} parameter", param.first);
+		return SGMatrix<float64_t>();
+	}
 }
--- a/src/shogun/kernel/MaternKernel.cpp
+++ b/src/shogun/kernel/MaternKernel.cpp
@@ -4,139 +4,188 @@
  * Authors: Gil Hoben
  */
 
+#include <shogun/distance/EuclideanDistance.h>
 #include <shogun/kernel/MaternKernel.h>
 
 using namespace shogun;
 
-MaternKernel::MaternKernel(): DistanceKernel(), nu(1.5)
+MaternKernel::MaternKernel() : ShiftInvariantKernel()
 {
-	init();
+	auto distance = std::make_shared<EuclideanDistance>();
+	distance->set_disable_sqrt(false);
+	m_distance = distance;
+	SG_ADD(
+	    &m_nu, "nu", "order of the modified Bessel function of the second kind",
+	    ParameterProperties::HYPER)
+	SG_ADD(
+	    &m_width, "width", "Kernel scale",
+	    ParameterProperties::HYPER | ParameterProperties::GRADIENT)
 }
 
-MaternKernel::MaternKernel(int32_t size, float64_t w, float64_t order, const std::shared_ptr<Distance>& dist): 
-	DistanceKernel(size, w, dist), nu(order)
+MaternKernel::MaternKernel(int32_t size, float64_t w, float64_t order)
+    : MaternKernel()
 {
-	init();
+	m_nu = order;
+	m_width = w;
+	set_cache_size(size);
+}
+
+MaternKernel::MaternKernel(float64_t w, float64_t order)
+    : MaternKernel(10, w, order)
+{
+}
+
+MaternKernel::MaternKernel(
+    const std::shared_ptr<Features>& l, const std::shared_ptr<Features>& r,
+    int32_t size, float64_t w, float64_t order)
+    : MaternKernel(size, w, order)
+{
+	init(l, r);
 }
 
 MaternKernel::~MaternKernel()
 {
+	cleanup();
 }
 
-bool MaternKernel::init(std::shared_ptr<Features> l, std::shared_ptr<Features> r)
+void MaternKernel::cleanup()
 {
-	ASSERT(distance)
-	DistanceKernel::init(l,r);
-	distance->init(l,r);
-	return init_normalizer();
+	Kernel::cleanup();
+	m_distance->cleanup();
 }
 
-void MaternKernel::init()
+bool MaternKernel::init(
+    std::shared_ptr<Features> l, std::shared_ptr<Features> r)
 {
-	SG_ADD(&nu, "nu", "order of the Bessel function of the second kind", ParameterProperties::HYPER)
+	cleanup();
+	return ShiftInvariantKernel::init(l, r);
 }
 
 float64_t MaternKernel::compute(int32_t idx_a, int32_t idx_b)
 {
 	float64_t result;
+	const float64_t dist = ShiftInvariantKernel::distance(idx_a, idx_b);
 
-	float64_t dist = distance->distance(idx_a, idx_b);
-
-	// first we can check if we should use one of the approximations which are cheaper to calculate
-	// these formulas were taken from Rasmussen's GPs for ML
-	if (Math::fequals(nu, 0.5, std::numeric_limits<float64_t>::epsilon()))
-		result = std::exp(-dist/width);
-	else if (Math::fequals(nu, 1.5, std::numeric_limits<float64_t>::epsilon()))
+	// first we can check if we should use one of the approximations which are
+	// cheaper to calculate these formulas were taken from Rasmussen's GPs for
+	// ML
+	if (Math::fequals(m_nu, 0.5, std::numeric_limits<float64_t>::epsilon()))
+		result = std::exp(-dist / m_width);
+	else if (Math::fequals(
+	             m_nu, 1.5, std::numeric_limits<float64_t>::epsilon()))
 	{
-		float64_t ratio = (std::sqrt(3) * dist) / width;
+		float64_t ratio = (std::sqrt(3) * dist) / m_width;
 		result = (1 + ratio) * std::exp(-ratio);
 	}
-	else if (Math::fequals(nu, 2.5, std::numeric_limits<float64_t>::epsilon()))
+	else if (Math::fequals(
+	             m_nu, 2.5, std::numeric_limits<float64_t>::epsilon()))
 	{
-		float64_t ratio = (std::sqrt(5) * dist) / width;
-		result = (1 + ratio + ((5 * std::pow(dist, 2)) / (3 * std::pow(width, 2)))) * std::exp(-ratio);
+		float64_t ratio = (std::sqrt(5) * dist) / m_width;
+		result = (1 + ratio +
+		          ((5 * std::pow(dist, 2)) / (3 * std::pow(m_width, 2)))) *
+		         std::exp(-ratio);
 	}
 	// if none of the above was true we compute the value the expensive way
-	// in theory if nu is large we could calculate this with the squared exponential covariance function
+	// in theory if nu is large we could calculate this with the squared
+	// exponential covariance function
 	else
 	{
-		float64_t ratio = std::sqrt(2 * nu) * dist / width;
-		// bessel function of the second kind
-#ifdef __clang__
-		// clang hasn't implemented mathematical special functions in the C++17 standard at this point
-		// note that yn casts nu to an integer
-		float64_t bessel = yn(nu, ratio);
+#ifdef _LIBCPP_VERSION
+		error(
+		    "Shogun compiled with libc++ does not support Matern kernels with "
+		    "width other than 0.5, 1.5 and 2.5. The current width is {}.",
+		    m_width);
 #else
-		float64_t bessel = std::cyl_neumann(nu, ratio);
+		const float64_t adjusted_dist =
+		    Math::fequals(dist, 0.0, std::numeric_limits<float64_t>::epsilon())
+		        ? std::numeric_limits<float32_t>::epsilon()
+		        : dist;
+		float64_t ratio = std::sqrt(2 * m_nu) * adjusted_dist / m_width;
+		// bessel function of the second kind
+		float64_t bessel = std::cyl_bessel_k(m_nu, ratio);
+		result = (std::pow(2, 1 - m_nu) / std::tgamma(m_nu)) *
+		         std::pow(ratio, m_nu) * bessel;
 #endif
-		result = (std::pow(2, 1 - nu) / std::tgamma(nu)) * std::pow(ratio, nu) * bessel;
 	}
 
 	return result;
 }
 
-
-SGMatrix<float64_t> MaternKernel::get_parameter_gradient(Parameters::const_reference param, index_t index)
+SGMatrix<float64_t> MaternKernel::get_parameter_gradient(
+    Parameters::const_reference param, index_t index)
 {
 	require(lhs, "Left hand side features must be set!");
 	require(rhs, "Right hand side features must be set!");
 
 	if (param.first == "width")
 	{
-		SGMatrix<float64_t> derivative=SGMatrix<float64_t>(num_lhs, num_rhs);
+		SGMatrix<float64_t> derivative = SGMatrix<float64_t>(num_lhs, num_rhs);
 		std::function<float64_t(float64_t)> gradient_func;
-		
-		// the gradients Matern wrt width were computed with WolframAlpha
-		if (Math::fequals(nu, 0.5, std::numeric_limits<float64_t>::epsilon()))
+
+		// the gradients Matern wrt m_width were computed with WolframAlpha
+		if (Math::fequals(m_nu, 0.5, std::numeric_limits<float64_t>::epsilon()))
 		{
-			const float64_t width_squared = std::pow(width, 2);
-			gradient_func = [&](const float64_t& dist){ 
-				return (dist * std::exp(-dist / width)) / width_squared;
+			const float64_t m_width_squared = std::pow(m_width, 2);
+			gradient_func = [&](const float64_t& dist) {
+				return (dist * std::exp(-dist / m_width)) / m_width_squared;
 			};
 		}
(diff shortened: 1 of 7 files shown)


karlnapf (commented):
gradients will be necessary for tuning hyperparameters in BO :)

karlnapf:
whats needed to merge this?

gf712:
> whats needed to merge this?

Gradients and unit tests :D

gf712:
and general clean up. I'll see if I have some time later this week.

src/shogun/kernel/MaternKernel.cpp
```diff
@@ -0,0 +1,85 @@
...
+}
+
+float64_t CMaternKernel::compute(int32_t idx_a, int32_t idx_b)
+{
+	float64_t result;
+
+	float64_t dist = distance->distance(idx_a, idx_b);
+
+	// first we can check if we should use one of the approximations which are cheaper to calculate
+	if (CMath::fequals(nu, 0.5, std::numeric_limits<float64_t>::epsilon()))
```
gf712: @vigsterkr do we have this somewhere else in the library now?

src/shogun/kernel/MaternKernel.cpp
```diff
@@ -0,0 +1,85 @@
...
+	// if none of the above was true we compute the value the expensive way
+	// in theory if nu is large we could calculate this with the squared exponential covariance function
+	else
+	{
+		float64_t ratio = std::sqrt(2 * nu) * dist / width;
+		// bessel function of the second kind
+#ifdef __clang__
+		// clang hasn't implemented mathematical special functions in the C++17 standard at this point
+		// note that yn casts nu to an integer
+		float64_t bessel = yn(nu, ratio);
```
gf712: need to check if new clang has this implemented

src/shogun/kernel/MaternKernel.cpp
```diff
@@ -0,0 +1,85 @@
...
+ * This software is distributed under BSD 3-clause license (see LICENSE file).
+ *
+ * Authors: Gil Hoben
+ */
+
+#include <shogun/kernel/MaternKernel.h>
+
+#include <cmath>
+
+CMaternKernel::CMaternKernel(): CDistanceKernel(), nu(0.0)
```
gf712: move the default value to class member definition

src/shogun/kernel/MaternKernel.h
```diff
@@ -0,0 +1,72 @@
...
+			 *
+			 * @param size cache size
+			 * @param width the kernel width
+			 * @param order the order of the bessel function of the second kind
+			 * @param dist distance to be used
+			 */
+			CMaternKernel(int32_t size, float64_t width, float64_t order, CDistance* dist);
+
+			/** destructor */
+			~CMaternKernel();
```
gf712: can be default, and needs override

src/shogun/kernel/MaternKernel.h
```diff
@@ -0,0 +1,72 @@
...
+
+			/** cleanup instance */
+			void cleanup();
+
+			/** return the kernel's name
+			 *
+			 * @return name
+			 */
+			virtual const char* get_name() const
```
gf712: override

src/shogun/kernel/MaternKernel.cpp
```diff
@@ -0,0 +1,85 @@
...
+{
+	ASSERT(distance)
+	CDistanceKernel::init(l,r);
+	distance->init(l,r);
+	return init_normalizer();
+}
+
+void CMaternKernel::init()
+{
+	nu = 1.5;
```
gf712: move default to declaration

src/shogun/kernel/MaternKernel.h
```diff
@@ -0,0 +1,72 @@
...
+			/** initialize kernel
+			 *
+			 * @param l features of left-hand side
+			 * @param r features of right-hand side
+			 * @return if initializing was successful
+			 */
+			virtual bool init(CFeatures* l, CFeatures* r);
+
+		protected:
+			float64_t compute(int32_t idx_a, int32_t idx_b);
```
gf712: override

src/shogun/kernel/MaternKernel.h
```diff
@@ -0,0 +1,72 @@
...
+				return "MaternKernel";
+			}
+
+			/** initialize kernel
+			 *
+			 * @param l features of left-hand side
+			 * @param r features of right-hand side
+			 * @return if initializing was successful
+			 */
+			virtual bool init(CFeatures* l, CFeatures* r);
```
gf712: override

src/shogun/kernel/MaternKernel.cpp
```diff
@@ -0,0 +1,85 @@
...
+	SG_UNREF(distance);
+}
+
+void CMaternKernel::cleanup()
+{
+}
+
+bool CMaternKernel::init(CFeatures* l, CFeatures* r)
+{
+	ASSERT(distance)
```
gf712: require?

src/shogun/kernel/MaternKernel.cpp
```diff
@@ -0,0 +1,85 @@
...
+	SG_REF(distance);
+	init();
+}
+
+CMaternKernel::~CMaternKernel()
+{
+	cleanup();
+	SG_UNREF(distance);
```
gf712: use automatic mem management

src/shogun/kernel/MaternKernel.cpp
```diff
@@ -0,0 +1,149 @@
...
+				auto bessel_nu_plus_1 = std::cyl_neumann(nu + 1, ratio);
+#endif
+				auto gradient = std::pow(2, (3 * nu) / 2 - 3 / 2);
+				gradient *= std::sqrt(nu) * dist;
+				gradient *= std::pow((std::sqrt(nu) * dist) / width, nu);
+				gradient *= bessel_nu_minus_1 - bessel_nu_plus_1;
+				gradient /= std::pow(width, 2) * std::tgamma(nu);
+
+				return gradient;
+			};
```
gf712, line 172: @karlnapf as you can see the gradients are pretty complex... I used WolframAlpha to compute them, but would like to test them.. Is there a numeric gradient estimation in shogun I could use?

src/shogun/kernel/MaternKernel.cpp
gf712: @karlnapf can we copy the implementation of bessel function of the second kind from https://www.boost.org/doc/libs/1_56_0/boost/math/special_functions/bessel.hpp? The problem is that clang doesn't have the implementation, even though it is part of the C++17 standard.

src/shogun/kernel/MaternKernel.cpp
karlnapf: I vaguely remember that we had something for that somewhere ....

src/shogun/kernel/MaternKernel.cpp
karlnapf: yep, just copy paste it somewhere appropriate (check license)

src/shogun/kernel/MaternKernel.cpp
karlnapf: i dont think atm

src/shogun/kernel/MaternKernel.cpp
karlnapf: actually, why don't you implement the kernel in jax and generate some simple test cases in there (maybe in a gist that could be put as a comment in the unit test)

gf712: I can check for matern 1/2, 3/2 and 5/2, but for the arbitrary order there is no way with JAX, as it hasn't implemented bessel of the second kind.. I'll have a look if I can test somewhere else.
Would be good to check the gradients though, like in sklearn unittests for kernels

karlnapf: yeah checking for nontrivial cases already is useful.

src/shogun/kernel/MaternKernel.h
```diff
@@ -0,0 +1,78 @@
...
+	 * It is defined as
+	 * \f[
+	 * k(x, y) = \frac{2 ^ {\nu - 1}}{\Gamma(\nu)}\Bigg(\sqrt{2\nu}\frac{D(x, y)}{\rho}\Bigg)^\nu + N_{\nu}\Bigg(\sqrt{2\nu}\frac{D(x, y)}{\rho}\Bigg)
+	 * \f]
+	 * Where:
+	 * 		\f$N_{\nu}\f$ is the Bessel function of the second kind with order \f$\nu\f$,
+	 *		\f$\rho\f$ is the kernel width, and
+	 *		\f$\Gamma\f$ is the gamma function
+	 **/
+	class MaternKernel: public DistanceKernel
```
gf712: @karlnapf do you know from what class I should inherit from?

src/shogun/kernel/MaternKernel.cpp
karlnapf: I think the finite diff stuff is somewhere in the nn code

src/shogun/kernel/MaternKernel.h
karlnapf: translation invariant kernel?

gf712: yea, thought so, will change it

tests/unit/kernel/MaternKernel_unittest.cc
```diff
@@ -0,0 +1,142 @@
...
+	auto kernel = std::make_shared<MaternKernel>(features, features, 10, 2.0, 2.5);
+	SGMatrix<float64_t> computed_kernel_matrix = kernel->get_kernel_matrix();
+
+	for (const auto& [result, reference]: zip_iterator(computed_kernel_matrix, expected_kernel_matrix))
+		EXPECT_NEAR(result, reference, 1E-7);		
+}
+
+TEST(MaternKernelTest_1_2, test_derivative_width)
+{
+	// gradient computed with JAX: https://gist.github.com/gf712/b10ecefd2379e8492c9802ac8cb329c9
```
gf712: @karlnapf getting the same results as JAX but different from sklearn..

karlnapf: ha! Is the kernel maybe parametrized differently?

src/shogun/kernel/MaternKernel.cpp
```diff
@@ -0,0 +1,170 @@
...
+using namespace shogun;
+
+
+MaternKernel::MaternKernel(): ShiftInvariantKernel()
+{
+	register_params();
+}
+
+MaternKernel::MaternKernel(int32_t size, float64_t w, float64_t order): 
+	ShiftInvariantKernel(), m_nu(order), m_width(w)
```
karlnapf: could you call `MaternKernel` here to avoid having to call register params again?

src/shogun/kernel/MaternKernel.cpp
```diff
@@ -0,0 +1,170 @@
...
+
+#include <shogun/distance/EuclideanDistance.h>
+#include <shogun/kernel/MaternKernel.h>
+
+using namespace shogun;
+
+
+MaternKernel::MaternKernel(): ShiftInvariantKernel()
+{
+	register_params();
```
karlnapf: I think with initializing parameters in the declaration and calling the base ctor in the other ones, we might not need these methods anymore, could just register directly here?

src/shogun/kernel/MaternKernel.cpp
```diff
@@ -0,0 +1,170 @@
...
+}
+
+float64_t MaternKernel::compute(int32_t idx_a, int32_t idx_b)
+{
+	float64_t result;
+	const float64_t dist = ShiftInvariantKernel::distance(idx_a, idx_b);
+
+	// first we can check if we should use one of the approximations which are cheaper to calculate
+	// these formulas were taken from Rasmussen's GPs for ML
+	if (Math::fequals(m_nu, 0.5, std::numeric_limits<float64_t>::epsilon()))
```
karlnapf, line 66: maybe we should add `allclose` methods to linalg to get rid of this CMath thing?

src/shogun/kernel/MaternKernel.cpp
```diff
@@ -0,0 +1,170 @@
...
+		result = std::exp(-dist/m_width);
+	else if (Math::fequals(m_nu, 1.5, std::numeric_limits<float64_t>::epsilon()))
+	{
+		float64_t ratio = (std::sqrt(3) * dist) / m_width;
```
karlnapf: is something like sqrt(3) constexpr? :)

tests/unit/kernel/MaternKernel_unittest.cc
```diff
@@ -0,0 +1,142 @@
...
+
+	auto features = std::make_shared<DenseFeatures<float64_t>>(matrix);
+
+	auto kernel = std::make_shared<MaternKernel>(features, features, 10, 2.0, 0.5);
+	auto params = kernel->get_params();
+	auto width = params.find("width");
+	SGMatrix<float64_t> dMatrix = kernel->get_parameter_gradient(*width);
+
+	for (const auto& [result, reference]: zip_iterator(dMatrix, expected_dmatrix))
+		EXPECT_NEAR(result, reference, 1E-7);
```
karlnapf: single precision?

src/shogun/kernel/MaternKernel.cpp
gf712: I think all compilers would precompute this with optimisations on

tests/unit/kernel/MaternKernel_unittest.cc
gf712: I’m not sure.. the kernel matrix is the same, it’s just the derivative that is different

src/shogun/kernel/MaternKernel.cpp
gf712: Yup, need to clean this up

src/shogun/kernel/MaternKernel.cpp
gf712: Yes, I can add that in a separate PR and merge that

tests/unit/kernel/MaternKernel_unittest.cc
gf712: Could you actually check the gist please? I think it’s correct.. kernel wrt to width right?

src/shogun/kernel/MaternKernel.cpp
```diff
@@ -0,0 +1,170 @@
...
+	return ShiftInvariantKernel::init(l,r);
+}
+
+void MaternKernel::register_params()
+{
+	auto distance=std::make_shared<EuclideanDistance>();
+	distance->set_disable_sqrt(false);
+	m_distance=distance;
+	SG_ADD(&m_nu, "nu", "order of the Bessel function of the second kind", ParameterProperties::HYPER)
+	SG_ADD(&m_width, "width", "Kernel scale", ParameterProperties::HYPER)
```
gf712: Also GRADIENT

tests/unit/kernel/MaternKernel_unittest.cc
gf712: You mean get epsilon from type traits?

tests/unit/kernel/MaternKernel_unittest.cc
gf712: It seems like the sklearn gradient is exactly double of what I’m getting..

gf712: So sklearn overestimates the gradient by a factor of `width`. I tested it numerically, and the JAX/WolframAlpha gradients are correct. I wonder why.. I suspect it is related to this distance calculation
https://github.com/scikit-learn/scikit-learn/blob/95d4f0841d57e8b5f6b2a570312e9d832e69debc/sklearn/gaussian_process/kernels.py#L1437-L1438

src/shogun/kernel/MaternKernel.cpp
```diff
@@ -0,0 +1,197 @@
...
+					        std::numeric_limits<float64_t>::epsilon())
+					        ? std::numeric_limits<float32_t>::epsilon()
+					        : dist;
+					float64_t ratio =
+					    std::sqrt(2 * m_nu) * adjusted_dist / width;
+					float64_t bessel = std::cyl_bessel_k(m_nu, ratio);
+					return (std::pow(2, 1 - m_nu) / std::tgamma(m_nu)) *
+					       std::pow(ratio, m_nu) * bessel;
+				};
+				return (k(m_width + epsilon) - k(m_width)) / epsilon;
```
gf712: @karlnapf the derivative of the matern kernel is a monster (I also think it is quite unstable outside the widths that are usually used, as I was getting very different result between WolframAlpha gradient and the numerical gradient. So I just put the numerical gradient (like sklearn does)

src/shogun/kernel/MaternKernel.cpp
```diff
@@ -0,0 +1,197 @@
...
+			};
+		}
+		else
+		{
+#ifdef __clang__
+			error(
+			    "Shogun compiled on clang does not support Matern kernels with "
+			    "width other than 0.5, 1.5 and 2.5. The current width is {}",
+			    m_width);
+#endif
```
gf712: @karlnapf I couldn't find the bessel function I need here (`kv` in scipy) so I think we should just wait until clang implements it. It should also be possible to compile this with clang with libstdc++, rather than libc++

src/shogun/kernel/MaternKernel.cpp
```diff
@@ -0,0 +1,197 @@
...
+		    Math::fequals(dist, 0.0, std::numeric_limits<float64_t>::epsilon())
+		        ? std::numeric_limits<float32_t>::epsilon()
+		        : dist;
+		float64_t ratio = std::sqrt(2 * m_nu) * adjusted_dist / m_width;
+		// bessel function of the second kind
+#ifdef _LIBCPP_VERSION
+		error(
+		    "Shogun compiled with libc++ does not support Matern kernels with "
+		    "width other than 0.5, 1.5 and 2.5. The current width is {}.",
+		    m_width);
```
gf712: it's too complicated to add support for cyl_bessel_k, so will just throw an error when running this part of the code when compiled with libc++. Note that realistically no one would ever use a value other than 0.5, 1.5 and 2.5..

src/shogun/kernel/MaternKernel.cpp
gf712: for reference https://www.wolframalpha.com/input/?i=derivative+(2^(1+-+nu)+%2F+Gamma(nu))+*+(sqrt(2+*+nu)+*+d+%2F+w)+^+nu+*+BesselK(nu%2C+sqrt(2+*+nu)+*+d+%2F+w)+wrt+w

src/shogun/kernel/MaternKernel.cpp
karlnapf: i see! cool!

gf712: see line 2 of the assembly in https://godbolt.org/z/xCXF3q. When optimisations are turned off this becomes a runtime call to sqrt

tests/unit/kernel/MaternKernel_unittest.cc
karlnapf: If that is possible I think it should be done or?

src/shogun/kernel/MaternKernel.cpp
karlnapf: ok!

karlnapf: maybe add some comments in the code so that this reasoning is transparent in case someone touches it later

src/shogun/kernel/MaternKernel.cpp
karlnapf: problem will be tests / examples when compiled without clang, but ok

tests/unit/kernel/MaternKernel_unittest.cc
karlnapf: maybe it has to do with the fact how they use the gradients downstream

gf712: maybe, maybe it also doesn't matter too much as long as they are in correct within the same factor.. Anyway, this should be correct

tests/unit/kernel/MaternKernel_unittest.cc
gf712: I am not sure, because the error propagates through each calculation. This is the tolerance value that got everything correct for all the tests without tweaking for each fixture

src/shogun/kernel/MaternKernel.cpp
gf712: sure! I think the main point is that It is more expensive to compute the gradient analytically than numerically. As for the value of the delta in the gradient calculation, it was what gave me the correct values for the width values that we compute analytically (i.e. I computed the 1/2 kernel gradient numerically and this delta gave a good result). I think it might be a bit high though?

src/shogun/kernel/MaternKernel.cpp
gf712: I guess, but the option is adding some fortran code from http://netlib.org/amos/

gf712: I can try and find out how stdlibc++ does it actually

karlnapf: good idea. Not so sure about Fortran code, never did that. Is it easy?

gf712: I copied the libstdc++ code :D

src/shogun/mathematics/bessel.cpp
```diff
@@ -0,0 +1,290 @@
+// Special functions -*- C++ -*-
+
+// Copyright (C) 2006-2018 Free Software Foundation, Inc.
+//
+// This file is part of the GNU ISO C++ Library.  This library is free
+// software; you can redistribute it and/or modify it under the
+// terms of the GNU General Public License as published by the
```
karlnapf: sadly, we cannot do that. It would need to live in shogun-gpl. But let's not add to that. Is there maybe a version in free-domain?

karlnapf: (gpl is viral ... it "infects" our bsd repository with gpl and then our repo also must be gpl)

gf712: OK, I'll keep looking!

gf712: is the boost license compatible? I'll try and rewrite their code into something that can work standalone

karlnapf: It is, but then we would need to include the boost license in our license .... gna

karlnapf: > Permission is hereby granted, free of charge, to any person or organization
obtaining a copy of the software and accompanying documentation covered by
this license (the "Software") to use, reproduce, display, distribute,
execute, and transmit the Software, and to prepare derivative works of the
Software, and to permit third-parties to whom the Software is furnished to
do so, all subject to the following:
> The copyright notices in the Software and this entire statement, including
the above license grant, this restriction and the following disclaimer,
must be included in all copies of the Software, in whole or in part, and
all derivative works of the Software, unless such copies or derivative
works are solely in the form of machine-executable object code generated by
a source language processor.

gf712: I think in that case we should just not have it for libc++. It is quite time consuming to get another implementation, and realistically we will not need it any time soon (this kernel is probably just going to be used for BO)

gf712: I was looking at how scipy does it, but then we need to add a bunch of fortran code and then translate the scipy C code to something for C++ types. There is a C translation of the main routines but it is but seems very innefficient

karlnapf: yeah ... I had similar issues a while ago when using pdf/cdf functions of the Gamma distribution. I eventually found a public domain version of some old-school code, a packet called cdflib `lib/external/cdflib.cpp`

karlnapf: it is crazy code, full of goto statements ;)

gf712: this is what I found https://github.com/cran/Bessel/blob/master/src/zbsubs.c. It's like reading Fortran written in C.. It can't be very efficient with so many jumps :p

src/shogun/kernel/MaternKernel.cpp
```diff
@@ -0,0 +1,198 @@
...
+
+#include <shogun/distance/EuclideanDistance.h>
+#include <shogun/kernel/MaternKernel.h>
+
+using namespace shogun;
+
+MaternKernel::MaternKernel() : ShiftInvariantKernel()
+{
+	auto distance = std::make_shared<EuclideanDistance>();
+	distance->set_disable_sqrt(false);
```
gf712, line 16: @karlnapf I turned this off, but I guess it could be true for the sake of efficiency?

karlnapf: what is that for again?

src/shogun/mathematics/bessel.cpp
karlnapf: use it :)

src/shogun/kernel/MaternKernel.cpp
gf712: It’s to avoid computing the square root of the Euclidean distance

src/shogun/mathematics/bessel.cpp
gf712: OK, I'll have some time during the weekend to clean it up, as this is written to work with Rcpp header files

src/shogun/mathematics/zbsubs.h
```diff
@@ -0,0 +1,110 @@
+
```
gf712, line 6: @karlnapf I am not sure what license to put here? This is mostly copied with some slight modifications to get it to compile without any additional libraries

gf712: there was no license in the repo where I copied this from (probably because it is just C code translated directly from FORTRAN)

karlnapf: It is probably public domain then. I suggest you try a bit to track the license down and if you cannot find anything State the source and that you modified it and say that it’s taken under public domain

gf712: it seems like they went for GPL... http://r-forge.r-project.org/projects/specfun/
I can just do the f2c myself for the zbsubs routines?

karlnapf: Gpl is non copy pastable
You can do them yourself but not sure how much work that would be?

gf712: I translated it with f2c, it was some effort, but it should be working fine now

src/shogun/mathematics/bessel.h
```diff
@@ -0,0 +1,59 @@
...
+			{
+				error("Overflow");
+			}
+			else if (ierr == 4)
+			{
+				error("|x|={} or nu={} too large", zr, nu);
+			}
+			else
+			{
+				error("Unexpected error {} returned from zbesk", ierr);
```
gf712: These error messages are similar to the R package, but I assume that is not an issue?

karlnapf: I think if there is any chance that a user will face those. They should be improved a tiny bit

src/shogun/mathematics/zbsubs.h
karlnapf: nice one!

src/shogun/mathematics/bessel.h
gf712: It is possible, I never had any issues, but I guess if you're using this kernel on any data something can go wrong..

gf712:
@karlnapf I think this is ready to merge now (just waiting for CI)

gf712:
@karlnapf shall I merge this?

src/shogun/kernel/MaternKernel.h
```diff
@@ -0,0 +1,140 @@
...
+		MaternKernel(float64_t width, float64_t nu);
+
+		/** constructor
+		 *
+		 * @param size cache size
+		 * @param width the kernel width
+		 * @param nu the order of the bessel function of the second kind
+		 * @param dist distance to be used
+		 */
+		MaternKernel(int32_t size, float64_t width, float64_t nu);
```
karlnapf: I'd suggest to remove this one ... the cache is useless for this kernel, so the size doesnt need to be set

src/shogun/kernel/MaternKernel.h
```diff
@@ -0,0 +1,140 @@
...
+		 * @param l features of left-hand side
+		 * @param r features of right-hand side
+		 * @param size cache size
+		 * @param width the kernel width
+		 * @param nu the order of the bessel function of the second kind
+		 * @param dist distance to be used
+		 */
+		MaternKernel(
+		    const std::shared_ptr<Features>& l,
+		    const std::shared_ptr<Features>& r, int32_t size, float64_t width,
```
karlnapf: same here, I'd remove the size

src/shogun/mathematics/zbsubs.h
```diff
@@ -0,0 +1,116 @@
...
+    double* yi, int* nz, double* tol, double* elim, double* alim);
+int zbuni_(
+    double* zr, double* zi, double* fnu, int* kode, int* n, double* yr,
+    double* yi, int* nz, int* nui, int* nlast, double* fnul, double* tol,
+    double* elim, double* alim);
+int zuni1_(
+    double* zr, double* zi, double* fnu, int* kode, int* n, double* yr,
+    double* yi, int* nz, int* nlast, double* fnul, double* tol, double* elim,
+    double* alim);
+int zuni2_(
```
karlnapf, line 111: tasty :)

tests/unit/kernel/MaternKernel_unittest.cc
```diff
@@ -0,0 +1,190 @@
...
+	                                           {0.07441661, 1.}};
+
+	auto features = std::make_shared<DenseFeatures<float64_t>>(matrix);
+
+	auto kernel = std::make_shared<MaternKernel>(
+	    features, features, cache_size, width, nu);
+	SGMatrix<float64_t> computed_kernel_matrix = kernel->get_kernel_matrix();
+
+	for (const auto & [ result, reference ] :
+	     zip_iterator(computed_kernel_matrix, expected_kernel_matrix))
```
karlnapf, line 30: very nice this zip iterator!

karlnapf (approved):

tests/unit/kernel/MaternKernel_unittest.cc
```diff
@@ -0,0 +1,190 @@
...
+	auto features = std::make_shared<DenseFeatures<float64_t>>(matrix);
+
+	auto kernel = std::make_shared<MaternKernel>(
+	    features, features, cache_size, width, nu);
+	auto params = kernel->get_params();
+	auto width_param = params.find("width");
+	SGMatrix<float64_t> dMatrix = kernel->get_parameter_gradient(*width_param);
+
+	SGMatrix<float64_t> computed_kernel_matrix = kernel->get_kernel_matrix();
+	kernel->put("width", width + epsilon);
```
karlnapf: class member or c++ setter (as discussed on irc)

tests/unit/kernel/MaternKernel_unittest.cc
gf712: Indeed! One of these days I need to fix it to work with temporary values, so we can have
```cpp
for (const auto& [X, y]: zip_iterator(DotIterator(feat), label)) {
...
}
```

karlnapf: yes that'd be ace!

karlnapf: maybe make an issue for this? a student might pick it up?

PR #4784, opened 2019-10-23
seeding fix for xvalmmd

### LINKED ISSUE #4783: why do we create twice the CrossvalidationSplitting objects
https://github.com/shogun-toolbox/shogun/blob/40f40f08b31b6cc73851201a4e1adf0ae7c39698/src/shogun/statistical_testing/internals/mmd/CrossValidationMMD.h#L198-L206

vigsterkr:
fix #4783

### DIFF
--- a/src/shogun/statistical_testing/internals/mmd/CrossValidationMMD.h
+++ b/src/shogun/statistical_testing/internals/mmd/CrossValidationMMD.h
@@ -195,16 +195,11 @@ struct CrossValidationMMD : PermutationMMD
 		SGVector<int64_t> dummy_labels_x(m_n_x);
 		SGVector<int64_t> dummy_labels_y(m_n_y);
 
-		auto instance_x=new CCrossValidationSplitting(new CBinaryLabels(dummy_labels_x), m_num_folds);
-		auto instance_y=new CCrossValidationSplitting(new CBinaryLabels(dummy_labels_y), m_num_folds);
-		random::seed(instance_x, prng);
-		random::seed(instance_y, prng);
-
-		m_kfold_x=unique_ptr<CCrossValidationSplitting>(instance_x);
-		m_kfold_y=unique_ptr<CCrossValidationSplitting>(instance_y);
+		m_kfold_x=std::make_unique<CCrossValidationSplitting>(new CBinaryLabels(dummy_labels_x), m_num_folds);
+		m_kfold_y=std::make_unique<CCrossValidationSplitting>(new CBinaryLabels(dummy_labels_y), m_num_folds);
 		random::seed(m_kfold_x.get(), prng);
 		random::seed(m_kfold_y.get(), prng);
-	
+
 		m_stack=unique_ptr<CSubsetStack>(new CSubsetStack());
 
 		const index_t size=m_n_x+m_n_y;
--- a/tests/unit/statistical_testing/internals/CrossValidationMMD_unittest.cc
+++ b/tests/unit/statistical_testing/internals/CrossValidationMMD_unittest.cc
@@ -107,6 +107,7 @@ TEST(CrossValidationMMD, biased_full)
 	kfold_p->put("seed", seed);
 	kfold_q->put("seed", seed);
 
+	std::mt19937_64 permPRNG(seed);
 	auto permutation_mmd=PermutationMMD();
 	permutation_mmd.m_stype=stype;
 	permutation_mmd.m_num_null_samples=num_null_samples;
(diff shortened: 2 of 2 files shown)


vigsterkr:
@lambday this change currently breaks some mmd unit tests :(
would just simply updating to new values be sufficient?
only prng changes here...

lambday:
@vigsterkr for CrossValidationMMD unit-tests, the checks are more fundamental it seems, cannot just update to the new values. Is the internal SGObject also using `std::mt19937_64` prng?

vigsterkr:
yep it's the same prng all over

vigsterkr:
@lambday note that if i would just add there 2 dummy calls for the prng, before creating the folds objects, things would pass...

src/shogun/statistical_testing/internals/mmd/CrossValidationMMD.h
```diff
@@ -195,16 +195,11 @@ struct CrossValidationMMD : PermutationMMD
 		SGVector<int64_t> dummy_labels_x(m_n_x);
 		SGVector<int64_t> dummy_labels_y(m_n_y);
```
vigsterkr, line 197: @lambday i'm pretty sure that if i would do `prng(); prng();`
this patch would pass

vigsterkr:
@lambday in fact i've just tested it locally and the hack i've written actually makes all the unit tests pass that are currently failing... so this is just the problem of having the right expected value/seed

lambday:
@vigsterkr adding dummy calls to the main codebase might not be the ideal way to go here I think. Would it be possible to create two different prng instances in the unit-test rather? e.g. one instance to pass to the `CrossValidationMMD` guy [here](https://github.com/shogun-toolbox/shogun/blob/develop/tests/unit/statistical_testing/internals/CrossValidationMMD_unittest.cc#L99) and another one to pass to the explicit testing logic [here](https://github.com/shogun-toolbox/shogun/blob/develop/tests/unit/statistical_testing/internals/CrossValidationMMD_unittest.cc#L137). Both should have the same seed. I think that might solve the problem as well?

lambday:
For the other failing unit-test, we can just update to the new value that it gets.

vigsterkr:
Not so sure I understand the rational for this separate prng...?
And why not just to update the seed or expected value? I mean it does not use one global prng, but objects that are nested and have prngs there the prng is shared... but eg the data generators in the unit test and xvalmmd doesn’t share one prng...

lambday:
@vigsterkr for `KernelSelectionMaxCrossValidation` unit-tests, we can just update the expected value with the new value we're getting here. 

But it's a bit tricky for the `CrossValidationMMD` unit-test. This test does not use a constant as the expected value. Rather, it computes the expected value independently using `PermutationMMD` within the test itself, and compares that value with that we get from the xval class. 

Now, these two independent computations would yield the same result, _if_ we feed each of them the same class of prng with the same state. 

But as per the present logic, we are using the same prng instance to compute _one after another_ - these two computations are not independent anymore. This is why the these are not giving the same result I am afraid.

I think if we can just create two instances of the prng with the same seed in these unit-tests, pass one to the shogun `CrossValidationMMD` class and pass the other one to the explicit computation logic in the test to the `PermutationMMD` class (like [this](https://gist.github.com/lambday/96894241f039e8eb5c09338a1570d66a)), then the tests would pass.

vigsterkr:
@lambday this is now how it works that we just pass on things with prngs :)
but note since 
https://github.com/shogun-toolbox/shogun/blob/develop/tests/unit/statistical_testing/internals/CrossValidationMMD_unittest.cc#L110
https://github.com/shogun-toolbox/shogun/blob/develop/tests/unit/statistical_testing/internals/CrossValidationMMD_unittest.cc#L94

these guys dont share a prng.... just saying... so then jsut updating the results should be fine right?

vigsterkr:
@lambday i guess this is what u had in mind https://github.com/shogun-toolbox/shogun/commit/7022a56181c68af5c5f2dab33d056e55d36ba107

note that every test in CrossValidationMMD unit tests fails atm with this change, but since the only check is
```
EXPECT_EQ(cv.m_rejections(current_run*num_folds+current_fold, k), p_value<alpha);
```

i'm not so sure what is update-able here...?

the other test that fails is `KernelSelectionMaxCrossValidation.quadratic_time_single_kernel_dense`, but i guess there the constant could be just updated to its new value:
```
694: The difference between selected_kernel->get_width() and 0.03125 is 0.09375, which exceeds 1E-10, where
694: selected_kernel->get_width() evaluates to 0.125,
694: 0.03125 evaluates to 0.03125, and
694: 1E-10 evaluates to 1e-10.
```

vigsterkr:
note that if i revert the change in CrossValidationMMD but keep the prng changes you've requested, there's still one test that fails: CrossValidationMMD.unbiased_incomplete

vigsterkr:
@lambday any ideas? since now we have 2 same type but separate prng seeded with the same seed... but the unit test still fails :(

lambday:
@vigsterkr let me check this out.. worst case scenario, we're gonna remove the explicit computation logic and put an array of constant expected values... but I want to understand why these are still not giving the same result..

vigsterkr:
@lambday any updates on this by any chance?

lambday:
@vigsterkr hi! I tried to reproduce the issue in my local and it seems like the issue is solved?

gf712:
@vigsterkr is this still an issue?

vigsterkr:
yep

PR #4786, opened 2019-10-25
visitor draft

ktiefe:


src/shogun/base/SGObject.h
```diff
@@ -65,6 +65,15 @@ bool dispatch_array_type(const CSGObject* obj, std::string_view name,
...
 #endif // SWIG
 
+struct InterfaceVisitor
+{
+	CSGObject* value;
+}
+
+Any::register_visitor<InterfaceType, InterfaceVisitor>([] (auto value, auto visitor) {
+        visitor->value = value;
+});
```
gf712: i think the idea is to put this in watch_param? @vigsterkr

src/shogun/base/SGObject.cpp
```diff
@@ -803,7 +803,9 @@ CSGObject* CSGObject::get(const std::string& name, index_t index) const
 CSGObject* CSGObject::get(std::string_view name, index_t index) const
 #endif
 {
-	auto* result = sgo_details::get_by_tag(this, name, sgo_details::GetByNameIndex(index));
+	//auto* result = sgo_details::get_by_tag(this, name, sgo_details::GetByNameIndex(index));
+	auto* result = sgo_details::get_by_tag(this, name);
```
vigsterkr, line 807: the problem here that imo this ignores the `index` parameter. so the visitor that you've written only works in the expected manner for `CSGObject::get(std::string_view name, std::nothrow_t)`. i need to see how we support get `CSGObject* CSGObject::get(std::string_view name, index_t index) const`. that's basically when the registered parameter is an array

vigsterkr (changes requested):
we need to think about how to deal with arrays

src/shogun/base/SGObject.h
```diff
@@ -1344,30 +1373,15 @@ CSGObject* get_if_possible(const CSGObject* obj, std::string_view name, GetByNam
...
-	return nullptr;
-}
-
 template<class T>
-CSGObject* get_by_tag(const CSGObject* obj, std::string_view name,
-		T&& how)
-{
-	return get_dispatch_all_base_types(obj, name, how);
-}
+CSGObject* get_by_tag(const CSGObject* obj, std::string_view name) {
```
vigsterkr, line 1377: this whole thing could be deleted and go into a private function that is implemented in .cpp

src/shogun/base/SGObject.h
```diff
@@ -857,6 +862,13 @@ class CSGObject
 	{
 		BaseTag tag(name);
 		create_parameter(tag, AnyParameter(make_any_ref(value), properties));
+		
+		if constexpr (is_sg_base<T>::value) {
```
vigsterkr, line 866: imo if you wrap this whole block into a constexpr function then you could just call that instead of copy pasting things 4 times.

src/shogun/base/SGObject.h
```diff
@@ -65,6 +65,11 @@ bool dispatch_array_type(const CSGObject* obj, std::string_view name,
 #endif // DOXYGEN_SHOULD_SKIP_THIS
 #endif // SWIG
 
+struct InterfaceVisitor
```
vigsterkr, line 68: this might be better to place into a separate .h file and include it here. place it under src/shogun/util/visitors

vigsterkr:
closing in favour of #4793 
sorry @ktiefe needed this fix...

PR #4789, opened 2019-11-03
use continuous memory block (vector) instead of a map to store parameters

vigsterkr:
coz it's faster and why not


geektoni:
marvellous

src/shogun/base/SGObject.cpp
```diff
@@ -70,7 +70,14 @@ namespace shogun
...
 			}
-			map[tag] = parameter;
+			map.emplace_back(tag, parameter);
+		}
+
+		auto find(const BaseTag& tag) const
+		{
+			for (auto it = map.cbegin(); it != map.cend(); it++)
+				if (it->first == tag)
+					return it;
```
gf712: shouldn't this return map.cend() after the loop?

gf712:
wouldn't it be even faster with std::array? N would have to be some large enough number

src/shogun/base/SGObject.cpp
vigsterkr: mostdef... this was f*!@# up during cleanup before commit

vigsterkr:
@gf712 this is more just like a 'dare you to do it' task... @geektoni didn't believe that this should be an easy patch ;)

iglesias:
> wouldn't it be even faster with std::array? N would have to be some large enough number

It may get too large for the stack.

vigsterkr:
@geektoni actually it was really just that simple patch... the problem is that we had a bug in AnyParameter copy ctor 🤦‍♂

vigsterkr:
anybody who's interested here are some benchmark numbers: https://pastebin.com/JNG55xmX

it's clear that there's a significant speedup using std::vector vs std::map... the only thing is that having the elements sorted in the std::vector in case of 6 registered params adds more overhead than gain....

iglesias:
Let's merge, then?

src/shogun/base/SGObject.cpp
```diff
@@ -56,59 +57,120 @@
 namespace shogun
 {
 
-	typedef std::map<BaseTag, AnyParameter> ParametersMap;
+	typedef std::vector<std::pair<BaseTag, AnyParameter>> ParametersMap;
 	typedef std::unordered_map<std::string_view, std::string_view> ObsParamsList;
 
+#define SORTED 0
```
geektoni: I guess this bunch of `#if` and `#else` will be removed once we choose the final implementation.

Looking at the benchmarks, I guess we could go with the unsorted option (unless we prefer to have the
parameter sorted for some reason). Since I don't think we will ever have like 1000 parameters registered we don't need to sort.

vigsterkr: yeah actually no... i've refactored a bit the whole SGObject::Self. this way it's easier to add a new type of backend. like a distributed key-value store? ;)
i opted to use template based solution instead of vtable, hoping that it's less amount of overhead.

src/shogun/base/SGObject.cpp
```diff
@@ -56,71 +68,9 @@
...
-				    return p.has_property(pprop) ||
-								(pprop==ParameterProperties::ALL &&
-								p.compare_mask(ParameterProperties::NONE));
-			    });
-			return result;
-		}
-
-		ParametersMap map;
-	};
+	class SGObject::Self: public detail::ParameterInterface<ParameterBackend> {};
```
vigsterkr, line 73: @lisitsyn like?

PR #4792, opened 2019-11-08
PlugINs!

vigsterkr:


vigsterkr:
based of the work of Sanuj from 2016 ;)

vigsterkr:
@sanuj finally the last remaining part of your code is about to get in develop :P

src/shogun/base/library.h
```diff
@@ -0,0 +1,95 @@
...
+
+        /** @return name of function that accesses Manifest
+         * of loaded library.
+         */
+        static std::string_view get_manifest_accessor_name()
+        {
+            return kManifestAccessorName;
+        }
+
+        friend void unload_library(Library&& lib);
```
gf712: why do you inject this to the outer namespace and have `void unload_library(Library&& lib);` 20 lines down?

vigsterkr: m_handle is private hence the friend part.

src/shogun/base/class_list.cpp
```diff
@@ -32,39 +32,83 @@
...
+							class_name, plugin);
+						continue;
+					}
 
-typedef SGObject* (*CreateFunction)(EPrimitiveType generic);
+					SG_TRACE("adding {} class of {} to class list", class_name, plugin);
+					class_list.emplace(class_name, [plugin, class_name](EPrimitiveType generic) {
+						// TODO: add library handle caching
+						auto lib = load_library(plugin);
+						auto metaclass = lib.manifest().class_by_name<SGObject>(class_name);
```
vigsterkr: @lisitsyn ok so if we wanna support(?) old school `SGO create("classname")` then things need to be exported as SGObject based... or i guess at least have in the manifest something like:
```
        template <typename T, enable_if_t<is_base_v<SGObject, T>>* = nullptr>
        MetaClass<SGObject> class_by_name(const std::string& name) const
        {
            auto clazz = find_class(name);
            return clazz.cast<MetaClass<SGObject>>();
        }
```

src/shogun/base/class_list.cpp
```diff
@@ -32,39 +32,83 @@
...
-REPLACE struct THIS
-};
+ClassList class_list()
+{
+	static const ClassList kClassList = create_class_list();
+	return kClassList;
+}
 
-std::shared_ptr<SGObject> shogun::create(const char* classname, EPrimitiveType generic)
+std::shared_ptr<SGObject> shogun::create(const std::string& classname, EPrimitiveType generic)
```
vigsterkr, line 67: @lisitsyn 
based on the current design, templates should be exported with different names so the whole `EPrimitiveType` is getting foobar.
say `template<T> DenseFeatures...`

for being able to call `create(...)` for that object currently we need to give different names, and i dont think there's a way around this... just like in case of swig... namely: IntDenseFeatures, RealFeatures etc.

vigsterkr:
ok so here essentially the c++ part works as it's expected.... now we would need some cmake guru to help us divide the project into smaller subprojects. any takers on this?

geektoni:
> ok so here essentially the c++ part works as it's expected.... now we would need some cmake guru to help us divide the project into smaller subprojects. any takers on this?

What about using this a "good first issue" for future GSoC students :) ? However, I have no idea about the complexity of it atm (or if we want to do it ourselves) ...

vigsterkr:
closing in favour of the work happening in feature/modular-graph branch

PR #4801, opened 2019-11-29
Add unit tests for utils::safe_convert

### LINKED ISSUE #4580: create unit tests for utils::safe_convert function

ArtemSkrebkov:
Fixes https://github.com/shogun-toolbox/shogun/issues/4580

### DIFF
--- a/tests/unit/utils/converter_unittest.cc
+++ b/tests/unit/utils/converter_unittest.cc
@@ -109,7 +109,10 @@ TYPED_TEST(Converters, canConvertFloatInfinityToReal)
 		ASSERT_NO_THROW(
 		    result = utils::safe_convert<TypeParam>(
 		        std::numeric_limits<float>::infinity()));
-		EXPECT_TRUE(std::isinf(result));
+		// would be better to use
+		// EXPECT_TRUE(std::isinf(result));
+		// but microsoft compiler does not compile it
+		EXPECT_FALSE(IS_FINITE(result));
 	}
 }
 
@@ -121,6 +124,9 @@ TYPED_TEST(Converters, canConvertFloatNaNToReal)
 		ASSERT_NO_THROW(
 		    result = utils::safe_convert<TypeParam>(
 		        std::numeric_limits<float>::quiet_NaN()));
-		EXPECT_TRUE(std::isnan(result));
+		// would be better to use
+		// EXPECT_TRUE(std::isnan(result));
+		// but microsoft compiler does not compile it
+		EXPECT_FALSE(IS_FINITE(result));
 	}
 }
--- a/tests/unit/utils/converter_unittest.cc
+++ b/tests/unit/utils/converter_unittest.cc
@@ -1,0 +1,126 @@
+#include <gtest/gtest.h>
+
+#include "sg_gtest_utilities.h"
+
+#include <shogun/lib/sg_types.h>
+#include <shogun/util/converters.h>
+
+#include <cstdint>
+#include <iostream>
+#include <limits>
+
+using namespace shogun;
+
+template <typename T>
+class Converters : public ::testing::Test
+{
+};
+
+SG_TYPED_TEST_CASE(Converters, sg_non_complex_types);
+
+TYPED_TEST(Converters, canDoCorrectConvertionOfUnsignedArg)
+{
+	TypeParam result = 0;
+	uint64_t value = 120ull;
+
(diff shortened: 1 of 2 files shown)


ArtemSkrebkov:
There are some controversial things with those tests. So please let me know what you think.

tests/unit/utils/converter_unittest.cc
```diff
@@ -0,0 +1,84 @@
...
+{
+	TypeParam result{0};
+	ASSERT_NO_THROW(result = utils::safe_convert<TypeParam>(true));
+	EXPECT_EQ(static_cast<TypeParam>(1), result);
+
+	ASSERT_NO_THROW(result = utils::safe_convert<TypeParam>(false));
+	EXPECT_EQ(static_cast<TypeParam>(0), result);
+}
+
+// TODO: either remove tests or fix conversion functions
```
vigsterkr: if you could add the fix to the converter utility would be great so we could enable these tests

tests/unit/utils/converter_unittest.cc
```diff
@@ -0,0 +1,84 @@
...
+	EXPECT_EQ(static_cast<TypeParam>(value), result);
+}
+
+TYPED_TEST(Converters, cannotConvertOutOfRange)
+{
+	auto isLongDouble = typeid(TypeParam) == typeid(long double);
+
+	if (isLongDouble)
+	{
+		std::cout << "Disaled for long double since it fails" << std::endl;
```
vigsterkr: this could be only a comment only no need to write it out and just quietly ignore this test. i'm just wondering why the test fails for long?

tests/unit/utils/converter_unittest.cc
```diff
@@ -0,0 +1,84 @@
+#include "sg_gtest_utilities.h"
+
+#include <gtest/gtest.h>
```
vigsterkr: move this to be the first included header plz

vigsterkr (changes requested):
sorry for the late review, here are some comments

tests/unit/utils/converter_unittest.cc
ArtemSkrebkov: Moved.

tests/unit/utils/converter_unittest.cc
ArtemSkrebkov: Fixed. The current implementation counts that passing Infinity and NaN to `safe_convert` will cause an overflow error. Is that OK?

tests/unit/utils/converter_unittest.cc
ArtemSkrebkov: Got it. The test is green after applying the fix I mentioned above.

tests/unit/utils/converter_unittest.cc
```diff
@@ -0,0 +1,78 @@
...
+	EXPECT_EQ(static_cast<TypeParam>(1), result);
+
+	ASSERT_NO_THROW(result = utils::safe_convert<TypeParam>(false));
+	EXPECT_EQ(static_cast<TypeParam>(0), result);
+}
+
+TYPED_TEST(Converters, cannotConvertFloatInfinity)
+{
+	EXPECT_THROW(
+	    utils::safe_convert<TypeParam>(std::numeric_limits<float>::infinity()),
```
vigsterkr: float inf -> double inf should work no?

tests/unit/utils/converter_unittest.cc
```diff
@@ -0,0 +1,78 @@
...
+	    std::overflow_error);
+}
+
+TYPED_TEST(Converters, cannotConvertFloatNaN)
+{
+	EXPECT_THROW(
+	    utils::safe_convert<TypeParam>(std::numeric_limits<float>::quiet_NaN()),
```
vigsterkr: float NAN -> double NAN should work no?

tests/unit/utils/converter_unittest.cc
ArtemSkrebkov: Agree. Updated `safe_convert` and added tests to cover this case.

tests/unit/utils/converter_unittest.cc
ArtemSkrebkov: Added a test to cover the case.

vigsterkr:
great stuff!

PR #4804, opened 2019-12-17
add resnet and inception v4

ZHOUKAI1:
deep learning is a very important way now, can you add more deep learning algorithm? thx


vigsterkr:
@ZHOUKAI1 i'm not so sure what's this patch about but there's a lot of noise /commits that are not related/... could you send in a PR that is actually just about your changes?

Issue #4639, opened 2019-05-16
Implement Levenstein string distance algorithm


karlnapf:
To be used in #4473

First standalone against `std::string`, and then also embedded in a `CDistance` specialization.

assignment6:
Hello,
Does this have to be in Python?  I can do in C ;-)

gf712:
Hey @assignment6, no, the idea is to have it in C++ so that it can be exposed to the other languages!

assignment6:
Ol.  But the other languages are based on C ;-), just sayin'

infinidoko:
Hi, please i would like to work on this if the issue is still available. I can implement the algorithm in c++.

karlnapf:
please go ahead

hariduttt:
Has it been implemented or I would like to work on it.

voldemoriarty:
https://bitbucket.org/clearer/iosifovich/src/master/

karlnapf:
We sometimes copy paste import code of license permits, it is standalone and minimal, and very specific to what we need (contains nothing else).

h-sigma:
[levenshtein.txt](https://github.com/shogun-toolbox/shogun/files/3396916/levenshtein.txt)
Would this suffice? I can rewrite it to not use standard library classes.

vigsterkr:
@h-sigma great start but for being able to review and give you feedback we would actually would like to have a pull request sent in with the changes you propose. for how to go further: https://github.com/shogun-toolbox/shogun/blob/develop/doc/readme/DEVELOPING.md

tthebst:
Hi, I want to try this. Anything special I need to know?

darelover:
Hi, I would like to work on this issue but I have few doubts:

1. Where should I put the file in the project?
2. Should i make it a header file or not?
3. Is there any header file that i need to include?

gf712:
> Hi, I would like to work on this issue but I have few doubts:
> 
>     1. Where should I put the file in the project?

You could probably put it in shogun/distance

>     2. Should i make it a header file or not?

Check out other distance implementations, such as EuclideanDistance.[h](https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/distance/EuclideanDistance.h)/[cpp](https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/distance/EuclideanDistance.cpp)

>     3. Is there any header file that i need to include?

[...]

Issue #4475, opened 2019-01-18
Suggest class name in factory instantiation


gf712:
## Intro
Instead of using directly the classes from C++, we use factories (see [here](https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/util/factory.h)). So for example in Python, instead of
```python
import shogun as sg
kernel = sg.GaussianKernel()
```
we have
```python
import shogun as sg
kernel = sg.kernel("GaussianKernel")
```
To find out what all possible names you can have a look at `src/shogun/base/class_list.cpp`, after compiling shogun.
The issue with this approach is that it is easy to mistype a name, e.g. `"Gaussiankernel"`, and after several hours of writing code these errors might happen more frequently and not so obvious to debug.
Right now you would get a message like this:
```python
SystemError: [ERROR] In file /home/gil/shogun/src/shogun/base/class_list.h line 51: Class Gaussiankernel with primitive type SGOBJECT does not exist.
```
## The task
Catch these errors with SWIG, for example in src/interfaces/shogun.i and try to find out what the user actually meant. This should be written in C++, which would make it accessible to all languages.

## Some pointers
- A good approach would be to calculate the distance between the string the user passed and all the classes in shogun
- All classes in shogun that are exposed via factories can be accessed with [shogun::available_objects()](https://github.com/shogun-toolbox/shogun/blob/e3e8cd197b4dde5a6c75136bea54e7fdc141a947/src/shogun/base/class_list.h#L71)
- A good distance metric for this is levenshtein, which is not implemented anywhere in shogun, but it is pretty straighforward (see [here](https://rosettacode.org/wiki/Levenshtein_distance#C.2B.2B)).
- Check #4473, it has a Python implementation of this

karlnapf:
Note: this should not be implemented in a standard c++ shogun source, but in the swig extension. E.g. see https://github.com/shogun-toolbox/shogun/blob/develop/src/interfaces/swig/shogun.i

PR #4806, opened 2019-12-28
Add feature LevensteinDistance

LiuYuHui:
Fix issue #4639


src/shogun/distance/LevensteinDistance.cpp
```diff
@@ -0,0 +1,68 @@
...
+}
+
+size_t
+LevensteinDistance::compute(const std::string& word1, const std::string& word2)
+{
+	require(word1.size() != 0, "Left hand side name cannot be empty!");
+	require(word2.size() != 0, "Right hand side name cannot be empty!");
+
+	int m = word1.size(), n = word2.size();
+	std::vector<std::vector<size_t>> dist(m + 1, std::vector<size_t>(n + 1));
```
gf712: you could use SGMatrix here instead

src/shogun/distance/LevensteinDistance.h
```diff
@@ -0,0 +1,56 @@
...
+ *
+ */
+#ifndef _LEVENSTEINDISTANCE_H__
+#define _LEVENSTEINDISTANCE_H__
+
+#include <string>
+#include <vector>
+namespace shogun
+{
+	class LevensteinDistance
```
gf712: This class should probably inherit from Distance no?

LiuYuHui: I find `Distance` have many api that LevensteinDistance don't need, such as `virtual bool init(std::shared_ptr<Features> lhs, std::shared_ptr<Features> rhs);`. Because the function of LevensteinDistance is that compare two class names which is `string`.

LiuYuHui: hi, @gf712, Can you give me some idea about implement Levenshtein algorithm for shogun Features?

LiuYuHui: Hi @gf712 , I have changed LevenshteinDistance inherit from Distance, Please have a look when you are free, thanks :)

karlnapf:
Now that we have this, we could make the auto-complete thing envisioned by @gf712 Not remember whether there is an issue for this?

gf712:
Yup, #4475

PR #4809, opened 2020-01-02
Fixes 4795

### LINKED ISSUE #4795: DenseSubSamplesFeatures to DenseFeatures conversion
While training a random forest, I generate a DenseSubSamplesFeatures<double> from a DenseFeatures<double>.
Compilation works.
At runtime, I get:
```
RuntimeError: Object of type DenseSubSamplesFeatures cannot be converted to type shogun::DenseFeatures<double>.
```

I'm working on commit 84175ba7c8c73dd4028f217daf7a5f4c25c2adaf.
I compile with g++ 9.1.0 on Arch Linux.

Here's the code that generates the DenseSubSamplesFeatures
https://github.com/lejeunel/glia/blob/itk-5.0.1/src/shogun_helpers.hxx#L83

LiuYuHui:


src/shogun/multiclass/tree/CARTree.cpp
```diff
@@ -69,97 +73,121 @@ CARTree::CARTree(SGVector<bool> attribute_types, EProblemType prob_type, int32_t
...
 {
 	require(data, "Data required for classification in apply_multiclass");
 
 	// apply multiclass starting from root
-	auto current=get_root()->as<bnode_t>();
+	auto current = get_root()->as<bnode_t>();
 
 	require(current, "Tree machine not yet trained.");
-	return apply_from_current_node(data->as<DenseFeatures<float64_t>>(), current)->as<MulticlassLabels>();
+	if (data->get_name() == "DenseSubSamplesFeatures")
```
gf712, line 114: You shouldn't compare strings, you can check the result of dynamic_cast. Btw SGObject has a `as` member function to do the casting:
```cpp 
if (auto* tmp = data->as<DenseSubSamplesFeatures<float64_t>>(data))
```

gf712: Also, doesn't get_name() return `const char*`? If so what you're doing here is compare pointer values, what you should use is [strcmp](https://en.cppreference.com/w/c/string/byte/strcmp)

LiuYuHui: Thanks for your review, I find that auto formatting make my commit a mess, so i closed this pr, i will create a new pr.

gf712: OK! I was wondering why it was closed. I will review it better once you updated it. 
Btw you don't need to close this PR, just continue working here and let us know when it's ready for review!

PR #4810, opened 2020-01-03
Fix 4795

### LINKED ISSUE #4795: DenseSubSamplesFeatures to DenseFeatures conversion

LiuYuHui:
Fix #4795

### DIFF
--- a/src/shogun/multiclass/tree/CARTree.cpp
+++ b/src/shogun/multiclass/tree/CARTree.cpp
@@ -119,7 +119,8 @@ std::shared_ptr<MulticlassLabels> CARTree::apply_multiclass(std::shared_ptr<Feat
 		SGMatrix<float64_t> feature_matrix(num_feat, num_vec);
 		for (size_t i = 0; i < num_vec; i++)
 		{
-			SGVector<float64_t> v = subfeat_data->get_computed_dot_feature_vector(i);
+			SGVector<float64_t> v =
+			    subfeat_data->get_computed_dot_feature_vector(i);
 			ASSERT(num_feat == v.vlen)
 			for (size_t j = 0; j < num_feat; j++)
 				feature_matrix.matrix[i * num_feat + j] = v.vector[j];
--- a/src/shogun/multiclass/tree/CARTree.cpp
+++ b/src/shogun/multiclass/tree/CARTree.cpp
@@ -121,9 +121,7 @@ std::shared_ptr<MulticlassLabels> CARTree::apply_multiclass(std::shared_ptr<Feat
 		{
 			SGVector<float64_t> v =
 			    subfeat_data->get_computed_dot_feature_vector(i);
-			ASSERT(num_feat == v.vlen)
-			for (size_t j = 0; j < num_feat; j++)
-				feature_matrix.matrix[i * num_feat + j] = v.vector[j];
+			linalg::add_col_vec(feature_matrix, i, v, feature_matrix, 0, 1);
 		}
 		return apply_from_current_node(
 		           std::make_shared<DenseFeatures<float64_t>>(feature_matrix),
--- a/src/shogun/multiclass/tree/CARTree.cpp
+++ b/src/shogun/multiclass/tree/CARTree.cpp
@@ -28,6 +28,8 @@
  * either expressed or implied, of the Shogun Development Team.
  */
 
+#include <algorithm>
+#include <iterator>
 #include <shogun/lib/View.h>
 #include <shogun/mathematics/Math.h>
 #include <shogun/mathematics/RandomNamespace.h>
@@ -119,10 +121,9 @@ std::shared_ptr<MulticlassLabels> CARTree::apply_multiclass(std::shared_ptr<Feat
 		SGMatrix<float64_t> feature_matrix(num_feat, num_vec);
 		for (size_t i = 0; i < num_vec; i++)
 		{
-			SGVector<float64_t> v =
-			    subfeat_data->get_computed_dot_feature_vector(i);
+			const auto& v = subfeat_data->get_computed_dot_feature_vector(i);
 			sg_memcpy(
-			    feature_matrix.data() + i * num_feat, v.data(),
+			    feature_matrix.get_column_vector(i), v.data(),
 			    num_feat * sizeof(float64_t));
 		}
 		return apply_from_current_node(
--- a/src/shogun/multiclass/tree/CARTree.cpp
+++ b/src/shogun/multiclass/tree/CARTree.cpp
@@ -114,14 +114,16 @@ std::shared_ptr<MulticlassLabels> CARTree::apply_multiclass(std::shared_ptr<Feat
 	if (auto subfeat_data =
 	        std::dynamic_pointer_cast<DenseSubSamplesFeatures<float64_t>>(data))
 	{
-		auto num_feat = subfeat_data->get_dim_feature_space();
+		size_t num_feat = subfeat_data->get_dim_feature_space();
 		auto num_vec = subfeat_data->get_num_vectors();
 		SGMatrix<float64_t> feature_matrix(num_feat, num_vec);
 		for (size_t i = 0; i < num_vec; i++)
 		{
 			SGVector<float64_t> v =
 			    subfeat_data->get_computed_dot_feature_vector(i);
-			linalg::add_col_vec(feature_matrix, i, v, feature_matrix, 0, 1);
+			sg_memcpy(
+			    feature_matrix.data() + i * num_feat, v.data(),
+			    num_feat * sizeof(float64_t));
 		}
 		return apply_from_current_node(
 		           std::make_shared<DenseFeatures<float64_t>>(feature_matrix),
--- a/src/shogun/multiclass/tree/CARTree.cpp
+++ b/src/shogun/multiclass/tree/CARTree.cpp
@@ -111,20 +111,18 @@ std::shared_ptr<MulticlassLabels> CARTree::apply_multiclass(std::shared_ptr<Feat
 	auto current=get_root()->as<bnode_t>();
 
 	require(current, "Tree machine not yet trained.");
-	if (auto tmp =
+	if (auto subfeat_data =
 	        std::dynamic_pointer_cast<DenseSubSamplesFeatures<float64_t>>(data))
 	{
-		auto num_feat = tmp->get_dim_feature_space();
-		auto num_vec = tmp->get_num_vectors();
-		shogun::SGMatrix<float64_t> feature_matrix;
-		for (auto i = 0; i < num_vec; i++)
+		auto num_feat = subfeat_data->get_dim_feature_space();
+		auto num_vec = subfeat_data->get_num_vectors();
+		SGMatrix<float64_t> feature_matrix(num_feat, num_vec);
+		for (size_t i = 0; i < num_vec; i++)
 		{
-			SGVector<float64_t> v = tmp->get_computed_dot_feature_vector(i);
+			SGVector<float64_t> v = subfeat_data->get_computed_dot_feature_vector(i);
 			ASSERT(num_feat == v.vlen)
-
-			for (auto j = 0; j < num_feat; j++)
-				feature_matrix.matrix[i * int64_t(num_feat) + j] =
-				    (float64_t)v.vector[j];
+			for (size_t j = 0; j < num_feat; j++)
+				feature_matrix.matrix[i * num_feat + j] = v.vector[j];
 		}
 		return apply_from_current_node(
 		           std::make_shared<DenseFeatures<float64_t>>(feature_matrix),
--- a/src/shogun/multiclass/tree/CARTree.cpp
+++ b/src/shogun/multiclass/tree/CARTree.cpp
@@ -111,7 +111,30 @@ std::shared_ptr<MulticlassLabels> CARTree::apply_multiclass(std::shared_ptr<Feat
 	auto current=get_root()->as<bnode_t>();
 
 	require(current, "Tree machine not yet trained.");
-	return apply_from_current_node(data->as<DenseFeatures<float64_t>>(), current)->as<MulticlassLabels>();
+	if (auto tmp =
+	        std::dynamic_pointer_cast<DenseSubSamplesFeatures<float64_t>>(data))
+	{
+		auto num_feat = tmp->get_dim_feature_space();
+		auto num_vec = tmp->get_num_vectors();
+		shogun::SGMatrix<float64_t> feature_matrix;
+		for (auto i = 0; i < num_vec; i++)
+		{
+			SGVector<float64_t> v = tmp->get_computed_dot_feature_vector(i);
+			ASSERT(num_feat == v.vlen)
+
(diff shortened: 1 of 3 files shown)


src/shogun/multiclass/tree/CARTree.cpp
```diff
@@ -109,7 +109,30 @@ std::shared_ptr<MulticlassLabels> CARTree::apply_multiclass(std::shared_ptr<Feat
 	auto current=get_root()->as<bnode_t>();
 
 	require(current, "Tree machine not yet trained.");
-	return apply_from_current_node(data->as<DenseFeatures<float64_t>>(), current)->as<MulticlassLabels>();
+	if (auto tmp =
+	        std::dynamic_pointer_cast<DenseSubSamplesFeatures<float64_t>>(data))
+	{
+		auto num_feat = tmp->get_dim_feature_space();
+		auto num_vec = tmp->get_num_vectors();
+		shogun::SGMatrix<float64_t> feature_matrix;
```
gf712: You don't need to add the namespace shogun here.
Doesn't this cause a segfault? You are allocating an empty matrix and then below you access the memory directly (`feature_matrix.matrix`)?

src/shogun/multiclass/tree/CARTree.cpp
```diff
@@ -109,7 +109,30 @@ std::shared_ptr<MulticlassLabels> CARTree::apply_multiclass(std::shared_ptr<Feat
 	auto current=get_root()->as<bnode_t>();
 
 	require(current, "Tree machine not yet trained.");
-	return apply_from_current_node(data->as<DenseFeatures<float64_t>>(), current)->as<MulticlassLabels>();
+	if (auto tmp =
```
gf712: maybe give this a more significant name, rather then tmp?
also use `->as<DenseSubSamplesFeatures<float64_t>>`

src/shogun/multiclass/tree/CARTree.cpp
```diff
@@ -109,7 +109,30 @@ std::shared_ptr<MulticlassLabels> CARTree::apply_multiclass(std::shared_ptr<Feat
...
+		auto num_vec = tmp->get_num_vectors();
+		shogun::SGMatrix<float64_t> feature_matrix;
+		for (auto i = 0; i < num_vec; i++)
+		{
+			SGVector<float64_t> v = tmp->get_computed_dot_feature_vector(i);
+			ASSERT(num_feat == v.vlen)
+
+			for (auto j = 0; j < num_feat; j++)
+				feature_matrix.matrix[i * int64_t(num_feat) + j] =
+				    (float64_t)v.vector[j];
```
gf712: the (float64_t) is not necessary i think?

src/shogun/multiclass/tree/CARTree.cpp
```diff
@@ -109,7 +109,30 @@ std::shared_ptr<MulticlassLabels> CARTree::apply_multiclass(std::shared_ptr<Feat
...
+		auto num_feat = tmp->get_dim_feature_space();
+		auto num_vec = tmp->get_num_vectors();
+		shogun::SGMatrix<float64_t> feature_matrix;
+		for (auto i = 0; i < num_vec; i++)
+		{
+			SGVector<float64_t> v = tmp->get_computed_dot_feature_vector(i);
+			ASSERT(num_feat == v.vlen)
+
+			for (auto j = 0; j < num_feat; j++)
+				feature_matrix.matrix[i * int64_t(num_feat) + j] =
```
gf712: rather than cast with int64_t(num_feat) declare j as int64_t. But I think it should be size_t, as this value should be unsigned

gf712:
Hey, thanks for the PR. Is there a way you could test these changes with unittests or a metaexample?

src/shogun/multiclass/tree/CARTree.cpp
LiuYuHui: use `->as<DenseSubSamplesFeatures<float64_t>>` will give me a error. Because the implemention of as is
``` cpp
       template<class T>
	inline std::shared_ptr<T> as()
	{
		auto c = std::dynamic_pointer_cast<T>(shared_from_this());
		if (c)
			return c;
		error(
			"Object of type {} cannot be converted to type {}.",
			this->get_name(),
			demangled_type<T>().c_str());
		return nullptr;
	}

File: src/shogun/multiclass/tree/CARTree.cpp (hunk 1 above)
Comment (LiuYuHui) [reply src/shogun/multiclass/tree/CARTree.cpp]: i will repair this as `SGMatrix<float64_t> feature_matrix(num_feat, num_vec);`

**Comment by LiuYuHui** [comment]
OK, I will construct a `DenseSubSamplesFeatures` from a `DenseFeatures`, and use `CARTree` to classify this features. Thanks for your review again.

File: src/shogun/multiclass/tree/CARTree.cpp (hunk 2 above)
Comment (gf712) [reply src/shogun/multiclass/tree/CARTree.cpp]: do you get a compile or runtime error?

Comment (LiuYuHui) [reply src/shogun/multiclass/tree/CARTree.cpp]: yes, I got a exception that `" cannot convert ‘std::shared_ptr<shogun::DenseSubSamplesFeatures<double> >’ to ‘shogun::DenseSubSamplesFeatures<double>*’ in return
   return sgo->as<T>(); 
"`.

Comment (gf712) [reply src/shogun/multiclass/tree/CARTree.cpp]: You are calling the wrong function, I think. That error comes from https://github.com/shogun-toolbox/shogun/blob/e900a9e602763e1ddc8ac98bb0448ad0a68593e0/src/shogun/base/SGObject.h#L659-L663
Did you write `data->as<DenseSubSamplesFeatures<float64_t>>`?

Comment (LiuYuHui) [reply src/shogun/multiclass/tree/CARTree.cpp]: I called `as` function like
```
 if (auto* subfeat_data = data->as<DenseSubSamplesFeatures<float64_t>>(data))
```
I don't know whether this way is right.

Comment (gf712) [reply src/shogun/multiclass/tree/CARTree.cpp]: ah yes, the ->as overload you want to call doesn't have a parameter, so it should be
`data->as<DenseSubSamplesFeatures<float64_t>>()`

Comment (LiuYuHui) [reply src/shogun/multiclass/tree/CARTree.cpp]: I change to this way
``` cpp
 if (auto subfeat_data = data->as<DenseSubSamplesFeatures<float64_t>>())
```
But, it can't pass unit test, a exception `C++ exception with description "Object of type DenseFeatures cannot be converted to type shogun::DenseSubSamplesFeatures<double>." thrown in the test body.` is thrown.
I think we should not use `as` function in this place, because if `std::shared_ptr<Features> data` can't converted to `std::shared_ptr<DenseSubSamplesFeatures<double>>`, we should not thrown a exception, we should handle this situation in else clause.

gf712: ah yes, that's true!

LiuYuHui:
Hi @gf712 , I made all required changes and added some test cases for my changes.
Please take a look when you are free. Thanks

LiuYuHui:
Hi @gf712 , I am still looking forward to your reply :)

gf712:
this looks good to me. maybe @karlnapf or @vigsterkr can have a quick look to see if this is the fix they needed for #4795?

LiuYuHui:
hi @vigsterkr, can you tell me whether this way is right? Thanks a lot.

gf712:
@vigsterkr can you have a look at this so we can merge? @LiuYuHui could you rebase?

src/shogun/multiclass/tree/CARTree.cpp
```diff
@@ -111,7 +111,29 @@ std::shared_ptr<MulticlassLabels> CARTree::apply_multiclass(std::shared_ptr<Feat
...
+		auto num_feat = subfeat_data->get_dim_feature_space();
+		auto num_vec = subfeat_data->get_num_vectors();
+		SGMatrix<float64_t> feature_matrix(num_feat, num_vec);
+		for (size_t i = 0; i < num_vec; i++)
+		{
+			SGVector<float64_t> v =
+			    subfeat_data->get_computed_dot_feature_vector(i);
+			ASSERT(num_feat == v.vlen)
+			for (size_t j = 0; j < num_feat; j++)
+				feature_matrix.matrix[i * num_feat + j] = v.vector[j];
```
karlnapf: this could be done using memset or a newly added linalg call (that could do memset)

tests/unit/multiclass/tree/CARTree_unittest.cc
```diff
@@ -193,9 +193,23 @@ TEST(CARTree, classify_nominal)
...
-
-
-
+	int feats_num = test_feats->get_num_vectors();
+	SGVector<int32_t> idx(feats_num);
+	for (int i = 0; i < feats_num; i++)
+	{
+		idx[i] = i;
+	}
+	// construct DenseSubSamplesFeatures from DensesFeatures
```
karlnapf: minor: redundant comment :)

src/shogun/multiclass/tree/CARTree.cpp
```diff
@@ -111,7 +111,29 @@ std::shared_ptr<MulticlassLabels> CARTree::apply_multiclass(std::shared_ptr<Feat
...
+	        std::dynamic_pointer_cast<DenseSubSamplesFeatures<float64_t>>(data))
+	{
+		auto num_feat = subfeat_data->get_dim_feature_space();
+		auto num_vec = subfeat_data->get_num_vectors();
+		SGMatrix<float64_t> feature_matrix(num_feat, num_vec);
+		for (size_t i = 0; i < num_vec; i++)
+		{
+			SGVector<float64_t> v =
+			    subfeat_data->get_computed_dot_feature_vector(i);
+			ASSERT(num_feat == v.vlen)
```
karlnapf: if this is something that a user could ever see, then it needs a proper error message with hints to to avoid the error

src/shogun/multiclass/tree/CARTree.cpp
```diff
@@ -111,7 +111,29 @@ std::shared_ptr<MulticlassLabels> CARTree::apply_multiclass(std::shared_ptr<Feat
 	auto current=get_root()->as<bnode_t>();
 
 	require(current, "Tree machine not yet trained.");
-	return apply_from_current_node(data->as<DenseFeatures<float64_t>>(), current)->as<MulticlassLabels>();
+	if (auto subfeat_data =
+	        std::dynamic_pointer_cast<DenseSubSamplesFeatures<float64_t>>(data))
+	{
+		auto num_feat = subfeat_data->get_dim_feature_space();
+		auto num_vec = subfeat_data->get_num_vectors();
+		SGMatrix<float64_t> feature_matrix(num_feat, num_vec);
```
karlnapf, line 121: actually, the is commodity code, it should be somewhere where it can be re-used, for example in the form of a copy constructor, or a method that returns the new feature object

LiuYuHui: add `DenseFeatures::DenseFeatures(std::shared_ptr<DenseSubSamplesFeatures<T>>& data);` ?

src/shogun/multiclass/tree/CARTree.cpp
```diff
@@ -111,7 +111,27 @@ std::shared_ptr<MulticlassLabels> CARTree::apply_multiclass(std::shared_ptr<Feat
...
+		for (size_t i = 0; i < num_vec; i++)
+		{
+			SGVector<float64_t> v =
+			    subfeat_data->get_computed_dot_feature_vector(i);
+			linalg::add_col_vec(feature_matrix, i, v, feature_matrix, 0, 1);
```
gf712: wouldn't it be better to use `sg_memcpy`? I am not sure if the compiler can optimise away the add in this situation and use memcpy instead

karlnapf: I agree. But even better would be a method that hides this stuff under the hood imo

LiuYuHui: you mean add something to linalg ?

karlnapf: I talked to @gf712 who reminded me that there is an column iterator for SGMatrix. So you can use that together with memset. No need to touch linalg then

src/shogun/multiclass/tree/CARTree.cpp
```diff
@@ -111,7 +111,29 @@ std::shared_ptr<MulticlassLabels> CARTree::apply_multiclass(std::shared_ptr<Feat
...
+	{
+		size_t num_feat = subfeat_data->get_dim_feature_space();
+		auto num_vec = subfeat_data->get_num_vectors();
+		SGMatrix<float64_t> feature_matrix(num_feat, num_vec);
+		for (size_t i = 0; i < num_vec; i++)
+		{
+			SGVector<float64_t> v =
+			    subfeat_data->get_computed_dot_feature_vector(i);
+			sg_memcpy(
+			    feature_matrix.data() + i * num_feat, v.data(),
```
karlnapf: I think there is a method that returns the pointer to the i-th column vector, might make it easier to read?
I also think there is an iterator for column vectors, which might make it even more clean :)

karlnapf (approved):

src/shogun/multiclass/tree/CARTree.cpp
```diff
@@ -111,7 +113,28 @@ std::shared_ptr<MulticlassLabels> CARTree::apply_multiclass(std::shared_ptr<Feat
...
-	return apply_from_current_node(data->as<DenseFeatures<float64_t>>(), current)->as<MulticlassLabels>();
+	if (auto subfeat_data =
+	        std::dynamic_pointer_cast<DenseSubSamplesFeatures<float64_t>>(data))
+	{
+		size_t num_feat = subfeat_data->get_dim_feature_space();
+		auto num_vec = subfeat_data->get_num_vectors();
+		SGMatrix<float64_t> feature_matrix(num_feat, num_vec);
+		for (size_t i = 0; i < num_vec; i++)
+		{
+			const auto& v = subfeat_data->get_computed_dot_feature_vector(i);
```
gf712, line 124: not sure why you are capturing a reference to a temporary value here, but seems like it works?

gf712: oh well, this is valid C++! Future reference https://blog.galowicz.de/2016/03/23/const_reference_to_temporary_object/

LiuYuHui: The lifetime of a temporary object can be extended by binding to a const lvalue reference :D, i learn it from [this](https://herbsutter.com/2008/01/01/gotw-88-a-candidate-for-the-most-important-const/).

PR #4812, opened 2020-01-10
[Feature Addition] Mish Activation function

### LINKED ISSUE #4797: Feature Request: Add Mish Activation
Mish is a new novel activation function proposed in this [paper](https://arxiv.org/abs/1908.08681).
It has shown promising results so far and has been adopted in several packages including:

- [TensorFlow-Addons](https://github.com/tensorflow/addons/tree/master/tensorflow_addons/activations)
- [SpaCy (Tok2Vec Layer)](https://github.com/explosion/spaCy)
- [Thinc - SpaCy's official NLP based ML library](https://github.com/explosion/thinc/releases/tag/v7.3.0)
- [Echo AI](https://github.com/digantamisra98/Echo)
- [Eclipse's deeplearning4j](https://github.com/eclipse/deeplearning4j/issues/8417)
- [CNTKX - Extension of Microsoft's CNTK](https://github.com/delzac/cntkx)
- [FastAI-Dev](https://github.com/fastai/fastai_dev/blob/0f613ba3205990c83de9dba0c8798a9eec5452ce/dev/local/layers.py#L441)
- [Darknet](https://github.com/AlexeyAB/darknet/commit/bf8ea4183dc265ac17f7c9d939dc815269f0a213)
- [Yolov3](https://github.com/ultralytics/yolov3/commit/444a9f7099d4ff1aef12783704e3df9a8c3aa4b3)
- [BeeDNN - Library in C++](https://github.com/edeforas/BeeDNN)
- [Gen-EfficientNet-PyTorch](https://github.com/rwightman/gen-efficientnet-pytorch)
- [dnet](https://github.com/umangjpatel/dnet/blob/master/nn/activations.py)

All benchmarks, analysis and links to official package implementations can be found in this [repository](https://github.com/digantamisra98/Mish) 

It would be nice to have Mish as an option within the activation function group.

This is the comparison of Mish with other conventional activation functions in a SEResNet-50 for CIFAR-10:  (Better accuracy and faster than GELU)
se50_1

thebhatman:
This PR resolves https://github.com/shogun-toolbox/shogun/issues/4797. Currently I have added the forward pass computation of the Mish Function.

### DIFF
--- a/src/shogun/mathematics/linalg/backend/eigen/SpecialPurposes.cpp
+++ b/src/shogun/mathematics/linalg/backend/eigen/SpecialPurposes.cpp
@@ -157,6 +157,18 @@ void LinalgBackendEigen::rectified_linear_impl(
 		result_eig(i) = Math::max((T)0, a_eig(i));
 }
 
+template <typename T>
+void LinalgBackendEigen::mish_impl(
+    const SGMatrix<T>& a, SGMatrix<T>& result) const
+{
+	typename SGMatrix<T>::EigenMatrixXtMap a_eig = a;
+	typename SGMatrix<T>::EigenMatrixXtMap result_eig = result;
+
+	for (index_t i = 0; i < a_eig.rows() * a_eig.cols(); ++i)
+		result_eig(i) = a_eig(i) * std::tanh(std::log(1 + std::exp(a_eig(i))));
+}
+
+
 /** Eigen3 softmax method */
 template <typename T, template <typename> class Container>
 void LinalgBackendEigen::softmax_impl(Container<T>& a) const
--- a/src/shogun/neuralnets/NeuralMishLayer.cpp
+++ b/src/shogun/neuralnets/NeuralMishLayer.cpp
@@ -29,7 +29,7 @@ void NeuralMishLayer::compute_activations(SGVector<float64_t> parameters,
 	}
 }
 
-void NeuralRectifiedLinearLayer::compute_local_gradients(
+void NeuralMishLayer::compute_local_gradients(
 		SGMatrix<float64_t> targets)
 {
 	if (targets.num_rows != 0)
--- a/src/shogun/neuralnets/NeuralMishLayer.cpp
+++ b/src/shogun/neuralnets/NeuralMishLayer.cpp
@@ -1,0 +1,59 @@
+/*
+ * Copyright (c) 2014, Shogun Toolbox Foundation
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+
+ * 1. Redistributions of source code must retain the above copyright notice,
+ * this list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ *
+ * 3. Neither the name of the copyright holder nor the names of its
+ * contributors may be used to endorse or promote products derived from this
+ * software without specific prior written permission.
+
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ *
+ * Written (W) 2020 Manjunath Bhat
+ */
+
+#include <shogun/neuralnets/NeuralMishLayer.h>
+#include <shogun/mathematics/Math.h>
+#include <shogun/lib/SGVector.h>
+
+using namespace shogun;
+
+NeuralMishLayer::NeuralMishLayer() : NeuralLinearLayer()
+{
+}
+
+NeuralMishLayer::NeuralMishLayer(int32_t num_neurons):
+NeuralLinearLayer(num_neurons)
+{
+}
+
+void NeuralMishLayer::compute_activations(SGVector<float64_t> parameters,
+    const std::vector<std::shared_ptr<NeuralLayer>>& layers)
+{
+	NeuralLinearLayer::compute_activations(parameters, layers);
+
+	int32_t len = m_num_neurons*m_batch_size;
+	for (int32_t i=0; i<len; i++)
+	{
+		m_activations[i] *= std::tanh(std::log(1 + std::exp(m_activations[i])));
+	}
+}
--- a/src/shogun/neuralnets/NeuralMishLayer.cpp
+++ b/src/shogun/neuralnets/NeuralMishLayer.cpp
@@ -23,7 +23,6 @@ void NeuralMishLayer::compute_activations(SGVector<float64_t> parameters,
 {
 	NeuralLinearLayer::compute_activations(parameters, layers);
 
-	int32_t len = m_num_neurons*m_batch_size;
 	for(auto& activation: m_activations)
 	{
 		activation *= std::tanh(std::log(1 + std::exp(activation)));
--- a/src/shogun/neuralnets/NeuralMishLayer.cpp
+++ b/src/shogun/neuralnets/NeuralMishLayer.cpp
@@ -28,3 +28,27 @@ void NeuralMishLayer::compute_activations(SGVector<float64_t> parameters,
 		activation *= std::tanh(std::log(1 + std::exp(activation)));
 	}
 }
+
+void NeuralRectifiedLinearLayer::compute_local_gradients(
+		SGMatrix<float64_t> targets)
+{
+	if (targets.num_rows != 0)
+	{
+		int32_t length = m_num_neurons*m_batch_size;
+		for (int32_t i=0; i<length; i++)
+		{
+			m_local_gradients[i] = (m_activations[i]-targets[i])/m_batch_size;
+		}
+	}
+	else
+	{
+		int32_t len = m_num_neurons*m_batch_size;
+		for (int32_t i=0; i< len; i++)
+		{
+			auto e = std::exp(m_activations[i]);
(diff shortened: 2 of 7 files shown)


src/shogun/neuralnets/NeuralMishLayer.cpp
```diff
@@ -0,0 +1,59 @@
+/*
```
gf712, line 1: You can switch the copyright notice to:
```
/*
 *
 * Authors: ...
 */
```
Keeps the files less verbose

tests/unit/neuralnets/NeuralMishLayer_unittest.cc
```diff
@@ -0,0 +1,91 @@
...
+ * activations
+ */
+TEST_F(NeuralMishLayerTest, compute_activations)
+{
+	int32_t seed = 100;
+	// initialize some random inputs
+	SGMatrix<float64_t> x;
+	std::shared_ptr<NeuralInputLayer> input;
+	std::mt19937_64 prng(seed);
+	std::tie(x, input) = setup_input_layer<float64_t>(12, 3, -10.0, 10.0, prng);
```
gf712: you could use structured bindings here, `auto [x, input] = ...`, as we are now on C++17

tests/unit/neuralnets/NeuralMishLayer_unittest.cc
```diff
@@ -0,0 +1,91 @@
...
+	{
+		A_ref[idx] = A_ref[idx] * std::tanh(std::log(1 + std::exp(A_ref[idx])));
+	}
+
+	// compare
+	EXPECT_EQ(A_ref.num_rows, A.num_rows);
+	EXPECT_EQ(A_ref.num_cols, A.num_cols);
+	for (int32_t i = 0; i < A.num_rows * A.num_cols; i++)
+	{
+		EXPECT_NEAR(A_ref[i], A[i], 1e-12);
```
gf712: the 1e-12 is a bit arbitrary, maybe use std::numeric_limits<float64_t>::epsilon?

src/shogun/neuralnets/NeuralMishLayer.cpp
```diff
@@ -0,0 +1,59 @@
...
+{
+}
+
+void NeuralMishLayer::compute_activations(SGVector<float64_t> parameters,
+    const std::vector<std::shared_ptr<NeuralLayer>>& layers)
+{
+	NeuralLinearLayer::compute_activations(parameters, layers);
+
+	int32_t len = m_num_neurons*m_batch_size;
+	for (int32_t i=0; i<len; i++)
```
gf712: you could use a ranged loop here no?
```cpp
for (auto& activation: m_activations)
{
  activation *= std::tanh(std::log(1 + std::exp(activation)));
}

Comment (gf712) [reply src/shogun/neuralnets/NeuralMishLayer.cpp]: or even better, std::transform with the same input and output iterators, that way it is clear that you are changing some values

File: tests/unit/neuralnets/NeuralMishLayer_unittest.cc (hunk 2 above)
Comment (gf712) [reply tests/unit/neuralnets/NeuralMishLayer_unittest.cc]: also maybe create some variables for those parameters, and use those instread of x.num_cols, layer->get_num_neurons(), and so on.. It just makes it easier to read the code imo

File: tests/unit/neuralnets/NeuralMishLayer_unittest.cc [hunk 5]
```diff
@@ -0,0 +1,94 @@
...
+
+/** Compares the activations computed using the layer against manually computed
+ * activations
+ */
+TEST_F(NeuralMishLayerTest, compute_activations)
+{
+	int32_t seed = 100;
+	// initialize some random inputs
+	SGMatrix<float64_t> x;
+	std::shared_ptr<NeuralInputLayer> input;
```
gf712: these lines can go no?
```
SGMatrix<float64_t> x;
std::shared_ptr<NeuralInputLayer> input;
```

src/shogun/neuralnets/NeuralMishLayer.cpp
```diff
@@ -0,0 +1,31 @@
...
+NeuralLinearLayer(num_neurons)
+{
+}
+
+void NeuralMishLayer::compute_activations(SGVector<float64_t> parameters,
+    const std::vector<std::shared_ptr<NeuralLayer>>& layers)
+{
+	NeuralLinearLayer::compute_activations(parameters, layers);
+
+	int32_t len = m_num_neurons*m_batch_size;
```
gf712: this is not needed anymore

tests/unit/neuralnets/NeuralMishLayer_unittest.cc
thebhatman: Yeah, thanks for pointing out. Will change it.

gf712: no worries, sorry for being so pedantic :D

thebhatman:
Also should I implement a custom gradient for Mish? Can you give me some pointers on how I should be implementing the gradient?

gf712:
> Also should I implement a custom gradient for Mish? Can you give me some pointers on how I should be implementing the gradient?

I am not sure how to do it, I have never written neural net stuff for shogun. But we don't have any support for autodiff right now so I am assuming you write it manually, like for softmax:
https://github.com/shogun-toolbox/shogun/blob/7db6e4ef151ed94154ec5db326eedf7fb7368ba4/src/shogun/neuralnets/NeuralSoftmaxLayer.h#L74-L84

maybe @karlnapf knows what needs to be done?

thebhatman:
Thank you! I have added local gradients for Mish. Currently I have defined gradients for when Mish is used in the hidden layers. At the output layer, I haven't changed the gradient as in both ReLU and Softmax, the gradient at output layer was same and Mean squared loss was assumed.. Please let me know if I should change the gradient at output layer.

src/shogun/neuralnets/NeuralMishLayer.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+void NeuralMishLayer::compute_local_gradients(
+		SGMatrix<float64_t> targets)
+{
+	if (targets.num_rows != 0)
+	{
+		int32_t length = m_num_neurons*m_batch_size;
+		for (int32_t i=0; i<length; i++)
+		{
+			m_local_gradients[i] = (m_activations[i]-targets[i])/m_batch_size;
```
karlnapf, line 40: this should be vectorized

karlnapf: also this code seems to be re-usable across different activation functions ... why have it down in here?

src/shogun/neuralnets/NeuralMishLayer.cpp
```diff
@@ -0,0 +1,54 @@
...
+		}
+	}
+	else
+	{
+		int32_t len = m_num_neurons*m_batch_size;
+		for (int32_t i=0; i< len; i++)
+		{
+			auto e = std::exp(m_activations[i]);
```
karlnapf, line 48: make sure to put a unit test that checks for correctness of this manually implemented gradient on a few simple non trivial cases...

karlnapf: again, could this be vectorized?

karlnapf (commented):
Thanks for the patch! I think it could be made a bit faster via avoiding explicit.

src/shogun/neuralnets/NeuralMishLayer.cpp
thebhatman: I had a doubt whether I had to implement gradient at the output layer separately. The other activation functions also had this. So I put it there. I'll remove it if it's not required.

src/shogun/neuralnets/NeuralMishLayer.cpp
thebhatman: I'm not really sure how to vectorize this. It would be really helpful if you could point me to a similar example where vectorized operations have been used....

gf712: by vectorized he means you could/should use Eigen. You can map the memory of the matrix to an eigen matrix and use their implementation of coefficient wise transformation.
Look for example at 
https://github.com/shogun-toolbox/shogun/blob/895760f011c5833cd290192f37bdef1457e0ebf8/src/shogun/mathematics/linalg/backend/eigen/BasicOps.cpp#L233-L243
This will compile with SIMD instructions if they are available and inline properly

thebhatman: I am thinking of adding an implementation of Mish here; something similar to this : https://github.com/shogun-toolbox/shogun/blob/7db6e4ef151ed94154ec5db326eedf7fb7368ba4/src/shogun/mathematics/linalg/backend/eigen/SpecialPurposes.cpp#L162

thebhatman: Would this be correct?

gf712: yes, you could do that!

karlnapf:
What's the state of this thing? Sorry for being a bit slow on our side atm

src/shogun/mathematics/linalg/backend/eigen/SpecialPurposes.cpp
```diff
@@ -157,6 +157,18 @@ void LinalgBackendEigen::rectified_linear_impl(
...
 
+template <typename T>
+void LinalgBackendEigen::mish_impl(
+    const SGMatrix<T>& a, SGMatrix<T>& result) const
+{
+	typename SGMatrix<T>::EigenMatrixXtMap a_eig = a;
+	typename SGMatrix<T>::EigenMatrixXtMap result_eig = result;
+
+	for (index_t i = 0; i < a_eig.rows() * a_eig.cols(); ++i)
+		result_eig(i) = a_eig(i) * std::tanh(std::log(1 + std::exp(a_eig(i))));
```
gf712, line 168: can you not write this as something like `result_eig = a_eig.array() * ((1 + a_eig.exp()).log()).tanh()`? That way you give eigen the best chance to vectorise this.

karlnapf:
Hi! Shall we try to finish this up? :)

digantamisra98:
Yeah @karlnapf. Sorry for the inactivity. Let me catch up with the commits again.

src/shogun/neuralnets/NeuralMishLayer.cpp
YashasSamaga: `std::log1p` instead of `std::log(1 + std::exp)` is better numerical accuracy but is still poor.

OpenCV CPU backend has nearly 2x faster and even more accurate [implementation](https://github.com/opencv/opencv/blob/4564b8a224545a54c545e9d2743cce3c0e1f1c5b/modules/dnn/src/layers/elementwise_layers.cpp#L760-L773).

src/shogun/mathematics/linalg/backend/eigen/SpecialPurposes.cpp
YashasSamaga: `log1p` instead of `log(1 + exp(x))` is better for numerical accuracy.

OpenCV CPU backend has a much faster (more than 2x) and accurate [implementation](https://github.com/opencv/opencv/blob/4564b8a224545a54c545e9d2743cce3c0e1f1c5b/modules/dnn/src/layers/elementwise_layers.cpp#L760-L773).

src/shogun/neuralnets/NeuralMishLayer.cpp
gf712: do you know what the accuracy tradeoff is?

src/shogun/mathematics/linalg/backend/eigen/SpecialPurposes.cpp
gf712: do you know what the accuracy tradeoff is?

YashasSamaga: Errors callculated against values computed using `tanh(log1p(exp(x)))` with 128-bit floats.

errors

YOLOv4 performance almost doubles: https://github.com/opencv/opencv/pull/17624

karlnapf: we could have the approximation somehwere, but it should be marked as such
(and I agree with @gf712 this can be vectorized, even the approximation)

PR #4815, opened 2020-02-01
Ruby nmatrix

vigsterkr:
This allows us to integrate with SciRuby stack, see the example notebook here: https://gist.github.com/vigsterkr/48a5f0523528bbab85bec04464ca2b6c

- [x] define typemaps for SGVector and SGMatrix
- [ ] fix SGMatrix -> NMatrix conversion: SGMatrix is column major while NMatrix is row-major, hence the `rb_nmatrix_dense_create` will not give the right output.
- [ ] Fix unsigned type (see FIXME's) mapping
- [ ] allow switching between `narray` typemap and `nmatrix` (compile time, i.e. cmake option)
- [ ] change meta example generator for ruby to use NMatrix instead of NArray


vigsterkr:
in order to get this fully working we still need #4177 fixed

karlnapf:
bump!

PR #4816, opened 2020-02-05
python style enumerate

gf712:
Python style `enumerate`, because why shouldn't C++ look like Python:

```cpp
for (const auto& [idx, label_i] : enumerate(*m_labels))
{
    auto c = static_cast<index_t>(label_i);
    centered_class[c].set_column(
	centered_class_col[c], mean_matrix.get_column(idx));
    ++centered_class_col[c];
}
```


vigsterkr:
👍 imo the macos build failure is unrelated - we have some flakiness there lately

vigsterkr (approved):

karlnapf:
Like! :)

gf712:
> +1 imo the macos build failure is unrelated - we have some flakiness there lately

yea i think so too..

PR #4818, opened 2020-02-12
simple draft of Linalg graph [WIP]

gf712:
very simple non working implementation as PoC

Need to:
- [ ]  figure out abstraction of data (Tensor?)
- [ ]  ...

In a way this is linalg with classes.


src/shogun/mathematics/graph/Types.h
```diff
@@ -0,0 +1,70 @@
...
+#ifndef SHOGUNTYPES_H_
+#define SHOGUNTYPES_H_
+
+#ifdef USE_NGRAPH
+#include <ngraph/ngraph.hpp>
+#endif
+
+namespace shogun
+{
+	enum class element_type
```
vigsterkr, line 16: arrow has a quite nice abstraction over these things:
https://github.com/apache/arrow/blob/master/cpp/src/arrow/type.h#L233
https://github.com/apache/arrow/blob/master/cpp/src/arrow/type.h#L460
https://github.com/apache/arrow/blob/master/cpp/src/arrow/type.h#L460

gf712:
@vigsterkr @karlnapf  what about an interface like this:
```cpp
auto X = SGVector<float64_t>(10);

auto tensor = std::make_shared<Tensor>(Shape{Shape::Dynamic}, element_type::float64);

auto input = std::make_shared<Input>(tensor);

auto output = std::make_shared<Add>(input, input);
// or with the operator overload
// auto output = input + input;

auto graph = std::make_shared<Graph>(input, output);

graph->build();
std::vector<std::shared_ptr<Result>> result = graph->eval(X);

// equivalent of auto shogun_result = X + X;
auto shogun_result = result[0]->as<SGVector<float64_t>>();

// for more complex graphs you can also fetch the tensors from other nodes
// after the graph is evaluated
auto input_tensor  = input->get_tensor();
assert(tensor == input_tensor);

auto graph = std::make_shared<Graph>(input, {input_tensor, output});
graph->build();
std::vector<std::shared_ptr<Result>> result = graph->eval(X);
shogun_result = result[1]->as<SGVector<float64_t>>();
assert(X == shogun_result)
```

For the interfaces Tensor can have a ->as_vector() and ->as_matrix() (or just ->as_array()) and the stored type is used when copying for example to a numpy array. Or can have a visitor pattern, but it might be overkill because this will only support a dozen of numeric types.

src/shogun/mathematics/graph/Types.h
gf712: Thanks I'll have a look!

src/shogun/io/fmt/fmt.h
```diff
@@ -3,8 +3,10 @@
 
 #if defined(USE_EXTERNAL_SPDLOG)
 #include <spdlog/fmt/fmt.h>
+#include <spdlog/fmt/bundled/ostream.h>
 #else
 #include <shogun/third_party/spdlog/fmt/fmt.h>
+#include <shogun/third_party/spdlog/fmt/bundled/ostream.h>
```
gf712, line 9: @vigsterkr is it ok to add the ostream.h? Makes it easier to format stuff with ftm, e.g. `format("{}", object_instance)`, where object instance has operator<< overloaded

vigsterkr: well lets have it in for the time being and we'll see later... if this helps to have a faster dev cycle, i dont see why we shouldnt have it :)

src/shogun/base/class_list.h
```diff
@@ -46,7 +46,7 @@ namespace shogun {
 		auto entry = clazzes.find(name);
 		if (entry != clazzes.end())
 		{
-			auto class_factory = entry->second().class_by_name<T>(name);
+			auto class_factory = entry->second().template class_by_name<T>(name);
```
gf712, line 49: @vigsterkr this was causing a compiler error

src/shogun/mathematics/graph/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
...
+set(SHOGUN_GRAPH_SOURCES
+	Graph.cpp
+	Tensor.cpp
+	ops/Add.cpp
+	ops/Input.cpp
+	)
+
+add_library(shogun-graph SHARED ${SHOGUN_GRAPH_HEADERS} ${SHOGUN_GRAPH_SOURCES} ${SHOGUN_CORE_HEADERS} ${CMAKE_BINARY_DIR}/src/shogun/lib/config.h)
+
+target_link_libraries(shogun-graph shogun-core)
```
gf712, line 51: @vigsterkr as you can see I don't have great cmake skills.
Not sure if each folder/module should have it's own CMakeLists.txt?
Also I had to change the path of SHOGUN_CORE_HEADERS files to an absolute path

vigsterkr: yeah i was trying to do this myself with kernels, features but locally that was a failure, but imo that would the best that actually each subfolder has it's own cmakelists.txt

src/shogun/mathematics/graph/Graph.cpp
```diff
@@ -0,0 +1,185 @@
...
+{
+	auto unordered_nodes = check_fully_connected(m_inputs, m_outputs);
+	build_backend_graph(unordered_nodes);
+}
+
+std::map<std::shared_ptr<Node>, Graph::STATUS> Graph::check_fully_connected(
+    const std::vector<std::shared_ptr<Input>>& inputs,
+    const std::vector<std::shared_ptr<Node>>& outputs)
+{
+	std::deque<std::shared_ptr<Node>> nodes_to_check;
```
vigsterkr: couldn't we simply use the iterator ctor of std::deque?
```
std::deque<std::shared_ptr<Node>> nodes_to_check(outputs.begin(), outputs.end())
```

src/shogun/mathematics/graph/Graph.cpp
```diff
@@ -0,0 +1,185 @@
...
+	for (const auto& node : outputs)
+	{
+		nodes_to_check.push_back(node);
+	}
+	std::unordered_set<std::shared_ptr<Node>> inputs_found;
+	std::map<std::shared_ptr<Node>, STATUS> unordered_nodes;
```
vigsterkr: could we use std::unordered_map?

src/shogun/mathematics/graph/Graph.cpp
```diff
@@ -0,0 +1,185 @@
...
+	switch (env->graph_backend())
+	{
+	case GRAPH::SHOGUN:
+	{
+		execute_shogun();
+	}
+	break;
+	case GRAPH::NGRAPH:
+	{
+#ifdef USE_NGRAPH
```
vigsterkr, line 77: i guess later we'll want to have this actually be runtime, so that you dont need ngraph to be present during compilation

src/shogun/mathematics/graph/Tensor.cpp
```diff
@@ -0,0 +1,15 @@
...
+#include <shogun/io/fmt/fmt.h>
+
+using namespace shogun;
+
+std::string Tensor::to_string() const
+{
+	return fmt::format("Tensor(shape={}, type={})", m_shape, m_type);
+}
+
+template Tensor::Tensor(const SGVector<float32_t>&);
```
vigsterkr, line 12: since the only function atm of Tensor in the implementation part is to_string don't you think that header only Tensor (all the implementation is there) would be better?

src/shogun/mathematics/graph/ops/Add.h
```diff
@@ -0,0 +1,274 @@
...
+			// or just use Eigen here
+			std::transform(
+			    static_cast<const T*>(input1),
+			    static_cast<const T*>(input1) + size,
+			    static_cast<const T*>(input2), static_cast<T*>(output),
+			    std::plus<T>());
+		}
+	};
+
+#ifdef USE_NGRAPH
```
vigsterkr, line 147: i guess this could be moved to a separate file and directory that is only related to ngraph and then have a shared lib defined like: shogun-ngraph.so

src/shogun/mathematics/graph/ops/Add.h
```diff
@@ -0,0 +1,274 @@
...
+
+	template <typename EngineImplementation>
+	IGNORE_IN_CLASSLIST class AddImpl: public OperationImpl
+	{
+	public:
+		AddImpl(const std::shared_ptr<Node>& node): OperationImpl(node) {}
+
+		virtual ~AddImpl() {}
+
+		template<typename T>
```
vigsterkr: is this being used somewhere? i mean the template param?

src/shogun/mathematics/graph/ops/Add.h
```diff
@@ -0,0 +1,274 @@
...
+		void evaluate()
+		{
+			this->evaluate_implementation();
```
vigsterkr: although in case of using pointer it's not as nice but if we dont need any template param then we could have:
```
void operator()()
```

src/shogun/mathematics/graph/ops/Input.cpp
```diff
@@ -0,0 +1,9 @@
+#include <shogun/mathematics/graph/ops/Input.h>
+
+using namespace shogun;
+
+// so this is a thing
+// template<> template<> void InputImpl<ShogunInput>::evaluate(const SGVector<float32_t>&);
```
vigsterkr: do we need these, i mean since the header has everything...

vigsterkr (commented):
some questions and comments

vigsterkr:
mmm sorry about not having this branch rebased on develop... do you want me to do that and then we can have this thing a bit less about rebase and more about graph...? or shall we just merge this into the branch, because honestly i'd be happy to chip in some parts of the graph development... for example i could help out to get this thing changed that we remove all the #ifdef NGRAPH macros and things are actually done runtime.

gf712:
sure, just merge it. Could you have a look at the manifest stuff btw? I added it and should be available with `available_operators()`. A lot is copy paste from the library stuff, so might give an idea of what can be abstracted away..

src/shogun/mathematics/graph/ops/Input.cpp
gf712: Not anymore, I figured tensor should be the type erase, any shape, representation.

src/shogun/mathematics/graph/CMakeLists.txt
```diff
@@ -1,29 +1,62 @@
...
+# 	)
+
+add_library(shogun-graph SHARED ${SHOGUN_GRAPH_HEADERS} 
+	${SHOGUN_GRAPH_SOURCES} 
+	${SHOGUN_CORE_HEADERS} 
+	${SHOGUN_CORE_GRAPH_SOURCES}
+	${SHOGUN_CORE_GRAPH_HEADERS}
+	${SHOGUN_GRAPH_IMPLEMENTATION_HEADERS}
+	${SHOGUN_GRAPH_IMPLEMENTATION_SOURCES}
+	${CMAKE_BINARY_DIR}/src/shogun/lib/config.h)
```
vigsterkr: imo this should be coming from shogun-core...

src/shogun/mathematics/graph/CMakeLists.txt
```diff
@@ -1,29 +1,62 @@
...
+# set(NGRAPH_GRAPH_IMPLEMENTATION_HEADERS
+# 	ops/ngraph/Add.h
+# 	)
+# set(NGRAPH_GRAPH_IMPLEMENTATION_SOURCES
+# 	ops/ngraph/Add.cpp
+# 	)
+
+add_library(shogun-graph SHARED ${SHOGUN_GRAPH_HEADERS} 
+	${SHOGUN_GRAPH_SOURCES} 
+	${SHOGUN_CORE_HEADERS}
```
vigsterkr: do we need this? shouldn't a `target_link_libraries(shogun-graph shogun-core)` solve this?

src/shogun/mathematics/graph/ops/Add.h
gf712: true, but then there's also build() need to clean up all these adhoc functions

src/shogun/mathematics/graph/Graph.cpp
gf712: ah, not used to  using deque, didn't realise there was this!

src/shogun/mathematics/graph/Graph.cpp
gf712: can be and should be

vigsterkr:
@gf712 ok i'll merge then this as is and i'll work on the dynamic loading of the graph lib (and on the way cleaning up the code redundancy part)... as well as try to come up with something with cmake to have a shogun-graph, shogun-ngraph shared libs....

gf712:
@vigsterkr This is what I am using to compile a simple test and what I am working towards:
```cpp
#include <shogun/mathematics/graph/Graph.h>
#include <shogun/mathematics/graph/ops/Add.h>
#include <shogun/mathematics/graph/ops/Input.h>

#include <shogun/mathematics/graph/operator_list.h>

#include <iostream>

using namespace shogun;


int main()
{
	available_operators();

	auto X = SGVector<float64_t>(10);

	auto input = std::make_shared<Input>(Shape{Shape::Dynamic}, element_type::FLOAT64);
	auto input1 = std::make_shared<Input>(Shape{2}, element_type::FLOAT64);

	auto intermediate = std::make_shared<Add>(input, input);

	std::cout << intermediate->get_tensors()[0] << '\n';
	std::cout << *intermediate << '\n';

	auto output = std::make_shared<Add>(input, input1);

	std::cout << output->get_tensors()[0] << '\n';
	std::cout << *output << '\n';

	auto graph = std::make_shared<Graph>(std::vector{input, input1}, std::vector<std::shared_ptr<Node>>{intermediate, output});

	graph->build();

	/*std::vector<std::shared_ptr<Tensor>> result =*/ graph->evaluate(std::vector{std::make_shared<Tensor>(X)});

	// auto shogun_result = result[0]->as<SGVector<float64_t>(); 
}
```

vigsterkr:
ok i'll make sure that by the end of my changes you dont need to call the `available_operators()` function and that it's all done automagically within shogunenv... we should add imo a unittest (say the one you've just wrote) and basically get that unit test work with different backends

src/shogun/mathematics/graph/CMakeLists.txt
gf712: yup it does, fixed this

src/shogun/mathematics/graph/Tensor.cpp
gf712: this was so that fmt wouldn't become a runtime dependency. But I guess Tensor is used in some other .cpp files

src/shogun/mathematics/graph/ops/Add.h
gf712: indeed, I started doing that `graph/ops/shogun/OperatorName.h`, `graph/ops/ngraph/OperatorName.h` and so on

gf712:
> ok i'll make sure that by the end of my changes you dont need to call the `available_operators()` function and that it's all done automagically within shogunenv... we should add imo a unittest (say the one you've just wrote) and basically get that unit test work with different backends

the available_operators() is just a sanity check. But yea all the runtime loading in the graph building is to be implemented

src/shogun/mathematics/graph/Tensor.cpp
vigsterkr: mmm ok i see... lets keep it as is and then we'll refactor lets just push this to a level that things are functional and then we cleanup on the end imo

PR #4820, opened 2020-02-14
add operator lookup

gf712:
Aim of this PR:

```cpp
auto X1 = SGVector<TypeParam>(10);
auto X2 = SGVector<TypeParam>(10);

X1.range_fill();
X2.range_fill();

auto expected_result1 = X1 + X2;
auto expected_result2 = expected_result1 + X2;


auto input = make_shared<Input>(Shape{Shape::Dynamic}, get_enum_from_type<TypeParam>::type);
auto input1 = make_shared<Input>(Shape{10}, get_enum_from_type<TypeParam>::type);

auto intermediate = make_shared<Add>(input, input);

auto output = make_shared<Add>(intermediate, input1);

auto graph = make_shared<Graph>(vector{input, input1}, vector<shared_ptr<Node>>{intermediate, output});

graph->build();

vector<shared_ptr<Tensor>> result = graph->evaluate(vector{make_shared<Tensor>(X1), make_shared<Tensor>(X2)});

auto result1 = result[0]->as<SGVector<TypeParam>>();
auto result2 = result[1]->as<SGVector<TypeParam>>();
```


src/shogun/mathematics/graph/Graph.cpp
```diff
@@ -138,8 +140,9 @@ void Graph::add_operator_node(const std::shared_ptr<Node>& node)
 	case GRAPH::XLA:
 	case GRAPH::TVM:
 	case GRAPH::SHOGUN:
-		break;
+		op = create_operator<OperatorShogunBackend>(std::string(node->get_operator_name()));
 	}
+	op->build(node);
```
gf712, line 175: @vigsterkr this adds the link between the "abstract representation" of the graph with the implementation

src/shogun/mathematics/graph/ops/shogun/Input.h
```diff
@@ -16,14 +16,19 @@ namespace shogun {
 			return "Input";
 		}
 
+		void evaluate() override
```
gf712: @vigsterkr let me know if you can think of a better way to feed data to the graph.
Basically the input node needs to be fed data, which comes from SGContainer. So it has to be different from the other nodes which can just grab data from the previous node

PR #4821, opened 2020-02-15
node and operator interface

gf712:
Start thinking about how to write some abstractions
- [ ] Unary operators
- [ ] Binary operators
- [ ] ngraph backend


src/shogun/mathematics/graph/ops/abstract/BinaryOperator.h
```diff
@@ -0,0 +1,95 @@
...
+
+#ifndef SHOGUNBINARYOPERATOR_H_
+#define SHOGUNBINARYOPERATOR_H_
+
+#include <shogun/mathematics/graph/OperatorImplementation.h>
+
+namespace shogun {
+
+	template <typename DerivedOperator, typename EngineImplementation>
+	class BinaryOperator : public OperatorImpl<EngineImplementation, BinaryOperator<DerivedOperator, DerivedOperator>>
```
gf712, line 15: This should be BinaryOperato<DerivedOperator, EngineOperator>

vigsterkr:
@gf712 i've took this patch and added on top my fixes.. it's already in the branch... now the operator are working... we'll have to discuss though some stuff :)

gf712:
OK! The issue was that I wasn’t passing the node type right? Are you working on unit test too then?

vigsterkr:
@gf712 yeah that was it.... the Add_test is there as an example, it's a bit shitty atm with different backends but it works...

PR #4822, opened 2020-02-15
refactor graph executor dynamic loading

vigsterkr:
change shape size type to int64_t


src/shogun/mathematics/graph/Graph.cpp
```diff
@@ -68,33 +67,12 @@ std::unordered_map<std::shared_ptr<Node>, Graph::STATUS> Graph::check_fully_conn
 
 std::vector<std::shared_ptr<Tensor>> Graph::evaluate(const std::vector<std::shared_ptr<Tensor>>& tensors)
 {
-	if (m_cached_operators.empty())
+	if (!m_executor)
 	{
-		error("Did you call Graph::build()?");
+		error("Graph has not been built!");
 	}
```
gf712: @vigsterkr so would this fallback to the default? so instead of an error just have a warning?

gf712 (approved):
Looks good to me

PR #4823, opened 2020-02-16
Add vcpkg installation instructions

LilyWangL:
Shogun is available as a port in vcpkg, a C++ library manager that simplifies installation for shogun and other project dependencies. Documenting the install process here will help users get started by providing a single set of commands to build shogun, ready to be included in their projects.

We also test whether our library ports build in various configurations (dynamic, static) on various platforms (OSX, Linux, Windows: x86, x64, UWP, ARM) to keep a wide coverage for users.

I'm a maintainer for vcpkg, and here is what the port script looks like. We try to keep the library maintained as close as possible to the original library.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,18 @@ Note that some folders are submodules that can be checked out with
 - *benchmarks* - speed benchmarks
 - *cmake* - cmake build scripts
 
+## Building shogun - Using vcpkg
+
+You can download and install shogun using the [vcpkg](https://github.com/Microsoft/vcpkg) dependency manager:
+
+    git clone https://github.com/Microsoft/vcpkg.git
+    cd vcpkg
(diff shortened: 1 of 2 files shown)


vigsterkr:
@LilyWangL thnx heaps for adding this! could you by any chance add this section to https://github.com/shogun-toolbox/shogun/blob/develop/doc/readme/INSTALL.md instead of README.md and just forcepush that change into your branch, so then i can just merge it right away!

LilyWangL:
> @LilyWangL thnx heaps for adding this! could you by any chance add this section to https://github.com/shogun-toolbox/shogun/blob/develop/doc/readme/INSTALL.md instead of README.md and just forcepush that change into your branch, so then i can just merge it right away!

Done.

PR #4824, opened 2020-02-16
ngraph ops implementation

gf712:
Can now switch between vanilla shogun and ngraph backends


src/shogun/mathematics/graph/ops/ngraph/Add.h
```diff
@@ -28,6 +28,15 @@ namespace shogun {
...
 		{
 		}	
+
+		void build_implementation() final
+		{
+			// TODO
+			const auto& node1 = this->m_abstract_node->get_input_nodes()[0];
+			const auto& node2 = this->m_abstract_node->get_input_nodes()[1];
+
+			std::make_shared<ngraph::op::Add>(node1, node2);
```
gf712: @vigsterkr somehow the nodes need to have access to the ngraph::Node. Which means then that the Node class in Shogun needs to be aware of the operator implementation? Or there is a NGraphNode that can be casted to?

src/shogun/mathematics/graph/NGraph.cpp
```diff
@@ -1,62 +1,126 @@
...
+	}
+	return op_it->second();
 }
 
-void NGraph::add_input_operator(const std::shared_ptr<Node>& node)
+void NGraph::add_input_operator(const std::shared_ptr<node::Node>& node)
 {
+	
+	auto input = get_operator(node);
+	m_lookup[node] = std::static_pointer_cast<detail::ngraph::InputNGraph>(input)->build_input(node);
```
gf712, line 114: @vigsterkr unfortunately ngraph can't infer the shape at execution time from dynamic. So need to the building of the ngraph graph at execution. So need change the top Graph abstraction..

gf712: nvm, just have to call a different op::Parameter constructor and create dynamic tensors

gf712:
@vigsterkr tests pass on both backends!

vigsterkr:
great work! 💃

PR #4826, opened 2020-02-18
Binary ops

gf712:
- [x] Subtract
- [x] Multiply
- [x] Divide
- [x] Logical operators interfaces TODO:
  - [x] add boolean type
  - [x] add binary logical operator -> returns boolean
- [x] AND
- [X] OR
- [X] XOR
- [x] Equals


src/shogun/mathematics/graph/CMakeLists.txt
```diff
@@ -90,8 +95,13 @@ IF (ENABLE_TESTING)
...
 	ADD_SHOGUN_TEST(nodes/Add_test)
+	ADD_SHOGUN_TEST(nodes/Subtract_test)
+
 	if (UNIX AND NOT APPLE)
 		set_property(TEST Add_test PROPERTY ENVIRONMENT "LD_LIBRARY_PATH=${PROJECT_BINARY_DIR}/src/shogun")
+		set_property(TEST Subtract_test PROPERTY ENVIRONMENT "LD_LIBRARY_PATH=${PROJECT_BINARY_DIR}/src/shogun")
 	endif()
 	add_dependencies(Add_test ${EXCUTORS})
+	add_dependencies(Subtract_test ${EXCUTORS})
+
```
gf712, line 139: @vigsterkr I am guessing this block could become a function/macro?

src/shogun/mathematics/graph/CMakeLists.txt
```diff
@@ -6,6 +6,7 @@ set(SHOGUN_CORE_GRAPH_HEADERS
 	${CMAKE_CURRENT_SOURCE_DIR}/Tensor.h
 	${CMAKE_CURRENT_SOURCE_DIR}/Types.h
 	${CMAKE_CURRENT_SOURCE_DIR}/nodes/Add.h
+	${CMAKE_CURRENT_SOURCE_DIR}/nodes/Subtract.h
```
gf712, line 15: @vigsterkr  I know it is meant to be bad practice in cmake, but maybe we should replace some of these with glob?

vigsterkr: glob is crazy.... but of course we can do it! we are doing it atm, so it's its all fine

src/shogun/mathematics/graph/CMakeLists.txt
vigsterkr: yep same as all the library setter magic for abi version etc.

src/shogun/mathematics/graph/Tensor.h
```diff
@@ -97,7 +100,8 @@ namespace shogun
 				if (m_shape.size() != shape.size())
 				{
 					error(
-					    "Mismatch in the number of dimensions, expected {}, but "
+					    "Mismatch in the number of dimensions, expected {}, "
+					    "but "
 					    "got {}",
 					    m_shape.size(), shape.size());
```
gf712, line 106: clang-format at it's best...

gf712:
@vigsterkr I think this is good to go. Would rather do the cmake stuff in a separate PR

PR #4827, opened 2020-02-18
added most "normal" types

gf712:
In ngraph we cannot use boolean types for any non-logical operation, which I guess makes sense and we should do the same


vigsterkr:
i'll merge this now, but i'll look into having something in shogun-core similar to what's in arrow that i've mentioned as we keep having these enums and stuff for primitive types all around shogun and would be great to have it unified - and use a more c++ like approach instead of ansi c, see enum

PR #4829, opened 2020-02-19
add helper function for binary operator testing

vigsterkr:


src/shogun/mathematics/graph/test/GraphTest.h
```diff
@@ -12,6 +13,37 @@ class GraphTest : public ::testing::Test
...
+		const shogun::SGVector<T>& expected_result1,
+		const shogun::SGVector<T>& expected_result2)
+	{
+		for (auto&& backend : this->m_backends)
+		{
+			graph->build(backend);
+
+			std::vector<std::shared_ptr<shogun::graph::Tensor>> result = graph->evaluate(
+				std::vector{std::make_shared<shogun::graph::Tensor>(X1),
+					std::make_shared<shogun::graph::Tensor>(X2)});
```
gf712, line 28: it would be good to check the result size, as it should be always be 2 in a binary op.

gf712 (changes requested):

src/shogun/mathematics/graph/test/GraphTest.h
gf712: ah no, not because it is a binary, but because it has two outputs. So not sure if there is anything else that can be checked here then

PR #4830, opened 2020-02-19
dot op implementation

gf712:
Transpose and reshape is used for column to row major conversion.
  - [x] add scalar dot 
  - [x] vector/matrix dot
  - [x] add "optimisation" where graphs with only coefficient wise computations don't convert column to row major

Next PR (out of scope for this PR):
  - add/finish MatMul and MatMulBias implementation
  -  add "optimisation" where ngraph uses MatMul to do the transpose in the backend

### DIFF
--- a/src/shogun/mathematics/graph/NGraph.cpp
+++ b/src/shogun/mathematics/graph/NGraph.cpp
@@ -36,7 +36,7 @@ std::vector<std::shared_ptr<Tensor>> NGraph::execute(
 	for (const auto& tensor : tensors)
 	{
 		const auto& shape = tensor->get_shape();
-		if (tensor->get_shape().size() < 2)
+		if (tensor->get_shape().size() < 2 || !m_requires_major_conversion)
 		{
 			auto& input =
 			    ngraph_input_tensors.emplace_back(backend->create_tensor(
@@ -44,7 +44,7 @@ std::vector<std::shared_ptr<Tensor>> NGraph::execute(
 			        to_ngraph_shape(shape)));
 			input->write(tensor->data(), tensor->size_in_bytes());
 		}
-		else if (tensor->get_shape().size() == 2)
+		else if (tensor->get_shape().size() == 2 && m_requires_major_conversion)
 		{
 			const auto ngraph_shape = to_ngraph_shape(shape);
 			auto& input =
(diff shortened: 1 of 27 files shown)


src/shogun/mathematics/graph/NGraph.cpp
```diff
@@ -10,12 +10,55 @@
...
+
+inline void row_to_column_major(void* dst, void* src, size_t type_size, size_t rows, size_t cols)
+{
+	size_t idx = 0;
+	for (size_t i = 0; i < rows; i++)
+	{
+		for (int j = 0; j < cols; ++j)
+		{
+			size_t new_idx = columnmajorindexfromrowmajorindex(idx, cols, rows);
+			memcpy((void*)&((char*)dst)[new_idx*type_size], (void*)&((char*)src)[idx*type_size], type_size);
```
gf712: @vigsterkr looks ugly but does the job... I can't be bothered to write a switch statement to cast to the correct type...
In any case we need to think more long term how to handle conversion from column to row major, this doesn't scale well.

src/shogun/mathematics/graph/NGraph.cpp
```diff
@@ -10,12 +10,55 @@
...
+			idx++;
```
gf712: this could be idx+=type_size but let's assume the compiler knows that too..

src/shogun/mathematics/graph/nodes/Dot_test.cc
```diff
@@ -0,0 +1,179 @@
...
+
+			vector<shared_ptr<Tensor>> result = graph->evaluate(
+			    vector{make_shared<Tensor>(X1), make_shared<Tensor>(X2)});
+
+			auto result1 = result[0]->as<SGVector<NumericType>>();
+
+			for (const auto& [expected_i, result_i] : zip_iterator(expected_result, result1))
+			{
+				EXPECT_EQ(expected_i, result_i);
+			}
```
gf712, line 136: @vigsterkr and then somehow this actually works!

src/shogun/mathematics/graph/nodes/Dot_test.cc
```diff
@@ -0,0 +1,179 @@
...
+		auto output = make_shared<node::Dot>(A, B);
+
+		auto graph = make_shared<Graph>(
+		    vector{A, B},
+		    vector<shared_ptr<node::Node>>{output});
+
+		for (auto&& backend : this->m_backends)
+		{
+			if (backend == GRAPH_BACKEND::SHOGUN)
+				continue;
```
gf712: didn't implement shogun operator yet

PR #4832, opened 2020-02-20
fix Tensor template instantiation

gf712:
This should fix the msvc compiler error in the Tensor class template instantiation


gf712:
@vigsterkr seems to have fix the template instantiation error

PR #4834, opened 2020-02-21
Add MeanSquared unittest

LiuYuHui:
Add MeanSquaredError/MeanAbsoluteError/MeanSquaredLogError unittest

### DIFF
--- a/tests/unit/evaluation/MeanSquared_unittest.cc
+++ b/tests/unit/evaluation/MeanSquared_unittest.cc
@@ -20,10 +20,12 @@ public:
 	std::shared_ptr<BinaryLabels> ground_truth_labels;
 	std::shared_ptr<RegressionLabels> prediect_reg_labels;
 	std::shared_ptr<RegressionLabels> gtruth_reg_labels;
+	float64_t eps;
 
 protected:
 	void SetUp() override
 	{
+		eps = std::numeric_limits<float64_t>::epsilon();
 		float64_t threshold = 0.5;
 		SGVector<float64_t> prediected_prob = {0.1, 0.4, 0.6, 0.9};
 		SGVector<float64_t> gtruth_prob = {0.1, 0.4, 0.4, 0.9};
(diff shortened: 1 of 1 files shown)


tests/unit/evaluation/MeanSquared_unittest.cc
```diff
@@ -0,0 +1,66 @@
...
+		prediect_reg_labels =
+		    std::make_shared<RegressionLabels>(prediected_reg);
+		gtruth_reg_labels = std::make_shared<RegressionLabels>(gtruth_reg);
+	}
+};
+
+TEST_F(MeanSquaredTest, MeanSquaredError)
+{
+	auto MSE = std::make_shared<MeanSquaredError>();
+	EXPECT_NEAR(MSE->evaluate(predicted_labels, predicted_labels), 0.0, 1e-11);
```
vigsterkr: could you please exchange the first 2 argument, i.e.:
```
EXPECT_NEAR(0.0, MSE->evaluate(predicted_labels, predicted_labels), 1e-11);
```
because the first value for EXPECT_NEAR is the expected value the 2nd argument is the actual value... and for the error use `std::numeric_limits<float64_t>::epsilon()`

so basically:
```
EXPECT_NEAR(0.0, MSE->evaluate(predicted_labels, predicted_labels), std::numeric_limits<float64_t>::epsilon());
```

tests/unit/evaluation/MeanSquared_unittest.cc
```diff
@@ -0,0 +1,66 @@
...
+#include "sg_gtest_utilities.h"
+#include <cmath>
+#include <memory>
+#include <shogun/evaluation/MeanAbsoluteError.h>
+#include <shogun/evaluation/MeanSquaredError.h>
+#include <shogun/evaluation/MeanSquaredLogError.h>
+#include <shogun/labels/BinaryLabels.h>
+#include <shogun/labels/RegressionLabels.h>
+
+class MeanSquaredTest : public ::testing::Test
```
vigsterkr, line 16: could you update this to use the type-templated tests? see an example here:
https://github.com/shogun-toolbox/shogun/blob/develop/tests/unit/base/SGObjectAll_unittest.cc#L29

probably some of these errors will fail for some types, but at least the tests for different floating point types should be tested and it should work

vigsterkr (changes requested):
thanks a lot for the patch! plz check some of my comments

tests/unit/evaluation/MeanSquared_unittest.cc
LiuYuHui: ok, I will change it

tests/unit/evaluation/MeanSquared_unittest.cc
LiuYuHui: When I try to use the type-templated tests, I find the parameter of `MSE->evaluate`is `std:shared_ptr<Labels>`, so, could you tell me which parameter should be templated?

vigsterkr:
great i'll wait till CI is green then i'll merge it!

PR #4835, opened 2020-02-21
registry singleton

gf712:


lisitsyn:
Why not? :)

gf712:
@lisitsyn pushed this in another commit

PR #4837, opened 2020-02-24
use if constexpr in DenseDistance

LiuYuHui:
use `if constexpr` to replace `explicit  template specialization`


gf712:
is get_feature_type() used anywhere in the codebase? I think it might be a good idea to remove it completely no? @vigsterkr

gf712:
@LiuYuHui disregard my previous comment, for now we are keeping this. I'll merge it

PR #4839, opened 2020-02-25
Graph MatMul implementation

gf712:
- [x] implement interface
- [x] implement ngraph call 
- [x] implement shogun version


cmake/external/spdlog.cmake
```diff
@@ -12,11 +12,12 @@ ExternalProject_Add(
 	URL_MD5 a966eea01f81551527853d282896cb4d
 	CMAKE_ARGS
 	-DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}
+	-DCMAKE_POSITION_INDEPENDENT_CODE=ON
```
gf712, line 15: @vigsterkr had to add this to get the fmt static library working

vigsterkr: yep you are right... sorry that slipped my attention.

gf712:
the current implementation is messy, but it works. I added the reduction axis calculation, so in theory could use Eigen::TensorMap in the future to add higher dimension tensors

Issue #4817, opened 2020-02-10
include and import snippets in examples 


mhangaard:
In https://www.shogun-toolbox.org/examples/latest/index.html all tutorials begin with setting up some features and labels types, but all includes and imports are skipped. It would be convenient for the reader if include and import snippets were added at the beginning of each tutorial.

karlnapf:
Yes, this is "planned" for some time, but we never got around to doing it. Thought we had an issue for it already but couldnt find it....
Anyways, the ideal would be to have a button "show full lusting" that takes the reader to exactly that. It should not be too tricky to modify our sphinx plugin to generate that. Nice entrance task.

thealphacod3r:
@karlnapf would love to work on this!
can you guide me a little bit through this??

krutikabapat:
@karlnapf Can you please elaborate what exactly is expected by show full lusting so that I can work on it.

vigsterkr:
@krutikabapat the full source of the generated metaexample of a given language

varsha753:
[shogun_doc.pdf](https://github.com/shogun-toolbox/shogun/files/4399081/shogun_doc.pdf)

varsha753:
Here are some example for implementation of Shogun algorithms in python with imports  
[shogun_doc.pdf](https://github.com/shogun-toolbox/shogun/files/4399127/shogun_doc.pdf)

karlnapf:
Closing as duplicate of #4020

karlnapf:
@varsha753 please don't spam our issues in this way

PR #4846, opened 2020-02-26
Helps in fixing #4817

naruarjun:
Helps in Fixing #4817

This adds the functionality to print entire code from a file generated by a metaexample, I have tested it and rendered cookbooks and it works well.

The command to be added to .rst files is 
```.. sgexample:: filename.sg:entire```

I will create a python script to add this to each file if this is merged.

### DIFF
--- a/doc/cookbook/extensions/sgexample.py
+++ b/doc/cookbook/extensions/sgexample.py
@@ -122,6 +122,8 @@ class ShogunExample(LiteralInclude):
             self.options['end-before'] = section
         elif section == 'end':
             self.options['start-after'] = section
+        elif section == 'entire':
+            pass
         else:
             self.options['start-after'] = section
             self.options['end-before'] = section


vigsterkr:
why do we need to change the rst file? you could just create in sphinx a thing to read in the whole file remove the comments and show the listing... i'm not so sure why it's required to extend the rst

naruarjun:
So I thought this would help us to keep it uniform, I will create a python script that creates a new rst file for each existing rst file and add this line to the new rst file, Then create a hyperlink button, so the user could view the final code in all languages.

naruarjun:
The existing files are not modified.

naruarjun:
Since a hyperlink is to be created, I thought easiest and hassle free way to display the code in a new page was this.

vigsterkr:
i dont see really the point... this could be a simple jquery script that basically loads in the listing and just shows it.... and no please no new page, that's super clunky UX wise

naruarjun:
Ok, I'll look into that approach.
I thought a new page because in the issue it said to add a show full listing button and when clicked, show entire code.

vigsterkr:
yeah but you can show the listing with a simple javascript snippet on the same page in a nice way

naruarjun:
I just want to clarify some specifics. Should the entire code be visible always on the page or should the code section toggle? Where on the page should I display the code and any other specifications you would like to mention?

vigsterkr:
@naruarjun there should be a button/hyperlink for getting the full listing. by default the full listing should not be shown....

gf712:
@naruarjun are you still working on this?

karlnapf:
looks stale to me, and there is not really any added code.
Can be re-opened if needed

PR #4854, opened 2020-02-27
fix CrossValidationStorage initialisation

### LINKED ISSUE #4848: MacOS MKL undefined behaviour
Right now there is undefined behaviour when running the tests. It's probably a coincidence that this only happens in MacOS test. What is probably happening is that a class member somewhere in `CrossValidationFoldStorage`, or it's base classes, is not being initialised and the comparison between two instances fails. To find the bug either look manually for what's wrong or use a sanitizer, or even a good IDE.

This error can be seen in Azure pipelines:
`/Users/runner/runners/2.165.0/work/1/s/tests/unit/base/SGObjectAll_unittest.cc:86: CrossValidationFoldStorage`

gf712:
This should fix #4848

### DIFF
--- a/src/shogun/evaluation/CrossValidationStorage.cpp
+++ b/src/shogun/evaluation/CrossValidationStorage.cpp
@@ -43,12 +43,6 @@ using namespace shogun;
 
 CrossValidationFoldStorage::CrossValidationFoldStorage() : EvaluationResult()
 {
-	m_current_run_index = 0;
-	m_current_fold_index = 0;
-	m_trained_machine = NULL;
-	m_test_result = NULL;
-	m_test_true_result = NULL;
-
 	SG_ADD(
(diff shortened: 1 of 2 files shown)


src/shogun/evaluation/CrossValidationStorage.cpp
```diff
@@ -75,9 +69,7 @@ CrossValidationFoldStorage::CrossValidationFoldStorage() : EvaluationResult()
 	    ParameterProperties::HYPER);
 }
 
-CrossValidationFoldStorage::~CrossValidationFoldStorage()
-{
-}
+CrossValidationFoldStorage::~CrossValidationFoldStorage() = default;
```
gf712, line 72: clang-tidy suggests using default here

PR #4859, opened 2020-02-27
Ported clasifer_multiclassmachine.py example and also eddited INTERFACE file to make proper hyperlinking for install.md

skymatte:


examples/meta/src/multiclass/multiclassmachine.sg
```diff
@@ -0,0 +1,33 @@
...
+#![create_machine]
+Machine classifier = machine("LibSVM",epsilon=0.00005)
+#![create_machine]
+
+#![create_strategy]
+MulticlassStrategy multiclass_one_vs_rest_strategy = multiclass_strategy("KernelMulticlassMachine")
+#![create_strategy]
+
+#![create_classifier]
+MulticlassMachine mc_classifier = multiclass_machine("LibSVMOneClass", kernel=k,classifier,labels_train)
```
gf712: this is not correct, you cannot pass arguments after named arguments. In any case, you should pass named arguments, because in the background this gets called to different .put("argname", arg) calls.

gf712 (changes requested):
Good start!

examples/meta/src/multiclass/multiclassmachine.sg
skymatte: In this comment do you mean that the "classifier" variable I am making, is wrong? Should I make a variable of another name?

examples/meta/src/multiclass/multiclassmachine.sg
```diff
@@ -0,0 +1,33 @@
...
+#![create_machine]
+Machine c = machine("LibSVM",epsilon=0.00005)
+#![create_machine]
+
+#![create_strategy]
+MulticlassStrategy multiclass_one_vs_rest_strategy = multiclass_strategy("KernelMulticlassMachine")
+#![create_strategy]
+
+#![create_classifier]
+MulticlassMachine mc_classifier = multiclass_machine("LibSVMOneClass", kernel=k,classifier=c,labels_train)
```
gf712, line 24: I mean you need to look at what the interface examples look like and you'll see! Run `make meta_examples` and have a look at the output files, in your case the python file will be in something like `build/examples/meta/python/multiclass/multiclassmachine.py`

gf712: look at the CI build error too: https://dev.azure.com/shogunml/shogun/_build/results?buildId=2676&view=logs&j=be3b03d7-4089-5ab3-9462-9af4d8b1dcdf&t=443ade06-68af-54ec-5bea-0e179d048f6f&l=93

gf712: and try to run these things locally too so you can debug more quickly!

PR #4866, opened 2020-03-01
Replace meta examples relative path with cmake variable

### LINKED ISSUE #4865: Refactor meta examples relative paths to absolute paths
Currently the meta examples are failing on windows build do the the fact that the relative path is not correct in all the meta examples, see for example: 
https://github.com/shogun-toolbox/shogun/blob/develop/examples/meta/src/binary/averaged_perceptron.sg#L1

on windows this should be:
```
File f_feats_train = csv_file("../../../data/classifier_binary_2d_linear_features_train.dat")
```

there's two way to solve this:
 1) change the meta example generator that basically fixes this relative path based on a platform
 2) change the meta examples generator in a way that it does a join over the path given here and an environment variable, say `SHOGUN_DATA` which contains the absolute path to the right directory. this of course adds more lines of code to the generated examples (in each languages there has to be some env var extractor code)..

gf712:
fixes #4865


vigsterkr:
@gf712 is there any chance you can just create the cookbook and paste here one of the examples ? :)

gf712:
@vigsterkr for the cookbook we don't show the load path. It'll be something like 
```
features_train = features(f_feats_train)
features_test = features(f_feats_test)
labels_train = labels(f_labels_train)
labels_test = labels(f_labels_test)
```
```

File: cmake/FindMetaExamples.cmake [hunk 1]
```diff
@@ -22,6 +22,8 @@ function(get_excluded_meta_examples)
     is_set(${SHOGUN_CONFIG} HAVE_LAPACK)
     is_set(${SHOGUN_CONFIG} USE_SVMLIGHT)
 
+  # TODO(geektoni): fix this meta example
+  LIST(APPEND evaluation/cross_validation_multiple_kernel_learning_weights_storage.sg)
```
gf712: @geektoni I brought this file back to develop because otherwise the cookbook build fails. However the actual example doesn't seem to work (I pretty much copied what you did in #4655), so I black flagged it. It fails because the cross validation objects doesn't seems to have any observable values? I get ` Observation index (0) is out of bound (total observations 0)` in the .get_observation(0) line in the meta example.

geektoni: Mmmh I'll try to have a look. Maybe it is missing some observer...

gf712: cool, thanks! this should be merged today into develop, so can fix it after that :)

PR #4869, opened 2020-03-02
add feature importances for CARTree

### LINKED ISSUE #4851: Feat. Importance and Feat. Count
Hello,

I'm working with tree based algorithm, such Random Forest, CART or C4.5. But i need to analyze which features and how often the algorithm uses to separate the data. 

Could you please implement those methods or functions that calculate the feature importance and feature count on those algorithms?

Thanks!

LiuYuHui:
fix #4851


vigsterkr (changes requested):
great stuff!! 
make sure that add a `watch_method` to `get_feature_importance`.

check the codebase for other examples. this way we'll be able to call this method nicely with .get() from the SWIG interfaces. and as usual a unit test for this would be great!

src/shogun/multiclass/tree/CARTree.cpp
```diff
@@ -491,10 +518,23 @@ std::shared_ptr<BinaryTreeMachineNode<CARTreeNodeData>> CARTree::CARTtrain(std::
 	right_child->data.transit_into_values=right_transit;
 	node->data.num_leaves=left_child->data.num_leaves+right_child->data.num_leaves;
 	node->data.weight_minus_branch=left_child->data.weight_minus_branch+right_child->data.weight_minus_branch;
-
+	node->data.impurity = node_impurity;
 	return node;
 }
-
+SGVector<float64_t> CARTree::get_feature_importance()
```
vigsterkr: mmm here this function *always* recomputes the feature importance as far as i understand. since `compute_feature_importance` is called here... could we call this on the end of the train? i mean the feature importance doesn't change after a train.

so basically what i would do is simply have a void function that is called after the train is done, it updates the internal `m_feature_importances`, and the CARTree::get_feature_importance simply returns that SGVector...

vigsterkr: and then i wouldn't add a watch_method but simply do an SG_ADD for the m_feature_importances (see other examples in the codebase)

src/shogun/multiclass/tree/CARTree.cpp
```diff
@@ -491,10 +518,23 @@ std::shared_ptr<BinaryTreeMachineNode<CARTreeNodeData>> CARTree::CARTtrain(std::
...
+	m_feature_importances = SGVector<float64_t>(m_num_feat);
+	compute_feature_importance(m_root->as<bnode_t>());
+	float64_t total_num_sample = m_root->data.total_weight;
+	linalg::scale(
+	    m_feature_importances, m_feature_importances, 1 / total_num_sample);
+	auto normalizer = linalg::sum(m_feature_importances);
+	if (normalizer > 0)
+		linalg::scale(
+		    m_feature_importances, m_feature_importances, 1 / normalizer);
+	return m_feature_importances;
```
gf712: did you compare the feature importance calculation with another implementation?

src/shogun/multiclass/tree/CARTree.h
```diff
@@ -469,6 +478,12 @@ class CARTree : public RandomMixin<TreeMachine<CARTreeNodeData>>
 
 	/** minimum number of feature vectors required in a node **/
 	int32_t m_min_node_size;
+
+	/**shores feature importances**/
```
gf712: stores?

src/shogun/multiclass/tree/CARTree.cpp
gf712: is this the same as in scikit learn?

LiuYuHui: yes, this implementation is the same as sklearn , actually, I borrow it from [this](https://github.com/scikit-learn/scikit-learn/blob/0abd95f742efea826df82458458fcbc0f9dafcb2/sklearn/tree/_tree.pyx#L1056)

src/shogun/multiclass/tree/CARTree.h
LiuYuHui: hhh, typo

src/shogun/multiclass/tree/CARTree.cpp
LiuYuHui: yes, the way of calling `compute_feature_importance` after the train is better, i will change it

src/shogun/multiclass/tree/CARTree.cpp
```diff
@@ -281,10 +281,30 @@ bool CARTree::train_machine(std::shared_ptr<Features> data)
 	{
 		prune_by_cross_validation(dense_features,m_folds);
 	}
-
 	return true;
 }
 
+void CARTree::compute_feature_importance(std::shared_ptr<bnode_t> node)
```
gf712: you should pass the argument as `const std::shared_ptr<bnode_t>& node`

src/shogun/multiclass/tree/CARTree.cpp
```diff
@@ -281,10 +281,30 @@ bool CARTree::train_machine(std::shared_ptr<Features> data)
...
+{
+	auto left = node->left();
+	auto right = node->right();
```
gf712: also could be `const auto& right = .. `?

src/shogun/multiclass/tree/CARTree.cpp
gf712: cool, thanks! seems to be correct :) maybe you could also add a unittest? not sure if it comparable to a result in scikit learn?

src/shogun/multiclass/tree/CARTree.cpp
LiuYuHui: i thought the size of shared_ptr is the same as the raw pointer, but const std::shared_ptr<bnode_t>& node is surely better :)

src/shogun/multiclass/tree/CARTree.cpp
LiuYuHui: yes, i will change it.

src/shogun/multiclass/tree/CARTree.cpp
gf712: also you should add a comment in the documentation saying that this is the implementation from scikit learn, as different tree implementations might slightly different definitions of feature importance

LiuYuHui: ok

src/shogun/multiclass/tree/CARTree.cpp
gf712: hmm it's not about the size. A shared_ptr has a reference counter, and if you copy it it will do an atomic operation to increment the count, which is very slow. So if you pass it as const reference you are just passing a reference, so no copy occurs, and the const makes sure you don't modify the reference (more or less).
A raw pointer is just a memory address (more or less).

src/shogun/multiclass/tree/CARTree.cpp
gf712: similarly here you avoid a copy, because this is just temporarily bound, e.g. the left and right variables are destroyed when you go out of scope. in theory you don't need to bind as a reference, as the compiler is required to do a copy elision here! So `const auto right = ` should suffice.
If you want to read more about references [this post is good](https://herbsutter.com/2020/02/23/references-simply/)

LiuYuHui: ok! thanks for telling me this.

src/shogun/multiclass/tree/CARTree.cpp
```diff
@@ -281,10 +281,30 @@ bool CARTree::train_machine(std::shared_ptr<Features> data)
...
+	if (left && right)
+	{
+		m_feature_importances[node->data.attribute_id] +=
+		    (node->data.impurity * node->data.total_weight -
+		     left->data.impurity * left->data.total_weight -
+		     right->data.impurity * right->data.total_weight);
+	}
+	if (left->data.num_leaves > 1)
```
vigsterkr: just a thought no need to implement this here right now, but instead of this recursive call we could fill up an array and then calculate the importance in a loop... that later could be openmp-ed? :)

tests/unit/multiclass/tree/CARTree_unittest.cc
```diff
@@ -198,6 +198,33 @@ TEST(CARTree, classify_nominal)
...
+	auto labels = std::make_shared<MulticlassLabels>(y);
+
+	auto c = std::make_shared<CARTree>();
+	c->set_labels(labels);
+	c->train(feats);
+	auto feat_import = c->get_feature_importance();
+	// those data are generated by below sklearn program
+	EXPECT_NEAR(0.111111, feat_import[0], 0.00001);
+	EXPECT_NEAR(0.222222, feat_import[1], 0.00001);
+	EXPECT_NEAR(0.666667, feat_import[2], 0.00001);
```
gf712, line 216: nice! thank you :)

src/shogun/multiclass/tree/CARTree.cpp
```diff
@@ -1452,6 +1524,9 @@ void CARTree::init()
 	m_sorted_features=SGMatrix<float64_t>();
 	m_sorted_indices=SGMatrix<index_t>();
 
+	SG_ADD(
+	    &m_feature_importances, "feature_importances", "feature importances");
```
gf712, line 1528: Could you add .get("feature_importances") in cartree.sg? So we can see how it works in interfaces

gf712 (commented):
I think the PR is almost there! thanks a lot :)

src/shogun/multiclass/tree/CARTree.cpp
LiuYuHui: done :)

vigsterkr:
awesome! thnx!

PR #4874, opened 2020-03-03
because circular

vigsterkr:
testing on the fly 🤷‍♂


gf712:
I think you need Kernel to be forward declared for KernelNormalizer. Or include the KernelNormalizer in all the kernel implementations that require it...

vigsterkr:
@gf712 currently normalizers has to include Kernel.h as they are all header only and do access kernel params... so going with the former solution....

PR #4876, opened 2020-03-04
Fixing Memory allocator/diallocator consistency

Khalifa1997:
In response to Issue of fixing memory allocations by SG_MALLOC


src/shogun/classifier/svm/LibLinear.cpp
```diff
@@ -117,7 +117,7 @@ bool LibLinear::train_machine(std::shared_ptr<Features> data)
 	}
 	SGVector<float64_t> w;
 	if (get_bias_enabled())
-		w = SGVector<float64_t>(SG_MALLOC(float64_t, num_feat + 1), num_feat);
+		w = SGVector<float64_t>(num_feat + 1, num_feat);
```
vigsterkr: mmm this one is actually a bit tricky.... since the idea is that the last element of W is the bias... but when u actually do most of the operations with W you actually want to have it's length to be the feat space dim... i.e. w.vlen should be num_feat but the size of w should be num_feat + 1 and the last element of w is used as the bias... from mem management perspective, apart from what's the size of the chunk u allocate the vlen doesn't matter... in other words:
```
w = SGVector<float64_t>(num_feat + 1);
w.vlen = num_feat;
```
could work... or
```
w = SGVector<float64_t>(SG_ALIGNED_ALLOC(num_feat + 1), num_feat);
```

the former is maybe easier to understand (although imo that vlen should be a const vlen and then this wouldn't be possible) and really follows the encapsulation concept as you dont expose the internal workings of the SGVector container. the latter is just a simple patch, but of course this breaks encapsulation, hence if tomorrow somebody refactors the whole SGVector internals this definitely need to be handled as a corner case (i.e. directly calling SG_ALIGNED_MALLOC).

vigsterkr (changes requested):
have you tried to compile this file? as imo it actually should fail as SGVector does not have an (int, int) ctor; see https://www.shogun.ml/api/latest/classshogun_1_1SGVector.html

src/shogun/distributions/Distribution.cpp
```diff
@@ -36,12 +36,11 @@ SGVector<float64_t> Distribution::get_log_likelihood()
 	ASSERT(features)
 
 	int32_t num=features->get_num_vectors();
-	float64_t* vec=SG_MALLOC(float64_t, num);
-
+	SGVector<float64_t> vec=SGVector<float64_t>(num);
```
vigsterkr:
```suggestion
	SGVector<float64_t> vec(num);
```

src/shogun/distributions/Distribution.cpp
```diff
@@ -36,12 +36,11 @@ SGVector<float64_t> Distribution::get_log_likelihood()
...
 	for (int32_t i=0; i<num; i++)
-		vec[i]=get_log_likelihood_example(i);
+		vec[i].vector=get_log_likelihood_example(i);
```
vigsterkr:
```suggestion
		vec[i]=get_log_likelihood_example(i);
```

vigsterkr (changes requested):
plz check the api of SGVector... there's indexing operator specified so u should be able to do vec[i]

src/shogun/distributions/Distribution.cpp
vigsterkr: no need for the extra copy ctor call

Khalifa1997:
Okay I will do check the API, and thanks!

src/shogun/classifier/svm/LibLinear.cpp
Khalifa1997: Okay so which one should I go for in this case

vigsterkr: @gf712 any opinion about this?

gf712: I think you should either use second one, or rewrite the algo so that you don’t have to use this hack... I agree that you shouldn’t change vlen, at least via the public interface

gf712: actually `w = SGVector<float64_t>(num_feat + 1);` should work fine, you just need to change `set_w` to use a slice of the vector, i.e. `m_w=w[0:num_feat+1]` and `b = w[num_feat+1]`. @vigsterkr from what i can see the number of features in the algo comes from the `prob` object and not SGVector.

src/shogun/classifier/svm/LibLinear.cpp
```diff
@@ -116,8 +117,9 @@ bool LibLinear::train_machine(std::shared_ptr<Features> data)
 		}
 	}
 	SGVector<float64_t> w;
+	index_t len =num_feat + 1;
```
vigsterkr: is this necessary? i mean it's only used the one place or?

src/shogun/classifier/svm/LibLinear.cpp
```diff
@@ -222,7 +224,7 @@ bool LibLinear::train_machine(std::shared_ptr<Features> data)
 		error("Error: unknown solver_type");
 		break;
 	}
-
+
```
vigsterkr, line 226: please try to watch for these unnecessary whitespace changes...

vigsterkr (changes requested):
some minor details

src/shogun/classifier/svm/LibLinear.cpp
Khalifa1997: Oh my bad okay, will do from now on.

src/shogun/classifier/svm/LibLinear.cpp
Khalifa1997: It's used in this one place only, cause this is what SG_ALIGNED_MALLOC takes, based on what I have read

Khalifa1997: I could change it however, np

src/shogun/distributions/PositionalPWM.cpp
```diff
@@ -217,7 +217,7 @@ SGMatrix<float64_t> PositionalPWM::get_scoring(int32_t d)
 		offs+=Math::pow((int32_t) m_w.num_rows,i+1);
 	int32_t rows=Math::pow((int32_t) m_w.num_rows,d);
 	int32_t cols=m_w.num_cols;
-	float64_t* scoring_matrix = SG_MALLOC(float64_t, rows*cols);
+	SGVector<float64_t>scoring_matrix(rows*cols);
```
vigsterkr: noup... this is almost exactly the same as the example i provided in the issue

src/shogun/distributions/PositionalPWM.cpp
```diff
@@ -217,7 +217,7 @@ SGMatrix<float64_t> PositionalPWM::get_scoring(int32_t d)
...
 	sg_memcpy(scoring_matrix,m_poim.vector+offs,rows*cols*sizeof(float64_t));
```
vigsterkr:
```suggestion
	sg_memcpy(scoring_matrix.matrix,m_poim.vector+offs,rows*cols*sizeof(float64_t));
```

src/shogun/distributions/PositionalPWM.cpp
vigsterkr:
```suggestion
	SGMatrix<float64_t> scoring_matrix(rows,cols);
```

src/shogun/distributions/PositionalPWM.cpp
```diff
@@ -217,7 +217,7 @@ SGMatrix<float64_t> PositionalPWM::get_scoring(int32_t d)
...
-	return SGMatrix<float64_t>(scoring_matrix,rows,cols);
+	return SGMatrix<float64_t>(scoring_matrix,index_t(rows),index_t(cols));
```
vigsterkr:
```suggestion
	return scoring_matrix;
```

vigsterkr (changes requested):
mmmm there are some conceptual problems in this patch.

src/shogun/classifier/svm/LibLinear.cpp
vigsterkr: the whitespace is still there :<

vigsterkr:
the malloc creates a memchunk of the size of the later used SGMatrix..... so there's no need for the SGVector wrappingm and i believe that this last patch wasn't even compiling locally... plz try to check that before send in a commit unless you are super blocked

Khalifa1997:
> the malloc creates a memchunk of the size of the later used SGMatrix..... so there's no need for the SGVector wrappingm and i believe that this last patch wasn't even compiling locally... plz try to check that before send in a commit unless you are super blocked

It's kinda weird it compiled for me when I used the make command, but yea I will pay more emphasis for that from now on I must of gotten it wrong but thanks for exampling however

vigsterkr:
@Khalifa1997 huh some crazy stuff happened with your branch. i guess you were rebasing something? in general would be great if you'd google around how to rebase a branch over another branch and some other 101 git operations. there are really tons of good material about this out there.

until you get of hold of that maybe just open a PR per commit/fix

src/shogun/features/SNPFeatures.cpp
```diff
@@ -401,8 +401,9 @@ SGMatrix<float64_t> SNPFeatures::get_histogram(bool normalize)
 		}
 	}
 	SG_FREE(h_normalizer);
-
-	return SGMatrix<float64_t>(h, nsym, string_length/2);
+	SGMatrix<float64_t>hMatrix(h, nsym, string_length/2,false);
+	SG_FREE(h);
```
vigsterkr: why this free?
maybe it would be good to think about what actually would be happening with h. you can play with a debugger and see what happens line by line

src/shogun/features/SNPFeatures.cpp
```diff
@@ -411,16 +412,15 @@ SGMatrix<float64_t> SNPFeatures::get_2x3_table(const std::shared_ptr<SNPFeatures
 	ASSERT(pos->strings->get_max_vector_length() == neg->strings->get_max_vector_length())
 	int32_t len=pos->strings->get_max_vector_length();
 
-	float64_t* table=SG_MALLOC(float64_t, 3*2*len/2);
 
 	SGMatrix<float64_t> p_hist=pos->get_histogram(false);
 	SGMatrix<float64_t> n_hist=neg->get_histogram(false);
-
+	SGMatrix<float64_t>table(2,3*len/2);
```
vigsterkr: plz use spaces

src/shogun/features/SNPFeatures.cpp
```diff
@@ -411,16 +412,15 @@ SGMatrix<float64_t> SNPFeatures::get_2x3_table(const std::shared_ptr<SNPFeatures
...
 
 	for (int32_t i=0; i<3*len/2; i++)
 	{
-		table[2*i]=p_hist.matrix[i];
-		table[2*i+1]=n_hist.matrix[i];
+		table.set_element(p_hist.matrix[i],0,2*i);
```
vigsterkr:
```
table.matrix[2*i]=p_hist.matrix[i];
table.matrix[2*i+1]=n_hist.matrix[i];
```

src/shogun/features/SNPFeatures.cpp
Khalifa1997: > @Khalifa1997 huh some crazy stuff happened with your branch. i guess you were rebasing something? in general would be great if you'd google around how to rebase a branch over another branch and some other 101 git operations. there are really tons of good material about this out there.
> 
> until you get of hold of that maybe just open a PR per commit/fix
Yup had plenty of problems with it, but yea will prob check some tutorials out




> why this free?
> maybe it would be good to think about what actually would be happening with h. you can play with a debugger and see what happens line by line


Okay will do, thanks!

Khalifa1997:
Nvm the commits, I had some issues with rebase but I have more grasp of it now :)

vigsterkr:
@Khalifa1997 it'd be good if you could sort out this PR (i mean the commits) as i'd really like to have it merged into develop, but unfortunately now in it's state its not possible.

src/shogun/features/SNPFeatures.cpp
```diff
@@ -401,8 +401,8 @@ SGMatrix<float64_t> SNPFeatures::get_histogram(bool normalize)
 		}
 	}
 	SG_FREE(h_normalizer);
-
-	return SGMatrix<float64_t>(h, nsym, string_length/2);
+	SGMatrix<float64_t>hMatrix(h, nsym, string_length/2,false);
```
vigsterkr: spaces plz

src/shogun/features/SNPFeatures.cpp
```diff
@@ -411,16 +411,15 @@ SGMatrix<float64_t> SNPFeatures::get_2x3_table(const std::shared_ptr<SNPFeatures
 	ASSERT(pos->strings->get_max_vector_length() == neg->strings->get_max_vector_length())
 	int32_t len=pos->strings->get_max_vector_length();
 
-	float64_t* table=SG_MALLOC(float64_t, 3*2*len/2);
 
 	SGMatrix<float64_t> p_hist=pos->get_histogram(false);
 	SGMatrix<float64_t> n_hist=neg->get_histogram(false);
-
+	SGMatrix<float64_t>table(2, 3*len/2);
```
vigsterkr: use spaces plz

src/shogun/features/StringFeatures.cpp
```diff
@@ -160,10 +160,9 @@ template<class ST> SGVector<ST> StringFeatures<ST>::get_feature_vector(int32_t n
 	int32_t l;
 	bool free_vec;
 	ST* vec=get_feature_vector(num, l, free_vec);
-	ST* dst=SG_MALLOC(ST, l);
-	sg_memcpy(dst, vec, l*sizeof(ST));
+	SGVector<ST>dst(vec,l);
```
vigsterkr: as said, use whitespace

src/shogun/features/SNPFeatures.cpp
vigsterkr: note that this is actually just putting one line up the matrix creation.... but the underlying memory chunk `h` is still created by SG_MALLOC/CALLOC see https://github.com/shogun-toolbox/shogun/blob/9bd877047b31461a05ef97f489a17b32c3d61427/src/shogun/features/SNPFeatures.cpp#L354

you should just create there:
```
SGMatrix<float64_t> h(nsym, string_length/2,false);
```
instead of the calloc, and return `h` itself

src/shogun/features/StringFeatures.cpp
```diff
@@ -160,10 +160,9 @@ template<class ST> SGVector<ST> StringFeatures<ST>::get_feature_vector(int32_t n
 	int32_t l;
 	bool free_vec;
 	ST* vec=get_feature_vector(num, l, free_vec);
-	ST* dst=SG_MALLOC(ST, l);
-	sg_memcpy(dst, vec, l*sizeof(ST));
```
vigsterkr, line 164: you should not remove this as without this the content of `dst` vector is nothing (or all 0s to be precise)

Khalifa1997: So do I leave it with the Malloc there

gf712: you can also do `sg_memcpy(dst.vector, vec, l*sizeof(ST))`;

gf712: @Khalifa1997 you should use the SGVector constructor with only the size parameter: `SGVector<ST> dst(l);` and then copy the data over: `sg_memcpy(dst.vector, vec, l*sizeof(ST));`
Right now the data in `vec` that you pass to the sgvector constructor gets deleted below (`free_feature_vector(vec, num, free_vec);`), so you need that copy. And then the sgvector will also try deallocate the memory that was deleted by `free_feature_vector`...

Khalifa1997:
> @Khalifa1997 it'd be good if you could sort out this PR (i mean the commits) as i'd really like to have it merged into develop, but unfortunately now in it's state its not possible.

Yea I agree, but how do you suggest going about this whenever I do
` git pull --rebase upstream develop`
I get those merge changes that I do, then I add the changes, then commit and push
now what happens is I get those ton of commits in the push, and I have no idea how this happens

src/shogun/features/StringFeatures.cpp
Khalifa1997: Got it! thanks

PR #4878, opened 2020-03-04
Some operator overloads for readability

gf712:


src/shogun/mathematics/graph/nodes/Add_test.cc
```diff
@@ -90,9 +90,9 @@ TYPED_TEST(GraphTest, matrix_add)
 	auto input1 = make_shared<node::Input>(
 	    Shape{10, Shape::Dynamic}, get_enum_from_type<NumericType>::type);
 
-	auto intermediate = make_shared<node::Add>(input, input);
+	auto intermediate = input + input;
```
gf712, line 93: @vigsterkr saving so many characters

gf712:
@vigsterkr I am not sure if `operator==` is a good idea, same with the other logical operators...?

vigsterkr:
i would keep them now not overloaded :)

Issue #4847, opened 2020-02-26
Change KMeans "mus" parameter to something more user friendly


gf712:
Right now "cluster centres" in KMeansBase is a parameter called "mus". This is probably not a good name, at least for user facing code. This should be changed in KMeansBase.cpp. Should be straightforward, but will probably require some test/example code refactoring.

zwasd:
Is `initial_centers` a good name?

gf712:
Well it's not initial_centers though, it's the cluster centres right? If you look at KMeansBase::get_cluster_centers() that returns mus. I think you'll need to register `mus_initial` and name that `initial_centers` and then remove `mus`. `mus` is already registered using `watch_method("cluster_centers", &KMeansBase::get_cluster_centers);`

gf712:
and then you'll need to add a callback function that checks the shape of the provided cluster centres (like in set_initial_centers), but that we can have a look at once a PR is started.

ana-simionescu:
How can I test the functionality after making my changes?

gf712:
Hi, you could either write a unittest or a meta example! There is information about both [here](https://github.com/shogun-toolbox/shogun/blob/develop/doc/readme/DEVELOPING.md#testing) :)

Khalifa1997:
Can I still work on this :D

krutikabapat:
@gf712 Basically we want the name of the parameter mus to be cluster_centers and mus_initial as initial_clusters. Right?

mukhituly:
I want to work on it. Can you guide me?

vigsterkr:
there are already several PRs open about this... although none of them actually finished.

PR #4879, opened 2020-03-04
#4847 Changed KMeans "mus" parameter

krutikabapat:
Changed Kmeans parameter to cluster_centers.

### DIFF
--- a/src/shogun/clustering/KMeans.cpp
+++ b/src/shogun/clustering/KMeans.cpp
@@ -171,7 +171,7 @@ void KMeans::Lloyd_KMeans(SGMatrix<float64_t> centers, int32_t num_centers)
 			}
 		}
 
-		observe<SGMatrix<float64_t>>(iter, "mus");
+		observe<SGMatrix<float64_t>>(iter, "cluster_centers");
 
 		if (iter%(max_iter/10) == 0)
 			io::info("Iteration[{}/{}]: Assignment of {} patterns changed.", iter, max_iter, changed);
@@ -185,10 +185,11 @@ void KMeans::Lloyd_KMeans(SGMatrix<float64_t> centers, int32_t num_centers)
 bool KMeans::train_machine(std::shared_ptr<Features> data)
 {
 	initialize_training(data);
-	Lloyd_KMeans(mus, k);
+	//Lloyd_KMeans(mus, k);
+	Lloyd_KMeans(cluster_centers,k);
 	compute_cluster_variances();
-	auto cluster_centres = std::make_shared<DenseFeatures<float64_t>>(mus);
-	distance->replace_lhs(cluster_centres);
+	auto cluster_cent = std::make_shared<DenseFeatures<float64_t>>(cluster_centers);
+	distance->replace_lhs(cluster_cent);
 	return true;
 }
 
--- a/src/shogun/clustering/KMeans.cpp
+++ b/src/shogun/clustering/KMeans.cpp
@@ -171,7 +171,7 @@ void KMeans::Lloyd_KMeans(SGMatrix<float64_t> centers, int32_t num_centers)
 			}
 		}
 
-		observe<SGMatrix<float64_t>>(iter, "cluster_centers");
+		observe<SGMatrix<float64_t>>(iter, "mus");
 
 		if (iter%(max_iter/10) == 0)
 			io::info("Iteration[{}/{}]: Assignment of {} patterns changed.", iter, max_iter, changed);
--- a/src/shogun/clustering/KMeansBase.cpp
+++ b/src/shogun/clustering/KMeansBase.cpp
@@ -82,7 +82,7 @@ void KMeansBase::set_random_centers()
 		lhs->free_feature_vector(vec, cluster_center_i);
 	}
 
-	observe<SGMatrix<float64_t>>(0, "cluster_centers");
+	observe<SGMatrix<float64_t>>(0, "mus");
 }
 
 void KMeansBase::compute_cluster_variances()
@@ -173,8 +173,8 @@ void KMeansBase::initialize_training(const std::shared_ptr<Features>& data)
 
 	if (initial_centers.matrix)
 	{
-		cluster_centers = initial_centers;
-		observe<SGMatrix<float64_t>>(0, "cluster_centers");
+		mus = mus_initial;
+		observe<SGMatrix<float64_t>>(0, "mus");
 	}
 	else
 	{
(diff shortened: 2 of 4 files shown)


LiuYuHui:
Are you test your change? it seems like your change does not pass the unit test.

krutikabapat:
@LiuYuHui I will check it again! Thank you

gf712:
@krutikabapat why did you close and delete?

krutikabapat:
There were some unit tests that were not passed, so I thought I will
rectify it once and send the PR again. Also by doing this, it would be easy
for others to send a PR for that issue if they already have completed it. :D

gf712:
Don't worry about that, just keep this open :) otherwise we keep getting notifications of PRs opening and closing. Also this way we can keep track of the PRs being worked on. Just close a PR if you plan on not working on it at all.

krutikabapat:
Really sorry about that. I am working hard to get the unit tests working. I
just wanted to know if restoring the branch will work? Also thanks a lot
for your suggestions.

gf712:
Yes this is fine. Let use know if you get stuck :) And push commits once you get it working locally.

krutikabapat:
I have reopened it. I will notify once I am done with it. :D.
Thank you!

src/shogun/clustering/KMeansBase.cpp
```diff
@@ -321,7 +321,7 @@ void KMeansBase::init()
 	SG_ADD(
 	    &use_kmeanspp, "kmeanspp", "Whether to use kmeans++",
 	    ParameterProperties::HYPER | ParameterProperties::SETTING);
-	SG_ADD(&mus, "mus", "Cluster centers", ParameterProperties::MODEL);
+	SG_ADD(&cluster_centers, "cluster_centers", "Cluster centers", ParameterProperties::MODEL);
```
vigsterkr:
```suggestion
	SG_ADD(&mus, "cluster_centers", "Cluster centers", ParameterProperties::MODEL);
```

src/shogun/clustering/KMeansMiniBatch.cpp
```diff
@@ -96,8 +96,8 @@ void KMeansMiniBatch::minibatch_KMeans()
 				c_alive[c]=(1.0-eta)*c_alive[c]+eta*x[c];
 			}
 		}
-		mus = rhs_mus->get_feature_matrix();
-		observe<SGMatrix<float64_t>>(i, "mus");
+		cluster_centers = rhs_mus->get_feature_matrix();
+		observe<SGMatrix<float64_t>>(i, "cluster_centers");
```
vigsterkr: keep it

src/shogun/clustering/KMeansBase.cpp
```diff
@@ -163,18 +163,18 @@ void KMeansBase::initialize_training(const std::shared_ptr<Features>& data)
...
-	memset(mus.matrix, 0, sizeof(float64_t)*centers_size);
+	memset(cluster_centers.matrix, 0, sizeof(float64_t)*centers_size);
 
-	if (mus_initial.matrix)
+	if (initial_centers.matrix)
 	{
-		mus = mus_initial;
-		observe<SGMatrix<float64_t>>(0, "mus");
+		cluster_centers = initial_centers;
+		observe<SGMatrix<float64_t>>(0, "cluster_centers");
```
vigsterkr: keep this line

src/shogun/clustering/KMeansBase.cpp
```diff
@@ -77,12 +77,12 @@ void KMeansBase::set_random_centers()
...
 
 		for (int32_t j=0; j<dimensions; j++)
-			mus(j,i)=vec[j];
+			cluster_centers(j,i)=vec[j];
 
 		lhs->free_feature_vector(vec, cluster_center_i);
 	}
 
-	observe<SGMatrix<float64_t>>(0, "mus");
+	observe<SGMatrix<float64_t>>(0, "cluster_centers");
```
vigsterkr: keep this line

src/shogun/clustering/KMeans.cpp
```diff
@@ -171,7 +171,7 @@ void KMeans::Lloyd_KMeans(SGMatrix<float64_t> centers, int32_t num_centers)
 			}
 		}
 
-		observe<SGMatrix<float64_t>>(iter, "mus");
+		observe<SGMatrix<float64_t>>(iter, "cluster_centers");
```
vigsterkr: keep this line

vigsterkr (changes requested):
i've marked which are the changes you should keep and suggested 1 place where you need a slightly different change.... all the rest you could just leave as it was originally

krutikabapat:
@vigsterkr Thank you so much for all the suggested changes. I have made the changes in the respective files. Please let me know if there are any further changes required. :D.

src/shogun/clustering/KMeans.cpp
```diff
@@ -185,10 +185,11 @@ void KMeans::Lloyd_KMeans(SGMatrix<float64_t> centers, int32_t num_centers)
 bool KMeans::train_machine(std::shared_ptr<Features> data)
 {
 	initialize_training(data);
-	Lloyd_KMeans(mus, k);
+	//Lloyd_KMeans(mus, k);
```
vigsterkr, line 188: i guess this got accidentally left in. just drop it.

src/shogun/clustering/KMeansBase.cpp
```diff
@@ -321,7 +321,7 @@ void KMeansBase::init()
 	SG_ADD(
 	    &use_kmeanspp, "kmeanspp", "Whether to use kmeans++",
 	    ParameterProperties::HYPER | ParameterProperties::SETTING);
-	SG_ADD(&mus, "mus", "Cluster centers", ParameterProperties::MODEL);
+	SG_ADD(&mus, "cluster_centers", "Cluster centers", ParameterProperties::MODEL);
```
vigsterkr, line 324: this wont work. if you rename the variable itself (mus -> cluster_centers) then you have to do it here as well.....

hence my previous suggestion that you do not need to rename any of the variable names, just how it's being registered here, and how it is called in the observe methods.

i'm pretty sure that this does not compile locally... so plz before you send in a patch make sure that it compiles... and then you'll see what are the errors you need to fix.

vigsterkr (changes requested):

src/shogun/clustering/KMeansBase.cpp
gf712, line 324: this is not going to work though, as a param is registered with the same name below... I think the idea is that this would become something like "cluster_centers_initial" and the user can put a matrix in that parameter as an inital guess of the cluster centres?

krutikabapat: @vigsterkr I am planning to rename the variables as mus -> cluster_centers and mus_initial -> initial_centers. Because at the end we want the name of the param **mus** to be changed right? And then I will register them and make changes in how they are called in the observe methods. Not sure, is it a good idea?

krutikabapat:
@LiuYuHui Why do you think it will fail for test cases?

LiuYuHui:
@krutikabapat imo,you should also change derived class such as `KMeansMiniBatch` :)

PR #4881, opened 2020-03-05
Move op registration to individual files

gf712:
Long story short:
 - OperatorFactory was moved to RuntimeNode.h so that all ops can be registered outside the executor implementation
 - this caused ODR violations, so OperatorFactory became inline
 - this meant that all operators were put in the same map, so shogun's Add was overriding ngraph's add -> runtime segfault
 - so I used two different OperatorFactory functions: shogun and ngraph
 - this is probably not a good design but get the ball rolling and we can separate all the definitions to .cpp files, which will reduce compile times


src/shogun/mathematics/graph/nodes/Add_test.cc
```diff
@@ -61,24 +60,24 @@ TYPED_TEST(GraphTest, matrix_add)
...
+	// }
+	// else
+	// {
+	// 	UniformIntDistribution<TypeParam> dist;
+	// 	auto gen = [&dist, &mersenne_engine]() {
+	// 		return dist(mersenne_engine);
+	// 	};
+	// 	generate(X1.begin(), X1.end(), gen);
+	// 	generate(X2.begin(), X2.end(), gen);
+	// }
```
gf712: temporary commented out to see if stuff compiles on all platforms

cmake/external/spdlog.cmake
```diff
@@ -17,7 +23,8 @@ ExternalProject_Add(
 	-DSPDLOG_BUILD_BENCH=OFF
 	-DSPDLOG_BUILD_TESTS=OFF
 	INSTALL_COMMAND ${CMAKE_COMMAND} -E copy_directory ${spdlog_SOURCE_DIR}/include ${spdlog_INCLUDE_DIR}
-    BUILD_BYPRODUCTS ${spdlog_DIR}/${CMAKE_STATIC_LIBRARY_PREFIX}spdlog${CMAKE_STATIC_LIBRARY_SUFFIX}
+    BUILD_BYPRODUCTS ${spdlog_LIBRARY_NAME}
+    BUILD_ALWAYS 1
```
gf712, line 27: @vigsterkr probably not best practice but it is the only way to switch from debug to release (and back) without cmake complaining and having to delete the build directory

vigsterkr (approved):

PR #4883, opened 2020-03-05
sign op

gf712:


src/shogun/mathematics/graph/Types.h
```diff
@@ -30,6 +30,14 @@ namespace shogun
...
 		};
 
+		inline bool is_unsigned(const element_type& type)
+		{
+			return type == element_type::BOOLEAN ||
+			       type == element_type::UINT8 ||
+			       type == element_type::UINT16 ||
+			       type == element_type::UINT32 || type == element_type::UINT64;
+		}
+
```
gf712, line 40: @vigsterkr would be nice to have a class for this :) like the arrow one

vigsterkr: ayay captain! i'll jump on this one, and actually try to clear up all these element type stories... only thing remains: if i unify this across shogun then it means that it's gonna end up in shogun-core. which is a problem if we are going to carve out shogun-graph from the main repo

gf712: i guess the shogun-core one would just use a subset of the features though? i mean, at worst copy paste or create the simplest submodule :D

Issue #4871, opened 2020-03-02
cleanup memory allocator/deallocator inconsistencies


vigsterkr:
In many cases in shogun there are still plain SG_MALLOC/CALLOC calls that allocates a memory region then that region is either passed to SGVector or SGMatrix but not only to wrap that region for interface conformity but as well to pass that memory region to be managed by SGMatrix/SGVector... i.e. when you see something like this:
```
float64_t mem = SG_MALLOC(float64_t, size);
....
SGVector<float64_t> v(mem, size);
```
in this case the SGVector will call a deallocator on the mem region, but it's actually not correct as SGVector might call a wrong type of deallocator (incompatible with the SG_MALLOC one).

the other case when you see something like this
```
float64_t mem = SG_MALLOC(float64_t, size);
....
SGVector<float64_t> v(mem, size, false);
```
is totally valid as in this case the mem management is not being handed over...

here's a patch that actually solves a problem like this:
https://github.com/shogun-toolbox/shogun/pull/4719/commits/edf6f4d2783fd9c870a06ddc94ee9fdd87d196f2

in order to find these cases the best course is to actually do:
```
cd src/shogun
git grep SG_MALLOC
```
and check what are those malloc are actually doing.

this is required to be fixed all along the whole library in order to get the windows build working properly....

tanaymeh:
I would like to fix a few of these throughout the codebase.
I will get back with PR as soon as I have done some work!

vigsterkr:
currently failing tests are:
KMeans.minibatch_fixed_centers
MeanRule.combine_matrix
MalsarTraceTest.train
MalsarClusteredTest.train
MalsarL12Test.train
NormOneTest.apply_to_vector
RandomForestTest.score_compare_sklearn_toydata
CARTree.comparable_with_sklearn
MulticlassLibLinearTest.train_and_apply
BaggingMachineTest.output_binary
Math.linspace_test
SGSparseMatrix.io_libsvm
ROCEvaluation.one

these will definitely have somewhere such a case mentioned above

Aakash1822:
Interested to work on it @vigsterkr. Please guide me

vigsterkr:
@Aakash1822 i've given example in the description and there's the list of the errors failing. you need to find similar (like in the example) the classes that are being used in those tests

PR #4884, opened 2020-03-05
Fixed SVMLight.cpp SG_MALLOC

jonpsy:
My first pull request for fixing SG_MALLOC, do comment !

@vigsterkr  @gf712

### DIFF
--- a/src/shogun/classifier/svm/SVMLight.cpp
+++ b/src/shogun/classifier/svm/SVMLight.cpp
@@ -1584,18 +1584,18 @@ void SVMLight::update_linear_component_mkl(
 	else // hope the kernel is fast ...
 	{
 		float64_t* w_backup = SG_MALLOC(float64_t, num_kernels);
-		float64_t* w1 = SG_MALLOC(float64_t, num_kernels);
+		SGVector<float64_t> weight_vector(num_weights);
 
 		// backup and set to zero
 		for (int32_t i=0; i<num_kernels; i++)
-		{
 			w_backup[i] = old_beta[i] ;
-			w1[i]=0.0 ;
-		}
+		
 		for (int32_t n=0; n<num_kernels; n++)
 		{
-			w1[n]=1.0 ;
-			kernel->set_subkernel_weights(SGVector<float64_t>(w1, num_weights,false));
+			
+			memset(weight_vector,0,num_kernels);
+			weight_vector[n] = 1.0;
+			kernel->set_subkernel_weights(weight_vector);
 
 			for (int32_t i=0;i<num;i++)
 			{
@@ -1605,14 +1605,14 @@ void SVMLight::update_linear_component_mkl(
 						W[j*num_kernels+n]+=(a[i]-a_old[i])*compute_kernel(i,j)*(float64_t)label[i];
 				}
 			}
-			w1[n]=0.0 ;
+			
 		}
 
 		// restore old weights
 		kernel->set_subkernel_weights(SGVector<float64_t>(w_backup,num_weights));
 
 		SG_FREE(w_backup);
-		SG_FREE(w1);
+		
 	}
 
 	call_mkl_callback(a, label, lin);
--- a/src/shogun/classifier/svm/SVMLight.cpp
+++ b/src/shogun/classifier/svm/SVMLight.cpp
@@ -1592,11 +1592,8 @@ void SVMLight::update_linear_component_mkl(
 		
 		for (int32_t n=0; n<num_kernels; n++)
 		{
-			
-			memset(w1,0,num_kernels);
 			w1[n] = 1.0;
 			kernel->set_subkernel_weights(w1);
-
 			for (int32_t i=0;i<num;i++)
 			{
 				if(a[i] != a_old[i])
@@ -1605,7 +1602,7 @@ void SVMLight::update_linear_component_mkl(
 						W[j*num_kernels+n]+=(a[i]-a_old[i])*compute_kernel(i,j)*(float64_t)label[i];
 				}
 			}
-			
+			w1[n] = 0.0;
 		}
 
 		// restore old weights
@@ -1634,7 +1631,7 @@ void SVMLight::update_linear_component_mkl_linadd(
 	ASSERT(num_weights==num_kernels)
 
 	float64_t* w_backup = SG_MALLOC(float64_t, num_kernels);
-	float64_t* w1 = SG_MALLOC(float64_t, num_kernels);
+	SGVector<float64_t> w1(num_weights);
 
 	// backup and set to one
 	for (int32_t i=0; i<num_kernels; i++)
@@ -1643,7 +1640,7 @@ void SVMLight::update_linear_component_mkl_linadd(
 		w1[i]=1.0 ;
 	}
 	// set the kernel weights
-	kernel->set_subkernel_weights(SGVector<float64_t>(w1, num_weights));
+	kernel->set_subkernel_weights(w1);
 
 	// create normal update (with changed alphas only)
 	kernel->clear_normal();
--- a/src/shogun/classifier/svm/SVMLight.cpp
+++ b/src/shogun/classifier/svm/SVMLight.cpp
@@ -1584,7 +1584,7 @@ void SVMLight::update_linear_component_mkl(
 	else // hope the kernel is fast ...
 	{
 		float64_t* w_backup = SG_MALLOC(float64_t, num_kernels);
-		SGVector<float64_t> weight_vector(num_weights);
+		SGVector<float64_t> w1(num_weights);
 
 		// backup and set to zero
 		for (int32_t i=0; i<num_kernels; i++)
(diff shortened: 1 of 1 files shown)


src/shogun/classifier/svm/SVMLight.cpp
```diff
@@ -1595,7 +1595,7 @@ void SVMLight::update_linear_component_mkl(
 		for (int32_t n=0; n<num_kernels; n++)
 		{
 			w1[n]=1.0 ;
-			kernel->set_subkernel_weights(SGVector<float64_t>(w1, num_weights));
+			kernel->set_subkernel_weights(SGVector<float64_t>(w1, num_weights,false));
```
vigsterkr: this would simply result in a memory leak, as `w1` would not be released later. you should fix the root of this, i.e. how `w1` is actually created

vigsterkr (changes requested):

src/shogun/classifier/svm/SVMLight.cpp
jonpsy: There is a SG_FREE(w1). I think [here](https://github.com/shogun-toolbox/shogun/issues/4871) you've mentioned that **false** would be fine because mem management is not handed over. 
I'm a bit confused

vigsterkr: mmm i think then you misunderstood my description of the issue....
what it was saying there that if you encounter a case where the `SGVector<float64_t>(memory, size, false)` constructor is being called that case should be ignored as that is fine...

jonpsy: I've made a workaround without using w1, is it valid ? 
```
	SGVector<float64_t> out(num_weights);
		
		
		for index in num_kernels:	
			memset(out,0,num_kernels);
			out[index] = 1.0;
			kernel->set_subkernel_weights(out); 
                        / * other codes */
```

vigsterkr: so i'd go around actually fixing the creation of w1

vigsterkr: i'm not so sure what this supposed to be? a metacode? for which section? plz push code that's way more easier to review than comments

jonpsy: Okay, just a second

src/shogun/classifier/svm/SVMLight.cpp
```diff
@@ -1584,18 +1584,18 @@ void SVMLight::update_linear_component_mkl(
 	else // hope the kernel is fast ...
 	{
 		float64_t* w_backup = SG_MALLOC(float64_t, num_kernels);
-		float64_t* w1 = SG_MALLOC(float64_t, num_kernels);
+		SGVector<float64_t> weight_vector(num_weights);
```
vigsterkr: keep the name `w1`

src/shogun/classifier/svm/SVMLight.cpp
```diff
@@ -1584,18 +1584,18 @@ void SVMLight::update_linear_component_mkl(
...
 			w_backup[i] = old_beta[i] ;
-			w1[i]=0.0 ;
-		}
+		
 		for (int32_t n=0; n<num_kernels; n++)
 		{
-			w1[n]=1.0 ;
-			kernel->set_subkernel_weights(SGVector<float64_t>(w1, num_weights,false));
+			
+			memset(weight_vector,0,num_kernels);
```
vigsterkr: there are 2 problems here:
 1) you set the whole `weight_vector` to 0 in the beginning of every iteration... hence you are actually nullifying the result of the previous iteration.
 2) there's no need to set the vector 0 in the beginning (above) since if you check SGVector's constructor you'll see that the allocated memory is actually set to 0 right after allocation

jonpsy (commented):
By using this, we don't have to depend on w1, where weight_vector is defined as 
``` SGVector<float64_t> weight_vector(num_weights); ``` before the loop

File: src/shogun/classifier/svm/SVMLight.cpp [hunk 4]
```diff
@@ -1605,14 +1605,14 @@ void SVMLight::update_linear_component_mkl(
 						W[j*num_kernels+n]+=(a[i]-a_old[i])*compute_kernel(i,j)*(float64_t)label[i];
 				}
 			}
-			w1[n]=0.0 ;
+
```
vigsterkr: why do you remove this one?

src/shogun/classifier/svm/SVMLight.cpp
```diff
@@ -1605,14 +1605,14 @@ void SVMLight::update_linear_component_mkl(
...
-			w1[n]=0.0 ;
+			
 		}
 
 		// restore old weights
 		kernel->set_subkernel_weights(SGVector<float64_t>(w_backup,num_weights));
 
 		SG_FREE(w_backup);
-		SG_FREE(w1);
+
```
vigsterkr: extra empty line... this should be simply removed

vigsterkr (changes requested):

src/shogun/classifier/svm/SVMLight.cpp
jonpsy: Noted

src/shogun/classifier/svm/SVMLight.cpp
jonpsy: Ok, so I should just use 
```
loop {
w1[n] = 1.0
/* code */
w1[n] = 0.0 
}
```

src/shogun/classifier/svm/SVMLight.cpp
jonpsy: I'll correct that

src/shogun/classifier/svm/SVMLight.cpp
jonpsy: Correct me if i'm wrong, w1 vector is set to ``` [ 0 0 0 . . .  0] ``` at the start of every iteration of loop, so I just put ```memset(w1,0,num_weights)``` at the beginning of each loop

vigsterkr: this only sets the nth element to 0... not the whole vector...

jonpsy: True, but we already have 
```
for (int32_t i=0; i<num_kernels; i++)
		{
			w_backup[i] = old_beta[i] ;
			w1[i]=0.0 ;        // All the values are zero
		}
``` 
And then in the loop :
```
loop {
w1[n] = 1       // some nth element in num_kernel
/*code*/
w1[n] = 0     //Value is back to 0
//By the end of this iteration w1 = [ 0 0 . . ]
}
```

vigsterkr: indeed. but setting the whole vector 0 instead of just one element is more costly and for nothing, as the other elements are already 0 as you have pointed out above.

jonpsy: Hmm. You're right, so shall I make the above-mentioned change? (w1[n] = 1, w1[n] = 0 instead of memset)

src/shogun/classifier/svm/SVMLight.cpp
```diff
@@ -1584,18 +1584,18 @@ void SVMLight::update_linear_component_mkl(
...
-		{
 			w_backup[i] = old_beta[i] ;
-			w1[i]=0.0 ;
-		}
+		
 		for (int32_t n=0; n<num_kernels; n++)
 		{
-			w1[n]=1.0 ;
-			kernel->set_subkernel_weights(SGVector<float64_t>(w1, num_weights));
+
```
vigsterkr: extra line...

src/shogun/classifier/svm/SVMLight.cpp
```diff
@@ -1584,18 +1584,18 @@ void SVMLight::update_linear_component_mkl(
...
 			w_backup[i] = old_beta[i] ;
-			w1[i]=0.0 ;
-		}
+		
 		for (int32_t n=0; n<num_kernels; n++)
 		{
-			w1[n]=1.0 ;
-			kernel->set_subkernel_weights(SGVector<float64_t>(w1, num_weights));
+			
+			memset(w1,0,num_kernels);
```
vigsterkr: this zeros out the whole vector.... instead of just one

src/shogun/classifier/svm/SVMLight.cpp
```diff
@@ -1584,18 +1584,18 @@ void SVMLight::update_linear_component_mkl(
...
-		float64_t* w1 = SG_MALLOC(float64_t, num_kernels);
+		SGVector<float64_t> w1(num_weights);
 
 		// backup and set to zero
 		for (int32_t i=0; i<num_kernels; i++)
-		{
 			w_backup[i] = old_beta[i] ;
-			w1[i]=0.0 ;
-		}
+
```
vigsterkr: extra line

vigsterkr (changes requested):

src/shogun/classifier/svm/SVMLight.cpp
vigsterkr: i'm not so sure why you keep trying to change the algo itself. the task is really just to replace the malloc with sgvector. you shouldn't have touched any of this code, it is 'good' as is...

the only thing that makes sense to remove is the initial zeroing of the vector as the SGVector constructor takes care of that.

jonpsy: Sorry I thought maybe we could optimize the algorithm, haha I got too excited !

jonpsy:
I noticed we had similar problem for w_backup also. So I fixed that too, I think SVMLight.cpp should work fine now :)

jonpsy:
If you approve of these changes I'll just do a git squash on this pull request.

vigsterkr:
no need i can squash from here ;)

PR #4889, opened 2020-03-06
Renamed "mus" parameter to "cluster_centers"

hmrishavbandy:


src/shogun/clustering/KMeansBase.h
```diff
@@ -130,13 +130,13 @@ class KMeansBase : public RandomMixin<DistanceMachine>
 		SGVector<float64_t> R;
 
 		/** Initial centers supplied */
-		SGMatrix<float64_t> mus_initial;
+		SGMatrix<float64_t> initial_centers;
```
vigsterkr, line 133: have you actually tried to compile this change? coz i believe it will actually fail during compilation. there are registration methods, see SG_ADD macro and watch_method and watch_param, which are actually responsible for the name of the parameter that you can get/set via the get/set api of SGObject.

hmrishavbandy: I haven't tried to compile the change as such . This actually looked a tad bit too easy and that was really naive on my part. I will refactor my code and submit it. Apologies for being hasty.

vigsterkr: yeah as a general rule it'd be good if you try to compile it locally and send a patch if it does actually compile.

hmrishavbandy:
Should I close this pull request and start fresh?

vigsterkr:
no need, you can just keep on doing your fix in this PR

PR #4893, opened 2020-03-08
refactor types in graph

vigsterkr:


src/shogun/mathematics/graph/Tensor.h
```diff
@@ -68,29 +67,29 @@ namespace shogun
 				if (m_data != nullptr)
 					error("Tensor already owns data!");
 				set_shape(shape);
-				m_data = allocator_dispatch(size(), m_type);
+				m_data = malloc(size_in_bytes());
```
gf712: should we use aligned memory?

src/shogun/mathematics/graph/Types.cpp
```diff
@@ -0,0 +1,156 @@
...
+{
+    namespace graph
+    {
+        template <>
+        std::shared_ptr<NumberType> from<bool>()
+        {
+            return boolean();
+        }
+        template <>
+        std::shared_ptr<NumberType> from<float>()
```
gf712, line 75: should this be float32_t? Not sure if it makes a difference?

src/shogun/mathematics/graph/Types.cpp
```diff
@@ -0,0 +1,156 @@
...
+        {
+            return float32();
+        }
+        template <>
+        std::shared_ptr<NumberType> from<double>()
```
gf712, line 80: same here. should this be float64_t?

src/shogun/mathematics/graph/Types.cpp
```diff
@@ -0,0 +1,156 @@
...
+                case element_type::UINT32:
+                    return uint32();
+                case element_type::UINT64:
+                    return uint64();
+                case element_type::FLOAT32:
+                    return float32();
+                case element_type::FLOAT64:
+                    return float64();
+                default:
+                    throw std::invalid_argument("Unknown type");
```
gf712: I think it would be better to have the error after the switch statement. So we get a compile time warning if a new type is added but there is no case to handle it

src/shogun/mathematics/graph/Tensor.h
vigsterkr: mmm yeah we could actually do SG_ALIGNED_MALLOC and take it from there

src/shogun/mathematics/graph/Types.h
```diff
@@ -27,110 +29,200 @@ namespace shogun
...
+		{
+			public:
+				static constexpr const char* type_name() { return "uint8"; }
+		};
+
+		class SHOGUN_ENGINE_EXPORT Int8Type:
+			public detail::IntegerTypeImpl<Int8Type, element_type::INT8, int8_t>
+		{
+			public:
+				static constexpr const char* type_name() { return "uint8"; }
```
gf712: `"int8_t"`

src/shogun/mathematics/graph/Types.h
```diff
@@ -27,110 +29,200 @@ namespace shogun
...
+		{
+			public:
+				Precision precision() const override;
+				static constexpr const char* type_name() { return "float64"; }
+		};
+
+		template <typename T>
+		std::shared_ptr<NumberType> from()
+		{
+			throw std::invalid_argument("Unknown type");
```
gf712: Could this be a static_assert?

src/shogun/mathematics/graph/nodes/Sign.h
```diff
@@ -23,7 +23,7 @@ namespace shogun
 				Sign(const std::shared_ptr<Node>& node) : UnaryNode(node)
 				{
 					const auto& type = m_types[0];
-					if (is_unsigned(type))
+					if (!type->is_signed())
```
gf712, line 26: cool!

src/shogun/mathematics/graph/test/GraphTest.h
```diff
@@ -54,8 +53,13 @@ class GraphTest : public ::testing::Test
...
+
 using GraphTypes = ::testing::Types<
-    bool, int8_t, int16_t, int32_t, int64_t, uint8_t, uint16_t, uint32_t,
-    uint64_t, float32_t, float64_t>;
+    shogun::graph::BooleanType, shogun::graph::UInt8Type,
+	shogun::graph::Int8Type, shogun::graph::UInt16Type,
+	shogun::graph::Int16Type, shogun::graph::UInt32Type,
+	shogun::graph::Int32Type, shogun::graph::Int64Type,
+	shogun::graph::UInt64Type, shogun::graph::Float32Type,
+	shogun::graph::Float64Type>;
```
gf712, line 63: Somewhat unrelated, but I am thinking if it would be good to add ::testing::Types, so we can exclude bool, or just have bool, and not unsigned, and so on..? Most tests have a `if constexpr` which doesn't test anything in one branch...

gf712:
Thanks! Looks good to me :)

src/shogun/mathematics/graph/test/GraphTest.h
vigsterkr: imo this is doable, with a bit different gtest construct, i.e. test suites.

PR #4897, opened 2020-03-08
Perceptron example

gf712:
@vigsterkr what should we use to generate some random data?

### DIFF
--- a/src/shogun/mathematics/graph/Graph_test.cc
+++ b/src/shogun/mathematics/graph/Graph_test.cc
@@ -1,0 +1,102 @@
+#include <gtest/gtest.h>
+
+#include "Graph.h"
+#include "nodes/Add.h"
+#include "nodes/Dot.h"
+#include "nodes/Input.h"
+#include "nodes/Multiply.h"
+#include "nodes/Sign.h"
+#include "nodes/Cast.h"
+#include "nodes/Equal.h"
+
+#include "test/GraphTest.h"
+
+#include <shogun/features/DataGenerator.h>
+
+#include <random>
+
+using namespace shogun;
+using namespace shogun::graph;
+using namespace std;
+
+TYPED_TEST(GraphTest, perceptron_stochastic_gradient_descent)
+{
+#if 0
+	using NumericType = typename TypeParam::c_type;
+
+	if constexpr (!std::is_same_v<NumericType, float64_t>)
+		return;
+	else
+	{
+		size_t max_iters = 100; 
+		int64_t num_feat = 5;
+		int64_t num_examples = 1000;
+		float64_t learning_rate = 0.001;
+
+
+		auto X = make_shared<node::Input>(Shape{num_feat}, element_type::FLOAT64);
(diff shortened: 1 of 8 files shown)


src/shogun/mathematics/graph/Graph_test.cc
```diff
@@ -0,0 +1,70 @@
+#include <gtest/gtest.h>
+
+#include "Graph.h"
+#include "nodes/Add.h"
+#include "nodes/Dot.h"
+#include "nodes/Input.h"
+#include "nodes/Multiply.h"
+#include "nodes/Sign.h"
+#include "nodes/Cast.h"
+#include "nodes/Equal.h"
```
gf712, line 10: @vigsterkr should probably put all nodes in a `graph/Nodes.h` file?

src/shogun/mathematics/graph/Graph_test.cc
```diff
@@ -0,0 +1,70 @@
...
+				for (const auto& idx : range(features.num_rows))
+				{
+					auto update = graph->evaluate(vector{make_shared<Tensor>(features.get_row_vector(idx)), 
+						make_shared<Tensor>(labels[idx]),
+						make_shared<Tensor>(w_vector),
+						make_shared<Tensor>(b_value)});
+					w_vector = update[0]->template as<SGVector<NumericType>>();
+					b_value = update[1]->template as<NumericType>();
+				}
+			}
```
gf712, line 84: @karlnapf it isn't jax, but once we get something like this working we can start transforming this into something more functional :) already have a design in mind!

vigsterkr:
@gf712 https://www.shogun.ml/api/latest/classshogun_1_1CDataGenerator.html ?

src/shogun/mathematics/graph/Graph_test.cc
```diff
@@ -0,0 +1,70 @@
...
+		auto w_vector = SGVector<float64_t>(num_feat);
+		w_vector.set_const(NumericType{1} / num_feat);
+		NumericType b_value{0};
+		SGMatrix<NumericType> features(num_examples, num_feat);
+		SGVector<NumericType> labels(num_examples);
+
+		for (auto&& backend : this->m_backends)
+		{
+			if (backend != GRAPH_BACKEND::SHOGUN)
+				continue;
```
gf712, line 67: @vigsterkr ngraph hangs when i run this, probably it is rebuilding the graph in each loop, because right now the graph is built during `execute` in the case of ngraph

gf712:
> @gf712 https://www.shogun.ml/api/latest/classshogun_1_1CDataGenerator.html ?

OK! And then at some point we need to write have a standalone thing? But I guess that's quite a bit further down the line

vigsterkr:
yeye most def.... i mean we should anyways have a much better random initializer for input :P

gf712:
@vigsterkr not sure if this is working, but I made some changes here that I'll need to start working on some other stuff.. Basically I want to generate data, but that requires linalg (required by Gaussian). So the perceptron example is there now but won't run it in the test suite until we sort out a better way to generate some random data.

PR #4899, opened 2020-03-09
add feature importances for RandomForest

LiuYuHui:


vigsterkr (changes requested):
great! could you please add a `watch_method` so that this is exposed via the parameter framework.

src/shogun/machine/RandomForest.h
```diff
@@ -137,6 +137,12 @@ class RandomForest : public BaggingMachine
 	 * @return number of randomly chosen features during each node split
 	 */
 	int32_t get_num_random_features() const;
+	/** get feature importances of previous trained, use Mean Decrease
+	 * Impurity(MDI)
+	 *
+	 * @return arrays of feature importance
+	 */
+	SGVector<float64_t> get_feature_importances();
```
gf712: If you don't cache the feature importances you could make this `const`, no?

LiuYuHui: ok,i will change it :)

src/shogun/machine/RandomForest.cpp
```diff
@@ -182,10 +183,32 @@ bool RandomForest::train_machine(std::shared_ptr<Features> data)
 	return BaggingMachine::train_machine();
 }
 
+SGVector<float64_t> RandomForest::get_feature_importances() const
```
karlnapf, line 186: minor: is the name "get" appropriate, rather than "compute"?

Issue #4259, opened 2018-04-16
remove num_lhs field in CKernel


karlnapf:
The field only causes problems. This is a refactoring exercise to remove the fiels, and instead query the underlying features.

Special cases might be `CCustomKernel` where no features are set. However, the base class should always query the underlying features.

The member function
```
virtual int32_t get_num_vec_lhs()
		{
			return num_lhs;
		}
```
would be changed to
virtual int32_t get_num_vec_lhs()
		{
			return lhs->get_num_vectors();
		}
```
or similar.

karlnapf:
it involves doing these changes and then checking whether they break anything. This involves reading the code, thinking, and running the tests. Please check here how to get started: https://github.com/shogun-toolbox/shogun/wiki/Getting-involved

divyanshu132:
@karlnapf I'm not able to perform all the tests, can you please help me out!!

vishal3410:
Hello @divyanshu132 I am working on this issue if u don't mind can you please leave this issue to me , i will let u know if i am not able to fix this

divyanshu132:
@vishal3410 no problem, it's completely fine!!

PR #4905, opened 2020-03-10
Removing num_lhs and num_rhs from Kernel

Khalifa1997:
In response to https://github.com/shogun-toolbox/shogun/issues/4259

### DIFF
--- a/src/shogun/kernel/Kernel.h
+++ b/src/shogun/kernel/Kernel.h
@@ -509,7 +509,7 @@ class Kernel : public SGObject
 		 */
 		virtual int32_t get_num_vec_lhs()
 		{
-			return num_lhs;
+			return lhs->get_num_vectors();
 		}
 
 		/** get number of vectors of rhs features
@@ -518,7 +518,7 @@ class Kernel : public SGObject
 		 */
 		virtual int32_t get_num_vec_rhs()
 		{
-			return num_rhs;
+			return rhs->get_num_vectors();
 		}
 
 		/** test whether features have been assigned to lhs and rhs
--- a/src/shogun/kernel/Kernel.h
+++ b/src/shogun/kernel/Kernel.h
@@ -199,9 +199,9 @@ class Kernel : public SGObject
 		 */
 		inline float64_t kernel(int32_t idx_a, int32_t idx_b)
 		{
-			require(idx_a>=0 && idx_b>=0 && idx_a<num_lhs && idx_b<num_rhs,
+			require(idx_a>=0 && idx_b>=0 && idx_a<get_num_vec_lhs() && idx_b<get_num_vec_rhs(),
 				"{}::kernel(): index out of Range: idx_a={}/{} idx_b={}/{}",
-				get_name(), idx_a,num_lhs, idx_b,num_rhs);
+				get_name(), idx_a,get_num_vec_lhs(), idx_b,get_num_vec_rhs());
 
 			return normalizer->normalize(compute(idx_a, idx_b), idx_a, idx_b);
 		}
@@ -256,9 +256,9 @@ class Kernel : public SGObject
 		virtual SGVector<float64_t> get_kernel_col(int32_t j)
 		{
 
-			SGVector<float64_t> col = SGVector<float64_t>(num_rhs);
+			SGVector<float64_t> col = SGVector<float64_t>(get_num_vec_rhs());
 
-			for (int32_t i=0; i!=num_rhs; i++)
+			for (int32_t i=0; i!=get_num_vec_rhs(); i++)
 				col[i] = kernel(i,j);
 
 			return col;
@@ -272,9 +272,9 @@ class Kernel : public SGObject
 		 */
 		virtual SGVector<float64_t> get_kernel_row(int32_t i)
 		{
-			SGVector<float64_t> row = SGVector<float64_t>(num_lhs);
+			SGVector<float64_t> row = SGVector<float64_t>(get_num_vec_lhs());
 
-			for (int32_t j=0; j!=num_lhs; j++)
+			for (int32_t j=0; j!=get_num_vec_lhs(); j++)
 				row[j] = kernel(i,j);
 
 			return row;


Khalifa1997:
@vigsterkr @karlnapf How would you go around about fixing those test files?
I reckon the line `SG_ADD( &num_rhs, "num_rhs", "Number of feature vectors on right hand side.");` in `Kernel::register_params()`
might be an issue, cause it's what's stopping me from removing all the num_rhs/num_lhs in the Kernel.cpp, then move forward and try to remove them from other kernels aswell.

LiuYuHui:
@Khalifa1997 , you can delete SG_ADD and then use watch_method instead

Khalifa1997:
> @Khalifa1997 , you can delete SG_ADD and then use watch_method instead

Interesting can you point me to the documentation please :)

gf712:
> > @Khalifa1997 , you can delete SG_ADD and then use watch_method instead
> 
> Interesting can you point me to the documentation please :)

do you know how to proceed? :)

Khalifa1997:
@gf712 I mean this is what I have done so far, although it does compile but I get many tests failing :(

gf712:
> @gf712 I mean this is what I have done so far, although it does compile but I get many tests failing :(

hmm, didn't we discuss on riot what had to be done? Sorry I am not sure anymore!

Khalifa1997:
@gf712 What we discussed was I should trace the failing tests to know what was failing however, even by using `ctest --output-on-failure` the output is just the file names

gf712:
`ctest --output-on-failure --verbose`?

Khalifa1997:
> `ctest --output-on-failure --verbose`?
I have added one failing test and the last portion of the my out when I run this.
https://pastebin.com/Abz8TE7J

vigsterkr:
yep those segfaults would need to be debugged. see manual for gdb or lldb

vigsterkr:
although i'm pretty sure that the segfault is because there's a null pointer deference when calling the get_num_... functions... you would need to check whether actually lhs or rhs is set (i.e. not null).

src/shogun/kernel/Kernel.h
```diff
@@ -509,7 +509,7 @@ class Kernel : public SGObject
 		 */
 		virtual int32_t get_num_vec_lhs()
 		{
-			return num_lhs;
+			return lhs->get_num_vectors();
```
vigsterkr, line 512: what if lhs == nullptr?

src/shogun/kernel/Kernel.h
```diff
@@ -518,7 +518,7 @@ class Kernel : public SGObject
 		 */
 		virtual int32_t get_num_vec_rhs()
 		{
-			return num_rhs;
+			return rhs->get_num_vectors();
```
vigsterkr, line 521: what if rhs == nullptr?

vigsterkr (changes requested):

src/shogun/kernel/Kernel.h
Khalifa1997: Hmm true, should it be something along the lines if(lhs) return lhs->get_num_vectors(); else return 0;

vigsterkr: well check what's the initial value of num_lhs and that should be returned if lhs is nullptr

Khalifa1997: I think it's intialized by zero cause that's according to `/shogun/tests/unit/kernel/ProductKernel_unittest.cc` which is one of the failing SEGFAULT tests

`append_kernel` uses `adjust_num_lhs_rhs_initialized` function that basically checks if `num_lhs` isn't equal to zero if it is it sets it using `get_num_vec_lhs`

also `init` function in Kernel sets the `num_lhs` to zero intially

vigsterkr: init does the initialization... hence that should be returned (0) if lhs or rhs is not set

Khalifa1997: @vigsterkr those segfaults errors are now fixed, how do you suggest I go about the other errors I have traced back the remaining errors, it's basically the same error over an over again say we take the `tests/unit/statistical_testing/internals/MultiKernelMMD_unittest.cc` test
on line 88 we start by constructing some new Gaussian vectors, which in turns calls `ShiftInvariantKernel::ShiftInvariantKernel(const std::shared_ptr<Features >&l, const std::shared_ptr<Features >&r) : Kernel(l, r, 0)` constructor which calls the kernel constructor setting `get_num_vec_lhs` and `get_num_vec_rhs` to be zero -since `lhs` and `rhs` are already nullptr.
Now the problem comes from the `inline float64_t kernel(int32_t idx_a, int32_t idx_b)` function inside kernel which has this require that would always break since idx_a is equal to zero and it has to be less than get_num_vec_lhs which in turn is equal to zero. anyways I will upload the pastebin so it may make a tiny bit more sense :grin: 
https://pastebin.com/dcntwX64

vigsterkr: sorry but it's really hard to debug by description. you'll have to try to resolve this yourself or create a very simple use-case where this specific bug is reproducible

Khalifa1997: @vigsterkr Thanks I will work on this

vigsterkr:
any update on this?

src/shogun/kernel/Kernel.h
gf712, line 521: Can you add something like this please? For both examples. Feel free to improve the error message
```suggestion
			require(rhs, "Kernel::rhs has not been set");
			return rhs->get_num_vectors();
```

karlnapf: Actually, the error message should read "Kernel right hand side features not set"

(User facing, to avoid putting variable/member names in error messages, but instead put the semantic equivalent)

gf712:
Almost there! Just see the comment and it should be good to merge.

PR #4901, opened 2020-03-09
Separate tensor from storage

gf712:
Separates abstraction of data locality and Tensors.
 - Tensors should be user facing
 - ShogunStorage is internal

Adds automatic memory management of ShogunStorage with std::shared_ptr, and guarantees that the correct deleter is called

PR #4907, opened 2020-03-10
exp op

gf712:


src/shogun/mathematics/graph/Tensor.h
```diff
@@ -174,17 +174,17 @@ namespace shogun
 					if (m_shape.size() > 1)
 						error("Tried to cast a multidimensional Tensor to a "
 						      "SGVector.");
-					if (from<typename Container::Scalar>() != m_type)
-						error("Type mismatch when casting from Tensor.");
+					if (*from<typename Container::Scalar>() != *m_type)
```
gf712: this is just temporary. I fixed the underlying problem in #4901

src/shogun/mathematics/graph/nodes/Exp_test.cc
```diff
@@ -0,0 +1,50 @@
...
+
+		auto input = make_shared<node::Input>(
+		    Shape{Shape::Dynamic}, TypeParam::type_id);
+
+		auto output = make_shared<node::Exp>(input);
+
+		auto graph = make_shared<Graph>(
+		    vector{input}, vector<shared_ptr<node::Node>>{output});
+
+		for (auto&& backend : this->m_backends)
```
karlnapf, line 34: question: is it good that tests for ops need to have all this stuff in them? I.e. I think it should be the same roughly for other ops, so some code could be reused?

gf712: @karlnapf I just wanted to be explicit with tests for now, otherwise I get myself lost because it's changing so often.. Once the API is more stable and logical, and the code works fine, I'll refactor the tests :)

gf712: also, I am working on the Array version now, which basically brings the functional version, and then need to combine that to the tests too

karlnapf: makes sense!

PR #4914, opened 2020-03-11
fix the sgvector sgmatrix conversion in shogun interface

vigsterkr:


gf712:
@LiuYuHui this fixes the java issue

PR #4916, opened 2020-03-12
fix hmm parallel

### LINKED ISSUE #4909: USE_HMMPARALLEL causes compiler errors
Currently there are compilation errors when compiling with `-DUSE_HMMPARALLEL=ON`. See for example [here](https://buildbot.shogun.ml/#/builders/22/builds/47/steps/5/logs/stdio).
I have not tried to compile locally and there could be a lot that has to be fixed in that code. Ideally all of this would be replaced with `std::thread`.

gf712:
Fix the parallel code in HMM. Also replaces pthread with the C++ standard std::thread.
Also fixed a potential data race.
fixes #4909

### DIFF
--- a/src/shogun/base/SGObject.h
+++ b/src/shogun/base/SGObject.h
@@ -961,6 +961,14 @@ protected:
 	 */
 	void add_callback_function(
 		std::string_view name, std::function<void()> method);
+
+	/** Get the shared_ptr from SGObject and cast it to the Derived class
+	 */
+	template <typename Derived>
+	std::shared_ptr<Derived> shared_from_base()
+	{
+		return std::static_pointer_cast<Derived>(shared_from_this());
+	}
 #endif
 
 public:
--- a/src/shogun/distributions/HMM.h
+++ b/src/shogun/distributions/HMM.h
@@ -380,7 +380,7 @@ class HMM : public RandomMixin<Distribution>
 		/// Datatype that is used in parrallel computation of viterbi
 		struct S_DIM_THREAD_PARAM
 		{
-			std::shared_ptr<CHMM> hmm;
+			std::shared_ptr<HMM> hmm;
 			int32_t dim;
 			float64_t prob_sum;
 		};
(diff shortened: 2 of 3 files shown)


src/shogun/base/SGObject.h
```diff
@@ -961,6 +961,14 @@ class SGObject: public std::enable_shared_from_this<SGObject>
...
 	void add_callback_function(
 		std::string_view name, std::function<void()> method);
+
+	/** Get the shared_ptr from SGObject and cast it to the Derived class
+	 */
+    template <typename Derived>
+    std::shared_ptr<Derived> shared_from_base()
+    {
+        return std::static_pointer_cast<Derived>(shared_from_this());
+    }
```
gf712: @vigsterkr this is useful when you need shared_from_this in a derived class

src/shogun/distributions/HMM.h
```diff
@@ -1240,8 +1240,12 @@ class HMM : public RandomMixin<Distribution>
 		float64_t mod_prob;
 
 		/// true if model probability is up to date
+		#ifdef USE_HMMPARALLEL
+		std::atomic<bool> mod_prob_updated;
```
gf712, line 1244: @vigsterkr not sure who the expert on this algo is, but I am pretty sure if you are using multiple threads this variable causes a race condition

vigsterkr: git blame? but honestly i think this algo has not been touched and only lingered around with minor fixups for a loooooong time

gf712:
@vigsterkr there are no tests to prove, but HMM parallel should be working now. I tested it on the two undocumented examples where it is used and no more segfault :) 
One of the buildbot nightly jobs should run this though

src/shogun/distributions/HMM.h
gf712: indeed, last time this seemed to have been actively developed was 2010 or so :D

vigsterkr:
💃 merge it!

gf712:
> dancer merge it!

yea, once CI approves I'll merge

vigsterkr:
Who needs CI? 🙃 “development on production for ever!” (tm)

gf712:
> Who needs CI? upside_down_face “development on production for ever!” (tm)

Should have merged straight to master and released it :)

PR #4918, opened 2020-03-13
add feature importance for C45

LiuYuHui:
if it look fine, i will update meta example and data :)

### DIFF
--- a/src/shogun/multiclass/tree/C45ClassifierTree.cpp
+++ b/src/shogun/multiclass/tree/C45ClassifierTree.cpp
@@ -32,6 +32,7 @@
 #include <shogun/lib/View.h>
 #include <shogun/mathematics/Math.h>
 #include <shogun/mathematics/Statistics.h>
+#include <shogun/mathematics/linalg/LinalgNamespace.h>
 #include <shogun/multiclass/tree/C45ClassifierTree.h>
 
 using namespace shogun;
@@ -141,8 +142,44 @@ bool C45ClassifierTree::train_machine(std::shared_ptr<Features> data)
 	feature_ids.range_fill();
 
 	set_root(C45train(data, m_weights, multiclass_labels(m_labels), feature_ids, 0));
-
+	// compute feature importances and normalize it
+	if (m_root)
+	{
+		m_feature_importances = SGVector<float64_t>(num_features);
+		m_feature_importances.zero();
+		compute_feature_importance(m_root);
+		float64_t total_num_sample = m_root->data.total_weight;
+		linalg::scale(
+		    m_feature_importances, m_feature_importances,
+		    1.0 / total_num_sample);
+		float64_t normalizer = linalg::sum(m_feature_importances);
+		if (normalizer > 0)
+			linalg::scale(
+			    m_feature_importances, m_feature_importances, 1.0 / normalizer);
+	}
 	return true;
+}
+
+void C45ClassifierTree::compute_feature_importance(
+    const std::shared_ptr<node_t>& node)
+{
+	auto childs = node->get_children();
+	m_feature_importances[node->data.attribute_id] +=
+	    node->data.impurity * node->data.total_weight;
+	for (const auto& child : childs)
+	{
+		m_feature_importances[node->data.attribute_id] -=
+		    child->data.impurity * child->data.total_weight;
+		if (child->data.attribute_id >= 0)
+		{
+			compute_feature_importance(child);
+		}
+	}
+}
+
+SGVector<float64_t> C45ClassifierTree::get_feature_importances() const
+{
+	return m_feature_importances;
 }
 
 std::shared_ptr<TreeMachineNode<C45TreeNodeData>> C45ClassifierTree::C45train(const std::shared_ptr<Features>& data, SGVector<float64_t> weights,
@@ -225,13 +262,17 @@ std::shared_ptr<TreeMachineNode<C45TreeNodeData>> C45ClassifierTree::C45train(co
 
 	// else get the feature with the highest informational gain. threshold is used for continuous features only.
 	float64_t max=0;
+	float64_t impurity = 0.0;
+	float64_t max_impurity = 0.0;
 	int32_t best_feature_index=-1;
 	float64_t threshold=0.;
 	for (int32_t i=0; i<feats->get_num_features(); i++)
 	{
 		if (m_nominal[feature_id_vector[i]])
 		{
-			float64_t gain=informational_gain_attribute(i,feats,weights,class_labels);
+			float64_t gain = informational_gain_attribute(
+			    i, feats, weights, class_labels, impurity);
+			max_impurity = std::max(max_impurity, impurity);
 			if (gain>=max)
 			{
 				max=gain;
@@ -268,7 +309,9 @@ std::shared_ptr<TreeMachineNode<C45TreeNodeData>> C45ClassifierTree::C45train(co
 					}
 
 					auto temp_feats=std::make_shared<DenseFeatures<float64_t>>(temp_feat_mat);
-					float64_t gain=informational_gain_attribute(0,temp_feats,weights,class_labels);
+					float64_t gain = informational_gain_attribute(
+					    0, temp_feats, weights, class_labels, impurity);
+					max_impurity = std::max(max_impurity, impurity);
 					if (gain>max)
 					{
 						threshold=z;
@@ -396,6 +439,7 @@ std::shared_ptr<TreeMachineNode<C45TreeNodeData>> C45ClassifierTree::C45train(co
 		// recursion over child nodes
 		auto child=C45train(new_data,new_weights,new_class_labels,new_feature_id_vector,level+1);
 		node->data.attribute_id=feature_id_vector[best_feature_index];
+		node->data.impurity = max_impurity;
 		if (m_nominal[feature_id_vector[best_feature_index]])
 			child->data.transit_if_feature_value=active_feature_value;
 		else
@@ -538,8 +582,10 @@ void C45ClassifierTree::prune_tree_from_current_node(const std::shared_ptr<Dense
 
 }
 
-float64_t C45ClassifierTree::informational_gain_attribute(int32_t attr_no, const std::shared_ptr<Features>& data,
-				SGVector<float64_t> weights, const std::shared_ptr<MulticlassLabels>& class_labels)
+float64_t C45ClassifierTree::informational_gain_attribute(
+    int32_t attr_no, const std::shared_ptr<Features>& data,
+    SGVector<float64_t> weights,
+    const std::shared_ptr<MulticlassLabels>& class_labels, float64_t& impurity)
 {
 	require(data,"Data required for information gain calculation");
 	require(data->get_feature_class()==C_DENSE,
@@ -626,6 +672,7 @@ float64_t C45ClassifierTree::informational_gain_attribute(int32_t attr_no, const
 	}
 
 	float64_t data_entropy=entropy(gain_labels,gain_weights);
+	impurity = gain;
 	gain = data_entropy-gain;
 
 	if (num_missing!=0)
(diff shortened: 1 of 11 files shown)


src/shogun/multiclass/tree/C45ClassifierTree.cpp
```diff
@@ -141,10 +142,46 @@ bool C45ClassifierTree::train_machine(std::shared_ptr<Features> data)
...
 	return true;
 }
 
+void C45ClassifierTree::compute_feature_importance(
+    const std::shared_ptr<node_t>& node)
+{
+	auto childs = node->get_children();
+	m_feature_importances[node->data.attribute_id] +=
+	    node->data.impurity * node->data.total_weight;
+	for (const auto& child : childs)
```
vigsterkr:
```suggestion
	#pragma omp parallel for shared(m_feature_importances)
	for (const auto& child : childs)
```

not so sure if its worth it though...

src/shogun/multiclass/tree/C45ClassifierTree.cpp
```diff
@@ -766,7 +813,9 @@ void C45ClassifierTree::init()
 	m_certainty=SGVector<float64_t>();
 	m_types_set=false;
 	m_weights_set=false;
-
+	m_feature_importances = SGVector<float64_t>();
+	SG_ADD(
+	    &m_feature_importances, "m_feature_importances", "feature importances");
```
vigsterkr:
```suggestion
	SG_ADD(
	    &m_feature_importances, "feature_importances", "feature importances");
```

vigsterkr: just to be consistent with the other trees

vigsterkr (commented):
some minor things

vigsterkr:
yep lets have the meta examples + data added if possible!

have you actually tested that omp on a bigger tree? :D it should work, but 🤷‍♂

LiuYuHui:
> 
> 
> yep lets have the meta examples + data added if possible!
> 
> have you actually tested that omp on a bigger tree? :D it should work, but 🤷‍♂

maybe i should use benchmark to compare these two situations :D

src/shogun/multiclass/tree/C45ClassifierTree.cpp
```diff
@@ -766,11 +814,12 @@ void C45ClassifierTree::init()
 	m_certainty=SGVector<float64_t>();
 	m_types_set=false;
 	m_weights_set=false;
-
+	m_feature_importances = SGVector<float64_t>();
+	SG_ADD(
+	    &m_feature_importances, "feature_importances", "feature importances");
```
gf712: This should either be registered with watch_method or pass the READONLY flag

gf712 (changes requested):

src/shogun/multiclass/tree/C45ClassifierTree.cpp
LiuYuHui: this ? `SG_ADD(
	    &m_feature_importances, "feature_importances", "feature importances",ParameterProperties::READONLY);`

gf712: yes!

vigsterkr: i've suggested @LiuYuHui that actually maybe it'd be better to unify this for trees... meaning the feature importance...

LiuYuHui: this pr or other pr?

vigsterkr: just do it (here)! (tm)

src/shogun/multiclass/tree/C45ClassifierTree.cpp
```diff
@@ -141,10 +142,47 @@ bool C45ClassifierTree::train_machine(std::shared_ptr<Features> data)
 	feature_ids.range_fill();
 
 	set_root(C45train(data, m_weights, multiclass_labels(m_labels), feature_ids, 0));
-
+	// compute feature importances and normalize it
```
karlnapf: minor: grammar ("it" is singular but importances are plural)

src/shogun/multiclass/tree/C45ClassifierTree.cpp
```diff
@@ -141,10 +142,47 @@ bool C45ClassifierTree::train_machine(std::shared_ptr<Features> data)
...
+	{
+		m_feature_importances = SGVector<float64_t>(num_features);
+		m_feature_importances.zero();
+		compute_feature_importance(m_root);
+		float64_t total_num_sample = m_root->data.total_weight;
+		linalg::scale(
+		    m_feature_importances, m_feature_importances,
+		    1.0 / total_num_sample);
+		float64_t normalizer = linalg::sum(m_feature_importances);
+		if (normalizer > 0)
```
karlnapf: Can it be zero?

src/shogun/multiclass/tree/C45ClassifierTree.cpp
```diff
@@ -141,10 +142,47 @@ bool C45ClassifierTree::train_machine(std::shared_ptr<Features> data)
...
 }
 
+void C45ClassifierTree::compute_feature_importance(
+    const std::shared_ptr<node_t>& node)
+{
+	auto childs = node->get_children();
+	m_feature_importances[node->data.attribute_id] +=
+	    node->data.impurity * node->data.total_weight;
+	#pragma omp parallel for shared(m_feature_importances)
+	for (const auto& child : childs)
```
karlnapf: minor: childs -> children

src/shogun/multiclass/tree/C45ClassifierTree.cpp
```diff
@@ -141,10 +142,47 @@ bool C45ClassifierTree::train_machine(std::shared_ptr<Features> data)
...
+	{
+		m_feature_importances[node->data.attribute_id] -=
+		    child->data.impurity * child->data.total_weight;
+		if (child->data.attribute_id >= 0)
+		{
+			compute_feature_importance(child);
```
karlnapf: is this a thread-safe (const method)?

karlnapf: just checked it is not, then the question is: is the multithreading safe?

src/shogun/multiclass/tree/C45ClassifierTree.h
```diff
@@ -146,6 +146,11 @@ class C45ClassifierTree : public TreeMachine<C45TreeNodeData>
 	/** clear feature types of various features */
 	void clear_feature_types();
 
+	/** get feature importance
```
karlnapf: minor: remove first line of comment, redundant

src/shogun/multiclass/tree/C45ClassifierTree.cpp
vigsterkr: hence shared(m_feature_importances)
but @LiuYuHui is actually working on some ~unification~ generalization atm

karlnapf: I guess shared only protects writing to the array?
but ok, lets see the unified version

src/shogun/multiclass/tree/FeatureImportanceTree.h
```diff
@@ -0,0 +1,69 @@
...
+namespace shogun
+{
+
+	/** @brief class FeatureImportanceTree, a base class for the tree which need
+	 * computing feature importances. This class is derived from TreeMachine<T>
+	 * and stores the feature importances
+	 *
+	 */
+	template <typename T>
+	class FeatureImportanceTree : public TreeMachine<T>
```
vigsterkr: Maybe `FeatureImportanceMixin` would be a better name?
@gf712 @karlnapf ?

gf712: Yes, and change the typename to something meaningful, e.g TreeType

karlnapf: I think having “mixin” in the name is not necessary as it is bulky, would rather put that in the api docstring. But +1 for a meaningful type name

gf712:
Thank you, this LGTM. Maybe let's wait until Azure Pipelines is back and then merge?

src/shogun/multiclass/tree/FeatureImportanceTree.h
```diff
@@ -0,0 +1,69 @@
...
+		virtual ~FeatureImportanceTree() = default;
+
+	private:
+		void compute_feature_importance_impl(
+		    const std::shared_ptr<TreeMachineNode<NodeType>>& node)
+		{
+			const auto& children = node->get_children();
+			m_feature_importances[node->data.attribute_id] +=
+			    node->data.impurity * node->data.total_weight;
+			for (const auto& child : children)
```
gf712: didn't you use omp here in a previous commit?

LiuYuHui: omp seems like don't support with range for

gf712: Ah, that's easy to fix, you just need have a oldschool for loop `for(int i =0; i < children.size(); ++i){auto& child = children[i]; ...}`. If the computation is quite expensive you could just use std::thread instead tbh. What do you think @vigsterkr?

src/shogun/multiclass/tree/FeatureImportanceTree.h
```diff
@@ -53,8 +53,10 @@ namespace shogun
 			const auto& children = node->get_children();
 			m_feature_importances[node->data.attribute_id] +=
 			    node->data.impurity * node->data.total_weight;
-			for (const auto& child : children)
+#pragma omp parallel for shared(m_feature_importances)
+			for (auto i = 0; i < children.size(); i++)
 			{
+				auto& child = children[i];
```
gf712: Const?

gf712: Did you also see any performance improvement with omp?

LiuYuHui: i have tested, but it seems like that there is no significant performance improvement

vigsterkr:
@LiuYuHui i guess the only thing is missing adding the shogun-data, right? could you update the data submodule (i've just merged your PR to shogun-data) and then we can get this merged! great stuff!

karlnapf:
:rocket:

karlnapf:
quick question. How do we know this implementation is correct and no error slipped in? I am always paranoid about that possibility (happened before) Maybe we could have a unit test? Or you think that is not necessary?

vigsterkr:
@karlnapf i was under the impression (i was told) that @LiuYuHui worked based on the scikitlearn implementation... hoping that he has actually checked its correctness :)

karlnapf:
Maybe we can open an entrance task that writes a unit test that compares numbers against sklearn? Or @LiuYuHui could do it? It also covers future refactors.  Just in case something slipped or might slip in the future.

gf712:
The unit tests use sklearn as a reference!

karlnapf:
but do we have one that checks the importances that were added?

vigsterkr:
you mean the toy data based?

LiuYuHui:
> 
> 
> Maybe we can open an entrance task that writes a unit test that compares numbers against sklearn? Or @LiuYuHui could do it? It also covers future refactors. Just in case something slipped or might slip in the future.

i have added CARTree unit test for feature importance, but it seems that sklearn just implement CARTree, and there is no C45 and id3 implementation in sklearn, so i haven't added C45 and id3 unittest

karlnapf:
A unit test of some corner cases where you can compute the results manually in a simple case would also do it. Just anything really that serves as a minimal line of defense against unintended changes later on. Or, alternatively another framework?

Issue #3702, opened 2017-03-15
Broken links in notebooks.


kirstymcnaught:
A few notebooks have web links that are not parsed correctly. Links missing _http[s]://www._ are opened as paths relative to the notebook server. Emails need 'mailto:' if they are intended to be a clickable link.

The following is a list of notebook files and 'bad' links.

doc/ipython-notebooks/clustering/GMM.ipynb
github.com/karlnapf
herrstrathmann.de

doc/ipython-notebooks/clustering/.ipynb_checkpoints/GMM-checkpoint.ipynb
github.com/karlnapf
herrstrathmann.de

ipython-notebooks/evaluation/xval_modelselection.ipynb
github.com/karlnapf
herrstrathmann.de

ipython-notebooks/statistical_testing/mmd_two_sample_testing.ipynb
github.com/karlnapf
herrstrathmann.de
soumyajitde.cse@gmail.com 
github.com/lambday

karlnapf:
Thanks for reporting.
Nice entrance task for GSoC students

deveshnag1:
@karlnapf I would like to work on this. I'm planning to apply for real world applications project. This doesn't come under entrance for any other project right?

karlnapf:
The entrance tasks are not really project related, especially not those easy ones

deveshnag1:
@karlnapf I think this can be closed now

karlnapf:
What about you run a linkchecker and then we close? :)

deveshnag1:
I'm not sure I follow

karlnapf:
I meant: you could run an automated tool that verifies the links in **all** the notebooks in the repository. If broken ones are found, you send another patch, otherwise we close this issue.

deveshnag1:
Sounds interesting. I'll get right to it.

Red-devilz:
Here is a list of broken-links, that are not yet fixed 

https://gist.github.com/Red-devilz/67dee8c8afc2502202b16466ff6da225

karlnapf:
Thanks for that! Really useful to have that list!

bhavukkalra:
Were all the links verified?
Could you please provide the list of broken links above. It doesn't seems to be opening(might be a broken link)
i.e
https://gist.github.com/Red-devilz/67dee8c8afc2502202b16466ff6da225
(list of broken links)(not opening)

vigsterkr:
[...]

PR #4919, opened 2020-03-13
added link checking functionality

bhavukkalra:
This is a feature that was required to be added to CI of shogun to make checking of broken links in Ipython notebooks more streamlined
(Usability add the path of the Ipython notebook file as a command line argument ,OUTPUT- Links that are broken)
also closes issue no. #3702

### DIFF
--- a/doc/ipython-notebooks/link_checker.py
+++ b/doc/ipython-notebooks/link_checker.py
@@ -44,11 +44,9 @@ if __name__ == '__main__':
     check_python()
     # enter name of file
     file_name = sys.argv[1]
-
     links = link_finder(file_name)
-    number_of_links = len(links)
     
-    for i in range(number_of_links):
-        return_code = link_checker(links[i])
+    for link in links:
+        return_code = link_checker(link)
         if(return_code != 200):
-            print(links[i]) 
+            print(link) 
--- a/doc/ipython-notebooks/link_checker.py
+++ b/doc/ipython-notebooks/link_checker.py
@@ -2,8 +2,6 @@ import json
 import re
 import subprocess
 import sys
-
-
 #requirement is python >= 3.5
 
 def Find(cell_contents):
@@ -12,7 +10,6 @@ def Find(cell_contents):
     return urls   
 
 def link_finder(file_name):
-    
     links = []
     f = open(file_name)
     s = f.read()
@@ -31,33 +28,14 @@ def link_checker(link_address):
     stdout, stderr = process.communicate()
     return (int(stdout))
 
-
-
-
-
 if __name__ == '__main__':
-
     # enter name of file
     file_name = sys.argv[1]
 
-
     links = link_finder(file_name)
-
-    
     number_of_links = len(links)
     
     for i in range(number_of_links):
         return_code = link_checker(links[i])
         if(return_code != 200):
-            print(links[i])
-
-
-    
-
-    
-
-   
-
-    
-          
-    
+            print(links[i]) 
--- a/doc/ipython-notebooks/link_checker.py
+++ b/doc/ipython-notebooks/link_checker.py
@@ -1,0 +1,63 @@
+import json
+import re
+import subprocess
+import sys
+
+
+#requirement is python >= 3.5
+
+def Find(cell_contents):
+    text = cell_contents[0]
+    urls = re.findall('http[s]?://(?:[a-zA-Z]|[0-9]|[$-_@.&+]|[!*\(\),]|(?:%[0-9a-fA-F][0-9a-fA-F]))+', text)
+    return urls   
+
+def link_finder(file_name):
+    
+    links = []
+    f = open(file_name)
+    s = f.read()
+    input_file = json.loads(s)
+    #convert given file to python dictionary fromat
+    cells_count = len(input_file['cells'])
+    for i in range(cells_count):
+        #Assuming all links are in markdown cells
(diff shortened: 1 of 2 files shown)


doc/ipython-notebooks/link_checker.py
```diff
@@ -0,0 +1,63 @@
...
+    cells_count = len(input_file['cells'])
+    for i in range(cells_count):
+        #Assuming all links are in markdown cells
+        if(input_file['cells'][i]['cell_type'] == 'markdown'):
+            urls_in_current_cell = Find(input_file['cells'][i]['source'])
+            links.extend(urls_in_current_cell)
+    return links       
+
+def link_checker(link_address):
+    process = subprocess.Popen(['curl','-s', '-o','/dev/null','-I','-w','%{http_code}',link_address],stdout=subprocess.PIPE, stderr=subprocess.PIPE,universal_newlines=True)
```
gf712: rather that CURL can you not use python's `request` package and check the return code?

doc/ipython-notebooks/link_checker.py
```diff
@@ -0,0 +1,63 @@
...
+
+    
+
+    
+
+   
+
+    
+          
+
```
gf712: I guess the white space is not required or?

doc/ipython-notebooks/link_checker.py
bhavukkalra: I actuallt taklked about it with @vigsterkr and he guided me to use curl
Could definitely do it with python requests package.
Will get on it

doc/ipython-notebooks/link_checker.py
bhavukkalra: Will remove them

doc/ipython-notebooks/link_checker.py
gf712: ah ok, well let's see what @vigsterkr thinks. It's just a suggestion :)

bhavukkalra: could you please confirm?

vigsterkr: curl was about the idea when it was a shellscript... in case of python i dont see the reason to call out with a subprocess and call curl. as @gf712 use request package or any basic python functionality that fetches an url... even urllib (which is by default part of python) could work.

bhavukkalra: OH that makes sense. My bad. Will get on it right now.Altering the link_checker function :)

doc/ipython-notebooks/link_checker.py
```diff
@@ -0,0 +1,41 @@
+import json
+import re
+import subprocess
+import sys
+#requirement is python >= 3.5
```
gf712, line 5: i think we only check for python3 in cmake, so if you require 3.5 then you should add a version check here and raise an error.

gf712:
@bhavukkalra  @vigsterkr so how will this script be called? Is this part of a cmake target?

doc/ipython-notebooks/link_checker.py
bhavukkalra: Will add that as well in the latest commit. Requests module does require python>=3.5
Will make a function for checking python version requirement.

gf712: I mean you can just use [urllib](https://docs.python.org/3/library/urllib.request.html#module-urllib.request) which is just part of python.

bhavukkalra: Oh. sorry for this.Didn't knew about urllib. That makes things much more easier.

bhavukkalra:
Quick question:(Checking python version)
should the program just raise an error exception with message that:python version >= 3.5 needed
or
just print the error message i.e python version >= 3.5 needed and exit the program and raise no exceptions?
what should be done here? rest of the functionality is implemented

gf712:
> Quick question:(Checking python version)
> should the program just raise an error exception with message that:python version >= 3.5 needed
> or
> just print the error message i.e python version >= 3.5 needed and exit the program and raise no exceptions?
> what should be done here? rest of the functionality is implemented

just raise an exception

bhavukkalra:
@vigsterkr @gf712 could you please review the changes in the latest commit and suggest if there is any modification needed?

doc/ipython-notebooks/link_checker.py
```diff
@@ -0,0 +1,54 @@
...
+
+if __name__ == '__main__':
+    check_python()
+    # enter name of file
+    file_name = sys.argv[1]
+
+    links = link_finder(file_name)
+    number_of_links = len(links)
+    
+    for i in range(number_of_links):
```
gf712: Can this not just be for link in links? No need for indexing or?

gf712 (commented):
Have you tested this locally?

doc/ipython-notebooks/link_checker.py
bhavukkalra: Yes, I have tested this locally.It gives the following output for classification.ipynb 
under the directory shogun/doc/ipython-notebooks/classification
(it is working)
There is no need of indexing.I am removing them
Screenshot 2020-03-14 at 2 14 27 PM
 :).

bhavukkalra:
removed indexing.could you please review the changes @gf712 @vigsterkr and suggest if any further improvements can be made or how can we increase the code readability.

PR #4922, opened 2020-03-13
Edited some phrases

nouranali:


doc/readme/ABOUT.md
```diff
@@ -6,8 +6,8 @@ Shogun is an open-source machine learning library that offers a wide range of ef
 
 #### Shogun is state-of-the-art
 
-* Efficient implementation (from standard to cutting edge algorithms), modern software architecture in C++.
-* Easy combination of multiple data representations, algorithm classes, and general-purpose tools for rapid prototyping of data pipelines.
+* An efficient implementation (from standard to cutting edge algorithms), modern software architecture in C++.
```
karlnapf, line 9: I don't really agree on those changes tbh

nouranali: Well I'm trying, so if you have a better idea you could suggest it

karlnapf: My suggestion is to leave the text as it is, the change does not fix anything or make it better to read/understand. If you spot actual typos/dead-links, these are very welcomed changes. Or if you are interested in improving the language of public facing bits of Shogun, have a look at in-code the class documentations (doxygen). Many classes have very bad docstrings. In contrast to these readmes that you are touching here, which are already quite tuned/edited, the docstrings have received very little love over time. For example, something that would immensely helpful would be to unify all such api documentation, or at least fix the really bad ones (of which there are many many)

karlnapf: And sorry if my previous comment sounded harsh, it wasnt meant this way :)

nouranali: actually it's quite helpful, kinda harsh but helpful 
thanks a lot I'll try to work on those docstrings

PR #4929, opened 2020-03-14
reordered a phrase

nouranali:


doc/OpenCV_docs/eigenfaces.cpp
```diff
@@ -144,7 +144,7 @@ int main()
     //
     // So in effect we just reduced the dimensions of each of our training images
     // by 12 times.
-    // Just this is the work of PCA in Eigenfaces face recognition algorithm.
+    // This is just the work of PCA in Eigenfaces face recognition algorithm.
```
karlnapf, line 147: what about we remove this sentence .. it doesnt really make sense anyways ;)

BTW there is not really a need to correct in-code comments that is not API documentation...it is way down on the priority list of things to improve

PR #4934, opened 2020-03-16
Functional style graph building

gf712:
This uses Jax style implementation to evaluate expressions.
```cpp
auto input1 = make_shared<Array>(X1);
auto input2 = make_shared<Array>(X2);
auto output = input1 + input2; // output is not evaluated, just has the information needed for computation
std::cout << output << '\n'; // now output is evaluated -> creates graph using default backend
```

### DIFF
--- a/src/shogun/base/ShogunEnv.h
+++ b/src/shogun/base/ShogunEnv.h
@@ -106,6 +106,17 @@ namespace shogun
 			return sg_graph_backend;
 		}
 
+		/** set the global graph backend
+		 *
+		 * @return the backend enum
+		 */
+		void set_graph_backend(GRAPH_BACKEND backend)
+		{
+			// if (!shogun::graph::available_backends().count(backend))
+				sg_graph_backend = backend;
+			// else
+			// 	error("Backend not available.");
+		}
 
 	private:
 		/** Default constructor
--- a/src/shogun/mathematics/graph/Array.h
+++ b/src/shogun/mathematics/graph/Array.h
@@ -1,0 +1,195 @@
+/*
+ * This software is distributed under BSD 3-clause license (see LICENSE file).
+ *
+ * Authors: Gil Hoben
+ */
+
+#ifndef SHOGUN_GRAPH_ARRAY_H_
+#define SHOGUN_GRAPH_ARRAY_H_
+
+#include <shogun/mathematics/graph/Tensor.h>
+#include <shogun/mathematics/graph/Graph.h>
+#include <shogun/mathematics/graph/nodes/Add.h>
+#include <shogun/mathematics/graph/nodes/Input.h>
+#include <shogun/mathematics/graph/nodes/MatMul.h>
+
+namespace shogun
+{
+	namespace graph
+	{
+		class LazyExpr
+		{
+			std::unordered_map<std::shared_ptr<node::Input>, std::shared_ptr<Tensor>> m_inputs;
+			std::shared_ptr<node::Node> m_output;
+
+		public:
+			LazyExpr(const std::shared_ptr<Tensor>& input_tensor)
+			{
+				auto input = std::make_shared<node::Input>(input_tensor->get_shape(), input_tensor->get_type()->type());
+				m_inputs.emplace(input, input_tensor);
+				m_output = input;
+			}
+
+			template<typename T>
+			void bind_input(const T& expr)
+			{
+				if constexpr(std::is_same_v<T, std::shared_ptr<LazyExpr>>)
+				{
+					for (const auto& [input_node, input_tensor]: expr->get_inputs())
+					{
+						if(!m_inputs.count(input_node))
+						{
+							m_inputs.emplace(input_node, input_tensor);
+						}
+					}
+				}
+			}
+
+			/** bind rhs nodes to lhs (this)
+			  */
+			template<typename OperatorType, typename ...Args, std::enable_if_t<std::is_base_of_v<node::Node, OperatorType>>* = nullptr>
+			void bind(Args&&... args)
+			{
+				(bind_input(std::forward<Args>(args)), ...);
+
+				m_output = std::make_shared<OperatorType>(m_output, return_node(std::forward<Args>(args))...);
+			}
+
+			const std::unordered_map<std::shared_ptr<node::Input>, std::shared_ptr<Tensor>>& get_inputs() const 
+			{
+				return m_inputs;
+			}
+
+			const std::shared_ptr<node::Node> get_output() const
+			{
+				return m_output;
+			}
+
+		private:
+			template <typename T>
+			auto return_node(const T& expr)
+			{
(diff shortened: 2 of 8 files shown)


src/shogun/mathematics/graph/Array.h
```diff
@@ -0,0 +1,237 @@
...
+			template <
+			    typename OperatorType, typename... Args,
+			    std::enable_if_t<std::is_base_of_v<node::Node, OperatorType>>* =
+			        nullptr>
+			void bind(Args&&... args)
+			{
+				(bind_input(std::forward<Args>(args)), ...);
+
+				m_output = std::make_shared<OperatorType>(
+				    m_output, return_node(std::forward<Args>(args))...);
```
gf712, line 79: @vigsterkr using all the C++17 features

src/shogun/mathematics/graph/Array_test.cc
```diff
@@ -0,0 +1,49 @@
...
+	auto expected_result2 = expected_result1 + X2;
+
+	auto input1 = make_shared<Array>(X1);
+	auto input2 = make_shared<Array>(X2);
+	auto intermediate = input1 + input1;
+	auto output = intermediate + input2;
+
+	auto result1 =
+	    intermediate->evaluate()->template as<SGVector<NumericType>>();
+	auto result2 = output->evaluate()->template as<SGVector<NumericType>>();
```
gf712: @karlnapf almost like JAX :) it's super inefficient right now, but it works

gf712: if we expose this to dynamically typed interfaces this would be hidden, i.e. `numpy_array = output.to_numpy()`

src/shogun/mathematics/graph/Array.h
```diff
@@ -0,0 +1,237 @@
...
+		std::shared_ptr<Array> matmul(
+		    const std::shared_ptr<Array>& lhs,
+		    const std::shared_ptr<Array>& rhs, const bool transpose_a,
+		    const bool transpose_b)
+		{
+			auto expr = lhs->get_lazy_expression()->copy();
+			expr->bind<node::MatMul>(
+			    rhs->get_lazy_expression(), transpose_a, transpose_b);
+			return std::make_shared<Array>(std::move(expr));
+		}
```
gf712, line 237: @vigsterkr is it worth coming up with a way to automatically expose nodes via this interface? I am thinking we could use ctags? Or it can just be done manually. This will have to be defined in a .cpp too, otherwise won't work with interfaces

karlnapf (commented):
Really cool!

So as long as I apply functional updates to the expressions, it doesnt evaluate stuff right?
What about control flow things, like if-else? I guess for that an evaluation is necessary?

In jax, one can really write big models/algorithms and only compile them down to jit at the very last step.

What about adding some things like an openmp parallel map? Is that hard?

karlnapf:
I guess it would be cool to express a few simple algorithms in this and take it from there:

* Linear/kernel regression (using cholesky)
* PCA (using SVD of Eigendecomposition)
* Multi layer dense neural network + loss + SGD
* KMeans

These would cover quite a substantial amount of the operations needed.

gf712:
> Really cool!
> 
> So as long as I apply functional updates to the expressions, it doesnt evaluate stuff right?

Indeed, everything is stored in LazyExpr, like in jax

> What about control flow things, like if-else? I guess for that an evaluation is necessary?
> In jax, one can really write big models/algorithms and only compile them down to jit at the very last step.

yes, if the value is required the graph has to be evaluated. We could think about how to avoid that, but it would be really difficult in C++ because of the lack of support for reflection. The only way I see it working is using llvm/clang frontend, but then it becomes compiler specific. We would have to do it like in tf where they have some utilities for control flow.
Btw this is not essential, we just need to cache the graph properly and the overhead can be massively reduced when something has to be evaluated.

> What about adding some things like an openmp parallel map? Is that hard?

you mean at the user facing API level? It wouldn't be required. The graph optimiser will detect parallelisation opportunities. You mean like jax's vmap right?

gf712:
> I guess it would be cool to express a few simple algorithms in this and take it from there:
> 
>     * Linear/kernel regression (using cholesky)

Should be doable now (after writing the op impl)

>     * PCA (using SVD of Eigendecomposition)

This is doable now (after writing the op impl)
 
>     * Multi layer dense neural network + loss + SGD

Requires autodiff implementation. I have to think how this will work. Probably the same way ngraph and stan do it with reverse autodiff?

>     * KMeans

This would be quite easy, so could be a good first example. Could write it as a jupyter notebook with C++ kernel

> 
> These would cover quite a substantial amount of the operations needed.

src/shogun/base/ShogunEnv.h
```diff
@@ -106,6 +106,17 @@ namespace shogun
...
+		 *
+		 * @return the backend enum
+		 */
+		void set_graph_backend(GRAPH_BACKEND backend)
+		{
+			// if (!shogun::graph::available_backends().count(backend))
+				sg_graph_backend = backend;
+			// else
+			// 	error("Backend not available.");
+		}
```
gf712, line 119: @vigsterkr the problem with graph backend setter is that we need to add something like `shogun::graph::available_backends()` to ShogunEnv

src/shogun/mathematics/graph/Array.h
```diff
@@ -0,0 +1,241 @@
...
+			{
+			}
+
+			template <typename T>
+			Array(const T& input) : m_output_tensor(std::make_shared<Tensor>(input))
+			{
+				m_lazy_expr = std::make_unique<LazyExpr>(m_output_tensor);
+			}
+
+			template <typename T, std::enable_if_t<std::is_rvalue_reference_v<T&&>>* = nullptr>
```
gf712, line 143: @vigsterkr does this look correct to you? Without the SFINAE `T&&` just matches everything

gf712:
@karlnapf In jax if you have something like
```python
y_ = np.dot(X, w)
y = np.dot(X, w) + b
```
Would you expect `y` being evaluated reusing `y_` or is it just recomputed?

karlnapf:
re-used if both are part of the same graph evaluation.... Although wouldnt this something that jit takes care of? LIke the hashing idea, but it is kinda doing stuff that maybe is better left to optimizers?

karlnapf:
In terms of autodiff, I think it would be good if there was `grad` that does backward mode to start with, but later on we probably also need forward/backward jacobian matrix and jacobian vector products

gf712:
> re-used if both are part of the same graph evaluation.... Although wouldnt this something that jit takes care of? LIke the hashing idea, but it is kinda doing stuff that maybe is better left to optimizers?

In this case it's a tricky one because I would imagine the optimiser sees the second one as a single gemm call. The issue later on will be how to fragment the graph, when to optimise... That's why I was wondering what you would expect jax todo in this situation

gf712:
Actually we might be able to integrate control flow by overloading the bool cast operator of Array

PR #4935, opened 2020-03-16
Updated KDE.ipynb

jonpsy:
@gf712, Updated KDE.ipynb , runs smoothly on conda package of shogun.

### DIFF
--- a/doc/ipython-notebooks/distributions/KernelDensity.ipynb
+++ b/doc/ipython-notebooks/distributions/KernelDensity.ipynb
@@ -5,28 +5,36 @@
    "metadata": {},
    "source": [
     "# Kernel Density Estimation"
-   ]
+   ],
+   "execution_count": null,
+   "outputs": []
   },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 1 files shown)


gf712:
Oh, this is the opposite of what has to be done. You need to fix ipython notebooks to use the new API, this is reverting to the old API. To do this don't use conda, you need to build shogun from source using the develop branch and the run the notebooks with this!

jonpsy:
Off to a good start!. So I should refactor the code such that, it works on the environment built from source?

gf712:
> Off to a good start!. So I should refactor the code such that, it works on the environment built from source?

Haha, don't worry, I wasn't very clear. Yes, that's it. We try to keep the notebooks up to date when we refactor the API, but there might be some that aren't working!

jonpsy:
Okay this should work when this #4951 is works for all platform

gf712:
OK, but now nothing has changed in this notebook right?

jonpsy:
Oh, yes! That's true!. Then this PR is for naught, it's all good.

Hmm...Shall I close it??

gf712:
OK, closing this in that case

PR #4937, opened 2020-03-16
Graph hash

gf712:
Hashes required:
- [x] Abstract node -> may contain runtime values
- [ ] Runtime node -> knows static shape and input. The graph can use this to find out if it requires recomputation of the graph
- [ ] Tensor hash with fingerprint (might be handled in PR by @vigsterkr)

The hash of the nodes is a combination of all hashes of all previous nodes. This might be too expensive for very simple graphs?

### DIFF
--- a/src/shogun/mathematics/graph/Graph_test.cc
+++ b/src/shogun/mathematics/graph/Graph_test.cc
@@ -19,6 +19,44 @@ using namespace shogun;
 using namespace shogun::graph;
 using namespace std;
 
+TYPED_TEST(GraphTest, abtract_graph_hash)
+{
+	using NumericType = typename TypeParam::c_type;
+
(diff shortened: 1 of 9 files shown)


src/shogun/mathematics/graph/Graph_test.cc
```diff
@@ -19,6 +19,44 @@ using namespace shogun;
...
+	    vector<shared_ptr<node::Node>>{output});
+
+	for (auto&& backend : this->m_backends)
+	{
+		graph1->build(backend);
+		graph2->build(backend);
+		graph3->build(backend);
+
+		EXPECT_EQ(*graph1, *graph2);
+		EXPECT_NE(*graph1, *graph3);
```
gf712, line 56: @vigsterkr is this what you would expect?

Issue #3000, opened 2016-02-22
Replace an undocumented example by a cookbook page


karlnapf:
Add a single page in the cookbook, replacing a python/octave/ ... example.

The patch should contain 3 changes
- added .rst file
- added .sg file
- remove .py, .m, ... files

See https://github.com/shogun-toolbox/docs/blob/master/EXAMPLES.md and https://github.com/shogun-toolbox/docs/blob/master/DEVELOPING.md for instructions.

NOTE: There are many many of those, so this needs many people solving small pieces

tbskm:
Ok, I've given it a shot for the Gaussian Naive Bayes example. I'm new to Shogun (and ML in general), so I would really appreciate it if someone could give it a quick look and let me know if everything looks ok.

https://github.com/tbskm/gaussianNaiveBayes-cookbook

karlnapf:
Hi there,
great, these are exactly the two files we need. Please send a Pull request against the cookbook feature branch (link above) rather than posting the files here. I will review it in there then.

A note for the others -- there are many more examples to go

sanuj:
@karlnapf I tried to build the cookbook in `doc/cookbook/build` by doing `cmake ..` and `make cookbook`. I tried the same in shogun's build dir also. [Here is the error gist.](https://gist.github.com/81ad03344c27a3d7c048.git)
`cmake -DBUILD_META_EXAMPLES ..` is also not working. What am I missing?

karlnapf:
This is just the link checker failing, are you online?
Also pull the latest version, it works on my machine

tbskm:
Hmm, pretty sure I would have been online, but I'm using a stock Broadcom card that frequently acts up with Linux, so my connection probably went down at some point.
Btw I've been busy with some school assignments, but I should be able to remove those unnecessary bits from the cookbook page either today or tomorrow. 
Thanks!



    _____________________________

sanuj:
@karlnapf It works now after pulling the latest version. Although `cmake -DBUILD_META_EXAMPLES ..` is still not working. The cookbook looks pretty neat.

karlnapf:
@tbskm I like the gaussian naive bayes example! Will merge soon.

karlnapf:
@sanuj lets not discuss this here ... irc maybe?

elenanst:
Hello, I'm working on replacing "converter_multidimensionalscaling_modular.py" with a meta example and a cookbook page.

[...]

PR #4942, opened 2020-03-18
Add meta example for Sparse Euclidian Distance #3000

Hephaestus12:
Changes to be committed:
	new file:   examples/meta/src/distance/**sparseeuclidean.sg**
	deleted:    examples/undocumented/python/**distance_sparseeuclidean.py**


examples/meta/src/distance/sparseeuclidean.sg
```diff
@@ -0,0 +1,20 @@
+CSVFile f_feats_a("../../data/fm_train_sparsereal.dat")
```
vigsterkr, line 1: wrong file name.... should be `sg.in`

vigsterkr (changes requested):

examples/meta/src/distance/sparseeuclidean.sg
gf712: also use `SHOGUN_DATA` instead of relative path

Hephaestus12: I don't understand.
This was the filename in the existing python example as well.
Also, I don't see any `sg.in` file in the `data` directory.

vigsterkr: it should be `examples/meta/src/distance/sparseeuclidean.sg.in` you'll see that all other examples have that extension....

and plz check the other examples for how the data file path is being defined.

vigsterkr: and as a general rule it'd be great if you could test (compile and run) any patch *before* you create a pull request

karlnapf: @Hephaestus12 just add make it more clear, @vigsterkr was not talking about the filename in the csv file, but about the filename of the meta example listing itself

Issue #4844, opened 2020-02-26
Automatic copy of the selected code snippet to clipboard


vigsterkr:
there are several code snippets at https://www.shogun.ml/examples/latest/index.html and it would be great that if a codesnippet is selected it would be automagically copied to the user's clipboard.

this is basically a jquery hack

krutikabapat:
@vigsterkr Would it be fine If we add a copy button near every code snippet and If the user clicks on it, the code snippet is copied?

vigsterkr:
@krutikabapat i mean that's necessary for sure, but there are way to solve this with JS that the selection gets copied on your clipboard....

krutikabapat:
Okay, got it :)

boladjivinny:
If this is still open, let me work on it

vigsterkr:
it is still open

boladjivinny:
@vigsterkr Can you direct me on where to find the related source code for the website?

vigsterkr:
https://github.com/shogun-toolbox/shogun/tree/develop/doc/cookbook

chinmay81098:
Hey, I have created a PR for this issue please review.

rajpratyush:
is this issue still open for work?
@vigsterkr

Issue #4020, opened 2017-12-16
"View full listing" for meta examples


karlnapf:
Every meta example on the website should have a tab that says "show full code listing", which allows a user to see all of the example code, not just the snippets

Requires some web-hacking skills... not THAT easy, but not impossible

Start from the sphinx plugin:

For that, you will have to modify the sphinx plugin we wrote to render the meta examples, see https://github.com/shogun-toolbox/shogun/blob/develop/doc/cookbook/extensions/sgexample.py

novate:
I would like to do this in my GSoC project, here are my initial ideas:
1. Enclose the unclosured initial meta example `.sg` code with `#![initial_code]`. This work could be done automatically with python.
2. Modify all `.rst` files. Add `.. sgexample:: something.sg:initial_code`. This work could also be done automatically with python.
3. To show full code, we could add `. sgexample:: something.sg` at the end of each `.rst` files, and modify `sgexample.py` to make the full code hide initially.

By doing procedures 1 and 2, we can make all cookbook with full content of example code. By doing procedure 3, we can add a “show full list” feature to our cookbook.

I'm currently learning `sphinx`, so I may have a more detailed idea on how to modify `sgexample.py`.

karlnapf:
Hi thanks for taking this up!

1. is not really necessary. We do explicitly extract code in `sgexample.py` so we can just extract all there.
2. modifying the `.rst` files is not really a good idea since it requires so many changes. Rather we want to change how the page is generated in the python script mentioned above
3. same as above

I think the best way to change things is to start playing with the web page template that is used and then take it from there

karlnapf:
Hey, any updates or questions here?

novate:
Yes, I do have some updates:
1. I agree that we'd better modify `.rst` file as few as we could, while I think that the `.rst` file may still be modified, for we want to show full code, however `.rst` files haven't contain all of them yet. Such a modification could be easy if we write a automated snippet. We could discuss on how to make the modification less and easy to be done.
2. I agree that the crucial part is to modify sphinx-related code, and I have checked some of the pages generated by `sphinx`:
[...]

PR #4946, opened 2020-03-19
Fix issue #4844

chinmay81098:
Automatic copy of the selected code snippet to clipboard feature added

### DIFF
--- a/doc/cookbook/source/templates/page.html
+++ b/doc/cookbook/source/templates/page.html
@@ -21,11 +21,23 @@
 jQuery(document).ready(function ($) {
     $('ul[id^=tabs]').tab();
     $('.highlight').mouseup(function() {
-      var copyText = window.getSelection().toString();
-      document.execCommand("copy");
+	    var timeout;
+	    var copyText = document.getSelection().toString();
       if(copyText){
-        alert("Copied!");
-      }
+        clearTimeout(timeout);
+        document.execCommand("copy");
+        $(this).tooltip("show");
+	    }    
+      var selected = this;
+      timeout = setTimeout(function() {
+      $(selected).tooltip("hide");
+      }, 1000);
+    });
+
+    $('.highlight').tooltip({
+	    trigger:'manual',
+      placement:'bottom',
+      title:'Copied!',
     });
 });
 </script>
--- a/doc/cookbook/source/templates/page.html
+++ b/doc/cookbook/source/templates/page.html
@@ -23,6 +23,9 @@ jQuery(document).ready(function ($) {
     $('.highlight').mouseup(function() {
       var copyText = window.getSelection().toString();
       document.execCommand("copy");
+      if(copyText){
+        alert("Copied!");
+      }
     });
 });
 </script>
--- a/doc/cookbook/source/templates/page.html
+++ b/doc/cookbook/source/templates/page.html
@@ -20,6 +20,10 @@
 <script>
 jQuery(document).ready(function ($) {
     $('ul[id^=tabs]').tab();
+    $('.highlight').mouseup(function() {
+      var copyText = window.getSelection().toString();
+      document.execCommand("copy");
+    });
 });
 </script>
 


vigsterkr:
yey works!
https://www.shogun.ml/cookbook_pr/99d6ad3b3e5f35805431f38570ba4c3d6aaa14a6/html/examples/binary/linear_support_vector_machine.html

geektoni:
Shouldn't we add some sort of "Copied!" notification to show the user what happened? :/

gf712:
> Shouldn't we add some sort of "Copied!" notification to show the user what happened? :/

I think so too! It took me a while to understand what was happening :D But thank you for adding this feature already!

chinmay81098:
> Shouldn't we add some sort of "Copied!" notification to show the user what happened? :/

Yes, thanks for the feedback. Its added now.

gf712:
@chinmay81098 how difficult would it to add something like [this](https://i.stack.imgur.com/3GBr2.png) rather than an alert?

chinmay81098:
> @chinmay81098 how difficult would it to add something like [this](https://i.stack.imgur.com/3GBr2.png) rather than an alert?

A bit, but I have tried to implement something similar. Kindly review.

geektoni:
This makes it more clear @chinmay81098 ! However, I am still thinking that maybe we would need something more "visible" since the user may not know that by selecting the code it will be also copied. 

Could it be possible to add some kind of button on the top right corner of the code snippets that copies everything in the snippet inside the clipboard (like [this](https://user-images.githubusercontent.com/3296552/77080192-4b573100-69f9-11ea-9ded-3a9e6f717d90.png))?

gf712:
> This makes it more clear @chinmay81098 ! However, I am still thinking that maybe we would need something more "visible" since the user may not know that by selecting the code it will be also copied.
> 
> Could it be possible to add some kind of button on the top right corner of the code snippets that copies everything in the snippet inside the clipboard (like [this](https://user-images.githubusercontent.com/3296552/77080192-4b573100-69f9-11ea-9ded-3a9e6f717d90.png))?

I think what @chinmay81098 seems to be pretty good already. That would be even nicer though. Do you think we could have both or would it be too much?

doc/cookbook/source/templates/page.html
```diff
@@ -20,6 +20,25 @@
...
+	    }    
+      var selected = this;
+      timeout = setTimeout(function() {
+      $(selected).tooltip("hide");
+      }, 1000);
+    });
+
+    $('.highlight').tooltip({
+	    trigger:'manual',
+      placement:'bottom',
```
gf712, line 39: can the tooltip be placed relative to the cursor? I am thinking it is a bit arbitrary that it is in the middle

karlnapf:
In the link above, how to make it work? :D sorry I am confused

vigsterkr:
https://www.shogun.ml/cookbook_pr/9546d036795c57e7d1fdf09dc13a14762df970fa/html/examples/binary/linear_support_vector_machine.html

commit hash

vigsterkr:
to make it work, just select text in code section

karlnapf:
Ah yes!

One thing I would like to suggest is also to have a button to view/copy the *whole* code listing (as in, including the bits that are not rendered)

Since we are touching this @chinmay81098 you wanna go ahead with that?

vigsterkr:
@karlnapf could we have that in another pr? :) that's quite some more hacking :)

karlnapf:
yes of course sorry :)
I think there is an issue for that already (although it might have been closed as stale)

karlnapf:
It is #4020

doc/cookbook/source/templates/page.html
chinmay81098: Yes, but for that I need to wrap the text in one single element and then add tooltip to it,right now each letter has its own tag that's why I had to target the complete box. Do I need to look into sgexample.py for this?

chinmay81098:
> > This makes it more clear @chinmay81098 ! However, I am still thinking that maybe we would need something more "visible" since the user may not know that by selecting the code it will be also copied.
> > Could it be possible to add some kind of button on the top right corner of the code snippets that copies everything in the snippet inside the clipboard (like [this](https://user-images.githubusercontent.com/3296552/77080192-4b573100-69f9-11ea-9ded-3a9e6f717d90.png))?
> 
> I think what @chinmay81098 seems to be pretty good already. That would be even nicer though. Do you think we could have both or would it be too much?

Both will be nice to have, but @karlnapf mentioned that there is already a pr for adding a button so I'll just follow that issue there and contribute if required.

Issue #4890, opened 2020-03-07
Error while compiling C++ files using local spdlog


krutikabapat:
Compilation errors while compiling C++ examples in ```shogun/src/shogun/any c++example file``` while using local installation of spdlog.  
The following errors are displayed:
```
/usr/local/include/shogun/io/SGIO.h:286:44: error: expected unqualified-id before ‘&’ token
        const fmt::experimental::string_view& msg) const;
                                            ^
/usr/local/include/shogun/io/SGIO.h:286:44: error: expected ‘)’ before ‘&’ token
/usr/local/include/shogun/io/SGIO.h:286:44: error: expected ‘;’ at end of member declaration
/usr/local/include/shogun/io/SGIO.h:286:46: error: ‘msg’ does not name a type
        const fmt::experimental::string_view& msg) const;
```

gf712:
Which version of spdlog did you use? Is this from the spdlog master branch?

krutikabapat:
@gf712 Yes it is from the spdlog master branch. I am using spdlog version 1.5.0

gf712:
@krutikabapat did it work with the 1.5.0 version then?

krutikabapat:
@gf712 I tried it with 1.5.0 only but It did not work. :'(

gf712:
@krutikabapat I seem to be able to compile it with a local 1.5.0 version. What error do you get with 1.5.0? If it is a linking issue, you need recompile your local spdlog with `-fPIC` (`cmake -DCMAKE_POSITION_INDEPENDENT_CODE=ON`)

PR #4948, opened 2020-03-19
Tried to correct some syntatical errors with issue #4890

sanjibansg:
Do inform about the correctness and any further development if required.


gf712 (commented):
Hmm this is just adding a syntax error... You should have a look at C++ class member function declarations!

gf712:
Also try to compile locally before sending any future patches please :)

PR #4949, opened 2020-03-20
fixes indentation(commands)

bhavukkalra:
fixes commands in R and Ruby interfaces not properly highlighted.Increases readability


karlnapf:
thanks!

PR #4950, opened 2020-03-20
GraphEnv

gf712:
Shogun depends on graph, not the other way around


src/shogun/mathematics/graph/Graph.cpp
```diff
@@ -22,12 +22,15 @@ Graph::Graph(
...
-	auto* env = ShogunEnv::instance();
-	build(env->graph_backend());
+	auto* env = GraphEnv::instance();
+	build(env->get_backend());
 }
 
 void Graph::build(GRAPH_BACKEND backend)
 {
+	auto* env = GraphEnv::instance();
+	std::lock_guard<std::mutex> lock(env->m_env_mutex);
```
gf712: @vigsterkr does this make sense? Basically lock the mutex here so that you can't change the backend in another thread whilst the graph is building

gf712:
@vigsterkr then the env would have some sort of device mapping, e.g. {"CPU1": ArchDetails} where ArchDetails would have information about things like SIMD instructions available

src/shogun/mathematics/graph/GraphEnv.cpp
```diff
@@ -0,0 +1,113 @@
...
+{
+	return m_backend;
+}
+
+void GraphEnv::set_backend(const GRAPH_BACKEND backend)
+{
+	if (available_backends().count(backend))
+	{
+		// mutex is locked here to avoid having graphs with mixed backend ops
+		std::lock_guard<std::mutex> lock(m_env_mutex);
```
gf712: @vigsterkr this is the other place where the mutex is locked

src/shogun/mathematics/graph/GraphExecutor.cpp
```diff
@@ -5,101 +5,19 @@
...
 #include <shogun/mathematics/graph/GraphExecutor.h>
-
-#include <mutex>
-#include <regex>
+#include "GraphEnv.h"
 
 using namespace shogun::graph;
 using namespace shogun::io;
 
-const ExecutorFactory& shogun::graph::backend_list()
```
gf712, line 15: also thought all of this should be part of GraphEnv rather than free functions

src/shogun/mathematics/graph/Graph.cpp
vigsterkr: yep!

vigsterkr (changes requested):
export export export....!
use SHOGUN_ENGINE_EXPORT macro! ;)

src/shogun/mathematics/graph/Graph.cpp
gf712: the only thing is that in the code above `build(env->get_backend());` in `Graph::build()` is not locked, so there is a chance that the backend changes there.. right?

gf712:
> export export export....!
> use SHOGUN_ENGINE_EXPORT macro! ;)

You mean for `GraphEnv`?

src/shogun/mathematics/graph/GraphEnv.cpp
vigsterkr: how about using shared/unique locks i.e. shared_mutex? https://en.cppreference.com/w/cpp/thread/shared_mutex

vigsterkr:
> > export export export....!
> > use SHOGUN_ENGINE_EXPORT macro! ;)
> 
> You mean for `GraphEnv`?

yep

vigsterkr:
see all the unresolved symbol errors: https://dev.azure.com/shogunml/shogun/_build/results?buildId=2950&view=logs&jobId=9ffa40d6-0008-5f71-b4ad-2956145303e6&j=9ffa40d6-0008-5f71-b4ad-2956145303e6&t=4c6a8d5c-d9f7-5d77-51f8-d43faa094742

src/shogun/mathematics/graph/GraphEnv.cpp
gf712: ah good point, I had in my mind this was C++20!

src/shogun/mathematics/graph/Graph.cpp
```diff
@@ -22,12 +22,15 @@ Graph::Graph(
...
-	auto* env = ShogunEnv::instance();
-	build(env->graph_backend());
+	auto* env = GraphEnv::instance();
+	build(env->get_backend());
 }
 
 void Graph::build(GRAPH_BACKEND backend)
 {
+	auto* env = GraphEnv::instance();
+	std::unique_lock lock(env->m_env_mutex);
```
gf712, line 32: @vigsterkr should this be shared?

PR #4951, opened 2020-03-20
Accept Feature arg

jonpsy:
Fix so that get_log_density() works even after passing Feature.  @vigsterkr , comments?

### DIFF
--- a/src/shogun/distributions/KernelDensity.cpp
+++ b/src/shogun/distributions/KernelDensity.cpp
@@ -94,9 +94,9 @@ bool KernelDensity::train(std::shared_ptr<Features> data)
 SGVector<float64_t> KernelDensity::get_log_density(const std::shared_ptr<Features>& test, int32_t leaf_size)
 {
 	require(test,"data not supplied");
-	std::shared_ptr<DenseFeatures<float64_t>> dense_feat =std::dynamic_pointer_cast<DenseFeatures<float64_t>>(test);
+	auto dense_feat =std::dynamic_pointer_cast<DenseFeatures<float64_t>>(test);
 	require(dense_feat,"Expected DenseFeatures<float64_t> type");
-	
+
 	if ((m_eval==EM_KDTREE_SINGLE) || (m_eval==EM_BALLTREE_SINGLE))
 		return tree->log_kernel_density(dense_feat->get_feature_matrix(),m_kernel_type,m_bandwidth,m_atol,m_rtol);
 
--- a/src/shogun/distributions/KernelDensity.cpp
+++ b/src/shogun/distributions/KernelDensity.cpp
@@ -94,8 +94,8 @@ bool KernelDensity::train(std::shared_ptr<Features> data)
 SGVector<float64_t> KernelDensity::get_log_density(const std::shared_ptr<Features>& test, int32_t leaf_size)
 {
 	require(test,"data not supplied");
-	require(test->get_feature_type == 10 && test->get_feature_class == 11,"Expecting DenseFeatures<float64_t>");
-	std::shared_ptr<DenseFeatures<float64_t>> dense_feat = std::static_pointer_cast<DenseFeatures<float64_t>>(test);
+	std::shared_ptr<DenseFeatures<float64_t>> dense_feat =std::dynamic_pointer_cast<DenseFeatures<float64_t>>(test);
+	require(dense_feat,"Expected DenseFeatures<float64_t> type");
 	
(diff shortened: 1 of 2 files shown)


src/shogun/distributions/KernelDensity.cpp
```diff
@@ -91,12 +91,14 @@ bool KernelDensity::train(std::shared_ptr<Features> data)
 	return true;
 }
 
-SGVector<float64_t> KernelDensity::get_log_density(const std::shared_ptr<DenseFeatures<float64_t>>& test, int32_t leaf_size)
+SGVector<float64_t> KernelDensity::get_log_density(const std::shared_ptr<Features>& test, int32_t leaf_size)
 {
 	require(test,"data not supplied");
-
+	require(test->get_feature_type == 10 && test->get_feature_class == 11,"Expecting DenseFeatures<float64_t>");
+	std::shared_ptr<DenseFeatures<float64_t>> dense_feat = std::static_pointer_cast<DenseFeatures<float64_t>>(test);
```
gf712: imo you should do a dynamic cast and check for nullptr, you shouldn't rely on `get_feature_type` that might get deprecated soon.

jonpsy: Aight I'll update that

jonpsy: I've used ```dynamic_pointer_cast``` as it's ```std::shared_ptr``` thoughts?

src/shogun/distributions/KernelDensity.cpp
```diff
@@ -91,12 +91,14 @@ bool KernelDensity::train(std::shared_ptr<Features> data)
 	return true;
 }
 
-SGVector<float64_t> KernelDensity::get_log_density(const std::shared_ptr<DenseFeatures<float64_t>>& test, int32_t leaf_size)
+SGVector<float64_t> KernelDensity::get_log_density(const std::shared_ptr<Features>& test, int32_t leaf_size)
 {
 	require(test,"data not supplied");
-
+	std::shared_ptr<DenseFeatures<float64_t>> dense_feat =std::dynamic_pointer_cast<DenseFeatures<float64_t>>(test);
```
gf712: `auto dense_feat = ...` you already show what the type is in the cast

jonpsy: True

src/shogun/distributions/KernelDensity.cpp
gf712: yes that's what you should use

src/shogun/distributions/KernelDensity.cpp
jonpsy: Done

src/shogun/distributions/KernelDensity.cpp
```diff
@@ -91,12 +91,14 @@ bool KernelDensity::train(std::shared_ptr<Features> data)
 	return true;
 }
 
-SGVector<float64_t> KernelDensity::get_log_density(const std::shared_ptr<DenseFeatures<float64_t>>& test, int32_t leaf_size)
+SGVector<float64_t> KernelDensity::get_log_density(const std::shared_ptr<Features>& test, int32_t leaf_size)
```
karlnapf, line 94: I don't really see the point of moving this check to runtime?

vigsterkr: @karlnapf unified/simplified API ;)

vigsterkr: features() returns Features... you wont be able to call this from SWIG API, as you dont have RealFeatures anymore...

karlnapf: sure thing. I didnt realise this is exposed to swig.

vigsterkr: well KDE notebook uses it :S

vigsterkr: i guess there could be a way to refactor this even more so the function is not exposed at all, but on the end somehow features should get to this object, i.e. it'll have to be general Features as well...

karlnapf: yeah, the public interface needs the base class for sure. I think this is good in the meantime and at some point we could/should refactor the distribution api....
Also with this, we can add a meta example for distributions!

gf712:
can you trigger the CI again to see if everything is working please?

PR #4952, opened 2020-03-21
cpu architecture identification

gf712:
identify available simd instructions at runtime. To be combined with GraphEnv


gf712:
Add gets dispatched properly on my machine which has avx2 instructions and uses 16 byte wide packets

gf712:
Merging this and working on other ops in future PR

Issue #4258, opened 2018-04-16
Make use of inheritance in class API docs


karlnapf:
There are many methods that are well documented in the base class (doxygen docs).
Then the subclass often doesnt provide any comments at all or, even worse, copy-pased from the baseclass.

But doxygen can in fact automatically use base class doc strings if subclasses do not provide one:

This depends on the INHERIT_DOCS configuration setting. When setting INHERIT_DOCS to YES the documentation will be copied, when setting it to NO the derived method will remain undocumented.

This task to to do through the method doc strings and try to
* use base class docs where possible (delete subclass docs and enable function in doxygen)
* if subclass docs are useful (something is different), write proper ones

j-petit:
I am looking into this.

nouranali:
I want to start in this issue, but i need more clarification, would you help

karlnapf:
Not sure what you mean, it is all described above?

PR #4955, opened 2020-03-23
Refactor RKS

jonpsy:
Tried to refactor some code using our linalg interface.I've deliberately left some parts as I've some doubts regarding it. I've done the benchmarks, here are the results

**Without changes**
wihtoutchange

**With changes**
withchanges

As we can see, the gain isn't substantial, more like +- 1 seconds.

**NOTE**: This PR needs rebase, the current build is breaking in my computer, so I've made PR w.r.t to the latest build which was working.

Thoughts? @karlnapf , @gf712 ?

### DIFF
--- a/src/shogun/features/RandomFourierDotFeatures.cpp
+++ b/src/shogun/features/RandomFourierDotFeatures.cpp
@@ -72,11 +72,6 @@ std::shared_ptr<Features> RandomFourierDotFeatures::duplicate() const
 	return std::make_shared<RandomFourierDotFeatures>(*this);
 }
 
-const char* RandomFourierDotFeatures::get_name() const
-{
-	return "RandomFourierDotFeatures";
-}
-
 float64_t RandomFourierDotFeatures::post_dot(float64_t dot_result, index_t par_idx) const
 {
 	dot_result += random_coeff(random_coeff.num_rows-1, par_idx);
--- a/src/shogun/features/RandomFourierDotFeatures.cpp
+++ b/src/shogun/features/RandomFourierDotFeatures.cpp
@@ -83,7 +83,7 @@ float64_t RandomFourierDotFeatures::post_dot(float64_t dot_result, index_t par_i
 	return std::cos(dot_result) * constant;
 }
 
-void RandomFourierDotFeatures::generate_random_params(SGVector<float64_t> vec)
+void RandomFourierDotFeatures::generate_random_params(SGVector<float64_t>& vec)
 {
 	NormalDistribution<float64_t> normal_dist;
 	UniformRealDistribution<float64_t> uniform_real_dist(0.0, 2 * Math::PI);
--- a/src/shogun/features/RandomFourierDotFeatures.cpp
+++ b/src/shogun/features/RandomFourierDotFeatures.cpp
@@ -83,11 +83,14 @@ float64_t RandomFourierDotFeatures::post_dot(float64_t dot_result, index_t par_i
 	return std::cos(dot_result) * constant;
 }
 
-SGVector<float64_t> RandomFourierDotFeatures::generate_random_parameter_vector()
+void RandomFourierDotFeatures::generate_random_params(SGVector<float64_t> vec)
 {
 	NormalDistribution<float64_t> normal_dist;
 	UniformRealDistribution<float64_t> uniform_real_dist(0.0, 2 * Math::PI);
-	SGVector<float64_t> vec(feats->get_dim_feature_space()+1);
+
+	/* we need space for both coefficients and offset */
+	ASSERT(vec.vlen == get_num_params());
+
 	switch (kernel)
 	{
 		case GAUSSIAN:
@@ -103,7 +106,12 @@ SGVector<float64_t> RandomFourierDotFeatures::generate_random_parameter_vector()
 		default:
 			error("Unknown kernel");
 	}
-	return vec;
+}
+
+int32_t RandomFourierDotFeatures::get_num_params() const
+{
+	/* returns number of features + offset */
+	return feats->get_dim_feature_space() + 1;
 }
 
 }
--- a/src/shogun/features/RandomFourierDotFeatures.cpp
+++ b/src/shogun/features/RandomFourierDotFeatures.cpp
@@ -88,7 +88,7 @@ void RandomFourierDotFeatures::generate_random_params(SGVector<float64_t>& vec)
 	NormalDistribution<float64_t> normal_dist;
 	UniformRealDistribution<float64_t> uniform_real_dist(0.0, 2 * Math::PI);
 
-	/* we need space for both coefficients and offset */
+	/* Ensure passed vector has appropriate size */
 	ASSERT(vec.vlen == get_num_params());
 
 	switch (kernel)
--- a/src/shogun/features/RandomFourierDotFeatures.h
+++ b/src/shogun/features/RandomFourierDotFeatures.h
@@ -45,10 +45,9 @@ class RandomFourierDotFeatures : public RandomKitchenSinksDotFeatures
 {
 public:
 
-	/** default constructor */
 	RandomFourierDotFeatures();
 
-	/** constructor that creates new random coefficients, basedon the kernel specified and the parameters
+	/** constructor that creates new random coefficients, based on the kernel specified and the parameters
 	 * of the kernel.
 	 *
 	 * @param features the dense features to use as a base
@@ -76,27 +75,19 @@ public:
 	 */
 	RandomFourierDotFeatures(const std::shared_ptr<File>& loader);
 
-	/** copy constructor */
 	RandomFourierDotFeatures(const RandomFourierDotFeatures& orig);
 
-	/** duplicate */
 	virtual std::shared_ptr<Features> duplicate() const;
 
-	/** destructor */
 	virtual ~RandomFourierDotFeatures();
 
-	/** @return object name */
-	virtual const char* get_name() const;
+	virtual const char* get_name() const
+	{
+		return "RandomFourierDotFeatures";
+	}
 
 protected:
 
-	/** subclass must override this to perform any operations
-	 * on the dot result between a feature vector and a parameter vector w
-	 *
-	 * @param dot_result the result of the dot operation
-	 * @param par_idx the idx of the parameter vector
-	 * @return the (optionally) modified result
-	 */
 	virtual float64_t post_dot(float64_t dot_result, index_t par_idx) const;
 
 	virtual void generate_random_params(SGVector<float64_t>& vec);
--- a/src/shogun/features/RandomFourierDotFeatures.h
+++ b/src/shogun/features/RandomFourierDotFeatures.h
@@ -103,7 +103,7 @@ protected:
 	 *
 	 * @return a random parameter vector
 	 */
-	virtual void generate_random_params(SGVector<float64_t> vec);
+	virtual void generate_random_params(SGVector<float64_t>& vec);
 
 	/** @return number of parameters per input dimension */
 	virtual int32_t get_num_params() const;
--- a/src/shogun/features/RandomFourierDotFeatures.h
+++ b/src/shogun/features/RandomFourierDotFeatures.h
@@ -101,7 +101,6 @@ protected:
 
 	virtual void generate_random_params(SGVector<float64_t>& vec);
 
-	/** @return number of parameters per input dimension */
 	virtual int32_t get_num_params() const;
 
 private:
--- a/src/shogun/features/RandomFourierDotFeatures.h
+++ b/src/shogun/features/RandomFourierDotFeatures.h
@@ -99,10 +99,6 @@ protected:
 	 */
 	virtual float64_t post_dot(float64_t dot_result, index_t par_idx) const;
 
-	/** Generates a random parameter vector, subclasses must override this
-	 *
-	 * @return a random parameter vector
-	 */
 	virtual void generate_random_params(SGVector<float64_t>& vec);
 
 	/** @return number of parameters per input dimension */
--- a/src/shogun/features/RandomFourierDotFeatures.h
+++ b/src/shogun/features/RandomFourierDotFeatures.h
@@ -103,7 +103,10 @@ protected:
 	 *
 	 * @return a random parameter vector
 	 */
-	virtual SGVector<float64_t> generate_random_parameter_vector();
+	virtual void generate_random_params(SGVector<float64_t> vec);
+
+	/** @return number of parameters per input dimension */
+	virtual int32_t get_num_params() const;
 
 private:
 	void init(KernelName kernel_name, const SGVector<float64_t>& params);
--- a/src/shogun/features/RandomKitchenSinksDotFeatures.cpp
+++ b/src/shogun/features/RandomKitchenSinksDotFeatures.cpp
@@ -38,9 +38,12 @@ SGMatrix<float64_t> RandomKitchenSinksDotFeatures::generate_random_coefficients(
 	auto num_dim = get_num_params();
 	SGMatrix<float64_t> random_params(num_dim,num_samples);
 
-	for(auto sample : range(num_samples))
-		generate_random_params(random_params.get_column(sample));
-	
+	for (auto sample : range(num_samples))
+	{
+		auto vec = random_params.get_column(sample);
+		generate_random_params(vec);
+	};
+
 	return random_params;
 }
 
--- a/src/shogun/features/RandomKitchenSinksDotFeatures.cpp
+++ b/src/shogun/features/RandomKitchenSinksDotFeatures.cpp
@@ -35,17 +35,12 @@ RandomKitchenSinksDotFeatures::RandomKitchenSinksDotFeatures(
 
 SGMatrix<float64_t> RandomKitchenSinksDotFeatures::generate_random_coefficients()
 {
-	SGVector<float64_t> vec = generate_random_parameter_vector();
-	SGMatrix<float64_t> random_params(vec.vlen, num_samples);
-	for (index_t dim=0; dim<random_params.num_rows; dim++)
-		random_params(dim, 0) = vec[dim];
+	auto num_dim = get_num_params();
+	SGMatrix<float64_t> random_params(num_dim,num_samples);
 
-	for (index_t sample=1; sample<num_samples; sample++)
-	{
-		vec = generate_random_parameter_vector();
-		for (index_t dim=0; dim<random_params.num_rows; dim++)
-			random_params(dim, sample) = vec[dim];
-	}
+	for(auto sample : range(num_samples))
+		generate_random_params(random_params.get_column(sample));
+	
 	return random_params;
 }
 
--- a/src/shogun/features/RandomKitchenSinksDotFeatures.h
+++ b/src/shogun/features/RandomKitchenSinksDotFeatures.h
@@ -224,7 +224,7 @@ protected:
 	 *
 	 * @return a random parameter vector
 	 */
-	virtual void generate_random_params(SGVector<float64_t> vec) = 0;
+	virtual void generate_random_params(SGVector<float64_t>& vec) = 0;
 	
 	/**  Get number of parameters per sample in the random_coeff matrix
 	 *
--- a/src/shogun/features/RandomKitchenSinksDotFeatures.h
+++ b/src/shogun/features/RandomKitchenSinksDotFeatures.h
@@ -226,10 +226,7 @@ protected:
 	 */
 	virtual void generate_random_params(SGVector<float64_t>& vec) = 0;
 	
-	/**  Get number of parameters per sample in the random_coeff matrix
-	 *
-	 * @return dimensionality
-	 */
+	/** @return number of parameters per input dimension */
 	virtual int32_t get_num_params() const = 0;
 private:
 	void init(std::shared_ptr<DotFeatures> dataset, int32_t K);
--- a/src/shogun/features/RandomKitchenSinksDotFeatures.h
+++ b/src/shogun/features/RandomKitchenSinksDotFeatures.h
@@ -224,7 +224,13 @@ protected:
 	 *
 	 * @return a random parameter vector
 	 */
-	virtual SGVector<float64_t> generate_random_parameter_vector()=0;
+	virtual void generate_random_params(SGVector<float64_t> vec) = 0;
+	
+	/**  Get number of parameters per sample in the random_coeff matrix
+	 *
+	 * @return dimensionality
+	 */
+	virtual int32_t get_num_params() const = 0;
 private:
 	void init(std::shared_ptr<DotFeatures> dataset, int32_t K);
 
--- a/tests/unit/features/RandomFourierDotFeatures_unittest.cc
+++ b/tests/unit/features/RandomFourierDotFeatures_unittest.cc
@@ -12,7 +12,7 @@ using namespace shogun;
 TEST(RandomFourierDotFeatures, dot_test)
 {
 	int32_t num_dims = 50;
-	int32_t vecs = 5;
+	int32_t num_vecs = 5;
 	int32_t D = 100;
 
 	SGMatrix<float64_t> w(num_dims+1,D);
@@ -24,8 +24,8 @@ TEST(RandomFourierDotFeatures, dot_test)
 		w(num_dims,i) = 1;
 	}
 
-	SGMatrix<int32_t> data(num_dims, vecs);
-	for (index_t i=0; i<vecs; i++)
+	SGMatrix<int32_t> data(num_dims, num_vecs);
+	for (index_t i=0; i<num_vecs; i++)
 	{
 		for (index_t j=0; j<num_dims; j++)
 			data(j,i) = i+j;
@@ -37,10 +37,10 @@ TEST(RandomFourierDotFeatures, dot_test)
 	auto r_feats = std::make_shared<RandomFourierDotFeatures>(
 			d_feats, D, GAUSSIAN, params, w);
 
-	SGMatrix<float64_t> cross_dot_matrix(vecs, vecs);
-	for (index_t i=0; i<vecs; i++)
+	SGMatrix<float64_t> cross_dot_matrix(num_vecs, num_vecs);
+	for (index_t i=0; i<num_vecs; i++)
 	{
-		for (index_t j=0; j<vecs; j++)
+		for (index_t j=0; j<num_vecs; j++)
 		{
 			cross_dot_matrix(i,j) = r_feats->dot(i,r_feats,j);
 		}
@@ -74,7 +74,7 @@ TEST(RandomFourierDotFeatures, dot_test)
 TEST(RandomFourierDotFeatures, dense_dot_test)
 {
 	int32_t num_dims = 50;
-	int32_t vecs = 5;
+	int32_t num_vecs = 5;
 	int32_t D = 100;
 
 	SGMatrix<float64_t> w(num_dims+1,D);
@@ -86,8 +86,8 @@ TEST(RandomFourierDotFeatures, dense_dot_test)
 		w(num_dims,i) = 1;
 	}
 
-	SGMatrix<int32_t> data(num_dims, vecs);
-	for (index_t i=0; i<vecs; i++)
+	SGMatrix<int32_t> data(num_dims, num_vecs);
+	for (index_t i=0; i<num_vecs; i++)
 	{
 		for (index_t j=0; j<num_dims; j++)
 			data(j,i) = i+j;
@@ -99,10 +99,10 @@ TEST(RandomFourierDotFeatures, dense_dot_test)
 	auto r_feats = std::make_shared<RandomFourierDotFeatures>(
 			d_feats, D, GAUSSIAN, params, w);
 
-	SGMatrix<float64_t> cross_dot_matrix(vecs, vecs);
-	for (index_t i=0; i<vecs; i++)
+	SGMatrix<float64_t> cross_dot_matrix(num_vecs, num_vecs);
+	for (index_t i=0; i<num_vecs; i++)
 	{
-		for (index_t j=0; j<vecs; j++)
+		for (index_t j=0; j<num_vecs; j++)
 		{
 			cross_dot_matrix(i,j) = r_feats->dot(i,r_feats,j);
 		}
@@ -125,7 +125,7 @@ TEST(RandomFourierDotFeatures, dense_dot_test)
 TEST(RandomFourierDotFeatures, add_to_dense_test)
 {
 	int32_t num_dims = 50;
-	int32_t vecs = 5;
+	int32_t num_vecs = 5;
 	int32_t D = 100;
 
 	SGMatrix<float64_t> w(num_dims+1,D);
@@ -137,8 +137,8 @@ TEST(RandomFourierDotFeatures, add_to_dense_test)
 		w(num_dims,i) = 1;
 	}
 
-	SGMatrix<int32_t> data(num_dims, vecs);
-	for (index_t i=0; i<vecs; i++)
+	SGMatrix<int32_t> data(num_dims, num_vecs);
+	for (index_t i=0; i<num_vecs; i++)
 	{
 		for (index_t j=0; j<num_dims; j++)
 			data(j,i) = i+j;
@@ -150,10 +150,10 @@ TEST(RandomFourierDotFeatures, add_to_dense_test)
 	auto r_feats = std::make_shared<RandomFourierDotFeatures>(
 			d_feats, D, GAUSSIAN, params, w);
 
-	SGMatrix<float64_t> cross_dot_matrix(vecs, vecs);
-	for (index_t i=0; i<vecs; i++)
+	SGMatrix<float64_t> cross_dot_matrix(num_vecs, num_vecs);
+	for (index_t i=0; i<num_vecs; i++)
 	{
-		for (index_t j=0; j<vecs; j++)
+		for (index_t j=0; j<num_vecs; j++)
 		{
 			cross_dot_matrix(i,j) = r_feats->dot(i,r_feats,j);
 		}
@@ -175,4 +175,25 @@ TEST(RandomFourierDotFeatures, add_to_dense_test)
 	}
 
 }
+
+TEST(RandomFourierDotFeatures, num_dims_embedding)
+{
+	int32_t num_dims = 50;
+	int32_t num_vecs = 5;
+	int32_t D = 100;
+
+	SGMatrix<float64_t> data(num_dims, num_vecs, false);
+	auto d_feats = std::make_shared<DenseFeatures<float64_t>>(data);
+	SGVector<float64_t> params(1);
+	params[0] = 8;
+	auto r_feats = std::make_shared<RandomFourierDotFeatures>(
+	    d_feats, D, GAUSSIAN, params);
+
+	auto rows = r_feats->get_dim_feature_space();
+	auto cols = r_feats->get_num_vectors();
+
+	EXPECT_EQ(rows, D);
+	EXPECT_EQ(cols, num_vecs);
+}
+
 
--- a/tests/unit/features/RandomFourierDotFeatures_unittest.cc
+++ b/tests/unit/features/RandomFourierDotFeatures_unittest.cc
@@ -13,11 +13,11 @@ TEST(RandomFourierDotFeatures, dot_test)
 {
 	int32_t num_dims = 50;
 	int32_t num_vecs = 5;
-	int32_t dim_output = 100;
+	int32_t num_dims_rff = 100;
 
-	SGMatrix<float64_t> w(num_dims+1,dim_output);
+	SGMatrix<float64_t> w(num_dims+1,num_dims_rff);
 
-	for (index_t i=0; i<dim_output; i++)
+	for (index_t i=0; i<num_dims_rff; i++)
 	{
 		for (index_t j=0; j<num_dims; j++)
 			w(j,i) = i+j;
@@ -35,7 +35,7 @@ TEST(RandomFourierDotFeatures, dot_test)
 	SGVector<float64_t> params(1);
 	params[0] = 8;
 	auto r_feats = std::make_shared<RandomFourierDotFeatures>(
-			d_feats, dim_output, GAUSSIAN, params, w);
+			d_feats, num_dims_rff, GAUSSIAN, params, w);
 
 	SGMatrix<float64_t> cross_dot_matrix(num_vecs, num_vecs);
 	for (index_t i=0; i<num_vecs; i++)
@@ -75,11 +75,11 @@ TEST(RandomFourierDotFeatures, dense_dot_test)
 {
 	int32_t num_dims = 50;
 	int32_t num_vecs = 5;
-	int32_t dim_output = 100;
+	int32_t num_dims_rff = 100;
 
-	SGMatrix<float64_t> w(num_dims+1,dim_output);
+	SGMatrix<float64_t> w(num_dims+1,num_dims_rff);
 
-	for (index_t i=0; i<dim_output; i++)
+	for (index_t i=0; i<num_dims_rff; i++)
 	{
 		for (index_t j=0; j<num_dims; j++)
 			w(j,i) = i+j;
@@ -97,7 +97,7 @@ TEST(RandomFourierDotFeatures, dense_dot_test)
 	SGVector<float64_t> params(1);
 	params[0] = 8;
 	auto r_feats = std::make_shared<RandomFourierDotFeatures>(
-			d_feats, dim_output, GAUSSIAN, params, w);
+			d_feats, num_dims_rff, GAUSSIAN, params, w);
 
 	SGMatrix<float64_t> cross_dot_matrix(num_vecs, num_vecs);
 	for (index_t i=0; i<num_vecs; i++)
@@ -114,7 +114,7 @@ TEST(RandomFourierDotFeatures, dense_dot_test)
 	SGVector<float64_t> precomputed_vec(vec, 5, false);
 	for (index_t i=0; i<5; i++)
 	{
-		SGVector<float64_t> ones(dim_output);
+		SGVector<float64_t> ones(num_dims_rff);
 		SGVector<float64_t>::fill_vector(ones.vector, ones.vlen, 1);
 		float64_t dot = r_feats->dot(i, ones);
 		EXPECT_NEAR(dot, vec[i], e);
@@ -126,11 +126,11 @@ TEST(RandomFourierDotFeatures, add_to_dense_test)
 {
 	int32_t num_dims = 50;
 	int32_t num_vecs = 5;
-	int32_t dim_output = 100;
+	int32_t num_dims_rff = 100;
 
-	SGMatrix<float64_t> w(num_dims+1,dim_output);
+	SGMatrix<float64_t> w(num_dims+1,num_dims_rff);
 
-	for (index_t i=0; i<dim_output; i++)
+	for (index_t i=0; i<num_dims_rff; i++)
 	{
 		for (index_t j=0; j<num_dims; j++)
 			w(j,i) = i+j;
@@ -148,7 +148,7 @@ TEST(RandomFourierDotFeatures, add_to_dense_test)
 	SGVector<float64_t> params(1);
 	params[0] = 8;
 	auto r_feats = std::make_shared<RandomFourierDotFeatures>(
-			d_feats, dim_output, GAUSSIAN, params, w);
+			d_feats, num_dims_rff, GAUSSIAN, params, w);
 
 	SGMatrix<float64_t> cross_dot_matrix(num_vecs, num_vecs);
 	for (index_t i=0; i<num_vecs; i++)
@@ -165,34 +165,34 @@ TEST(RandomFourierDotFeatures, add_to_dense_test)
 	SGVector<float64_t> precomputed_vec(vec, 5, false);
 	for (index_t i=0; i<5; i++)
 	{
-		SGVector<float64_t> zeros(dim_output);
+		SGVector<float64_t> zeros(num_dims_rff);
 		SGVector<float64_t>::fill_vector(zeros.vector, zeros.vlen, 0);
 		r_feats->add_to_dense_vec(1, i, zeros.vector, zeros.vlen, false);
 		float64_t sum = 0;
-		for (index_t j=0; j<dim_output; j++)
+		for (index_t j=0; j<num_dims_rff; j++)
 			sum += zeros[j];
 		EXPECT_NEAR(sum, vec[i], e);
 	}
 
 }
 
-TEST(RandomFourierDotFeatures, num_dims_rffs)
+TEST(RandomFourierDotFeatures, dim_output)
 {
 	int32_t num_dims = 50;
 	int32_t num_vecs = 5;
-	int32_t dim_output = 100;
+	int32_t num_dims_rff = 100;
 
 	SGMatrix<float64_t> data(num_dims, num_vecs, false);
 	auto d_feats = std::make_shared<DenseFeatures<float64_t>>(data);
 	SGVector<float64_t> params(1);
 	params[0] = 8;
 	auto r_feats = std::make_shared<RandomFourierDotFeatures>(
-	    d_feats, dim_output, GAUSSIAN, params);
+	    d_feats, num_dims_rff, GAUSSIAN, params);
 
 	auto rows = r_feats->get_dim_feature_space();
 	auto cols = r_feats->get_num_vectors();
 
-	EXPECT_EQ(rows, dim_output);
+	EXPECT_EQ(rows, num_dims_rff);
 	EXPECT_EQ(cols, num_vecs);
 }
 
--- a/tests/unit/features/RandomFourierDotFeatures_unittest.cc
+++ b/tests/unit/features/RandomFourierDotFeatures_unittest.cc
@@ -176,7 +176,7 @@ TEST(RandomFourierDotFeatures, add_to_dense_test)
 
 }
 
-TEST(RandomFourierDotFeatures, num_dims_embedding)
+TEST(RandomFourierDotFeatures, num_dims_rffs)
 {
 	int32_t num_dims = 50;
 	int32_t num_vecs = 5;
--- a/tests/unit/features/RandomFourierDotFeatures_unittest.cc
+++ b/tests/unit/features/RandomFourierDotFeatures_unittest.cc
@@ -13,11 +13,11 @@ TEST(RandomFourierDotFeatures, dot_test)
 {
 	int32_t num_dims = 50;
 	int32_t num_vecs = 5;
-	int32_t D = 100;
+	int32_t dim_output = 100;
 
-	SGMatrix<float64_t> w(num_dims+1,D);
+	SGMatrix<float64_t> w(num_dims+1,dim_output);
 
-	for (index_t i=0; i<D; i++)
+	for (index_t i=0; i<dim_output; i++)
 	{
 		for (index_t j=0; j<num_dims; j++)
 			w(j,i) = i+j;
@@ -35,7 +35,7 @@ TEST(RandomFourierDotFeatures, dot_test)
 	SGVector<float64_t> params(1);
 	params[0] = 8;
 	auto r_feats = std::make_shared<RandomFourierDotFeatures>(
-			d_feats, D, GAUSSIAN, params, w);
+			d_feats, dim_output, GAUSSIAN, params, w);
 
 	SGMatrix<float64_t> cross_dot_matrix(num_vecs, num_vecs);
 	for (index_t i=0; i<num_vecs; i++)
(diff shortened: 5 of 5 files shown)


src/shogun/features/RandomFourierDotFeatures.cpp
```diff
@@ -90,15 +90,13 @@ SGVector<float64_t> RandomFourierDotFeatures::generate_random_parameter_vector()
...
-		case GAUSSIAN:
-			for (index_t i=0; i<vec.vlen-1; i++)
-			{
-				vec[i] = std::sqrt((float64_t)1 / kernel_params[0]) *
-				         std::sqrt(2.0) * normal_dist(m_prng);
-			}
-
+		case GAUSSIAN:{
+			auto parameter = std::sqrt((float64_t)1 / kernel_params[0]) *
+			std::sqrt(2.0) * normal_dist(m_prng);
```
gf712: this is incorrect, you are now just generating a single random number, rather than drawing from the distribution `vlen` times.

src/shogun/features/RandomFourierDotFeatures.cpp
```diff
@@ -90,15 +90,13 @@ SGVector<float64_t> RandomFourierDotFeatures::generate_random_parameter_vector()
...
+			linalg::add_scalar(vec,parameter);
```
gf712: here you would have to set vec to zeros before adding anything to it

gf712:
@NanuSai to show benchmarks could you run just the one you are targetting by running the binary inside `build/bin`? That will show you the breakdown for all the different dimensions and sizes of inputs that are being benchmarked. In future show that instead of the total running time.

jonpsy:
> @NanuSai to show benchmarks could you run just the one you are targetting by running the binary inside `build/bin`? That will show you the breakdown for all the different dimensions and sizes of inputs that are being benchmarked. In future show that instead of the total running time.

Hey, yea sorry for that. The docs just said to run ```ctest -L benchmark``` to run, but since our target is RandomFourierDotFeatures
We have **Without changes** : 27.06 seconds
                **With changes** : 26.21 seconds
But yes, next time I'll do better, thanks !

src/shogun/features/RandomFourierDotFeatures.cpp
karlnapf: @NanuSai you could write a unit test before you refactor things, that helps catching such bugs

src/shogun/features/RandomFourierDotFeatures.cpp
jonpsy: I thought SGVectors are initialized to 0s on CPU by default? And if we aren't using linalg in this file, then I don't think setting values to 0 would be necessary.

src/shogun/features/RandomFourierDotFeatures.cpp
jonpsy: True, but to generate random parameter for each instance we're forced to use loops (or is there a better way?) and if that's the case I think the current implementation is good, at least for this file.

EDIT: On second thought, We can use OpenMP parallelization for generating the random numbers, that ought to improve performance.

src/shogun/features/RandomFourierDotFeatures.cpp
```diff
@@ -90,16 +90,16 @@ SGVector<float64_t> RandomFourierDotFeatures::generate_random_parameter_vector()
 	SGVector<float64_t> vec(feats->get_dim_feature_space()+1);
 	switch (kernel)
 	{
-		case GAUSSIAN:
+		case GAUSSIAN:		
+			#pragma omp parallel for
```
gf712: Is the result deterministic? Do you get the same result with the same seed?

src/shogun/features/RandomFourierDotFeatures_benchmark.cc
```diff
@@ -9,6 +9,7 @@
 #include "shogun/features/DotFeatures_benchmark.h"
 #include "shogun/features/RandomFourierDotFeatures.h"
 #include "shogun/mathematics/UniformIntDistribution.h"
+#include "shogun/mathematics/linalg/LinalgNamespace.h"
```
gf712: Why did you add the header file here?

src/shogun/features/RandomFourierDotFeatures.cpp
gf712: Also you might observe that there are repeating patterns in the matrix because each thread would start from the same state.. I know that can happen with rand...

gf712: http://www.thesalmons.org/john/random123/papers/random123sc11.pdf You can see it’s a bit more difficult than that... For a nice explanation on how to achieve this https://github.com/google/jax/blob/master/design_notes/prng.md

src/shogun/features/RandomFourierDotFeatures_benchmark.cc
jonpsy: I thought I'd add a unit test inside this benchmark, using LinalgNamespace, too early though.

src/shogun/features/RandomFourierDotFeatures.cpp
jonpsy: I suppose you're talking about [this](https://stackoverflow.com/questions/4287531/how-to-generate-random-numbers-in-parallel)?

gf712: for example, yes. The point is that first you have no guarantee that omp will schedule the threads in the same way for different runs, then there is the issue with dispatching the state... long story short, this is something we will have to work on at some point, but I don't think what you wrote does what you expect it to. This is also beyond the scope of this PR imo

jonpsy:
Adding omp parallelization now is causing some troubles and after your opinions I agree it's going a bit off topic. For now the only change is ```linalg::add_vector``` in ```RandomKitchenSinksDotFeatures.cpp```.  I'll open up a new PR regarding the omp parallelization and refer this PR

src/shogun/features/RandomFourierDotFeatures_benchmark.cc
karlnapf: unit tests go to unit tests, benchmarks go to benchmarks

src/shogun/features/RandomFourierDotFeatures.cpp
```diff
@@ -96,7 +96,6 @@ SGVector<float64_t> RandomFourierDotFeatures::generate_random_parameter_vector()
 				vec[i] = std::sqrt((float64_t)1 / kernel_params[0]) *
 				         std::sqrt(2.0) * normal_dist(m_prng);
 			}
-
```
karlnapf: pls no whitespace changes

src/shogun/features/RandomKitchenSinksDotFeatures.cpp
```diff
@@ -37,14 +38,12 @@ SGMatrix<float64_t> RandomKitchenSinksDotFeatures::generate_random_coefficients(
 {
 	SGVector<float64_t> vec = generate_random_parameter_vector();
 	SGMatrix<float64_t> random_params(vec.vlen, num_samples);
-	for (index_t dim=0; dim<random_params.num_rows; dim++)
-		random_params(dim, 0) = vec[dim];
+	linalg::add_col_vec(random_params,0,vec,random_params);
```
karlnapf: does this do the same thing? It assigns, rather than adding?

jonpsy: Since it already initiated with 0, I think adding and assigning won't make a difference here. I'm using this btw https://www.shogun-toolbox.org/api/latest/namespaceshogun_1_1linalg.html#a604e1a0ccc54502a649d7d81cc20dd10.

src/shogun/features/RandomFourierDotFeatures.cpp
karlnapf: make sure to check the diff. there are now two witespace changes as you added tabs

src/shogun/features/RandomKitchenSinksDotFeatures.cpp
karlnapf: but from this code, it is not obvious that they were initialized with zero. So rather assign, easier to read

jonpsy: Done !

src/shogun/features/RandomKitchenSinksDotFeatures.cpp
```diff
@@ -37,14 +38,13 @@ SGMatrix<float64_t> RandomKitchenSinksDotFeatures::generate_random_coefficients(
 {
 	SGVector<float64_t> vec = generate_random_parameter_vector();
 	SGMatrix<float64_t> random_params(vec.vlen, num_samples);
-	for (index_t dim=0; dim<random_params.num_rows; dim++)
-		random_params(dim, 0) = vec[dim];
+	linalg::range_fill(random_params,0.0);
+	linalg::add_col_vec(random_params,0,vec,random_params);
```
jonpsy: I was wondering, why did we separate the case for col(0) when it can be put in the loop below?

src/shogun/features/RandomKitchenSinksDotFeatures.cpp
```diff
@@ -37,14 +38,13 @@ SGMatrix<float64_t> RandomKitchenSinksDotFeatures::generate_random_coefficients(
 {
 	SGVector<float64_t> vec = generate_random_parameter_vector();
 	SGMatrix<float64_t> random_params(vec.vlen, num_samples);
-	for (index_t dim=0; dim<random_params.num_rows; dim++)
-		random_params(dim, 0) = vec[dim];
+	linalg::range_fill(random_params,0.0);
```
gf712: why `range_fill`?

jonpsy: I'm initiating the values with zero.

gf712: but `range_fill` is to initialise values from `[0, 1,..., size]`, no? there's a function for zeroing out

jonpsy: You're right. ```random_params.zero()``` should do the trick !. This also "assigns" rather than add.

src/shogun/features/RandomKitchenSinksDotFeatures.cpp
```diff
@@ -37,14 +38,13 @@ SGMatrix<float64_t> RandomKitchenSinksDotFeatures::generate_random_coefficients(
 {
 	SGVector<float64_t> vec = generate_random_parameter_vector();
 	SGMatrix<float64_t> random_params(vec.vlen, num_samples);
-	for (index_t dim=0; dim<random_params.num_rows; dim++)
-		random_params(dim, 0) = vec[dim];
+	random_params.zero();
```
karlnapf: all this is overly complicated.
There should just be a boradcasting style method to fill a matrix with a given vector (either column or row) in linalg.
```
linalg::set(random_params, vec); // needs to be implemented
```

Or a single simple loop using memset, in the lines of (which would also be the implementation in linalg)
```
for i in range(num_samples)
  sg_memcpy(random_params.data() + i*num_samples, vec.data(), vec.vlen); // note I didnt test this
```
The above could even be shortened with a c++ `for_each` or similar

jonpsy: Okay there has been work on creating column iterator for SGMatrix. Once that's merged we can do 
``` std::for_each(random_params.begin_col(),random_params.end_col(),[](const auto& random_vec){ random_vec = generate_random_vector();} ```

Comment (karlnapf) [reply src/shogun/features/RandomKitchenSinksDotFeatures.cpp]: btw there is also a simple `set_column` method of SGMatrix

Comment (jonpsy) [reply src/shogun/features/RandomKitchenSinksDotFeatures.cpp]: Aye, this seems to do the trick. It's compiling well and unit tests pass.  Any changes you recommend before we finalize?

**Comment by karlnapf** [comment]
yes lgtm. Thanks!

File: src/shogun/features/RandomKitchenSinksDotFeatures.cpp [hunk 10]
```diff
@@ -37,14 +38,11 @@ SGMatrix<float64_t> RandomKitchenSinksDotFeatures::generate_random_coefficients(
 {
 	SGVector<float64_t> vec = generate_random_parameter_vector();
```
karlnapf: actually no, you are generating the vec once and then throw it away without looking at it

jonpsy: I was gonna change it anyway 😅. I made a review on it remember?

jonpsy: So nothing besides this, then

src/shogun/features/RandomKitchenSinksDotFeatures.cpp
jonpsy: Okay, this is handled !

src/shogun/features/RandomKitchenSinksDotFeatures.cpp
jonpsy: vec.vlen has num_feature_dim_space + 1  i.e. num_samples  + 1 . So should I hardcode that? in 
```
random_params(num_samples + 1 , num_samples)
```

karlnapf: what I mean is that you create the vector and the you overwrite it in the first iteration of the loop

src/shogun/features/RandomKitchenSinksDotFeatures.cpp
```diff
@@ -35,16 +36,13 @@ RandomKitchenSinksDotFeatures::RandomKitchenSinksDotFeatures(
 
 SGMatrix<float64_t> RandomKitchenSinksDotFeatures::generate_random_coefficients()
 {
-	SGVector<float64_t> vec = generate_random_parameter_vector();
-	SGMatrix<float64_t> random_params(vec.vlen, num_samples);
-	for (index_t dim=0; dim<random_params.num_rows; dim++)
-		random_params(dim, 0) = vec[dim];
+	SGVector<float64_t> vec;
```
karlnapf: you can move this variable inside the loop right?

src/shogun/features/RandomKitchenSinksDotFeatures.cpp
```diff
@@ -35,16 +36,13 @@ RandomKitchenSinksDotFeatures::RandomKitchenSinksDotFeatures(
...
+	SGMatrix<float64_t> random_params(num_samples + 1, num_samples);
 
-	for (index_t sample=1; sample<num_samples; sample++)
+	for (index_t sample=0; sample<num_samples; sample++)
 	{
 		vec = generate_random_parameter_vector();
```
gf712:
```suggestion
		auto vec = generate_random_parameter_vector();
```
and remove the vec from above!

src/shogun/features/RandomKitchenSinksDotFeatures.cpp
jonpsy: right, but we would be redeclaring it every loop iteration won't we?. That's why I defined it outside.

gf712: any good compiler will see that, so it won't harm the codegen. Writing declarations like this is more of ANSI C style :D

gf712:
@karlnapf merge?

src/shogun/features/RandomKitchenSinksDotFeatures.cpp
jonpsy: Thanks for everything till now ^^

karlnapf: @NanuSai make sure to read a bit on C++ compiler optimizations. The code you write is not necessarily what will be executed on the machine, as gil pointed out above. If you look at the assembly of the compiled file, you will see that defining the variable before or in the loop leads to exactly the same machine code.

src/shogun/features/RandomKitchenSinksDotFeatures.cpp
```diff
@@ -35,16 +36,12 @@ RandomKitchenSinksDotFeatures::RandomKitchenSinksDotFeatures(
...
-		random_params(dim, 0) = vec[dim];
+	SGMatrix<float64_t> random_params(num_samples + 1, num_samples);
 
-	for (index_t sample=1; sample<num_samples; sample++)
+	for (index_t sample=0; sample<num_samples; sample++)
 	{
-		vec = generate_random_parameter_vector();
-		for (index_t dim=0; dim<random_params.num_rows; dim++)
-			random_params(dim, sample) = vec[dim];
+		auto vec = generate_random_parameter_vector();
```
karlnapf: technically we could even use `const auto&` here to avoid one instantiation :) @gf712 ?

karlnapf:
> @karlnapf merge?

yes!

src/shogun/features/RandomKitchenSinksDotFeatures.cpp
gf712: yes, we could, and maybe even should!

src/shogun/features/RandomKitchenSinksDotFeatures.cpp
```diff
@@ -6,6 +6,7 @@
 
 #include <shogun/features/DenseFeatures.h>
 #include <shogun/features/RandomKitchenSinksDotFeatures.h>
+#include <shogun/mathematics/linalg/LinalgNamespace.h>
```
gf712: is this needed btw?

jonpsy: I suppose not, btw I think we should change the PR name because linalg isn't used at all

src/shogun/features/RandomKitchenSinksDotFeatures.cpp
jonpsy: we will :)

karlnapf: does the set_column method then need to accept const references?

jonpsy: yes

src/shogun/features/RandomKitchenSinksDotFeatures.cpp
```diff
@@ -35,16 +35,12 @@ RandomKitchenSinksDotFeatures::RandomKitchenSinksDotFeatures(
 
 SGMatrix<float64_t> RandomKitchenSinksDotFeatures::generate_random_coefficients()
 {
-	SGVector<float64_t> vec = generate_random_parameter_vector();
-	SGMatrix<float64_t> random_params(vec.vlen, num_samples);
-	for (index_t dim=0; dim<random_params.num_rows; dim++)
-		random_params(dim, 0) = vec[dim];
+	SGMatrix<float64_t> random_params(num_samples + 1, num_samples);
```
karlnapf: why has the dimension now changed? the number of rows should be the dimension of the feature space

karlnapf: not sure where this `num_samples + 1` comes from?

jonpsy: ``` generate_random_parameter_vector()``` returns a vec of vlen  ```get_feature_dim_space() + 1 ```. 

```get_feature_dim_space()```  returns num_samples

So in essence, vec.vlen = num_samples + 1.

Comment (karlnapf) [reply src/shogun/features/RandomKitchenSinksDotFeatures.cpp]: `generate_random_parameter_vector` is a purely virtual method though, so it could return different things in the future. I realise now why there was a single vector generated first, and then the remaining...in order to be sure about the dimensions of the matrix. I did't realise this before. You will have to generate one parameter vector, then allocates the matrix, and then copy it to the first column of the matrix, then the rest.
BTW I am puzzled what this matrix represents actually. Usually one needs to store frequencies (coefficients) that are dim x num_rf, see e.g. [here](https://github.com/karlnapf/kernel_exp_family/blob/master/kernel_exp_family/kernels/kernels.py#L316). So num_samples+1 x num_samples doesnt make sense to me. Maybe you could investigate that?

Comment (karlnapf) [reply src/shogun/features/RandomKitchenSinksDotFeatures.cpp]: btw one optimization that could be done here is (once you know the dimension of the matrix), that you dont generate random coefficients and then copy them, but rather provide a reference to a vector to the `generate_random...` method so that it can write to the matrix memory straight away. Saves us copying the whole matrix once.

Comment (karlnapf) [reply src/shogun/features/RandomKitchenSinksDotFeatures.cpp]: actually thinking about all this more. I am not sure how much worth it even is doing these optimizations here, as the design is really not that good (see my email with many suggestions how to improve it)

Comment (karlnapf) [reply src/shogun/features/RandomKitchenSinksDotFeatures.cpp]: Ok for now let's do this

* add a method that can tell us the number of rows of this matrix, so that we can allocate it without having to generate a random vec or hard-code anything.
* add a way for the generate random vector to write the results to a pre-specified memory block (the column of the matrix)
* then loop over the columns and write the random parameters directly to the matrix column wise

Comment (karlnapf) [reply src/shogun/features/RandomKitchenSinksDotFeatures.cpp]:
```
auto d = .... // figure out the dimension via a method or similar
SGMatrix<float64_t> random_params(d, num_samples);
for (auto i : range(num_samples))
  generate_random_parameter_vector(random_params.get_column(i)); // not sure this exists, but something in these lines will
```

Comment (jonpsy) [reply src/shogun/features/RandomKitchenSinksDotFeatures.cpp]: > `generate_random_parameter_vector` is a purely virtual method though, so it could return different things in the future. I realise now why there was a single vector generated first, and then the remaining...in order to be sure about the dimensions of the matrix. I did't realise this before. You will have to generate one parameter vector, then allocates the matrix, and then copy it to the first column of the matrix, then the rest.
> BTW I am puzzled what this matrix represents actually. Usually one needs to store frequencies (coefficients) that are dim x num_rf, see e.g. [here](https://github.com/karlnapf/kernel_exp_family/blob/master/kernel_exp_family/kernels/kernels.py#L316). So num_samples+1 x num_samples doesnt make sense to me. Maybe you could investigate that?

I was thinking the same  about why num_samples + 1 is taken, apparently it stores numbers from gaussiann in [0 .... num_samples] and 1 number from uniform distrib in [num_sample+1]. But isn't the number sampled from uniform_distrb to act as an offset ?.

Comment (karlnapf) [reply src/shogun/features/RandomKitchenSinksDotFeatures.cpp]: Ah yes of course it’s the offset. 
Another thing to refactor... but that’s for later

I suggest you add a method that returns the number of parameters per random feature (use it inside the random vec generation) and then allocate the matrix using that.

Comment (karlnapf) [reply src/shogun/features/RandomKitchenSinksDotFeatures.cpp]: Storing the offset in the last component means that we cannot do matrix vector products for the projection. That would be another improvement (in tidiness and speed)

Comment (jonpsy) [reply src/shogun/features/RandomKitchenSinksDotFeatures.cpp]: Okay upon investigation here's what I found.
```get_dim_feature_space()``` returns the dim of feature space(K) of the final RFF transformed feature( K x m ) { K= number of samples, m = number of examples}
However, our ```random_param``` matrix must have dimensions ```dim_feature_space x num_samples```.(n x m) . Here the feature_space is of the **input vector**(n) not of the final vector(k) but the value used if k. 

Let's continue this over the mailing list, meanwhile, I'll investigate further.

Comment (karlnapf) [reply src/shogun/features/RandomKitchenSinksDotFeatures.cpp]: This is all expceted. Here is a python implementation I wrote a while ago that might make it easier to understand [link](https://github.com/karlnapf/kernel_exp_family/blob/master/kernel_exp_family/kernels/kernels.py#L316)

File: src/shogun/features/RandomFourierDotFeatures.cpp [hunk 16]
```diff
@@ -83,27 +83,27 @@ float64_t RandomFourierDotFeatures::post_dot(float64_t dot_result, index_t par_i
 	return std::cos(dot_result) * constant;
 }
 
-SGVector<float64_t> RandomFourierDotFeatures::generate_random_parameter_vector()
+void RandomFourierDotFeatures::generate_random_parameter_vector(SGMatrix<float64_t>& mat, index_t col)
```
karlnapf: rather provide a result vector than a matrix

src/shogun/features/RandomFourierDotFeatures.cpp
```diff
@@ -83,27 +83,27 @@ float64_t RandomFourierDotFeatures::post_dot(float64_t dot_result, index_t par_i
...
 {
 	NormalDistribution<float64_t> normal_dist;
 	UniformRealDistribution<float64_t> uniform_real_dist(0.0, 2 * Math::PI);
-	SGVector<float64_t> vec(feats->get_dim_feature_space()+1);
+	auto len = mat.num_rows + 1;
```
karlnapf: it is a bad idea to rely on the input for the size. Rather assert/require that the input (vector) has dimension of feature space + 1 as size.

src/shogun/features/RandomKitchenSinksDotFeatures.cpp
```diff
@@ -8,6 +8,7 @@
 #include <shogun/features/RandomKitchenSinksDotFeatures.h>
 #include <typeinfo>
 #include <utility>
+#include<iostream>
```
karlnapf: that should be removed

src/shogun/features/RandomKitchenSinksDotFeatures.cpp
```diff
@@ -35,20 +36,15 @@ RandomKitchenSinksDotFeatures::RandomKitchenSinksDotFeatures(
 
 SGMatrix<float64_t> RandomKitchenSinksDotFeatures::generate_random_coefficients()
 {
-	SGVector<float64_t> vec = generate_random_parameter_vector();
-	SGMatrix<float64_t> random_params(vec.vlen, num_samples);
-	for (index_t dim=0; dim<random_params.num_rows; dim++)
-		random_params(dim, 0) = vec[dim];
+	SGMatrix<float64_t> random_params(feats->get_dim_feature_space() + 1 , num_samples);
```
karlnapf: as explained before, this feature space dim +1 shouldnt be harcoded in here. You need to add a purely virtual method to the base class that returns the number of dimensions. OR you first ask for the first parameter vector, then allocate the matrix and fill the rest. This is a general class that could work with other parametrizations so we don't want to fixate this here.

tests/unit/features/RandomFourierDotFeatures_unittest.cc
```diff
@@ -176,3 +176,30 @@ TEST(RandomFourierDotFeatures, add_to_dense_test)
 
 }
 
+TEST(RandomFourierDotFeatures, random_coeff_dim)
+{
+	int32_t num_dims = 50;
+	int32_t vecs = 5;
+	int32_t D = 100;
+
+	SGMatrix<int32_t> data(num_dims, vecs);
```
karlnapf: could use a matrix full of constants to reduce code here if youre not checking for the results

tests/unit/features/RandomFourierDotFeatures_unittest.cc
```diff
@@ -176,3 +176,30 @@ TEST(RandomFourierDotFeatures, add_to_dense_test)
...
+	for (index_t i=0; i<vecs; i++)
+	{
+		for (index_t j=0; j<num_dims; j++)
+			data(j,i) = i+j;
+	}
+
+	auto d_feats = std::make_shared<DenseFeatures<int32_t>>(data);
+	SGVector<float64_t> params(1);
```
karlnapf, line 187: this is another thing to refactor, this vector of params (in another PR) rather I would take a kernel instance and get the parameters from that

tests/unit/features/RandomFourierDotFeatures_unittest.cc
```diff
@@ -176,3 +176,30 @@ TEST(RandomFourierDotFeatures, add_to_dense_test)
 
 }
 
+TEST(RandomFourierDotFeatures, random_coeff_dim)
```
karlnapf: this test is for the gaussian rff, and it should be named accordingly I guess?

tests/unit/features/RandomFourierDotFeatures_unittest.cc
```diff
@@ -176,3 +176,30 @@ TEST(RandomFourierDotFeatures, add_to_dense_test)
...
+{
+	int32_t num_dims = 50;
+	int32_t vecs = 5;
```
karlnapf: could you use consistent variable names?

tests/unit/features/RandomFourierDotFeatures_unittest.cc
```diff
@@ -176,3 +176,30 @@ TEST(RandomFourierDotFeatures, add_to_dense_test)
...
+	{
+		for (index_t j=0; j<num_dims; j++)
+			data(j,i) = i+j;
+	}
+
+	auto d_feats = std::make_shared<DenseFeatures<int32_t>>(data);
+	SGVector<float64_t> params(1);
+	params[0] = 8;
+	auto r_feats = std::make_shared<RandomFourierDotFeatures>(
+			d_feats, D, GAUSSIAN, params);
```
karlnapf: ẁe could make this test parametrized so that it tests the embedding for all kernels?

src/shogun/features/RandomFourierDotFeatures.cpp
gf712: Indeed, what you wrote here is something you’d see in C (the reference would be a pointer). I think this is google style C++, but we don’t use it, as in modern C++ the compiler optimises the return value allocation and we can even have multiple return values.

tests/unit/features/RandomFourierDotFeatures_unittest.cc
```diff
@@ -176,3 +176,30 @@ TEST(RandomFourierDotFeatures, add_to_dense_test)
...
+	
+	auto rows = r_feats->get_random_coefficients().num_rows;
```
karlnapf: actually I think this is maybe not the best thing to unit test. You see there can be many ways to compute the embedding, but what counts is that the dimension of the embedding is correct, rather than some internal method

src/shogun/features/RandomFourierDotFeatures.cpp
```diff
@@ -83,27 +83,27 @@ float64_t RandomFourierDotFeatures::post_dot(float64_t dot_result, index_t par_i
...
+void RandomFourierDotFeatures::generate_random_parameter_vector(SGMatrix<float64_t>& mat, index_t col)
 {
 	NormalDistribution<float64_t> normal_dist;
 	UniformRealDistribution<float64_t> uniform_real_dist(0.0, 2 * Math::PI);
-	SGVector<float64_t> vec(feats->get_dim_feature_space()+1);
+	auto len = mat.num_rows + 1;       
+
 	switch (kernel)
 	{
 		case GAUSSIAN:
```
gf712, line 91: @karlnapf is there a reason why we can’t reuse the kernel implementation of a Kernel class?

gf712: For example GaussianKernel in this case

src/shogun/features/RandomFourierDotFeatures.cpp
karlnapf: I don't think the compiler would optimise out the method allocating a new sgvector and the an memcpy call subsequently. Rather, we would allocate the matrix first, and then call this method here with wrapped column vectors (There is a method in SGMatrix that returns a non memory managed column vector)

karlnapf: or would it @gf712 ? That would blow my mind :)

src/shogun/features/RandomFourierDotFeatures.cpp
karlnapf: The code to sample spectral densities of the kernel (happens to be a gaussian for the gaussian kernel) doesnt really exist in the kernel class, but I agree we should add it there. The OOP way to do this would be to add a virtual method `sample_spectral_density` to `Kernel`, and then throw a runtime error if that is not implemented. Then all this switching here would be unnecessary. Furthermore, the shitty passing the kernel bandwidth to the embedding class would not happen (instead it would just read it from the kernel object). I have suggested that to @NanuSai already.

src/shogun/features/RandomFourierDotFeatures.cpp
gf712: No, what I mean is the sgmatrix allocation is done on the calling stack. So that it isn’t copied from this stack to the caller. That’s what RVO does. The sgmatrix data pointer is in the heap, so that can’t be optimised here.

src/shogun/features/RandomKitchenSinksDotFeatures.cpp
jonpsy: Sorry, i used that for debugging forgot to remove.

karlnapf: debugging with cout is extremely inefficient. Check out gdb, which makes everything much easier

vigsterkr: SG_DEBUG and if you use it consciously you can even leave it in the code.

vigsterkr:
@NanuSai those screenshots about benchmarks doesn't tell too much... you should actually use `-VV` to have a better understandings. that's where you get the details of the gbenchmark

vigsterkr:
`more like +- 1 seconds` is not true as most of the tests has that diff what you are claiming for the random fourier dot.... again use `-VV` for ctest to actually see what the real diff is

src/shogun/features/RandomKitchenSinksDotFeatures.cpp
karlnapf: +1

tests/unit/features/RandomFourierDotFeatures_unittest.cc
jonpsy: I don't quite get what you mean, num_dims(n) and vecs(m) is used throughout the unit test file. I was just following the convention.

src/shogun/features/RandomKitchenSinksDotFeatures.cpp
jonpsy: Question,  the random_params matrix transform our dimensionality of feature space from n -> k. So, shouldn't it's rows always be feature space dim  + 1 ?.

tests/unit/features/RandomFourierDotFeatures_unittest.cc
karlnapf: the point of touching this code is that it is quite badly written, and we are trying to improve things whenever we write something new or touch existing stuff. consistent variables names simply means that you use the same pattern everywhere: if you prepend "num" to denote a number, then you do don't omit this in the next variable.

src/shogun/features/RandomKitchenSinksDotFeatures.cpp
karlnapf: no, *as explained above* that is only the case if you use RFF features with the offset structure that is used here. But that can change for different random feature classes, maybe some RFF classes dont even use offsets. Rule of thumb: never hardcode such things, always put them into a method. That is the whole point of doing this in an object oriented design.

tests/unit/features/RandomFourierDotFeatures_unittest.cc
jonpsy: Hmm, how about we create a test_fixture? I mean there's a lot of repetition here.

tests/unit/features/RandomFourierDotFeatures_unittest.cc
jonpsy: Yeah, that's apt

tests/unit/features/RandomFourierDotFeatures_unittest.cc
jonpsy: Actually, I agree. I've added an ```ASSERT``` in  ```generate_randm_param_vector()```, so it does the job for us.

jonpsy: And since it handles the KERNEL based switching(to be modified) it acts like a parameterized test as well. So I guess this test doesn't serve much purpose.

EDIT: In the existing unit test cases, none of them used ```generate_random_parameter_vector()```  hence the error slipped past us, which we fixed. This unit test calls the constructor which relies on this method. Maybe not for dimension check, but maybe we can use this test case to check if the method we talked about is working correctly or not.

src/shogun/features/RandomFourierDotFeatures.cpp
```diff
@@ -83,17 +83,21 @@ float64_t RandomFourierDotFeatures::post_dot(float64_t dot_result, index_t par_i
...
 }
 
-SGVector<float64_t> RandomFourierDotFeatures::generate_random_parameter_vector()
+void RandomFourierDotFeatures::generate_random_parameter_vector(SGVector<float64_t>&& vec)
 {
 	NormalDistribution<float64_t> normal_dist;
 	UniformRealDistribution<float64_t> uniform_real_dist(0.0, 2 * Math::PI);
-	SGVector<float64_t> vec(feats->get_dim_feature_space()+1);
+
+	/* Since it uses offset technique, assert that input matrix dim
```
karlnapf: "we need space for coefficients and offset"

src/shogun/features/RandomFourierDotFeatures.cpp
```diff
@@ -83,17 +83,21 @@ float64_t RandomFourierDotFeatures::post_dot(float64_t dot_result, index_t par_i
...
-SGVector<float64_t> RandomFourierDotFeatures::generate_random_parameter_vector()
+void RandomFourierDotFeatures::generate_random_parameter_vector(SGVector<float64_t>&& vec)
 {
 	NormalDistribution<float64_t> normal_dist;
 	UniformRealDistribution<float64_t> uniform_real_dist(0.0, 2 * Math::PI);
-	SGVector<float64_t> vec(feats->get_dim_feature_space()+1);
+
+	/* Since it uses offset technique, assert that input matrix dim 
+	feature space + 1 is equal to the length of vector */
+	ASSERT(vec.vlen == feats->get_dim_feature_space() + 1);
```
karlnapf: use the method you added

src/shogun/features/RandomFourierDotFeatures.h
```diff
@@ -103,7 +103,13 @@ class RandomFourierDotFeatures : public RandomKitchenSinksDotFeatures
...
 	 * @return a random parameter vector
 	 */
-	virtual SGVector<float64_t> generate_random_parameter_vector();
+	virtual void generate_random_parameter_vector(SGVector<float64_t>&& vec);
+
+	/** obtain the dimensionality of the feature space of random coefficients
+	 *
+	 * @return dimensionality
+	 */
+	virtual int32_t get_dim_feature_space_coeff() const;
```
karlnapf: "get_num_parameters", and state in the doc that this is per sample

tests/unit/features/RandomFourierDotFeatures_unittest.cc
```diff
@@ -176,3 +176,30 @@ TEST(RandomFourierDotFeatures, add_to_dense_test)
...
+	auto cols = r_feats->get_random_coefficients().num_cols;
```
karlnapf: why calling it twice?

tests/unit/features/RandomFourierDotFeatures_unittest.cc
```diff
@@ -176,3 +176,30 @@ TEST(RandomFourierDotFeatures, add_to_dense_test)
 
 }
 
+TEST(RandomFourierDotFeatures, gauss_coeff_dim)
+{
+	int32_t num_dims = 50;
+	int32_t num_vecs = 5;
+	int32_t D = 100;
+
+	SGMatrix<int32_t> data(num_dims, num_vecs);
```
karlnapf: no need for this data. Just set to ones or zeros

tests/unit/features/RandomFourierDotFeatures_unittest.cc
karlnapf: I dont think you fixed an error here?

karlnapf: ok, again. You need to check the dimension of the **transformed** data. Not the coefficient dimensions. It is better to test interface input/output, not internal states.

tests/unit/features/RandomFourierDotFeatures_unittest.cc
jonpsy: I'll instantiate data with random constants as you said before, if that's not a problem?.

tests/unit/features/RandomFourierDotFeatures_unittest.cc
jonpsy: My bad, I took it for random_coef matrix. About my second comment, I want a unit test which generates random coefficients rather than have it predefined. So, in the added test I created a unit test which uses the said method and checks for the output dimensions, that won't be a problem no?

tests/unit/features/RandomFourierDotFeatures_unittest.cc
```diff
@@ -176,3 +176,58 @@ TEST(RandomFourierDotFeatures, add_to_dense_test)
 
 }
 
+TEST(RandomFourierDotFeatures, gauss_coeff_dim)
+{
+	int32_t num_dims = 50;
+	int32_t num_vecs = 5;
+	int32_t D = 100;
+
+	float64_t mat[] = {
```
karlnapf: what is this? use sgmatrix, and set it to const values. **The dimensions of the projection dont depend on the content**

tests/unit/features/RandomFourierDotFeatures_unittest.cc
```diff
@@ -176,3 +176,58 @@ TEST(RandomFourierDotFeatures, add_to_dense_test)
...
+	auto d_feats = std::make_shared<DenseFeatures<float64_t>>(data);
+	SGVector<float64_t> params(1);
+	params[0] = 8;
+	auto r_feats = std::make_shared<RandomFourierDotFeatures>(
+			d_feats, D, GAUSSIAN, params);
+
+	auto rows = r_feats->get_dim_feature_space();
+	auto cols = r_feats->get_num_vectors();
+
+	EXPECT_EQ(rows,D);
```
karlnapf: yes this is what we want to test, good

tests/unit/features/RandomFourierDotFeatures_unittest.cc
```diff
@@ -176,3 +176,58 @@ TEST(RandomFourierDotFeatures, add_to_dense_test)
 
 }
 
+TEST(RandomFourierDotFeatures, gauss_coeff_dim)
+{
+	int32_t num_dims = 50;
+	int32_t num_vecs = 5;
+	int32_t D = 100;
```
karlnapf: could you name this num_dims_rff or something that is less confusing?

tests/unit/features/RandomFourierDotFeatures_unittest.cc
jonpsy: Right, all it needs is the num_dims of the matrix to construct the random coefficients matrix.

jonpsy:
I've noticed we're moving to use override in our repo. I've also noticed there is already a PR which is handling all these cases at once, shall I update my PR to use override or should I leave it be and let the other person handle it?

tests/unit/features/RandomFourierDotFeatures_unittest.cc
karlnapf: my point is we dont need any random data in here. That just adds noise to the code, and is not needed

karlnapf:
Don't worry about the override

jonpsy:
Good to merge, @karlnapf ?

karlnapf:
Let's do some more tidy up ... I'll comment

src/shogun/features/RandomFourierDotFeatures.cpp
```diff
@@ -83,11 +83,14 @@ float64_t RandomFourierDotFeatures::post_dot(float64_t dot_result, index_t par_i
 	return std::cos(dot_result) * constant;
 }
 
-SGVector<float64_t> RandomFourierDotFeatures::generate_random_parameter_vector()
+void RandomFourierDotFeatures::generate_random_parameter_vector(SGVector<float64_t>&& vec)
```
karlnapf: do we need this to be rvalue?

@gf712 ?

karlnapf: let's rename this to "generate_random_params"

src/shogun/features/RandomFourierDotFeatures.cpp
```diff
@@ -103,7 +106,11 @@ SGVector<float64_t> RandomFourierDotFeatures::generate_random_parameter_vector()
 		default:
 			error("Unknown kernel");
 	}
-	return vec;
+}
+
+int32_t RandomFourierDotFeatures::get_num_params() const
+{
+	return feats->get_dim_feature_space() + 1;
```
karlnapf, line 109: add a comment that this is number of features plus an offset

src/shogun/features/RandomFourierDotFeatures.h
```diff
@@ -103,7 +103,13 @@ class RandomFourierDotFeatures : public RandomKitchenSinksDotFeatures
 	 *
 	 * @return a random parameter vector
 	 */
-	virtual SGVector<float64_t> generate_random_parameter_vector();
+	virtual void generate_random_parameter_vector(SGVector<float64_t>&& vec);
+
+	/**  Get number of parameters per sample in the random_coeff matrix
```
karlnapf: dont reference variables in comments.
Just "@return number of parameters per input dimension"

tests/unit/features/RandomFourierDotFeatures_unittest.cc
```diff
@@ -176,3 +176,24 @@ TEST(RandomFourierDotFeatures, add_to_dense_test)
...
+	SGVector<float64_t> params(1);
+	params[0] = 8;
+	auto r_feats = std::make_shared<RandomFourierDotFeatures>(
+			d_feats, D, GAUSSIAN, params);
+
+	auto rows = r_feats->get_dim_feature_space();
+	auto cols = r_feats->get_num_vectors();
+
+	EXPECT_EQ(rows,D);
+	EXPECT_EQ(cols,num_vecs);
```
karlnapf: minor: we use a space after comas...(some other places as well)

src/shogun/features/RandomFourierDotFeatures.cpp
gf712: no, should be lvalue ref. rvalue ref implies that you are transferring ownership. In this case you won't see a difference, but this is bad practice, as the caller should now assume that the parameter `vec` has transferred it's data ownership to this function, which I don't think is the case? All you're doing here is mutate `vec`, right?

jonpsy: About mutating, you're correct, we're only feeding values to  vec inside the method.

jonpsy: @gf712 but get_column would return a temporary value we'll get "cannot bind non const lvalue to rvalue", if we take const lvalue then we can't modify it. So isn't && the only way?

gf712: I see. Imo you should then store get_column vector in a variable and then pass that to this function.

jonpsy: Rather, make it pass-by-value?

karlnapf: no, reference is better (this will be called in a loop) so as gil said above

jonpsy:
A gentle nudge, is this mergeable? @karlnapf

src/shogun/features/RandomKitchenSinksDotFeatures.h
```diff
@@ -224,7 +224,13 @@ class RandomKitchenSinksDotFeatures : public RandomMixin<DotFeatures>
 	 *
 	 * @return a random parameter vector
 	 */
-	virtual SGVector<float64_t> generate_random_parameter_vector()=0;
+	virtual void generate_random_params(SGVector<float64_t>& vec) = 0;
+	
+	/**  Get number of parameters per sample in the random_coeff matrix
```
karlnapf: minor. simplify to one line
"@return number of parameters per random feature dimension"

tests/unit/features/RandomFourierDotFeatures_unittest.cc
karlnapf: why resolved?

src/shogun/features/RandomFourierDotFeatures.cpp
```diff
@@ -83,11 +83,14 @@ float64_t RandomFourierDotFeatures::post_dot(float64_t dot_result, index_t par_i
...
 }
 
-SGVector<float64_t> RandomFourierDotFeatures::generate_random_parameter_vector()
+void RandomFourierDotFeatures::generate_random_params(SGVector<float64_t>& vec)
 {
 	NormalDistribution<float64_t> normal_dist;
 	UniformRealDistribution<float64_t> uniform_real_dist(0.0, 2 * Math::PI);
-	SGVector<float64_t> vec(feats->get_dim_feature_space()+1);
+
+	/* we need space for both coefficients and offset */
```
karlnapf: remove, or write something that is appropriate "Ensure passed vector has appropriate size"

src/shogun/features/RandomKitchenSinksDotFeatures.h
karlnapf: actually the superclass has a good comment, so this can just be removed

jonpsy: Sorry, I think you meant inside the RandomFourierDotFeatures.h ?

karlnapf: sorry it is not the superclass. Yes it should be the same comment as in RandomFourierDotFeatures

jonpsy: Mind if I shift it from fourier.h -> sinks.h? (for both generate_random_parms and get_num_params)

karlnapf: what do you mean by "shift"?

jonpsy: remove from fourier, put it inside sinks

karlnapf: I dont understand :D

jonpsy: In RandomFourierDotFeatures.h, we have protected methods ```generate_random_params``` and ```get_num_params``` right?. I suggest removing their docs because it is already documented in RandomKitchenSinks.h.

karlnapf: I see, yes. That's what I wrote above: if a superclass has a good docstring for a purely virtual method it is not always necessary to re-state that in the subclass. doxygen already place the superclass docstrings in subclasses, see #4258

tests/unit/features/RandomFourierDotFeatures_unittest.cc
karlnapf: this is still not resolved. Pls just don't mark conversations as resolved....it doesn't really help

jonpsy: Really sorry!. By the way, I changed the name to ```num_dims_embedding``` I guess that wouldn't do? 

EDIT: nvm, num_dims_rff it is!

src/shogun/features/RandomKitchenSinksDotFeatures.h
jonpsy: Okay, done shifting. Thanks for the details :) , resolving this..?

jonpsy:
Hey, I've also did some minor polish
- Remove inherited docs
- Minor grammar fix
- Move get_name() from .cpp to .h

src/shogun/features/RandomKitchenSinksDotFeatures.h
karlnapf: I actually prefer to not resolve anything as that hides our interactions. Ppl can hide conversations if they want

src/shogun/features/RandomFourierDotFeatures.cpp
karlnapf: @NanuSai in my opinion this should be next (after we have merged this)

tests/unit/features/RandomFourierDotFeatures_unittest.cc
karlnapf: +1 But here you have still "D"

jonpsy: problem, every subtest in this RFF_unittest uses D for "dim_output" shall I change for all?

karlnapf: yes I think it is a good idea

karlnapf: with "yes" I meant "change all occurences" ... not.`dim_output`, which is not a good name, I suggested `dims_rff` or `num_dims_rff` above

jonpsy: Right, misunderstanding from my part. So I'll name the test "dim_output"  and the "D" variable as "num_dims_rff". All is well?

karlnapf:
:tada:

Issue #4954, opened 2020-03-22
Fix Eigen::Map to set the correct memory alignment


vigsterkr:
there are many places where we use Eigen::Map for wrapping SGVector and SGMatrix in the codebase. These Maps are currently not really constructed the best way, as both SGVector and SGMatrix memory chunk is 16 bytes aligned, but this is not specified to Eigen.

for example
```
Map<MatrixXd> eigen_cov(cov.matrix, cov.num_rows, cov.num_cols);
```

should be actually:
```
Map<MatrixXd, Eigen::Aligned16> eigen_cov(cov.matrix, cov.num_rows, cov.num_cols);
```

or better yet it'd be better to port all those Map definitions to use the
```
operator EigenMatrixXtMap()
```
(or vector) as in that case if we change the alignment, then it could be changed in one part of the code instead of keep changing it all over again.

see linalg codebase how one can use the EigenMatrixXtMap/EigenVectorXtMap

gf712:
I approve :+1:

vigsterkr:
https://github.com/shogun-toolbox/shogun/commit/8fb6c23865f7edae90924e8efa6f3fba8add8b3a fixes the mapping definition for alignment

YuriYumeYuu:
I would like to contribute to shogun. Is this issue still open or is there some work left?

karlnapf:
if you had checked the above reference yourself, you would have seen that somebody is actively doing this atm

PR #4956, opened 2020-03-24
Eigen Fix

Abdelrahman350:
Hello, I have made a fix required in this issue https://github.com/shogun-toolbox/shogun/issues/4954 
I hope this helps.


karlnapf (commented):
What about using the operator that viktor suggested in the issue? Then these settings are all at one global place ....

karlnapf:
please dont close / reopen PRs...it spams our mailboxes and burdens the CI

Abdelrahman350:
I am sorry, @karlnapf.

gf712:
@Abdelrahman350 do you know how many of these are left? I have 89 across 23 files. Not sure if you are using `grep` but if not that will make your life much easier. Would be good fix all of that in this PR to get CI working again. Thanks!

Abdelrahman350:
Yes, I am using grep already. Thank you for your advice.

gf712:
> Yes, I am using grep already. Thank you for your advice.

No worries, btw there might be more than that, because I only searched for `Eigen::Map`, but something like `Map<` should find them all with few, if any, false positives.

Abdelrahman350:
Yes, @gf712  I am aware of that and I am using `Map<` for searching.

Abdelrahman350:
I have a question here. I have the first 7 commits that are unnecessary and I want to remove them or drop them. Can anyone tell me how to drop those first 7 commits, please?. If it's not possible, please ignore them while reviewing and start from that commit `79859eb`. Thanks
@gf712 @karlnapf

gf712:
@Abdelrahman350 if you want to "remove" commits you can use `git revert`. Have a look online on how to use it.

karlnapf:
No need to drop anything, just keep on adding commits to the PR. We will squash all of them into one when we merge it. The readme for developing also has info on that

vigsterkr:
you would have just needed to rebase... the problem that this is now not having all the reverted stuff that i was forced to do...

Abdelrahman350:
Ok, thank you very much.

> you would have just needed to rebase... the problem that this is now not having all the reverted stuff that i was forced to do...

So, is there something else I should do?

vigsterkr:
rebase over latest develop and dont bring back the alignment changes with your own commits

Abdelrahman350:
> rebase over latest develop and dont bring back the alignment changes with your own commits

what I understand that you want me to rebase my branch over the latest develop version, am I right?
I don't know what do you mean by alignment changes I used the same alignment as it was before my commits and didn't change it. I use the default text editor of Ubuntu.

I also noticed that you changed the `EigenMatrixXtMap` definition I compiled my changes after pulling your commits and everything works fine.

Abdelrahman350:
Ok Mr @vigsterkr  I made the rebase you told me to do, thanks for your help. I hope I did it is right.

src/shogun/distributions/classical/GaussianDistribution.cpp
```diff
@@ -142,7 +138,7 @@ SGVector<float64_t> GaussianDistribution::log_pdf_multiple(SGMatrix<float64_t> s
 
 	/* one quadratic part x^T C^-1 x for each sample x */
 	SGVector<float64_t> result(num_samples);
-	Map<VectorXd> eigen_result(result.vector, result.vlen);
+	SGVector<double>::EigenVectorXtMap eigen_result = result;
```
karlnapf, line 141: @vigsterkr could we have a templated function that does this so it is a bit less redundant?
`auto eigen_result = result.eigen()` or `auto eigen_result = eigen(result)`

karlnapf: @gf712 what do you think about that?

gf712: @karlnapf I actually had a similar discussion with @vigsterkr. This would then potentially enable us to return the correctly aligned Eigen::Map, i.e. 16/32/64 byte boundary. But then again, with the graph stuff this might be not that urgent, as there all the alignment will happen behind the scences at runtime.
Anyway, should be possible to have a `constexpr EigenVectorXtMap SGVector::eigen()`?

karlnapf: It is not urgent, but as people are doing this, we can at least make the code a tiny bit less verbose?
@Abdelrahman350 it would be cool if you added such a method to SGVector/SGMatrix

Abdelrahman350: Just to make it clear @karlnapf . What should the method do exactly?

vigsterkr: @karlnapf i have to check whether this could be somehow be made to auto-kick-in for functions where the arg is that... the nice thing is that this is super implicit :) i see the point of you wanting to have less verbose/less boilerplate code. but adding that function that you are mentioning actually adding redundancy to the SGVector codebase as you'll have 2 functions that are doing the very same thing :P

karlnapf: @Abdelrahman350 it would save us from writing `SGVector<double>::EigenVectorXtMap` every time, the function would return the mapped type.
@vigsterkr yes automatic would be best of course :) But we actually don't have EigenMap arguments in any functions, usually these maps are created inside functions that take sgvector as arguments.... I mean it was just an idea. Some functions have like 10 such maps in them and writing `SGVector<double>::EigenVectorXtMap` 10 times takes quite some space. Maybe it would be a templated util function instead of an sgvector method....not sure

Abdelrahman350: Ok, @karlnapf I have an idea to implement a function `auto eigen = globalMapping(m)` and use function overloading to do the same issue with vectors also. 
`SGVector<double> globalMapping(SGVector<float64_t> result){    SGVector<double>::EigenVectorXtMap eigen_result = result 
}`
Do you want me to proceed with this idea?

karlnapf: give it a try and we can have a look :)

PR #4957, opened 2020-03-24
packet math

gf712:
here is the idea:
 - we loop through data using packets (store + load)
 - dispatch the packets to use the correct arch dependent operation implementation (e.g. `_mm256_add_ps`)
 - dispatching is done with void* to avoid implementations knowing about other architecture specific implementations

This allows us to then do optimisations by merging various ops and perform computations whilst the data is still in the registers, e.g. `c = a + a - b;` would become (more or less) `load a, load b, add, sub, store c, loop`, rather than `load a, add, store tmp, (same thing happening in the loop), load tmp, load b, sub, store c`. So optimisation is done at the register level rather than vector/matrix/tensor, which is much more efficient in particular with large matrices that are larger than L3 cache.

We can also reuse the eigen implementations, the issue is that they only support part of the simd types (for example there is no int16), but we can add those.

This is what Eigen does, but at compile time. We are moving it to runtime. Then the more complex ops, like dot, can still be handled by Eigen, compiled for the specifc arch target.


src/shogun/mathematics/graph/CMakeLists.txt
```diff
@@ -230,6 +230,7 @@ target_link_libraries(shogun-vanilla-executor shogun-vanilla-executor-sse41)
 target_link_libraries(shogun-vanilla-executor shogun-vanilla-executor-avx)
 target_link_libraries(shogun-vanilla-executor shogun-vanilla-executor-avx512f)
 
+target_compile_options(shogun-vanilla-executor PRIVATE -mavx)
```
gf712: temporary, just to make it work on my machine

src/shogun/mathematics/graph/ops/shogun/Add.h
```diff
@@ -10,6 +10,7 @@
 #include <shogun/mathematics/graph/nodes/Add.h>
 #include <shogun/mathematics/graph/ops/abstract/BinaryOperator.h>
 #include <shogun/mathematics/graph/CPUArch.h>
+#include <immintrin.h>
```
gf712: temporary, this would be in a file specialised for loading/storing __m256 types

src/shogun/mathematics/graph/ops/shogun/Add.h
```diff
@@ -55,7 +56,35 @@ namespace shogun
...
+										static_cast<T*>(output)+i);
+								}
+							}
+						}
+						while(i < size)
+						{
+							*(static_cast<T*>(output)+i) = *(static_cast<const T*>(input1)+i) + *(static_cast<const T*>(input2)+i); 
+							i++;
+						}
+					}
```
gf712, line 88: this logic would be done by an executor

src/shogun/mathematics/graph/ops/shogun/Add.h
```diff
@@ -55,7 +56,35 @@ namespace shogun
...
+						size_t i = 0;
+						if (size > 32/sizeof(T))
+						{
+							size_t remainder = size % (32/sizeof(T));
+							for (;i<size-remainder; i+=32/sizeof(T))
+							{
+								if constexpr(std::is_same_v<float, T>)
+								{
+									const auto packet1 = _mm256_loadu_ps(static_cast<const T*>(input1)+i);
+									const auto packet2 = _mm256_loadu_ps(static_cast<const T*>(input2)+i);
```
gf712: this would be done by the packet implementation

src/shogun/mathematics/graph/ops/shogun/details/add_avx.cpp
```diff
@@ -4,43 +4,42 @@
...
-	    void* input1, void* input2, void* output, const size_t size)
+	    const T2* input1, const T2* input2, T2* output)
 	{
-		std::transform(
-		    static_cast<const bool*>(input1),
-		    static_cast<const bool*>(input1) + size,
-		    static_cast<const bool*>(input2), static_cast<bool*>(output),
-		    std::plus<bool>());
+		if constexpr (std::is_same_v<T1, float>)
+			*static_cast<__m256*>(output) = _mm256_add_ps(*static_cast<const __m256*>(input1), *static_cast<const __m256*>(input2));
```
gf712: the actual packet math

src/shogun/mathematics/graph/ops/shogun/details/add_avx.cpp
```diff
@@ -4,43 +4,42 @@
...
+		else
+		{
+			std::transform(
+			    static_cast<const T2*>(input1),
+			    static_cast<const T2*>(input1) + 32/sizeof(T2),
+			    static_cast<const T2*>(input2), static_cast<T2*>(output),
+			    std::plus<T2>());
```
gf712: if there is no support for loading/storing the packet carries a void* and the compiler decides how to implement the op using something like std::transform

src/shogun/mathematics/graph/ops/shogun/Add.h
gf712: This could have been something like `Eigen::internal::ploadu<Packet8f>(static_cast<const T*>(input1)+i)`, if we want to use the eigen implementations

src/shogun/mathematics/graph/ops/shogun/Add.h
```diff
@@ -55,7 +56,35 @@ namespace shogun
...
+								}
+								else
+								{
+									add_kernel_implementation_avx<T>(static_cast<const T*>(input1)+i, 
+										static_cast<const T*>(input2)+i, 
+										static_cast<T*>(output)+i);
+								}
+							}
+						}
+						while(i < size)
```
gf712: this should be manually unrolled with some sort of dispatching `unroll_op<T, remainder>(func, input1, input2, output)`

src/shogun/mathematics/graph/ops/shogun/Add.h
gf712: and then can use either the Eigen threadpool or our own using C++17 if the user wants multithreading

src/shogun/mathematics/graph/ops/shogun/Add.h
```diff
@@ -55,7 +56,35 @@ namespace shogun
 					if (CPU_arch->has_avx512f())
 						add_kernel_implementation_avx512f<T>(input1, input2, output, size);
 					else if (CPU_arch->has_avx())
-						add_kernel_implementation_avx<T>(input1, input2, output, size);
+					{
+						size_t i = 0;
+						if (size > 32/sizeof(T))
+						{
+							size_t remainder = size % (32/sizeof(T));
+							for (;i<size-remainder; i+=32/sizeof(T))
```
gf712: `32` because it's the width in bytes of a YMM avx register

gf712:
@vigsterkr: we would then have a `kernel_implementation_packet` for computing packets (this is then used by the optimiser to merge ops) and `kernel_implementation_tensor` for computing whole tensors in one go.
The idea with all this is that we don't rely on the prefetcher to do what we intend to do..

src/shogun/mathematics/graph/ops/shogun/Add.h
```diff
@@ -10,6 +10,7 @@
 #include <shogun/mathematics/graph/nodes/Add.h>
 #include <shogun/mathematics/graph/ops/abstract/BinaryOperator.h>
 #include <shogun/mathematics/graph/CPUArch.h>
+#include <Eigen/Core>
```
gf712: here to make things work

src/shogun/mathematics/graph/ops/shogun/Add.h
karlnapf: probably good to put a comment in the source for that?

src/shogun/mathematics/graph/ops/shogun/details/add_avx.cpp
```diff
@@ -4,43 +4,42 @@
...
-	template void add_kernel_implementation_avx<int32_t>(void*, void*, void*, const size_t);
-	template void add_kernel_implementation_avx<int64_t>(void*, void*, void*, const size_t);
-	template void add_kernel_implementation_avx<uint8_t>(void*, void*, void*, const size_t);
-	template void add_kernel_implementation_avx<uint16_t>(void*, void*, void*, const size_t);
-	template void add_kernel_implementation_avx<uint32_t>(void*, void*, void*, const size_t);
-	template void add_kernel_implementation_avx<uint64_t>(void*, void*, void*, const size_t);
-	template void add_kernel_implementation_avx<float>(void*, void*, void*, const size_t);
-	template void add_kernel_implementation_avx<double>(void*, void*, void*, const size_t);
+	template void add_kernel_implementation_avx<bool>(const bool*, const bool*, bool*);
+	template void add_kernel_implementation_avx<int8_t>(const int8_t*, const int8_t*, int8_t*);
```
karlnapf: is it possible to automate those using a list of types?

src/shogun/mathematics/graph/ops/shogun/Add.h
gf712: this logic would be moved to an executor, this is just PoC for now :) but yes, I will also replace these things with some Eigen definitions

src/shogun/mathematics/graph/ops/shogun/details/add_avx.cpp
gf712: the easiest would be to just put a macro that instantiates all the types

src/shogun/mathematics/graph/ops/shogun/Add.h
```diff
@@ -55,7 +56,35 @@ namespace shogun
...
+						size_t i = 0;
+						if (size > 32/sizeof(T))
+						{
+							size_t remainder = size % (32/sizeof(T));
+							for (;i<size-remainder; i+=32/sizeof(T))
+							{
+								if constexpr(std::is_same_v<float, T>)
+								{
+									const auto packet1 = Eigen::internal::ploadu<Eigen::internal::Packet8f>(static_cast<const T*>(input1)+i);
+									const auto packet2 = Eigen::internal::ploadu<Eigen::internal::Packet8f>(static_cast<const T*>(input2)+i);
```
gf712: `ploadu` would be `pload` if we can guarantee that memory is aligned. Right now we do 16 byte alignment which wouldn't work here, AVX requires 32-byte

src/shogun/mathematics/graph/ops/shogun/Packet.h
```diff
@@ -0,0 +1,97 @@
...
+			// void store_avx512f(const aligned_vector& input1, void* output);
+
+			enum class RegisterType
+			{
+				SSE = 16,
+				AVX = 32,
+				AVX512 = 64
+			};
+
+			struct Packet
```
gf712: @karlnapf basically this will be the abstraction of the data on the stack. And then the respective implementation (depending on CPU architecture) is called and gets the value from the variant. This actually works on my laptop...

src/shogun/mathematics/graph/ops/shogun/details/add_avx.cpp
```diff
@@ -4,43 +4,54 @@
...
+		void* input1, void* input2, void* output)
 	{
-		std::transform(
-		    static_cast<const bool*>(input1),
-		    static_cast<const bool*>(input1) + size,
-		    static_cast<const bool*>(input2), static_cast<bool*>(output),
-		    std::plus<bool>());
+		static_cast<Packet*>(output)->m_data = padd(
+			std::get<Packet8f>(static_cast<const Packet*>(input1)->m_data), 
+			std::get<Packet8f>(static_cast<const Packet*>(input2)->m_data));
```
gf712, line 45: @karlnapf @vigsterkr the packet data lives on the stack because the variant is basically a union so it can store data on stack instead of heap. This means that we can "type erase" __m128 and so on.. :)

src/shogun/mathematics/graph/ops/shogun/details/load_avx.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+using namespace Eigen::internal;
+
+namespace shogun::graph::op {
+			
+	template <typename T>
+	aligned_vector load_avx(void* input1);
+
+	template <typename T>
+	aligned_vector load_avx(void* data)
```
gf712, line 18: the only detail missing is here. I am not 100% sure that msvc will allow us to send data back and forth using variant... At worst I can cast it to void* as I did with all the rest

src/shogun/mathematics/graph/ops/shogun/Add.h
```diff
@@ -54,8 +59,66 @@ namespace shogun
 					auto* CPU_arch = CPUArch::instance();
 					if (CPU_arch->has_avx512f())
 						add_kernel_implementation_avx512f<T>(input1, input2, output, size);
+					else if (CPU_arch->has_avx2())
```
gf712, line 75: @vigsterkr so this isn't sustainable, but we can easily add so many ops that Eigen doesn't support, for example here I added addition for int32 using AVX2 instructions. The problem is just that we need to check for specific type here, i.e. there should be something in the condition checking if avx2 is available and that the type is int32. The float versions are available in AVX (which I guess is safe to assume is a subset of AVX2)

src/shogun/mathematics/graph/ops/shogun/details/add_avx2.cpp
```diff
@@ -0,0 +1,76 @@
...
+			std::get<Packet4d>(static_cast<const Packet*>(input2)->m_data));
+	}
+
+	template<>
+	void add_kernel_implementation_avx<int32_t>(
+		void* input1, void* input2, void* output)
+	{
+		auto p1 = std::get<Packet8i>(static_cast<const Packet*>(input1)->m_data);
+		auto p2 = std::get<Packet8i>(static_cast<const Packet*>(input2)->m_data);
+		static_cast<Packet*>(output)->m_data = _mm256_add_epi32(p1, p2);
```
gf712, line 72: @vigsterkr  @karlnapf  not available in Eigen! can add also uint8, int8, uint16, int16 and int64

src/shogun/mathematics/graph/ops/shogun/details/add_avx.cpp
```diff
@@ -4,43 +4,82 @@
...
-		Out.device(Eigen::DefaultDevice{}) = A + B;
+		std::transform(
+		    reinterpret_cast<const T*>(&vec1),
+		    reinterpret_cast<const T*>(&vec1) + AVX_BYTESIZE/sizeof(T),
+		    reinterpret_cast<const T*>(&vec2),
+		    reinterpret_cast<T*>(&result),
+		    std::plus<T>());
+
+		static_cast<Packet*>(output)->m_data = result;
 	}
```
gf712, line 37: I have no idea how the compiler/runtime manage to get this to work... there are so many code smells here but this is a way to generalise integer math when there are no intrinsics functions

src/shogun/mathematics/graph/ops/shogun/details/add_avx.cpp
```diff
@@ -4,43 +4,82 @@
...
-		    static_cast<const bool*>(input1),
-		    static_cast<const bool*>(input1) + size,
-		    static_cast<const bool*>(input2), static_cast<bool*>(output),
-		    std::plus<bool>());
+		const auto& vec1 = std::get<bool*>(static_cast<const Packet*>(input1)->m_data);
+		const auto& vec2 = std::get<bool*>(static_cast<const Packet*>(input2)->m_data);
+		auto& out = std::get<bool*>(static_cast<const Packet*>(output)->m_data);
+
+		std::transform(vec1, vec1 + AVX_BYTESIZE/sizeof(bool), vec2, out, std::plus<bool>());
+	}
```
gf712: I'll work on a more efficient boolean version with bitsets at some point

gf712:
Got avx2 and sse working locally, can't test avx512 on my personal laptop, but should work..
Next I will try and abstract away the execution loop

gf712:
Now the abstraction is at the BinaryOperator level, which mean that the graph should be able to see that there is a packet implementation and that it can merge some ops. Working on an executor that evaluates the subexpressions to avoid direct communication between graph and operator (as being done now)

Issue #1139, opened 2013-06-03
Remove all Math-based code from SGMatrix/SGVector


karlnapf:
Since our matrix and vector classes are intended to be just wrappers for different languages rather than a representation of matrix/vector-like objects, they should not contain mathematical operations like mean/scale/zeros etc.

In addition, CStatistics and CMath contain a lot of mathematical operations on matrix/vector objects.

Instead, we should use eigen3 for such tasks. SGMatrix and SGVector should contain a method that returned a mapped eigen3 object, which then can be used to compute things.

This will simply our code and prevent us from re-inventing the wheel many times.
Can be done one after another.

PR #4958, opened 2020-03-24
remove add and operator+ from SGVector and replace with linalg::add calls

Hephaestus12:
remove `add` from SGVector
remove operator`+` from SGVector
remove operator`+=` from SGVector
refactor code to call `linalg::add` method
#2747
#1139
#2582

Code compiles successfully
While running `make test`, a few unit tests failed but they were failing on the original code(before making these changes) as well.


karlnapf:
LGTM

Hephaestus12:
> LGTM

So, are there any other changes I should make before this gets merged?

src/shogun/lib/SGVector.h
```diff
@@ -389,23 +383,6 @@ template<class T> class SGVector : public SGReferencedData
...
-
-		/** Inplace addition operator */
-		SGVector<T> operator+= (SGVector<T> x)
-		{
-			add(x);
-			return *this;
-		}
-
-		/** Inplace addition operator for sparse vector */
-		SGVector<T> operator+= (SGSparseVector<T>& x)
```
karlnapf, line 403: btw is this defined in linalg?

gf712: good point. There is no specialisation for SGSparseVector in linalg..

Hephaestus12: Should I let this remain as it was then? And only replace the SGVector calls?

gf712: I think so

karlnapf: @gf712 If no compiler errors happen, I would just remove as unused?

gf712: I guess so. Just thinking if there is something hiding behind a `#ifdef`..

karlnapf: I say remove and fix the errors when they appear :) We want to clean up these classes in any case ...

gf712: OK, in that case can merge once the CI is fixed, just to be sure

karlnapf: one thing I am always concerned about is changing code that is not covered by any tests (at least integration) .... but in this case I think we can justify it?

gf712:
> So, are there any other changes I should make before this gets merged?

Could you force push so that the CI is triggered again? It is working again.

Hephaestus12:
Force pushed

gf712:
It seems like some calls require the std::complex version of `add`. For `add` it should be pretty straighforward as Eigen has an implementation for that. You have to go to `LinalgBackendBase.h` and in the macro where `add` is implemented, write the specialisation for SGVector<complex128_t>.
```cpp
// this is what you have
DEFINE_FOR_ALL_PTYPE(BACKEND_GENERIC_IN_PLACE_ADD, SGVector)
DEFINE_FOR_ALL_PTYPE(BACKEND_GENERIC_IN_PLACE_ADD, SGMatrix)
// this is what you add
BACKEND_GENERIC_IN_PLACE_ADD(std::complex128_t, SGVector);
BACKEND_GENERIC_IN_PLACE_ADD(std::complex128_t, SGMatrix);
```
And that should work. @karlnapf should probably have a more generic way of adding std::complex partial specialisation?

karlnapf:
no that would be the way with the current system (tm)

Hephaestus12:
@gf712 I understood what you are asking to be done.
However, in the file: `LinalgMacros.h` line 33, `DEFINE_FOR_ALL_PTYPE` has been defined and it does include the `complex128_t` instance.

`LinalgMacros.h` line 33:
```
#define DEFINE_FOR_ALL_PTYPE(METHODNAME, Container)                            \
	METHODNAME(bool, Container);                                               \
	METHODNAME(char, Container);                                               \
	METHODNAME(int8_t, Container);                                             \
	METHODNAME(uint8_t, Container);                                            \
	METHODNAME(int16_t, Container);                                            \
	METHODNAME(uint16_t, Container);                                           \
	METHODNAME(int32_t, Container);                                            \
	METHODNAME(uint32_t, Container);                                           \
	METHODNAME(int64_t, Container);                                            \
	METHODNAME(uint64_t, Container);                                           \
	METHODNAME(float32_t, Container);                                          \
	METHODNAME(float64_t, Container);                                          \
	METHODNAME(floatmax_t, Container);                                         \
	METHODNAME(complex128_t, Container);
```
Is the addition you asked for still necessary?

gf712:
sorry i meant for `LinalgBackendEigen.h`

gf712:
it has `DEFINE_FOR_NUMERIC_PTYPE`

Hephaestus12:
Can we rather just replace DEFINE_FOR_NUMERIC_PTYPE with DEFINE_FOR_ALL_PTYPE

Hephaestus12:
> It seems like some calls require the std::complex version of `add`. For `add` it should be pretty straighforward as Eigen has an implementation for that. You have to go to `LinalgBackendBase.h` and in the macro where `add` is implemented, write the specialisation for SGVector<complex128_t>.
> 
> ```c++
> // this is what you have
> DEFINE_FOR_ALL_PTYPE(BACKEND_GENERIC_IN_PLACE_ADD, SGVector)
> DEFINE_FOR_ALL_PTYPE(BACKEND_GENERIC_IN_PLACE_ADD, SGMatrix)
> // this is what you add
> BACKEND_GENERIC_IN_PLACE_ADD(std::complex128_t, SGVector);
> BACKEND_GENERIC_IN_PLACE_ADD(std::complex128_t, SGMatrix);
> ```
> 
> And that should work. @karlnapf should probably have a more generic way of adding std::complex partial specialisation?

Replacing this in LinalgBackendEigen.h is not compiling successfully.

gf712:
what's the error?

Hephaestus12:
> It seems like some calls require the std::complex version of `add`. For `add` it should be pretty straighforward as Eigen has an implementation for that. You have to go to `LinalgBackendBase.h` and in the macro where `add` is implemented, write the specialisation for SGVector<complex128_t>.
> 
> ```c++
> // this is what you have
> DEFINE_FOR_ALL_PTYPE(BACKEND_GENERIC_IN_PLACE_ADD, SGVector)
> DEFINE_FOR_ALL_PTYPE(BACKEND_GENERIC_IN_PLACE_ADD, SGMatrix)
> // this is what you add
> BACKEND_GENERIC_IN_PLACE_ADD(std::complex128_t, SGVector);
> BACKEND_GENERIC_IN_PLACE_ADD(std::complex128_t, SGMatrix);
> ```
> 
> And that should work. @karlnapf should probably have a more generic way of adding std::complex partial specialisation?

when I do this, I get the following error:
```
../../../src/shogun/libshogun.so.18.0: undefined reference to `shogun::LinalgBackendEigen::add(shogun::SGMatrix<std::complex<double> > const&, shogun::SGMatrix<std::complex<double> > const&, std::complex<double>, std::complex<double>, shogun::SGMatrix<std::complex<double> >&) const'
../../../src/shogun/libshogun.so.18.0: undefined reference to `shogun::LinalgBackendEigen::add(shogun::SGVector<std::complex<double> > const&, shogun::SGVector<std::complex<double> > const&, std::complex<double>, std::complex<double>, shogun::SGVector<std::complex<double> >&) const'
```


> Can we rather just replace DEFINE_FOR_NUMERIC_PTYPE with DEFINE_FOR_ALL_PTYPE

And when I do this, I get this error:
```
../../../src/shogun/libshogun.so.18.0: undefined reference to `vtable for shogun::LinalgBackendEigen'
```

gf712:
OK, can you just push what you have please?

Hephaestus12:
Pushed.

src/shogun/mathematics/linalg/LinalgBackendEigen.h
```diff
@@ -55,6 +55,8 @@ namespace shogun
 	    Type beta, Container<Type>& result) const;
 		DEFINE_FOR_NUMERIC_PTYPE(BACKEND_GENERIC_IN_PLACE_ADD, SGVector)
 		DEFINE_FOR_NUMERIC_PTYPE(BACKEND_GENERIC_IN_PLACE_ADD, SGMatrix)
+		BACKEND_GENERIC_IN_PLACE_ADD(complex128_t, SGVector);
+		BACKEND_GENERIC_IN_PLACE_ADD(complex128_t, SGMatrix);
```
gf712, line 59: The issue is that you only added this to the declaration. You have to add this to the implementation too, which is in `BasicOps.cpp`

gf712: And then let's see if Eigen let's you add complex numbers...

Hephaestus12:
Compiling successfully although a few tests are still failing

tests/unit/lib/SGVector_unittest.cc
```diff
@@ -194,14 +171,6 @@ TEST(SGVectorTest,complex128_tests)
...
 	SGVector<complex128_t> b=a.clone();
 
-	// test ::operator+ and []
-	a=a+b;
-	for (index_t i=0; i<a.vlen; ++i)
-	{
-		EXPECT_NEAR(a[i].real(), 10.0, 1E-14);
-		EXPECT_NEAR(a[i].imag(), 12.0, 1E-14);
-	}
-
```
gf712, line 204: you need to replace the call here right? `a=a+b` should be `a=linalg::add(a,b)`, otherwise you won't be getting the same results below...

gf712 (changes requested):
Almost there

tests/unit/lib/SGVector_unittest.cc
Hephaestus12: Since this is a test for the + operator, isn't it better to just remove the code?

gf712: I think this might be the only place left where we test complex numbers though... But yes ideally remove this and add complex128_t add test in linalg unittests

tests/unit/lib/SGVector_unittest.cc
```diff
@@ -193,9 +170,9 @@ TEST(SGVectorTest,complex128_tests)
 	SGVector<complex128_t> a(10);
 	a.set_const(complex128_t(5.0, 6.0));
 	SGVector<complex128_t> b=a.clone();
-
+	
 	// test ::operator+ and []
-	a=a+b;
+	a=linalg::add(a,b);
```
karlnapf: i think we would just remove the addition tests of SGVector ... linalg::add is tested already

gf712: But not for complex numbers though?

karlnapf: sure, but then we would have a test for linalg, and not sgvector, this is not testing sgvector::operator+ anymore, or am I wrong?

gf712: Indeed, that’s the other option I suggested above

Hephaestus12: So what do you suggest we do here?
Should I simply remove the test and open another issue?

gf712: Can you go to `Eigen3_operations_unittest.cc`, create a new class inheriting from ::testing::Test, then pass it to `SG_TYPED_TEST_CASE(..., sg_all_primitive_types, bool);` and use that new class in `TYPED_TEST` where we test for `add`. Then you can delete this SGVector test :)

gf712: And give that class a meaningful name like the other ones in `Eigen3_operations_unittest.cc` please!

tests/unit/lib/SGVector_unittest.cc
Hephaestus12: Which file should this test for complex addition be added though?
I see a test for the add function in the file : `tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc`
Should the complex test be added here?

gf712: see my comment [here](https://github.com/shogun-toolbox/shogun/pull/4958/files/daf8ae841f6bda16a2dd0732fa05398f3cab59ba#diff-c6e86f9deed123416d12ab100ec0c9e6), it explains all you need to do

tests/unit/lib/SGVector_unittest.cc
Hephaestus12: Yes I will do that. However, I have a doubt, Doesn't the existing code to that currently as well? reference
line 26 
```
template <typename T>
class LinalgBackendEigenAllTypesTest : public ::testing::Test
{
};
```

line 43
```
SG_TYPED_TEST_CASE(
    LinalgBackendEigenAllTypesTest, sg_all_primitive_types, bool, complex128_t);
```
line 50
```
TYPED_TEST(LinalgBackendEigenAllTypesTest, SGVector_add)
{
```

gf712: Unfortunately it doesn't do all the types like it says... You can see on line 43 that bool and complex128_t are excluded from the types being test

Hephaestus12:
A few other changes needed to be done which have been done to allow for complex numbers to be tested

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
```diff
@@ -64,8 +71,12 @@ TYPED_TEST(LinalgBackendEigenAllTypesTest, SGVector_add)
 	auto result = add(A, B, alpha, beta);
 
 	for (index_t i = 0; i < 9; ++i)
+	{
 		EXPECT_NEAR(
-		    alpha * A[i] + beta * B[i], result[i], get_epsilon<TypeParam>());
+	    	std::real(alpha * A[i] + beta * B[i]), std::real(result[i]), std::real(get_epsilon<TypeParam>()));
+		EXPECT_NEAR(
+	    	std::imag(alpha * A[i] + beta * B[i]), std::imag(result[i]), std::imag(get_epsilon<TypeParam>()));
+	}
```
gf712, line 79: @karlnapf is this OK with you?

tests/unit/lib/SGVector_unittest.cc
gf712: You have to remove the code below as it depends on the add you remove here. Either that or adapt the results.

Hephaestus12: I'll replace the 10.0 and 12.0 with 5.0 and 6.0 respectively.

gf712: thanks!

Hephaestus12:
Is this fine?

gf712:
> Is this fine?

Not yet. I mean look at the output of the CI.. Could you run this locally until the test passes? You are modifying the wrong test case.

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
karlnapf: yes lgtm

tests/unit/lib/SGVector_unittest.cc
```diff
@@ -98,29 +98,6 @@ TEST(SGVectorTest,setget)
...
-	for (int i=0; i < c.vlen; ++i)
-		EXPECT_EQ(c[i], a[i]+b[i]);
-
-	c = a + a;
-	EXPECT_EQ(c.vlen, 10);
-	for (int i=0; i < c.vlen; ++i)
-		EXPECT_EQ(c[i], 2*a[i]);
-}
-
 TEST(SGVectorTest,norm)
```
karlnapf, line 101: future note: the norm should also be removed :)

Hephaestus12:
@gf712 I rectified the tests. Now The SGVector unit test is successfully passing.
A few other tests are failing though( MeanSquaredError test and a few meta examples). Should that be my concern? Or should I push the code?

tests/unit/lib/SGVector_unittest.cc
Hephaestus12: Yes, @karlnapf  I shall get to that asap once this PR gets merged :)

gf712:
> @gf712 I rectified the tests. Now The SGVector unit test is successfully passing.
> A few other tests are failing though( MeanSquaredError test and a few meta examples). Should that be my concern? Or should I push the code?

Yup, push what you have then! Those errors are probably unrelated and build specific. I think the only way to get everything passing locally 100% of the time is to use the docker image. I think the only thing that was failing in your case was the SGVector test!

gf712:
Thanks!

PR #4961, opened 2020-03-25
Eigen_Fix

Abdelrahman350:
Hello, I added the EigenMatrixXtMap operator. If this is right I will do the same to the whole other matrices and vectors.


gf712:
could you push this to #4956 as it seems to be about the same? Only open one PR per fix please.

PR #4962, opened 2020-03-26
Change KMeans "mus" parameter 

### LINKED ISSUE #4847: Change KMeans "mus" parameter to something more user friendly

LiuYuHui:
fix #4847

### DIFF
--- a/doc/ipython-notebooks/clustering/KMeans.ipynb
+++ b/doc/ipython-notebooks/clustering/KMeans.ipynb
@@ -243,7 +243,7 @@
     "initial_centers = array([[0.,10.],[50.,50.]])\n",
     "\n",
     "# initial centers passed\n",
-    "kmeans = sg.machine(\"KMeans\", k=k, distance=distance, mus=initial_centers)"
+    "kmeans = sg.machine(\"KMeans\", k=k, distance=distance, initial_centers=initial_centers)"
    ]
   },
   {
@@ -288,7 +288,7 @@
     "new_initial_centers = array([[5.,5.],[0.,100.]])\n",
     "\n",
     "# set new initial centers\n",
-    "kmeans.put(\"mus\", new_initial_centers)"
+    "kmeans.put(\"initial_centers\", new_initial_centers)"
    ]
   },
   {
--- a/doc/ipython-notebooks/clustering/KMeans.ipynb
+++ b/doc/ipython-notebooks/clustering/KMeans.ipynb
@@ -288,7 +288,7 @@
     "new_initial_centers = array([[5.,5.],[0.,100.]])\n",
     "\n",
     "# set new initial centers\n",
-    "kmeans.put(\"initial_centers\", new_initial_centers)"
+    "kmeans.put(\" initial_centers\", new_initial_centers)"
    ]
   },
   {
--- a/doc/ipython-notebooks/clustering/KMeans.ipynb
+++ b/doc/ipython-notebooks/clustering/KMeans.ipynb
@@ -288,7 +288,7 @@
     "new_initial_centers = array([[5.,5.],[0.,100.]])\n",
     "\n",
     "# set new initial centers\n",
-    "kmeans.put(\" initial_centers\", new_initial_centers)"
+    "kmeans.put(\"initial_centers\", new_initial_centers)"
    ]
   },
   {
--- a/src/shogun/clustering/KMeans.cpp
+++ b/src/shogun/clustering/KMeans.cpp
@@ -133,9 +133,7 @@ void KMeans::Lloyd_KMeans(SGMatrix<float64_t> centers, int32_t num_centers)
 					}
 					else
 					{
-						/*  cluster_centers(:,j)=zeros(dim,1) ; */
-						for (j=0; j<dim; j++)
-							centers(j, cluster_assignments_i)=0;
+						centers.get_column(cluster_assignments_i).zero();
 					}
 
 				}
--- a/src/shogun/clustering/KMeans.cpp
+++ b/src/shogun/clustering/KMeans.cpp
@@ -149,7 +149,6 @@ void KMeans::Lloyd_KMeans(SGMatrix<float64_t> centers, int32_t num_centers)
 		/* Update Step : Calculate new means */
 		if (!fixed_centers)
 		{
-			/* cluster_centers=zeros(dim, num_centers) ; */
 			centers.zero();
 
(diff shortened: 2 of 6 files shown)


src/shogun/clustering/KMeansBase.cpp
```diff
@@ -321,7 +320,9 @@ void KMeansBase::init()
 	SG_ADD(
 	    &use_kmeanspp, "kmeanspp", "Whether to use kmeans++",
 	    ParameterProperties::HYPER | ParameterProperties::SETTING);
-	SG_ADD(&mus, "mus", "Cluster centers", ParameterProperties::MODEL);
+	SG_ADD(
+	    &cluster_centers, "cluster_centers", "Cluster centers",
+	    ParameterProperties::MODEL);
 
-	watch_method("cluster_centers", &KMeansBase::get_cluster_centers);
```
gf712, line 326: I would replace the SG_ADD above with the watch_method. And then you should have another SG_ADD for initial cluster centres.

gf712 (commented):
We need to find a way to initialise KMeans with clusters centres provided by the user using something like `.put("initial_clusters")`. Then when KMeans runs it will either initialise cluster centres itself or use the user provided values.

src/shogun/clustering/KMeans.cpp
```diff
@@ -133,7 +133,7 @@ void KMeans::Lloyd_KMeans(SGMatrix<float64_t> centers, int32_t num_centers)
 					}
 					else
 					{
-						/*  mus(:,j)=zeros(dim,1) ; */
+						/*  cluster_centers(:,j)=zeros(dim,1) ; */
```
karlnapf: Id suggest to remove this comment as this is evident from the code. Also, one could use a direct zero-ing of the memory here rather than a loop

src/shogun/clustering/KMeans.cpp
```diff
@@ -149,7 +149,7 @@ void KMeans::Lloyd_KMeans(SGMatrix<float64_t> centers, int32_t num_centers)
 		/* Update Step : Calculate new means */
 		if (!fixed_centers)
 		{
-			/* mus=zeros(dim, num_centers) ; */
+			/* cluster_centers=zeros(dim, num_centers) ; */
```
karlnapf: and here as well

src/shogun/clustering/KMeansBase.cpp
```diff
@@ -77,12 +76,12 @@ void KMeansBase::set_random_centers()
 		SGVector<float64_t> vec=lhs->get_feature_vector(cluster_center_i);
 
 		for (int32_t j=0; j<dimensions; j++)
-			mus(j,i)=vec[j];
+			cluster_centers(j, i) = vec[j];
```
karlnapf: no loop necessary here, maybe touching it now is an opportunity to either use sg_memset or add a function to linalg?

src/shogun/clustering/KMeansBase.cpp
```diff
@@ -163,18 +162,18 @@ void KMeansBase::initialize_training(const std::shared_ptr<Features>& data)
...
 	R=SGVector<float64_t>(k);
 
-	mus=SGMatrix<float64_t>(dimensions, k);
+	cluster_centers = SGMatrix<float64_t>(dimensions, k);
 	/* cluster_centers=zeros(dimensions, k) ; */
-	memset(mus.matrix, 0, sizeof(float64_t)*centers_size);
+	memset(cluster_centers.matrix, 0, sizeof(float64_t) * centers_size);
 
-	if (mus_initial.matrix)
+	if (initial_centers.matrix)
```
LiuYuHui, line 169: > We need to find a way to initialise KMeans with clusters centres provided by the user using something like .put("initial_clusters"). Then when KMeans runs it will either initialise cluster centres itself or use the user provided values.

@gf712 i guess this is what you want?

src/shogun/clustering/KMeansBase.cpp
```diff
@@ -321,7 +317,11 @@ void KMeansBase::init()
...
 	    ParameterProperties::HYPER | ParameterProperties::SETTING);
-	SG_ADD(&mus, "mus", "Cluster centers", ParameterProperties::MODEL);
-
 	watch_method("cluster_centers", &KMeansBase::get_cluster_centers);
+	SG_ADD(
+	    &initial_centers, "initial_centers", "Initial centers",
+	    ParameterProperties::HYPER);
+	// add this to check initial centers
+	add_callback_function(
+	    "initial_centers", [&]() { set_initial_centers(initial_centers); });
```
gf712, line 326: Yes, like this! cool :)

src/shogun/clustering/KMeans.cpp
gf712: yup, you could  `cluster_centers.get_column(cluster_assignments_i).zero()`

src/shogun/clustering/KMeansBase.cpp
gf712: yup, exactly, thanks!

src/shogun/clustering/KMeans.cpp
LiuYuHui: cool, i didn't find a good way to do this zero-ing.

gf712:
Could you trigger the CI now that it is working again?

gf712:
I think this is ready to be merged. @vigsterkr what do you think?

vigsterkr:
👍 lets wait till CIs are green and then lets merge it! \o/

doc/ipython-notebooks/clustering/KMeans.ipynb
```diff
@@ -288,7 +288,7 @@
     "new_initial_centers = array([[5.,5.],[0.,100.]])\n",
     "\n",
     "# set new initial centers\n",
-    "kmeans.put(\"mus\", new_initial_centers)"
+    "kmeans.put(\" initial_centers\", new_initial_centers)"
```
gf712:
```suggestion
    "kmeans.put(\"initial_centers\", new_initial_centers)"
```

gf712 (changes requested):
just saw this was causing issues in notebook

gf712:
Good work!

PR #4965, opened 2020-03-27
Added SKlearnvsSHOGUN on KRR

jonpsy:
A ipython notebook which benchmarks Sklearn vs Shogun and highlights our API. Does this suffice for our ipython notebook for "RandomFourierFeatures"? @karlnapf.


gf712:
Can you not get the data via the sklearn API or somewhere else rather than pushing it to the codebase?

jonpsy:
Hmmm, I can do that, I was just aiming for a large enough dataset so that we can properly see how skearn and Shogun KRR scales with data.

gf712:
I understand what you are trying to do, but you are forcing all users who clone shogun to download these datasets... Can you not just add some utility python function that requests the data and parse that? Basically like [this](https://scikit-learn.org/stable/modules/generated/sklearn.datasets.fetch_california_housing.html#sklearn.datasets.fetch_california_housing).

jonpsy:
I can see three ways to go by solving the problem (in order of my preference :D )
1. I can put the data in our $SHOGUN_DATA_DIR folder, users download it by default anyway.
2. I can use  some curl mechanism to download the dataset from kaggle
3. I could use sklearn dataset.

gf712:
Personally 2 is the best way, but maybe @vigsterkr or @karlnapf think otherwise?

jonpsy:
pandas offer quite slick way of directly taking data from url(2nd method). I can vey well use that, but I'll wait for their comments before I proceed with it.

vigsterkr:
i'd just use pandas for this... the thing is that the notebook needs a lot of work in the explanation part... and on the other hand i'm not so sure what is the exact aim here?
btw would be nice to have a notebook that has all the outputs as well....

jonpsy:
> i'd just use pandas for this... the thing is that the notebook needs a lot of work in the explanation part... and on the other hand i'm not so sure what is the exact aim here?
> btw would be nice to have a notebook that has all the outputs as well....

1.Since you too agree with the pandas method, I'll remove the data folder and directly make pandas read data from url.
2. I tried explaining the best I can, if possible could you point out which part needs more clearer/vivid explanation.
3. I didn't quite understand what you meant by "has all the outputs" can you help make me understand :D ?

vigsterkr:
3. some gist or some place where we (the reviewer) could see what are the outputs when you ran this notebook on your machine or somewhere....

jonpsy:
>     1. some gist or some place where we (the reviewer) could see what are the outputs when you ran this notebook on your machine or somewhere....

https://github.com/NanuSai/shogun-benchmarks/blob/master/notebooks/SKLEARN_KRR%20_vs_SHOGUN_KRR.ipynb 
Hope it helps :D

doc/Sklearn_docs/SKLEARN_KRR _vs_SHOGUN_KRR copy.ipynb
```diff
@@ -0,0 +1,401 @@
+{
```
karlnapf, line 1: I think this needs a different path.

I also wonder whether we really need to have a comparison inside the repository. I would either

* make a notebook that showcases Shogun's KRR (or better: add it to the existing KRR notebook with some explanations), inside the shogun repo.
* make a notebook that compares things as part of your application. For that you dont need to send a PR, but can rather just post the gist link in your application

vigsterkr:
TBH: i dont see a point of having a benchmarking comparison using jupyter notebook.

vigsterkr:
and i've already pointed you out that no matter what models with RMSE 15571108 is like... what's the point?

karlnapf:
thats some big RMSE :D

jonpsy:
We're training on a small subset of the data maybe that's why we're getting poor performance?

vigsterkr:
@NanuSai the thing is that with this RMSE it's probably possible to write a function that just randomly spits back some numbers... it'll be waaaaay faster than any trained model, but still as the outcome the whole thing gets to a moo points. i.e. what's the value of this 'exercise'/notebook? a) trained models are useless b) we have some benchmark numbers that are going via python's GIL...

doc/Sklearn_docs/SKLEARN_KRR _vs_SHOGUN_KRR copy.ipynb
jonpsy: In the proposal you've mentioned about filling gaps in our notebooks by a PR. So I went ahead and did this. Anyway, I can add the contents inside our "regression.ipynb" that's what you suggest right?.

karlnapf:
The regression notebook features KRR and has nice results and visualisations, you can add the nystrom krr there and describe it a bit, which would be useful. As mentioned above, if you want to do a comparison (which can be useful for your application), dont do it in a PR, and make sure to compare things that make sense as @vigsterkr said.
Also feel free to ping us in irc if you have more questions

doc/Sklearn_docs/SKLEARN_KRR _vs_SHOGUN_KRR copy.ipynb
karlnapf: yes, see below. Also before you do stuff, try to reflect for a second whether it is useful, and what it is for. That helps you focus your efforts.

jonpsy:
Very well, so shall we close this PR then? I can integrate some of my code to the "regression.ipynb" I guess. 
About the point of the exercise, I just wanted to compare shogun and sklearn on KRR so that we could see how our Shogun works in its python form and how it stands up against other libraries like sklearn. 
But I agree, the metrics are going all over the place.

PR #4968, opened 2020-03-27
Updated Convolutional Neural Network

Dynamo13:
Used keras.layers to add the hidden layers (Maxpooling2D and Convolutional layers) and flatten layer followed by dense layers to give the output .Finally adam optimiser was used with accuracy as metrics to evalute the model


karlnapf:
nope

PR #4973, opened 2020-03-28
update notebooks

gf712:
- Various fixes to make python notebooks work
- Added free functions to go around states of objects with new API
- Added `structured_model` and `factor_type` factories for StructuredModel
- Added CI job and script to run all notebooks (to replace the buildbot job)


karlnapf (commented):
This is grep right?
I think some of them will not run without errors in Py3, or did you run them?

gf712:
> This is grep right?
> I think some of them will not run without errors in Py3, or did you run them?

I fixed the notebook job on buildbot so I am running stuff until all the notebooks are functional. But yes, I just did `grep` and I am just running through the PR until it works...

karlnapf:
great initiative! :)

gf712:
> great initiative! :)

Thanks! I have slowly made through all the buildbot jobs... Hopefully all will be done soon and we can release everything working :)

src/shogun/regression/LeastAngleRegression.cpp
```diff
@@ -429,9 +429,6 @@ SGMatrix<ST> LeastAngleRegression::cholesky_delete(SGMatrix<ST>& R, int32_t i_ki
 	return nR;
 }
 
-template bool LeastAngleRegression::train_machine_templated<floatmax_t>(std::shared_ptr<DenseFeatures<floatmax_t>> data);
-template bool LeastAngleRegression::train_machine_templated<float64_t>(std::shared_ptr<DenseFeatures<float64_t>> data);
-template bool LeastAngleRegression::train_machine_templated<float32_t>(std::shared_ptr<DenseFeatures<float32_t>> data);
```
gf712, line 434: pretty sure these templates are instantiated by CRTP

doc/ipython-notebooks/neuralnets/neuralnets_digits.ipynb
```diff
@@ -118,18 +114,22 @@
...
+    "net_no_reg.put(\"seed\", 10)\n",
     "net_no_reg.quick_connect()\n",
     "net_no_reg.initialize_neural_network()\n",
     "\n",
     "net_l2 = NeuralNetwork(layers)\n",
+    "net_l2.put(\"seed\", 10)\n",
     "net_l2.quick_connect()\n",
     "net_l2.initialize_neural_network()\n",
     "\n",
     "net_l1 = NeuralNetwork(layers)\n",
```
karlnapf: if you touch bits like this, and the ctors are already available through factories, it would be great if you could replace them, e.g. https://github.com/shogun-toolbox/shogun/blob/develop/examples/meta/src/neural_nets/feedforward_net_classification.sg.in

Of course, only if this works out of the box (e.g. when there are examples to copy paste it from).and you have the nerves ;)

gf712: Ah cool, I thought it wasn't working properly with factories, because the old API is still here. I'll have a look. Now the CI is finally actually running the job.

gf712: actually this is not being run by the interfaces...

gf712: @karlnapf How do you think we should expose `DeepAutoencoder::pre_train(std::shared_ptr<Features>)`?

gf712: and DeepAutoencoder::std::shared_ptr<DenseFeatures<float64_t>> data>)?

gf712: @vigsterkr thoughts? ^

vigsterkr: ain't got no DenseFeatures in SWIG :(

karlnapf: this is one of the cases where it is a bit of work. We need to provide an abstraction for such operations in the base api. So far, we managed to work around this, via something like
```
nn.put('pre_train_features', f)
nn.run('pre_train') # throws error if above wasnt done by the user
```
This is also done for things like labels (runtime error if not set), kernels, and more specific situations. I am not the biggest fan (especially for labels, which imo should be passed to `::train`, but that is another discussion), but there *will* be methods that are specific to particular classes and wont be in the base api.

Another solution could be a `watch_method` with a single sgobject parameter (see our discussion about adding one with an index), which is then dynamically casted to the desired class.
```
nn.run('pre_train', f)
```

But extending the base API in this matter is also at risk of making things complicated.

The root problem really is the design of the NN class here. Under Shogun's paradigm, it is bad style of relying on such mechanics. So a re-design of the class would be best. Where something like "pre-train" is not necessary, or called inside `::train`

doc/ipython-notebooks/classification/MKL.ipynb
```diff
@@ -694,30 +685,30 @@
...
     "\n",
-    "mkl = sg.MKLMulticlass(1.2, kernel, labels)\n",
+    "mkl = sg.machine(\"MKLMulticlass\", C=1.2, kernel=kernel, \n",
+    "                 labels=labels, mkl_eps=0.001, mkl_norm=1)\n",
     "\n",
-    "mkl.set_epsilon(1e-2)\n",
-    "mkl.set_mkl_epsilon(0.001)\n",
-    "mkl.set_mkl_norm(1)\n",
+    "# FIX ME: need a way to interact with SVM\n",
+    "# mkl.set_epsilon(1e-2)\n",
```
gf712: @karlnapf @vigsterkr need to come up with a way to handle these situations. Would we .get() the SVM and manipulate it?

src/shogun/classifier/mkl/MKLMulticlass.cpp
```diff
@@ -34,9 +32,20 @@ MKLMulticlass::MKLMulticlass(float64_t C, std::shared_ptr<Kernel> k, std::shared
...
+{
 	mkl_eps=0.01;
 	max_num_mkl_iters=999;
 	pnorm=1;
+
+	SG_ADD(&mkl_eps, "mkl_eps", "MKL Epsilon");
+	SG_ADD(&max_num_mkl_iters, "max_num_mkl_iters", "MKL max number of iterations");
+	SG_ADD(
+	    &pnorm, "mkl_norm", "MKL norm", ParameterProperties::NONE | ParameterProperties::CONSTRAIN,
+	    SG_CONSTRAINT(greater_than_or_equal(1.0)));
```
gf712, line 48: @karlnapf added this to MKLMulticlass, should be fine no? Nothing was being exposed to parameter framework.

src/shogun/machine/FeatureDispatchCRTP.h
```diff
@@ -35,7 +35,7 @@ namespace shogun
 	protected:
 		virtual bool train_dense(std::shared_ptr<Features> data)
 		{
-			auto this_casted = this->template as<P>();
+			auto* this_casted = static_cast<P*>(this);
```
gf712, line 38: no need for dynamic casting here

src/shogun/regression/LinearRidgeRegression.cpp
```diff
@@ -39,12 +39,13 @@ void LinearRidgeRegression::init()
 
 	SG_ADD(&m_tau, "tau", "Regularization parameter", ParameterProperties::HYPER);
 	SG_ADD(
-	    &m_use_bias, "use_bias", "Whether or not to fit an offset term");
+	    &m_use_bias, "use_bias", "Whether or not to fit an offset term", 
+	    ParameterProperties::SETTING);
```
gf712, line 43: @karlnapf there is this issue now with crossvalidation that if we don't set a parameter to either hyperparameter or setting it won't be copied/cloned... Need to continue going through all the parameters in all algos and set something..

src/shogun/distributions/KernelDensity.cpp
```diff
@@ -91,12 +91,15 @@ bool KernelDensity::train(std::shared_ptr<Features> data)
 	return true;
 }
 
-SGVector<float64_t> KernelDensity::get_log_density(const std::shared_ptr<DenseFeatures<float64_t>>& test, int32_t leaf_size)
+SGVector<float64_t> KernelDensity::get_log_density(const std::shared_ptr<Features>& test, int32_t leaf_size)
 {
 	require(test,"data not supplied");
 
+	auto dense_feat = std::dynamic_pointer_cast<DenseFeatures<float64_t>>(test);
+	require(dense_feat, "Expected Features to be of type DenseFeatures.");
```
gf712: @karlnapf is it fine to have this userfacing?

doc/ipython-notebooks/classification/MKL.ipynb
karlnapf: it is commented out here, so it didnt work before I guess? :)
Is this something that would need to be applied recursively? I.e. all svms need to have set the epsilon. Or is it something of the MKL class?

src/shogun/classifier/mkl/MKLMulticlass.cpp
karlnapf: yep, otherwise it is also not serialized and loading will reset it to the default setting

src/shogun/regression/LinearRidgeRegression.cpp
karlnapf: yes, I think this whitelisting idea is maybe not the best. What about blacklisting parameters instead in xval? Like readonly ones?

src/shogun/distributions/KernelDensity.cpp
karlnapf: can't this be moved to the crtp dispatcher? There is an error msg in there. But I guess it is some work. So then it is fine like this as well. Although people will not understand it. I also don't like using classnames in user-facing msgs (as they tend to change). What about say "dense real features"

doc/ipython-notebooks/classification/MKL.ipynb
gf712: isn't there only a single SVM? In any case, what happens is that `MulticlassSVM::set_epsilon` is called, which casts the underlying `m_machine` to SVM and calls `set_epsilon` on that

src/shogun/distributions/KernelDensity.cpp
gf712: Hmm it would be some work, because this is not the train method. We would need a dispatcher for get_log_density?

doc/ipython-notebooks/classification/MKL.ipynb
karlnapf: I see. Maybe
`mkl.get('svm').put('epsilon')`

Since we are using oop design, and the user provides an object, why not do that?

Or we somehow take over the SVM parameters with the MKL class so that we can use "put"?

src/shogun/distributions/KernelDensity.cpp
karlnapf: ah I see, maybe later right? :)

doc/ipython-notebooks/classification/MKL.ipynb
gf712: Yup, that's what I did, just get.("machine").put(), it keeps the interface cleaner imo

src/shogun/regression/LinearRidgeRegression.cpp
gf712: Yup, that is probably better... Just need to label all the labels/features as readonly

src/shogun/distributions/KernelDensity.cpp
gf712: indeed.. Otherwise there will be too much in this PR

src/shogun/regression/LinearRidgeRegression.cpp
karlnapf: ah no! Thats why we did it this way. Features should not be readonly.
We need to whitelist as otherwise we have no way to distinguish between model parameters and things like feautres. I remember now.
I put a PR in to test exactly such things (explicit xval by hand against the class we have), that they have the same results. If we combine this with random parameter content, this will catch missing parameter registrations

doc/ipython-notebooks/gaussian_process/variational_classifier.ipynb
```diff
@@ -1279,7 +1278,7 @@
     "    name=inf.get_name()\n",
     "    prbs=gpc.get_probabilities(features_boundary)\n",
     "    plot_helper(plots[0],x1_boundary,x2_boundary,prbs,n_boundary,name,x1s,x2s,\"without optimizing %d inducing points\"%n_inducing) \n",
-    "    x_inducing=inf.get_inducing_features().get_computed_dot_feature_matrix()\n",
+    "    x_inducing=inf.get_inducing_features().get(\"feature_matrix\")  \n",
```
gf712, line 1280: @vigsterkr how do you call `get_computed_dot_feature_matrix` with a Features object?

gf712: pretty sure what I wrote here isn't right

karlnapf: have a look at the plot, that makes it clear
I actually think this might work

doc/ipython-notebooks/neuralnets/autoencoders.ipynb
```diff
@@ -391,7 +399,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "conv_nn = ae.convert_to_neural_network(NeuralSoftmaxLayer(10))\n",
+    "conv_nn = sg.convert_to_neural_network(ae, sg.layer(\"NeuralSoftmaxLayer\", num_neurons=10), 0.01)\n",
```
gf712, line 402: @karlnapf @vigsterkr would this be a reasonable tradeoff? Imo much more readable than putting an output layer and then calling get() to get a net Machine out...

gf712: also, each free function adds about 100 LoC in SWIG

karlnapf: I say do it.

But one question: Could we maybe overload the factory for this?

gf712: Hmmm not sure I understand. You mean the cast functions?

karlnapf: the factory for labels can for example take a vector or a csv file and then returns a labels object, potentially converted. See what I mean?

gf712: hmm, I see. You mean `sg.machine(ae, sg.layer("NeuralSoftmaxLayer", num_neurons=10), sigma=0.01)` would convert DeepAutoencoder to a NeuralNetwork with an output layer for inference?

gf712:
@karlnapf @vigsterkr shall I merge this and now we have notebooks in CI? It’s an extra 25 minutes of testing but probably still quicker than MSVC build...

doc/ipython-notebooks/neuralnets/autoencoders.ipynb
karlnapf: Something in those lines

gf712: OK. I can look into this in another PR

karlnapf: actually rather than using the factory, we should use the casting functions we have `as_neural_network`, `as_kernel`, etc

gf712: OK, I'll write an issue for this then!

gf712:
Removed PR notebook from buildbot for the time being. Can be brought back when we update the script to only run specific notebooks, rather than using all the donbot cores for each PR :D

PR #4977, opened 2020-03-29
Update INSTALL.md

ANOLASC:
I found that the release in nightly was not up to date, and I modified  the description.


karlnapf:
Closing this as:

* you havent actually changed the nightly link
* the link works
* the change of "latest stable" to "shogun 6.1.3" you made is not good as then we would need to change the file with every release, which easily slips and leads to outdated readmes ....

PR #4986, opened 2020-04-01
Feature/kalman tool box

Abdelrahman350:
This is a new feature branch where I want to start building shogun's Kalman toolbox.


Abdelrahman350:
Hello, I want to know how to begin implementing the standard Kalman filter. I know the Algorithm, but I need to discuss the software requirements needed. The idea I have is to implement all needed models in a .h file and include this file in the Kalman.h file and implement a parameterized constructor that handles the various models. something like in this toolbox in MATLAB [https://www.mathworks.com/help/driving/ref/multiobjecttracker-system-object.html](url).

vigsterkr:
@Abdelrahman350 before doing anything further make sure you cleanup your branch of the Eigen. shogun depends already on eigen and it's available no need to import it...

Abdelrahman350:
> @Abdelrahman350 before doing anything further make sure you cleanup your branch of the Eigen. shogun depends already on eigen and it's available no need to import it...

Ok, this will be my next commit. Furthermore I want to use shogun's mathematics library directly. Thank you for your advice.

src/shogun/kalman/kalman_filter.h
```diff
@@ -0,0 +1,52 @@
+#ifndef KALMAN_FILTER_H_
+#define KALMAN_FILTER_H_
+#include <shogun/mathematics/eigen3.h>
```
vigsterkr, line 3: never in a header

src/shogun/kalman/kalman_filter.h
```diff
@@ -0,0 +1,52 @@
...
+
+using Eigen::MatrixXd;
+using Eigen::VectorXd;
+
+class KalmanFilter {
```
vigsterkr, line 8: plz observe how other methods are being implemented... they should inherit from SGObject etc etc... i suppose this is going to be a transformer? then should inherit from Transformer

src/shogun/kalman/kalman_filter.h
```diff
@@ -0,0 +1,52 @@
...
+  /**
+   * Destructor
+   */
+  virtual ~KalmanFilter();
+
+  /**
+   * Predict Predicts the state and the state covariance
+   *   using the process model
+   */
+  void Predict();
```
vigsterkr, line 24: no we dont do this directly in an object

src/shogun/kalman/kalman_filter.h
```diff
@@ -0,0 +1,52 @@
...
+
+  /**
+   * Updates the state and
+   * @param z The measurement at k+1
+   */
+  void Update(const VectorXd &z);
```
vigsterkr, line 30: no we dont expose eigen types to the user

src/shogun/kalman/kalman_filter.h
```diff
@@ -0,0 +1,52 @@
...
+ 
+  // state vector
+  VectorXd x_;
+
+  // state covariance matrix
+  MatrixXd P_;
```
vigsterkr, line 36: we dont use eigen matrix directly...

vigsterkr (changes requested):
you should really first try to check out more the codebase to understand how shogun works.
and don't just create a directory in shogun kalman and put stuff there...

Abdelrahman350:
> you should really first try to check out more the codebase to understand how shogun works.
> and don't just create a directory in shogun kalman and put stuff there...

Ok, thank you for your notices I will work on it @vigsterkr .

PR #4989, opened 2020-04-02
Update the undocumented graphical python examples

vinnik-dmitry07:


examples/undocumented/python/graphical/mclda.py
```diff
@@ -1,63 +1,57 @@
...
+plt.plot(x_neg, y_neg, 'ro')
 
 # train qda
-labels = MulticlassLabels( np.concatenate([np.zeros(N), np.ones(N)]) )
+labels = MulticlassLabels(np.concatenate([np.zeros(N), np.ones(N)]))
 pos = np.array([x_pos, y_pos])
 neg = np.array([x_neg, y_neg])
-features = RealFeatures( np.array(np.concatenate([pos, neg], 1)) )
+
+features_ = features(np.array(np.concatenate([pos, neg], 1)))
```
gf712: @karlnapf maybe we should stop `from shogun import *` to avoid this kind of thing happening?

karlnapf: yes absolutely, we should use `import shogun as sg` and then `sg.features` (actually also in meta examples, Ill open an issue)

karlnapf: #4990

examples/undocumented/python/graphical/lda.py
```diff
@@ -1,32 +1,31 @@
-from pylab import figure,pcolor,scatter,contour,colorbar,show,subplot,plot,connect
+import matplotlib.pyplot as plt
```
karlnapf, line 1: I would put changes to python graphical example in a different PR than the C++ code below. EDIT. sorry I only now realised that those are cleanups....fine to have a single PR then

Also, note that graphical python examples are pretty much the bottom of our priority list, so let's get these changes in but let's not do more of these. Instead, I suggest you focus on sending some C++ code

examples/undocumented/python/graphical/mclda.py
```diff
@@ -1,63 +1,57 @@
-from shogun import RealFeatures
 from shogun import MulticlassLabels
 from shogun import MCLDA
-from pylab import pcolor, contour, colorbar, connect, show, plot, axis
-
+import matplotlib.pyplot as plt
 import numpy as np
-import util
+from shogun import features
```
karlnapf: actually I just looked through the changes here, the only functional change is replacing RealFeatures with Features.
The file above has no functional changes as far as I can see.

examples/undocumented/python/graphical/mclda.py
```diff
@@ -1,63 +1,57 @@
-from shogun import RealFeatures
 from shogun import MulticlassLabels
```
karlnapf: could we have a `import shogun as sg` as per the issue I referenced?

karlnapf: (and remove all the other imports?)

examples/undocumented/python/graphical/util.py
```diff
@@ -1,10 +1,14 @@
 """ Utilities for matplotlib examples """
 
 import pylab
-from numpy import ones, array, double, meshgrid, reshape, linspace, \
-	concatenate, ravel, pi, sinc
+from numpy import ones, array, double, meshgrid, linspace, concatenate, ravel, pi, sinc
```
karlnapf: import numpy as np

src/shogun/distributions/KernelDensity.h
```diff
@@ -50,8 +50,8 @@ enum EEvaluationMode
 
 /** @brief This class implements the kernel density estimation technique. Kernel density estimation is a non-parametric
  * way to estimate an unknown pdf. The pdf at a query point given finite training samples is calculated using the
- * following formula : \\
- * \f$pdf(x')= \frac{1}{nh} \sum_{i=1}^n K(\frac{||x-x_i||}{h})\f$ \\
+ * following formula: \n
```
karlnapf: I think there are centred math tex commands in doxygen, so you dont have to do newlines by hand. See e.g. GaussianKernel.h
```
 * \f[
 * k({\bf x},{\bf x'})= exp(-\frac{||{\bf x}-{\bf x'}||^2}{\tau})
 * \f]
```

src/shogun/multiclass/tree/KDTree.h
```diff
@@ -40,7 +40,7 @@ namespace shogun
 {
 
 /** @brief This class implements KD-Tree.
- * cf. http://www.autonlab.org/autonweb/14665/version/2/part/5/data/moore-tutorial.pdf
+ * cf. https://web.archive.org/web/20140327060654if_/http://www.autonlab.org:80/autonweb/14665/version/2/part/5/data/moore-tutorial.pdf
```
karlnapf: is this really a good idea? Why not put a reference as a citation (a different/better reference, or wikipedia)

examples/undocumented/python/graphical/lda.py
```diff
@@ -1,32 +1,31 @@
...
 from shogun import *
 import util
 
-util.set_title('LDA')
```
karlnapf, line 5: for the record. next time please make a smaller PR title. This one is way too epic ;)

examples/undocumented/python/graphical/lda.py
```diff
@@ -1,32 +1,31 @@
...
-plot(neg[0,:], neg[1,:], "b.")
+neg = util.get_realdata(False)
+plt.plot(neg[0, :], neg[1, :], "b.")
 
 # train lda
-labels=util.get_labels()
-features=util.get_realfeatures(pos, neg)
-lda=LDA(gamma, features, labels)
+labels = util.get_labels()
+features = util.get_realfeatures(pos, neg)
```
karlnapf, line 20: while we are at it `get_real_features`

examples/undocumented/python/graphical/lda.py
```diff
@@ -1,32 +1,31 @@
...
+lda = LDA(gamma, features, labels)
```
karlnapf: white we are at it. Could you port this to the new API usage, something in the lines of
```
lda = machine("LDA", gamma=gamma, labels=labels)
lda.train(features)
```

examples/undocumented/python/graphical/mclda.py
```diff
@@ -1,63 +1,57 @@
...
+cov_neg = [[100, 50], [20, 3]]
 
-x_neg, y_neg = np.random.multivariate_normal(mean_neg, cov_neg, 500).T
-plot(x_neg, y_neg, 'ro');
+x_neg, y_neg = np.random.multivariate_normal(mean_neg, cov_neg, N).T
+plt.plot(x_neg, y_neg, 'ro')
 
 # train qda
-labels = MulticlassLabels( np.concatenate([np.zeros(N), np.ones(N)]) )
+labels = MulticlassLabels(np.concatenate([np.zeros(N), np.ones(N)]))
```
karlnapf: use `sg.labels` factory (new api)

examples/undocumented/python/graphical/mclda.py
```diff
@@ -1,63 +1,57 @@
...
 pos = np.array([x_pos, y_pos])
 neg = np.array([x_neg, y_neg])
-features = RealFeatures( np.array(np.concatenate([pos, neg], 1)) )
+
+features_ = features(np.array(np.concatenate([pos, neg], 1)))
 
 lda = MCLDA()
```
karlnapf: new API

examples/undocumented/python/graphical/mclda.py
```diff
@@ -1,63 +1,57 @@
...
 lda.set_labels(labels)
```
karlnapf: use `put` (or rather a kwarg when using the `machine` factory)

examples/undocumented/python/graphical/mclda.py
```diff
@@ -1,63 +1,57 @@
...
+x2_min = min(1.2 * ys)
 
 x1 = np.linspace(x1_min, x1_max, size)
 x2 = np.linspace(x2_min, x2_max, size)
 
 x, y = np.meshgrid(x1, x2)
 
-dense = RealFeatures( np.array((np.ravel(x), np.ravel(y))) )
+dense = features(np.array((np.ravel(x), np.ravel(y))))
 dense_labels = lda.apply(dense).get_labels()
```
karlnapf: use `get`

examples/undocumented/python/graphical/lda.py
vinnik-dmitry07: I made these changes for vigsterkr can run this example, because there, and I guessed in some other places, was also the problem with `const std::shared_ptr<DenseFeatures<float64_t>>&`. I mean I did this, not because of the priorities. But now I have understood that it is a SWIG problem.

src/shogun/multiclass/tree/KDTree.h
vinnik-dmitry07: I am not the author of code, so I decided to put what the author meant when was writing this comment

vinnik-dmitry07: I saw that link from the archive works, I do not really investigate content. Sorry, will do better next time.

examples/undocumented/python/graphical/mclda.py
vinnik-dmitry07: Aye

examples/undocumented/python/graphical/lda.py
karlnapf: sure thing. the patch is welcome, let's get it in, just wanted to make sure you are on the same page as us

src/shogun/multiclass/tree/KDTree.h
karlnapf: no worries at all. I just think dead-links should be removed rather than pulled out from the internet archive ;) wikepedia is probably fine here anyways.

examples/undocumented/python/graphical/lda.py
karlnapf: and it does make sense to improve the whole listing and make it ready for the new api, so then we don't have to touch it again anytime soon.

examples/undocumented/python/graphical/classifier_gaussian_process_binary_classification.py
```diff
@@ -1,61 +1,52 @@
...
-        return
+def gaussian_process_binary_classification_laplace(X_train, y_train, n_test=50):
     import shogun as sg
-    import numpy as np
 
     # convert training data into Shogun representation
-    train_features = RealFeatures(X_train)
-    train_labels = BinaryLabels(y_train)
+    train_features = sg.features(X_train)
+    train_labels = sg.BinaryLabels(y_train)
```
karlnapf: sg.labels should work

examples/undocumented/python/graphical/classifier_gaussian_process_binary_classification.py
```diff
@@ -1,61 +1,52 @@
...
-    test_features = RealFeatures(X_test)
+    test_features = sg.features(X_test)
 
     # create Gaussian kernel with width = 2.0
-    kernel = sg.kernel("GaussianKernel", log_width=np.log(2.0))
+    kernel = sg.kernel('GaussianKernel', log_width=np.log(2.0))
 
     # create zero mean function
-    mean = ZeroMean()
+    mean = sg.ZeroMean()
```
karlnapf, line 49: is there a factory for mean functions? I think so as we have a gp meta example

examples/undocumented/python/graphical/classifier_gaussian_process_binary_classification.py
```diff
@@ -73,35 +64,35 @@ def gaussian_process_binary_classification_laplace(X_train, y_train, n_test=50):
     # inf = SingleLaplacianInferenceMethod(kernel, train_features, mean, train_labels, lik)
 
     # specify EP approximation inference method
-    inf = EPInferenceMethod(kernel, train_features, mean, train_labels, lik)
+    inf = sg.EPInferenceMethod(kernel, train_features, mean, train_labels, lik)
 
     # create and train GP classifier, which uses Laplace approximation
-    gp = GaussianProcessClassification(inf)
+    gp = sg.GaussianProcessClassification(inf)
```
karlnapf, line 71: all those should have a factory, meta examples should guide you

karlnapf (commented):
Thanks for the update.
I'd like to encourage you (again) to focus on c++ code though. These graphical example are not super important. But since you have made the effort in polishing them, let's finish the job (on those you touched), but please don't add new ones to the list for now.
I made a few comments, mostly regarding factories. Try to use them where you can. If none exist, open an issue for it (you don't have to do it here)

Finally, I think we should add those examples to the CI (with a fake matplotlib), so that they get at least executed. Thoughts @vigsterkr ?

vinnik-dmitry07:
> I'd like to encourage you (again) to focus on c++ code though
> but please don't add new ones to the list for now

Okidoki. Btw these examples look really nice, should be in the showroom, but it does not work :(.
image

vinnik-dmitry07:
Also here is a problem with SPE_LOCAL (not caused by changes).

karlnapf:
We could add those to notebooks for example, but let's do that later.
Any idea what the problem with SPE is?

examples/undocumented/python/graphical/classifier_gaussian_process_binary_classification.py
vinnik-dmitry07: image

vinnik-dmitry07: How to run this method from the machine? (get_probabilities)

karlnapf: just checked, it is not possible currently, and we need to refactor this stuff (meta example is also not yet ported to new api I realised). Ignore my comment then

vinnik-dmitry07:
> We could add those to notebooks for example, but let's do that later.
> Any idea what the problem with SPE is?

I finally figured out that m_k should be about 400 (from total 500 points), also along the way I fixed some bug and added missed parameter.
image

examples/undocumented/python/graphical/classifier_perceptron_graphical.py
```diff
@@ -1,78 +1,75 @@
...
-	from shogun import RealFeatures, BinaryLabels
-	from shogun import Perceptron
-	from shogun import MSG_INFO
+    # 2D data
+    _DIM = 2
 
-	# 2D data
-	_DIM = 2
+    # To get the nice message that the perceptron has converged
+    # dummy = sg.BinaryLabels()
```
karlnapf: this line is unnecessary or?
I'd also remove the comment about the info loglevel

examples/undocumented/python/graphical/converter_ffsep_bss.py
```diff
@@ -7,12 +7,9 @@
 Kevin Hughes 2013
```
karlnapf: question: could we group all the ica examples into one? The code is more or less the same. We could just have a loop over all the different ica methods

examples/undocumented/python/graphical/mclda.py
karlnapf: `lda.apply(dense).get('labels')`

examples/undocumented/python/graphical/util.py
```diff
@@ -1,115 +1,117 @@
...
-
-	return x, y, z
+    data = np.concatenate(np.array(
+        (-np.ones(NUM_EXAMPLES, dtype=np.double), np.ones(NUM_EXAMPLES, dtype=np.double))
+    ))
+    if raw:
+        return data
+    else:
+        if type == 'binary':
+            return sg.BinaryLabels(data)
```
karlnapf: `labels` factory pls

karlnapf: and then you shouldn't need the if then else, as the factory does that

src/shogun/converter/StochasticProximityEmbedding.cpp
```diff
@@ -29,19 +28,19 @@ void StochasticProximityEmbedding::init()
 {
 	SG_ADD(&m_k, "m_k", "Number of neighbors");
 	SG_ADD(&m_tolerance, "m_tolerance", "Regularization parameter");
-	SG_ADD(&m_max_iteration, "max_iteration", "maximum number of iterations");
+	SG_ADD(&m_num_updates, "m_num_updates", "SPE number of updates");
+	SG_ADD(&m_max_iteration, "max_iteration", "Maximum number of iterations");
```
karlnapf, line 32: @gf712 are we still using SG_ADD; or is the templated watch already there?

src/shogun/converter/StochasticProximityEmbedding.h
```diff
@@ -135,39 +137,36 @@ class StochasticProximityEmbedding : public EmbeddingConverter
...
 		ESPEStrategy m_strategy;
 
 		/** number of neighbours in local strategy */
 		int32_t m_k;
 
 		/** regularization parameter */
-		float32_t m_tolerance;
+		float64_t m_tolerance;
 
-		/** number of apdates per SPE iteration */
```
karlnapf, line 162: "apdates" :laughing:

karlnapf (commented):
looks good, few more minor fixes and then we can merge it

examples/undocumented/python/graphical/classifier_gaussian_process_binary_classification.py
karlnapf: #4996

examples/undocumented/python/graphical/classifier_perceptron_graphical.py
vinnik-dmitry07: Yes, Perceptron now does not use SG_INFO, I will remove it.

examples/undocumented/python/graphical/classifier_gaussian_process_binary_classification.py
vinnik-dmitry07: I did not find.

vinnik-dmitry07:
Sorry, I remember that I do not have to touch more files but to test nothing breaks after changing this: https://github.com/shogun-toolbox/shogun/pull/4989#discussion_r403453180 I had to update kernel_ridge_regression.

vinnik-dmitry07:
Accidentally-closed

examples/undocumented/python/graphical/kernel_ridge_regression_sinc.py
```diff
@@ -1,31 +1,30 @@
...
-gk=GaussianKernel(feat, feat, width)
-krr=KernelRidgeRegression()
-krr.set_labels(lab)
-krr.set_kernel(gk)
-krr.set_tau(1e-6)
+feat = sg.features(X)
+lab = sg.labels(Y.flatten())
+gk = sg.kernel('GaussianKernel', log_width=width)
+gk.init(feat, feat)
+krr = sg.machine('KernelRidgeRegression')
```
karlnapf: could you use python `**kwargs`, i.e.
`krr = sg.machine('KernelRidgeRregression", labels=lab, kernel=gk, tau=1e-3)` (and for all the other instantiations as well). Makes everything look more like Python :)

examples/undocumented/python/graphical/kernel_ridge_regression_sinc.py
```diff
@@ -1,31 +1,30 @@
...
-plot([XE[0,200]], [YE200], '+', hold=True)
-#print YE[200], YE200
+XE, YE = util.compute_output_plot_isolines_sine(krr, gk, feat, regression=True)
+YE200 = krr.apply_one(200)
 
-connect('key_press_event', util.quit)
-show()
+plt.plot(XE[0], YE, label='test output')
+plt.plot([XE[0, 200]], [YE200], '+')
+# print(YE[200], YE200)
```
karlnapf: remove

src/shogun/converter/StochasticProximityEmbedding.h
```diff
@@ -7,72 +7,74 @@
...
+	{
 
 	public:
-
 		/** constructor */
 		StochasticProximityEmbedding();
 
 		/** destructor */
-		virtual ~StochasticProximityEmbedding();
+		~StochasticProximityEmbedding() override;
```
karlnapf: no virtual destructor?

src/shogun/converter/StochasticProximityEmbedding.h
```diff
@@ -102,31 +104,31 @@ class StochasticProximityEmbedding : public EmbeddingConverter
...
 		 */
-		int32_t get_nupdates() const;
+		int32_t get_num_updates() const;
 
 		/** setter for the maximum number of iterations
 		 *
 		 * @param max_iteration the maximum number of iterations
 		 */
-		void set_max_iteration(const int32_t max_iteration);
+		void set_max_iteration(int32_t max_iteration);
```
karlnapf, line 131: ++

src/shogun/converter/StochasticProximityEmbedding.h
```diff
@@ -135,39 +137,36 @@ class StochasticProximityEmbedding : public EmbeddingConverter
 		int32_t get_max_iteration() const;
 
 		/** get name */
-		virtual const char* get_name() const;
+		const char* get_name() const override;
```
karlnapf: I am actually not sure what our policy here is, we used to keep the `virtual`. But it is not needed with `override` I guess.
@gf712 @vigsterkr thoughts?

src/shogun/converter/StochasticProximityEmbedding.h
gf712: Was the const it causing issues?

src/shogun/converter/StochasticProximityEmbedding.cpp
gf712: The PR is mostly ready I think. I can have a look tomorrow!

src/shogun/converter/StochasticProximityEmbedding.h
vinnik-dmitry07: I follow https://en.cppreference.com/w/cpp/language/virtual#In_detail

src/shogun/converter/StochasticProximityEmbedding.h
gf712: This is the correct way to do it, but I think it causes lots of compiler warnings... because we always use virtual... We should write a libtooling script to replace all the virtual get_name with override, or even just use sed

src/shogun/converter/StochasticProximityEmbedding.h
vinnik-dmitry07: More here https://clang.llvm.org/extra/clang-tidy/checks/modernize-use-override.html

gf712: Hmm but this is a destructor. You are overriding the base class destructor so there will be leaks..

src/shogun/converter/StochasticProximityEmbedding.h
vinnik-dmitry07: Clang-tidy says: https://clang.llvm.org/extra/clang-tidy/checks/readability-avoid-const-params-in-decls.html

gf712: I see. OK!

src/shogun/converter/StochasticProximityEmbedding.h
vinnik-dmitry07: "override" keyword simply means "this function is marked as virtual in some base class"

vinnik-dmitry07: Concluding everything, all StochasticProximityEmbedding base classes till SGObject (EmbeddingConverter, Converter, Transformer) should have the override keyword instead of virtual.

vinnik-dmitry07: Imagine we have a base class B having a method X and derived classes D_1, D_2, ..., D_N declared so: class D_1 : B; class D_2: D_1; ... 
The most correct way is to declare X as virtual in B and as override it in D_1, ..., D_N-1 and as final in D_N.

gf712: Just had a look online. This override doesn’t do anything. The destructor is always virtual, because it needs to lookup the vtable along the hierarchy to know how to destruct all class members. The override only ensure that the base class is virtual, which is a given in the way shogun is designed from SGObject down. Imo this syntax is confusing and doesn’t bring anything. It says override, but it doesn’t, and to me at lest it looks like a bug (even though it actually isn’t).

gf712: Btw I know how virtual functions work. My point was that overriding a destructor imo would imply overriding the behaviour of the parent classes. It doesn’t, because the destructor override here is a noop.

gf712: https://github.com/isocpp/CppCoreGuidelines/issues/721

vinnik-dmitry07: > It says override, but it doesn’t

[C++11 Standart](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2013/n3690.pdf)
"Even though destructors are not inherited, a destructor in a derived class **_overrides_** a base class destructor." -- Page 233.
<br>
"The use of the _virtual_ specifier in the declaration of an overriding function is legal but redundant (has empty semantics)." -- Page 231.

gf712: But in that case what's the point of `override`, if it overriden anyway? Other than to make sure that the base class has a virtual destructor, which we know it does? Anyway. up to you, my point is that then we need to change every destructor in shogun, otherwise we just end up with a mix and match of patterns and code styles.

vinnik-dmitry07: > Anyway. up to you, my point is that then we need to change every destructor in shogun

I can do it in an hour without any future problems.

>But in that case what's the point of override, if it overriden anyway?

C++ headers were invented to developers who use someone else code. When you first look at a header file, you do not know anything about the base class. As using _virtual_ specifier in the derived class has no sense, we should specify somehow that the method overrides the base method. Plus using override by analogy to writing _private_ specifier helps us not to shoot ourselves in the foot.

vinnik-dmitry07: https://github.com/chromium/chromium/search?q=override&unscoped_q=override
Chromium guys agree with me.

gf712: go for it if you think that is worth it :)

src/shogun/converter/StochasticProximityEmbedding.h
gf712: @vinnik-dmitry07 whilst you are at it can you also fix all of these, please?

src/shogun/converter/StochasticProximityEmbedding.h
gf712: for reference I think @vinnik-dmitry07 means this https://github.com/chromium/chromium/blob/2ca8c5037021c9d2ecc00b787d58a31ed8fc8bcb/tools/clang/plugins/tests/virtual_specifiers.txt

karlnapf: @vinnik-dmitry07 could you hold your horses for now and just follow shogun's style? We don't want to mix up ways we do things, as this leads to redundant discussions and compiler warnings. If you are keen in changing the way this is done in shogun, please open an issue and a separate PR that applies what you want to do to all of shogun.

karlnapf: Of course such a PR is extremely welcome. (Although I suggest you defer this to later and focus on sending something related to your project proposal for now)

karlnapf (approved):
LGTM
thanks a lot for the tidy up!

karlnapf:
somebody should eyeball this and then we can merge it

gf712:
To be on the safe side can you trigger CI, please? Now all the builds are working again!

PR #4992, opened 2020-04-03
use pythonic import

### LINKED ISSUE #4990: pythonic imports of shogun
We want to replace imports in the lines of

* `from shogun import *`
* `from shogun import features`

with `import shogun as sg`, and then call shogun methods with that handle, i.e. `sg.features` instead of just `features`

This is as the number of names of shogun API objects that clash with potential variable names (features) is growing and we want to avoid confusion.

This needs to be done in (in order of priority)

* ~~meta examples (see `examples/meta/generator/targets/python.json`)~~
* notebooks
* all other python code in the repo

LiuYuHui:
fixes part of #4990

### DIFF
--- a/examples/meta/generator/targets/python.json
+++ b/examples/meta/generator/targets/python.json
@@ -5,14 +5,13 @@
         "IncludeInterfacedClasses": true, 
         "IncludeEnums": true,
         "IncludeGlobalFunctions": true,
-        "DependencyListElement": "from shogun import $typeName",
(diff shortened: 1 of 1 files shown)


gf712:
Cool! Looks good. What do you think @karlnapf?

karlnapf:
looks good. Note there is more code that does the old style import (see issue)

gf712:
Yup, I'll merge but remove "fix" so it doesn't close the issue

Issue #4311, opened 2018-05-29
kNN and nearest neighbours


vigsterkr:
Current design of kNN is a bit broken as even though we have various kNN engines that can solve the classification problem itself, but when just simply trying to query the nearest neighbors of a given set of points with `nearest_neighbors` then it simply does a brute-force check for that...

Saurabh7:
I will take a look at this ..

ashutosh-b-b:
Is this still open? Can I take up this?

vigsterkr:
@ashutosh-b-b sure go ahead!

Hephaestus12:
@vigsterkr Can you have a look at this? :)

PR #4994, opened 2020-04-03
Improve nearest neighbour calculation for KNN algorithm

Hephaestus12:
Compiles.
Passes tests.
How should I add unit tests? There are no unit tests for the pre-existing `nearest_neighbours()` function.
#4311

### DIFF
--- a/src/shogun/multiclass/BruteKNNSolver.cpp
+++ b/src/shogun/multiclass/BruteKNNSolver.cpp
@@ -19,14 +19,13 @@ KNNSolver(k, q, num_classes, min_label, train_labels)
 {
 	std::cout<<"entered BruteKNNSolver::constructor\n";
 	init();
-	//nn=NN;
 }
 
 std::shared_ptr<MulticlassLabels> BruteKNNSolver::classify_objects(std::shared_ptr<Distance> knn_distance, const int32_t num_lab, SGVector<int32_t>& train_lab, SGVector<float64_t>& classes)
 {
 	auto output=std::make_shared<MulticlassLabels>(num_lab);
+	
 	//get the k nearest neighbors of each example
-	//SGMatrix<index_t> NN = this->nn;
 	compute_nearest_neighbours(knn_distance);
 
 	//from the indices to the nearest neighbors, compute the class labels
@@ -52,7 +51,6 @@ SGVector<int32_t> BruteKNNSolver::classify_objects_k(std::shared_ptr<Distance> k
 	SGVector<int32_t> output(m_k*num_lab);
 
 	//get the k nearest neighbors of each example
-	//SGMatrix<index_t> NN = this->nn;
 	compute_nearest_neighbours(knn_distance);
 
 	for (index_t i = 0; i < num_lab && (!cancel_computation()); i++)
@@ -77,6 +75,7 @@ bool BruteKNNSolver::train_KNN(std::shared_ptr<Distance> knn_distance)
 bool BruteKNNSolver::compute_nearest_neighbours(std::shared_ptr<Distance> knn_distance)
 {
 	std::cout<<"entered BruteKNNSolver::compute_nearest_neighbours()\n";
+	
 	//number of examples to which kNN is applied
 	int32_t n = knn_distance->get_num_vec_rhs();
 
@@ -98,9 +97,7 @@ bool BruteKNNSolver::compute_nearest_neighbours(std::shared_ptr<Distance> knn_di
 		std::pair<float64_t, index_t> pairt[m_train_labels.vlen];
 	
 		std::cout<<"i is "<<i<<'\n';
-		//std::cout<<" before sorting dists is "<<'\n';
 		dists.display_vector("before sorting dists");
-		//std::cout<<"before sorting train_idxs is "<<'\n';
 		train_idxs.display_vector("before sorting train_idxs");
 
 		// Storing the respective array elements in pairs.
@@ -122,15 +119,12 @@ bool BruteKNNSolver::compute_nearest_neighbours(std::shared_ptr<Distance> knn_di
 		SG_DEBUG("{}", train_idxs.to_string());
 
 			
-		//std::cout<<"after sorting dists is "<<'\n';
 		dists.display_vector("after sorting dists");
-		//std::cout<<"after sorting train_idxs is "<<'\n';
 		train_idxs.display_vector("after sorting train_idxs is");
 
 		//only considering the first k elements
 		SGVector<index_t> nearest_k_train_idxs(train_idxs.vector, m_k, false);
 
-		//std::cout<<"nearest_k_train_idxs is "<<'\n';
 		nearest_k_train_idxs.display_vector("nearest_k_train_idxs");
 
 		m_NN.set_column(i, nearest_k_train_idxs);
--- a/src/shogun/multiclass/BruteKNNSolver.cpp
+++ b/src/shogun/multiclass/BruteKNNSolver.cpp
@@ -16,7 +16,7 @@ KNNSolver(k, q, num_classes, min_label, train_labels)
 	nn=NN;
 }
 
-std::shared_ptr<MulticlassLabels> BruteKNNSolver::classify_objects(std::shared_ptr<Distance> knn_distance, const int32_t num_lab, SGVector<int32_t>& train_lab, SGVector<float64_t>& classes) const
+std::shared_ptr<MulticlassLabels> BruteKNNSolver::classify_objects(std::shared_ptr<Distance> knn_distance, const int32_t num_lab, SGVector<int32_t>& train_lab, SGVector<float64_t>& classes)
 {
 	auto output=std::make_shared<MulticlassLabels>(num_lab);
 	//get the k nearest neighbors of each example
@@ -40,7 +40,7 @@ std::shared_ptr<MulticlassLabels> BruteKNNSolver::classify_objects(std::shared_p
 	return output;
 }
 
-SGVector<int32_t> BruteKNNSolver::classify_objects_k(std::shared_ptr<Distance> knn_distance, const int32_t num_lab, SGVector<int32_t>& train_lab, SGVector<int32_t>& classes) const
+SGVector<int32_t> BruteKNNSolver::classify_objects_k(std::shared_ptr<Distance> knn_distance, const int32_t num_lab, SGVector<int32_t>& train_lab, SGVector<int32_t>& classes)
 {
 	SGVector<int32_t> output(m_k*num_lab);
 
@@ -57,4 +57,18 @@ SGVector<int32_t> BruteKNNSolver::classify_objects_k(std::shared_ptr<Distance> k
 	}
 
 	return output;
+}
+
+bool BruteKNNSolver::train_KNN(std::shared_ptr<Distance> knn_distance)
+{
+	//TODO complete this
+	
+	return true;
+}
+
+bool BruteKNNSolver::compute_nearest_neighbours()
+{
+	//TODO complete this
+
+	return true;
 }
--- a/src/shogun/multiclass/BruteKNNSolver.cpp
+++ b/src/shogun/multiclass/BruteKNNSolver.cpp
@@ -6,21 +6,28 @@
 #include <shogun/base/progress.h>
 #include <shogun/lib/Signal.h>
 #include <shogun/multiclass/BruteKNNSolver.h>
+#include <shogun/multiclass/KNNSolver.h>
+#include <shogun/machine/DistanceMachine.h>
+
+#include <iostream>
+#include <algorithm>
 
 using namespace shogun;
 
-BruteKNNSolver::BruteKNNSolver(const int32_t k, const float64_t q, const int32_t num_classes, const int32_t min_label, const SGVector<int32_t> train_labels, const SGMatrix<index_t> NN):
+BruteKNNSolver::BruteKNNSolver(const int32_t k, const float64_t q, const int32_t num_classes, const int32_t min_label, const SGVector<int32_t> train_labels):
 KNNSolver(k, q, num_classes, min_label, train_labels)
 {
+	std::cout<<"entered BruteKNNSolver::constructor\n";
 	init();
-	nn=NN;
+	//nn=NN;
 }
 
 std::shared_ptr<MulticlassLabels> BruteKNNSolver::classify_objects(std::shared_ptr<Distance> knn_distance, const int32_t num_lab, SGVector<int32_t>& train_lab, SGVector<float64_t>& classes)
 {
 	auto output=std::make_shared<MulticlassLabels>(num_lab);
 	//get the k nearest neighbors of each example
-	SGMatrix<index_t> NN = this->nn;
+	//SGMatrix<index_t> NN = this->nn;
+	compute_nearest_neighbours(knn_distance);
 
 	//from the indices to the nearest neighbors, compute the class labels
 	for (auto i : SG_PROGRESS(range(num_lab)))
@@ -29,7 +36,7 @@ std::shared_ptr<MulticlassLabels> BruteKNNSolver::classify_objects(std::shared_p
 			break;
 		//write the labels of the k nearest neighbors from theirs indices
 		for (index_t j=0; j<m_k; j++)
-			train_lab[j] = m_train_labels[ NN(j,i) ];
+			train_lab[j] = m_train_labels[ m_NN(j,i) ];
 
 		//get the index of the 'nearest' class
 		index_t out_idx = choose_class(classes.vector, train_lab.vector);
@@ -45,13 +52,14 @@ SGVector<int32_t> BruteKNNSolver::classify_objects_k(std::shared_ptr<Distance> k
 	SGVector<int32_t> output(m_k*num_lab);
 
 	//get the k nearest neighbors of each example
-	SGMatrix<index_t> NN = this->nn;
+	//SGMatrix<index_t> NN = this->nn;
+	compute_nearest_neighbours(knn_distance);
 
 	for (index_t i = 0; i < num_lab && (!cancel_computation()); i++)
 	{
 		//write the labels of the k nearest neighbors from theirs indices
 		for (index_t j=0; j<m_k; j++)
-			train_lab[j] = m_train_labels[ NN(j,i) ];
+			train_lab[j] = m_train_labels[ m_NN(j,i) ];
 
 		choose_class_for_multiple_k(output.vector+i, classes.vector, train_lab.vector, num_lab);
 	}
@@ -66,9 +74,68 @@ bool BruteKNNSolver::train_KNN(std::shared_ptr<Distance> knn_distance)
 	return true;
 }
 
-bool BruteKNNSolver::compute_nearest_neighbours()
+bool BruteKNNSolver::compute_nearest_neighbours(std::shared_ptr<Distance> knn_distance)
 {
-	//TODO complete this
+	std::cout<<"entered BruteKNNSolver::compute_nearest_neighbours()\n";
+	//number of examples to which kNN is applied
+	int32_t n = knn_distance->get_num_vec_rhs();
 
+	std::cout << "n is " << n << '\n';
+	std::cout << "k is " << m_k << '\n';
+
+	//distances to train data
+	SGVector<float64_t> dists(m_train_labels.vlen);
+	//indices to train data
+	SGVector<index_t> train_idxs(m_train_labels.vlen);
+
+	for (int i = 0; i<n; i++)
+	{
+		//COMPUTATION_CONTROLLERS
+		distances_lhs(dists,0,m_train_labels.vlen-1,i);
+
+		train_idxs.range_fill(0);
+
+		std::pair<float64_t, index_t> pairt[m_train_labels.vlen];
+	
+		std::cout<<"i is "<<i<<'\n';
+		//std::cout<<" before sorting dists is "<<'\n';
+		dists.display_vector("before sorting dists");
+		//std::cout<<"before sorting train_idxs is "<<'\n';
+		train_idxs.display_vector("before sorting train_idxs");
+
+		// Storing the respective array elements in pairs.
+		for (int j = 0; j < m_train_labels.vlen; j++)
+		{
+			pairt[j].first = dists[j];
+			pairt[j].second = train_idxs[j];
+		}
+
+		std::sort(pairt, pairt + m_train_labels.vlen);
+
+		for (int j = 0; j < m_train_labels.vlen; j++)
+		{
+			dists[j] = pairt[j].first;
+			train_idxs[j] = pairt[j].second;
+		}
+
+		SG_DEBUG("\nQuick sort query {}", i);
+		SG_DEBUG("{}", train_idxs.to_string());
+
+			
+		//std::cout<<"after sorting dists is "<<'\n';
+		dists.display_vector("after sorting dists");
+		//std::cout<<"after sorting train_idxs is "<<'\n';
+		train_idxs.display_vector("after sorting train_idxs is");
+
+		//only considering the first k elements
+		SGVector<index_t> nearest_k_train_idxs(train_idxs.vector, m_k, false);
+
+		//std::cout<<"nearest_k_train_idxs is "<<'\n';
+		nearest_k_train_idxs.display_vector("nearest_k_train_idxs");
+
+		m_NN.set_column(i, nearest_k_train_idxs);
+	}
+	knn_distance->reset_precompute();
+	std::cout<<"exiting BruteKNNSolver::compute_nearest_neighbours()\n";
 	return true;
 }
--- a/src/shogun/multiclass/KNN.cpp
+++ b/src/shogun/multiclass/KNN.cpp
@@ -105,14 +105,10 @@ bool KNN::train_machine(std::shared_ptr<Features> data)
 	return true;
 }
 
-SGMatrix<index_t> KNN::nearest_neighbors(KNN_SOLVER knn_solver/*=KNN_BRUTE*/)
+SGMatrix<index_t> KNN::nearest_neighbors()
 {
 	//number of examples to which kNN is applied
 	int32_t n=distance->get_num_vec_rhs();
-
-	require(
-	    n >= m_k,
-	    "K ({}) must not be larger than the number of examples ({}).", m_k, n);
 
 	//distances to train data
 	SGVector<float64_t> dists(m_train_labels.vlen);
@@ -124,7 +120,7 @@ SGMatrix<index_t> KNN::nearest_neighbors(KNN_SOLVER knn_solver/*=KNN_BRUTE*/)
 	distance->precompute_lhs();
 	distance->precompute_rhs();
 
-	switch (knn_solver)
+	switch (m_knn_solver)
 	{
 	case KNN_BRUTE:
 	{
@@ -136,26 +132,20 @@ SGMatrix<index_t> KNN::nearest_neighbors(KNN_SOLVER knn_solver/*=KNN_BRUTE*/)
 			distances_lhs(dists,0,m_train_labels.vlen-1,i);
 
 			//fill in an array with 0..num train examples-1
-			for (int32_t j=0; j<m_train_labels.vlen; j++)
-				train_idxs[j]=j;
+			train_idxs.range_fill(0);
 
 			//sort the distance vector between test example i and all train examples
 			Math::qsort_index(dists.vector, train_idxs.vector, m_train_labels.vlen);
 
-#ifdef DEBUG_KNN
-			io::print("\nQuick sort query {}\n", i);
-			for (int32_t j=0; j<m_k; j++)
-				io::print("{} ", train_idxs[j]);
-			io::print("\n");
-#endif
+			SG_DEBUG("\nQuick sort query {}", i);
+			SG_DEBUG("{}", train_idxs.to_string());
+			SG_DEBUG("");
 
 			//fill in the output the indices of the nearest neighbors
-			for (int32_t j=0; j<m_k; j++)
-				NN(j,i) = train_idxs[j];
+			sg_memcpy(&NN.matrix[m_k * i], train_idxs.vector, sizeof(index_t) * m_k);
 		}
 
 		distance->reset_precompute();
-
 		break;
 	}
 	case KNN_KDTREE:
--- a/src/shogun/multiclass/KNN.cpp
+++ b/src/shogun/multiclass/KNN.cpp
@@ -122,93 +122,20 @@ bool KNN::train_machine(std::shared_ptr<Features> data)
 SGMatrix<index_t> KNN::nearest_neighbors()
 {
 	std::cout<<"entered KNN::nearest_neighbors()";
-	//number of examples to which kNN is applied
-	/* int32_t n=distance->get_num_vec_rhs();
-	std::cout<<"Number of vectors RHS is :"<< n << '\n'; */
 
-	//distances to train data
-//	SGVector<float64_t> dists(m_train_labels.vlen);
-	//indices to train data
-//	SGVector<index_t> train_idxs(m_train_labels.vlen);
 	//pre-allocation of the nearest neighbors
 	SGMatrix<index_t> NN;
 
 	distance->precompute_lhs();
 	distance->precompute_rhs();
-	//std::cout<<"n is "<<n<<'\n';
-	std::cout<<"k is "<<m_k<<'\n';
+
 	switch (m_knn_solver)
 	{
 	case KNN_BRUTE:
-	{
-		std::cout<<"ENTERING CASE: KNN_BRUTE"<<'\n';
-		/* for (auto i : SG_PROGRESS(range(n)))
-		{
-			COMPUTATION_CONTROLLERS
-			distances_lhs(dists,0,m_train_labels.vlen-1,i);
-
-			train_idxs.range_fill(0);
-
-			std::pair<float64_t, index_t> pairt[m_train_labels.vlen];
-
-			
-			std::cout<<"i is "<<i<<'\n';
-			//std::cout<<" before sorting dists is "<<'\n';
-			dists.display_vector("before sorting dists");
-			//std::cout<<"before sorting train_idxs is "<<'\n';
-			train_idxs.display_vector("before sorting train_idxs");
-
-			// Storing the respective array elements in pairs.
-			for (int j = 0; j < m_train_labels.vlen; j++)
-			{
-				pairt[j].first = dists[j];
-				pairt[j].second = train_idxs[j];
-			}
-
-			sort(pairt, pairt + m_train_labels.vlen);
-
-			for (int j = 0; j < m_train_labels.vlen; j++)
-			{
-				dists[j] = pairt[j].first;
-				train_idxs[j] = pairt[j].second;
-			}
-
-			SG_DEBUG("\nQuick sort query {}", i);
-			SG_DEBUG("{}", train_idxs.to_string());
-
-			
-			//std::cout<<"after sorting dists is "<<'\n';
-			dists.display_vector("after sorting dists");
-			//std::cout<<"after sorting train_idxs is "<<'\n';
-			train_idxs.display_vector("after sorting train_idxs is");
-
-			//only considering the first k elements
-			SGVector<index_t> nearest_k_train_idxs(train_idxs.vector, m_k, false);
-
-			//std::cout<<"nearest_k_train_idxs is "<<'\n';
-			nearest_k_train_idxs.display_vector("nearest_k_train_idxs");
-
-			NN.set_column(i, nearest_k_train_idxs);
-		}
-
-		distance->reset_precompute(); */
-		solver->compute_nearest_neighbours(distance);
-		NN = solver->get_nearest_neighbours();
-		break;
-	}
 	case KNN_KDTREE:
 	{
-		//auto lhs = distance->get_lhs();
-		//auto kd_tree = std::make_shared<KDTree>(m_leaf_size);
-		//kd_tree->build_tree(lhs->as<DenseFeatures<float64_t>>());
-
-		//auto query = distance->get_rhs();
-		//std::shared_ptr<shogun::KDTree> kd_tree = solver->get_kd_tree();
-		//kd_tree->query_knn(query->as<DenseFeatures<float64_t>>(), m_k);
-		
 		solver->compute_nearest_neighbours(distance);
 		NN = solver->get_nearest_neighbours();
-
 		break;
 	}
 	}
@@ -315,8 +242,6 @@ SGMatrix<int32_t> KNN::classify_for_multiple_k()
 
 	io::info("{} test examples", num_lab);
 
-	//init_solver(m_knn_solver);
-
 	SGVector<int32_t> output = solver->classify_objects_k(distance, num_lab, train_lab, classes);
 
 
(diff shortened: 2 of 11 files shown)


src/shogun/multiclass/KNN.cpp
```diff
@@ -123,33 +124,53 @@ SGMatrix<index_t> KNN::nearest_neighbors()
...
+			//fill in an array with 0..num train examples-1
+			for (int32_t j=0; j<m_train_labels.vlen; j++)
+				train_idxs[j]=j;
 
-		//sort the distance vector between test example i and all train examples
-		Math::qsort_index(dists.vector, train_idxs.vector, m_train_labels.vlen);
+			//sort the distance vector between test example i and all train examples
+			Math::qsort_index(dists.vector, train_idxs.vector, m_train_labels.vlen);
 
 #ifdef DEBUG_KNN
```
vigsterkr: i'd remove this

src/shogun/multiclass/KNN.cpp
```diff
@@ -123,33 +124,53 @@ SGMatrix<index_t> KNN::nearest_neighbors()
...
-		io::print("\nQuick sort query {}\n", i);
-		for (int32_t j=0; j<m_k; j++)
-			io::print("{} ", train_idxs[j]);
-		io::print("\n");
+			io::print("\nQuick sort query {}\n", i);
```
vigsterkr: change this to SG_DEBUG

vigsterkr: remove \n from the end of the string

src/shogun/multiclass/KNN.cpp
```diff
@@ -123,33 +124,53 @@ SGMatrix<index_t> KNN::nearest_neighbors()
...
+			for (int32_t j=0; j<m_k; j++)
+				io::print("{} ", train_idxs[j]);
```
vigsterkr: use SG_DEBUG instead of io::print

vigsterkr (changes requested):
good start! even though i would continue the refactor as this is a DistanceMachine which is essentially a Machine, that has a train interface. the idea is that in `nearest_neighbors` you dont (re)train when you query the nearest neighbours, but actually use the trained KNN to do the query.

Hephaestus12:
> good start! even though i would continue the refactor as this is a DistanceMachine which is essentially a Machine, that has a train interface. the idea is that in `nearest_neighbors` you dont (re)train when you query the nearest neighbors, but actually use the trained KNN to do the query.

The `solver`(of type `KNNSolver`) for the query is only made when `init_solver()` is called from `apply_multiclass()`.

Moreover, the instance of the `KDTree` is only made when `classify_objects()`(defined in the `KDTreeKNNSolver` class) is called from the `apply_multiclass()` function.

Should I just use the already defined `solver`(which is defined as a protected member variable) and call the `KDTree` instance from there?
Should I somehow send the `KDTree` instance back to the `KNN` class so that `kd_tree->get_knn_indices()` can be directly called?

I'm sorry for the confusion, just that I don't see this `nearest_neighbours()` function being called anywhere or an example so I didn't know if it was going to be called before or after the `apply_multiclass()` function.

src/shogun/multiclass/KNN.cpp
```diff
@@ -104,7 +105,7 @@ bool KNN::train_machine(std::shared_ptr<Features> data)
 	return true;
 }
 
-SGMatrix<index_t> KNN::nearest_neighbors()
+SGMatrix<index_t> KNN::nearest_neighbors(KNN_SOLVER knn_solver/*=KNN_BRUTE*/)
```
karlnapf: not sure I understand ...

src/shogun/multiclass/KNN.cpp
```diff
@@ -123,33 +124,53 @@ SGMatrix<index_t> KNN::nearest_neighbors()
...
+			for (int32_t j=0; j<m_k; j++)
+				io::print("{} ", train_idxs[j]);
+			io::print("\n");
 #endif
 
-		//fill in the output the indices of the nearest neighbors
-		for (int32_t j=0; j<m_k; j++)
-			NN(j,i) = train_idxs[j];
+			//fill in the output the indices of the nearest neighbors
+			for (int32_t j=0; j<m_k; j++)
```
karlnapf: no loop necessary here. `sg_memcpy` could do the job for example, or others as well

karlnapf:
> How should I add unit tests?

Try to come up with some very simple corner cases where you know what the results should be, and then maybe a single very simple but nontrivial example with actual content.
Does this work as expected (as say verified by a python notebook w a plot)?

src/shogun/multiclass/KNN.cpp
```diff
@@ -123,33 +124,53 @@ SGMatrix<index_t> KNN::nearest_neighbors()
...
+			COMPUTATION_CONTROLLERS
+			//lhs idx 0..num train examples-1 (i.e., all train examples) and rhs idx i
+			distances_lhs(dists,0,m_train_labels.vlen-1,i);
 
-		//fill in an array with 0..num train examples-1
-		for (int32_t j=0; j<m_train_labels.vlen; j++)
-			train_idxs[j]=j;
+			//fill in an array with 0..num train examples-1
+			for (int32_t j=0; j<m_train_labels.vlen; j++)
+				train_idxs[j]=j;
```
gf712: Since you are touching this code can you replace it with train_idxs.range_fill please? :)

src/shogun/multiclass/KNN.cpp
gf712: Loop could be replaced with SGVector to_string no?

src/shogun/multiclass/KNN.cpp
Hephaestus12: I converted the `nearest_neighbour()` unparameterized function to one with a single parameter of type `KNN_SOLVER`. The `/*=KNN_BRUTE*/` is just to let the reader know that the default value is `KNN_BRUTE` as stated in the `KNN.h` file.

src/shogun/multiclass/KNN.cpp
Hephaestus12: Okay, I'll change this.

src/shogun/multiclass/KNN.cpp
Hephaestus12: Okay will do.

Hephaestus12:
> Does this work as expected (as say verified by a python notebook w a plot)?

I haven't checked this, should I add the python notebook somewhere? or will this just be for personal verification?

src/shogun/multiclass/KNN.cpp
Hephaestus12: Or should I use the `display_vector() `function?

vigsterkr:
Thats deprecated, use to_string() and sg_debug

src/shogun/multiclass/KNN.cpp
Hephaestus12: How should I get a pointer to the ith row of the matrix for the first parameter of `sg_memcpy()`? Will something like `(NN.matrix+m_k*sizeof(index_t))` work?

Hephaestus12:
> > good start! even though i would continue the refactor as this is a DistanceMachine which is essentially a Machine, that has a train interface. the idea is that in `nearest_neighbors` you dont (re)train when you query the nearest neighbors, but actually use the trained KNN to do the query.
> 
> The `solver`(of type `KNNSolver`) for the query is only made when `init_solver()` is called from `apply_multiclass()`.
> 
> Moreover, the instance of the `KDTree` is only made when `classify_objects()`(defined in the `KDTreeKNNSolver` class) is called from the `apply_multiclass()` function.
> 
> Should I just use the already defined `solver`(which is defined as a protected member variable) and call the `KDTree` instance from there?
> Should I somehow send the `KDTree` instance back to the `KNN` class so that `kd_tree->get_knn_indices()` can be directly called?
> 
> I'm sorry for the confusion, just that I don't see this `nearest_neighbours()` function being called anywhere or an example so I didn't know if it was going to be called before or after the `apply_multiclass()` function.

@vigsterkr What can I do about this?

src/shogun/multiclass/KNN.cpp
Hephaestus12: Or should I use the `SGMatrix<T>::set_column()` function?

src/shogun/multiclass/KNN.cpp
karlnapf: no `display_vector`, but do what was suggested above

src/shogun/multiclass/KNN.cpp
karlnapf: that would do it as well!

Hephaestus12:
Is this fine?

tests/unit/multiclass/KNN_unittest.cc
```diff
@@ -226,3 +226,89 @@ TEST(KNN, classify_multiple_kdtree)
...
+
+
+TEST(KNN, nearest_neighbours_brute)
+{
+	int32_t feats = 2;
+
+	SGVector< float64_t > lab(5);
+	lab[0] = 0.0; lab[1] = 0.0; lab[2] = 1.0; lab[3] = 1.0; lab[4] = 0.0;
+
+	SGMatrix< float64_t > feat(feats, 5);
```
karlnapf: minor: we usually format this without spaces

tests/unit/multiclass/KNN_unittest.cc
```diff
@@ -226,3 +226,89 @@ TEST(KNN, classify_multiple_kdtree)
...
+	feat(0,0) = 2.0; feat(0,1) = 3.0; feat(0,2) = 6.0; feat(0,3) = 7.0; feat(0,4) = 3.5;
```
karlnapf: could you just use a list initializer for the matrix? easier to read

tests/unit/multiclass/KNN_unittest.cc
```diff
@@ -226,3 +226,89 @@ TEST(KNN, classify_multiple_kdtree)
...
+	feat(1,0) = 1.0; feat(1,1) = 1.0; feat(1,2) = 1.0; feat(1,3) = 1.0; feat(1,4) = 2.0;
+	
+	SGVector<index_t> train (4);
+	SGVector<index_t> test (1);
+	train[0] = 0; train[1] = 1; train[2] = 2; train[3] = 3;
```
karlnapf: also list initializer

tests/unit/multiclass/KNN_unittest.cc
```diff
@@ -226,3 +226,89 @@ TEST(KNN, classify_multiple_kdtree)
...
+
+	features_test->add_subset(test);
+	labels_test->add_subset(test);
+
+	auto dist = std::make_shared<EuclideanDistance>(features, features_test);
+	knn->set_distance(dist);
+
+	SGMatrix<index_t> NN = knn->nearest_neighbors();
+	
+	EXPECT_EQ(feat(0, NN(0,0)), 3.0); EXPECT_EQ(feat(1, NN(0,0)), 1.0);
```
karlnapf: I think you need to use a real equals for this to consistently work (at least check whether this macro works for reals)

tests/unit/multiclass/KNN_unittest.cc
```diff
@@ -226,3 +226,89 @@ TEST(KNN, classify_multiple_kdtree)
...
+	EXPECT_EQ(feat(0, NN(1,0)), 2.0); EXPECT_EQ(feat(1, NN(1,0)), 1.0);
```
karlnapf: where do these test results come from? Maybe write a comment.
Also, is this a good test case? What would it catch?

Hephaestus12: This is a small and non-trivial test case you had mentioned. The test results have been worked out by hand.

>Maybe write a comment.

Will, do.
> Also, is this a good test case? What would it catch?

I  don't quite understand what you mean by this? Since we are directly checking if the nearest neighbours are the right ones, any error would lead to a wrong answer right?

tests/unit/multiclass/KNN_unittest.cc
Hephaestus12: Okay.

tests/unit/multiclass/KNN_unittest.cc
Hephaestus12: Okay

tests/unit/multiclass/KNN_unittest.cc
Hephaestus12: Okay.

tests/unit/multiclass/KNN_unittest.cc
Hephaestus12: Replaced with `EXPECT_DOUBLE_EQ`

tests/unit/multiclass/KNN_unittest.cc
karlnapf: I could for example imagine making the test smaller while still testing exactly the same type of thing. But with a comment it should suffice

src/shogun/multiclass/KNN.cpp
```diff
@@ -123,33 +120,47 @@ SGMatrix<index_t> KNN::nearest_neighbors()
...
+		{
+			COMPUTATION_CONTROLLERS
+			//lhs idx 0..num train examples-1 (i.e., all train examples) and rhs idx i
+			distances_lhs(dists,0,m_train_labels.vlen-1,i);
+
+			//fill in an array with 0..num train examples-1
+			train_idxs.range_fill(0);
+
+			//sort the distance vector between test example i and all train examples
+			Math::qsort_index(dists.vector, train_idxs.vector, m_train_labels.vlen);
```
karlnapf: could you use c++ sorting here? we can to eventually drop the math class

src/shogun/multiclass/KNN.cpp
```diff
@@ -123,33 +120,47 @@ SGMatrix<index_t> KNN::nearest_neighbors()
...
+
+			SG_DEBUG("\nQuick sort query {}", i);
+			SG_DEBUG("{}", train_idxs.to_string());
+			SG_DEBUG("");
+
+			//fill in the output the indices of the nearest neighbors
+			sg_memcpy(&NN.matrix[m_k * i], train_idxs.vector, sizeof(index_t) * m_k);
```
karlnapf: there are methods in sgmatrix/sgvector that do this I think? (make it more readable)

tests/unit/multiclass/KNN_unittest.cc
```diff
@@ -226,3 +226,79 @@ TEST(KNN, classify_multiple_kdtree)
 
 
 }
+
+
+TEST(KNN, nearest_neighbours_brute)
+{
+	//simple non-trivial test
+	//initialize labels
```
karlnapf: minor nitpick: remove the comment (redundant)

tests/unit/multiclass/KNN_unittest.cc
```diff
@@ -226,3 +226,79 @@ TEST(KNN, classify_multiple_kdtree)
...
+	SGVector<float64_t> lab({0.0, 0.0, 1.0, 1.0, 0.0});
+	//initialize features
+	SGMatrix<float64_t> feat({{2.0, 1.0}, {3.0, 1.0}, {6.0, 1.0}, {7.0, 1.0}, {3.5, 2.0}});
+	//divide training and testing parts
+	SGVector<index_t> train ({0, 1, 2, 3});
+	SGVector<index_t> test ({4});
+
+	auto labels = std::make_shared<MulticlassLabels>(lab);
+	auto features = std::make_shared<DenseFeatures<float64_t> >(feat);
+	auto features_test = features->clone()->as<DotFeatures>();
```
karlnapf: why not use the same features then? why do we need to clone?

tests/unit/multiclass/KNN_unittest.cc
```diff
@@ -226,3 +226,79 @@ TEST(KNN, classify_multiple_kdtree)
...
+
+	int32_t k=3;
+	auto distance = std::make_shared<EuclideanDistance>();
+	auto knn=std::make_shared<KNN> (k, distance, labels, KNN_BRUTE);
+
+	features->add_subset(train);
+	labels->add_subset(train);
+	knn->train(features);
+
+	features_test->add_subset(test);
```
karlnapf: I see.
Let's not use this method, as it is deprecated. There is a `view` function somewhere, that creates a new instance using the name underlying memory

tests/unit/multiclass/KNN_unittest.cc
```diff
@@ -226,3 +226,79 @@ TEST(KNN, classify_multiple_kdtree)
...
+	labels_test->add_subset(test);
+
+	auto dist = std::make_shared<EuclideanDistance>(features, features_test);
+	knn->set_distance(dist);
+
+	SGMatrix<index_t> NN = knn->nearest_neighbors();
+	
+	EXPECT_DOUBLE_EQ(feat(0, NN(0,0)), 3.0); EXPECT_DOUBLE_EQ(feat(1, NN(0,0)), 1.0);
```
karlnapf: picking up on our discussion, could you make this easier to read

* a comment might help
* fewer features might help
* it should be immediately obvious why you assert this

tests/unit/multiclass/KNN_unittest.cc
```diff
@@ -226,3 +226,79 @@ TEST(KNN, classify_multiple_kdtree)
...
+
+	features->add_subset(train);
+	labels->add_subset(train);
+	knn->train(features);
+
+	features_test->add_subset(test);
+	labels_test->add_subset(test);
+
+	auto dist = std::make_shared<EuclideanDistance>(features, features_test);
+	knn->set_distance(dist);
```
karlnapf: could you use `put` ? (working towards plugins) and use a constexpr for the name

src/shogun/multiclass/KNN.cpp
Hephaestus12: Yes, that is true. There is a `set_column()` function that I can use.
However, for that, I will need to reduce the size of the `train_idxs` vector to`m_k` or,
copy just the first `m_k` elements to a new array before using the `set_column()` function.
Is that okay?

karlnapf: You could create a new SGVector with a smaller length on the same memory block (and disable the memory management via the boolean flag in the ctor), and then pass that.

karlnapf: (definitely no copying)

tests/unit/multiclass/KNN_unittest.cc
Hephaestus12: Done.

tests/unit/multiclass/KNN_unittest.cc
Hephaestus12: Done.

src/shogun/multiclass/KNN.cpp
Hephaestus12: Okay.

src/shogun/multiclass/KNN.cpp
Hephaestus12: Is this fine?

tests/unit/multiclass/KNN_unittest.cc
Hephaestus12: Is this fine?

tests/unit/multiclass/KNN_unittest.cc
Hephaestus12: is this fine?

tests/unit/multiclass/KNN_unittest.cc
Hephaestus12: I'm sorry, I didn't quite understand what you mean by this.
I shouldn't use the `set_distance()` function?

karlnapf: you can use the `put` method instead, as once we have plugins, you only have the `Machine` interface in here

src/shogun/multiclass/KNN.cpp
```diff
@@ -123,33 +121,66 @@ SGMatrix<index_t> KNN::nearest_neighbors()
...
+    		// Storing the respective array 
+    		// elements in pairs. 
+    		for (int j = 0; j < m_train_labels.vlen; j++)  
+    		{ 
+        		pairt[j].first = dists[j]; 
+        		pairt[j].second = train_idxs[j]; 
+    		} 
+  
+    		// Sorting the pair array. 
+   			 sort(pairt, pairt + m_train_labels.vlen);
```
karlnapf: whitespace

and does this compile?

Hephaestus12: > whitespace

Really sorry about that.
> and does this compile?

But yes, this does compile. Even the tests passed successfully.

Hephaestus12:
Is this fine?

src/shogun/multiclass/KNN.cpp
```diff
@@ -123,33 +121,66 @@ SGMatrix<index_t> KNN::nearest_neighbors()
...
+  
+			// Storing the respective array 
+			// elements in pairs. 
+			for (int j = 0; j < m_train_labels.vlen; j++)  
+			{ 
+				pairt[j].first = dists[j]; 
+				pairt[j].second = train_idxs[j]; 
+			} 
+
+			// Sorting the pair array.
```
karlnapf: minor nitpick: no need for this comment. The code shows it

src/shogun/multiclass/KNN.cpp
```diff
@@ -123,33 +121,66 @@ SGMatrix<index_t> KNN::nearest_neighbors()
...
+			for (int j = 0; j < m_train_labels.vlen; j++)  
+			{ 
+				pairt[j].first = dists[j]; 
+				pairt[j].second = train_idxs[j]; 
+			} 
+
+			// Sorting the pair array. 
+			sort(pairt, pairt + m_train_labels.vlen); 
+
+			// Modifying original arrays
```
karlnapf: same here, not a useful comment

src/shogun/multiclass/KNN.cpp
```diff
@@ -123,33 +121,66 @@ SGMatrix<index_t> KNN::nearest_neighbors()
...
+		for (auto i : SG_PROGRESS(range(n)))
+		{
+			COMPUTATION_CONTROLLERS
+			//lhs idx 0..num train examples-1 (i.e., all train examples) and rhs idx i
+			distances_lhs(dists,0,m_train_labels.vlen-1,i);
+
+			//fill in an array with 0..num train examples-1
+			train_idxs.range_fill(0);
+
+			std::pair<float64_t, index_t> pairt[m_train_labels.vlen];
```
karlnapf: do you mean `paired` here?

src/shogun/multiclass/KNN.cpp
```diff
@@ -123,33 +121,66 @@ SGMatrix<index_t> KNN::nearest_neighbors()
...
+	case KNN_BRUTE:
+	{
+		//for each test example
+		for (auto i : SG_PROGRESS(range(n)))
+		{
+			COMPUTATION_CONTROLLERS
+			//lhs idx 0..num train examples-1 (i.e., all train examples) and rhs idx i
+			distances_lhs(dists,0,m_train_labels.vlen-1,i);
+
+			//fill in an array with 0..num train examples-1
```
karlnapf: another comment to remove

src/shogun/multiclass/KNN.cpp
```diff
@@ -123,33 +121,66 @@ SGMatrix<index_t> KNN::nearest_neighbors()
...
-		//fill in the output the indices of the nearest neighbors
-		for (int32_t j=0; j<m_k; j++)
-			NN(j,i) = train_idxs[j];
+	case KNN_BRUTE:
+	{
+		//for each test example
+		for (auto i : SG_PROGRESS(range(n)))
+		{
+			COMPUTATION_CONTROLLERS
+			//lhs idx 0..num train examples-1 (i.e., all train examples) and rhs idx i
```
karlnapf: and this one. also whitespace issues below

src/shogun/multiclass/KNN.cpp
```diff
@@ -123,33 +121,66 @@ SGMatrix<index_t> KNN::nearest_neighbors()
...
-			io::print("{} ", train_idxs[j]);
-		io::print("\n");
-#endif
-
-		//fill in the output the indices of the nearest neighbors
-		for (int32_t j=0; j<m_k; j++)
-			NN(j,i) = train_idxs[j];
+	case KNN_BRUTE:
+	{
+		//for each test example
```
karlnapf: remove

src/shogun/multiclass/KNN.cpp
```diff
@@ -123,33 +121,66 @@ SGMatrix<index_t> KNN::nearest_neighbors()
...
+				train_idxs[j] = pairt[j].second; 
+			} 
+
+			SG_DEBUG("\nQuick sort query {}", i);
+			SG_DEBUG("{}", train_idxs.to_string());
+			SG_DEBUG("");
+
+			SGVector<index_t> nearest_k_train_idxs(train_idxs.vector, m_k, 0);
+
+			//fill in the output the indices of the nearest neighbors
```
karlnapf: remove

src/shogun/multiclass/KNN.cpp
```diff
@@ -123,33 +121,66 @@ SGMatrix<index_t> KNN::nearest_neighbors()
...
+			{ 
+				dists[j] = pairt[j].first; 
+				train_idxs[j] = pairt[j].second; 
+			} 
+
+			SG_DEBUG("\nQuick sort query {}", i);
+			SG_DEBUG("{}", train_idxs.to_string());
+			SG_DEBUG("");
+
+			SGVector<index_t> nearest_k_train_idxs(train_idxs.vector, m_k, 0);
```
karlnapf: Could you do `false`  rather than 0?

karlnapf: a comment here is useful, "view on first k elements",

src/shogun/multiclass/KNN.cpp
```diff
@@ -123,33 +121,66 @@ SGMatrix<index_t> KNN::nearest_neighbors()
...
+			// Modifying original arrays 
+			for (int j = 0; j < m_train_labels.vlen; j++)  
+			{ 
+				dists[j] = pairt[j].first; 
+				train_idxs[j] = pairt[j].second; 
+			} 
+
+			SG_DEBUG("\nQuick sort query {}", i);
+			SG_DEBUG("{}", train_idxs.to_string());
+			SG_DEBUG("");
```
karlnapf: remove the empty line print

src/shogun/multiclass/KNN.cpp
```diff
@@ -123,33 +121,66 @@ SGMatrix<index_t> KNN::nearest_neighbors()
...
+			// Storing the respective array 
+			// elements in pairs. 
+			for (int j = 0; j < m_train_labels.vlen; j++)  
+			{ 
+				pairt[j].first = dists[j]; 
+				pairt[j].second = train_idxs[j]; 
+			} 
+
+			// Sorting the pair array. 
+			sort(pairt, pairt + m_train_labels.vlen);
```
karlnapf: so if this compiles, this means that there is a `using namespace std` lingering somehwere, which is bad. Could you check where that is? Need to follow the includes ... best use grep or something

src/shogun/multiclass/KNN.cpp
```diff
@@ -123,33 +121,66 @@ SGMatrix<index_t> KNN::nearest_neighbors()
...
+		}
+
+		distance->reset_precompute();
+		break;
 	}
+	case KNN_KDTREE:
+	{
+		auto lhs = distance->get_lhs();
+		auto kd_tree = std::make_shared<KDTree>(m_leaf_size);
+		kd_tree->build_tree(lhs->as<DenseFeatures<float64_t>>());
```
karlnapf: I have a stupid question. If you build the tree in here, as opposed to say in the train method, what do we gain? KD-trees make it fast to *look up* nearest neighbors once precomputed....

tests/unit/multiclass/KNN_unittest.cc
```diff
@@ -226,3 +227,67 @@ TEST(KNN, classify_multiple_kdtree)
 
 
 }
+
+
+TEST(KNN, nearest_neighbours_brute)
```
karlnapf: ok now that the nests are a bit cleaner, it is time to remove redundant code. Both tests are almost identical, so what you need to do is to use googletest to generate the code for both test cases for you. See eg this tutorial on how to do this:
http://sandordargo.com/blog/2019/04/24/parameterized-testing-with-gtest

It is also done in many shogun unit tests, so there are plenty of examples

karlnapf:
> Is this fine?

we get emails when you update the PR, so you dont need to cause another email being sent by asking straight away after you have pushed. Feel free to ping us if nothing happens for a few day, but this way you just create spam in our inboxes.

src/shogun/multiclass/KNN.cpp
Hephaestus12: If you are referring to the identifier, I meant pair-t which to think of, is really silly.
I can change it to anything else.

src/shogun/multiclass/KNN.cpp
Hephaestus12: `using namespace std;` is found in 22 files in the `src/shogun` directory out of which 11 instances are in the `src/shogun/io` directory.
Should I remove them and refactor the code accordingly?

Or should that be in another PR?

src/shogun/multiclass/KNN.cpp
Hephaestus12: No, we don't gain anything at all by building the tree in the `nearest_neighbours() `function. I just did this to prevent changing a lot of existing code.

Currently the KD-tree is built in two functions:
i) `nearest_neighbours()` 
ii) when `classify_objects()` (defined in the `KDTreeKNNSolver` class) is called from the `apply_multiclass()` function.

I'm assuming `nearest_neighbours()` will be called before `apply_multiclass()` is called. Hence it won't be possible to use that instance of the KD-tree.
One solution is to build the KD-tree in the `train()` function after checking if the `m_knn_solver` is of type `KNN_KDTree`. However, we'll need to think of some way to store it in some sort of a member variable and then send it to the `classify_objects()` function defined in the `KDTreeKNNSolver` class.

src/shogun/multiclass/KNN.cpp
Hephaestus12: Removed.

src/shogun/multiclass/KNN.cpp
Hephaestus12: Removed.

src/shogun/multiclass/KNN.cpp
Hephaestus12: Removed.

src/shogun/multiclass/KNN.cpp
Hephaestus12: Removed.
Whitespace issues solved.

src/shogun/multiclass/KNN.cpp
Hephaestus12: Removed.

src/shogun/multiclass/KNN.cpp
gf712: I am not sure I understand what the code does?

Hephaestus12: Basically, I'm merging the two arrays into an array of pairs before sorting it so that the `trairtn_idxs` gets sorted according to the ascending order of the corresponding distances.

src/shogun/multiclass/KNN.cpp
karlnapf: I think this calls for some kind of re-design of the KNN classes.

Definitely you want to build the tree in the train method, and then subsequently you use it in the apply method.

src/shogun/multiclass/KNN.cpp
karlnapf: It is ok for using the namespace in .cpp files. In header files we need to avoid it.
Could you find out which changes you need to make such that this file doesnt compile anymore?

Hephaestus12:
This isn't ready to merge yet. Is something like this better?
I'll move the nearest neighbour calculation for the BruteForce to the respective `BruteKNNSover` class.

However, I am getting one compile-time error I'm not able to understand:
```
Linking CXX executable statistical_testing/cpp-statistical_testing-quadratic_time_maximum_mean_discrepancy
../../../src/shogun/libshogun.so.18.0: undefined reference to `shogun::KNNSolver::train_KNN(std::shared_ptr<shogun::Distance>) const'
../../../src/shogun/libshogun.so.18.0: undefined reference to `typeinfo for shogun::KNNSolver'
../../../src/shogun/libshogun.so.18.0: undefined reference to `shogun::KNNSolver::compute_nearest_neighbours()'
../../../src/shogun/libshogun.so.18.0: undefined reference to `vtable for shogun::KNNSolver'
examples/meta/cpp/CMakeFiles/cpp-statistical_testing-quadratic_time_maximum_mean_discrepancy.dir/build.make:101: recipe for target 'examples/meta/cpp/statistical_testing/cpp-statistical_testing-quadratic_time_maximum_mean_discrepancy' failed
```

src/shogun/multiclass/KNNSolver.h
```diff
@@ -83,9 +84,15 @@ class KNNSolver : public DistanceMachine
 		 */
 		 virtual SGVector<int32_t> classify_objects_k(std::shared_ptr<Distance> d, const int32_t num_lab, SGVector<int32_t>& train_lab, SGVector<int32_t>& classes) const = 0;
 
+		virtual bool train_KNN(std::shared_ptr<Distance> knn_distance) const;
+
+		virtual bool compute_nearest_neighbours();
```
gf712: The linker error you are getting is because you declared these functions but never defined them. You just have the definition in the derived class right? In that case these should be pure virtual functions (whether that is good design, I don’t know). You could also CRTP instead.

Hephaestus12:
@karlnapf I am working on this, however, there is a design-related dilemma I am stuck in.
To build the KD-tree, we need the test features. If you are saying that we need to build the KD-tree when we train the KNN object, then we'll need to pass the testing features(the data-points from which to find the nearest neighbours) to the object when we train itself. Won't that be incorrect/ bad design?

Also, I'll need to make some changes to the `CoverTreeKNNSolver` class which is in the `src/gpl` directory. What should I do about that?

karlnapf:
> @karlnapf I am working on this, however, there is a design-related dilemma I am stuck in.
> To build the KD-tree, we need the test features. If you are saying that we need to build the KD-tree when we train the KNN object, then we'll need to pass the testing features(the data-points from which to find the nearest neighbours) to the object when we train itself. Won't that be incorrect/ bad design?
> 
> Also, I'll need to make some changes to the `CoverTreeKNNSolver` class which is in the `src/gpl` directory. What should I do about that?

Maybe I understand this wrongly, but I think you will need to build the kd_tree at training time on the training features only, and then at test time you query that tree with all points from the test data. For that the tree structure doesnt need to change.

Re the gpl submodule, it is a git submodule, you can edit it locally, then you have to send a PR to the shogun gpl repository, and update the submodule version in the main develop repository. It works just like shogun-data. If you have questions, search online for git submodules and how they work.

src/shogun/multiclass/KNN.cpp
```diff
@@ -249,49 +315,56 @@ SGMatrix<int32_t> KNN::classify_for_multiple_k()
...
+	std::cout<<"entered KNN::init_solver(KNN_SOLVER knn_solver)\n";
 	switch (knn_solver)
 	{
 	case KNN_BRUTE:
 	{
-		SGMatrix<index_t> NN = nearest_neighbors();
-		solver = std::make_shared<BruteKNNSolver>(m_k, m_q, m_num_classes, m_min_label, m_train_labels, NN);
-
+		//SGMatrix<index_t> NN = nearest_neighbors();
+		solver = std::make_shared<BruteKNNSolver>(m_k, m_q, m_num_classes, m_min_label, m_train_labels);
```
Hephaestus12, line 290: I changed the code for the `BruteKNNSolver` class to accommodate the new design, however, for some reason the program control isn't entering the` BruteKNNSolver()` constructor to create the `solver` instance. Can someone please help me out?

Please excuse the unnecessary print statements and comments, they are just for debugging and will be removed when merging.

Hephaestus12:
> Maybe I understand this wrongly, but I think you will need to build the kd_tree at training time on the training features only, and then at test time you query that tree with all points from the test data. For that the tree structure doesnt need to change.

Thanks!
I did this and now the KDTreeKNNSolver is working perfectly :)

src/shogun/multiclass/KNN.cpp
karlnapf: use gdb, or `SG_DEBUG` that is way quicker than recompiling and printing

src/shogun/multiclass/BruteKNNSolver.cpp
```diff
@@ -6,21 +6,28 @@
 #include <shogun/base/progress.h>
 #include <shogun/lib/Signal.h>
 #include <shogun/multiclass/BruteKNNSolver.h>
+#include <shogun/multiclass/KNNSolver.h>
+#include <shogun/machine/DistanceMachine.h>
+
+#include <iostream>
```
karlnapf, line 12: remove

src/shogun/multiclass/BruteKNNSolver.cpp
```diff
@@ -6,21 +6,28 @@
...
+#include <algorithm>
 
 using namespace shogun;
 
-BruteKNNSolver::BruteKNNSolver(const int32_t k, const float64_t q, const int32_t num_classes, const int32_t min_label, const SGVector<int32_t> train_labels, const SGMatrix<index_t> NN):
+BruteKNNSolver::BruteKNNSolver(const int32_t k, const float64_t q, const int32_t num_classes, const int32_t min_label, const SGVector<int32_t> train_labels):
 KNNSolver(k, q, num_classes, min_label, train_labels)
 {
+	std::cout<<"entered BruteKNNSolver::constructor\n";
```
karlnapf, line 20: this is a classic candidate for a SG_DEBUG....although one wouldnt see it in but maybe another helper method

karlnapf (commented):
It is really hard to see what is going on right now.

Hephaestus12:
> It is really hard to see what is going on right now.

I'll clean up the code and push again.

Hephaestus12:
Just to clear out the flow of control:
first `KNN::KNN()` is called(constructor)
    this calls `KNN::init()`

once this is complete, we train the machine by calling:
`KNN::train_machine()` 
    which calls `KNN::init_solver()`
        This function makes a new `KNNSolver` instance of the respective type(Brute, KDTree, etc)
    We then use the `solver` to call the `train_KNN()` method of the respective class(Brute, KDTree,      etc)

Then to get the nearest neighbours, we call the following:
`KNN::nearest_neighbors()`
    which calls `compute_nearest_neighbours()` using the solver we created in the init_solver
    and then we call `get_nearest_neighbours()` to get the neighbours.

On the other hand, to classify, we call:
`KNN::apply_multiclass()`
    which calls `classify_objects()` of the respective `solver` type defined in the respective classes

Here, the program control is not able to make a `solver` in the `init_solver()` function for the `BRUTE_SOLVER` case. (It isn't even entering the `BruteKNNSolver()` constructor)

tests/unit/multiclass/KNN_unittest.cc
```diff
@@ -226,3 +227,63 @@ TEST(KNN, classify_multiple_kdtree)
...
+	knn->put("distance", (std::shared_ptr<shogun::Distance>)dist);
+
+	SGMatrix<index_t> NN = knn->nearest_neighbors();
+	
+	EXPECT_DOUBLE_EQ(feat(0, NN(0,0)), 3.0);	 //3.0 is closest to 3.5
+	EXPECT_DOUBLE_EQ(feat(0, NN(1,0)), 2.0);	 //2.0 is second-closest to 3.5
+	EXPECT_DOUBLE_EQ(feat(0, NN(2,0)), 6.0);	 //6.0 is third-closest to 3.5
+}
+
+INSTANTIATE_TEST_CASE_P(
```
karlnapf, line 284: does this pass the tests?
I think in addition to some unit testing, could you also make some simple plots on some 2d data. So we can see whether this really does meaningful stuff? Alternatively, you could test the results on a nontrivial but simple dataset against a reference lib. And also I think you could test whether the different solvers return the same values on simple nontrivial data?

PR #4995, opened 2020-04-04
Refactor C++ factories

LiuYuHui:
- [x] refactor all C++ factories in factory.h
- [x] Replace factory calls in meta examples
- [x] Fix notebooks
- [x] Fix python graphical examples


src/shogun/util/factory.h
```diff
@@ -99,7 +100,8 @@ namespace shogun
 	template <class T>
 	std::shared_ptr<Features> features(SGMatrix<T> mat)
```
vigsterkr: we might wanna move this to details namespace? @gf712

src/shogun/util/factory.h
```diff
@@ -333,5 +335,63 @@ namespace shogun
...
+	class StringsFeatures;
+	class FeatureSubset;
+
+	template <
+	    typename TypeName, typename... Args,
+	    std::enable_if_t<
+	        traits::is_any_of_v<
+	            TypeName, Features, StringsFeatures, FeatureSubset,
+	            FeatureSubset, CSVFile, LibSVMFile, PipelineBuilder, Kernel>,
+	        TypeName>* = nullptr>
```
gf712: I’d prefer having a static assert with a meaningful message in the function rather than SFINAE

src/shogun/util/factory.h
gf712: Agreed

gf712 (changes requested):

src/shogun/util/factory.h
```diff
@@ -333,5 +335,63 @@ namespace shogun
...
 	}
+
+	class StringsFeatures;
+	class FeatureSubset;
+
+	template <
+	    typename TypeName, typename... Args,
+	    std::enable_if_t<
+	        traits::is_any_of_v<
+	            TypeName, Features, StringsFeatures, FeatureSubset,
```
karlnapf: question: can we avoid having to write these class names in-place here, rather than putting them into a list somewhere else, together with other base type definitions?

src/shogun/util/factory.h
```diff
@@ -333,5 +335,63 @@ namespace shogun
...
+		{
+			return kernel(std::forward<Args>(args)...);
+		}
+	}
+
+	template <
+	    typename TypeName,
+	    std::enable_if_t<
+	        !traits::is_any_of_v<
+	            TypeName, Features, StringsFeatures, FeatureSubset,
```
karlnapf: eg the list appears again here (see comment above)

src/shogun/util/factory.h
vigsterkr: there was an initiative to put the base interface classes into base_types.h see sg_inferface

src/shogun/util/factory.h
```diff
@@ -333,5 +335,63 @@ namespace shogun
...
+	    std::enable_if_t<
+	        traits::is_any_of_v<
+	            TypeName, Features, StringsFeatures, FeatureSubset,
+	            FeatureSubset, CSVFile, LibSVMFile, PipelineBuilder, Kernel>,
+	        TypeName>* = nullptr>
+	std::shared_ptr<TypeName> create(Args&&... args)
+	{
+		if constexpr (std::is_same_v<TypeName, Features>)
+		{
+			return features(std::forward<Args>(args)...);
```
karlnapf: so then all these "old" methods will be hidden into somewhere and not be exposed anymore I guess?

karlnapf: ah this is the comment above by the others, nevermind :)

src/shogun/util/factory.h
karlnapf: ++

src/shogun/util/factory.h
```diff
@@ -333,5 +335,63 @@ namespace shogun
...
+		}
+		else if constexpr (std::is_same_v<TypeName, StringsFeatures>)
```
karlnapf: would it make sense to generative these if then else statement automatically somehow using the fact that we can one-to-one relate these factory names with the type?

karlnapf (commented):
great initiative @LiuYuHui :)

src/shogun/util/factory.h
LiuYuHui: The function of this SFINAE is to distinguish between two kinds of `create`, one accepts a name, the other one not.

src/shogun/util/factory.h
LiuYuHui: yes, i will be change those to details namespace when all refactor stuff is done, otherwise, there are too many errors :)

src/shogun/util/factory.h
gf712: Ah I see!

src/shogun/util/factory.h
```diff
@@ -333,5 +335,63 @@ namespace shogun
...
+	}
+
+	template <
+	    typename TypeName,
+	    std::enable_if_t<
+	        !traits::is_any_of_v<
+	            TypeName, Features, StringsFeatures, FeatureSubset,
+	            FeatureSubset, CSVFile, LibSVMFile, PipelineBuilder, Kernel>,
+	        TypeName>* = nullptr>
+	std::shared_ptr<TypeName> create(const std::string& name)
```
gf712: Could you have string_view for C++ and std::string for SWIG?

src/shogun/util/factory.h
gf712: you can also the overload the other way around. Specialise for string types, and just have this without any SFINAE. Should be easier to maintain.. And then have a static_assert at the end of the constexpr that throws a compile time error if it cannot handle. Downside is that you need to handle, const char*, std::string and std::string_view, or check if the type is implicitly convertible to std::string...

examples/meta/generator/targets/cpp.json
```diff
@@ -120,7 +120,13 @@
...
         },
         "StaticCall": "$typeName::$method($arguments)",
-        "GlobalCall": "$method($arguments)",
+        "GlobalCall": {
+            "Default": "$method($arguments)",
+            "create_features": "create<Features>($arguments)",
+            "create_distance": "create<Distance>($arguments)",
+            "create_machine": "create<Machine>($arguments)",
+            "create_csv": "create<CSVFile>($arguments)"
+        },
```
LiuYuHui, line 151: does it look good? if so, i will change other meta examples.

vigsterkr: 👍

karlnapf: :rocket:

src/shogun/util/factory.h
```diff
@@ -333,5 +336,55 @@ namespace shogun
...
+			return kernel(std::forward<Args>(args)...);
+		}
+		else
+		{
+			static_assert(
+			    (sizeof...(Args) == 1) ||
+			        (std::is_constructible_v<Args, std::string> && ...),
+			    "The create function should accept a string");
+			return create_object<TypeName>(
+			    (std::string{std::forward<Args>(args)}.c_str())...);
```
gf712, line 378: Perfect!

karlnapf (commented):
great. Let's deploy it for all targets! I like it that this both solves the naming conflict issues, and cleans up the c++ api!

src/shogun/util/factory.h
```diff
@@ -333,5 +336,55 @@ namespace shogun
...
+		else if constexpr (std::is_same_v<TypeName, Kernel>)
+		{
+			return kernel(std::forward<Args>(args)...);
+		}
+		else
+		{
+			static_assert(
+			    (sizeof...(Args) == 1) ||
+			        (std::is_constructible_v<Args, std::string> && ...),
+			    "The create function should accept a string");
```
karlnapf: nitpicking: `The create function requires a string argument.`

gf712:
> great. Let's deploy it for all targets! I like it that this both solves the naming conflict issues, and cleans up the c++ api!

Also means that we are freeing up the module name in Python. So we could at some point look into shogun.kernel.GaussianKernel() to make it more pythonic? This would work with a partial function, so still using factories behind the scenes. Or would that be too far off the unified interface idea?

karlnapf:
> > great. Let's deploy it for all targets! I like it that this both solves the naming conflict issues, and cleans up the c++ api!
> 
> Also means that we are freeing up the module name in Python. So we could at some point look into shogun.kernel.GaussianKernel() to make it more pythonic? This would work with a partial function, so still using factories behind the scenes. Or would that be too far off the unified interface idea?

I'd like that. I'd also like to actually drop the "Kernel" from "GaussianKernel" and (and use namespaces in c++)
`sg.create_kernel("Gaussian")`, `create<Kernel>("Gaussian")`
But maybe later?

vigsterkr:
the thing is that the name string is 'generated' from get_name()... so what you suggest @karlnapf that actually any kernel is ending with *Kernel suffix... which i'm supposing that atm is true, but what if not... it could be circumvented that the get_name() is somehow returning a string like ObjectName+BaseInterfaceName.... it gets tricky a bit though....:( but this would only work for Kernel? i suppose for you wouldn't like to have create<Machine>("LibSVM") but that would remain LibSVM in get_name and not LibSVMMachine.... or?

i see your point with GaussianKernel, but it's somewhat an edge case atm i feel. `create<Distance>("Euclidean")` would be maybe another candidate for the trick above.... what else?

karlnapf:
@vigsterkr  I also agree with what you say that automatically making these names is tricky.

 I was actually talking about renaming the c++ classes, and simply don't repeat the base class name in subclasses, in particular not if it is one of the API base classes. And then maybe have some thing like a more nested namespace system, like namespace kernel, which contains "Gaussian", "Spectrum", etc.
Not sure this is the best way to go about this, just thinking out loud that `kernel("GaussianKernel")` is not very pretty

Just browsing the doxygen api. Each of the cases below represents cluster of repeated base class names (there are many more...)
* F1Measure -> F1
* FITCInferenceMethod -> FITC
* MulticlassOVREvaluation -> MulticlassOVR
* MultitaskKernelMaskPairNormalizer -> MultitaskMaskPair
* SoftMaxLikelihood -> SoftMax
* StratifiedCrossValidationSplitting -> Stratified

vigsterkr:
yeah the namespace could be cleaned up, but note that the way both now and in the plugin system as well (see feature branch) classes are being created/registered for create_object method is via the `get_name` function. so no matter what namespace and class name refactor you do that's what matter.

karlnapf:
> yeah the namespace could be cleaned up, but note that the way both now and in the plugin system as well (see feature branch) classes are being created/registered for create_object method is via the `get_name` function. so no matter what namespace and class name refactor you do that's what matter.

ah yes that is true. In that case, I think it kinda sucks that we have a flat namespace for shogun object creation. Maybe in the future we can think about syncing: folder names, namespace names, and make `get_name` return something like `kernel.Gaussian`, and then have `create<Kernel>('kernel.Gaussian')` or `create_kernel('Gaussian')`. But later I guess :)

src/shogun/util/factory.h
```diff
@@ -315,12 +318,12 @@ namespace shogun
...
 	}
 
-	std::shared_ptr<File> csv_file(std::string fname, char rw = 'r')
+	std::shared_ptr<CSVFile> csv_file(std::string fname, char rw = 'r')
 	{
 		return std::make_shared<CSVFile>(fname.c_str(), rw);
 	}
 
-	std::shared_ptr<File> libsvm_file(std::string fname, char rw = 'r')
+	std::shared_ptr<LibSVMFile> libsvm_file(std::string fname, char rw = 'r')
```
gf712, line 324: Is there a reason why you changed the type here?

examples/meta/generator/targets/python.json
```diff
@@ -81,7 +81,7 @@
             "get_string": "$object.get($arguments)"
         },
         "StaticCall": "sg.$typeName.$method($arguments)",
-        "GlobalCall": "sg.$method($arguments$kwargs)",
+        "GlobalCall": {"Default": "sg.shogun.$method($arguments)"},
```
gf712: I am not sure I understand why you changed this?

LiuYuHui: i changed the examples/meta/generator/translate.py, so the"GlobalCall" should be a list.

src/shogun/util/factory.h
LiuYuHui:
``` cpp
template <typename TypeName, typename... Args>
std::shared_ptr<TypeName> create(Args&&... args)
```
The **TypeName** of return value should be the same as the **TypeName** of create<TypeName>.

src/shogun/util/factory.h
```diff
@@ -333,5 +336,113 @@ namespace shogun
...
+	// The function of those two class is to distinguish between string_feature
+	// and features_subset
+	class StringsFeatures;
+	class FeatureSubset;
+
+	template <typename TypeName, typename... Args>
+	std::shared_ptr<std::conditional_t<
+	    traits::is_any_of_v<TypeName, StringsFeatures, FeatureSubset>, Features,
+	    TypeName>>
+	create(Args... args)
```
LiuYuHui: The SWIG don't handle with rvalue reference, so i don't use `perfect forwarding ` in here.

gf712: I would hide the forwarding reference version with a macro guard.

examples/meta/generator/targets/cpp.json
```diff
@@ -120,7 +120,37 @@
             "get_string": "$object->get<std::string>($arguments)"
         },
         "StaticCall": "$typeName::$method($arguments)",
-        "GlobalCall": "$method($arguments)",
+        "GlobalCall": {
+            "Default": "$method($arguments)",
```
karlnapf, line 124: we really need to think about having the base classes defined in one global place rather than re-stating them over and over again. @gf712 maybe we could have them in a header file, and then a python script can parse that and take them out for situations like this one here?

src/interfaces/swig/factory.i
```diff
@@ -2,11 +2,41 @@
...
+%template(create_features) shogun::create_features<uint16_t>;
+%template(create_features) shogun::create_features<int64_t>;
+
 %template(features) shogun::features<uint16_t>;
 %template(features) shogun::features<int64_t>;
 #endif //SWIGJAVA
 
+%template(create_svm) shogun::create<shogun::SVM,std::string>;
+%template(create_evaluation) shogun::create<shogun::Evaluation, std::string>;
+%template(create_multiclass_strategy) shogun::create<shogun::MulticlassStrategy, std::string>;
```
karlnapf, line 66: same here ....

examples/meta/generator/targets/cpp.json
gf712: Hmm, I guess so. For plugins I think shogun-core would need to have all base class definitions?

karlnapf: Yep!

src/shogun/util/factory.h
```diff
@@ -337,5 +334,110 @@ namespace shogun
...
+			        (std::is_constructible_v<Args, std::string> && ...),
+			    "The create function requires a string argument.");
+			return create_object<TypeName>(
+			    (std::string{std::forward<Args>(args)}.c_str())...);
+		}
+	}
+
+	// The SWIG now only support one template parameter in variadic template, so
+	// we have to write those create_ wrapper function
+	std::shared_ptr<Features> create_features(
```
gf712: Could you hide the templates to fix SWIG's lack of support for variadic templates with `#ifdef SWIG`? There is no need to have this in the C++ library, imo. @vigsterkr?

src/shogun/util/factory.h
```diff
@@ -337,5 +334,110 @@ namespace shogun
 	{
 		return std::make_shared<PipelineBuilder>();
 	}
+	} // namespace details
+	template <typename TypeName, typename... Args>
+	std::shared_ptr<TypeName> create(Args... args)
```
gf712, line 341: What about
```suggestion
#ifdef SWIG
	std::shared_ptr<TypeName> create(Args... args)
#else
	std::shared_ptr<TypeName> create(Args&&... args)
#endif
```

src/interfaces/swig/shogun.i
```diff
@@ -403,8 +403,8 @@ PUT_ADD(LossFunction)
 PUT_ADD(StructuredModel)
 PUT_ADD(FactorType)
 
-%template(kernel) kernel<float64_t, float64_t>;
-%template(features) features<float64_t>;
+%template(create_kernel) create_kernel<float64_t, float64_t>;
+%template(create_features) create_features<float64_t>;
```
gf712: @vigsterkr @karlnapf would it make sense to move these to factory.i? Or is it required to be defined here in SWIG? Same with all the `PUT_ADD`..

karlnapf: yep move them. Think I tried that and had problems but I forgot the details :)

src/shogun/util/factory.h
LiuYuHui: i added `ifdef SWIG`, but it didn't compile

vigsterkr:
U meant ifndef or?

src/shogun/util/factory.h
gf712: What’s the error? Compiler or swig?

LiuYuHui: the error comes from swig, it seems like that after added `ifdef SWIG`, SWIG still use the `create(Args&&... args)`

karlnapf: `#ifndef` for the win

gf712: Like here?
https://github.com/shogun-toolbox/shogun/blob/1c235d83bc609080a9e62ecb1fc8580a0b7a6388/src/shogun/base/SGObject.h#L251-L255

LiuYuHui: yes, like this https://pastebin.com/qSKLfGiL, i don't know why it don't work

gf712: OK, I suspect that SWIG is not parsing the template and function signature separately. Try to write the template with the function signature, so you’d have template <typename... Args> twice

src/shogun/util/factory.h
LiuYuHui: after added `ifdef SWIG`, there are errors that those create_ can't be found

doc/ipython-notebooks/classification/Classification.ipynb
```diff
@@ -84,11 +84,11 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "shogun_feats_linear = sg.features(sg.csv_file(os.path.join(SHOGUN_DATA_DIR, 'toy/classifier_binary_2d_linear_features_train.dat')))\n",
-    "shogun_labels_linear = sg.labels(sg.csv_file(os.path.join(SHOGUN_DATA_DIR, 'toy/classifier_binary_2d_linear_labels_train.dat')))\n",
+    "shogun_feats_linear = sg.create_features(sg.create_csv(os.path.join(SHOGUN_DATA_DIR, 'toy/classifier_binary_2d_linear_features_train.dat')))\n",
```
gf712: @karlnapf @vigsterkr would it make sense to rename `create_csv` to `load_csv`. It goes against the pattern but create implies (imo) that we are creating a CSV file?

karlnapf: we could also name all of those `sg_*`The names before more were more ambiguous (i.e. also allow for casting/conversions). But not sure. `load_csv` is also good

vigsterkr: load_csv is good... as well as read_csv :) either is fine

gf712: yea, `read_csv` would be like pandas which is better

LiuYuHui: ok, i will change create_csv to read_csv.

gf712: for consistency you should do the same with all the other file parsers, imo

src/shogun/util/factory.h
```diff
@@ -337,5 +334,121 @@ namespace shogun
...
+	{
+		return create<Labels>(file);
+	}
+
+	std::shared_ptr<File> read_csv(std::string fname, char rw = 'r')
+	{
+		return create<CSVFile>(fname, rw);
+	}
+
+	std::shared_ptr<File> create_libsvm(std::string fname, char rw = 'r')
```
gf712: To stay consistent
```suggestion
	std::shared_ptr<File> read_libsvm(std::string fname, char rw = 'r')
```

src/shogun/util/factory.h
```diff
@@ -337,5 +334,121 @@ namespace shogun
...
+			    (sizeof...(Args) == 1) ||
+			        (std::is_constructible_v<Args, std::string> && ...),
+			    "The create function requires a string argument.");
+			return create_object<TypeName>(
+			    (std::string{std::forward<Args>(args)}.c_str())...);
+		}
+	}
+
+	// The SWIG now only support one template parameter in variadic template, so
+	// we have to write those create_ wrapper function
```
gf712: could you move these swig specific things to `factory.i` please? You can just write C++ code there using `%{...C++ code...%}` syntax

gf712 (commented):
I think this is almost ready. See comments :)

gf712:
LGTM, @karlnapf @vigsterkr?

karlnapf:
Yes looks good from my side.

gf712:
Great work! Thanks!

karlnapf:
++ really nice, well done!

src/shogun/util/factory.h
```diff
@@ -337,5 +334,71 @@ namespace shogun
 	{
 		return std::make_shared<PipelineBuilder>();
 	}
+	} // namespace details
+	// fix me
+	//#ifdef SWIG
+	template <typename TypeName, typename... Args>
+	std::shared_ptr<TypeName> create(Args... args)
+	//#else
```
karlnapf, line 342: this should be removed

tests/unit/utils/factory_unittest.cc
```diff
@@ -54,7 +55,7 @@ TEST(Factory, DISABLED_string_features_from_file)
 	file_save->close();
 
 	auto file_load = std::make_shared<CSVFile>(filename.c_str(), 'r');
-	auto obj = string_features(file_load, DNA, PT_CHAR);
+	auto obj = create_string_features(file_load, DNA, PT_CHAR);
```
karlnapf, line 58: I am a bit confused why we still have python style methods in c++ though :)

LiuYuHui: i forget to delete those codes, will repair in #5011.
https://github.com/shogun-toolbox/shogun/blob/376bf8f500d3856f5ffe6f145dff84cbf55de104/src/shogun/util/factory.h#L382-L402

PR #4997, opened 2020-04-04
[WIP]Make column iterator for SGMatrix

jonpsy:
Currently I'm getting this build error, I'm unsure if I should use pointer for m_mat. I'd like to know your thoughts ! :D. The unit-test case need some changes too, but first I want to clear the errors out.

https://pastebin.com/13tiQs3J

### DIFF
--- a/src/shogun/lib/SGMatrix.h
+++ b/src/shogun/lib/SGMatrix.h
@@ -41,7 +41,15 @@ template<class T> class SGMatrix : public SGReferencedData
 	public:
 		typedef RandomIterator<T> iterator;
 		typedef ConstRandomIterator<T> const_iterator;
+#ifndef SWIG	
+	public:
+		class column_iterator;
+		class const_column_iterator;
 
+	private:
+		template<class MatrixType>
+		struct column_struct;
+#endif
 	public:
 		typedef Eigen::Matrix<T,-1,-1,0,-1,-1> EigenMatrixXt;
 		typedef Eigen::Map<EigenMatrixXt,0,Eigen::Stride<0,0> > EigenMatrixXtMap;
@@ -258,6 +266,24 @@ template<class T> class SGMatrix : public SGReferencedData
 		/** Returns a const iterator to the element following the last element of the container. */
 		const_iterator end() const noexcept { return const_iterator(matrix + (num_rows * num_cols)); }
 
+		/** Returns an iterator to the first column of the container. */
+		column_iterator begin_column() noexcept { return column_iterator(*this); }
+
+		/** Returns an iterator to the column following the last column of the container. */
+		column_iterator end_column() noexcept { return column_iterator(*this,num_cols); }
+
+		/** Returns a const iterator to the first column of the container. */
+		const_column_iterator begin_column() const noexcept { return const_column_iterator(*this); }
+
+		/** Returns a const iterator to the column following the last column of the container. */
+		const_column_iterator end_column() const noexcept { return const_column_iterator(*this,num_cols); }
+
+		/** Returns a helper struct to help iterate over columns*/
+		auto columns() noexcept { return column_struct<decltype(*this)>(*this); }
+
+		/** Returns a helper struct to help const iterate over columns*/
+		auto columns() const noexcept { return column_struct<decltype(*this)>(*this); }
+			
 #endif // SWIG should skip this part
 
 		/** Get element at index
@@ -527,5 +553,264 @@ template<class T> class SGMatrix : public SGReferencedData
 		/** GPU Matrix structure. Stores pointer to the data on GPU. */
 		std::shared_ptr<GPUMemoryBase<T>> gpu_ptr;
 };
+#ifndef SWIG 
+    template<typename T>
+	class SGMatrix<T>::column_iterator
+	{
+	public:
+		// Iterator traits
+		using difference_type = index_t;
+		using value_type = SGVector<T>;
+		using pointer = SGVector<T>*;
+		using reference = SGVector<T>&;
+		using iterator_category = std::random_access_iterator_tag;
+
+		column_iterator(SGMatrix<T> mat, index_t idx = 0)
+			: m_mat(mat), m_col_idx(idx)
+		{
+		}
+		column_iterator(const column_iterator& other)
+			: m_mat(other.m_mat), m_col_idx(other.m_col_idx)
+		{
+		}
+
+		column_iterator& operator++()
+		{
+			++m_col_idx;
+			return *this;
+		}
+		column_iterator operator++(int)
+		{
+			column_iterator retval(*this);
+			++(*this);
+			return retval;
+		}
+		column_iterator& operator--()
+		{
+			--m_col_idx;
+			return *this;
+		}
+		column_iterator operator--(int)
+		{
+			column_iterator retval(*this);
+			--(*this);
+			return retval;
+		}
+
+		bool operator==(const column_iterator& other) const
+		{
+			return m_mat == other.m_mat && m_col_idx == other.m_col_idx;
+		}
+		bool operator!=(const column_iterator& other) const
+		{
+			return !(*this == other);
+		}
+
+		value_type operator*()
+		{
+			return m_mat.get_column(m_col_idx);
+		}
+
+		column_iterator& operator+=(difference_type d)
+		{
+			m_col_idx += d;
+			return *this;
+		}
+		column_iterator& operator-=(difference_type d)
+		{
+			m_col_idx -= d;
+			return *this;
+		}
+
+		column_iterator operator+(difference_type d) const
+		{
+			return column_iterator(m_mat, m_col_idx + d);
+		}
+		column_iterator operator-(difference_type d) const
+		{
+			return column_iterator(m_mat, m_col_idx - d);
+		}
+
+		value_type operator[](difference_type d) const
+		{
+			return *column_iterator(m_mat, m_col_idx + d);
+		}
+
+		bool operator<(const column_iterator& other) const
+		{
+			return m_col_idx < other.m_col_idx;
+		}
+		bool operator>(const column_iterator& other) const
+		{
+			return m_col_idx > other.m_col_idx;
+		}
+		bool operator<=(const column_iterator& other) const
+		{
+			return m_col_idx <= other.m_col_idx;
+		}
+		bool operator>=(const column_iterator& other) const
+		{
+			return m_col_idx >= other.m_col_idx;
+		}
+
+		difference_type operator-(const column_iterator& other) const
+		{
+			return m_col_idx - other.m_col_idx;
+		}
+
+	private:
+		SGMatrix<T> m_mat;
+		difference_type m_col_idx;
+	};
+	template<typename T>
+	class SGMatrix<T>::const_column_iterator
+	{
+	public:
+		// Iterator traits
+		using difference_type = index_t;
+		using value_type = const SGVector<T>;
+		using pointer = const SGVector<T>*;
+		using reference = const SGVector<T>&;
+		using iterator_category = std::random_access_iterator_tag;
+
+		const_column_iterator(const SGMatrix<T> mat, index_t idx = 0)
+			: m_mat(mat), m_col_idx(idx)
+		{
+		}
+		const_column_iterator(const const_column_iterator& other)
+			: m_mat(other.m_mat), m_col_idx(other.m_col_idx)
+		{
+		}
+
+		const_column_iterator& operator++()
+		{
+			++m_col_idx;
+			return *this;
+		}
+		const_column_iterator operator++(int)
+		{
+			const_column_iterator retval(*this);
+			++(*this);
+			return retval;
+		}
+		const_column_iterator& operator--()
+		{
+			--m_col_idx;
+			return *this;
+		}
+		const_column_iterator operator--(int)
+		{
+			const_column_iterator retval(*this);
+			--(*this);
+			return retval;
+		}
+
+		bool operator==(const const_column_iterator& other) const
+		{
+			return m_mat == other.m_mat && m_col_idx == other.m_col_idx;
+		}
+		bool operator!=(const const_column_iterator& other) const
+		{
+			return !(*this == other);
+		}
+
+		value_type operator*()
+		{
+			return m_mat.get_column(m_col_idx);
+		}
+
+		const_column_iterator& operator+=(difference_type d)
+		{
+			m_col_idx += d;
+			return *this;
+		}
+		const_column_iterator& operator-=(difference_type d)
+		{
+			m_col_idx -= d;
+			return *this;
+		}
+
+		const_column_iterator operator+(difference_type d) const
+		{
+			return const_column_iterator(m_mat, m_col_idx + d);
+		}
+		const_column_iterator operator-(difference_type d) const
+		{
+			return const_column_iterator(m_mat, m_col_idx - d);
+		}
+
+		value_type operator[](difference_type d) const
+		{
+			return *const_column_iterator(m_mat, m_col_idx + d);
+		}
+
+		bool operator<(const const_column_iterator& other) const
+		{
+			return m_col_idx < other.m_col_idx;
+		}
+		bool operator>(const const_column_iterator& other) const
+		{
+			return m_col_idx > other.m_col_idx;
+		}
+		bool operator<=(const const_column_iterator& other) const
+		{
+			return m_col_idx <= other.m_col_idx;
+		}
+		bool operator>=(const const_column_iterator& other) const
+		{
+			return m_col_idx >= other.m_col_idx;
+		}
+
+		difference_type operator-(const const_column_iterator& other) const
+		{
+			return m_col_idx - other.m_col_idx;
+		}
+
+	private:
+		const SGMatrix<T> m_mat;
+		difference_type m_col_idx;
+	};
+	template<typename T>
+	template<class MatrixType>
+	struct SGMatrix<T>::column_struct
+	{
+		column_struct(MatrixType mat):m_mat(mat)
+		{
+		}
+
+		auto begin()
+		{
+			return m_mat.begin_column();
+		}
+
+		auto end()
+		{
+			return m_mat.end_column();
+		}	
+
+		auto begin() const
+		{
+			return m_mat.begin_column();
+		}
+
+		auto end() const
+		{
+			return m_mat.end_column();
+		}	
+
+		auto cbegin() const
+		{
+			return m_mat.begin_column();
+		}
+
+		auto cend() const
+		{
+			return m_mat.end_column();
+		}	
+
+	private:
+		MatrixType m_mat;
+	};
+#endif
 }
 #endif // __SGMATRIX_H__
--- a/tests/unit/lib/SGMatrix_unittest.cc
+++ b/tests/unit/lib/SGMatrix_unittest.cc
@@ -808,8 +808,7 @@ TEST(SGMatrixTest, column_iterator)
 	for (auto vec : mat.columns())
 		linalg::add_scalar(vec, 1.0);
 
-	auto range = std::distance(begin_col,end_col);
-	while (range-- > 0)
-		EXPECT_EQ(*mat.get_column(range), *mat_copy.get_column(range) + 1.0);
+	for (auto col_idx = 0; col_idx < mat.num_cols; col_idx++)
+		EXPECT_EQ(*mat.get_column(col_idx), *mat_copy.get_column(col_idx) + 1.0);
 }
 
--- a/tests/unit/lib/SGMatrix_unittest.cc
+++ b/tests/unit/lib/SGMatrix_unittest.cc
@@ -761,3 +761,55 @@ TEST(SGMatrixTest,iterator)
 	for (auto v: mat)
 		EXPECT_EQ(mat[index++], v);
 }
+
+TEST(SGMatrixTest, column_iterator)
+{
+	constexpr index_t size = 5;
+	SGMatrix<float64_t> mat(size, size);
+	linalg::range_fill(mat, 1.0);
+
+	auto begin_col = mat.begin_column();
+	auto end_col = mat.end_column();
+
+	EXPECT_EQ(mat.num_cols, std::distance(begin_col, end_col));
+	EXPECT_EQ(mat.get_column(0), *begin_col);
+	++begin_col;
+	EXPECT_EQ(mat.get_column(1), *begin_col);
+	--begin_col;
+	EXPECT_EQ(mat.get_column(0), *begin_col++);
+	EXPECT_TRUE(begin_col != end_col);
+	++begin_col;
+	EXPECT_EQ(mat.get_column(2), *begin_col--);
+	--begin_col;
+	EXPECT_EQ(mat.get_column(0), *begin_col);
+	begin_col += 2;
+	EXPECT_EQ(mat.get_column(2), *begin_col);
+	begin_col -= 2;
+	EXPECT_EQ(mat.get_column(0), *begin_col);
+	EXPECT_EQ(mat.get_column(1), begin_col[1]);
+
+	auto new_itr = begin_col + 2;
+	EXPECT_EQ(mat.get_column(2), *new_itr);
+
+	// range-based loop should work as well
+	auto index = 0;
+	for (const auto& i : mat.columns())
+		EXPECT_EQ(i, mat.get_column(index++));
+
+	const SGMatrix<float64_t> const_mat(mat);
+
+	index = 0;
+	for (const auto& i : const_mat.columns())
+		EXPECT_EQ(i, const_mat.get_column(index++));
+
+	// Modifying value of column vector
+	auto mat_copy = mat.clone();
+
+	for (auto vec : mat.columns())
+		linalg::add_scalar(vec, 1.0);
+
+	auto range = std::distance(begin_col,end_col);
+	while (range-- > 0)
+		EXPECT_EQ(*mat.get_column(range), *mat_copy.get_column(range) + 1.0);
+}
+


src/shogun/lib/SGMatrix.h
```diff
@@ -258,6 +301,13 @@ template<class T> class SGMatrix : public SGReferencedData
 		/** Returns a const iterator to the element following the last element of the container. */
 		const_iterator end() const noexcept { return const_iterator(matrix + (num_rows * num_cols)); }
 
+		/** Returns an iterator to the first element of the container. */
+		column_iterator begin_column() noexcept { return column_iterator(matrix); }
```
vigsterkr:
```suggestion
		column_iterator begin_column() noexcept { return column_iterator(*this); }
```

vigsterkr: SGMatrix has no member called matrix...

tests/unit/lib/SGMatrix_unittest.cc
```diff
@@ -761,3 +761,32 @@ TEST(SGMatrixTest,iterator)
...
+	col_itr -= 1;
+	EXPECT_EQ(mat.get_column(2),*col_itr);
+	EXPECT_EQ(mat.get_column(3),col_itr[1]);
+	
+	auto new_itr = col_itr + 2;
+	EXPECT_EQ(mat.get_column(4),*new_itr);
+
+	//range-based loop should work as well
+	int idx;
+	std::for_each(mat.begin_column(),mat.end_column(),[&](SGVector<float64_t>* vec){ EXPECT_EQ(*vec,mat.get_column(idx++));});
```
vigsterkr:
```suggestion
	std::for_each(mat.begin_column(),mat.end_column(),[&idx](auto* vec){ EXPECT_EQ(*vec,mat.get_column(idx++));});
```

src/shogun/lib/SGMatrix.h
```diff
@@ -258,6 +301,13 @@ template<class T> class SGMatrix : public SGReferencedData
...
+
+		/** Returns an iterator to the element following the last column of the container. */
+		column_iterator end_column() noexcept { return column_iterator(matrix + (num_rows * num_cols)); }
```
vigsterkr: definitely not...

vigsterkr (changes requested):

src/shogun/lib/SGMatrix.h
jonpsy: Hmm 
```column_iterator end_column() noexcept{ return column_iterator(*this,num_cols); }``` ?

**Comment by karlnapf** [comment]
It doesnt compile?

File: src/shogun/lib/SGMatrix.h (hunk 3 above)
Comment (gf712) [reply src/shogun/lib/SGMatrix.h]: Yes

File: src/shogun/lib/SGMatrix.h [hunk 4]
```diff
@@ -41,6 +41,49 @@ template<class T> class SGMatrix : public SGReferencedData
...
+			bool operator < (const column_iterator &other) const { return m_idx < other.m_idx; }
+			bool operator > (const column_iterator &other) const { return m_idx > other.m_idx; }
+			bool operator <= (const column_iterator &other) const { return m_idx <= other.m_idx; }
+			bool operator >= (const column_iterator &other) const { return m_idx >= other.m_idx; }
+
+			difference_type operator - (const column_iterator &other) const {return m_idx - other.m_idx;}
+
+		private:
+			SGMatrix<T>& m_mat;
+			index_t m_idx;
```
gf712: Maybe call this m_column_idx, so it is clear what it is tracking

tests/unit/lib/SGMatrix_unittest.cc
```diff
@@ -761,3 +761,31 @@ TEST(SGMatrixTest,iterator)
...
+	col_itr -= 1;
+	EXPECT_EQ(mat.get_column(2),*col_itr);
+	EXPECT_EQ(mat.get_column(3),col_itr[1]);
+	
+	auto new_itr = col_itr + 2;
+	EXPECT_EQ(mat.get_column(4),*new_itr);
+
+	//range-based loop should work as well
+	int idx;
+	std::for_each(mat.begin_column(),mat.end_column(),[&idx](auto* vec){ EXPECT_EQ(*vec,mat.get_column(idx++));});
```
gf712: Imo std::for_each is sufficient. The tests above are a bit convoluted..

gf712:
> It doesnt compile?

@NanuSai has to fix the end_column() class function

tests/unit/lib/SGMatrix_unittest.cc
jonpsy: I'll fix that.

src/shogun/lib/SGMatrix.h
jonpsy: Sure.

jonpsy: How about ```m_col_idx``` ?

tests/unit/lib/SGMatrix_unittest.cc
gf712: Thanks :)

src/shogun/lib/SGMatrix.h
gf712: Sure!

tests/unit/lib/SGMatrix_unittest.cc
jonpsy: Have a check on them, see if it can be improved?

src/shogun/lib/SGMatrix.h
jonpsy: .begin_column() and .end_column() are working perfectly fine !.  There's a small build error during for_each comparison. Need to fix that and we're done.

tests/unit/lib/SGMatrix_unittest.cc
```diff
@@ -761,3 +761,37 @@ TEST(SGMatrixTest,iterator)
...
 		EXPECT_EQ(mat[index++], v);
 }
+
+TEST(SGMatrixTest,column_iterator)
+{
+	constexpr index_t size=5;
+	SGMatrix<float64_t> mat(size, size);
+	linalg::range_fill(mat, 1.0);
+
+	SGMatrix<float64_t>::column_iterator col_itr(mat);
```
jonpsy: instead of this, should I use 
``` 
auto begin_col = mat.begin_column();
auto end_col = mat.end_column();
EXPECT_EQ(mat.num_cols, std::distance(begin_col, end_col));
EXPECT_TRUE(begin_col != end_col);
``` 

Sounds better to me (I compiled and tested it, it works )

tests/unit/lib/SGMatrix_unittest.cc
```diff
@@ -761,3 +761,37 @@ TEST(SGMatrixTest,iterator)
...
+	EXPECT_EQ(mat.get_column(1),*col_itr);
+	EXPECT_EQ(mat.get_column(2),col_itr[1]);
+	
+	
+	auto new_itr = col_itr + 2;
+	EXPECT_EQ(mat.get_column(3),*new_itr);
+
+	//range-based loop should work as well
+	int idx;
+	std::for_each(mat.begin_column(),mat.end_column(),[&idx,mat](auto* vec){ EXPECT_EQ(*vec,mat.get_column(idx++));});
```
jonpsy: This doesn't work and throws this: https://pastebin.com/qArHsKcU

gf712: should be `std::for_each(mat.begin_column(),mat.end_column(),[&idx,mat](const auto& vec){ EXPECT_EQ(*vec,mat.get_column(idx++));}); `
You are passing a non-pointer type to a lambda expecting a pointer

jonpsy: Okay using that gives https://pastebin.com/XVvURTpb. 
I think for_each deferences it for us, so we don't need to use ```*v``` again. It compiles succesfully when we use ```v```

gf712: You are dereferencing a non-pointer type without an operator* overload. Yes, you have to write it as `v`

jonpsy: Right!.

tests/unit/lib/SGMatrix_unittest.cc
jonpsy: Okay, I implemented the same.

tests/unit/lib/SGMatrix_unittest.cc
```diff
@@ -761,3 +761,38 @@ TEST(SGMatrixTest,iterator)
...
+	begin_col -= 1;
+	EXPECT_EQ(mat.get_column(1),*begin_col);
+	EXPECT_EQ(mat.get_column(2),begin_col[1]);
+	
+	auto new_itr = begin_col + 2;
+	EXPECT_EQ(mat.get_column(3),*new_itr);
+
+	//range-based loop should work as well
+	int idx=0;
+	std::for_each(mat.begin_column(),mat.end_column(),[&idx,mat](const auto& vec){ EXPECT_EQ(vec,mat.get_column(idx++));});
```
gf712: doesn't this cover all the test above?

jonpsy: Hm, we do want to check if += , -= are working, right?

gf712: += would be handled by for_each no? Anyway, I guess more tests can’t hurt!

jonpsy: Haha yea, that's true. I meant += d btw. for_each increases by one only.And then there's also the deferencing one ```[]```

src/shogun/lib/SGMatrix.h
```diff
@@ -41,6 +41,49 @@ template<class T> class SGMatrix : public SGReferencedData
...
+
+			column_iterator operator + (difference_type d) const { return column_iterator(m_mat,m_col_idx+d); }
+			column_iterator operator - (difference_type d) const { return column_iterator(m_mat,m_col_idx-d); }
+
+			value_type operator [] (difference_type d) const { return *column_iterator(m_mat,m_col_idx+d); }
+
+			bool operator < (const column_iterator &other) const { return m_col_idx < other.m_col_idx; }
+			bool operator > (const column_iterator &other) const { return m_col_idx > other.m_col_idx; }
+			bool operator <= (const column_iterator &other) const { return m_col_idx <= other.m_col_idx; }
+			bool operator >= (const column_iterator &other) const { return m_col_idx >= other.m_col_idx; }
```
gf712: Could you clang format this please? There is a mixture of styles here, such as spaces in operators and references on the variable vs on the type

gf712 (changes requested):

src/shogun/lib/SGMatrix.h
jonpsy: Sure!

jonpsy:
Woops ! Coming right up!

jonpsy:
I was thinking we should make mat.columns() so that we can have something like.. 
```for(auto& i: mat.columns ()) ```

Shall we go for that in this PR? I think since we're already knees deep in it, let's do it.

**Comment by gf712** [comment]
> I was thinking we should make mat.columns() so that we can have something like..
> `for(auto& i: mat.columns ()) `
> 
> Shall we go for that in this PR? I think since we're already knees deep in it, let's do it.

Yes, got for it. Do you know how to do it? You basically need columns to return a wrapper with begin and end class functions.

**Comment by jonpsy** [comment]
I've made it, thanks for the guidance !.Unit tests are passing, maybe the description could be improved but other than that I think it's all good :!

**Comment by LiuYuHui** [comment]
> 
> 
> I've made it, thanks for the guidance !.Unit tests are passing, maybe the description could be improved but other than that I think it's all good :!

it seems like that there are issues related to the SWIG in your code, maybe it can be fixed by just adding the  `#ifndef SWIG ` :)

**Comment by jonpsy** [comment]
@LiuYuHui , thanks for the input. Correct me if I'm wrong, columns() method is already inside #ifndef,  I think you mean the implementation of "helper_struct"?.

File: src/shogun/lib/SGMatrix.h [hunk 10]
```diff
@@ -41,7 +41,244 @@ template<class T> class SGMatrix : public SGReferencedData
...
+			bool operator==(const const_column_iterator& other) const
+			{
+				return m_mat == other.m_mat && m_col_idx == other.m_col_idx;
+			}
+			bool operator!=(const const_column_iterator& other) const
+			{
+				return !(*this == other);
+			}
+
+			value_type operator*()
```
gf712: Here you can return reference

src/shogun/lib/SGMatrix.h
```diff
@@ -41,7 +41,244 @@ template<class T> class SGMatrix : public SGReferencedData
...
+			const_column_iterator operator+(difference_type d) const
+			{
+				return const_column_iterator(m_mat, m_col_idx + d);
+			}
+			const_column_iterator operator-(difference_type d) const
+			{
+				return const_column_iterator(m_mat, m_col_idx - d);
+			}
+
+			value_type operator[](difference_type d) const
```
gf712: Same here

src/shogun/lib/SGMatrix.h
```diff
@@ -41,7 +41,244 @@ template<class T> class SGMatrix : public SGReferencedData
...
+		
+			auto begin()
+			{
+				return m_mat.begin_column();
+			}
+
+			auto end()
+			{
+				return m_mat.end_column();
+			}
```
gf712: Maybe also add the const versions?

jonpsy: Oh right!, do I need another struct for this ?

jonpsy: Hmm, maybe I can overload inside the struct. I'll try and let y'know ! :D

gf712: No, you can just do the same thing as in the iterators below. Just add begin/end const and add a columns const in SGMatrix

src/shogun/lib/SGMatrix.h
jonpsy: Correct me if I'm wrong but m_mat.get_column(m_col_idx) returns an SGVector<T>, so it's a local variable. We can't return a reference of local variable from functions right? (We can but that involves using value_type).

jonpsy:
Okay I've added support for const iterator. After much thoughts, I don't think there's any way to go about this without creating two helper_struct. I've also added a unit test to check it's working.  Would love to hear your feedback !

gf712:
> Okay I've added support for const iterator. After much thoughts, I don't think there's any way to go about this without creating two helper_struct. I've also added a unit test to check it's working. Would love to hear your feedback !

technically you can, whether it is worth it is another question... @vigsterkr @karlnapf?
https://github.com/shogun-toolbox/shogun/blob/d80e9babce1f3f1da24e59e84b9025da77b2d483/src/shogun/util/container_iterators.h#L60-L67

src/shogun/lib/SGMatrix.h
gf712: Ah yes, sorry

gf712:
LGTM. @vigsterkr should the iterator definitions be inside of SGMatrix? They take up so much space (in terms of LoC). If we keep them there we should have a single iterator that uses `std::conditional_t` to distinguish between const versions, reducing LoC to pretty much half..

jonpsy:
Okay that!!! Yes that would be amazing to have!!! I was thinking the same on code size, personally Id love to implement it

gf712:
> Okay that!!! Yes that would be amazing to have!!! I was thinking the same on code size, personally Id love to implement it

OK, go ahead. You should just have to add an additional template parameter in the struct and the condition in the type definitions. You can do this with both the iterator and the helper struct

jonpsy:
If everything is good to go, now would be an excellent time to suggest a better name than "helper_struct".

src/shogun/lib/SGMatrix.h
```diff
@@ -809,5 +554,254 @@ template<class T> class SGMatrix : public SGReferencedData
...
+		helper_struct(ST& mat):m_mat(mat)
+		{
+		}
+
+		auto begin()
+		{
+			return m_mat.begin_column();
+		}
+
+		auto end()
```
vigsterkr, line 786: `auto begin() const` and `auto end() const` is still missing...

src/shogun/lib/SGMatrix.h
```diff
@@ -37,267 +37,19 @@ namespace shogun
...
-			}
+	public:
+		class column_iterator;
+		class const_column_iterator;
 
-		private:
-			SGMatrix<T>& m_mat;
-		};
+	private:
+		template<class ST>
```
vigsterkr:
```suggestion
		template<class MatrixType>
```

src/shogun/lib/SGMatrix.h
```diff
@@ -37,267 +37,19 @@ namespace shogun
 template<class T> class SGMatrix : public SGReferencedData
 {
 	friend class LinalgBackendEigen;
-
+
```
vigsterkr: noise

vigsterkr (changes requested):

src/shogun/lib/SGMatrix.h
```diff
@@ -527,5 +554,264 @@ template<class T> class SGMatrix : public SGReferencedData
...
+		}
+
+	private:
+		const SGMatrix<T>& m_mat;
+		difference_type m_col_idx;
+	};
+#ifndef SWIG
+	template<typename T>
+	template<class MatrixType>
+	struct SGMatrix<T>::helper_struct
```
gf712: I think just rename this to something more meaningful and it’s good to merge!

gf712 (changes requested):
Looks great, thanks!

tests/unit/lib/SGMatrix_unittest.cc
```diff
@@ -761,3 +761,45 @@ TEST(SGMatrixTest,iterator)
...
+	auto index = 0;
+	for(const auto& i : mat.columns())
+		EXPECT_EQ(i,mat.get_column(index++));
+	
+	const SGMatrix<float64_t> const_mat(mat);
+
+	index = 0;
+	for(const auto& i : const_mat.columns())
+		EXPECT_EQ(i,const_mat.get_column(index++));
+}
```
gf712, line 813: what about testing for modifying the value of the column vector?
```cpp
auto mat_copy = mat.clone();
for(auto& vec : mat_copy.columns())
	linalg::add(vec, 1, vec); //can't remember signature but basically vec += 1
for(const auto& [result, original] : zip_iterator(mat_copy.columns(), mat.columns()))
{
        auto reference = linalg::add(original, 1) // again can't remember signature but something like reference = original + 1 
	EXPECT_EQ(result, reference);
}
```

jonpsy: It's this
```c++
1	//Modifying value of column vector
2	auto mat_copy = mat.clone();
3
4	for(const auto& vec : mat_copy.columns())
5		mat_copy.set_column(index,linalg::add(vec,vec));
6	
7	for(const auto& [result, original] : zip_iterator(mat_copy.columns(),mat.columns()))
8	{
9			auto reference = linalg::add(original,original)
10			EXPECT_EQ(result, reference);
11	}
```
However zip_iterator is showing at line 7. Here's the pastebin https://pastebin.com/DfjXGkuR
```
no instance of constructor "zip_iterator" matches the argument list -- argument types are: (shogun::SGMatrix<float64_t>::column_struct<shogun::SGMatrix<float64_t> &>, shogun::SGMatrix<float64_t>::column_struct<shogun::SGMatrix<float64_t> &>)
```

src/shogun/lib/SGMatrix.h
```diff
@@ -527,5 +553,264 @@ template<class T> class SGMatrix : public SGReferencedData
...
+		difference_type operator-(const const_column_iterator& other) const
+		{
+			return m_col_idx - other.m_col_idx;
+		}
+
+	private:
+		const SGMatrix<T>& m_mat;
+		difference_type m_col_idx;
+	};
+#ifndef SWIG
```
karlnapf: note that all of the iterator code should be hidden from swig, not just parts

karlnapf (commented):
nice thanks for this

src/shogun/lib/SGMatrix.h
jonpsy: Ah cool, btw should I write "SWIG Should skip this part". I mean it's apparent anyway, thoughts?

karlnapf: no need imo

tests/unit/lib/SGMatrix_unittest.cc
```diff
@@ -761,3 +761,61 @@ TEST(SGMatrixTest,iterator)
...
+	for (const auto& vec : mat_copy.columns())
+		mat_copy.set_column(index++, linalg::add(vec, vec));
+
+	index = 0;
+	for (const auto& vec : mat.columns())
+		mat.set_column(index++, linalg::add(vec,vec));
+
+	auto idx1 = 0 ,idx2 =0;
+	while(index < std::distance(begin_col,end_col))
+		EXPECT_EQ(mat.get_column(idx1++),mat_copy.get_column(idx2++));
```
gf712: I think you missed the point of this test... You are now doing twice exactly the same and checking if the output is the same (which is a pretty basic assumption and I don't think we need to test for).
The test I suggested was to check whether updating the values in a loop was actually working when using the iterator. Please read the comment with the code snippet. What I meant was:
1. Mutate the vector from the iterator of a matrix
2. Compare in a separate loop so that we can be sure that mutating the vector returned by the iterator is actually changing the matrix itself and not a local copy. Hence you should either have a matrix with reference results or compute a reference value in a different way (e.g. do the same operation as you did in 1 with a local copy and compare that to the test matrix)

tests/unit/lib/SGMatrix_unittest.cc
```diff
@@ -761,3 +761,55 @@ TEST(SGMatrixTest,iterator)
...
+
+	// Modifying value of column vector
+	auto mat_copy = mat.clone();
+
+	for (auto vec : mat.columns())
+		linalg::add_scalar(vec, 1.0);
+
+	auto range = std::distance(begin_col,end_col);
+	while (range-- > 0)
+		EXPECT_EQ(*mat.get_column(range), *mat_copy.get_column(range) + 1.0);
```
jonpsy: I believe this covers "mutating column vector"
* mat_copy gets a local copy of the original matrix
* We use a for loop then modify the vector returned by mat.columns() by adding 1.0 to it.
* Finally, we assert if the original matrix is changed or not (by comparing the difference in values)

gf712: OK, looks good. But could you write the while as a for loop, please? It is just increasing the code complexity.. Why not just `for (auto col_idx = 0; col_idx < mat.n_cols; col_idx++)`?

gf712:
Let's just wait for CI and it is good from my side. @vigsterkr @karlnapf what do you think?

tests/unit/lib/SGMatrix_unittest.cc
```diff
@@ -761,3 +761,54 @@ TEST(SGMatrixTest,iterator)
...
+	const SGMatrix<float64_t> const_mat(mat);
+
+	index = 0;
+	for (const auto& i : const_mat.columns())
+		EXPECT_EQ(i, const_mat.get_column(index++));
+
+	// Modifying value of column vector
+	auto mat_copy = mat.clone();
+
+	for (auto vec : mat.columns())
```
gf712, line 808: actually can you make this `auto&`?

jonpsy: There's a problem, mat.columns() returns an rvalue.

gf712: ok, then just leave it like this

karlnapf: we could add one version that returns a reference to the temporary SGVector? Would that make sense @gf712 ?

tests/unit/lib/SGMatrix_unittest.cc
karlnapf: or `for (auto col_idx in range(mat.n_cols))`

src/shogun/lib/SGMatrix.h
```diff
@@ -41,7 +41,15 @@ template<class T> class SGMatrix : public SGReferencedData
 	public:
 		typedef RandomIterator<T> iterator;
 		typedef ConstRandomIterator<T> const_iterator;
+#ifndef SWIG
```
karlnapf, line 44: @gf712 do we have to put all this code into SGMatrix.h? It bloats the file quite a bit? I'd prefer to have it somewhere else, where other iterators sit. But you or @vigsterkr might see this differently?

jonpsy: Not a pro opinion but, I did try to have a go at pushing all this code to our ```<shogun/utils/iterators.h>```. This, however, creates a circular dependency issue because our iterator class  SGMatrix as an internal member and SGMatrix #includes iterator file.

karlnapf: Yes, I feared that. Maybe the others have an idea

gf712: I mean it is doable, not sure if it will help. So you could have the declaration in iterator.h with forward declaration in SGMatrix and then implement the iterator in SGMatrix.cpp. This should work.

tests/unit/lib/SGMatrix_unittest.cc
gf712: I don’t think that would work, because then that reference would be to a value that is stored somewhere. So the iterator would keep a “cached” vector, which could be fine. I mean the iterator is stateful anyway (we already store the index)

karlnapf:
What's the latest here? Would be cool to turn to get this in? :)

gf712:
I think it’s ready to be merged!

karlnapf:
Thanks a lot for the patch @NanuSai 
Want to apply the new iterator to a few places in the codebase?

jonpsy:
> Thanks a lot for the patch @NanuSai
> Want to apply the new iterator to a few places in the codebase?

I'm game!

PR #5001, opened 2020-04-07
virtual -> override

vinnik-dmitry07:
The build is successful. Unit tests 98% (as before changes). Everything works - I have tested many times.
image
\<offtop>Heiko, I follow your directions so I am working on the RBF networks\</offtop>


gf712:
Was this with clang tidy?

vinnik-dmitry07:
> Was this with clang tidy?

Yes Clang-Tidy is built in CLion. But it was not one click. I did several troubleshoots along the way: there was some bugs overriding a pure virtual method with a pure method: https://youtrack.jetbrains.com/issue/CPP-19811. Also I faced some strange compiler behavior: I really do not know why, but `noexcept(false)` and `override` do not work together.

gf712:
Interesting. Anyway, there is command line tool for clang tidy where you can just pass the rule you want to check and it will do everything. Not sure if you would have the same issues.

karlnapf:
LGTM

gf712 (changes requested):
@vinnik-dmitry07 you need to fix all of these, because now there are so many compiler warnings if you look at the CI.
```
/src/src/shogun/kernel/Kernel.h:946:16: warning: 'load_serializable_post' overrides a member function but is not marked 'override' [-Winconsistent-missing-override]
                virtual void load_serializable_post() noexcept(false);
                             ^
/src/src/shogun/base/SGObject.h:746:15: note: overridden virtual function is here
        virtual void load_serializable_post() noexcept(false);
```

karlnapf:
@vinnik-dmitry07 given that this is nontrivial, I suggest one last time that you park this PR for now and instead  work on some project related c++ code. Such cleanups are part of GSoC anyways

Issue #5000, opened 2020-04-07
Add a Poisson regression model to Shogun


Khalifa1997:
As needed for my GSoC project, it would be quite beneficial for me and future Shogun users aswell to have a possion regression model, especially for count-based problems. If it's possible I'd like to be working on this issue

karlnapf:
As said on irc, this is very welcome and useful. You will need to write a new machine class (potentially inheriting from LinearMachine but not necessarily). I think it might be nice to make this modular wrt likelihood (and link) functions, although we can certainly start without that (as long as the likelihood codes inside the class are separate from the rest). As you will need to use gradient descent to fit this model, make sure to check out Shogun's optimization algorithms, which you should use. Re gradient, you could start by implementing it by hand, and then make that also modular wrt the likelihood function. I suggest you start by sending class headers, a diagram how things are intended to interact, then some pseudo-code, and then we can take it from there. Check how other libs are dealing with design/modularity/optimization. You could also start by writing a Python (object oriented, roughly following Shogun's API) draft of everything if that helps.

karlnapf:
https://en.wikipedia.org/wiki/Poisson_regression

karlnapf:
Here is a list of common link functions and likelihoods: https://www.statsmodels.org/stable/glm.html

geektoni:
As I mentioned in the email, also this lib here could be used as an inspiration https://github.com/glm-tools/pyglmnet

Khalifa1997:
![Untitled Diagram](https://user-images.githubusercontent.com/26578518/78711106-bd46da00-7916-11ea-8d73-7549eb294492.jpg)
This is a class-diagram of what I propose, we'd inherit from the `Machine` class since we are using the `train()` and `apply()` functions anyways. I have added a question mark on exposure since I am not sure if its applicable to all GLMs. Also to further add onto that, I don't think that different GLMs would have anything that's unique to them so no need to implement a unique class per GLM (they would all just share the same data members namely: betas/weight)

vigsterkr:
linearmachine?

Khalifa1997:
> linearmachine?

if you mean inheriting from LinearMachine well the issue comes that it already has things like `bias` that aren't applicable at all -to my knowledge- to GLMs

geektoni:
> > linearmachine?
> 
> if you mean inheriting from LinearMachine well the issue comes that it already has things like `bias` that aren't applicable at all -to my knowledge- to GLMs

[...]

PR #5004, opened 2020-04-12
Added GLM

Khalifa1997:
in reponse to #5000 
The `train machine` function is currently empty, I will fill it up as soon as I get some feedback on a few things
In my head, I am using a `DescendUpdaterWithCorrection`/`DescendUpdater` inorder to have access to the function `update_variable` inside my class, yet. Which params should I pass to it, I know I could use `get_name()` to get which specific optimizer this and downcast it to this optimizer but I feel like there is a better way to go around it than this

Edit: I just saw that GLM.h has some extra doxygen comments, I will remove them on my next commit

### DIFF
--- a/src/shogun/regression/GLM.cpp
+++ b/src/shogun/regression/GLM.cpp
@@ -46,11 +46,8 @@ SGVector<float64_t> GLM::log_likelihood(
 void GLM::init()
 {
 	SG_ADD(
-	    &m_tau, "tau", "Regularization parameter",
+	    &m_tau, "tau", "L2 Regularization parameter",
 	    ParameterProperties::SETTING);
-	SG_ADD(
-	    &m_lambda, "lambda", "Regularization parameter lambda",
-	    ParameterProperties::HYPER);
 	SG_ADD(
 	    (std::shared_ptr<SGObject>*)&m_descend_updater, "descend_updater",
 	    "Descend Updater used for updating weights",
@@ -64,12 +61,10 @@ void GLM::init()
 }
 GLM::GLM(
     const std::shared_ptr<DescendUpdater>& descend_updater,
-    DistributionFamily family, LinkFunction link_fn, float64_t tau,
-    float64_t lambda)
+    DistributionFamily family, LinkFunction link_fn, float64_t tau)
     : LinearMachine()
 {
 	m_tau = tau;
-	m_lambda = lambda;
 	m_link_fn = link_fn;
 	m_descend_updater = descend_updater;
 	m_family = family;
--- a/src/shogun/regression/GLM.cpp
+++ b/src/shogun/regression/GLM.cpp
@@ -12,7 +12,7 @@ GLM::GLM() : LinearMachine()
 {
 	init();
 }
-SGVector<float64_t> log_likelihood(
+SGVector<float64_t> GLM::log_likelihood(
     const std::shared_ptr<DenseFeatures<float64_t>>& features,
     const std::shared_ptr<Labels>& label)
 {
@@ -21,15 +21,13 @@ SGVector<float64_t> log_likelihood(
 	ASSERT(vector_count > 0 && label->get_num_labels() == vector_count)
 	// Array of Lambdas
 	SGVector<float64_t> lambda(vector_count);
+	SGVector<float64_t> beta = LinearMachine::get_w();
+	float64_t beta0 = LinearMachine::get_bias();
 	for (auto i = 0; i < vector_count; i++)
 	{
 		SGVector<float64_t> feature_vector = features->get_feature_vector(i);
-		// Assume beta is the same as the feature vector
-		SGVector<float64_t> beta = feature_vector.clone();
-		// Assume beta0 is the same as the first element in the feature vector
-		float64_t b0 = feature_vector.get_element(0);
 		float64_t res = linalg::dot(feature_vector, beta);
-		lambda.set_element(log(1 + std::exp(b0 + res)), i);
+		lambda.set_element(log(1 + std::exp(beta0 + res)), i);
 	}
 	SGVector<float64_t> likelihood(vector_count);
 	SGVector<float64_t> labels = label->get_values();
@@ -42,15 +40,14 @@ SGVector<float64_t> log_likelihood(
 	    linalg::element_prod(labels, log_lambda), lambda, 1.0, -1.0);
 	SGVector<float64_t> likelihood_clone = likelihood.clone();
 	for (auto i = 0; i < vector_count; i++)
-		likelihood.set_element(
-		    SGVector<float64_t>::sum(likelihood_clone, i), i);
+		likelihood.set_element(linalg::sum(likelihood_clone.slice(0, i)), i);
 	return likelihood;
 }
 void GLM::init()
 {
 	SG_ADD(
-	    &m_alpha, "alpha", "Weighting parameter between L1 and L2 Penalty",
-	    ParameterProperties::HYPER);
+	    &m_tau, "tau", "Regularization parameter",
+	    ParameterProperties::SETTING);
 	SG_ADD(
 	    &m_lambda, "lambda", "Regularization parameter lambda",
 	    ParameterProperties::HYPER);
@@ -67,11 +64,11 @@ void GLM::init()
 }
 GLM::GLM(
     const std::shared_ptr<DescendUpdater>& descend_updater,
-    DistributionFamily family, LinkFunction link_fn, float64_t alpha,
+    DistributionFamily family, LinkFunction link_fn, float64_t tau,
     float64_t lambda)
     : LinearMachine()
 {
-	m_alpha = alpha;
+	m_tau = tau;
 	m_lambda = lambda;
 	m_link_fn = link_fn;
 	m_descend_updater = descend_updater;
--- a/src/shogun/regression/GLM.cpp
+++ b/src/shogun/regression/GLM.cpp
@@ -1,12 +1,13 @@
 #include <shogun/lib/config.h>
 
+#include <shogun/features/DenseFeatures.h>
 #include <shogun/labels/RegressionLabels.h>
 #include <shogun/lib/SGVector.h>
 #include <shogun/mathematics/Math.h>
 #include <shogun/mathematics/linalg/LinalgNamespace.h>
 #include <shogun/mathematics/linalg/LinalgSpecialPurposes.h>
-
 #include <shogun/regression/GLM.h>
+
 #include <utility>
 using namespace shogun;
 
@@ -25,18 +26,21 @@ float64_t GLM::log_likelihood(
 	SGVector<float64_t> lambda(vector_count);
 	SGVector<float64_t> beta = LinearMachine::get_w();
 	float64_t beta0 = LinearMachine::get_bias();
+	auto feature_matrix = features->get_feature_matrix();
+	auto res =
+	    linalg::matrix_prod(SGMatrix(beta, 1, beta.vlen), feature_matrix);
+	linalg::add_scalar(res, beta0);
+	auto exponent = linalg::exponent(res);
 	for (auto i = 0; i < vector_count; i++)
 	{
-		SGVector<float64_t> feature_vector = features->get_feature_vector(i);
-		float64_t res = linalg::dot(feature_vector, beta);
-		lambda.set_element(log(1 + std::exp(beta0 + res)), i);
+		lambda[i] = log(1 + exponent[i]);
 	}
 	SGVector<float64_t> likelihood(vector_count);
 	SGVector<float64_t> labels = label->get_values();
 	SGVector<float64_t> log_lambda(vector_count);
 
 	for (auto i = 0; i < vector_count; i++)
-		log_lambda.set_element(log(lambda.get_element(i)), i);
+		log_lambda[i] = log(lambda[i]);
 
 	likelihood = linalg::add(
 	    linalg::element_prod(labels, log_lambda), lambda, 1.0, -1.0);
@@ -61,14 +65,16 @@ SGVector<float64_t> GLM::log_likelihood_derivative(
 	SGMatrix<float64_t> s(z.num_rows, z.num_cols);
 	linalg::logistic(z, s);
 	SGVector<float64_t> q(vector_count);
+	linalg::add_scalar(z, beta0);
+	auto exponent = linalg::exponent(z);
 	for (auto i = 0; i < vector_count; i++)
 	{
-		q.set_element(log(1 + std::exp(z.get_element(0, i))), i);
+		q[i] = log(1 + exponent[i]);
 	}
 	float64_t beta0_grad =
 	    linalg::sum(s) -
 	    linalg::sum(linalg::element_prod(label->get_values(), SGVector(s)));
-	result.set_element(beta0_grad, 0);
+	result[0] = beta0_grad;
 	return result;
 }
 void GLM::init()
--- a/src/shogun/regression/GLM.cpp
+++ b/src/shogun/regression/GLM.cpp
@@ -26,7 +26,7 @@ void GLM::init()
 	    ParameterProperties::SETTING);
 }
 GLM::GLM(
-    std::shared_ptr<DescendUpdater> descend_updater, Family family,
+    const std::shared_ptr<DescendUpdater>& descend_updater, Family family,
     LinkFunction Link_fn, float64_t alpha, float64_t lambda)
     : LinearMachine()
 {
--- a/src/shogun/regression/GLM.cpp
+++ b/src/shogun/regression/GLM.cpp
@@ -1,6 +1,7 @@
 #include <shogun/lib/config.h>
 
 #include <shogun/labels/RegressionLabels.h>
+#include <shogun/lib/SGMatrix.h>
 #include <shogun/mathematics/Math.h>
 #include <shogun/mathematics/linalg/LinalgNamespace.h>
 #include <shogun/regression/GLM.h>
@@ -11,28 +12,44 @@ GLM::GLM() : LinearMachine()
 {
 	init();
 }
-
+SGVector<float64_t>
+log_likelihood(SGVector<float64_t> features, float64_t label)
+{
+	ASSERT(features.size() > 0)
+	// Assume weights is the same as the feature vector
+	SGVector<float64_t> weights = SGMatrix(features.clone());
+	// Assume b0 to be the element in the beginning of the feature vector
+	float64_t b0 = features.get_element(0);
+}
 void GLM::init()
 {
-	SG_ADD(&m_alpha, "alpha", "alpha parameter", ParameterProperties::HYPER);
-	SG_ADD(&m_lambda, "lambda", "lambda parameter", ParameterProperties::HYPER);
+	SG_ADD(
+	    &m_alpha, "alpha", "Weighting parameter between L1 and L2 Penalty",
+	    ParameterProperties::HYPER);
+	SG_ADD(
+	    &m_lambda, "lambda", "Regularization parameter lambda",
+	    ParameterProperties::HYPER);
 	SG_ADD(
 	    (std::shared_ptr<SGObject>*)&m_descend_updater, "descend_updater",
 	    "Descend Updater used for updating weights",
 	    ParameterProperties::SETTING);
-	SG_ADD(&m_family, "family", "family used", ParameterProperties::SETTING);
 	SG_ADD(
-	    &m_linkfn, "linkfn", "Link function used",
+	    &m_family, "family", "Distribution Family used",
+	    ParameterProperties::SETTING);
+	SG_ADD(
+	    &m_link_fn, "link_fn", "Link function used",
 	    ParameterProperties::SETTING);
 }
 GLM::GLM(
-    const std::shared_ptr<DescendUpdater>& descend_updater, Family family,
-    LinkFunction Link_fn, float64_t alpha, float64_t lambda)
+    const std::shared_ptr<DescendUpdater>& descend_updater,
+    DistributionFamily family, LinkFunction link_fn, float64_t alpha,
+    float64_t lambda)
     : LinearMachine()
 {
 	m_alpha = alpha;
 	m_lambda = lambda;
-	m_linkfn = Link_fn;
+	m_link_fn = link_fn;
 	m_descend_updater = descend_updater;
 	m_family = family;
+	init();
 }
--- a/src/shogun/regression/GLM.cpp
+++ b/src/shogun/regression/GLM.cpp
@@ -24,8 +24,8 @@ float64_t GLM::log_likelihood(
 	ASSERT(vector_count > 0 && label->get_num_labels() == vector_count)
 
 	SGVector<float64_t> lambda(vector_count);
-	SGVector<float64_t> beta = LinearMachine::get_w();
-	float64_t beta0 = LinearMachine::get_bias();
+	auto beta = get_w();
+	auto beta0 = get_bias();
 	auto feature_matrix = features->get_feature_matrix();
 	auto res =
 	    linalg::matrix_prod(SGMatrix(beta, 1, beta.vlen), feature_matrix);
@@ -36,7 +36,7 @@ float64_t GLM::log_likelihood(
 		lambda[i] = log(1 + exponent[i]);
 	}
 	SGVector<float64_t> likelihood(vector_count);
-	SGVector<float64_t> labels = label->get_values();
+	auto labels = label->get_values();
 	SGVector<float64_t> log_lambda(vector_count);
 
 	for (auto i = 0; i < vector_count; i++)
@@ -55,9 +55,9 @@ SGVector<float64_t> GLM::log_likelihood_derivative(
 	auto feature_count = features->get_num_features();
 	ASSERT(vector_count > 0 && label->get_num_labels() == vector_count)
 	SGVector<float64_t> result(vector_count + 1);
-	SGVector<float64_t> beta = LinearMachine::get_w();
-	float64_t beta0 = LinearMachine::get_bias();
-	SGMatrix<float64_t> z = linalg::matrix_prod(
+	auto beta = get_w();
+	auto beta0 = get_bias();
+	auto z = linalg::matrix_prod(
 	    SGMatrix<float64_t>(beta, 1, beta.vlen),
 	    features->get_feature_matrix()); // Z is 1xN Matrix where N is the
 	                                     // number of vectors
@@ -96,11 +96,10 @@ void GLM::init()
 GLM::GLM(
     const std::shared_ptr<DescendUpdater>& descend_updater,
     DistributionFamily family, LinkFunction link_fn, float64_t tau)
-    : LinearMachine()
+    : GLM()
 {
 	m_tau = tau;
 	m_link_fn = link_fn;
 	m_descend_updater = descend_updater;
 	m_family = family;
-	init();
 }
--- a/src/shogun/regression/GLM.cpp
+++ b/src/shogun/regression/GLM.cpp
@@ -13,37 +13,35 @@ GLM::GLM() : LinearMachine()
 	init();
 }
 SGVector<float64_t> log_likelihood(
-    const std::shared_ptr<DenseFeatures<float64_t>> features,
-    std::shared_ptr<Labels> label)
+    const std::shared_ptr<DenseFeatures<float64_t>>& features,
+    const std::shared_ptr<Labels>& label)
 {
-	int32_t vector_count = features->get_num_vectors();
-	int32_t feature_count = features->get_num_features();
+	auto vector_count = features->get_num_vectors();
+	auto feature_count = features->get_num_features();
 	ASSERT(vector_count > 0 && label->get_num_labels() == vector_count)
 	// Array of Lambdas
 	SGVector<float64_t> lambda(vector_count);
-	for (int32_t i = 0; i < vector_count; i++)
+	for (auto i = 0; i < vector_count; i++)
 	{
 		SGVector<float64_t> feature_vector = features->get_feature_vector(i);
 		// Assume beta is the same as the feature vector
 		SGVector<float64_t> beta = feature_vector.clone();
 		// Assume beta0 is the same as the first element in the feature vector
 		float64_t b0 = feature_vector.get_element(0);
-		SGVector<float64_t> result(feature_count);
-		SGVector<float64_t>::vector_multiply(
-		    result, feature_vector, beta, feature_vector.vlen);
-		float64_t res = SGVector<float64_t>::sum(result);
+		float64_t res = linalg::dot(feature_vector, beta);
 		lambda.set_element(log(1 + std::exp(b0 + res)), i);
 	}
 	SGVector<float64_t> likelihood(vector_count);
+	SGVector<float64_t> labels = label->get_values();
+	SGVector<float64_t> log_lambda(vector_count);
 
-	for (int32_t i = 0; i < vector_count; i++)
-	{
-		float64_t y_i = label->get_value(i);
-		likelihood.set_element(
-		    y_i * log(lambda.get_element(i)) - lambda.get_element(i), i);
-	}
+	for (auto i = 0; i < vector_count; i++)
+		log_lambda.set_element(log(lambda.get_element(i)), i);
+
+	likelihood = linalg::add(
+	    linalg::element_prod(labels, log_lambda), lambda, 1.0, -1.0);
 	SGVector<float64_t> likelihood_clone = likelihood.clone();
-	for (int32_t i = 0; i < vector_count; i++)
+	for (auto i = 0; i < vector_count; i++)
 		likelihood.set_element(
 		    SGVector<float64_t>::sum(likelihood_clone, i), i);
 	return likelihood;
--- a/src/shogun/regression/GLM.cpp
+++ b/src/shogun/regression/GLM.cpp
@@ -4,6 +4,8 @@
 #include <shogun/lib/SGVector.h>
 #include <shogun/mathematics/Math.h>
 #include <shogun/mathematics/linalg/LinalgNamespace.h>
+#include <shogun/mathematics/linalg/LinalgSpecialPurposes.h>
+
 #include <shogun/regression/GLM.h>
 #include <utility>
 using namespace shogun;
@@ -19,7 +21,7 @@ SGVector<float64_t> GLM::log_likelihood(
 	auto vector_count = features->get_num_vectors();
 	auto feature_count = features->get_num_features();
 	ASSERT(vector_count > 0 && label->get_num_labels() == vector_count)
-	// Array of Lambdas
+
 	SGVector<float64_t> lambda(vector_count);
 	SGVector<float64_t> beta = LinearMachine::get_w();
 	float64_t beta0 = LinearMachine::get_bias();
@@ -42,6 +44,34 @@ SGVector<float64_t> GLM::log_likelihood(
 	for (auto i = 0; i < vector_count; i++)
 		likelihood.set_element(linalg::sum(likelihood_clone.slice(0, i)), i);
 	return likelihood;
+}
+
+SGVector<float64_t> GLM::log_likelihood_derivative(
+    const std::shared_ptr<DenseFeatures<float64_t>>& features,
+    const std::shared_ptr<Labels>& label)
+{
+	auto vector_count = features->get_num_vectors();
+	auto feature_count = features->get_num_features();
+	ASSERT(vector_count > 0 && label->get_num_labels() == vector_count)
+	SGVector<float64_t> result(vector_count + 1);
+	SGVector<float64_t> beta = LinearMachine::get_w();
+	float64_t beta0 = LinearMachine::get_bias();
+	SGMatrix<float64_t> z = linalg::add_scalar(
+	    linalg::matrix_prod(
+	        SGMatrix(beta, 1, beta.vlen), features->get_feature_matrix()),
+	    beta0); // Z is 1xN Matrix where N is the number of vectors
+	SGMatrix<float64_t> s(z.num_rows, z.num_cols);
+	linalg::logistic(z, s);
+	SGVector<float64_t> q(vector_count);
+	for (auto i = 0; i < vector_count; i++)
+	{
+		q.set_element(log(1 + std::exp(z.get_element(0, i))), i);
+	}
+	float64_t beta0_grad =
+	    linalg::sum(s) +
+	    linalg::sum(linalg::dot(label->get_values(), SGVector(s)));
+	result.set_element(beta0_grad, 0);
+	return result;
 }
 void GLM::init()
 {
--- a/src/shogun/regression/GLM.cpp
+++ b/src/shogun/regression/GLM.cpp
@@ -1,0 +1,24 @@
+#include <shogun/lib/config.h>
+
+#include <shogun/labels/RegressionLabels.h>
+#include <shogun/mathematics/Math.h>
+#include <shogun/mathematics/linalg/LinalgNamespace.h>
+#include <shogun/regression/GLM.h>
+#include <utility>
+using namespace shogun;
+
+GLM::GLM(): LinearMachine(){
+    init();
+}
+
+void GLM::init(){
+
+}
+GLM::GLM(DescendUpdaterWithCorrection* descend_updater, Family family, LinkFunction Link_fn, 
+	float64_t alpha, float64_t lambda): LinearMachine(){
+        m_alpha= alpha;
+        m_lambda= lambda;
+        m_linkfn= Link_fn;
+        m_descend_updater = descend_updater;
+        m_family= family;
+    }
--- a/src/shogun/regression/GLM.cpp
+++ b/src/shogun/regression/GLM.cpp
@@ -14,7 +14,7 @@ GLM::GLM() : LinearMachine()
 {
 	init();
 }
-SGVector<float64_t> GLM::log_likelihood(
+float64_t GLM::log_likelihood(
     const std::shared_ptr<DenseFeatures<float64_t>>& features,
     const std::shared_ptr<Labels>& label)
 {
@@ -40,10 +40,7 @@ SGVector<float64_t> GLM::log_likelihood(
 
 	likelihood = linalg::add(
 	    linalg::element_prod(labels, log_lambda), lambda, 1.0, -1.0);
-	SGVector<float64_t> likelihood_clone = likelihood.clone();
-	for (auto i = 0; i < vector_count; i++)
-		likelihood.set_element(linalg::sum(likelihood_clone.slice(0, i)), i);
-	return likelihood;
+	return linalg::sum(likelihood);
 }
 
 SGVector<float64_t> GLM::log_likelihood_derivative(
@@ -56,10 +53,11 @@ SGVector<float64_t> GLM::log_likelihood_derivative(
 	SGVector<float64_t> result(vector_count + 1);
 	SGVector<float64_t> beta = LinearMachine::get_w();
 	float64_t beta0 = LinearMachine::get_bias();
-	SGMatrix<float64_t> z = linalg::add_scalar(
-	    linalg::matrix_prod(
-	        SGMatrix(beta, 1, beta.vlen), features->get_feature_matrix()),
-	    beta0); // Z is 1xN Matrix where N is the number of vectors
+	SGMatrix<float64_t> z = linalg::matrix_prod(
+	    SGMatrix<float64_t>(beta, 1, beta.vlen),
+	    features->get_feature_matrix()); // Z is 1xN Matrix where N is the
+	                                     // number of vectors
+	linalg::add_scalar(z, beta0);
 	SGMatrix<float64_t> s(z.num_rows, z.num_cols);
 	linalg::logistic(z, s);
 	SGVector<float64_t> q(vector_count);
@@ -68,8 +66,8 @@ SGVector<float64_t> GLM::log_likelihood_derivative(
 		q.set_element(log(1 + std::exp(z.get_element(0, i))), i);
 	}
 	float64_t beta0_grad =
-	    linalg::sum(s) +
-	    linalg::sum(linalg::dot(label->get_values(), SGVector(s)));
+	    linalg::sum(s) -
+	    linalg::sum(linalg::element_prod(label->get_values(), SGVector(s)));
 	result.set_element(beta0_grad, 0);
 	return result;
 }
--- a/src/shogun/regression/GLM.cpp
+++ b/src/shogun/regression/GLM.cpp
@@ -1,7 +1,7 @@
 #include <shogun/lib/config.h>
 
 #include <shogun/labels/RegressionLabels.h>
-#include <shogun/lib/SGMatrix.h>
+#include <shogun/lib/SGVector.h>
 #include <shogun/mathematics/Math.h>
 #include <shogun/mathematics/linalg/LinalgNamespace.h>
 #include <shogun/regression/GLM.h>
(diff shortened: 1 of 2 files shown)


src/shogun/regression/GLM.cpp
```diff
@@ -0,0 +1,24 @@
...
+using namespace shogun;
+
+GLM::GLM(): LinearMachine(){
+    init();
+}
+
+void GLM::init(){
+
+}
+GLM::GLM(DescendUpdaterWithCorrection* descend_updater, Family family, LinkFunction Link_fn,
```
gf712: Could you use shared_ptr instead of raw pointer

src/shogun/regression/GLM.h
```diff
@@ -0,0 +1,104 @@
...
+#include <shogun/machine/FeatureDispatchCRTP.h>
+#include <shogun/machine/LinearMachine.h>
+#include <shogun/optimization/DescendUpdaterWithCorrection.h>
+#include <shogun/regression/Regression.h>
+
+namespace shogun
+{
+enum Family
+{
+	normal,
```
gf712: We usually write the enum values in caps

src/shogun/regression/GLM.h
```diff
@@ -0,0 +1,104 @@
...
+		/** default destructor */
+		virtual ~GLM() {}
+
+		/** train model
+		 *
+		 * @param data training data 
+		 *
+		 * @return whether training was successful
+		 */
+		virtual bool train_machine(Features* data=NULL) {return true;};
```
gf712: Same here, shouldn’t this be shared_ptr? I think this is actually not overriding the base class train_machine but actually creating a new function that is virtual

src/shogun/regression/GLM.cpp
```diff
@@ -0,0 +1,24 @@
...
+
+#include <shogun/labels/RegressionLabels.h>
+#include <shogun/mathematics/Math.h>
+#include <shogun/mathematics/linalg/LinalgNamespace.h>
+#include <shogun/regression/GLM.h>
+#include <utility>
+using namespace shogun;
+
+GLM::GLM(): LinearMachine(){
+    init();
```
gf712: The class members would be registered here with watch_method/SG_ADD

src/shogun/regression/GLM.h
```diff
@@ -0,0 +1,104 @@
...
+{
+	public:
+		/** problem type */
+		MACHINE_PROBLEM_TYPE(PT_REGRESSION);
+
+
+		/** Default constructor */
+		GLM();
+
+		/** default constructor
```
gf712: This isn’t a default constructor though

gf712 (commented):
I think it’s a good start, but need to see how it will be implemented in train_machine!

src/shogun/regression/GLM.cpp
gf712: Btw this should be the base class, ie DescendUpdater

src/shogun/regression/GLM.h
vigsterkr: afaik this is protected in Machine. i mean the version with shared_ptr<Features>

src/shogun/regression/GLM.h
```diff
@@ -0,0 +1,104 @@
...
+		 *
+		 * @param data training data 
+		 *
+		 * @return whether training was successful
+		 */
+		virtual bool train_machine(Features* data=NULL) {return true;};
+
+		/** @return object name */
+		virtual const char* get_name() const { return "GLM"; }
+
```
vigsterkr, line 69: no need for this many new lines :)

src/shogun/regression/GLM.h
```diff
@@ -0,0 +1,104 @@
...
+
+
+	protected:
+
+		DescendUpdaterWithCorrection* m_descend_updater;
```
vigsterkr: shared_ptr plz

src/shogun/regression/GLM.cpp
vigsterkr: second that

src/shogun/regression/GLM.cpp
```diff
@@ -0,0 +1,24 @@
...
+}
+
+void GLM::init(){
+
+}
+GLM::GLM(DescendUpdaterWithCorrection* descend_updater, Family family, LinkFunction Link_fn, 
+	float64_t alpha, float64_t lambda): LinearMachine(){
+        m_alpha= alpha;
```
vigsterkr: nitpicking: plz be consistent with equation either `<variable>= value` or `<variable> = value`

src/shogun/regression/GLM.h
```diff
@@ -0,0 +1,104 @@
...
+	inverse
+};
+/** @brief Class GLM implements Generalized Linear Models, such as poisson, gamma, binomial
+ */
+class GLM : public LinearMachine
+{
+	public:
+		/** problem type */
+		MACHINE_PROBLEM_TYPE(PT_REGRESSION);
+
```
vigsterkr, line 45: clang-format?

vigsterkr (changes requested):

src/shogun/regression/GLM.cpp
gf712: Sorry, not here, below in ::init()

src/shogun/regression/GLM.h
gf712: Yea that is the issue that #5001 would flag at compile time. Should use override and then the compiler complains if there is no virtual function with the same signature in the base class.

src/shogun/regression/GLM.cpp
```diff
@@ -7,18 +7,32 @@
...
-	float64_t alpha, float64_t lambda): LinearMachine(){
-        m_alpha= alpha;
-        m_lambda= lambda;
-        m_linkfn= Link_fn;
-        m_descend_updater = descend_updater;
-        m_family= family;
-    }
\ No newline at end of file
+GLM::GLM(
+    std::shared_ptr<DescendUpdater> descend_updater, Family family,
```
vigsterkr: `const std::shared_ptr<DescendUpdater>& descend_updater`

src/shogun/regression/GLM.h
```diff
@@ -17,88 +17,96 @@
...
+		 *
+		 * @param descend_updater chosen Descend Updater algorithm
+		 * @param Linkfn the link function check
+		 * https://www.rdocumentation.org/packages/stats/versions/3.6.2/topics/family
+		 * @param Family the family check
+		 * https://www.rdocumentation.org/packages/stats/versions/3.6.2/topics/family
+		 * @param alpha alpha
+		 * @param lambda lambda
+		 */
+		GLM(std::shared_ptr<DescendUpdater>,
```
vigsterkr: `const std::shared_ptr<DescendUpdater>& descend_updater`

vigsterkr (changes requested):

src/shogun/regression/GLM.cpp
Khalifa1997: funny it built the first time, changed it tho

src/shogun/regression/GLM.h
```diff
@@ -0,0 +1,112 @@
...
+		GLM();
+
+		/** Constructor
+		 *
+		 * @param descend_updater chosen Descend Updater algorithm
+		 * @param Linkfn the link function check
+		 * https://www.rdocumentation.org/packages/stats/versions/3.6.2/topics/family
+		 * @param Family the family check
+		 * https://www.rdocumentation.org/packages/stats/versions/3.6.2/topics/family
+		 * @param alpha alpha
```
karlnapf: what are alpha and lambda? could you also describe them in the docs if you put docs? (otherwise they are pointless)

src/shogun/regression/GLM.h
```diff
@@ -0,0 +1,112 @@
...
+		 * @param lambda lambda
+		 */
+		GLM(const std::shared_ptr<DescendUpdater>& descend_updater,
```
karlnapf, line 55: could you remove those ctors? The new API design uses the default ctor with subsequent `put` calls

src/shogun/regression/GLM.h
```diff
@@ -0,0 +1,112 @@
...
+		 * @param learn_rate Learning rate
+		 * @param Linkfn the link function check
+		 * https://www.rdocumentation.org/packages/stats/versions/3.6.2/topics/family
+		 * @param Family the distribution/family check
+		 * https://www.rdocumentation.org/packages/stats/versions/3.6.2/topics/family
+		 * @param alpha alpha
+		 * @param lambda lambda
+		 */
+
+		/** default destructor */
```
karlnapf: no need for such comments

src/shogun/regression/GLM.h
```diff
@@ -0,0 +1,112 @@
...
+		GLM(const std::shared_ptr<DescendUpdater>& descend_updater,
+		    Family family = POISSON_DISTRIBUTION, LinkFunction Link_fn = LOG,
+		    float64_t alpha = 0.5, float64_t lambda = 0.1);
+
+		/** standard constructor
+		 * @param data features
+		 * @param labs labels
+		 * @param learn_rate Learning rate
+		 * @param Linkfn the link function check
+		 * https://www.rdocumentation.org/packages/stats/versions/3.6.2/topics/family
```
karlnapf: such references would go into the class docs, not the ctor docs

src/shogun/regression/GLM.h
```diff
@@ -0,0 +1,112 @@
...
+		 * @param descend_updater chosen Descend Updater algorithm
+		 * @param Linkfn the link function check
+		 * https://www.rdocumentation.org/packages/stats/versions/3.6.2/topics/family
+		 * @param Family the family check
+		 * https://www.rdocumentation.org/packages/stats/versions/3.6.2/topics/family
+		 * @param alpha alpha
+		 * @param lambda lambda
+		 */
+		GLM(const std::shared_ptr<DescendUpdater>& descend_updater,
+		    Family family = POISSON_DISTRIBUTION, LinkFunction Link_fn = LOG,
```
karlnapf: Link_fn -> link_fn

src/shogun/regression/GLM.h
```diff
@@ -0,0 +1,112 @@
...
+		{
+			return "GLM";
+		}
+
+	protected:
+		std::shared_ptr<DescendUpdater> m_descend_updater;
+		float64_t m_alpha;
+		float64_t m_lambda;
+		Family m_family;
+		LinkFunction m_linkfn;
```
karlnapf: pls use consistent variable names, i.e. "m_link_fn"

src/shogun/regression/GLM.h
```diff
@@ -0,0 +1,112 @@
...
+#include <shogun/regression/Regression.h>
+
+namespace shogun
+{
+	enum Family
+	{
+		NORMAL_DISTRIBUTION,
+		EXPONENTIAL_DISTRIBUTION,
+		GAMMA_DISTRIBUTION,
+		BINOMIAL_DISTRIBUTION,
```
karlnapf, line 26: As I said before, I am not very sure of representing these things as enums, this will lead to spaghetti code with that many. BUT you can leave it for now and just focus on the poisson regression. Once that is done we can think about this again

src/shogun/regression/GLM.h
```diff
@@ -0,0 +1,112 @@
...
+
+#include <shogun/features/Features.h>
+#include <shogun/machine/FeatureDispatchCRTP.h>
+#include <shogun/machine/LinearMachine.h>
+#include <shogun/optimization/DescendUpdater.h>
+#include <shogun/regression/Regression.h>
+
+namespace shogun
+{
+	enum Family
```
karlnapf: if you put this enum into the shogun namespace, it should have a better name

src/shogun/regression/GLM.h
vigsterkr: in SWIG... i would still have this for c++

src/shogun/regression/GLM.cpp
```diff
@@ -0,0 +1,38 @@
...
+using namespace shogun;
+
+GLM::GLM() : LinearMachine()
+{
+	init();
+}
+
+void GLM::init()
+{
+	SG_ADD(&m_alpha, "alpha", "alpha parameter", ParameterProperties::HYPER);
```
karlnapf: please avoid empty comments/docs/etc as "alpha parameter" for a varialbe called "alpha". If you put a description, why not say what it is or what it does?

src/shogun/regression/GLM.h
karlnapf: fair enough. in that case, please make sure the class works with the default ctor. We had/have many bugs where a class segfaults in the train method as the default ctor didnt initialize things and the train method didnt check that. Don't rely that a user used this ctor, write separate tests for both

src/shogun/regression/GLM.h
```diff
@@ -0,0 +1,112 @@
+/*
+ * This software is distributed under BSD 3-clause license (see LICENSE file).
+ *
+ * Authors: Ahmed Khalifa
+ */
+/*
+References:
+https://www.rdocumentation.org/packages/stats/versions/3.6.2/topics/family
```
karlnapf: I'd put this into the class docs.
But actually, don't just blindly put references, we dont really need reference to other libs in here. Rather a reference to wikipedia describing GLM (or Poisson regression) in general, in a conceptual/mathy sense

karlnapf (commented):
This is good progress from the first version! But they key thing now will be to write a train method. I suggest you add a log_likelihood method first, and then one that computes the gradient of it. And then finally you can call the descent updater in the train method. See wikipedia for the expressions. As said, I would start to do this for Poisson regression first, and then once it works we can modularise it

src/shogun/regression/GLM.h
```diff
@@ -0,0 +1,112 @@
...
+https://www.statsmodels.org/stable/generated/statsmodels.genmod.generalized_linear_model.GLM.html#statsmodels.genmod.generalized_linear_model.GLM
+http://glm-tools.github.io/pyglmnet/api.html
+*/
+#ifndef _GENERALIZEDLINEARMODEL_H__
+#define _GENERALIZEDLINEARMODEL_H__
+
+#include <shogun/lib/config.h>
+
+#include <shogun/features/Features.h>
+#include <shogun/machine/FeatureDispatchCRTP.h>
```
karlnapf, line 13: good point, would be good if this class worked for multiple feature types. But you are not using this atm I think?

src/shogun/regression/GLM.h
karlnapf: and also please make sure that the default parameters in the ctor match the default parameters when using the default one (another source of confusion). I would remove the default values in the ctor to be on the safe side actually, and instead you can initialize the values in the `::init` method

src/shogun/regression/GLM.cpp
```diff
@@ -0,0 +1,38 @@
...
+GLM::GLM(
+    const std::shared_ptr<DescendUpdater>& descend_updater, Family family,
+    LinkFunction Link_fn, float64_t alpha, float64_t lambda)
+    : LinearMachine()
+{
+	m_alpha = alpha;
+	m_lambda = lambda;
+	m_linkfn = Link_fn;
+	m_descend_updater = descend_updater;
+	m_family = family;
```
karlnapf, line 104: you need to call the init method here  as well

vigsterkr: i would rather call the default GLM above... that way no need to call init.

src/shogun/regression/GLM.cpp
```diff
@@ -0,0 +1,38 @@
...
+	    ParameterProperties::SETTING);
+	SG_ADD(&m_family, "family", "family used", ParameterProperties::SETTING);
+	SG_ADD(
+	    &m_linkfn, "linkfn", "Link function used",
+	    ParameterProperties::SETTING);
+}
+GLM::GLM(
+    const std::shared_ptr<DescendUpdater>& descend_updater, Family family,
+    LinkFunction Link_fn, float64_t alpha, float64_t lambda)
+    : LinearMachine()
```
vigsterkr:
```suggestion
    : GLM()
```

vigsterkr: this way init will be called before setting the new values

src/shogun/regression/GLM.h
gf712: You can even have the values inline with the class member declaration

src/shogun/regression/GLM.cpp
```diff
@@ -0,0 +1,38 @@
...
+	SG_ADD(&m_alpha, "alpha", "alpha parameter", ParameterProperties::HYPER);
+	SG_ADD(&m_lambda, "lambda", "lambda parameter", ParameterProperties::HYPER);
+	SG_ADD(
+	    (std::shared_ptr<SGObject>*)&m_descend_updater, "descend_updater",
+	    "Descend Updater used for updating weights",
+	    ParameterProperties::SETTING);
+	SG_ADD(&m_family, "family", "family used", ParameterProperties::SETTING);
+	SG_ADD(
+	    &m_linkfn, "linkfn", "Link function used",
+	    ParameterProperties::SETTING);
```
karlnapf, line 94: you need to initialize all values to defaults here. This brings up an interesting question. What is the default descend updater? I think it would be good to have one set so that users are not forced to pass one (tedious).

src/shogun/regression/GLM.cpp
```diff
@@ -0,0 +1,38 @@
...
+}
+GLM::GLM(
+    const std::shared_ptr<DescendUpdater>& descend_updater, Family family,
```
karlnapf: Another comment as I saw a concversation in irc re swig and the fact that the descend updater is not exposed yet. I would not worry about swig and the interfaces for now. Let's focus on the c++ code for now, possibly with unit tests.
The next step here (after incorporating the feedback you got) is to implement the gradient of the log-likelihood, and the next thing then is passing that to a descent updater. For the latter, you can start with plain old gradient descent, no need to be more fancy (for now).

src/shogun/regression/GLM.cpp
Khalifa1997: I think since they are just linear models at the end of the day, there isn't really alot of parameters to learn like in Neural Networks, SGD would work best here which I believe is `GradientDescendUpdater`

karlnapf: The loss is convex (for Poisson regression with log link function), so a second order method like Newton will be better. But we can change that later, best is to start with something simple, then work it up from there.

src/shogun/regression/GLM.h
Khalifa1997: I haven't explored all the feature types aren't all features the same? (aka. just numbers at the end of the day) please do enlighten me :D

Khalifa1997:
I have updated the class I will try to list all the changes I have made
1. Changed the ctor structure such that the default values are inline in the class variable declarations so the default ctor problem no longer exists
2. Removed pointless comments and changed variable naming + added some description to values like alpha/lambda
3. I tried implementing log_likelihood and I know how, based on [this](http://glm-tools.github.io/pyglmnet/tutorial.html#poisson-log-likelihood) the only problem is I was using `SGVector` and `SGMatrix` but since I am going to have to use things like the dot operator maybe using `Features` class would be the better idea here - and label class for the label- but Idk which type of `Features` should I be using, `DotFeatures`?
4. Have yet to decide a default `DescendUpdater`

src/shogun/regression/GLM.h
karlnapf: read the code :)

src/shogun/regression/GLM.cpp
```diff
@@ -0,0 +1,55 @@
...
+using namespace shogun;
+
+GLM::GLM() : LinearMachine()
+{
+	init();
+}
+SGVector<float64_t>
+log_likelihood(SGVector<float64_t> features, float64_t label)
+{
+	ASSERT(features.size() > 0)
```
karlnapf: You would write code that implements the log likelihood for all data at once

src/shogun/regression/GLM.cpp
```diff
@@ -0,0 +1,55 @@
...
+	// Assume weights is the same as the feature vector
+	SGVector<float64_t> weights = SGMatrix(features.clone());
+	// Assume b0 to be the element in the beginning of the feature vector
+	float64_t b0 = features.get_element(0);
```
karlnapf: this code doesnt do anything yet

karlnapf:
I just read through the code, and many of the things we commented on are not addressed yet.

RE your question about the features, read some other classes, e.g. ridge regression, that will make it clear

Khalifa1997:
I have updated the log_likelihood function such that it takes dense features, and computes the likelihood, as you could tell tho the naming of variables is abit off, so I am open to suggestions in that area. As for the gradient calculation I am using [this](http://glm-tools.github.io/pyglmnet/tutorial.html#gradient-descent) implementation, however I am not sure if shogun has a sigmoid function or would I have to make a small function for it?

src/shogun/regression/GLM.cpp
```diff
@@ -7,18 +7,32 @@
...
-void GLM::init(){
-
+void GLM::init()
+{
+	SG_ADD(&m_alpha, "alpha", "alpha parameter", ParameterProperties::HYPER);
+	SG_ADD(&m_lambda, "lambda", "lambda parameter", ParameterProperties::HYPER);
+	SG_ADD(
+	    (std::shared_ptr<SGObject>*)&m_descend_updater, "descend_updater",
+	    "Descend Updater used for updating weights",
+	    ParameterProperties::SETTING);
```
Khalifa1997, line 88: as for those lines, I think that's what causing the build to fail

Khalifa1997: any thoughts? @karlnapf

src/shogun/regression/GLM.cpp
```diff
@@ -0,0 +1,82 @@
...
+#include <utility>
+using namespace shogun;
+
+GLM::GLM() : LinearMachine()
+{
+	init();
+}
+SGVector<float64_t> log_likelihood(
+    const std::shared_ptr<DenseFeatures<float64_t>> features,
+    std::shared_ptr<Labels> label)
```
vigsterkr:
```suggestion
    const std::shared_ptr<DenseFeatures<float64_t>>& features,
    const std::shared_ptr<Labels>& label)
```

src/shogun/regression/GLM.cpp
```diff
@@ -0,0 +1,82 @@
...
+{
+	int32_t vector_count = features->get_num_vectors();
+	int32_t feature_count = features->get_num_features();
```
vigsterkr:
```suggestion
	auto vector_count = features->get_num_vectors();
	auto feature_count = features->get_num_features();
```

vigsterkr: in general try to use for these kind of statements auto

src/shogun/regression/GLM.cpp
```diff
@@ -0,0 +1,82 @@
...
+	ASSERT(vector_count > 0 && label->get_num_labels() == vector_count)
+	// Array of Lambdas
+	SGVector<float64_t> lambda(vector_count);
+	for (int32_t i = 0; i < vector_count; i++)
+	{
+		SGVector<float64_t> feature_vector = features->get_feature_vector(i);
```
vigsterkr: auto

src/shogun/regression/GLM.cpp
```diff
@@ -0,0 +1,82 @@
...
+		float64_t b0 = feature_vector.get_element(0);
+		SGVector<float64_t> result(feature_count);
+		SGVector<float64_t>::vector_multiply(
+		    result, feature_vector, beta, feature_vector.vlen);
+		float64_t res = SGVector<float64_t>::sum(result);
+		lambda.set_element(log(1 + std::exp(b0 + res)), i);
+	}
+	SGVector<float64_t> likelihood(vector_count);
+
+	for (int32_t i = 0; i < vector_count; i++)
```
vigsterkr: this whole thing does the following:
likelihood = y*log(lambda) - lambda
this is definitely vectorizable... please use linalg framework to do this. instead of doing this vanilla code

src/shogun/regression/GLM.cpp
```diff
@@ -0,0 +1,82 @@
...
+	{
+		float64_t y_i = label->get_value(i);
+		likelihood.set_element(
+		    y_i * log(lambda.get_element(i)) - lambda.get_element(i), i);
+	}
+	SGVector<float64_t> likelihood_clone = likelihood.clone();
```
vigsterkr: auto

src/shogun/regression/GLM.cpp
```diff
@@ -0,0 +1,82 @@
...
+	SGVector<float64_t> lambda(vector_count);
+	for (int32_t i = 0; i < vector_count; i++)
+	{
+		SGVector<float64_t> feature_vector = features->get_feature_vector(i);
+		// Assume beta is the same as the feature vector
+		SGVector<float64_t> beta = feature_vector.clone();
+		// Assume beta0 is the same as the first element in the feature vector
+		float64_t b0 = feature_vector.get_element(0);
+		SGVector<float64_t> result(feature_count);
+		SGVector<float64_t>::vector_multiply(
```
vigsterkr: linalg

src/shogun/regression/GLM.cpp
```diff
@@ -0,0 +1,82 @@
...
+		    result, feature_vector, beta, feature_vector.vlen);
+		float64_t res = SGVector<float64_t>::sum(result);
```
vigsterkr: linalg

vigsterkr (changes requested):
initial problems...

src/shogun/regression/GLM.cpp
```diff
@@ -0,0 +1,80 @@
...
+	auto feature_count = features->get_num_features();
+	ASSERT(vector_count > 0 && label->get_num_labels() == vector_count)
+	// Array of Lambdas
+	SGVector<float64_t> lambda(vector_count);
+	for (auto i = 0; i < vector_count; i++)
+	{
+		SGVector<float64_t> feature_vector = features->get_feature_vector(i);
+		// Assume beta is the same as the feature vector
+		SGVector<float64_t> beta = feature_vector.clone();
+		// Assume beta0 is the same as the first element in the feature vector
```
karlnapf: we don't do that, this is a trick to make the math look easier, but we will treat bias and weights separately please

src/shogun/regression/GLM.cpp
```diff
@@ -0,0 +1,80 @@
...
+SGVector<float64_t> log_likelihood(
+    const std::shared_ptr<DenseFeatures<float64_t>>& features,
+    const std::shared_ptr<Labels>& label)
+{
+	auto vector_count = features->get_num_vectors();
+	auto feature_count = features->get_num_features();
+	ASSERT(vector_count > 0 && label->get_num_labels() == vector_count)
+	// Array of Lambdas
+	SGVector<float64_t> lambda(vector_count);
+	for (auto i = 0; i < vector_count; i++)
```
karlnapf, line 34: you don't need to loop here if using dense features. You can do a matrix vector product

src/shogun/regression/GLM.cpp
```diff
@@ -0,0 +1,80 @@
...
+		float64_t b0 = feature_vector.get_element(0);
+		float64_t res = linalg::dot(feature_vector, beta);
+		lambda.set_element(log(1 + std::exp(b0 + res)), i);
+	}
+	SGVector<float64_t> likelihood(vector_count);
+	SGVector<float64_t> labels = label->get_values();
+	SGVector<float64_t> log_lambda(vector_count);
+
+	for (auto i = 0; i < vector_count; i++)
+		log_lambda.set_element(log(lambda.get_element(i)), i);
```
karlnapf: don't we have element wise log in linalg? if not we should have it

src/shogun/regression/GLM.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+	likelihood = linalg::add(
+	    linalg::element_prod(labels, log_lambda), lambda, 1.0, -1.0);
+	SGVector<float64_t> likelihood_clone = likelihood.clone();
+	for (auto i = 0; i < vector_count; i++)
+		likelihood.set_element(
```
karlnapf: same here, please do with this matrix algebra and elementwise operations

src/shogun/regression/GLM.cpp
```diff
@@ -0,0 +1,80 @@
...
+	init();
+}
+SGVector<float64_t> log_likelihood(
+    const std::shared_ptr<DenseFeatures<float64_t>>& features,
+    const std::shared_ptr<Labels>& label)
+{
+	auto vector_count = features->get_num_vectors();
+	auto feature_count = features->get_num_features();
+	ASSERT(vector_count > 0 && label->get_num_labels() == vector_count)
+	// Array of Lambdas
```
karlnapf: remove such comments

src/shogun/regression/GLM.cpp
```diff
@@ -0,0 +1,80 @@
...
+	    linalg::element_prod(labels, log_lambda), lambda, 1.0, -1.0);
+	SGVector<float64_t> likelihood_clone = likelihood.clone();
+	for (auto i = 0; i < vector_count; i++)
+		likelihood.set_element(
+		    SGVector<float64_t>::sum(likelihood_clone, i), i);
+	return likelihood;
+}
+void GLM::init()
+{
+	SG_ADD(
```
karlnapf, line 82: this is for the elastic net stuff, which we don't need here for now. Just the L2 reguariser is enough

src/shogun/regression/GLM.cpp
```diff
@@ -0,0 +1,80 @@
...
+	    &m_alpha, "alpha", "Weighting parameter between L1 and L2 Penalty",
+	    ParameterProperties::HYPER);
+	SG_ADD(
+	    &m_lambda, "lambda", "Regularization parameter lambda",
```
karlnapf: Add "L2".

karlnapf (commented):
well done on adding the log likelihood. Though you should vectorize it.

Next step is to implement the gradient of that

src/shogun/regression/GLM.cpp
Khalifa1997: I don't think there is. I had checked in the documentation and didn't find anything for it

src/shogun/regression/GLM.cpp
Khalifa1997: you mean use `bias` and `m_w` for weights?

src/shogun/regression/GLM.cpp
Khalifa1997: so just removing alpha and lambda and adding tau as in `LinearRidgeRegression` ?

src/shogun/regression/GLM.cpp
Khalifa1997: there is no `log` operation in `linalg` thus I don't think it can be done without a loop

src/shogun/regression/GLM.h
geektoni: We could make the `GLM` class accept a `GLMDistribution` object instead. It will contain the log-likelihood and gradients of the given distribution (e.g., `GLMDistributionPoisson`). Then, the `GLM` class will only call the methods of the `GLMDistribution` to train itself.  

However, this is obviously out of the scope of this PR.

src/shogun/regression/GLM.cpp
lgoetz: yes, for example; the point is, you should not assume the bias is the first element of the feature vector

src/shogun/regression/GLM.cpp
lgoetz: see src/shogun/distributions/LinearHMM.cpp for example

src/shogun/regression/GLM.cpp
lgoetz: Yes (or use lambda instead of tau). And please add "L2" to the description to clarify it's a ridge penalty

src/shogun/regression/GLM.cpp
karlnapf: If it is not in linalg, add it there. We have something for elementwise ops in there iirc.
I don't think LinearHMM is a good place to look .. It is way too old school ;)

src/shogun/regression/GLM.cpp
Khalifa1997: @lgoetz yes I have changed that in my last commit thanks :)

src/shogun/regression/GLM.cpp
Khalifa1997: Done, I think I will use tau for this method to keep it consistent with the other models such as LinearRidgeRegression + I am using the naming lambda for the likelihood array

src/shogun/regression/GLM.cpp
Khalifa1997: @karlnapf do I start another PR and add it?

Khalifa1997:
Now I am going to briefly sum up what I have added here:
the expected result of the log_likelihood_derivative should be of count `len(beta) + 1 //for the beta0 element`  so I have made a result array for those things.
Now I am facing two problems atm
1. there is no log function in linalg
2. there is no elementwise division in linalg

now on line 70 in the .cpp file  it should of been `y*s / q` but since there is no linalg I would have to do it without vectorization which would make the code much much more ugly.

So I would love to have some feedback on what to do now thanks!

N.B: implementing grad_beta should be very straight forward once the issues I have mentionned are solved

src/shogun/regression/GLM.cpp
lgoetz: > If it is not in linalg, add it there. We have something for elementwise ops in there iirc.
> I don't think LinearHMM is a good place to look .. It is way too old school ;)

sorry ;)

src/shogun/regression/GLM.cpp
```diff
@@ -0,0 +1,100 @@
...
+	auto vector_count = features->get_num_vectors();
+	auto feature_count = features->get_num_features();
+	ASSERT(vector_count > 0 && label->get_num_labels() == vector_count)
+
+	SGVector<float64_t> lambda(vector_count);
+	SGVector<float64_t> beta = LinearMachine::get_w();
+	float64_t beta0 = LinearMachine::get_bias();
+	for (auto i = 0; i < vector_count; i++)
+	{
+		SGVector<float64_t> feature_vector = features->get_feature_vector(i);
```
karlnapf: auto

src/shogun/regression/GLM.cpp
karlnapf: @Khalifa1997 why do you resolve conversations when you haven't addressed the comment by @vigsterkr ?
This just results in us giving redundant feedback -> wasting time

src/shogun/regression/GLM.cpp
```diff
@@ -0,0 +1,100 @@
...
+		float64_t res = linalg::dot(feature_vector, beta);
```
karlnapf: dont do a loop here but a matrix vector product

src/shogun/regression/GLM.cpp
```diff
@@ -0,0 +1,100 @@
...
+	likelihood = linalg::add(
+	    linalg::element_prod(labels, log_lambda), lambda, 1.0, -1.0);
+	return linalg::sum(likelihood);
+}
+
+SGVector<float64_t> GLM::log_likelihood_derivative(
+    const std::shared_ptr<DenseFeatures<float64_t>>& features,
+    const std::shared_ptr<Labels>& label)
+{
+	auto vector_count = features->get_num_vectors();
```
karlnapf, line 54: it would be good if you could modularise this.
The log likelihood consists of multiple parts, activation, data terms, etc. You could put each of those into a helper method to make it more readable

src/shogun/regression/GLM.cpp
```diff
@@ -0,0 +1,100 @@
...
+	    SGMatrix<float64_t>(beta, 1, beta.vlen),
+	    features->get_feature_matrix()); // Z is 1xN Matrix where N is the
+	                                     // number of vectors
+	linalg::add_scalar(z, beta0);
+	SGMatrix<float64_t> s(z.num_rows, z.num_cols);
+	linalg::logistic(z, s);
+	SGVector<float64_t> q(vector_count);
+	for (auto i = 0; i < vector_count; i++)
+	{
+		q.set_element(log(1 + std::exp(z.get_element(0, i))), i);
```
karlnapf: q[i] z[i]

src/shogun/regression/GLM.cpp
karlnapf: Not sure what you mean

src/shogun/regression/GLM.h
karlnapf: that, or have locally defined helper classes that are then instantiated based on an enum.
But the code for single cases (eg posison) should be in a single place (likelihood contribution, gradients, etc)

karlnapf (commented):
Cool, great that you have the gradient in place now

You should slowly start thinking about a unit test where you compare likelihood / gradient values with a reference implementation on some fixed simple data

Also made some minor comments

src/shogun/regression/GLM.cpp
Khalifa1997: he had mentionned auto to be used in the for loop and I think I had address them all in my code +he mentionned the ctor thing when it took default values. I fixed this by giving default values inline with member attributes. but if I had missed any comment, then yea I appologise and ofc u can point it out :)

karlnapf: You can always use auto if possible. The only case where I would write the type is if it would be unclear/confusing otherwise :)

src/shogun/regression/GLM.cpp
```diff
@@ -0,0 +1,100 @@
...
+    const std::shared_ptr<DenseFeatures<float64_t>>& features,
+    const std::shared_ptr<Labels>& label)
+{
+	auto vector_count = features->get_num_vectors();
+	auto feature_count = features->get_num_features();
+	ASSERT(vector_count > 0 && label->get_num_labels() == vector_count)
+
+	SGVector<float64_t> lambda(vector_count);
+	SGVector<float64_t> beta = LinearMachine::get_w();
+	float64_t beta0 = LinearMachine::get_bias();
```
gf712: Do we override get_w/get_bias? As in is there a particular reason why you specify LinearMachine::?

src/shogun/regression/GLM.cpp
gf712: And use linalg?

src/shogun/regression/GLM.cpp
Khalifa1997: I get an error when I test as you could tell in the CI, in the SGObject test, so I have a feeling this is what causing it. what do you think?

src/shogun/regression/GLM.cpp
Khalifa1997: @gf712  there is no linalg for log :S,
and by your comment @karlnapf do you mean this is how I should get the elements?

src/shogun/regression/GLM.cpp
Khalifa1997: For some reason, I get get_w not declared in this scope without me doing so

Edit: now it works after I fixed some of my includes.. ig :D

src/shogun/regression/GLM.cpp
Khalifa1997: Okay will do that in my next commit thanks:)

src/shogun/regression/GLM.cpp
Khalifa1997: I have changed this part to return `linalg::sum(likelihood)` so should be done

src/shogun/regression/GLM.cpp
Khalifa1997: Done.

src/shogun/regression/GLM.cpp
Khalifa1997: Done.

src/shogun/regression/GLM.cpp
vigsterkr: i'm not so sure how includes changed things...

src/shogun/regression/GLM.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+SGVector<float64_t> GLM::log_likelihood_derivative(
+    const std::shared_ptr<DenseFeatures<float64_t>>& features,
+    const std::shared_ptr<Labels>& label)
+{
+	auto vector_count = features->get_num_vectors();
+	auto feature_count = features->get_num_features();
+	ASSERT(vector_count > 0 && label->get_num_labels() == vector_count)
+	SGVector<float64_t> result(vector_count + 1);
+	SGVector<float64_t> beta = LinearMachine::get_w();
```
vigsterkr: auto....

vigsterkr: get_w() should just work.

src/shogun/regression/GLM.cpp
```diff
@@ -0,0 +1,106 @@
...
+	float64_t beta0 = LinearMachine::get_bias();
```
vigsterkr: auto

src/shogun/regression/GLM.cpp
```diff
@@ -0,0 +1,106 @@
...
+	SGMatrix<float64_t> z = linalg::matrix_prod(
```
vigsterkr: auto

src/shogun/regression/GLM.cpp
```diff
@@ -0,0 +1,106 @@
...
+	SGMatrix<float64_t> s(z.num_rows, z.num_cols);
+	linalg::logistic(z, s);
+	SGVector<float64_t> q(vector_count);
+	linalg::add_scalar(z, beta0);
+	auto exponent = linalg::exponent(z);
+	for (auto i = 0; i < vector_count; i++)
+	{
+		q[i] = log(1 + exponent[i]);
+	}
+	float64_t beta0_grad =
```
vigsterkr, line 74: auto

src/shogun/regression/GLM.cpp
vigsterkr: when you report such thing, could you please make sure that you copy here the link to the CI line where you think the error is....

src/shogun/regression/GLM.cpp
```diff
@@ -0,0 +1,106 @@
...
+	    &m_family, "family", "Distribution Family used",
+	    ParameterProperties::SETTING);
+	SG_ADD(
+	    &m_link_fn, "link_fn", "Link function used",
+	    ParameterProperties::SETTING);
+}
+GLM::GLM(
+    const std::shared_ptr<DescendUpdater>& descend_updater,
+    DistributionFamily family, LinkFunction link_fn, float64_t tau)
+    : LinearMachine()
```
vigsterkr: call `GLM()` instead of LinearMachine()... this way you dont need the redundancy to call init()....

src/shogun/regression/GLM.cpp
```diff
@@ -0,0 +1,106 @@
...
+{
+	m_tau = tau;
+	m_link_fn = link_fn;
+	m_descend_updater = descend_updater;
+	m_family = family;
+	init();
```
vigsterkr: remove it once you've fixed the ctor i've mentioned above.

src/shogun/regression/GLM.h
```diff
@@ -0,0 +1,91 @@
...
+		 *
+		 * @param descend_updater chosen Descend Updater algorithm
+		 * @param link_fn the link function
+		 * @param Family the family
+		 * @param tau L2-Regularization parameter
+		 */
+		GLM(const std::shared_ptr<DescendUpdater>& descend_updater,
+		    DistributionFamily family, LinkFunction link_fn, float64_t tau);
+
+		virtual ~GLM(){};
```
vigsterkr: i have explicitly stated that these should be `override` instead of virtual. you have marked this resolved, but it has not been changed at all....

src/shogun/regression/GLM.h
```diff
@@ -0,0 +1,91 @@
...
+
+		/** train model
+		 * @param data training data
+		 * @return whether training was successful
+		 */
+		virtual bool train_machine(std::shared_ptr<Features> data = NULL)
```
vigsterkr: it's still protected... and as said before if you would be using `override` then you would get an error actually

src/shogun/regression/GLM.h
```diff
@@ -0,0 +1,91 @@
...
+		{
+			return true;
+		};
+
+		/** @return object name */
+		virtual const char* get_name() const
```
vigsterkr: `override`...

src/shogun/regression/GLM.h
```diff
@@ -0,0 +1,91 @@
...
+	 * gamma, binomial
+	 */
+	class GLM : public LinearMachine
+	{
+	public:
+		/** problem type */
+		MACHINE_PROBLEM_TYPE(PT_REGRESSION);
+
+		GLM();
+		float64_t log_likelihood(
```
vigsterkr: this is nitpicking but still: ctors should be first and then class methods... although i'm not sure whether these really need to be public?

vigsterkr (changes requested):

src/shogun/regression/GLM.h
```diff
@@ -0,0 +1,96 @@
...
+
+		float64_t log_likelihood(
+		    const std::shared_ptr<DenseFeatures<float64_t>>& features,
+		    const std::shared_ptr<Labels>& label);
+
+		SGVector<float64_t> log_likelihood_derivative(
+		    const std::shared_ptr<DenseFeatures<float64_t>>& features,
+		    const std::shared_ptr<Labels>& label);
+
+		virtual bool
```
vigsterkr: you can just drop `virtual`

src/shogun/regression/GLM.h
```diff
@@ -0,0 +1,96 @@
...
+		train_machine(std::shared_ptr<Features> data = NULL) override
+		{
+			return true;
+		};
+
+		/** @return object name */
+		virtual const char* get_name() const override
```
vigsterkr: you can drop virtual here as well.... but get_name is public.

src/shogun/regression/GLM.h
```diff
@@ -0,0 +1,96 @@
...
+		 *
+		 * @param descend_updater chosen Descend Updater algorithm
+		 * @param link_fn the link function
+		 * @param Family the family
+		 * @param tau L2-Regularization parameter
+		 */
+		GLM(const std::shared_ptr<DescendUpdater>& descend_updater,
+		    DistributionFamily family, LinkFunction link_fn, float64_t tau);
+
+		virtual ~GLM() override{};
```
vigsterkr: you can drop virtual...

vigsterkr (changes requested):

src/shogun/regression/GLM.cpp
gf712: The issue is probably that m_descend_updater (`DescendUpdater`) is never initialised, so you are serialising a nullptr, and I think that might be causing issues... @vigsterkr ?

gf712: Hence https://dev.azure.com/shogunml/shogun/_build/results?buildId=3277&view=logs&j=c857963e-3b4d-5192-8776-ae81779de3af&t=542541c1-2650-5e44-f82f-a76d8c4f5255&l=216

vigsterkr: mmm but this should catch that:
https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/io/serialization/JsonSerializer.cpp#L178

src/shogun/regression/GLM.cpp
```diff
@@ -0,0 +1,105 @@
...
+	auto beta = get_w();
+	auto beta0 = get_bias();
+	auto feature_matrix = features->get_feature_matrix();
+	auto res =
+	    linalg::matrix_prod(SGMatrix(beta, 1, beta.vlen), feature_matrix);
+	linalg::add_scalar(res, beta0);
+	auto exponent = linalg::exponent(res);
+	for (auto i = 0; i < vector_count; i++)
+	{
+		lambda[i] = log(1 + exponent[i]);
```
karlnapf, line 36: std::

gf712:
Implemented in #5005

PR #5007, opened 2020-04-18
Adding Log to linalg

Khalifa1997:


gf712:
Could you add a unit test please? There is a file with all the linalg test cases

gf712:
LGTM

Khalifa1997:
> SGVector a{1,2,3,4}

so `SGMatrix<TypeParam> a{1,2,3,4};` ?

gf712:
No, it’s a matrix, so you need to specify it as a column vector SGMatrix<TypeParam> a{{1,2,3,4}};. There should be some examples in that file..

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
```diff
@@ -1003,6 +1003,34 @@ TYPED_TEST(
 		EXPECT_NEAR(c[i] * b[i], a[i], get_epsilon<TypeParam>());
 }
 
+TYPED_TEST(LinalgBackendEigenNonIntegerTypesTest, SGVector_log)
+{
+	const index_t len = 4;
+	TypeParam data[] = {1, 2, 3, 4};
+	SGVector<TypeParam> a(data, len, false);
```
gf712: This is not what I wrote...
```suggestion
	SGVector<TypeParam> a{1, 2, 3, 4};
```

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
```diff
@@ -1003,6 +1003,34 @@ TYPED_TEST(
...
+	EXPECT_NEAR(result[1], 0.30102999566, get_epsilon<TypeParam>());
+	EXPECT_NEAR(result[2], 0.47712125472, get_epsilon<TypeParam>());
+	EXPECT_NEAR(result[3], 0.60205999132, get_epsilon<TypeParam>());
+}
+
+TYPED_TEST(LinalgBackendEigenNonIntegerTypesTest, SGMatrix_log)
+{
+	const index_t n = 2;
+	TypeParam data[] = {1, 2, 3, 4};
+	SGMatrix<TypeParam> a(data, n, n, false);
```
gf712: same here.. Why not use the list initialiser?

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
Khalifa1997: I followed the other examples, I will change it now

gf712: Ah, that is odd... Anyway, what I showed you here is the correct way to do it, because we have a constructor just for this!

Khalifa1997:
> why `log10`? I would expect `log` to be by default base e, no?

I think I saw a tutorial that log base 10 is used, which one do you recommend for this problem?

gf712:
> > why `log10`? I would expect `log` to be by default base e, no?
> 
> I think I saw a tutorial that log base 10 is used, which one do you recommend for this problem?

For a function that says `log` I would expect it to use `log_e`, if you need log10, then you should call it `log10`

Khalifa1997:
> > > why `log10`? I would expect `log` to be by default base e, no?
> > 
> > 
> > I think I saw a tutorial that log base 10 is used, which one do you recommend for this problem?
> 
> For a function that says `log` I would expect it to use `log_e`, if you need log10, then you should call it `log10`

Yea you're right I just looked up the documentation for GLMs, its natural log that's used so I am gonna revert that change. sorry :P

gf712:
> Yea you're right I just looked up the documentation for GLMs, its natural log that's used so I am gonna revert that change. sorry :P

Nw. This is unrelated to GLMs btw, this is a comment about what the function name says and what it actually does!

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
Khalifa1997: Cool I never knew that, It would be quite handy when I add the element-wise divison

Khalifa1997:
@gf712 CI produces this error `error: no matching constructor for initialization of 'SGMatrix<LinalgBackendEigenNonIntegerTypesTest_SGMatrix_log_Test<float>::TypeParam>' (aka 'SGMatrix<float>')
        SGMatrix<TypeParam> a{1, 2, 3, 4};
`

Khalifa1997:
> I think we could use std::log here in a loop that way we keep the number of characters down

you mean `std::log(1)`?

gf712:
> @gf712 CI produces this error `error: no matching constructor for initialization of 'SGMatrix<LinalgBackendEigenNonIntegerTypesTest_SGMatrix_log_Test<float>::TypeParam>' (aka 'SGMatrix<float>') SGMatrix<TypeParam> a{1, 2, 3, 4}; `

You need to pay more attention to the code you write and/or the comments... "No, it’s a matrix, so you need to specify it as a column vector **SGMatrix a{{1,2,3,4}};**. There should be some examples in that file.."

Khalifa1997:
> > @gf712 CI produces this error `error: no matching constructor for initialization of 'SGMatrix<LinalgBackendEigenNonIntegerTypesTest_SGMatrix_log_Test<float>::TypeParam>' (aka 'SGMatrix<float>') SGMatrix<TypeParam> a{1, 2, 3, 4}; `
> 
> You need to pay more attention to the code you write and/or the comments... "No, it’s a matrix, so you need to specify it as a column vector **SGMatrix a{{1,2,3,4}};**. There should be some examples in that file.."

ahh mb

gf712:
> > I think we could use std::log here in a loop that way we keep the number of characters down
> 
> you mean `std::log(1)

As in iterate over the result in a loop and compare to the result of std::log of the original matrix.

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
```diff
@@ -1003,6 +1003,24 @@ TYPED_TEST(
 		EXPECT_NEAR(c[i] * b[i], a[i], get_epsilon<TypeParam>());
 }
 
+TYPED_TEST(LinalgBackendEigenNonIntegerTypesTest, SGVector_log)
+{
+	SGVector<TypeParam> a{{1.0, 2.0, 3.0}};
```
gf712:
```suggestion
	SGVector<TypeParam> a{1.0, 2.0, 3.0};
```

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
```diff
@@ -1003,6 +1003,24 @@ TYPED_TEST(
...
+	auto result = linalg::log(a);
+
+	for (index_t i = 0; i < 3; i++)
+		EXPECT_NEAR(result[i], std::log(i+1), get_epsilon<TypeParam>());
```
gf712: why not
```suggestion
		EXPECT_NEAR(result[i], std::log(a[i]), get_epsilon<TypeParam>());
```

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
```diff
@@ -1003,6 +1003,24 @@ TYPED_TEST(
...
+}
+
+TYPED_TEST(LinalgBackendEigenNonIntegerTypesTest, SGMatrix_log)
+{	
+	SGMatrix<TypeParam> a{{1.0, 2.0, 3.0}};
+	auto result = linalg::log(a);
+
+	for (index_t i = 0; i < 3; i++)
+		EXPECT_NEAR(result[i], std::log(i+1), get_epsilon<TypeParam>());
```
gf712: same here
```suggestion
		EXPECT_NEAR(result[i], std::log(a[i])), get_epsilon<TypeParam>());
```

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
Khalifa1997: This is the same line that didn't work last time..

gf712: No it isn't... the error you showed was with a matrix, this is a vector

Khalifa1997: Ohh I understand now.. I will edit it now and push thanks!

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
```diff
@@ -1003,6 +1003,24 @@ TYPED_TEST(
 		EXPECT_NEAR(c[i] * b[i], a[i], get_epsilon<TypeParam>());
 }
 
+TYPED_TEST(LinalgBackendEigenNonIntegerTypesTest, SGVector_log)
+{
+	SGVector<TypeParam> a{1.0, 2.0, 3.0};
+	auto result = linalg::log(a);
+
+	for (index_t i = 0; i < 3; i++)
```
karlnapf: replace 3 by a.vlen

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
```diff
@@ -1003,6 +1003,24 @@ TYPED_TEST(
...
+		EXPECT_NEAR(result[i], std::log(a[i]), get_epsilon<TypeParam>());
+}
+
+TYPED_TEST(LinalgBackendEigenNonIntegerTypesTest, SGMatrix_log)
+{	
+	SGMatrix<TypeParam> a{{1.0, 2.0, 3.0}};
+	auto result = linalg::log(a);
+
+	for (index_t i = 0; i < 3; i++)
```
karlnapf: same here, make the loop dependent over the matrix size

Khalifa1997: Done.

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
Khalifa1997: Done

karlnapf: No need to write this, as we get an email every time you push :)

karlnapf (commented):
lgtm can be merged once the ci is happy

Khalifa1997:
@karlnapf CI is happy :)

PR #5008, opened 2020-04-18
Added Elementwise division

Khalifa1997:
Added elementwise division with [Eigen's division operator](https://eigen.tuxfamily.org/dox/classEigen_1_1VectorwiseOp.html#aa154a8407885a9ad51504180f6d7a163) one thing to note however, the case whereby there is division by zero isn't yet handled

### DIFF
--- a/src/shogun/mathematics/linalg/LinalgBackendBase.h
+++ b/src/shogun/mathematics/linalg/LinalgBackendBase.h
@@ -405,6 +405,56 @@ namespace shogun
 #undef BACKEND_GENERIC_IN_PLACE_BLOCK_ELEMENT_PROD
 
 /**
+ * Wrapper method of in-place vector elementwise division.
+ *
+ * @see linalg::element_div
+ */
+#define BACKEND_GENERIC_IN_PLACE_VECTOR_ELEMENT_DIV(Type, Container)          \
+	virtual void element_div(                                                 \
+	    const Container<Type>& a, const Container<Type>& b,                    \
+	    Container<Type>& result) const                                         \
+	{                                                                          \
+		not_implemented(SOURCE_LOCATION);;                                                    \
+	}
+		DEFINE_FOR_ALL_PTYPE(
+		    BACKEND_GENERIC_IN_PLACE_VECTOR_ELEMENT_DIV, SGVector)
+#undef BACKEND_GENERIC_IN_PLACE_VECTOR_ELEMENT_DIV
+
+/**
+ * Wrapper method of in-place matrix elementwise division.
+ *
+ * @see linalg::element_div
+ */
+#define BACKEND_GENERIC_IN_PLACE_MATRIX_ELEMENT_DIV(Type, Container)          \
+	virtual void element_div(                                                 \
+	    const Container<Type>& a, const Container<Type>& b,                    \
+	    Container<Type>& result, bool transpose_A, bool transpose_B) const     \
+	{                                                                          \
+		not_implemented(SOURCE_LOCATION);;                                                    \
+	}
+		DEFINE_FOR_ALL_PTYPE(
+		    BACKEND_GENERIC_IN_PLACE_MATRIX_ELEMENT_DIV, SGMatrix)
+#undef BACKEND_GENERIC_IN_PLACE_MATRIX_ELEMENT_DIV
+
+/**
+ * Wrapper method of in-place matrix block elementwise division.
+ *
+ * @see linalg::element_div
+ */
+#define BACKEND_GENERIC_IN_PLACE_BLOCK_ELEMENT_DIV(Type, Container)           \
+	virtual void element_div(                                                 \
+	    const linalg::Block<Container<Type>>& a,                               \
+	    const linalg::Block<Container<Type>>& b, Container<Type>& result,      \
+	    bool transpose_A, bool transpose_B) const                              \
+	{                                                                          \
+		not_implemented(SOURCE_LOCATION);;                                                    \
+	}
+		DEFINE_FOR_ALL_PTYPE(
+		    BACKEND_GENERIC_IN_PLACE_BLOCK_ELEMENT_DIV, SGMatrix)
+
+#undef BACKEND_GENERIC_IN_PLACE_BLOCK_ELEMENT_DIV
+
+/**
  * Wrapper method of in-place exponent method.
  *
  * @see linalg::exponent
--- a/src/shogun/mathematics/linalg/LinalgBackendBase.h
+++ b/src/shogun/mathematics/linalg/LinalgBackendBase.h
@@ -428,7 +428,7 @@ namespace shogun
 #define BACKEND_GENERIC_IN_PLACE_MATRIX_ELEMENT_DIV(Type, Container)          \
 	virtual void element_div(                                                 \
 	    const Container<Type>& a, const Container<Type>& b,                    \
-	    Container<Type>& result, bool transpose_A, bool transpose_B) const     \
+	    Container<Type>& result) const     \
 	{                                                                          \
 		not_implemented(SOURCE_LOCATION);;                                                    \
 	}
@@ -444,8 +444,7 @@ namespace shogun
 #define BACKEND_GENERIC_IN_PLACE_BLOCK_ELEMENT_DIV(Type, Container)           \
 	virtual void element_div(                                                 \
 	    const linalg::Block<Container<Type>>& a,                               \
-	    const linalg::Block<Container<Type>>& b, Container<Type>& result,      \
-	    bool transpose_A, bool transpose_B) const                              \
+	    const linalg::Block<Container<Type>>& b, Container<Type>& result) const                              \
 	{                                                                          \
 		not_implemented(SOURCE_LOCATION);;                                                    \
 	}
--- a/src/shogun/mathematics/linalg/LinalgBackendBase.h
+++ b/src/shogun/mathematics/linalg/LinalgBackendBase.h
@@ -437,23 +437,6 @@ namespace shogun
 #undef BACKEND_GENERIC_IN_PLACE_MATRIX_ELEMENT_DIV
 
 /**
- * Wrapper method of in-place matrix block elementwise division.
- *
- * @see linalg::element_div
- */
-#define BACKEND_GENERIC_IN_PLACE_BLOCK_ELEMENT_DIV(Type, Container)           \
-	virtual void element_div(                                                 \
-	    const linalg::Block<Container<Type>>& a,                               \
-	    const linalg::Block<Container<Type>>& b, Container<Type>& result) const                              \
-	{                                                                          \
-		not_implemented(SOURCE_LOCATION);;                                                    \
-	}
-		DEFINE_FOR_ALL_PTYPE(
-		    BACKEND_GENERIC_IN_PLACE_BLOCK_ELEMENT_DIV, SGMatrix)
-
-#undef BACKEND_GENERIC_IN_PLACE_BLOCK_ELEMENT_DIV
-
-/**
  * Wrapper method of in-place exponent method.
  *
  * @see linalg::exponent
--- a/src/shogun/mathematics/linalg/LinalgBackendViennaCL.h
+++ b/src/shogun/mathematics/linalg/LinalgBackendViennaCL.h
@@ -119,9 +119,9 @@ namespace shogun
 	/** Implementation of @see LinalgBackendBase::element_div */
 	#define BACKEND_GENERIC_IN_PLACE_MATRIX_ELEMENT_DIV(Type, Container) \
 	virtual void element_div(const Container<Type>& a, const Container<Type>& b,\
-		Container<Type>& result, bool transpose_A, bool transpose_B) const \
+		Container<Type>& result) const \
 	{  \
-		element_div_impl(a, b, result, transpose_A, transpose_B); \
+		element_div_impl(a, b, result); \
 	}
 	DEFINE_FOR_ALL_PTYPE(BACKEND_GENERIC_IN_PLACE_MATRIX_ELEMENT_DIV, SGMatrix)
 	#undef BACKEND_GENERIC_IN_PLACE_MATRIX_ELEMENT_DIV
@@ -412,40 +412,16 @@ namespace shogun
 		/** ViennaCL matrix in-place elementwise division method */
 		template <typename T>
 		void element_div_impl(
-		    const SGMatrix<T>& a, const SGMatrix<T>& b, SGMatrix<T>& result,
-		    bool transpose_A, bool transpose_B) const
+		    const SGMatrix<T>& a, const SGMatrix<T>& b, SGMatrix<T>& result) const
 		{
 			GPUMemoryViennaCL<T>* a_gpu = cast_to_viennacl(a);
 			GPUMemoryViennaCL<T>* b_gpu = cast_to_viennacl(b);
 			GPUMemoryViennaCL<T>* result_gpu = cast_to_viennacl(result);
 
-			if (transpose_A && transpose_B)
-				result_gpu->data_matrix(result.num_rows, result.num_cols) =
-				    viennacl::linalg::element_div(
-				        viennacl::trans(
-				            a_gpu->data_matrix(a.num_rows, a.num_cols)),
-				        viennacl::trans(
-				            b_gpu->data_matrix(b.num_rows, b.num_cols)));
-
-			else if (transpose_A)
-				result_gpu->data_matrix(result.num_rows, result.num_cols) =
-				    viennacl::linalg::element_div(
-				        viennacl::trans(
-				            a_gpu->data_matrix(a.num_rows, a.num_cols)),
-				        b_gpu->data_matrix(b.num_rows, b.num_cols));
-
-			else if (transpose_B)
-				result_gpu->data_matrix(result.num_rows, result.num_cols) =
-				    viennacl::linalg::element_div(
-				        a_gpu->data_matrix(a.num_rows, a.num_cols),
-				        viennacl::trans(
-				            b_gpu->data_matrix(b.num_rows, b.num_cols)));
-
-			else
-				result_gpu->data_matrix(result.num_rows, result.num_cols) =
-				    viennacl::linalg::element_div(
-				        a_gpu->data_matrix(a.num_rows, a.num_cols),
-				        b_gpu->data_matrix(b.num_rows, b.num_cols));
+			result_gpu->data_matrix(result.num_rows, result.num_cols) =
+				viennacl::linalg::element_div(
+				    a_gpu->data_matrix(a.num_rows, a.num_cols),
+				    b_gpu->data_matrix(b.num_rows, b.num_cols));
 		}
 		
 		/** ViennaCL logistic method. Calculates f(x) = 1/(1+exp(-x)) */
--- a/src/shogun/mathematics/linalg/LinalgBackendViennaCL.h
+++ b/src/shogun/mathematics/linalg/LinalgBackendViennaCL.h
@@ -116,16 +116,6 @@ namespace shogun
 	DEFINE_FOR_ALL_PTYPE(BACKEND_GENERIC_IN_PLACE_MATRIX_ELEMENT_PROD, SGMatrix)
 	#undef BACKEND_GENERIC_IN_PLACE_MATRIX_ELEMENT_PROD
 
-	/** Implementation of @see LinalgBackendBase::element_div */
-	#define BACKEND_GENERIC_IN_PLACE_MATRIX_ELEMENT_DIV(Type, Container) \
-	virtual void element_div(const Container<Type>& a, const Container<Type>& b,\
-		Container<Type>& result) const \
-	{  \
-		element_div_impl(a, b, result); \
-	}
-	DEFINE_FOR_ALL_PTYPE(BACKEND_GENERIC_IN_PLACE_MATRIX_ELEMENT_DIV, SGMatrix)
-	#undef BACKEND_GENERIC_IN_PLACE_MATRIX_ELEMENT_DIV
-
 	/** Implementation of @see LinalgBackendBase::logistic */
 	#define BACKEND_GENERIC_LOGISTIC(Type, Container) \
 	virtual void logistic(const Container<Type>& a, Container<Type>& result) const \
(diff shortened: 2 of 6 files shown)


gf712:
I guess you get NaN values in zero division? That is the behaviour in C++ I think.
I am not sure I understand why you have the transpose versions?

Khalifa1997:
> I guess you get NaN values in zero division? That is the behaviour in C++ I think.

You are right, It does evaluate to NaN well or -NaN.

> I am not sure I understand why you have the transpose versions?

I wanted to stay consistent with `element_prod` as it also takes transpose as an option, and it's a good feature to add imo

gf712:
> > I guess you get NaN values in zero division? That is the behaviour in C++ I think.
> 
> You are right, It does evaluate to NaN well or -NaN.

Actually it is undefined behaviour, so Eigen will populate the zero division results with an internal representation of `inf`, see https://godbolt.org/z/-WAuL-. @karlnapf should linalg throw? I don't know what happens if for example you invert a singular matrix? In that case you would also have zero division no?

> > I am not sure I understand why you have the transpose versions?
> 
> I wanted to stay consistent with `element_prod` as it also takes transpose as an option, and it's a good feature to add imo

Only if absolutely necessary.. You are adding a lot of code, that takes a long time to compile.. @karlnapf ?

Khalifa1997:
> > > I guess you get NaN values in zero division? That is the behaviour in C++ I think.
> > 
> > 
> > You are right, It does evaluate to NaN well or -NaN.
> 
> Actually it is undefined behaviour, so Eigen will populate the zero division results with an internal representation of `inf`, see https://godbolt.org/z/-WAuL-. @karlnapf should linalg throw? I don't know what happens if for example you invert a singular matrix? In that case you would also have zero division no?
> 
Hmm interesting, I think this can be done by an if condition before division like this, `bool has_zero = (B.array() == 0.0).any();` then I would use the error function incase this line is true

gf712:
> > > > I guess you get NaN values in zero division? That is the behaviour in C++ I think.
> > > 
> > > 
> > > You are right, It does evaluate to NaN well or -NaN.
> > 
> > 
> > Actually it is undefined behaviour, so Eigen will populate the zero division results with an internal representation of `inf`, see https://godbolt.org/z/-WAuL-. @karlnapf should linalg throw? I don't know what happens if for example you invert a singular matrix? In that case you would also have zero division no?
> 
> Hmm interesting, I think this can be done by an if condition before division like this, `bool has_zero = (B.array() == 0.0).any();` then I would use the error function incase this line is true

Yes, but then everyone has to pay the price of that check.. What if I am dividing by something I know can't be zero? For example divide by a vector of distances?

Khalifa1997:
> > > > > I guess you get NaN values in zero division? That is the behaviour in C++ I think.
> > > > 
> > > > 
> > > > You are right, It does evaluate to NaN well or -NaN.
> > > 
> > > 
> > > Actually it is undefined behaviour, so Eigen will populate the zero division results with an internal representation of `inf`, see https://godbolt.org/z/-WAuL-. @karlnapf should linalg throw? I don't know what happens if for example you invert a singular matrix? In that case you would also have zero division no?
> > 
> > 
> > Hmm interesting, I think this can be done by an if condition before division like this, `bool has_zero = (B.array() == 0.0).any();` then I would use the error function incase this line is true
> 
> Yes, but then everyone has to pay the price of that check.. What if I am dividing by something I know can't be zero? For example divide by a vector of distances?

Yea, I understand.. we will see what @karlnapf thinks

src/shogun/mathematics/linalg/LinalgBackendViennaCL.h
```diff
@@ -399,7 +409,45 @@ namespace shogun
 				        a_gpu->data_matrix(a.num_rows, a.num_cols),
 				        b_gpu->data_matrix(b.num_rows, b.num_cols));
 		}
+		/** ViennaCL matrix in-place elementwise division method */
```
karlnapf: no need to populate this, as it doesnt work anyways

src/shogun/mathematics/linalg/backend/eigen/BasicOps.cpp
```diff
@@ -446,6 +477,63 @@ void LinalgBackendEigen::element_prod_impl(
 	result_eig = a_eig.array() * b_eig.array();
 }
 
+/* Helper method to compute elementwise division with Eigen */
```
karlnapf, line 489: @gf712 just another illustration for the fact that we need a data structure for linear algebra. That can for example have a ".T" flag inside :)

karlnapf:
IMO we don't need the transpose stuff. But I agree it is good to stay consistent with what we have. Not sure what to do. In terms of dividing by zero I am clearly in favor of NaN

karlnapf:
lgtm in general, given that the current linalg defines things like this.

gf712:
> IMO we don't need the transpose stuff. But I agree it is good to stay consistent with what we have. Not sure what to do. In terms of dividing by zero I am clearly in favor of NaN

@karlnapf what do you mean with the second part? That we should not throw with zero division and assume the caller knows the risk?

gf712:
> lgtm in general, given that the current linalg defines things like this.

My issue is that with my 16GB RAM laptop I struggle already with compiling shogun in debug mode. All the transpose combinations will make shogun uncompilable soon, for some ppl at least..

src/shogun/mathematics/linalg/backend/eigen/BasicOps.cpp
gf712: yup, like eigen.. Something being worked on in the graph stuff!

karlnapf:
Actually just checked and python throws an error, but numpy array operations return inf. I think following that is good

karlnapf:
let's not have the transpose then. It is not needed, and actually doesnt even make that much sense for element wise operations

Khalifa1997:
> let's not have the transpose then. It is not needed, and actually doesnt even make that much sense for element wise operations

So remove transpose and keep the NaN?

src/shogun/mathematics/linalg/LinalgBackendViennaCL.h
karlnapf: i.e. remove

src/shogun/mathematics/linalg/LinalgNamespace.h
```diff
@@ -1156,6 +1156,165 @@ namespace shogun
...
+			    "Dimension mismatch! A({} x {}) vs B({} x {})", A.m_row_size,
+			    A.m_col_size, B.m_row_size, B.m_col_size);
+
+			require(
+			    A.m_row_size == result.num_rows && A.m_col_size == result.num_cols,
+			    "Dimension mismatch! A({} x {}) vs result({} x {})",
+			    A.m_row_size, A.m_col_size, result.num_rows, result.num_cols);
+
+			require(
+			    !result.on_gpu(),
```
karlnapf: @gf712 as we disabled this iirc. What about a quick refactor where all `.on_gpu` calls are removed?

src/shogun/mathematics/linalg/backend/eigen/BasicOps.cpp
karlnapf: :rocket:

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
```diff
@@ -1002,6 +1002,117 @@ TYPED_TEST(
...
+	{
+		a[i] = i;
+		b[i] = i + 1;
+		c[i] = i;
+	}
+
+	element_div(a, b, a);
+	for (index_t i = 0; i < len; ++i)
+		EXPECT_NEAR(c[i] / b[i], a[i], get_epsilon<TypeParam>());
+}
```
karlnapf, line 1121: pls also test the corner cases where matrix dimensions do not match, or where inf is returned (can be on very small minimal element)

src/shogun/mathematics/linalg/LinalgNamespace.h
gf712: we still have it. What we removed were the calls from SGVector/Matrix if ViennaCL is not used to compile the library

src/shogun/mathematics/linalg/LinalgBackendBase.h
```diff
@@ -404,6 +404,55 @@ namespace shogun
...
+
+/**
+ * Wrapper method of in-place matrix block elementwise division.
+ *
+ * @see linalg::element_div
+ */
+#define BACKEND_GENERIC_IN_PLACE_BLOCK_ELEMENT_DIV(Type, Container)           \
+	virtual void element_div(                                                 \
+	    const linalg::Block<Container<Type>>& a,                               \
+	    const linalg::Block<Container<Type>>& b, Container<Type>& result) const                              \
```
gf712: Do we need/use these?

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
Khalifa1997: I am using something like this
`EXPECT_THROW(element_div(SGMatrix<TypeParam> (m+1, m), d), ShogunException);`
or would you prefer that or should I use smth else

src/shogun/mathematics/linalg/LinalgBackendBase.h
Khalifa1997: I wanted to remain consistent with `element_prod`.. what do you think @karlnapf

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
```diff
@@ -1002,6 +1002,159 @@ TYPED_TEST(
...
+{
+	const auto m = 2;
+	SGMatrix<TypeParam> A(m, m);
+	SGMatrix<TypeParam> B(m, m);
+
+	for (auto i : range(m * m))
+	{
+		A[i] = i;
+		B[i] = i + 1;
+	}
```
gf712: what about `std::iota`? cleans this up a bit

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
```diff
@@ -1002,6 +1002,159 @@ TYPED_TEST(
...
+	const auto m = 2;
+	SGMatrix<TypeParam> A(m, m);
+	SGMatrix<TypeParam> B(m, m);
+	SGMatrix<TypeParam> result(m, m);
+
+	for (auto i : range(m * m))
+	{
+		A[i] = i;
+		B[i] = i + 1;
+	}
```
gf712: `std::iota`

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
```diff
@@ -1002,6 +1002,159 @@ TYPED_TEST(
...
+	for (auto i : range(nrows))
+		for (auto j : range(ncols))
+		{
+			A(i, j) = i * 10 + j + 1;
+			B(j, i) = i + j + 1;
+		}
+
+	const auto m = 2;
+	auto A_block = linalg::block(A, 0, 0, m, m);
+	auto B_block = linalg::block(B, 0, 0, m, m);
```
gf712: are you going to need these block operations? I think these should be added when needed, otherwise it's just binary bloat

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
```diff
@@ -1002,6 +1002,159 @@ TYPED_TEST(
...
+	const index_t len = 4;
+	SGVector<TypeParam> a(len);
+	SGVector<TypeParam> b(len);
+	SGVector<TypeParam> c(len);
+
+	for (index_t i = 0; i < len; ++i)
+	{
+		a[i] = i;
+		b[i] = i + 1;
+	}
```
gf712: `range_fill()`

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
```diff
@@ -1002,6 +1002,159 @@ TYPED_TEST(
...
+	SGVector<TypeParam> a(len);
+	SGVector<TypeParam> b(len);
+	SGVector<TypeParam> c(len);
+
+	for (index_t i = 0; i < len; ++i)
+	{
+		a[i] = i;
+		b[i] = i + 1;
+		c[i] = i;
+	}
```
gf712: `range_fill`

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
Khalifa1997: so do you recommend I remove the whole thing? aka the block `element_div`?

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
```diff
@@ -1002,6 +1002,159 @@ TYPED_TEST(
...
+
+	element_div(a, b, a);
+	for (index_t i = 0; i < len; ++i)
+		EXPECT_NEAR(c[i] / b[i], a[i], get_epsilon<TypeParam>());
```
gf712: btw, there is a `zip_iterator`, so you can write this python style
```cpp
for (const auto& [result, el1, el2]: zip_iterator(a,b,c))
    EXPECT_NEAR(el2 / el1, result, get_epsilon<TypeParam>());
```
Same in all the other loops where you check the result. Another thing that you should try to use in your code, because it cleans it up a lot

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
gf712: yes, unless you need it ofc! :)

gf712: but can wait for @karlnapf input or maybe @vigsterkr has thoughts on this?

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
```diff
@@ -1002,6 +1003,145 @@ TYPED_TEST(
...
+TYPED_TEST(
+    LinalgBackendEigenAllTypesTest, SGMatrix_elementwise_division_in_place)
+{
+	const auto m = 2;
+	SGMatrix<TypeParam> A(m, m);
+	SGMatrix<TypeParam> B(m, m);
+	SGMatrix<TypeParam> result(m, m);
+
+	std::iota(A.matrix, A.matrix + m*2, 0);
+	std::iota(B.matrix, B.matrix + m*2, 1);
```
gf712: why `m*2`? surely `m*m`.. Also, you should get the matrix length with `SGMatrix::size()`

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
```diff
@@ -1002,6 +1003,145 @@ TYPED_TEST(
...
 		EXPECT_NEAR(c[i] * b[i], a[i], get_epsilon<TypeParam>());
 }
+TYPED_TEST(LinalgBackendEigenAllTypesTest, SGMatrix_elementwise_division)
+{
+	const auto m = 2;
+	SGMatrix<TypeParam> A(m, m);
+	SGMatrix<TypeParam> B(m, m);
+
+	std::iota(A.matrix, A.matrix + m*2, 0);
+	std::iota(B.matrix, B.matrix + m*2, 1);
```
gf712: use `SGMatrix::size()`

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
gf712: I guess you put `m*2` because `m*2==m*m` in this case... But this is can lead to bugs in refactors, and is also unclear

Khalifa1997:
This is quite strange, even though the bits of code I had added which should theoretically handle division by zero, it just breaks the whole tests even creating test errors in stuff like elementwise division of matrices of different sizes -even though those particular tests have passed before adding the division by zero test- is there anything I am missing here?.. @gf712 @karlnapf

gf712:
> This is quite strange, even though the bits of code I had added which should theoretically handle division by zero, it just breaks the whole tests even creating test errors in stuff like elementwise division of matrices of different sizes -even though those particular tests have passed before adding the division by zero test- is there anything I am missing here?.. @gf712 @karlnapf

OK, first, the other tests are still passing. 
Second, the issue is that zero division is only defined for floating point types (that's why those test pass if you look at the CI output). Integer types are undefined and therefore throw.
You should add a `if constexpr` to check that the test type is floating point type.

Khalifa1997:
@gf712 is this good to merge? I believe the tests only failed cause it couldn't pull the docker image correctly

gf712:
> @gf712 is this good to merge? I believe the tests only failed cause it couldn't pull the docker image correctly

Just waiting for @karlnapf's input about the block operations

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
```diff
@@ -1002,6 +1004,158 @@ TYPED_TEST(
...
+	EXPECT_THROW(element_div(C, A, B), ShogunException);
+	EXPECT_THROW(element_div(D, A, B), ShogunException);
+	EXPECT_THROW(element_div(E, A, B), ShogunException);
+
+	if constexpr (std::is_floating_point<TypeParam>::value){
+	B.zero();
+	element_div(A, B, result);
+	typename SGMatrix<TypeParam>::EigenMatrixXtMap result_eig = result;
+    	EXPECT_TRUE(result_eig.array().isInf().any() && result_eig.array().isNaN().any());
+	}
```
gf712, line 1071: Could you indent this please

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
```diff
@@ -1002,6 +1004,158 @@ TYPED_TEST(
...
+
+	for (const auto& [result, el1, el2]: zip_iterator(c,b,a))
+    		EXPECT_NEAR(el2 / el1, result, get_epsilon<TypeParam>());
+
+	if constexpr (std::is_floating_point<TypeParam>::value){
+	b.zero();
+	c = element_div(a, b);
+	typename SGVector<TypeParam>::EigenRowVectorXtMap result_eig = c;
+    	EXPECT_TRUE(result_eig.array().isInf().any() && result_eig.array().isNaN().any());
+	}
```
gf712, line 1094: indent

src/shogun/mathematics/linalg/LinalgNamespace.h
```diff
@@ -1156,6 +1156,165 @@ namespace shogun
...
+		 * @return The result of the operation
+		 */
+		template <typename T>
+		SGVector<T> element_div(const SGVector<T>& a, const SGVector<T>& b)
+		{
+			require(
+			    a.vlen == b.vlen, "Dimension mismatch! A({}) vs B({})",
+			    a.vlen, b.vlen);
+
+			SGVector<T> result = a.clone();
```
gf712, line 1251: @karlnapf @vigsterkr  wouldn't it be cheaper in these situations to just allocate an empty vector/matrix? This is an allocation + memcpy, and the copied results will be overwritten

Khalifa1997: I saw the `.clone` in my linalg functions if convenient I could issue a PR to clean them all

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
karlnapf: do it

karlnapf: we just added those for some specific cases we had

src/shogun/mathematics/linalg/LinalgNamespace.h
karlnapf: yes! This is actually something we should fix shogun wide!

karlnapf: #5014

karlnapf: #5014

karlnapf (commented):
LGTM

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
gf712: just this missing

gf712 (changes requested):
Almost ready to merge!

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
Khalifa1997: Do I remove block operation as @karlnapf suggested?

karlnapf: yes let's do that

src/shogun/mathematics/linalg/LinalgNamespace.h
```diff
@@ -1156,6 +1156,125 @@ namespace shogun
...
+		SGMatrix<T> element_div(
+		    const Block<SGMatrix<T>>& A, const Block<SGMatrix<T>>& B)
+		{
+
+			SGMatrix<T> result(A.m_row_size, A.m_col_size);
+
+			element_div(A, B, result);
+
+			return result;
+		}
```
gf712: I thought we decided to remove this?

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
```diff
@@ -1002,6 +1004,121 @@ TYPED_TEST(
...
+	EXPECT_THROW(element_div(C, A, B), ShogunException);
+	EXPECT_THROW(element_div(D, A, B), ShogunException);
+	EXPECT_THROW(element_div(E, A, B), ShogunException);
+
+	if constexpr (std::is_floating_point<TypeParam>::value){
+	B.zero();
+	element_div(A, B, result);
+	typename SGMatrix<TypeParam>::EigenMatrixXtMap result_eig = result;
+    		EXPECT_TRUE(result_eig.array().isInf().any() && result_eig.array().isNaN().any());
+	}
```
gf712, line 1071: Could you please indent the whole block? This is just like any other `if` branch.

tests/unit/mathematics/linalg/operations/Eigen3_operations_unittest.cc
```diff
@@ -1002,6 +1004,121 @@ TYPED_TEST(
...
+	c = element_div(a, b);
+
+	for (const auto& [result, el1, el2]: zip_iterator(c,b,a))
+    		EXPECT_NEAR(el2 / el1, result, get_epsilon<TypeParam>());
+
+	if constexpr (std::is_floating_point<TypeParam>::value){
+	b.zero();
+	c = element_div(a, b);
+	typename SGVector<TypeParam>::EigenRowVectorXtMap result_eig = c;
+    		EXPECT_TRUE(result_eig.array().isInf().any() && result_eig.array().isNaN().any());
```
gf712, line 1093: same here

src/shogun/mathematics/linalg/LinalgNamespace.h
Khalifa1997: mb must of not seen it, fixed it

PR #5012, opened 2020-04-21
Add sin,cos to linalg

jonpsy:
This should be quick, let's add this beauty 

Thoughts, @gf712?

### DIFF
--- a/src/shogun/mathematics/linalg/LinalgBackendEigen.h
+++ b/src/shogun/mathematics/linalg/LinalgBackendEigen.h
@@ -245,6 +245,22 @@ namespace shogun
 		DEFINE_FOR_ALL_PTYPE(BACKEND_GENERIC_LOG, SGMatrix)
 #undef BACKEND_GENERIC_LOG
 
+/** Implementation of @see linalg::sin */
+#define BACKEND_GENERIC_SIN(Type, Container)                              \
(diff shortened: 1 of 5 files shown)


src/shogun/mathematics/linalg/LinalgBackendEigen.h
```diff
@@ -688,6 +704,22 @@ namespace shogun
 		template <typename T>
 		void log_impl(const SGMatrix<T>& a, SGMatrix<T>& result) const;
 
+		/** Eigen3 matrix sin method */
```
karlnapf, line 707: maybe we dont need to say "Eigen3" here since the file is for eigen (I know you just took that from the other code, but still)

karlnapf (commented):
LGTM.
@gf712 looks good? If so feel free to merge.

src/shogun/mathematics/linalg/LinalgBackendEigen.h
jonpsy: Off the record, I can't say I understood our linalg to its entirety(which i'm planning to change :D),  so I did a quick swoop from ahmeds code for the time being. Anyway you want me to change all such occurrences or only the one you mentioned?

karlnapf: no need to change this everywhere (in general when we make comments you dont need to change other code, only your own)

PR #5025, opened 2020-05-04
Replaced clone calls in linalg (fixes #5014)

### LINKED ISSUE #5014: remove `.clone` in linalg methods
Linalg methods that return newly allocated vectors sometimes clone an input vector, instead of just allocating empty memory. That is unnecessary and could be changed.

gf712:


src/shogun/lib/SGMatrix.cpp
```diff
@@ -407,6 +407,12 @@ T* SGMatrix<T>::clone_matrix(const T* matrix, int32_t nrows, int32_t ncols)
 	return result;
 }
 
+template <class T>
+SGMatrix<T> SGMatrix<T>::empty_like(const SGMatrix<T>& other)
```
karlnapf, line 411: what about just `empty`?

karlnapf (approved):

src/shogun/lib/SGMatrix.cpp
gf712: sure, this was just to keep it in line with numpy

karlnapf: ah I see. Think that is good

gf712: shall I merge this like it is now?

Issue #5030, opened 2020-05-14
Clean up constructors


gf712:
Throughout shogun we have various constructors with lots of repetitions, e.g. calls to `register_params`. We could avoid using `register_params` by using [delegating constructors](https://docs.microsoft.com/en-us/cpp/cpp/delegating-constructors?view=vs-2019) and then only have a single constructor responsible for parameter registration. This would avoid bugs with initialization, which lead to undefined behaviour, and parameter registration.

karlnapf:
++
Could you maybe post an example PR where this is done, then people have an easier time reproducing it

gf712:
Sure, I'll start a PR where I pick a class and clean up the constructors and tag it here

jonpsy:
I think I can give this a shot ;D

ellikamishra:
I can try this!

GuoxinYin:
Hello, is anyone working on this right now? If no, I can work on this!

ArhamaQ:
hey, I am  a beginner but have strong C++ knowledge. I think I can resolve this issue. if it is still open. can you assign me this issue?

PR #5031, opened 2020-05-14
SigmoidKernel constructors cleanup

gf712:
Example of how to refactor constructors #5030
For future reference: remember that the member initializer list is guaranteed to be executed before the body of the constructor.


karlnapf (commented):
LGTM

Issue #5026, opened 2020-05-12
Enable models to emit lambda functions.


geektoni:
We might want to emit some information from a given model which may be costly to compute and it could slow down the training ( see #4655 for an example of such scenario ). This is important because those kinds of information would be computed even if the user wants to observe different parameters. 

The solution here would be to emit _lambdas_ instead, which can then be executed to obtain the real value. I think there are two possible ways to proceed:

1.  The object emits a lambda, which is then executed by the `ParameterObserver` (if the user wants to observe that particular parameter). The result of the lambda is then stored and shown to the user. In such a case, the end-user will not manage directly the emitted lambdas and it will see only the final result. We might experience anyway reduced training speed in this scenario since the observer will need to execute the lambda anyway;
2. The object emits a lambda, which is stored as a normal observation by the `ParameterObserver`. The end-user will need to execute this lambda in order to retrieve its value. In such a case, the end-user will be aware of the existence of these lambdas. Basically, besides emitting scalars/vectors/matrices, we introduce also the possibility of emitting functions.

@karlnapf @gf712 what do you think is the best way to implement this?

gf712:
You mean this would be like a promise that can the be computed asynchronously? Could you give an example of a lambda you would want to pass please? From the example you give would it be something like:
```cpp
observe(i, "oob_error", "Out-of-bag Error", [=](){return get_oob_error()};
```
And then you tell the observer whether to compute it?
The problem if you store all these lambdas is that you are storing a copy of the object each time. Also if you mean passing a lambda as an end user if you are using a target interface you won't have access to this feature (I doubt SWIG can handle this, but I think something like [binder](https://github.com/RosettaCommons/binder) can)

geektoni:
@gf712 yes, the example you wrote is exactly what I had in mind :) Then if the user wants to observe the parameter `oob_error`, once the observer captures the lambda, the observer will execute it and it will show to the final user just the result (not the lambda itself). All of this I think will be done asynchronously thanks to RxCpp. 

I guess the second solution would be harder to implement because of the interfaces (SWIG support for lambdas seems quite limited at the moment http://www.swig.org/Doc4.0/SWIGDocumentation.html#CPlusPlus11_lambda_functions_and_expressions)

karlnapf:
We wouldnt want to store the lambda. The class just offers the lambda to the observer, which can decide to call it (and store the information), or not. It shouldnt be valid after the call is over

karlnapf:
[...]

PR #5033, opened 2020-05-14
Extend observe() to emit std::function objects

geektoni:
This addresses #5026. I still need to work around some details, but this is kind of what I had in mind to make it work. At the moment it doesn't pass the tests though. @karlnapf @gf712

### DIFF
--- a/src/shogun/lib/observers/ObservedValueTemplated.h
+++ b/src/shogun/lib/observers/ObservedValueTemplated.h
@@ -28,6 +28,23 @@ namespace shogun
 		 * @param value the observed value
 		 */
 		ObservedValueTemplated(
+		    const int64_t step, const std::string_view name, const T value)
+		    : ObservedValue(step, name), m_observed_value(value)
+		{
+			this->watch_param(
+			    name, &m_observed_value,
+			    AnyParameterProperties(
+			        "Unamed observed value", ParameterProperties::READONLY));
+			m_any_value = make_any(m_observed_value);
+		}
+
+		/**
+		 * Constructor
+		 * @param step step
+		 * @param name the observed value's name
+		 * @param value the observed value
+		 */
+		ObservedValueTemplated(
 		    const int64_t step, const std::string_view name,
 		    const std::string_view description, const T value)
 		    : ObservedValue(step, name), m_observed_value(value)
@@ -53,6 +70,12 @@ namespace shogun
 		{
 			this->watch_param(name, &m_observed_value, properties);
 			m_any_value = make_any(m_observed_value);
+		}
+
+		ObservedValueTemplated(const ObservedValueTemplated& other)
+		    : ObservedValue(other)
+		{
+			m_observed_value = other.m_observed_value;
 		}
 
 		/**
--- a/src/shogun/lib/observers/ParameterObserver.cpp
+++ b/src/shogun/lib/observers/ParameterObserver.cpp
@@ -81,4 +81,4 @@ bool ParameterObserver::observes(const AnyParameterProperties& property)
 index_t ParameterObserver::get_num_observations() const
 {
 	return utils::safe_convert<index_t>(m_observations.size());
-};
+}
--- a/tests/unit/lib/observers/ParameterObserverTest_unittest.cc
+++ b/tests/unit/lib/observers/ParameterObserverTest_unittest.cc
@@ -8,6 +8,7 @@
 
 #include <shogun/lib/config.h>
 
+#include <functional>
 #include <shogun/lib/observers/ParameterObserverLogger.h>
 #include <vector>
 
@@ -18,6 +19,10 @@ class MockEmitter : public SGObject
 public:
 	MockEmitter() : SGObject()
 	{
+		this->watch_param(
+		    "member_variable", &member_variable,
+		    AnyParameterProperties(
+		        "A member variable", ParameterProperties::NONE));
 	}
 	virtual ~MockEmitter()
 	{
@@ -41,7 +46,18 @@ public:
 		observe<int32_t>(1, "a", 1, p);
 		observe<int32_t>(1, "b", 1, p2);
 		observe<int32_t>(1, "None", 1, p3);
+
+		std::function<int32_t()> lambda_function = [this]() {
+			this->member_variable++;
+			return 20;
+		};
+
+		observe<std::function<int32_t()>>(
+		    1, "method", "method description", lambda_function);
 	}
+
+private:
+	int32_t member_variable = 0;
 };
 
 class ParameterObserverTest : public ::testing::Test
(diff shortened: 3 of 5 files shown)


src/shogun/base/SGObject.h
```diff
@@ -1129,6 +1129,37 @@ class SGObject: public std::enable_shared_from_this<SGObject>
...
+	 * @param step time step
+	 * @param name name of this function
+	 * @param method the function we want to emit
+	 */
+	template <class T>
+	void observe(
+		const int64_t step, std::string_view name,
+		std::function<T()> method) const
+	{
+		if (get_num_subscriptions() == 0)
```
karlnapf: I guess we could have more involved filters here right?
Like. "check if user requested to observe this name"?

src/shogun/lib/observers/ObservedValueTemplated.h
```diff
@@ -21,6 +21,23 @@ namespace shogun
...
+		 * @param value the observed value
+		 */
+		ObservedValueTemplated(
+		    const int64_t step, const std::string_view name, const T value)
+		    : ObservedValue(step, name), m_observed_value(value)
+		{
+			this->watch_param(
+			    name, &m_observed_value,
+			    AnyParameterProperties(
+			        "Unamed observed value", ParameterProperties::READONLY));
```
karlnapf, line 37: un-named? (also there is a type)

tests/unit/lib/observers/ParameterObserverTest_unittest.cc
```diff
@@ -41,6 +41,10 @@ class MockEmitter : public SGObject
 		observe<int32_t>(1, "a", 1, p);
 		observe<int32_t>(1, "b", 1, p2);
 		observe<int32_t>(1, "None", 1, p3);
+
+		std::function<int32_t()> lambda_function = []() { return 10; };
```
karlnapf: I think this method should have something in it that changes its state. So that you later on can assert that it was *not* called

tests/unit/lib/observers/ParameterObserverTest_unittest.cc
```diff
@@ -138,6 +142,18 @@ TEST_F(ParameterObserverTest, filter_all_property)
...
+TEST_F(ParameterObserverTest, filter_method)
+{
+	std::vector<std::string> method_params = {"method"};
+	std::shared_ptr<ParameterObserver> observer(
+	    new ParameterObserverLogger(method_params));
+	emitter.subscribe(observer);
+	emitter.emit_value();
+
+	EXPECT_EQ(observer->get<int32_t>("num_observations"), 1);
+    EXPECT_EQ(observer->get_observation(0)->get_any().as<int32_t>(), 10);
```
karlnapf: nice.
Another test should be to test that the method is not called as written above

src/shogun/base/SGObject.h
geektoni: mmh it won't be exactly trivial to do so. SGObject does not store directly the observers, just their "subscriptions". The filtering is done automatically by RxCpp (have a look at the `SGObject::subscribe`).

src/shogun/lib/observers/ObservedValueTemplated.h
geektoni: I'm not sure about this constructor. I may change it later on.

tests/unit/lib/observers/ParameterObserverTest_unittest.cc
geektoni: It makes sense. I will update the tests accordingly.

src/shogun/lib/observers/ParameterObserver.cpp
```diff
@@ -82,3 +82,17 @@ index_t ParameterObserver::get_num_observations() const
 {
 	return utils::safe_convert<index_t>(m_observations.size());
 };
+
+std::shared_ptr<ObservedValue> ParameterObserver::execute_lambda(const TimedObservedValue &value)
+{
+    try
+    {
+        auto func =
+            value.first->get_any().as<std::function<std::shared_ptr<ObservedValue>()>>();
```
geektoni: @gf712 I have a question here. Let's say my `Any` object contains an `ObservedValueTemplated` pointer object. The class `ObservedValueTemplated` inherits from `ObservedValue`. Ideally, I should be able to cast my `Any` to an `ObservedValue` pointer object right (e.g., `any.as<std::shared_ptr<ObservedValue>>()`)?

I thought it would have been possible to do so thanks to the inheritance, but it does not seem to work :/ is this an intended behaviour (or am I missing something)?

gf712: Yea, it doesn't work off the bat. You need to register that inheritance using
https://github.com/shogun-toolbox/shogun/blob/7014676a95ec27b51473f037c4c137083075d61e/src/shogun/lib/any.h#L1259-L1275

gf712: This is basically a vtable, but it is only populated with SGObject derived classes to SGObject and some primitive types + std::string. I guess you could always register it there? @vigsterkr

src/shogun/base/SGObject.h
karlnapf: I see....then we are kind of back to square one right?
At least the code is now here rather than in the rf model ....

geektoni: Well, the main difference is that we can at least limit the execution of expensive functions now. 

For instance, in the BaggingMachine case, the `compute_oob()` method was called before emitting all observations. Now, the `compute_oob()` is called only if an observable wants it and it is executed by the observable rather than the model.

karlnapf: ah maybe I got that wrong then. Cool! :)

tests/unit/lib/observers/ParameterObserverTest_unittest.cc
```diff
@@ -41,7 +46,18 @@ class MockEmitter : public SGObject
 		observe<int32_t>(1, "a", 1, p);
 		observe<int32_t>(1, "b", 1, p2);
 		observe<int32_t>(1, "None", 1, p3);
+
+		std::function<int32_t()> lambda_function = [this]() {
+			this->member_variable++;
+			return 20;
+		};
+
+		observe<std::function<int32_t()>>(
```
geektoni, line 55: Apparently, `Any` is smarter than I thought :) with the current implementation we can emit `std::function` directly without changing anything in the framework. We can also execute them in a fairly simple way by just calling `get<lambda return type>("name of the lambda")`. See the unit tests for an example.

Next question. Some of these operations could be quite costly and if the user calls them multiple times then they will be executed from scratch every time. Shall we somehow cache the result of these executions, so that the user will just need to call the lambda once and then it will retrieve only the pre-computed result?

This caching will happen only for the stored observations though.

karlnapf: could you make an example of such an operation?

geektoni: > could you make an example of such an operation?

mmh I don't have any real example now. I was thinking about possible use cases/corner cases that we may face.

gf712: @geektoni I am modifying things a bit in #5037, so we use visitor patterns instead of `any_cast`. The only things you will have to keep in mind is that the visitor will only call a function when you get it if you have the `ParameterProperties::FUNCTION`, which is added in that PR. About the caching, you could use a `static` variable inside the lambda.

gf712: actually, maybe don't use static, that might cause massive issues

gf712: If you want to cache actually, use `std::optional`. Put the result there. If it is set it returns the value, otherwise computes it.

PR #5036, opened 2020-05-15
#5031, Kernel constructor cleanup

jonpsy:
Ding dong! , did anyone order cleanup? #5030 
@gf712

### DIFF
--- a/src/shogun/kernel/ANOVAKernel.cpp
+++ b/src/shogun/kernel/ANOVAKernel.cpp
@@ -7,26 +7,27 @@
 
 #include <shogun/mathematics/Math.h>
 #include <shogun/kernel/ANOVAKernel.h>
-#include <shogun/mathematics/Math.h>
 
 using namespace shogun;
 
-ANOVAKernel::ANOVAKernel(): DotKernel(0), cardinality(1.0)
+ANOVAKernel::ANOVAKernel(): DotKernel(0)
 {
-	register_params();
+	SG_ADD(
+	    &m_cardinality, "cardinality", "Kernel cardinality.",
+	    ParameterProperties::HYPER);
 }
 
-ANOVAKernel::ANOVAKernel(int32_t cache, int32_t d)
-: DotKernel(cache), cardinality(d)
+ANOVAKernel::ANOVAKernel(int32_t size, int32_t d)
+: ANOVAKernel()
 {
-	register_params();
+	set_cache_size(size);
+	m_cardinality = d;
 }
 
 ANOVAKernel::ANOVAKernel(
 	const std::shared_ptr<DenseFeatures<float64_t>>& l, const std::shared_ptr<DenseFeatures<float64_t>>& r, int32_t d, int32_t cache)
-  : DotKernel(cache), cardinality(d)
+  : ANOVAKernel(cache, d)
 {
-	register_params();
 	init(l, r);
 }
 
@@ -88,20 +89,14 @@ float64_t ANOVAKernel::compute_rec2(int32_t idx_a, int32_t idx_b)
 	return result;
 }
 
-void ANOVAKernel::register_params()
-{
-	SG_ADD(&cardinality, "cardinality", "Kernel cardinality.", ParameterProperties::HYPER);
-}
-
-
 float64_t ANOVAKernel::compute_recursive1(float64_t* avec, float64_t* bvec, int32_t len)
 {
-	int32_t DP_len=(cardinality+1)*(len+1);
+	int32_t DP_len=(m_cardinality+1)*(len+1);
 	float64_t* DP = SG_MALLOC(float64_t, DP_len);
 
 	ASSERT(DP)
-	int32_t d=cardinality;
-	int32_t offs=cardinality+1;
+	int32_t d=m_cardinality;
+	int32_t offs=m_cardinality+1;
 
 	ASSERT(DP_len==(len+1)*offs)
 
@@ -128,15 +123,15 @@ float64_t ANOVAKernel::compute_recursive1(float64_t* avec, float64_t* bvec, int3
 
 float64_t ANOVAKernel::compute_recursive2(float64_t* avec, float64_t* bvec, int32_t len)
 {
-	float64_t* KD = SG_MALLOC(float64_t, cardinality+1);
-	float64_t* KS = SG_MALLOC(float64_t, cardinality+1);
+	float64_t* KD = SG_MALLOC(float64_t, m_cardinality+1);
+	float64_t* KS = SG_MALLOC(float64_t, m_cardinality+1);
 	float64_t* vec_pow = SG_MALLOC(float64_t, len);
 
 	ASSERT(vec_pow)
 	ASSERT(KS)
 	ASSERT(KD)
 
-	int32_t d=cardinality;
+	int32_t d=m_cardinality;
 	for (int32_t i=0; i < len; i++)
 		vec_pow[i] = 1;
 
@@ -171,18 +166,4 @@ float64_t ANOVAKernel::compute_recursive2(float64_t* avec, float64_t* bvec, int3
 	SG_FREE(KD);
 
 	return result;
-}
-
-std::shared_ptr<ANOVAKernel> ANOVAKernel::obtain_from_generic(const std::shared_ptr<Kernel>& kernel)
-{
-	if (!kernel)
-		return NULL;
-
-	require(kernel->get_kernel_type()==K_ANOVA, "Provided kernel is "
-				"not of type CANOVAKernel, but type {}!",
-				kernel->get_kernel_type());
-
-	/* since an additional reference is returned */
-
-	return std::static_pointer_cast<ANOVAKernel>(kernel);
 }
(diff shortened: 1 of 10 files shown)


src/shogun/kernel/ANOVAKernel.h
```diff
@@ -105,32 +86,21 @@ class ANOVAKernel: public DotKernel
...
-
-	/** register params */
-	void register_params();
 
 private:
 	float64_t compute_recursive1(float64_t* avec, float64_t* bvec, int32_t len);
 	float64_t compute_recursive2(float64_t* avec, float64_t* bvec, int32_t len);
 
 protected:
+	virtual float64_t compute(int32_t idx_a, int32_t idx_b);
```
gf712, line 96: minor, but might be worth leaving compute where it was? So that the class members are separate from the class functions?

gf712:
Thanks! Are you going to push more kernel refactors here or shall I merge when it passes CI?

src/shogun/kernel/ANOVAKernel.cpp
```diff
@@ -7,44 +7,36 @@
...
 }
 
 ANOVAKernel::ANOVAKernel(
 	const std::shared_ptr<DenseFeatures<float64_t>>& l, const std::shared_ptr<DenseFeatures<float64_t>>& r, int32_t d, int32_t cache)
-  : DotKernel(cache), cardinality(d)
+  : ANOVAKernel(cache,d)
 {
-	register_params();
-	init(l, r);
+	ASSERT(DotKernel::init(l, r));
```
karlnapf: I am curious, why is that changed?

src/shogun/kernel/ANOVAKernel.h
```diff
@@ -105,32 +86,21 @@ class ANOVAKernel: public DotKernel
...
 	float64_t compute_rec2(int32_t idx_a, int32_t idx_b);
 
-	/** Casts the given kernel to CANOVAKernel.
-	 * @param kernel Kernel to cast. Must be CANOVAKernel. Might be NULL
-	 * @return casted CANOVAKernel object, NULL if input was NULL
+	/** Casts the given kernel to ANOVAKernel.
+	 * @param kernel Kernel to cast. Must be ANOVAKernel. Might be NULL
+	 * @return casted ANOVAKernel object, NULL if input was NULL
 	 */
 	static std::shared_ptr<ANOVAKernel> obtain_from_generic(const std::shared_ptr<Kernel>& kernel);
```
karlnapf: I think these are not used anymore, so can be remove (see what happens if you do that)

src/shogun/kernel/AUCKernel.h
```diff
@@ -54,66 +52,19 @@ namespace shogun
...
-			return "AUCKernel";
-		}
-
-		/** return feature class the kernel can deal with
-		 *
-		 * @return feature class SIMPLE
-		 */
-		virtual EFeatureClass get_feature_class()
-		{
-			return C_DENSE;
```
karlnapf, line 89: removing the whole method? Does the superclass already have it?

src/shogun/kernel/ANOVAKernel.h
jonpsy: I see your point

src/shogun/kernel/AUCKernel.h
jonpsy: No. Scroll down a bit, I just used inline and made it into one line not removed

src/shogun/kernel/ANOVAKernel.cpp
jonpsy: Just removed init(l,r) and pasted its code here. Also DotKernel::init(l,r) returns a bool, so I asserted that.

jonpsy:
> Thanks! Are you going to push more kernel refactors here or shall I merge when it passes CI?

There are a lot of kernels, some of them offer subtle complications. Also that I'm cleaning the docs too and the internal member names as well, so that'd be difficult for you guys to review(in case something bad goes down). I was instead planning on sending these in rounds.

EDIT: I'll add a few more simple cases, I'll @ you when I'm done ? :)  
EDIT2: Please merge? @gf712  :D

src/shogun/kernel/ANOVAKernel.h
jonpsy: Makes sense, removed it.

src/shogun/kernel/AUCKernel.h
karlnapf: ah I see, sorry :)

src/shogun/kernel/AUCKernel.h
```diff
@@ -54,66 +52,20 @@ namespace shogun
...
 
-		/** return feature class the kernel can deal with
-		 *
-		 * @return feature class SIMPLE
-		 */
-		virtual EFeatureClass get_feature_class()
-		{
-			return C_DENSE;
-		}
+		inline virtual const char* get_name() const { return "AUCKernel"; }
```
karlnapf: `inline` can be removed, it does nothing for us.
if you want to force use `SG_FORCE_INLINE` but I dont think that is needed here

gf712: Yea, the compiler will do that for you since it’s such a small cost to inline it

src/shogun/kernel/LogKernel.h
```diff
@@ -56,47 +56,23 @@ class LogKernel: public Kernel
...
-	 */
 	virtual float64_t compute(int32_t idx_a, int32_t idx_b);
 
-private:
-	void init();
-
 protected:
-	/// distance to be used
-	std::shared_ptr<Distance> m_distance;
+	std::shared_ptr<Distance> m_distance = NULL;
```
gf712: No need to set a shared ptr to NULL, the default constructor sets the pointer to nullptr

src/shogun/kernel/LogKernel.cpp
```diff
@@ -11,46 +11,32 @@
...
-{
-	init();
-	ASSERT(m_distance)
-	
-	init(std::move(l), std::move(r));
+: LogKernel(10, degree, std::move(dist))
+{	
+	Kernel::init(l,r);
+	ASSERT(m_distance->init(l,r));
+	init_normalizer();
```
gf712, line 34: Where did this come from?

src/shogun/kernel/LogKernel.cpp
```diff
@@ -11,46 +11,32 @@
...
-	
-}
-
-bool LogKernel::init(std::shared_ptr<Features> l, std::shared_ptr<Features> r)
-{
-	ASSERT(m_distance)
-	Kernel::init(l,r);
-	m_distance->init(l,r);
-	return init_normalizer();
-}
```
gf712, line 48: Deleting this is getting you the build error. This is implementing a pure virtual method from kernel

src/shogun/kernel/LogKernel.cpp
jonpsy: removing its init(l,r) method (used only inside LogKernel constructor and pasted the code here.

gf712: Ok, but you need to keep the init(l,r) method so I guess just leave it there

jonpsy: Do we absolutely need to? I removed init(l,r) from other classes as well.

src/shogun/kernel/LogKernel.h
jonpsy: I modified the default constructor to do only SG_ADD. So I guess I'd need to set default values to internal members no?

src/shogun/kernel/LogKernel.cpp
gf712: Yes
https://github.com/shogun-toolbox/shogun/blob/7014676a95ec27b51473f037c4c137083075d61e/src/shogun/kernel/Kernel.h#L455

gf712: Actually it isn’t pure virtual so I guess you don’t need to

src/shogun/kernel/LogKernel.h
gf712: No need for shared ptr, it will by default own a nullptr

src/shogun/kernel/LogKernel.cpp
jonpsy: Very well, so I'll restore the init(l,r) methods. Maybe we can do something about them in future

src/shogun/kernel/LogKernel.h
gf712: But primitive types require initialisation yes, otherwise it’s undefined behaviour

src/shogun/kernel/LogKernel.cpp
jonpsy: > Actually it isn’t pure virtual so I guess you don’t need to

Yes but there are some methods (other than constructors) which are calling init(l,r).

gf712: Ah ok, then yes keep it

gf712:
Thanks! :)

PR #5037, opened 2020-05-18
Use visitors in put and get

gf712:
This replaces the `make_any` in `put` with a visitor pattern (saves creating a new Any and replacing the old with the new one; instead we change the stored value directly with the visitor). For symmetry I also do the same with `get`, but I don't think it makes any difference to using `any_cast`. Also had to change some  of the getter signatures to have access to mutable and immutable references.

This will be really useful to register things like `std::optional` and `std::variant` without changing the functionality in `Any`.

### DIFF
--- a/.ci/ci.yml
+++ b/.ci/ci.yml
@@ -62,6 +62,7 @@ jobs:
 - job: swig
   displayName: Linux SWIG Interface
   dependsOn: libshogun
+  timeoutInMinutes: 90
 
   pool:
     vmImage: ubuntu-16.04
--- a/src/shogun/base/AnyParameter.h
+++ b/src/shogun/base/AnyParameter.h
@@ -46,6 +46,8 @@ namespace shogun
 
 		CONSTRAIN = 1u << 13,
 
+		FUNCTIONAL = 1u << 14,
+
 		ALL = std::numeric_limits<uint32_t>::max(),
 	};
 
@@ -182,7 +184,12 @@ namespace shogun
 		{
 		}
 
-		Any get_value() const
+		const Any& get_value() const
+		{
+			return m_value;
+		}
+
+		Any& get_value()
 		{
 			return m_value;
 		}
--- a/src/shogun/base/AnyParameter.h
+++ b/src/shogun/base/AnyParameter.h
@@ -39,14 +39,17 @@ namespace shogun
 		// has automatically computed value
 		AUTO = 1u << 10,
 
+		// is const
 		READONLY = 1u << 11,
 
-		// an executable function
-		RUNFUNCTION = 1u << 12,
+		// a class member function with side effects
+		FUNCTION = 1u << 12,
 
-		CONSTRAIN = 1u << 13,
+		// a const class member function
+		CONSTFUNCTION = 1u << 13,
 
-		FUNCTIONAL = 1u << 14,
+		// parameter is constrained, e.g. has to be positive
+		CONSTRAIN = 1u << 14,
 
 		ALL = std::numeric_limits<uint32_t>::max(),
 	};
@@ -60,7 +63,8 @@ namespace shogun
 	        {ParameterProperties::SETTING, "SETTING"},
 	        {ParameterProperties::AUTO, "AUTO"},
 	        {ParameterProperties::READONLY, "READONLY"},
-	        {ParameterProperties::RUNFUNCTION, "RUNFUNCTION"},
+	        {ParameterProperties::FUNCTION, "FUNCTION"},
+	        {ParameterProperties::CONSTFUNCTION, "CONSTFUNCTION"},
 	        {ParameterProperties::CONSTRAIN, "CONSTRAIN"}};
 
 	enableEnumClassBitmask(ParameterProperties);
--- a/src/shogun/base/SGObject.h
+++ b/src/shogun/base/SGObject.h
@@ -543,7 +543,7 @@ public:
 		const auto& value = param.get_value();
 		try
 		{
-			if (param.get_properties().has_property(ParameterProperties::FUNCTIONAL))
+			if (param.get_properties().has_property(ParameterProperties::CONSTFUNCTION))
 			{
 				ParameterGetterInterface<ReturnType, std::function<ReturnType()>> visitor{result};
 				value.visit_with(&visitor);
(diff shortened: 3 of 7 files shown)


src/shogun/base/SGObject.h
```diff
@@ -298,30 +310,20 @@ class SGObject: public std::enable_shared_from_this<SGObject>
 	 */
 	template <typename T,
 		      typename std::enable_if_t<!is_string<T>::value>* = nullptr>
-	void put(const Tag<T>& _tag, const T& value) noexcept(false)
```
gf712, line 301: removed these as they don't do anything for us. This is just useful if the template type determines if it is noexcept

src/shogun/base/SGObject.h
```diff
@@ -519,68 +521,48 @@ class SGObject: public std::enable_shared_from_this<SGObject>
...
-		// we won't be there
-		return any_cast<T>(value);
-	}
 
-	template <typename T, typename std::enable_if_t<is_sg_base<T>::value>* = nullptr>
-	std::shared_ptr<T> get(const Tag<T>& _tag) const noexcept(false)
-	{
-		const Any value = get_parameter(_tag).get_value();
-		try
+		if constexpr (is_string<T>::value)
```
gf712, line 523: simplified a bunch of overloads with `if constexpr`

src/shogun/base/SGObject.h
```diff
@@ -784,6 +766,38 @@ class SGObject: public std::enable_shared_from_this<SGObject>
...
+		else
+		{
+			Any::register_visitor<T, ParameterGetterInterface<T, T>>(
+				[](T* value, auto* visitor)
+				{
+					visitor->m_value = *value;	
+				}
+			);
+		}
+	}
```
gf712, line 786: @vigsterkr I adapted from your PR. Can then put all the visitor registrations in this function

src/shogun/base/SGObject.h
```diff
@@ -1065,15 +1087,70 @@ class SGObject: public std::enable_shared_from_this<SGObject>
 	 * @param _tag name information of parameter
 	 * @param value new value of parameter
 	 */
-	void update_parameter(const BaseTag& _tag, const Any& value, bool do_checks = true);
+	template <typename T>
+	void update_parameter(const BaseTag& _tag, const T& value, bool do_checks = true)
```
gf712, line 1078: made this template dependent so that we can have all these parameter handling branches in a single location

src/shogun/base/SGObject.h
```diff
@@ -1065,15 +1087,70 @@ class SGObject: public std::enable_shared_from_this<SGObject>
...
+			if (!msg.empty())
+			{
+				require(!do_checks,
+					"{}::{} cannot be updated because it must be: {}!",
+					get_name(), _tag.name().c_str(), msg.c_str());
+			}
+		}
+		if constexpr (std::is_same_v<T, Any>)
+		{
+			param.set_value(value);
```
gf712, line 1100: this is needed right now for auto init, but I will refactor it in another PR and can then remove this

src/shogun/lib/any.h
```diff
@@ -1229,7 +1225,7 @@ namespace shogun
 		static void register_caster(std::function<To(From)> caster);
 
 		template <class Type, class State>
-		static void register_visitor(std::function<void(Type, State*)> visitor);
+		static void register_visitor(std::function<void(Type*, State*)> visitor);
```
gf712, line 1228: it is more convenient to pass around pointers rather than values

src/shogun/base/SGObject.h
```diff
@@ -1160,6 +1237,7 @@ class SGObject: public std::enable_shared_from_this<SGObject>
 	{
 		auto param = this->get_parameter(BaseTag(name));
 		auto cloned = any_cast<T>(param.get_value());
+		param.get_properties().remove_property(ParameterProperties::FUNCTIONAL);
```
gf712: This is a bit awkward, but the observers pass on properties as is, meaning the value of a function call still has a FUNCTIONAL property, eg cluster centres in KMeans unit test

src/shogun/base/SGObject.h
```diff
@@ -930,10 +949,12 @@ class SGObject: public std::enable_shared_from_this<SGObject>
 	{
 		AnyParameterProperties properties(
 			"Dynamic parameter",
-			ParameterProperties::READONLY);
-		std::function<T()> bind_method =
-			std::bind(method, dynamic_cast<const S*>(this));
+			ParameterProperties::READONLY | ParameterProperties::FUNCTIONAL);
+		std::function<T()> bind_method = [this, method](){
+			return (static_cast<const S*>(this) ->* method)();
+		};
```
gf712, line 941: this should be more efficient than std::bind

src/shogun/base/AnyParameter.h
```diff
@@ -39,12 +39,17 @@ namespace shogun
 		// has automatically computed value
 		AUTO = 1u << 10,
 
+		// is const
 		READONLY = 1u << 11,
 
-		// an executable function
-		RUNFUNCTION = 1u << 12,
+		// a class member function with side effects
```
karlnapf, line 45: do we want to allow this even? I'd prefer if we didnt ... landslide risk

gf712: This is used by some classes, like RandomMixin to set the random seed (`set_random_seed` is not const). The safety net we have here is that you can only call these functions using `run`, rather than `get`.

.ci/ci.yml
```diff
@@ -62,6 +62,7 @@ jobs:
 - job: swig
   displayName: Linux SWIG Interface
   dependsOn: libshogun
+  timeoutInMinutes: 90
```
vigsterkr, line 65: as well? :D

src/shogun/base/SGObject.h
vigsterkr: 👍

vigsterkr (approved):
⛵

.ci/ci.yml
gf712: it should be unrelated. It's just with the python notebook job it can go just above 60 minutes, so just set it to 90

PR #5044, opened 2020-05-22
Removed link to AppVeyor CL

craftyshark:
We no longer use AppVeyor CL. The link in the windows Ready to Build packages led to a dead link. Not sure if we want any more clarification in that section.


gf712:
I think this is fine for now, thanks! We still need to get all the tests passing with MSVC first anyway.

PR #5046, opened 2020-05-23
handle nullptr by returning None in Python

gf712:
Currently getting a nullptr, e.g. `k.get("lhs")` when kernel is uninitialized, will cause a segfault when Python tries to print it (this can easily happen when using the REPL. The solution, imo, is to return Python `None` in this situation.


karlnapf:
I agree!

PR #5048, opened 2020-05-29
compile static libspdlog

gf712:
Compile spdlog static library. This makes it easier to link libshogun to other external libraries. We already use this in the plugins branch


cmake/external/spdlog.cmake
```diff
@@ -1,20 +1,45 @@
...
+
 include(ExternalProject)
 ExternalProject_Add(
-	spdlog
-	PREFIX ${CMAKE_BINARY_DIR}/spdlog
+	SpdLog
+	PREFIX ${spdlog_PREFIX}
 	DOWNLOAD_DIR ${spdlog_SOURCE_DIR}
 	SOURCE_DIR ${spdlog_SOURCE_DIR}
 	URL https://github.com/gabime/spdlog/archive/v1.3.1.tar.gz
```
gf712, line 17: I wanted to upgrade to 1.5 but this causes issues with finding spdlog in the build folder for some reason

karlnapf:
++

PR #5051, opened 2020-06-02
zip_iterator and enumerate for lvalues and rvalues

### LINKED ISSUE #5050: Add temporary value support for zip_iterator
Currently zip_iterator can only handle lvalues, but it would be nice to have it work with rvalues.
This would probably require quite a bit of template programming and refactoring `zip_iterator`.

What currently works
```cpp
auto dot_iterator = DotIterator(feat);
for (const auto& [X, y]: zip_iterator(dot_iterator, labels)] {
...
}
```

What should also work
```cpp
for (const auto& [X, y]: zip_iterator(DotIterator(feat), labels)] {
...
}
```

Currently all the iterable objects passed to `zip_iterator` are stored by reference in a tuple to avoid copying them, but this means that `zip_iterator(DotFeatures(feat), labels)` gives a compile time error as we are trying to access a reference to a temporary. We would have to somehow have a constructor and tuple type pack that would have either a lvalue or reference to lvalue depending on the iterator type.

```cpp
// given the instantiation
zip_iterator(DotIterator(feat), labels)
// this iterator would have a tuple type:
std::tuple<DotIterator, std::shared_ptr<Labels>&> m_container_tuples;
```

gf712:
This fixes #5050, but we need to have a move constructor in the class iterators e.g. `DotIterator::iterator`, otherwise it will use the copy constructor.
See https://godbolt.org/z/L6P2e7. `B` does not have a move constructor, so it is constructed and copied, whereas `A` is constructed and moved.

### DIFF
--- a/src/shogun/classifier/Perceptron.cpp
+++ b/src/shogun/classifier/Perceptron.cpp
@@ -15,6 +15,7 @@
 #include <shogun/lib/observers/ObservedValueTemplated.h>
 #include <shogun/mathematics/Math.h>
 #include <shogun/mathematics/linalg/LinalgNamespace.h>
+#include <shogun/util/zip_iterator.h>
 
 using namespace shogun;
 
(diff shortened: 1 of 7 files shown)


src/shogun/classifier/Perceptron.cpp
```diff
@@ -62,12 +63,9 @@ void Perceptron::iteration()
 	SGVector<float64_t> w = get_w();
 
 	auto labels = binary_labels(m_labels)->get_int_labels();
-	auto iter_train_labels = labels.begin();
 
-	for (const auto& v : DotIterator(features))
+	for (const auto& [v, true_label] : zip_iterator(DotIterator(features), labels))
```
gf712, line 67: @karlnapf we can then replace all of these loops

karlnapf: yes that is nice!

karlnapf (approved):

Issue #5045, opened 2020-05-23
Should put and get do implicit conversions?


gf712:
When we call `put` and `get` we expect the cast types to match exactly:
https://github.com/shogun-toolbox/shogun/blob/93a649bb9bacfd342875eae40115d37e74d1f0f8/tests/unit/base/SGObject_unittest.cc#L486-L487
However, does this make sense? In C++ it is valid for me to call a function with an implicit conversion:
```cpp
void f(double a);
int main()
{
    f(1);
}
```
Similarly, it is valid to implicitly cast a return type (so a getter can return int, but we store it as float64_t). So the test from above would fail, because the types would be casted implicitly, if we were using a getter (rather than `SGObject::get`):
```cpp
float64_t data = get_int(); // no compiler error
```

From a technical perspective this is possible to implement. My question is: would it bring more bugs than it would avoid? For example, on one hand a user might lose information by casting float to int, but on the other the program won't crash after 10 hours of training because of the type mismatch. I would also not want shogun to become like JS where everything has *some* result because the type system is too flexible.

karlnapf:
I'd prefer if we had a more relaxed typing with the parameter framework. Then we also wouldnt have to do these horrendeous workarounds in swig that I added to avoid problems with this.
E.g. in java everything is a float64, in octave matrices with one element automatically get converted to scalars, 1.0 gets converted to int, etc.
I'd basically expect the same implicit casting as C++ compilers do?

gf712:
I guess that means every arithmetic type can be casted to any other arithmetic type?

karlnapf:
what happens in c if you do `int a = 3.4`?

gf712:
It depends. With -Wconversion you get a warning and I guess if you combined that with -Werror you get a compiler error. Without that warning the floating point type is floored, ie a=3

karlnapf:
I see....mmmh. Probably we would want to follow the behaviour that we have in our build, i.e. without those flags?

gf712:
Yes, I agree. I think that at least all integer types should be implicitly converted between them using the safe conversion helpers we already have. The same for floating points. Just from int to float I am not sure about...

karlnapf:
What I would do then is to have everything that doesnt loose information automatically. And then the ones which might loose information can simply print a runtime warning from shogun

PR #5056, opened 2020-06-04
Implicit conversions for put/get

gf712:
Implicit conversions as discussed in #5045. Would have to add all the primitive types, and probably come up with some better abstraction. This does the same as the Any::CastingRegistry, but by putting all this logic in SGObject we can add some very specific edge cases, like `std::variant`. I think this separates `Any` as a "type safe" type erasure abstraction from the parameters that use it.

### DIFF
--- a/tests/unit/base/SGObject_unittest.cc
+++ b/tests/unit/base/SGObject_unittest.cc
@@ -483,9 +483,10 @@ TEST(SGObject, tags_set_get_int)
 
 	EXPECT_THROW(obj->get<int32_t>("foo"), ShogunException);
 	obj->put(MockObject::kInt, 10);
+	EXPECT_NO_THROW(obj->put(MockObject::kInt, 10.0));
 	EXPECT_EQ(obj->get(Tag<int32_t>(MockObject::kInt)), 10);
-	EXPECT_THROW(obj->get<float64_t>(MockObject::kInt), ShogunException);
-	EXPECT_THROW(obj->get(Tag<float64_t>(MockObject::kInt)), ShogunException);
+	EXPECT_EQ(obj->get<float64_t>(MockObject::kInt), 10.0);
+	EXPECT_EQ(obj->get(Tag<float64_t>(MockObject::kInt)), 10.0);
 	EXPECT_EQ(obj->get<int>(MockObject::kInt), 10);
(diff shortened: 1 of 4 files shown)


tests/unit/base/SGObject_unittest.cc
```diff
@@ -483,9 +483,10 @@ TEST(SGObject, tags_set_get_int)
 
 	EXPECT_THROW(obj->get<int32_t>("foo"), ShogunException);
 	obj->put(MockObject::kInt, 10);
+	EXPECT_NO_THROW(obj->put(MockObject::kInt, 10.0));
 	EXPECT_EQ(obj->get(Tag<int32_t>(MockObject::kInt)), 10);
-	EXPECT_THROW(obj->get<float64_t>(MockObject::kInt), ShogunException);
-	EXPECT_THROW(obj->get(Tag<float64_t>(MockObject::kInt)), ShogunException);
+	EXPECT_EQ(obj->get<float64_t>(MockObject::kInt), 10.0);
+	EXPECT_EQ(obj->get(Tag<float64_t>(MockObject::kInt)), 10.0);
```
gf712, line 489: @karlnapf so this won't throw anymore

karlnapf: yep.
But everything that looses information throws?
maybe a test for some examples?

gf712: by lose information you mean something like 1.3 becomes 1? Right now I think this function only supports checks for overflow, e.g. int -10 is converted to unsigned int.

karlnapf: Yes, we wouldnt want that to happen for now I think.

gf712: @vigsterkr would it make sense to extend safe_convert to check floating point casting to integers and see if the delta is larger than fepsilon?

karlnapf: I always thought that the safe_convert functions would be for overflow checks, not for loss of information when going from discrete to floating?

karlnapf: (on the other hand, why not :) )

PR #5057, opened 2020-06-05
Remove Machine subclasses from SWIG

gf712:
Removed some more Machine subclasses from SWIG


karlnapf:
yes!
Swig compilation should get faster now slowly ... :)

gf712:
> yes!
> Swig compilation should get faster now slowly ... :)

Not by much :D still got 850k loc for the Python wrapper...

karlnapf:
I wonder where the majority is coming from these day. I once did an experiment an excluded all the classes but CMachine and it compiled in a few secs :)

gf712:
I can have a look tomorrow, just need to parse the swig generated code with clang and get the function names. Can probably figure out with that what is generating the most functions. I suspect it's all the feature and label template instantiations in SWIG

gf712:
@karlnapf I am actually removing a bunch of stuff in another branch and removed SGVector/Matrix and co, and that made a huge difference. We were exposing the whole thing, even though we just use the language native vector/matrix libraries (it made up about 15% of the SWIG code)

karlnapf:
This is strange, I had a go on removing all SG* stuff from swig quite some years ago and we shrinked the memory requirement to compile swig by more than a 1gig back then. I though it was all removed. But yeah totally remove all those things

PR #5010, opened 2020-04-19
[WIP] RFFGaussPreproc

jonpsy:
Addressing #5009

PR #5059, opened 2020-06-07
add inplace cos

jonpsy:
@gf712  @karlnapf . Inplace cos method for #5010

### DIFF
--- a/src/shogun/mathematics/linalg/LinalgNamespace.h
+++ b/src/shogun/mathematics/linalg/LinalgNamespace.h
@@ -1325,6 +1325,41 @@ namespace shogun
 			return result;
 		}
 
+		/** Performs the operation B = cos(A)
+		 *
+		 * This version returns the result in-place.
+		 * User should pass an appropriately pre-allocated memory matrix
+		 * 
+		 * @param a Input matrix
+		 * @param result Result matrix
+		 */
+		template<typename T>
+		void cos(const SGMatrix<T>& a, SGMatrix<T>& result)
+		{
+			require(
+			    a.num_rows == result.num_rows && a.num_cols == result.num_cols,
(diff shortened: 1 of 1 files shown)


gf712:
If it’s in place why do you have two arguments?

jonpsy:
I think I've used the wrong terminology. This method takes preallocated matrix as input, so it can "act as" a place in.
Edit: I think that's good because if it had only 1 param its definition will clash with the cos which returns container.

src/shogun/mathematics/linalg/LinalgNamespace.h
```diff
@@ -1325,6 +1325,41 @@ namespace shogun
 			return result;
 		}
 
+		/** Performs the operation B = cos(A)
+		 *
+		 * This version returns the result in-place.
```
karlnapf: remove this line

karlnapf (commented):
ehm, where is the code for the cos? This seems to be just the wrapper...

jonpsy:
Sorry, didn't catch that. This is just adding void cos(SGMatrix& a, SGMatrix& result) and similarly for SGVector. The backend for cos, sin and log is already implemented...

karlnapf:
But if it was already implemented, why do we need this additional wrapping code?

karlnapf:
I still dont understand: the backend already had the thing implemented but the interface had not?

jonpsy:
Precisely! LinalgNamespace.h uses ```infer_backend->cos``` all the interfaces. There lacked an interface for the following.
```c++
<template T>
void cos(SGMatrix<T>& A, SGMatrix<T>& result)
```
which I added.

jonpsy:
I've pushed the changes of LinalgNamespace in main PR for CI check just to confirm things.

PR #5064, opened 2020-06-12
More swig cleanup

gf712:
This reduces SWIG generated files for python by about 300k loc.. Down to ~620k!


src/interfaces/swig/Library.i
```diff
@@ -26,312 +23,53 @@
...
+SGCONTAINER_NOVALUE_WRAPPER(uint16_t)
+SGCONTAINER_NOVALUE_WRAPPER(uint32_t)
+SGCONTAINER_NOVALUE_WRAPPER(uint64_t)
+SGCONTAINER_NOVALUE_WRAPPER(int16_t)
+SGCONTAINER_NOVALUE_WRAPPER(int32_t)
+SGCONTAINER_NOVALUE_WRAPPER(int64_t)
+SGCONTAINER_NOVALUE_WRAPPER(float32_t)
+SGCONTAINER_NOVALUE_WRAPPER(float64_t)
+SGCONTAINER_NOVALUE_WRAPPER(floatmax_t)
+SGCONTAINER_NOVALUE_WRAPPER(std::complex<double>)
```
gf712, line 69: this removes sgcontainers from the interface but allows to still write the typemaps

src/interfaces/swig/Features.i
```diff
@@ -577,40 +458,7 @@ RANDOM_INTERFACE(RandomKitchenSinksDotFeatures)
...
-    %template(ULongIntMatrixFeatures)  MatrixFeatures<uint64_t>;
-#endif
-#ifdef USE_FLOATMAX
-    %template(LongRealMatrixFeatures) MatrixFeatures<floatmax_t>;
-#endif
-#ifdef USE_FLOAT32
-    %template(ShortRealMatrixFeatures) MatrixFeatures<float32_t>;
-#endif
 #ifdef USE_FLOAT64
     %template(RealMatrixFeatures) MatrixFeatures<float64_t>;
```
gf712, line 427: the only template that is actually used in the library

src/interfaces/swig/Features.i
```diff
@@ -496,48 +419,6 @@ namespace shogun
...
-#endif
-#ifdef USE_FLOATMAX
-    %template(LongRealSubsetFeatures) DenseSubsetFeatures<floatmax_t>;
-#endif
-#ifdef USE_FLOAT32
-    %template(ShortRealSubsetFeatures) DenseSubsetFeatures<float32_t>;
-#endif
-#ifdef USE_FLOAT64
-    %template(RealSubsetFeatures) DenseSubsetFeatures<float64_t>;
-#endif
```
gf712, line 538: I did not find any instance where we return this type

src/shogun/util/factory.h
```diff
@@ -164,6 +171,39 @@ namespace shogun
...
+			case PT_UINT64:
+			{
+				return std::make_shared<StringFeatures<uint64_t>>(
+				    file, alphabet_type);
+			}
+			default:
+				not_implemented(SOURCE_LOCATION);
+			}
+
+			return nullptr;
```
gf712, line 192: @karlnapf  not sure if it makes sense to add all these overloads?

examples/undocumented/python/distance_canberraword.py
```diff
@@ -10,20 +10,16 @@
 parameter_list = [[traindna,testdna,3,0,False],[traindna,testdna,3,0,False]]
 
 def distance_canberraword (fm_train_dna=traindna,fm_test_dna=testdna,order=3,gap=0,reverse=False):
-	from shogun import StringCharFeatures, StringWordFeatures, DNA
-
-	charfeat=StringCharFeatures(DNA)
-	charfeat.set_features(fm_train_dna)
-	feats_train=StringWordFeatures(charfeat.get_alphabet())
-	feats_train.obtain_from_char(charfeat, order-1, order, gap, reverse)
+	charfeat=sg.create_string_features(fm_train_dna, sg.DNA)
```
karlnapf, line 13: sg.dna -> "DNA" ?

karlnapf (commented):
> This reduces SWIG generated files for python by about 300k loc.. Down to ~620k!

this is great!

src/shogun/util/factory.h
karlnapf: you mean having all the possible instantiations? I think we generally should only ever have stuff that is actually used anywhere?

src/interfaces/swig/Features.i
karlnapf: yeah good. It is kind of strange that we instantiate all the templates even if not used ....
we should remove this one at some point as well, with a factory that has a similar signature as the ctor maybe?

src/shogun/util/factory.h
gf712: I mean it only makes sense to represent a string with unsigned integer types right?

examples/meta/generator/targets/java.json
```diff
@@ -6,7 +6,7 @@
         "IncludeGlobalFunctions": false,
         "DependencyListElement": "import org.shogun.$typeName;",
         "DependencyListSeparator": "\n",
-        "ExcludeImport": ["StringCharList", "StringWordList"]
+        "ExcludeImport": ["StringCharList", "StringWordList", "RealVector", "RealMatrix", "BoolMatrix", "ByteMatrix", "WordMatrix", "IntMatrix", "LongIntMatrix", "ShortRealMatrix", "BoolVector", "IntVector", "ByteVector", "LongIntVector", "ShortRealVector", "WordVector"]
```
gf712, line 9: I am not sure if this is the way to do it, but Java tries to import all of these types. However we don’t need them and I deleted them from SWIG (these were the sgcontainer templates)

examples/meta/generator/targets/java.json
```diff
@@ -107,14 +109,14 @@
...
+            "BoolVector": "$identifier.get($indices) == 1.0 ? true : false",
+            "BoolMatrix": "$identifier.get($indices) == 1.0 ? true : false"
         },
         "Assign": {
             "Vector": "$identifier.put($indices, $expr)",
             "Matrix": "$identifier.put($indices, $expr)",
-            "BoolVector": "$identifier.set_element($expr, $indices)",
-            "BoolMatrix": "$identifier.set_element($expr, $indices)"
+            "BoolVector": "$identifier.put($indices, $expr ? 1.0 : 0.0)",
+            "BoolMatrix": "$identifier.put($indices, $expr ? 1.0 : 0.0)"
```
gf712, line 119: This is also not great, but at least we consistently use DoubleMatrix, rather than mix with a shogun type

examples/undocumented/python/distance_canberraword.py
gf712: That might be tricky because this actually uses the constructor rather than put/get

src/interfaces/swig/Library.i
```diff
@@ -8,11 +8,8 @@
 #define PROTOCOLS_SGVECTOR(class_name, type_name, format_str, typecode)
 #endif
 
-%shared_ptr(shogun::Cache)
-%shared_ptr(shogun::ListElement)
 %shared_ptr(shogun::Signal)
 %shared_ptr(shogun::Time)
```
gf712: I think it’s OK to remove Time? I mean all languages have ways to measure process walltime and cpu time

src/interfaces/swig/Features.i
gf712: Yes, that should be doable, would have something like `create_features("MatrixFeatures", dims=(n_rows, n_cols), data=mat)`?

examples/meta/generator/targets/csharp.json
```diff
@@ -101,14 +103,14 @@
         "Access": {
             "Vector": "$identifier[$indices]",
             "Matrix": "$identifier[$indices]",
-            "BoolVector": "$identifier.get_element($indices)",
-            "BoolMatrix": "$identifier.get_element($indices)"
+            "BoolVector": "$identifier[$indices] == 1 ? true : false",
+            "BoolMatrix": "$identifier[$indices] == 1 ? true : false"
```
gf712, line 107: same as java, because boolean arrays are always a pain..

examples/undocumented/python/kernel_sparse_gaussian.py
```diff
@@ -7,11 +7,10 @@
 parameter_list = [[traindat,testdat,1.1],[traindat,testdat,1.2]]
 
 def kernel_sparse_gaussian (fm_train_real=traindat,fm_test_real=testdat,width=1.1 ):
-	from shogun import SparseRealFeatures
 	import shogun as sg
 
-	feats_train=SparseRealFeatures(fm_train_real)
-	feats_test=SparseRealFeatures(fm_test_real)
+	feats_train=sg.create_features(fm_train_real)
+	feats_test=sg.create_features(fm_test_real)
```
gf712: @karlnapf in this case we aren't loading the matrix as SparseFeatures, because it isn't a sparse scipy matrix... Should there be `sg.create_sparse_features`?

examples/undocumented/python/distance_canberraword.py
karlnapf: It needs to be though as the type won’t be available in the future...
But can be postponed I guess

examples/meta/generator/targets/java.json
karlnapf: I think I used the Sg type because of the strange api of double matrix. Would prefer if none of those were imported actually
If this works... good!

src/interfaces/swig/Library.i
karlnapf: Remove

examples/undocumented/python/kernel_sparse_gaussian.py
karlnapf: Think so yes. And it should be overloaded to take both a sparse matrix but also a dense matrix (for testing etc)

examples/meta/generator/targets/csharp.json
karlnapf: Think the matrix meta example might be a blocker... curious

examples/undocumented/python/kernel_sparse_gaussian.py
gf712: OK, makes sense. The issue I have now is that SGSparseMatrix is not part of the visitor pattern atm.. but working on it in another PR.

examples/meta/generator/targets/csharp.json
gf712: No, it works fine :)

examples/undocumented/python/distance_canberraword.py
gf712: Can have a dispatch in StringFeatures for that

src/shogun/features/Features.cpp
```diff
@@ -49,7 +49,7 @@ Features::~Features()
 
 void Features::init()
 {
-	set_default_mask(ParameterProperties::READONLY);
+	// set_default_mask(ParameterProperties::READONLY);
```
gf712, line 52: @karlnapf I am thinking if all parameters that are READONLY should be std::optional. This would mean the user can still set a value that is READONLY, if it is not set, and if it is set then it will throw. The current implementation only works when you can create the object via the constructor, so something like this won't work:
```python
hashed_feats=sg.create_features("HashedDocDotFeatures", num_bits=num_bits, 
                                doc_collection=f, tokenizer=tokenizer, 
                                should_normalize=normalize)
```
because it will call the default constructor and then try to put the parameters

karlnapf: So this will allow setting values once?
I think this wont be appropriate everywhere, though this is a case where it is.
I guess our "current solution" for this type of probem is offering different factories, but in this case here this doesnt really make sense.
I guess that means we kind of need two types. READONLY for parameters that can never be set from outside (like labels in Machine). And then things like the mask here (not sure what a good name would be). Many of this second type will need callbacks attached to them...

BTW there are also differences for serialization. READONLY would not necessarily be serialized I guess? Or do we actually need a third switch to turn off serialization for particular parameters?

examples/undocumented/python/distance_canberraword.py
karlnapf: couldnt we just force the user to set this with put rather than in the factory?

src/shogun/features/Features.cpp
gf712: >I guess that means we kind of need two types. READONLY for parameters that can never be set from outside (like labels in Machine). And then things like the mask here (not sure what a good name would be). Many of this second type will need callbacks attached to them...

In terms of name it could be something simple `WRITE_ONCE`. We wouldn't need a callback though, we just have the logic implemented like with readonly. And the std::optional would be handled similarly to AutoValue, ie. by the visitor. I assume you mean a callback to check if the value has been set?

>BTW there are also differences for serialization. READONLY would not necessarily be serialized I guess? Or do we actually need a third switch to turn off serialization for particular parameters?

That is a good q. I am not sure how it works rn... I guess it works because we don't serialise the features/labels which are the only parameters with this property?

karlnapf: ++ for WRITE_ONCE .... re the callback, I think for many cases we will need one (think some statistics are initialized once the value is written), but that is not a problem as we can actually do that

I think for serialization we are currently manually excluding a few parameter properties (like FUNCTION ... not sure about READONLY). Maybe it might be better to have another flag to explicitly disable, as if we rely on say READONLY and that keeps on changing its semantics, we might introduce bugs ...

gf712: > ++ for WRITE_ONCE .... re the callback, I think for many cases we will need one (think some statistics are initialized once the value is written), but that is not a problem as we can actually do that

Oh I see, I think callbacks are solvable yes, just need to make sure they are called after the value is set, which I think is already the case.

> I think for serialization we are currently manually excluding a few parameter properties (like FUNCTION ... not sure about READONLY). Maybe it might be better to have another flag to explicitly disable, as if we rely on say READONLY and that keeps on changing its semantics, we might introduce bugs ...

OK, I will have a look into this. I will fix this in another PR, and then when that is merged can fix this here.

vigsterkr: serialization only checks whether the value is visitable... if it is, it'll be serialized

karlnapf: ah yes thx, and functions are e.g. not visitable.
The thing is that we sometimes register things that *are* visit-able, but which shouldnt be serialized. This includes for example tables that are created in the ctor and might be exposed via READONLY.

gf712:
Odd, somehow managed to close this PR by accident. Down to 570k loc

src/interfaces/swig/Library.i
gf712: gone :)

examples/undocumented/python/kernel_combined.py
```diff
@@ -25,17 +24,17 @@ def kernel_combined (fm_train_real=traindat,fm_test_real=testdat,fm_train_dna=tr
...
 	feats_test.append_feature_obj(subkfeats_test)
 	kernel.add("kernel_array", subkernel)
 
-	subkfeats_train=StringCharFeatures(fm_train_dna, DNA)
-	subkfeats_test=StringCharFeatures(fm_test_dna, DNA)
-	subkernel=LocalAlignmentStringKernel(10)
+	subkfeats_train=sg.create_string_features(fm_train_dna, sg.DNA)
+	subkfeats_test=sg.create_string_features(fm_test_dna, sg.DNA)
+	subkernel=sg.create_kernel("LocalAlignmentStringKernel")
 	feats_train.append_feature_obj(subkfeats_train)
```
karlnapf: btw this should work via the new interface already via `add` ?

src/shogun/distributions/HMM.cpp
```diff
@@ -212,10 +212,11 @@ HMM::HMM(int32_t p_N, int32_t p_M, Model* p_model, float64_t p_PSEUDO)
 }
 
 HMM::HMM(
-	std::shared_ptr<StringFeatures<uint16_t>> obs, int32_t p_N, int32_t p_M,
+	const std::shared_ptr<Features>& obs, int32_t p_N, int32_t p_M,
 	float64_t p_PSEUDO)
 : RandomMixin<Distribution>(), iterations(150), epsilon(1e-4), conv_it(5)
 {
+	auto casted_obs = obs->as<StringFeatures<uint16_t>>();
```
karlnapf, line 219: easy :)

src/shogun/kernel/string/WeightedDegreeStringKernel.cpp
```diff
@@ -565,12 +560,11 @@ bool WeightedDegreeStringKernel::set_wd_weights_by_type(EWDKernType p_type)
...
 	ASSERT(p_type==E_WD) /// if we know a better weighting later on do a switch
 
-	SG_FREE(weights);
-	weights=SG_MALLOC(float64_t, degree);
+	weights=SGMatrix<float64_t>(degree, 1);
 	weights_degree=degree;
 	weights_length=1;
 
-	if (weights)
+	if (weights.size() > 0)
```
karlnapf, line 538: just wondering whether these files are covered by any sort of tests ... as bugs might slip through otherwise? (Even if unlikely)

gf712: there are no unittests.. It might be worth adding something that is run before these changes

examples/undocumented/python/kernel_combined.py
gf712: True, another class that can be removed :)

src/shogun/features/CombinedFeatures.cpp
```diff
@@ -162,6 +162,16 @@ void CombinedFeatures::init()
...
+	add_callback_function("feature_array", [this](){
+		int32_t n=feature_array.back()->get_num_vectors();
+		if (get_num_vectors()>0 && n!=get_num_vectors())
+		{
+			feature_array.pop_back();
+			error("Number of feature vectors does not match (expected {}, "
+					"obj has {})", get_num_vectors(), n);
+		}
+		num_vec=n;
+	});
```
gf712, line 174: @karlnapf these callbacks are becoming really awkward. The issue is if something hasn't been set then we need to reset to the old state. In this case this is possible because I can pop the last element of the vector (which would be invalid here). The problem is that callbacks are only triggered after updating the parameter. What do you think about adding a customisable constraint where you can just pass lambda that checks the new parameter value before replacing the old parameter value?

src/shogun/kernel/string/WeightedDegreeStringKernel.cpp
karlnapf: think integration tests are fine for that... but then if we port the string kernel ones you will get many conflicts in here....any idea how to best do this? we could involve the gsoc students maybe?

gf712: Hmm I think the easiest is if I just run the code in the develop branch and copy paste the results into a unittest?

gf712: The issue is that the interface cannot be accessed from outside? Oh but we don't need it I guess. Yes, @LiuYuHui or @Hephaestus12 could write a meta example for this then. Shouldn't be difficult for me to rebase. It would just be good to do that soon so I can merge this as it will reduce the interface compile time by quite a bit

karlnapf: I personally think unit tests should be hand-designed to cover corner cases or reference libraries and not be created from outputs, this is what our integration meta tests are for

gf712: > I personally think unit tests should be hand-designed to cover corner cases or reference libraries and not be created from outputs, this is what our integration meta tests are for

yes agreed. For some reason I thought we couldn't write a meta example with this kernel. But we can right? Just cannot use factories

src/shogun/features/CombinedFeatures.cpp
karlnapf: this is a design flaw, having the check in here. Algorithms should check this when training imo.

having said that, there should definitely be checks that are done *before* changing, just like the constraints. Can't you have a custom constraint here btw?

src/shogun/kernel/string/WeightedDegreeStringKernel.cpp
karlnapf: Yeah. The GSoC students could prioritize the meta examples to port being the string kernel ones. And then they have to work with what API is there (no factories yet). @LiuYuHui @Hephaestus12 let us know if you have any questions, it would be great if you both could port a few string kernel examples next

src/shogun/features/CombinedFeatures.cpp
gf712: Well, you can but you need to create a new struct which is quite verbose, so I was thinking I could add a customisable constraint where you just have a lambda.
```cpp
SG_ADD(&array, "feature_array", SG_CONSTRAINT([this](const auto& new_element){
  if (get_num_vectors()>0 && new_element->get_num_vectors()!=get_num_vectors())
  {
	error("Number of feature vectors does not match (expected {}, obj has {})", get_num_vectors(), new_element->get_num_vectors());
  }
}))
```

I am not sure why the check there is flawed though? Would you not want to know about an issue as soon as possible, ie. before training?

gf712: And then that would have to also have a callback to update `num_vec` (we could have that in the constraint but then we are mixing functionality)

src/shogun/kernel/string/WeightedDegreeStringKernel.cpp
gf712: the examples that would be highest priority would be `kernel_weighted_degree_string.py` and `kernel_weighted_degree_position_string.py`. Those are the ones I had to refactor quite a bit

src/shogun/features/CombinedFeatures.cpp
karlnapf: num_vec should just be removed and always be queried from the first feature (if exists), and then everying can be matched against that.

src/shogun/kernel/string/WeightedDegreeStringKernel.cpp
LiuYuHui: how to handle this? https://github.com/shogun-toolbox/shogun/blob/55bb4a991f9336c78225545c9ddb530ca31e73cc/examples/undocumented/python/kernel_weighted_degree_string.py#L19-L22

gf712: you could write the array manually, have a look at the matrix_types meta example. It's not ideal but should work
https://github.com/shogun-toolbox/shogun/blob/55bb4a991f9336c78225545c9ddb530ca31e73cc/examples/meta/src/meta_api/matrix_types.sg.in#L56-L62

Hephaestus12: How do you replace this line?
just `create_features()` doesn't seem to work.

https://github.com/shogun-toolbox/shogun/blob/41b6247f0a8c0f9ce45b388ae085d42d90bdd16d/examples/undocumented/python/kernel_linear_string.py#L14-L15

gf712: There is a create_string_features function with that same signature

gf712:
@karlnapf shall I merge this?

karlnapf:
absolutey! :rocket:

Issue #5054, opened 2020-06-04
Label assertion and mapping in Machine


gf712:
Currently some classification algorithms check whether the input Labels are valid, e.g. the class labels are continuous `[0, 1, ..., n_classes-1]`, which leads to a lot of duplicate code.
These checks should be done by the Machine base class when training is performed. The Machine will then store the mapping of any Label input to an internal encoding, e.g. a binary classification task would map {10,20} -> {-1,+1} using a `BinaryLabelEncoder` class, and similarly there would be a `MulticlassLabelsEncoder` class for multiclass tasks. The properly encoded Labels are then dispatched to the `train_machine` method. When `apply` is called the returned Labels are mapped back to the user input Labels space using the `LabelEncoder`.

The tasks (in order):
- [x] write a `LabelEncoder` base class and respective `BinaryLabelEncoder` and `MulticlassLabelsEncoder` derived classes. These should also check that the Labels are valid, e.g. cannot transform {-1, 0, 1} to BinaryLabels. #5067 
- [ ] add `LabelEncoder` as a `Machine` class member
- [ ] fit the `LabelEncoder` and transform input in `train` and then perform inverse operation in `apply`
- [ ] Remove label checks from `Machine` subclasses, since algorithms are now guaranteed to receive a valid Label representation
- [ ] xvalidation would use its own mapping that it passes on to each fold's `Machine` in order to keep the same mapping across folds

Most of this code already exists, but it is spread around the code base

karlnapf:
a lot of the conversion code is inside the labels classes already, so can be re-used.
E.g. [here](https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/labels/MulticlassLabels.cpp#L210) and [here](https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/labels/MulticlassLabels.cpp#L244)

Also note that some of this code is already used within the old approach, where algorithm classes convert the labels to the appropriate form (rather than the base class doing it as outlined above). See e.g. [here](https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/classifier/LDA.cpp#L87). This would just be removed with the approach described above as the algorithms are guaranteed to receive the appropriate labels.
Finally, this old approach currently in use might cause bugs/wrong results when used within xvalidation as the mappings (might) change across folds....

PR #5067, opened 2020-06-15
Add label encoder

LiuYuHui:
Related to #5054 
Add LabelEncoder/MulticlassLabelsEncoder/BinaryLabelEncoder

### DIFF
--- a/src/shogun/labels/BinaryLabelEncoder.h
+++ b/src/shogun/labels/BinaryLabelEncoder.h
@@ -16,7 +16,9 @@
 #include <unordered_set>
 namespace shogun
 {
-
+	/** @brief Implements a reversible mapping from
+	 * any form of labels to binary labels (+1, -1).
+	 */
 	class BinaryLabelEncoder : public LabelEncoder
 	{
 	public:
@@ -24,29 +26,22 @@ namespace shogun
 
 		~BinaryLabelEncoder() = default;
 
-		/** Fit label encoder
-		 *
-		 * @param Target values.
-		 * @return SGVector which contains unique labels.
-		 */
 		SGVector<float64_t> fit(const std::shared_ptr<Labels>& labs) override
 		{
 			const auto result_vector = labs->as<DenseLabels>()->get_labels();
 			check_is_valid(result_vector);
 			if (!can_convert_float_to_int(result_vector))
 			{
+				std::set<float64_t> s(
+				    result_vector.begin(), result_vector.end());
 				io::warn(
-				    "({}, {}) have been converted to (-1, 1).",
-				    result_vector[0], result_vector[1]);
+				    "({}, {}) have been converted to (-1, 1).", *s.begin(),
+				    *s.end());
 			}
 			auto result_labels = fit_impl(result_vector);
 			return result_labels;
 		}
-		/** Transform labels to normalized encoding.
-		 *
-		 * @param  Target values to be transformed.
-		 * @return Labels transformed to be normalized.
-		 */
+
 		std::shared_ptr<Labels>
 		transform(const std::shared_ptr<Labels>& labs) override
 		{
@@ -57,34 +52,29 @@ namespace shogun
 			std::transform(
 			    transformed_vec.begin(), transformed_vec.end(),
 			    transformed_vec.begin(), [](float64_t e) {
-				    if (std::abs(e - 0.0) <=
-				        std::numeric_limits<float64_t>::epsilon())
-					    return -1.0;
-				    else
-					    return e;
+				    return Math::fequals(
+				               e, 0.0,
+				               std::numeric_limits<float64_t>::epsilon())
+				               ? -1.0
+				               : e;
 			    });
 			return std::make_shared<BinaryLabels>(transformed_vec);
 		}
-		/** Transform labels back to original encoding.
-		 *
-		 * @param normailzed encoding labels
-		 * @return original encoding labels
-		 */
+
 		std::shared_ptr<Labels>
 		inverse_transform(const std::shared_ptr<Labels>& labs) override
 		{
 			auto normalized_labels = labs->as<BinaryLabels>();
 			normalized_labels->ensure_valid();
 			auto normalized_vector = normalized_labels->get_labels();
-			check_is_valid(normalized_vector);
 			std::transform(
 			    normalized_vector.begin(), normalized_vector.end(),
 			    normalized_vector.begin(), [](float64_t e) {
-				    if (std::abs(e + 1.0) <=
-				        std::numeric_limits<float64_t>::epsilon())
-					    return 0.0;
-				    else
-					    return e;
+				    return Math::fequals(
+				               e, -1.0,
+				               std::numeric_limits<float64_t>::epsilon())
+				               ? 0.0
+				               : e;
 			    });
 			auto origin_vec = inverse_transform_impl(normalized_vector);
 			SGVector<int32_t> result_vev(origin_vec.vlen);
@@ -93,11 +83,7 @@ namespace shogun
 			    [](auto&& e) { return static_cast<int32_t>(e); });
 			return std::make_shared<BinaryLabels>(result_vev);
 		}
-		/** Fit label encoder and return encoded labels.
-		 *
-		 * @param Target values.
-		 * @return Labels transformed to be normalized.
-		 */
+
 		std::shared_ptr<Labels>
 		fit_transform(const std::shared_ptr<Labels>& labs) override
 		{
@@ -118,11 +104,10 @@ namespace shogun
 			    std::unordered_set<float64_t>(vec.begin(), vec.end());
 			require(
 			    unique_set.size() == 2,
-			    "Binary labels should contain only two elements, can not interpret ({}) as binary labels",
+			    "Cannot interpret ({}) as binary labels, need exactly two "
+			    "classes.",
 			    fmt::join(unique_set, ", "));
 		}
-
-		
 	};
 } // namespace shogun
 
--- a/src/shogun/labels/BinaryLabelEncoder.h
+++ b/src/shogun/labels/BinaryLabelEncoder.h
@@ -32,11 +32,14 @@ namespace shogun
 		SGVector<float64_t> fit(const std::shared_ptr<Labels>& labs) override
 		{
 			const auto result_vector = labs->as<DenseLabels>()->get_labels();
+			check_is_valid(result_vector);
+			if (is_convert_float_to_int(result_vector))
+			{
+				io::warn(
+				    "({}, {}) have been converted to (-1, 1).",
+				    result_vector[0], result_vector[1]);
+			}
 			auto result_labels = fit_impl(result_vector);
-			require(
-			    unique_labels.size() == 2,
-			    "Binary Labels should contain only two elements");
-
 			return result_labels;
 		}
 		/** Transform labels to normalized encoding.
@@ -48,11 +51,7 @@ namespace shogun
 		transform(const std::shared_ptr<Labels>& labs) override
 		{
 			const auto result_vector = labs->as<DenseLabels>()->get_labels();
-			require(
-			    std::unordered_set<float64_t>(
-			        result_vector.begin(), result_vector.end())
-			            .size() == 2,
-			    "Binary Labels should contain only two elements");
+			check_is_valid(result_vector);
 			auto transformed_vec = transform_impl(result_vector);
 
 			std::transform(
@@ -77,12 +76,7 @@ namespace shogun
 			auto normalized_labels = labs->as<BinaryLabels>();
 			normalized_labels->ensure_valid();
 			auto normalized_vector = normalized_labels->get_labels();
-			require(
-			    std::unordered_set<float64_t>(
-			        normalized_vector.begin(), normalized_vector.end())
-			            .size() == 2,
-			    "Binary Labels should contain only two elements");
-
+			check_is_valid(normalized_vector);
 			std::transform(
 			    normalized_vector.begin(), normalized_vector.end(),
 			    normalized_vector.begin(), [](float64_t e) {
@@ -115,6 +109,32 @@ namespace shogun
 		virtual const char* get_name() const
 		{
 			return "BinaryLabelEncoder";
+		}
+
+	private:
+		void check_is_valid(const SGVector<float64_t>& vec)
+		{
+			const auto unique_set =
+			    std::unordered_set<float64_t>(vec.begin(), vec.end());
+			require(
+			    unique_set.size() == 2,
+			    "Binary labels should contain only two elements, ({}) have "
+			    "been detected.",
+			    fmt::join(unique_set, ", "));
+		}
+
+		bool is_convert_float_to_int(const SGVector<float64_t>& vec) const
+		{
+			SGVector<int32_t> converted(vec.vlen);
+			std::transform(
+			    vec.begin(), vec.end(), converted.begin(),
+			    [](auto&& e) { return static_cast<int32_t>(e); });
+			return std::equal(
+			    vec.begin(), vec.end(), converted.begin(),
+			    [](auto&& e1, auto&& e2) {
+				    return std::abs(e1 - e2) >
+				           std::numeric_limits<float64_t>::epsilon();
+			    });
 		}
 	};
 } // namespace shogun
--- a/src/shogun/labels/BinaryLabelEncoder.h
+++ b/src/shogun/labels/BinaryLabelEncoder.h
@@ -33,7 +33,7 @@ namespace shogun
 		{
 			const auto result_vector = labs->as<DenseLabels>()->get_labels();
 			check_is_valid(result_vector);
-			if (!can_convert_float_to_int(result_vector))
+			if (print_warning && !can_convert_float_to_int(result_vector))
 			{
 				io::warn(
 				    "({}, {}) have been converted to (-1, 1).",
@@ -122,19 +122,7 @@ namespace shogun
 			    fmt::join(unique_set, ", "));
 		}
 
-		bool can_convert_float_to_int(const SGVector<float64_t>& vec) const
-		{
-			SGVector<int32_t> converted(vec.vlen);
-			std::transform(
-			    vec.begin(), vec.end(), converted.begin(),
-			    [](auto&& e) { return static_cast<int32_t>(e); });
-			return std::equal(
-			    vec.begin(), vec.end(), converted.begin(),
-			    [](auto&& e1, auto&& e2) {
-				    return std::abs(e1 - e2) <
-				           std::numeric_limits<float64_t>::epsilon();
-			    });
-		}
+		
 	};
 } // namespace shogun
 
--- a/src/shogun/labels/BinaryLabelEncoder.h
+++ b/src/shogun/labels/BinaryLabelEncoder.h
@@ -33,7 +33,7 @@ namespace shogun
 		{
 			const auto result_vector = labs->as<DenseLabels>()->get_labels();
 			check_is_valid(result_vector);
-			if (is_convert_float_to_int(result_vector))
+			if (!can_convert_float_to_int(result_vector))
 			{
 				io::warn(
 				    "({}, {}) have been converted to (-1, 1).",
@@ -118,12 +118,11 @@ namespace shogun
 			    std::unordered_set<float64_t>(vec.begin(), vec.end());
 			require(
 			    unique_set.size() == 2,
-			    "Binary labels should contain only two elements, ({}) have "
-			    "been detected.",
+			    "Binary labels should contain only two elements, can not interpret ({}) as binary labels",
 			    fmt::join(unique_set, ", "));
 		}
 
-		bool is_convert_float_to_int(const SGVector<float64_t>& vec) const
+		bool can_convert_float_to_int(const SGVector<float64_t>& vec) const
 		{
 			SGVector<int32_t> converted(vec.vlen);
 			std::transform(
@@ -132,7 +131,7 @@ namespace shogun
 			return std::equal(
 			    vec.begin(), vec.end(), converted.begin(),
 			    [](auto&& e1, auto&& e2) {
-				    return std::abs(e1 - e2) >
+				    return std::abs(e1 - e2) <
 				           std::numeric_limits<float64_t>::epsilon();
 			    });
 		}
--- a/src/shogun/labels/BinaryLabelEncoder.h
+++ b/src/shogun/labels/BinaryLabelEncoder.h
@@ -38,8 +38,7 @@ namespace shogun
 				    "({}, {}) have been converted to (-1, 1).", *s.begin(),
 				    *s.end());
 			}
-			auto result_labels = fit_impl(result_vector);
-			return result_labels;
+			return fit_impl(result_vector);
 		}
 
 		std::shared_ptr<Labels>
--- a/src/shogun/labels/BinaryLabelEncoder.h
+++ b/src/shogun/labels/BinaryLabelEncoder.h
@@ -33,7 +33,7 @@ namespace shogun
 		{
 			const auto result_vector = labs->as<DenseLabels>()->get_labels();
 			check_is_valid(result_vector);
-			if (print_warning && !can_convert_float_to_int(result_vector))
+			if (!can_convert_float_to_int(result_vector))
 			{
 				io::warn(
 				    "({}, {}) have been converted to (-1, 1).",
--- a/src/shogun/labels/BinaryLabelEncoder.h
+++ b/src/shogun/labels/BinaryLabelEncoder.h
@@ -13,7 +13,7 @@
 #include <shogun/labels/DenseLabels.h>
 #include <shogun/labels/LabelEncoder.h>
 #include <shogun/lib/SGVector.h>
-
+#include <unordered_set>
 namespace shogun
 {
 
@@ -32,13 +32,12 @@ namespace shogun
 		SGVector<float64_t> fit(const std::shared_ptr<Labels>& labs) override
 		{
 			const auto result_vector = labs->as<DenseLabels>()->get_labels();
-			fit_impl(result_vector);
+			auto result_labels = fit_impl(result_vector);
 			require(
 			    unique_labels.size() == 2,
-			    "BinaryLabel should contain only two elements");
+			    "Binary Labels should contain only two elements");
 
-			return SGVector<float64_t>(
-			    unique_labels.begin(), unique_labels.end());
+			return result_labels;
 		}
 		/** Transform labels to normalized encoding.
 		 *
@@ -50,9 +49,10 @@ namespace shogun
 		{
 			const auto result_vector = labs->as<DenseLabels>()->get_labels();
 			require(
-			    std::set<float64_t>(result_vector.begin(), result_vector.end())
+			    std::unordered_set<float64_t>(
+			        result_vector.begin(), result_vector.end())
 			            .size() == 2,
-			    "BinaryLabel should contain only two elements");
+			    "Binary Labels should contain only two elements");
 			auto transformed_vec = transform_impl(result_vector);
 
 			std::transform(
@@ -74,7 +74,14 @@ namespace shogun
 		std::shared_ptr<Labels>
 		inverse_transform(const std::shared_ptr<Labels>& labs) override
 		{
-			auto normalized_vector = labs->as<DenseLabels>()->get_labels();
+			auto normalized_labels = labs->as<BinaryLabels>();
+			normalized_labels->ensure_valid();
+			auto normalized_vector = normalized_labels->get_labels();
+			require(
+			    std::unordered_set<float64_t>(
+			        normalized_vector.begin(), normalized_vector.end())
+			            .size() == 2,
+			    "Binary Labels should contain only two elements");
 
 			std::transform(
 			    normalized_vector.begin(), normalized_vector.end(),
@@ -85,14 +92,12 @@ namespace shogun
 				    else
 					    return e;
 			    });
-			require(
-			    std::set<float64_t>(
-			        normalized_vector.begin(), normalized_vector.end())
-			            .size() == 2,
-			    "BinaryLabel should contain only two elements");
-
-			return std::make_shared<BinaryLabels>(
-			    inverse_transform_impl(normalized_vector));
+			auto origin_vec = inverse_transform_impl(normalized_vector);
+			SGVector<int32_t> result_vev(origin_vec.vlen);
+			std::transform(
+			    origin_vec.begin(), origin_vec.end(), result_vev.begin(),
+			    [](auto&& e) { return static_cast<int32_t>(e); });
+			return std::make_shared<BinaryLabels>(result_vev);
 		}
 		/** Fit label encoder and return encoded labels.
 		 *
--- a/src/shogun/labels/BinaryLabelEncoder.h
+++ b/src/shogun/labels/BinaryLabelEncoder.h
@@ -1,0 +1,117 @@
+/*
+ * This software is distributed under BSD 3-clause license (see LICENSE file).
+ *
+ * Authors: Yuhui Liu
+ *
+ */
+#ifndef _BINARYLABELENCODER__H__
+#define _BINARYLABELENCODER__H__
+
+#include <memory>
+#include <shogun/base/SGObject.h>
+#include <shogun/labels/BinaryLabels.h>
+#include <shogun/labels/DenseLabels.h>
+#include <shogun/labels/LabelEncoder.h>
+#include <shogun/lib/SGVector.h>
+
+namespace shogun
+{
+
+	class BinaryLabelEncoder : public LabelEncoder
+	{
+	public:
+		BinaryLabelEncoder() = default;
+
+		~BinaryLabelEncoder() = default;
+
+		/** Fit label encoder
+		 *
+		 * @param Target values.
+		 * @return SGVector which contains unique labels.
+		 */
+		SGVector<float64_t> fit(const std::shared_ptr<Labels>& labs) override
+		{
+			const auto result_vector = labs->as<DenseLabels>()->get_labels();
+			fit_impl(result_vector);
+			require(
+			    unique_labels.size() == 2,
+			    "BinaryLabel should contain only two elements");
+
+			return SGVector<float64_t>(
+			    unique_labels.begin(), unique_labels.end());
+		}
+		/** Transform labels to normalized encoding.
+		 *
+		 * @param  Target values to be transformed.
+		 * @return Labels transformed to be normalized.
+		 */
+		std::shared_ptr<Labels>
+		transform(const std::shared_ptr<Labels>& labs) override
+		{
+			const auto result_vector = labs->as<DenseLabels>()->get_labels();
+			require(
+			    std::set<float64_t>(result_vector.begin(), result_vector.end())
+			            .size() == 2,
+			    "BinaryLabel should contain only two elements");
+			auto transformed_vec = transform_impl(result_vector);
+
+			std::transform(
+			    transformed_vec.begin(), transformed_vec.end(),
+			    transformed_vec.begin(), [](float64_t e) {
+				    if (std::abs(e - 0.0) <=
+				        std::numeric_limits<float64_t>::epsilon())
+					    return -1.0;
+				    else
+					    return e;
+			    });
+			return std::make_shared<BinaryLabels>(transformed_vec);
+		}
+		/** Transform labels back to original encoding.
+		 *
+		 * @param normailzed encoding labels
+		 * @return original encoding labels
+		 */
+		std::shared_ptr<Labels>
+		inverse_transform(const std::shared_ptr<Labels>& labs) override
+		{
+			auto normalized_vector = labs->as<DenseLabels>()->get_labels();
+
+			std::transform(
+			    normalized_vector.begin(), normalized_vector.end(),
+			    normalized_vector.begin(), [](float64_t e) {
+				    if (std::abs(e + 1.0) <=
+				        std::numeric_limits<float64_t>::epsilon())
+					    return 0.0;
+				    else
+					    return e;
+			    });
+			require(
+			    std::set<float64_t>(
+			        normalized_vector.begin(), normalized_vector.end())
+			            .size() == 2,
+			    "BinaryLabel should contain only two elements");
+
+			return std::make_shared<BinaryLabels>(
+			    inverse_transform_impl(normalized_vector));
+		}
+		/** Fit label encoder and return encoded labels.
+		 *
+		 * @param Target values.
+		 * @return Labels transformed to be normalized.
+		 */
+		std::shared_ptr<Labels>
+		fit_transform(const std::shared_ptr<Labels>& labs) override
+		{
+			const auto result_vector = labs->as<DenseLabels>()->get_labels();
+			return std::make_shared<BinaryLabels>(
+			    transform_impl(fit_impl(result_vector)));
+		}
+
+		virtual const char* get_name() const
+		{
+			return "BinaryLabelEncoder";
+		}
+	};
+} // namespace shogun
+
+#endif
--- a/src/shogun/labels/LabelEncoder.h
+++ b/src/shogun/labels/LabelEncoder.h
@@ -1,0 +1,106 @@
+/*
+ * This software is distributed under BSD 3-clause license (see LICENSE file).
+ *
+ * Authors: Yuhui Liu
+ *
+ */
+
+#ifndef _LABELENCODER__H__
+#define _LABELENCODER__H__
+
+#include <algorithm>
+#include <map>
+#include <memory>
+#include <set>
+#include <shogun/base/SGObject.h>
+#include <shogun/lib/SGVector.h>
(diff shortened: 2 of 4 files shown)


src/shogun/labels/BinaryLabelEncoder.h
```diff
@@ -0,0 +1,117 @@
...
+
+		/** Fit label encoder
+		 *
+		 * @param Target values.
+		 * @return SGVector which contains unique labels.
+		 */
+		SGVector<float64_t> fit(const std::shared_ptr<Labels>& labs) override
+		{
+			const auto result_vector = labs->as<DenseLabels>()->get_labels();
+			fit_impl(result_vector);
```
gf712: I don't think it will work having a single `fit_impl`. I think BinaryLabels require {-1,+1}, but I could have labels {0,10}. So I would expect that mapping to happen in the BinaryLabelEncoder? Otherwise I will just get {0,10} here, right?

gf712: ah that is done in transform, but then how do you keep the original mapping? What if I fit to a set that starts with label 0 and then test set first label is 10? Would I still get the same mapping?

src/shogun/labels/BinaryLabelEncoder.h
```diff
@@ -0,0 +1,117 @@
...
+			    std::set<float64_t>(result_vector.begin(), result_vector.end())
+			            .size() == 2,
+			    "BinaryLabel should contain only two elements");
+			auto transformed_vec = transform_impl(result_vector);
+
+			std::transform(
+			    transformed_vec.begin(), transformed_vec.end(),
+			    transformed_vec.begin(), [](float64_t e) {
+				    if (std::abs(e - 0.0) <=
+				        std::numeric_limits<float64_t>::epsilon())
```
gf712: `Math::fequals` does this

src/shogun/labels/BinaryLabelEncoder.h
```diff
@@ -0,0 +1,117 @@
...
+					    return -1.0;
+				    else
+					    return e;
```
gf712: Here you assume that the labels are either 0 or 1 right?

src/shogun/labels/BinaryLabelEncoder.h
```diff
@@ -0,0 +1,117 @@
...
+		SGVector<float64_t> fit(const std::shared_ptr<Labels>& labs) override
+		{
+			const auto result_vector = labs->as<DenseLabels>()->get_labels();
+			fit_impl(result_vector);
+			require(
+			    unique_labels.size() == 2,
+			    "BinaryLabel should contain only two elements");
+
+			return SGVector<float64_t>(
+			    unique_labels.begin(), unique_labels.end());
```
gf712: isn't this the result of `fit_impl`?

tests/unit/labels/LabelsEncoder_unittest.cc
```diff
@@ -0,0 +1,66 @@
...
+#include <shogun/labels/BinaryLabelEncoder.h>
+#include <shogun/labels/LabelEncoder.h>
+#include <shogun/labels/MulticlassLabels.h>
+#include <shogun/labels/MulticlassLabelsEncoder.h>
+using namespace shogun;
+
+TEST(BinaryLabelEncoder, fit_transform)
+{
+	auto label_encoder = std::make_shared<BinaryLabelEncoder>();
+	SGVector<int32_t> vec{-1, -1, 1, -1, 1};
```
gf712, line 17: what about something with values that are not {-1,1}. Could you also test for something with more than two labels to make sure the exception is thrown?

tests/unit/labels/LabelsEncoder_unittest.cc
```diff
@@ -0,0 +1,66 @@
...
+	for (int i = 0; i < 5; i++)
+	{
+		EXPECT_EQ(vec[i], inv_result[i]);
+	}
+}
+
+TEST(MulticlassLabelsEncoder, fit_transform)
+{
+	auto label_encoder = std::make_shared<MulticlassLabelsEncoder>();
+	SGVector<float64_t> vec{1, 2, 2, 6};
```
gf712: I am assuming something with negative values and 0 would also work?

src/shogun/labels/BinaryLabelEncoder.h
LiuYuHui: i assume that the labels of fit should have the same categories as the labels of transform.

tests/unit/labels/LabelsEncoder_unittest.cc
```diff
@@ -0,0 +1,66 @@
...
+	auto result_labels = label_encoder->transform(origin_labels);
+	auto result_vec = result_labels->as<MulticlassLabels>()->get_labels();
+	EXPECT_EQ(0, result_vec[0]);
+	EXPECT_EQ(1, result_vec[1]);
+	EXPECT_EQ(1, result_vec[2]);
+	EXPECT_EQ(2, result_vec[3]);
+
+	auto inv_result = label_encoder->inverse_transform(result_labels)
+	                      ->as<MulticlassLabels>()
+	                      ->get_labels();
```
gf712, line 122: could you also test for something that isn't the result? For example something like `{2,0,1,0}`?

gf712: Also what happens when you transform with something that wasn't fitted? For example if now there was a labels 3, but your labels space is {1,2,6}?

gf712 (commented):
Good start! I have some questions. But I think I just misunderstood the design. It would be good to add more tests and think of all possible corner cases

src/shogun/labels/BinaryLabelEncoder.h
gf712: Yes, that makes sense I think. I was just a bit confused at the start :D

src/shogun/labels/BinaryLabelEncoder.h
LiuYuHui: The labels have been transformed by `auto transformed_vec = transform_impl(result_vector);`, so the labels should only contain 0 or 1 now.

gf712: Oh I see, so if you have something like {0,10} -> {0,1} -> {-1,+1}? Have you checked everything works fine with negative labels as well?

tests/unit/labels/LabelsEncoder_unittest.cc
LiuYuHui: if values are not {-1, 1}, they will be transformed to {-1, 1}, i will add more unite test.

src/shogun/labels/BinaryLabelEncoder.h
LiuYuHui: yes, it exactly is what I did, for the situation with negative labels, i will add more test.

src/shogun/labels/BinaryLabelEncoder.h
vigsterkr: plz dont change it to Math::  ;) we wanna kill that thing sooner or later

gf712: hmm I guess we need to replace this with a utility function somewhere then? Because this is quite a common operation

src/shogun/labels/LabelEncoder.h
```diff
@@ -0,0 +1,106 @@
...
+		}
+
+		SGVector<float64_t>
+		inverse_transform_impl(const SGVector<float64_t>& result_vector)
+		{
+			SGVector<float64_t> original_vector(result_vector.vlen);
+			std::transform(
+			    result_vector.begin(), result_vector.end(),
+			    original_vector.begin(),
+			    [& normalized_to_origin = normalized_to_origin](const auto& e) {
```
gf712: btw if you don't rename a variable you can just write it as `[&normalized_to_origin]`

LiuYuHui: normalized_to_origin is a class member, so if I want to use normalized_to_origin in lambda, I guess I have to write `[& normalized_to_origin = normalized_to_origin]`?

src/shogun/labels/BinaryLabelEncoder.h
```diff
@@ -0,0 +1,117 @@
...
+		 * @return Labels transformed to be normalized.
+		 */
+		std::shared_ptr<Labels>
+		transform(const std::shared_ptr<Labels>& labs) override
+		{
+			const auto result_vector = labs->as<DenseLabels>()->get_labels();
+			require(
+			    std::set<float64_t>(result_vector.begin(), result_vector.end())
+			            .size() == 2,
+			    "BinaryLabel should contain only two elements");
```
karlnapf: could you pls re-use the code we have for these transformations? I put some links in the issue.
Or at least then delete the old code

karlnapf: "BinaryLabel" -> "Binary labels".
And also please print what the values are so that the user can see what is wrong. See also the existing code for that

src/shogun/labels/BinaryLabelEncoder.h
```diff
@@ -0,0 +1,117 @@
...
+				        std::numeric_limits<float64_t>::epsilon())
+					    return 0.0;
+				    else
+					    return e;
+			    });
+			require(
+			    std::set<float64_t>(
+			        normalized_vector.begin(), normalized_vector.end())
+			            .size() == 2,
+			    "BinaryLabel should contain only two elements");
```
karlnapf: See comment above

src/shogun/labels/BinaryLabelEncoder.h
```diff
@@ -0,0 +1,122 @@
...
+		 * @param Target values.
+		 * @return SGVector which contains unique labels.
+		 */
+		SGVector<float64_t> fit(const std::shared_ptr<Labels>& labs) override
+		{
+			const auto result_vector = labs->as<DenseLabels>()->get_labels();
+			auto result_labels = fit_impl(result_vector);
+			require(
+			    unique_labels.size() == 2,
+			    "Binary Labels should contain only two elements");
```
karlnapf: nitpick: "Binary labels" (the idea is to not replicate class/variable/type names in user facing error messages).
Still would be good to print the unique labels in this error message

src/shogun/labels/BinaryLabelEncoder.h
karlnapf: could be an issue, and then while we are at it can also replace all the CMath::fequals calls

src/shogun/labels/BinaryLabelEncoder.h
```diff
@@ -0,0 +1,122 @@
...
+		inverse_transform(const std::shared_ptr<Labels>& labs) override
+		{
+			auto normalized_labels = labs->as<BinaryLabels>();
+			normalized_labels->ensure_valid();
+			auto normalized_vector = normalized_labels->get_labels();
+			require(
+			    std::unordered_set<float64_t>(
+			        normalized_vector.begin(), normalized_vector.end())
+			            .size() == 2,
+			    "Binary Labels should contain only two elements");
```
karlnapf: could we maybe put this string somewhere so it is not replicated so many times.... as well as the check code itself?

tests/unit/labels/LabelsEncoder_unittest.cc
karlnapf: I guess you want to test

* binary labels are passed (no-op)
* labels with two unique values are passed (and transformed into binary)
* labels with fewer or more than two unique values are passed (error)

tests/unit/labels/LabelsEncoder_unittest.cc
```diff
@@ -0,0 +1,169 @@
...
+	{
+		EXPECT_EQ(vec[i], inv_result[i]);
+	}
+
+	SGVector<int32_t> test_vec{-1, -1, -1, -1, -1, 1};
+	auto test_labels = std::make_shared<BinaryLabels>(test_vec);
+	auto inv_test = label_encoder->inverse_transform(test_labels)
+	                    ->as<BinaryLabels>()
+	                    ->get_labels();
+	EXPECT_EQ(-100, inv_test[0]);
```
karlnapf: could you use a loop (or macro) and the original vector so make this less verbose?

karlnapf (commented):
BTW I think there should be a warning printed if a user passes labels that are not integers. E.g. {2.13434, 3.42342343} shouldnt be mapped to {-1,+1} without warning

src/shogun/labels/BinaryLabelEncoder.h
LiuYuHui: but the amount of the unique labels may be exceed more than two, such as if `{0,1,2,3,4,5,6,7,8}` is passed to the method, should we print all values?

vigsterkr: yeah the idea is that the user gets a feedback what is the the problem with his input. i.e. you print out that we've detected `{0,1,2}` as unique labels in his data set... this should help one to figure out/fix the input...

karlnapf: ++ what viktor says here.
Also keep in mind that the unique number of labels will be relatively small (unless someone passes regression labels .... so maybe you could cap them at length 10 or so and make a "...", although I don't really think that is necessary)

LiuYuHui:
> BTW I think there should be a warning printed if a user passes labels that are not integers. E.g. {2.13434, 3.42342343} shouldnt be mapped to {-1,+1} without warning

but it seems like the `Labels` class use `SGVector<float64_t>` to store the values, even the values are `SGVector<int32_t>`, so i think there is no way to figure out what the values are, because those values have been converted to `float64_t`.

gf712:
> > BTW I think there should be a warning printed if a user passes labels that are not integers. E.g. {2.13434, 3.42342343} shouldnt be mapped to {-1,+1} without warning
> 
> but it seems like the `Labels` class use `SGVector<float64_t>` to store the values, even the values are `SGVector<int32_t>`, so i think there is no way to figure out what the values are, because those values have been converted to `float64_t`.

But you can cast the elements to int and then compare the difference between element and integer cast right? If that difference is larger than epsilon then there is an issue, i.e. it does not represent an integer. You could do this when you populate the set. It will slow down the encoding fitting massively, but this shouldn't be the performance critical part anyway

src/shogun/labels/BinaryLabelEncoder.h
```diff
@@ -0,0 +1,142 @@
...
+	private:
+		void check_is_valid(const SGVector<float64_t>& vec)
+		{
+			const auto unique_set =
+			    std::unordered_set<float64_t>(vec.begin(), vec.end());
+			require(
+			    unique_set.size() == 2,
+			    "Binary labels should contain only two elements, ({}) have "
+			    "been detected.",
+			    fmt::join(unique_set, ", "));
```
gf712, line 108: neat! Didn't know this was a thing :D

src/shogun/labels/BinaryLabelEncoder.h
```diff
@@ -0,0 +1,142 @@
...
+			SGVector<int32_t> converted(vec.vlen);
+			std::transform(
+			    vec.begin(), vec.end(), converted.begin(),
+			    [](auto&& e) { return static_cast<int32_t>(e); });
+			return std::equal(
+			    vec.begin(), vec.end(), converted.begin(),
+			    [](auto&& e1, auto&& e2) {
+				    return std::abs(e1 - e2) >
+				           std::numeric_limits<float64_t>::epsilon();
+			    });
```
gf712: I think you need to rename this function, it is a bit confusing what you get back. Imo it should be `can_convert_float_to_int` and then you just have to invert the logic to `std::abs(e1 - e2) < std::numeric_limits<float64_t>::epsilon();`, and then fix the logic in each call

tests/unit/labels/LabelsEncoder_unittest.cc
```diff
@@ -0,0 +1,170 @@
...
+	{
+		EXPECT_EQ(expected_inv_test[i], inv_test[i]);
+	}
+}
+
+TEST(BinaryLabelEncoder, more_than_two_labels)
+{
+	auto label_encoder = std::make_shared<BinaryLabelEncoder>();
+	SGVector<int32_t> vec{-100, 200, -100, 200, -100, 42};
+	auto origin_labels = std::make_shared<BinaryLabels>(vec);
```
gf712, line 84: @karlnapf doesn't BinaryLabels throw in this situation? And if not shouldn't it?
Also shouldn't this test be done with MulticlassLabels?

src/shogun/labels/LabelEncoder.h
```diff
@@ -0,0 +1,106 @@
...
+#include <algorithm>
+#include <map>
+#include <memory>
+#include <set>
+#include <shogun/base/SGObject.h>
+#include <shogun/lib/SGVector.h>
+namespace shogun
+{
+
+	class LabelEncoder : public SGObject
```
gf712, line 23: Should this inherit from `Transformer`?

LiuYuHui: It seems like `Transformer` transform `Features`, not `Labels`.
https://github.com/shogun-toolbox/shogun/blob/41b6247f0a8c0f9ce45b388ae085d42d90bdd16d/src/shogun/transformer/Transformer.h#L60-L62

gf712: Ah makes sense, nvm!

src/shogun/labels/BinaryLabelEncoder.h
karlnapf: nit: "Cannot interpret {} as binary labels"

tests/unit/labels/LabelsEncoder_unittest.cc
karlnapf: I think it doesnt atm, but I am not sure.

src/shogun/labels/LabelEncoder.h
gf712: ah yes, you're right!

tests/unit/labels/LabelsEncoder_unittest.cc
```diff
@@ -0,0 +1,170 @@
...
+	}
+}
+
+TEST(MulticlassLabelsEncoder, negative_labels)
+{
+	auto eps = std::numeric_limits<float64_t>::epsilon();
+	auto label_encoder = std::make_shared<MulticlassLabelsEncoder>();
+	SGVector<float64_t> vec{-100.1, 200.4, -2.868, 6.98, -2.868};
+	auto origin_labels = std::make_shared<MulticlassLabels>(vec);
+	auto unique_vec = label_encoder->fit(origin_labels);
```
gf712, line 136: so will this give the user a warning? Because they are using non-integer representations?

karlnapf: I think it should (though maybe it should be possible to de-activate it? not sure if this can be spamy)

gf712: I assume this would only happen once when you fit thought no? And the user can also change the io level.

src/shogun/labels/LabelEncoder.h
```diff
@@ -0,0 +1,125 @@
...
+		 */
+		virtual std::shared_ptr<Labels>
+		fit_transform(const std::shared_ptr<Labels>&) = 0;
+
+		virtual const char* get_name() const
+		{
+			return "LabelEncoder";
+		}
+
+		void set_print_warning(bool print_warning){
```
gf712: Imo this shouldn’t be needed. The user should set this with log level, eg switch from warn to error

vigsterkr: ++

src/shogun/labels/BinaryLabelEncoder.h
```diff
@@ -0,0 +1,129 @@
...
+		/** Fit label encoder
+		 *
+		 * @param Target values.
+		 * @return SGVector which contains unique labels.
+		 */
+		SGVector<float64_t> fit(const std::shared_ptr<Labels>& labs) override
+		{
+			const auto result_vector = labs->as<DenseLabels>()->get_labels();
+			check_is_valid(result_vector);
+			if (print_warning && !can_convert_float_to_int(result_vector))
```
vigsterkr:
```suggestion
			if (!can_convert_float_to_int(result_vector))
```
spdlog and log level will take care of this

src/shogun/labels/BinaryLabelEncoder.h
```diff
@@ -0,0 +1,129 @@
...
+			{
+				io::warn(
+				    "({}, {}) have been converted to (-1, 1).",
+				    result_vector[0], result_vector[1]);
+			}
+			auto result_labels = fit_impl(result_vector);
+			return result_labels;
```
vigsterkr:
```suggestion
			return fit_impl(result_vector);
```

src/shogun/labels/LabelEncoder.h
```diff
@@ -0,0 +1,125 @@
...
+			    [](auto&& e) { return static_cast<int32_t>(e); });
+			return std::equal(
+			    vec.begin(), vec.end(), converted.begin(),
+			    [](auto&& e1, auto&& e2) {
+				    return std::abs(e1 - e2) <
+				           std::numeric_limits<float64_t>::epsilon();
+			    });
+		}
+
+		bool print_warning = true;
```
vigsterkr: drop it

src/shogun/labels/MulticlassLabelsEncoder.h
```diff
@@ -0,0 +1,89 @@
...
+
+		/** Fit label encoder
+		 *
+		 * @param Target values.
+		 * @return SGVector which contains unique labels.
+		 */
+		SGVector<float64_t> fit(const std::shared_ptr<Labels>& labs) override
+		{
+			const auto result_vector = labs->as<DenseLabels>()->get_labels();
+			if (print_warning && !can_convert_float_to_int(result_vector))
```
vigsterkr:
```suggestion
			if (!can_convert_float_to_int(result_vector))
```

vigsterkr (changes requested):

tests/unit/labels/LabelsEncoder_unittest.cc
karlnapf: in grid search it would happen quite a few times. But yes with the loglevel that is best taken care of

src/shogun/labels/BinaryLabelEncoder.h
```diff
@@ -0,0 +1,129 @@
...
+#include <shogun/base/SGObject.h>
+#include <shogun/labels/BinaryLabels.h>
+#include <shogun/labels/DenseLabels.h>
+#include <shogun/labels/LabelEncoder.h>
+#include <shogun/lib/SGVector.h>
+#include <unordered_set>
+namespace shogun
+{
+
+	class BinaryLabelEncoder : public LabelEncoder
```
karlnapf, line 22: Class doc: "Implements a reversible mapping from any form of labels to binary labels (+1, -1)."

src/shogun/labels/BinaryLabelEncoder.h
```diff
@@ -0,0 +1,129 @@
...
+	{
+	public:
+		BinaryLabelEncoder() = default;
+
+		~BinaryLabelEncoder() = default;
+
+		/** Fit label encoder
```
karlnapf: Could we move this docstring to the base class and just leave it empty here (as it contains nothing relevant for this subclass)

src/shogun/labels/BinaryLabelEncoder.h
```diff
@@ -0,0 +1,129 @@
...
+		 * @return SGVector which contains unique labels.
+		 */
+		SGVector<float64_t> fit(const std::shared_ptr<Labels>& labs) override
+		{
+			const auto result_vector = labs->as<DenseLabels>()->get_labels();
+			check_is_valid(result_vector);
+			if (!can_convert_float_to_int(result_vector))
+			{
+				io::warn(
+				    "({}, {}) have been converted to (-1, 1).",
```
karlnapf: is that always true? Wouldnt you want to print the actual conversion? I.e. is result_vector[0] always mapped to -1?

src/shogun/labels/BinaryLabelEncoder.h
```diff
@@ -0,0 +1,129 @@
...
+				    result_vector[0], result_vector[1]);
+			}
+			auto result_labels = fit_impl(result_vector);
+			return result_labels;
+		}
+		/** Transform labels to normalized encoding.
```
karlnapf: same here, move to base class and remove here

src/shogun/labels/BinaryLabelEncoder.h
karlnapf: I think we should either introduce this utility function now, or just use CMath::fequals and include this in a larger refactor (it is copy paste)

src/shogun/labels/BinaryLabelEncoder.h
```diff
@@ -0,0 +1,129 @@
...
+			    transformed_vec.begin(), [](float64_t e) {
+				    if (std::abs(e - 0.0) <=
+				        std::numeric_limits<float64_t>::epsilon())
+					    return -1.0;
+				    else
+					    return e;
+			    });
+			return std::make_shared<BinaryLabels>(transformed_vec);
+		}
+		/** Transform labels back to original encoding.
```
karlnapf: base class

src/shogun/labels/BinaryLabelEncoder.h
```diff
@@ -0,0 +1,129 @@
...
+		 *
+		 * @param normailzed encoding labels
+		 * @return original encoding labels
+		 */
+		std::shared_ptr<Labels>
+		inverse_transform(const std::shared_ptr<Labels>& labs) override
+		{
+			auto normalized_labels = labs->as<BinaryLabels>();
+			normalized_labels->ensure_valid();
```
karlnapf, line 67: what does this do? The same as the `check_is_valid`below? I think it might require that at least two labels exist, so not sure that is appropriate here?

src/shogun/labels/BinaryLabelEncoder.h
```diff
@@ -0,0 +1,129 @@
...
+			auto normalized_vector = normalized_labels->get_labels();
+			check_is_valid(normalized_vector);
+			std::transform(
+			    normalized_vector.begin(), normalized_vector.end(),
+			    normalized_vector.begin(), [](float64_t e) {
+				    if (std::abs(e + 1.0) <=
```
karlnapf: `Math::fequals` or utility func

src/shogun/labels/BinaryLabelEncoder.h
```diff
@@ -0,0 +1,129 @@
...
+					    return e;
+			    });
+			auto origin_vec = inverse_transform_impl(normalized_vector);
+			SGVector<int32_t> result_vev(origin_vec.vlen);
+			std::transform(
+			    origin_vec.begin(), origin_vec.end(), result_vev.begin(),
+			    [](auto&& e) { return static_cast<int32_t>(e); });
+			return std::make_shared<BinaryLabels>(result_vev);
+		}
+		/** Fit label encoder and return encoded labels.
```
karlnapf: base class

src/shogun/labels/BinaryLabelEncoder.h
```diff
@@ -0,0 +1,129 @@
...
+		}
+
+	private:
+		void check_is_valid(const SGVector<float64_t>& vec)
+		{
+			const auto unique_set =
+			    std::unordered_set<float64_t>(vec.begin(), vec.end());
+			require(
+			    unique_set.size() == 2,
+			    "Binary labels should contain only two elements, can not interpret ({}) as binary labels",
```
karlnapf:
```suggestion
			    "Cannot interpret ({}) as binary labels, need exactly two classes.",
```

src/shogun/labels/BinaryLabelEncoder.h
```diff
@@ -0,0 +1,129 @@
...
+			    [](auto&& e) { return static_cast<int32_t>(e); });
+			return std::make_shared<BinaryLabels>(result_vev);
+		}
+		/** Fit label encoder and return encoded labels.
+		 *
+		 * @param Target values.
+		 * @return Labels transformed to be normalized.
+		 */
+		std::shared_ptr<Labels>
+		fit_transform(const std::shared_ptr<Labels>& labs) override
```
karlnapf, line 87: Can't this be implemented in the base class?

src/shogun/labels/BinaryLabelEncoder.h
```diff
@@ -0,0 +1,129 @@
...
+			return "BinaryLabelEncoder";
+		}
+
+	private:
+		void check_is_valid(const SGVector<float64_t>& vec)
+		{
+			const auto unique_set =
+			    std::unordered_set<float64_t>(vec.begin(), vec.end());
+			require(
+			    unique_set.size() == 2,
```
karlnapf, line 105: @gf712 I think it might not be good to assert this as sometimes labels might only contain one class. But I guess this will pop up if a problem and we can change it then :)

src/shogun/labels/BinaryLabelEncoder.h
karlnapf: Actually udpate. Simply use docstring of base class

src/shogun/labels/MulticlassLabelsEncoder.h
```diff
@@ -0,0 +1,89 @@
...
+			return std::make_shared<MulticlassLabels>(
+			    transform_impl(result_vector));
+		}
+		/** Transform labels back to original encoding.
+		 *
+		 * @param normailzed encoding labels
+		 * @return original encoding labels
+		 */
+		std::shared_ptr<Labels>
+		inverse_transform(const std::shared_ptr<Labels>& labs) override
```
karlnapf, line 52: don't we also need a check valid here? Something that ensures that the labels are contiguous? [0,1,2,3,4, ... ] no gaps.
I wrote some code for this in the links I posted. Either re-use or remove my code :)

src/shogun/labels/MulticlassLabelsEncoder.h
```diff
@@ -0,0 +1,89 @@
...
+#include <shogun/base/SGObject.h>
+#include <shogun/labels/DenseLabels.h>
+#include <shogun/labels/LabelEncoder.h>
+#include <shogun/labels/MulticlassLabels.h>
+#include <shogun/lib/SGVector.h>
+
+namespace shogun
+{
+
+	class MulticlassLabelsEncoder : public LabelEncoder
```
karlnapf, line 22: same comments as for binary labels class

src/shogun/labels/LabelEncoder.h
karlnapf: "Implements a reversible mapping from any form of labels to one of Shogun's target label spaces (binary, multi-class, etc)."

src/shogun/labels/LabelEncoder.h
```diff
@@ -0,0 +1,121 @@
...
+		bool can_convert_float_to_int(const SGVector<float64_t>& vec) const
+		{
+			SGVector<int32_t> converted(vec.vlen);
+			std::transform(
+			    vec.begin(), vec.end(), converted.begin(),
+			    [](auto&& e) { return static_cast<int32_t>(e); });
+			return std::equal(
+			    vec.begin(), vec.end(), converted.begin(),
+			    [](auto&& e1, auto&& e2) {
+				    return std::abs(e1 - e2) <
```
karlnapf: CMath::fequals or utility pls :)

src/shogun/labels/LabelEncoder.h
```diff
@@ -0,0 +1,121 @@
...
+				           std::numeric_limits<float64_t>::epsilon();
+			    });
+		}
+
+		std::set<float64_t> unique_labels;
+		std::unordered_map<float64_t, float64_t> normalized_to_origin;
```
karlnapf: `inverse_mapping`

karlnapf (commented):
Made some more comments. Nice work :)

src/shogun/labels/BinaryLabelEncoder.h
LiuYuHui: the type of return value is different, one is `shared_ptr<BinaryLabels>`, the other one is `shared_ptr<MulticlassLabels>`.
``` cpp
std::shared_ptr<Labels> BinaryLabelEncoder::fit_transform(const std::shared_ptr<Labels>& labs) override
{
    const auto result_vector = labs->as<DenseLabels>()->get_labels();
    return std::make_shared<BinaryLabels>(
	          transform_impl(fit_impl(result_vector)));
}
```
```cpp
std::shared_ptr<Labels> MulticlassLabelsEncoder::fit_transform(const std::shared_ptr<Labels>& labs) override
{
    const auto result_vector = labs->as<DenseLabels>()->get_labels();
    return std::make_shared<MulticlassLabels>(
	          transform_impl(fit_impl(result_vector)));
}

File: src/shogun/labels/BinaryLabelEncoder.h (hunk 29 above)
Comment (LiuYuHui) [reply src/shogun/labels/BinaryLabelEncoder.h:67]: yes, I should remove `check_is_valid` and only use `ensure_valid` here, `ensure_valid` have more strict constraint as it ensures labels have only contained {-1, +1}.

File: src/shogun/labels/MulticlassLabelsEncoder.h (hunk 35 above)
Comment (LiuYuHui) [reply src/shogun/labels/MulticlassLabelsEncoder.h:52]: I am thinking whether we need a check valid here, as inverse_transform is to map from internal encoding to origin encoding. for example, {100, 100, 200, 300} -> {0, 0, 1, 2}, {0, 0, 1, 2} are transformed by internal encoding, but it is not continuous

Comment (karlnapf) [reply src/shogun/labels/MulticlassLabelsEncoder.h:52]: ah you are right of course :)

File: src/shogun/labels/BinaryLabelEncoder.h [hunk 39]
```diff
@@ -16,37 +16,33 @@
 #include <unordered_set>
 namespace shogun
 {
-
+	/** @brief 	Implements a reversible mapping from
```
karlnapf: whitespaces are weird here. Could you clean up?

src/shogun/labels/BinaryLabelEncoder.h
karlnapf: typo can not -> cannot

src/shogun/labels/LabelEncoder.h
```diff
@@ -16,7 +16,11 @@
 #include <shogun/lib/SGVector.h>
 namespace shogun
 {
-
+	/** @brief Implements a reversible mapping from any
+	 * form of labels to one of Shogun's target label spaces
+	 * (binary, multi-class, etc).
+	 *
```
karlnapf: nit: remove this line

karlnapf (commented):
Cool, looks good. What is missing?

LiuYuHui:
> Cool, looks good. What is missing?

i think all things have been done.

src/shogun/labels/LabelEncoder.h
```diff
@@ -0,0 +1,124 @@
...
+			std::transform(
+			    result_vector.begin(), result_vector.end(),
+			    original_vector.begin(),
+			    [& inverse_mapping = inverse_mapping](const auto& e) {
+				    return inverse_mapping[e];
+			    });
+			return original_vector;
+		}
+
+		bool can_convert_float_to_int(const SGVector<float64_t>& vec) const
```
karlnapf, line 132: I wonder whether this should be templated (for both float and int type) and live somewhere where other conversion tools (safe_convert) live...this might be useful elesewhere
@gf712 thoughts?

src/shogun/labels/MulticlassLabelsEncoder.h
```diff
@@ -0,0 +1,75 @@
...
+#include <memory>
+#include <shogun/base/SGObject.h>
+#include <shogun/labels/DenseLabels.h>
+#include <shogun/labels/LabelEncoder.h>
+#include <shogun/labels/MulticlassLabels.h>
+#include <shogun/lib/SGVector.h>
+
+namespace shogun
+{
+	/** @brief 	Implements a reversible mapping from
```
karlnapf: some whitespace glitches

karlnapf (commented):
Just wondering: would it make sense to avoid doing any conversion if the labels are already in the correct format? (i.e. -1,1 or 0,1,2,3,4 ....,num_classes-1)? Or is that a pointless optimization as not making a difference anyways?

LiuYuHui:
> Just wondering: would it make sense to avoid doing any conversion if the labels are already in the correct format? (i.e. -1,1 or 0,1,2,3,4 ....,num_classes-1)? Or is that a pointless optimization as not making a difference anyways?

I think we should not add this optimization, when {-1, 1} or {0, 1, 2, num_classes-1} are passed in `fit/transform`, we still need to maintain the mapping, as we don't know what values will be passed in `inverse_transform`.

karlnapf:
> > Just wondering: would it make sense to avoid doing any conversion if the labels are already in the correct format? (i.e. -1,1 or 0,1,2,3,4 ....,num_classes-1)? Or is that a pointless optimization as not making a difference anyways?
> 
> I think we should not add this optimization, when {-1, 1} or {0, 1, 2, num_classes-1} are passed in `fit/transform`, we still need to maintain the mapping, as we don't know what values will be passed in `inverse_transform`.

Well that could be dealt with ... say if no inverse mapping was computed, the labels are not mapped backwards, but simply used as returned by the machine. @gf712 @vigsterkr what are your thoughts?

karlnapf:
I think otherwise, we can merge this. Any objections (apart from the question above)?

gf712:
> > > Just wondering: would it make sense to avoid doing any conversion if the labels are already in the correct format? (i.e. -1,1 or 0,1,2,3,4 ....,num_classes-1)? Or is that a pointless optimization as not making a difference anyways?
> > 
> > 
> > I think we should not add this optimization, when {-1, 1} or {0, 1, 2, num_classes-1} are passed in `fit/transform`, we still need to maintain the mapping, as we don't know what values will be passed in `inverse_transform`.
> 
> Well that could be dealt with ... say if no inverse mapping was computed, the labels are not mapped backwards, but simply used as returned by the machine. @gf712 @vigsterkr what are your thoughts?

Yes, makes sense to me to have this, but then you need to check if the encoder has been fitted (boolean class member) and then you can see if the map is empty -> if empty noop

src/shogun/labels/BinaryLabelEncoder.h
gf712: hmm, in what situation would there only be one label?

karlnapf: When predicting, although I am not sure this is ever called in that case.

src/shogun/labels/LabelEncoder.h
gf712: Yes, ideally this would be templated

src/shogun/labels/BinaryLabelEncoder.h
gf712: seems like this is only called in fit and transform, so should be fine

src/shogun/labels/LabelEncoder.h
gf712: @LiuYuHui could you do this change so that when `CMath::fequals` is replaced we can just find and replace these things?

src/shogun/labels/BinaryLabelEncoder.h
karlnapf: yep + 1 and if it becomes a  problem we just change it

karlnapf:
BTW once the above discussion is resolved, we could merge this.
But there is a follow up thing to do (could go into a single PR)
* embed this into Machine
* remove the old conversion methods
* all machine implementations can then perform dynamic casts and assume the correct label type is given to them

And then another PR for using all this within xvalidation

src/shogun/labels/LabelEncoder.h
```diff
@@ -0,0 +1,146 @@
...
+			return std::equal(
+			    vec.begin(), vec.end(), converted.begin(),
+			    [&](auto&& e1, auto&& e2) {
+				    return Math::fequals(e1, static_cast<float64_t>(e2), eps);
+			    });
+		}
+
+		std::set<float64_t> unique_labels;
+		std::unordered_map<float64_t, float64_t> inverse_mapping;
+		const float64_t eps = std::numeric_limits<float64_t>::epsilon();
```
gf712:
```suggestion
		constexpr float64_t eps = std::numeric_limits<float64_t>::epsilon();
```

LiuYuHui: when I changed `const` to `constexpr`, I got an error:  `non-static data member ‘eps’ declared ‘constexpr’`

gf712: ah yes, it should be `static constexpr float64_t eps = std::numeric_limits<float64_t>::epsilon();`

gf712:
I think this can merged now :)

karlnapf:
Great! :)

As a next step, can I suggest a PR that
* uses this class inside Machine::train
* removes all the conversion code in BinaryLabels and MulticlassLabels
* replaces all usages of `binary_labels` and `multiclass_labels` conversion methods in `train_machine` methods with `->as<BinaryLabels>()` etc? I.e. a simple dynamic cast
* Fixes all the examples/tests that will break through that :)

Issue #5003, opened 2020-04-11
-DENABLE_TESTING=OFF leads to ADD_CUSTOM_COMMAND error


vinnik-dmitry07:
```
CMake Error at CMakeLists.txt:561 (ADD_CUSTOM_COMMAND):
  ADD_CUSTOM_COMMAND Wrong syntax.  A TARGET or OUTPUT must be specified.
```

```
IF((ENABLE_TESTING AND NOT TRAVIS_DISABLE_UNIT_TESTS AND EXISTS ${CMAKE_SOURCE_DIR}/tests/unit)
	OR (EXISTS ${CMAKE_SOURCE_DIR}/examples AND BUILD_META_EXAMPLES))
	# set CTAGS_FILE when tests/unit or examples/meta are added to the project
	# according to the logic below
	SET(CTAGS_FILE ${CMAKE_CURRENT_BINARY_DIR}/tags CACHE INTERNAL "" FORCE)
ENDIF()

IF(ENABLE_TESTING AND NOT BUILD_DASHBOARD_REPORTS)
	enable_testing()
ENDIF()


FIND_PACKAGE(Ctags)
# TODO: the arg for -R should rather be the where libshogun's headers are residing
ADD_CUSTOM_COMMAND(OUTPUT ${CTAGS_FILE} # <- CTAGS_FILE is empty here
    COMMAND ${CTAGS_EXECUTABLE} -f ${CTAGS_FILE}
    # classes, enums, functions
    --c++-kinds=cfgp
    --fields=+im
    --exclude=*.cpp
    --exclude=third_party
    --exclude=GoogleMock
    --exclude=rxcpp
    --exclude=interfaces
    --exclude=Eigen3
    --exclude=external
    --languages=c++
    -R ${CMAKE_SOURCE_DIR})
```

karlnapf:
I cannot reproduce this locally, could you post the exact cmake command and cmake version?

vinnik-dmitry07:
It happens when
```
-DBUILD_META_EXAMPLES=OFF 
-DENABLE_TESTING=OFF
```
are together.

karlnapf:
doesnt happen here

vinnik-dmitry07:
```
cmake -DCMAKE_BUILD_TYPE=Debug -DENABLE_TESTING=OFF -DBUILD_META_EXAMPLES=OFF -G "CodeBlocks - Unix Makefiles" /home/arkan0id/CLionProjects/shogun
```
```
[...]

PR #5068, opened 2020-06-16
ctags detection should be only called when needed

vigsterkr:
but then it's required

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -555,24 +555,33 @@ IF(ENABLE_TESTING AND NOT BUILD_DASHBOARD_REPORTS)
 	enable_testing()
 ENDIF()
 
-FIND_PACKAGE(Ctags)
-# TODO: the arg for -R should rather be the where libshogun's headers are residing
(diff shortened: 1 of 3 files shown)


vigsterkr:
@gf712 right?

CMakeLists.txt
```diff
@@ -555,24 +555,26 @@ IF(ENABLE_TESTING AND NOT BUILD_DASHBOARD_REPORTS)
...
-    --exclude=rxcpp
-    --exclude=interfaces
-    --exclude=Eigen3
-    --exclude=external
-    --languages=c++
-    -R ${CMAKE_SOURCE_DIR})
-
-ADD_CUSTOM_TARGET(ctags DEPENDS ${CTAGS_FILE})
+IF (DEFINED CTAGS_FILE)
+    FIND_PACKAGE(Ctags REQUIRED)
```
gf712: why required? if it doesn't find it then the command will not run right?

Hephaestus12:
Is it possible to set this CTAGS_FILE manually while running cmake?

vigsterkr:
this should fix #5003

Issue #5047, opened 2020-05-28
Chain SGObject::put


gf712:
I was wondering if it would make sense for `SGObject::put` to return `this` by reference? That way we could chain various put calls:
```cpp
auto k = create<Kernel>("GaussianKernel")
         .put("log_width, 1.)
         .put("cache_size", 200);
```

karlnapf:
yep I think that would be nice

mr3543:
hey I was having a look at this issue and thought I might be able to help. I was under the impression `create<Kernel>("GaussianKernel")` would return a `std::shared_ptr<Kernel>` so chaining the put calls would have to look something like 

`auto k = create<Kernel>("GaussianKernel")
->put("log_width",1)
->put("cache_size",200);`

and `put` would have to return a `shared_ptr`. I just started looking at this repo though so I could be off here.

gf712:
Yup, that’s correct :) so you would probably have to return something like shared_from_this() in put

mr3543:
cool, thanks!

mr3543:
I tried making the change but there are a few test cases (358,432,443,470,476) that make objects without shared pointers and call `put`. These throw a `bad_weak_ptr` exception, see https://stackoverflow.com/questions/27697973/shared-from-this-causing-bad-weak-ptr. 

I can try to change the instantiation in these test cases, but wasn't sure if that's something the team is ok with.

gf712:
Hmm, try to get it to run and send a PR and then we can have a look :)

theartful:
@gf712 
We can achieve chaining like effect in dynamic languages by using kwargs.
So
``` python
kernel = create_kernel("GaussianKernel")
kernel.put(log_width=1, cache_size=200)
```
we can even modify python's put to return the same python object, so we can do
```
kernel = create_kernel("GaussianKernel").put(log_width=1, cache_size=200)
```

For C++ we can accept a list of pairs using template parameter packs
```C++
kernel.put({"log_width", 1}, {"cache_size", 200})
```
but it's not as attractive.

jonpsy:
Smells like #5139. I'll add this to my list. Although, I'm a bit worried about its implications in SWIG interfaces.

PR #5069, opened 2020-06-17
return shared_from_this from SGObject::put

mr3543:
changed `put` in `SGObject.h` to return `shared_ptr<SGObject>`. see issue [#5407](https://github.com/shogun-toolbox/shogun/issues/5047)


gf712:
I think it’s safe to assume that SGObjects are always on the heap and managed by a shared_ptr. Imo those tests that are failing should reflect this. @vigsterkr ?

vigsterkr:
@gf712 mmmmmm this is a back and forth story :) for a long time that was actually the assumption...

PR #5072, opened 2020-06-20
Refactor gaussian process machine

LiuYuHui:


tests/unit/regression/GaussianProcessRegression_unittest.cc
```diff
@@ -191,23 +191,24 @@ TEST(GaussianProcessRegression, apply_regression_on_training_features)
...
 	// Gaussian likelihood with sigma = 0.25
-	auto liklihood=std::make_shared<GaussianLikelihood>(0.25);
+	std::shared_ptr<LikelihoodModel> liklihood=std::make_shared<GaussianLikelihood>(0.25);
 
 	// specify GP regression with exact inference
-	auto inf=std::make_shared<ExactInferenceMethod>(kernel, features_train,
-			mean, labels_train, liklihood);
-
+	auto inf=std::make_shared<ExactInferenceMethod>();
+	inf->put("kernel", kernel);
```
karlnapf: could you use either string constants here (defined in the class you are putting to), or use c++ setters?

src/shogun/regression/GaussianProcessRegression.cpp
```diff
@@ -39,35 +39,9 @@ std::shared_ptr<RegressionLabels> GaussianProcessRegression::apply_regression(st
...
-
-	}
-	else
-	{
-		result=std::make_shared<RegressionLabels>(get_mean_vector(data));
-		if (m_compute_variance)
-			result->put("current_values", get_variance_vector(data));
-	}
-
+	result=std::make_shared<RegressionLabels>(get_mean_vector(data));
```
gf712: minor thing, you now have the declaration and definition side by side, you can just put all in the same line

src/shogun/regression/GaussianProcessRegression.cpp
```diff
@@ -39,35 +39,9 @@ std::shared_ptr<RegressionLabels> GaussianProcessRegression::apply_regression(st
...
+	if (m_compute_variance)
+		result->put("current_values", get_variance_vector(data));
```
gf712: can this be `result->set_values(get_variance_vector(data))`?

tests/unit/regression/GaussianProcessRegression_unittest.cc
```diff
@@ -191,23 +191,24 @@ TEST(GaussianProcessRegression, apply_regression_on_training_features)
 	auto labels_train=std::make_shared<RegressionLabels>(lab_train);
 
 	// choose Gaussian kernel with sigma = 0.02 and zero mean function
-	auto kernel=std::make_shared<GaussianKernel>(10, 0.02);
-	auto mean=std::make_shared<ZeroMean>();
+	std::shared_ptr<Kernel> kernel=std::make_shared<GaussianKernel>(10, 0.02);
+	std::shared_ptr<MeanFunction> mean=std::make_shared<ZeroMean>();
```
gf712: why not `auto`? I guess it didn't work with put?

src/shogun/classifier/GaussianProcessClassification.cpp
```diff
@@ -71,18 +71,6 @@ std::shared_ptr<MulticlassLabels> GaussianProcessClassification::apply_multiclas
 	require(m_method->supports_multiclass(), "{} with {} doesn't support "
 			"multi classification\n", m_method->get_name(), lik->get_name());
 
-	// if regression data equals to NULL, then apply classification on training
-	// features
-	if (!data)
-	{
-		if (m_method->get_inference_type()==INF_SPARSE)
```
karlnapf, line 78: what about these types of checks? that this doesnt support sparse inference methods?

src/shogun/regression/GaussianProcessRegression.cpp
```diff
@@ -37,37 +37,9 @@ std::shared_ptr<RegressionLabels> GaussianProcessRegression::apply_regression(st
...
-	std::shared_ptr<RegressionLabels> result;
-
-	// if regression data equals to NULL, then apply regression on training
-	// features
-	if (!data)
-	{
-		std::shared_ptr<Features> feat;
-
-		// use inducing features for FITC inference method
-		if (m_method->get_inference_type()==INF_FITC_REGRESSION)
```
karlnapf, line 50: same question here, where does this logic live now?

src/shogun/regression/GaussianProcessRegression.cpp
```diff
@@ -37,37 +37,9 @@ std::shared_ptr<RegressionLabels> GaussianProcessRegression::apply_regression(st
...
-		{
-			auto fitc_method = m_method->as<FITCInferenceMethod>();
-			feat=fitc_method->get_inducing_features();
```
karlnapf, line 53: ah  nevermind. This basically fell back to predicting on the inducing features in case there was no training data provided. But that doesnt make sense anyways...it should just predict on the data provided and if a user wants to predict at the inducing points only, he has to pass those into the train method.

karlnapf (approved):
LGTM

karlnapf:
@gf712 I'll merge if you don't object?

gf712:
> @gf712 I'll merge if you don't object?

Yup, let's merge! @LiuYuHui what other non parametric models are left? Something that would be nice is if in the issue you created you could have a list of the class names and then have a tickbox so we can all keep track of the progress more easily :)

Example:

- [x] GaussianProcess
- [x] NearestCentroid
...

tests/unit/regression/GaussianProcessRegression_unittest.cc
```diff
@@ -198,16 +198,17 @@ TEST(GaussianProcessRegression, apply_regression_on_training_features)
 	auto liklihood=std::make_shared<GaussianLikelihood>(0.25);
 
 	// specify GP regression with exact inference
-	auto inf=std::make_shared<ExactInferenceMethod>(kernel, features_train,
```
gf712, line 201: but shouldn't this constructor be changed so it works without labels?

tests/unit/regression/GaussianProcessRegression_unittest.cc
```diff
@@ -452,16 +453,17 @@ TEST(GaussianProcessRegression,apply_regression_scaled_kernel)
 	auto lik=std::make_shared<GaussianLikelihood>(0.25);
 
 	// specify GP regression with exact inference
-	auto inf=std::make_shared<ExactInferenceMethod>(kernel, features_train,
```
gf712, line 455: same here

tests/unit/regression/GaussianProcessRegression_unittest.cc
karlnapf: yes, the ctor should be modified eventually
@LiuYuHui you can decide whether you want to change it in this PR, or in subsequent ones?

karlnapf: (I like smaller PRs more, so if you're ok doing this in another one, I will merge this one)

gf712: Sounds good, we can go through this when we merge into develop

LiuYuHui:
> > @gf712 I'll merge if you don't object?
> 
> Yup, let's merge! @LiuYuHui what other non parametric models are left? Something that would be nice is if in the issue you created you could have a list of the class names and then have a tickbox so we can all keep track of the progress more easily :)
> 
> Example:
> 
> * [x]  GaussianProcess
> * [x]  NearestCentroid
>   ...

I think `KernelMachine` is the only left part of non-parametric models but maybe have other edge cases.

PR #5075, opened 2020-06-22
Refactor KernelMachine

LiuYuHui:
i think `KernelMachine` is the only left part of the non-parametric model but maybe have other edge cases.

### DIFF
--- a/src/shogun/machine/KernelMachine.cpp
+++ b/src/shogun/machine/KernelMachine.cpp
@@ -46,8 +46,6 @@ KernelMachine::KernelMachine() : NonParametricMachine()
 KernelMachine::KernelMachine(const std::shared_ptr<Kernel>& k, SGVector<float64_t> alphas,
         SGVector<int32_t> svs, float64_t b) : KernelMachine()
 {
-    init();
-
     int32_t num_sv=svs.vlen;
     ASSERT(num_sv == alphas.vlen)
     create_new_model(num_sv);
@@ -59,8 +57,6 @@ KernelMachine::KernelMachine(const std::shared_ptr<Kernel>& k, SGVector<float64_
 
 KernelMachine::KernelMachine(const std::shared_ptr<KernelMachine>& machine) : KernelMachine()
 {
-	init();
-
 	SGVector<float64_t> alphas = machine->get_alphas().clone();
 	SGVector<int32_t> svs = machine->get_support_vectors().clone();
 	float64_t bias = machine->get_bias();
--- a/src/shogun/machine/KernelMachine.cpp
+++ b/src/shogun/machine/KernelMachine.cpp
@@ -38,13 +38,13 @@ struct S_THREAD_PARAM_KERNEL_MACHINE
 };
 #endif // DOXYGEN_SHOULD_SKIP_THIS
 
-KernelMachine::KernelMachine() : Machine()
+KernelMachine::KernelMachine() : NonParametricMachine()
 {
     init();
 }
 
 KernelMachine::KernelMachine(const std::shared_ptr<Kernel>& k, SGVector<float64_t> alphas,
-        SGVector<int32_t> svs, float64_t b) : Machine()
+        SGVector<int32_t> svs, float64_t b) : NonParametricMachine()
 {
     init();
 
@@ -57,7 +57,7 @@ KernelMachine::KernelMachine(const std::shared_ptr<Kernel>& k, SGVector<float64_
     set_bias(b);
 }
 
-KernelMachine::KernelMachine(const std::shared_ptr<KernelMachine>& machine) : Machine()
+KernelMachine::KernelMachine(const std::shared_ptr<KernelMachine>& machine) : NonParametricMachine()
 {
 	init();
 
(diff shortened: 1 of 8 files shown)


gf712:
lgtm!

src/shogun/regression/KRRNystrom.h
```diff
@@ -100,12 +100,13 @@ less than number of data points ({})", m_num_rkhs_basis, n);
...
 	};
 
-	bool train_machine(std::shared_ptr<Features >data) override;
-
 	/** @return object name */
 	virtual const char* get_name() const override { return "KRRNystrom"; }
 
 protected:
+
+	bool train_machine(std::shared_ptr<Features >data) override;
```
karlnapf, line 107: minor whitespace glitch

src/shogun/regression/KernelRidgeRegression.cpp
```diff
@@ -66,21 +65,10 @@ bool KernelRidgeRegression::solve_krr_system()
...
-		kernel->init(data, data);
-	}
-	ASSERT(kernel && kernel->has_features())
-
-	if (m_labels->get_num_labels() != kernel->get_num_vec_rhs())
-	{
-		error("Number of labels does not match number of kernel"
-			" columns (num_labels={} cols={}", m_labels->get_num_labels(), kernel->get_num_vec_rhs());
-	}
+	require(m_labels->get_num_labels() == data->get_num_vectors(),
```
karlnapf, line 68: Wait, wouldnt labels be provided as a parameter here? Or has the member been set elsewhere? I think in both parametric and nonparametric, the train method should have feats,labels as parameters. In nonparametric these are then stored locally

LiuYuHui: yes, the m_labels have been set in `train`.
https://github.com/shogun-toolbox/shogun/blob/3d99460a74f40649645488f095a75ea21be38df6/src/shogun/machine/Machine.cpp#L74-L77

gf712: The idea is that when the refactor is done we can remove m_labels from Machine, and then we can remove the set_labels in train

src/shogun/machine/KernelMachine.cpp
```diff
@@ -38,13 +38,13 @@ struct S_THREAD_PARAM_KERNEL_MACHINE
...
+KernelMachine::KernelMachine() : NonParametricMachine()
 {
     init();
 }
 
 KernelMachine::KernelMachine(const std::shared_ptr<Kernel>& k, SGVector<float64_t> alphas,
-        SGVector<int32_t> svs, float64_t b) : Machine()
+        SGVector<int32_t> svs, float64_t b) : NonParametricMachine()
 {
     init();
```
vigsterkr:
```suggestion
        SGVector<int32_t> svs, float64_t b) : KernelMachine()
{
```

src/shogun/machine/KernelMachine.cpp
```diff
@@ -57,7 +57,7 @@ KernelMachine::KernelMachine(const std::shared_ptr<Kernel>& k, SGVector<float64_
     set_bias(b);
 }
 
-KernelMachine::KernelMachine(const std::shared_ptr<KernelMachine>& machine) : Machine()
+KernelMachine::KernelMachine(const std::shared_ptr<KernelMachine>& machine) : NonParametricMachine()
 {
 	init();
```
vigsterkr:
```suggestion
KernelMachine::KernelMachine(const std::shared_ptr<KernelMachine>& machine) : KernelMachine()
{
```

vigsterkr (changes requested):
some minor things

src/shogun/regression/KernelRidgeRegression.cpp
karlnapf: but then we need to touch everything *again* ... imo the labels should be passed to `train_machine` (also to be consistent with feature being passed). We should never (in train) rely on labels/features being retrieved from members.

gf712: but then we need to change everything in one go no? I think the idea was to do these small PRs for npmachine, and when they are done we can have a big PR where we just replace the function signatures?
The idea is that then we can also uncomment this
https://github.com/shogun-toolbox/shogun/blob/3d99460a74f40649645488f095a75ea21be38df6/src/shogun/machine/NonParametricMachine.h#L22-L24

karlnapf: i see. so signatures come later. sorry about my moans then :)
It is just all these files need to be touched again. But I guess it will be considerably easier once these changes here are done

gf712: I think at the start we looked at also changing the signatures, but then we also thought about making things like `train_machine` pure virtual, so then we will need to change all that too. Also could make `apply` `const`. So doing all that when the labels are separated more from the implementation is imo simpler, but I also don't have to write the code :D

src/shogun/machine/KernelMachine.cpp
LiuYuHui: sorry, I don't understand why I should change `NonParametricMachine` to `KernelMachine`.

vigsterkr: https://www.variadic.xyz/2012/01/17/c11-delegating-constructors/

karlnapf:
Anything else? Otherwise I think we can merge this?

PR #5076, opened 2020-06-23
port kernel linear meta example

LiuYuHui:


gf712:
Just need the data file :)

karlnapf (commented):
LGTM (with data file)

(please prioritize other meta examples from now on, we have a lot of ones for distance/kernel)

PR #5077, opened 2020-06-23
kernel normalizer fix

gf712:
Fixes cyclic dependency between Kernel and KernelNormalizer and KernelNormalizer cleanup

### DIFF
--- a/src/shogun/io/SGIO.cpp
+++ b/src/shogun/io/SGIO.cpp
@@ -146,7 +146,7 @@ private:
 	std::shared_ptr<spdlog::sinks::sink> stderr_sink;
 };
 
-SGIO::SGIO() : show_progress(false)
+SGIO::SGIO() : show_progress(false), syntax_highlight(false)
 {
 	init_default_sink();
 	init_default_logger();
--- a/src/shogun/kernel/normalizer/SqrtDiagKernelNormalizer.h
+++ b/src/shogun/kernel/normalizer/SqrtDiagKernelNormalizer.h
@@ -58,7 +58,7 @@ class SqrtDiagKernelNormalizer : public KernelNormalizer
 
 		/** initialization of the normalizer
          * @param k kernel */
-		virtual bool init(Kernel* k)
+		bool init(Kernel* k) override
 		{
 			ASSERT(k)
 			num_sqrtdiag_lhs=k->get_num_vec_lhs();
--- a/src/shogun/kernel/normalizer/SqrtDiagKernelNormalizer.h
+++ b/src/shogun/kernel/normalizer/SqrtDiagKernelNormalizer.h
@@ -27,42 +27,38 @@ namespace shogun
 class SqrtDiagKernelNormalizer : public KernelNormalizer
 {
 	public:
+		SqrtDiagKernelNormalizer(): KernelNormalizer()
+		{
+			SG_ADD(&sqrtdiag_lhs, "sqrtdiag_lhs",
+				"sqrt(K(x,x)) for left hand side examples.")
+
+			SG_ADD(&sqrtdiag_rhs, "sqrtdiag_rhs",
+				"sqrt(K(x,x)) for right hand side examples.")
+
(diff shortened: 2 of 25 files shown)


gf712:
@LiuYuHui this should fix the issue you had with the meta example

src/shogun/kernel/normalizer/SqrtDiagKernelNormalizer.h
```diff
@@ -121,45 +117,35 @@ class SqrtDiagKernelNormalizer : public KernelNormalizer
...
-				if (v[i]==0.0)
-					v[i]=1e-16; /* avoid divide by zero exception */
+					func = [&cwk] (const int32_t i) { return std::sqrt(cwk->compute_helper(i,i, true)); };
 			}
+			else 
+				func = [&k] (const int32_t i) { return std::sqrt(k->compute(i,i));};
 
-			return (v!=NULL);
+			for (int32_t i=0; i<num; i++)
+				v[i]= std::max(func(i), 1e-16);
```
gf712: @karlnapf is it safe to assume that a kernel value is always positive? Because otherwise I need to change this back. This is just making sure that the value is never exactly zero.

src/shogun/transfer/multitask/MultitaskKernelMaskPairNormalizer.h
```diff
@@ -58,22 +59,20 @@ class MultitaskKernelMaskPairNormalizer: public KernelNormalizer
...
 
-
 		//same as first-element normalizer
 		auto old_lhs=k->lhs;
 		auto old_rhs=k->rhs;
 		k->lhs=old_lhs;
 		k->rhs=old_lhs;
 
-
 		if (std::string(k->get_name()) == "WeightedDegree") {
```
gf712, line 76: @karlnapf this does nothing afaik. I couldn't find a kernel with this name.. I wanted to change this to a dynamic cast, but I am not sure what this is meant to be...

src/shogun/kernel/normalizer/SqrtDiagKernelNormalizer.h
karlnapf: nope ... kernels can be negative pointwise. The kernel matrix needs to be psd for any input ...

src/shogun/transfer/multitask/MultitaskKernelMaskPairNormalizer.h
karlnapf: LOL, probably old code?
Remove?

gf712: I guess.. I assume what this is meant to be is `if (dynamic_cast<WeightedDegreeStringKernel*>(k))`.. either that or this is meant to be any WeightedKernel, which would also include `WeightedDegreeRBFKernel` and `WeightedDegreePositionStringKernel`

src/shogun/io/SGIO.cpp
```diff
@@ -146,7 +146,7 @@ class SGIO::RedirectSink : public spdlog::sinks::base_sink<std::mutex>
 	std::shared_ptr<spdlog::sinks::sink> stderr_sink;
 };
 
-SGIO::SGIO() : show_progress(false)
+SGIO::SGIO() : show_progress(false), syntax_highlight(false)
```
gf712, line 149: @vigsterkr valgrind was complaining about this being uninitialised. I am assuming by default this should be `false`?

vigsterkr: 👍

src/shogun/transfer/multitask/MultitaskKernelMaskPairNormalizer.h
gf712: OK, so I did some digging around and at the time this was written that was the name returned by WeightedDegreeStringKernel::get_name() so I’ll fix this!

gf712:
I'll merge this when CI is done

PR #5081, opened 2020-06-25
Example/kernel oligo string

Hephaestus12:


Hephaestus12:
[This](https://dev.azure.com/shogunml/shogun/_build/results?buildId=3749&view=logs&j=089c709a-44eb-5f6e-96e7-15e9ee1ff5bf) is where there seems to be some error causing it to timeout. What can be done about this?

gf712:
> [This](https://dev.azure.com/shogunml/shogun/_build/results?buildId=3749&view=logs&j=089c709a-44eb-5f6e-96e7-15e9ee1ff5bf) is where there seems to be some error causing it to timeout. What can be done about this?

nothing, sometimes the examples overrun the timeout, but it seems like everything ran successfully :)

karlnapf:
LGTM

karlnapf:
thx @Hephaestus12 
@gf712 do we still need to prioritize string kernel examples?
Otherwise I would say let's do kernel/distance examples last from now on and focus on other ones ....

gf712:
yup, other examples would be better now (and more interesting/challenging). The only one that would be good to merge asap is #5078

PR #5082, opened 2020-06-27
dynamic cast constraint

gf712:
Not quite automagic yet, but I am not sure we would want this constraint to be added without being explicit..

```python
In [1]: import shogun as sg                                                                                                                                                                                 

In [2]: k = sg.create_kernel("HistogramWordStringKernel")                                                                                                                                                   

In [3]: k.put("estimate", sg.create_machine("KMeans"))                                                                                                                                                      
[06/27/20 12:11:39 error] of type shogun::PluginEstimate
[06/27/20 12:11:39 error] HistogramWordStringKernel::estimate cannot be updated because it must be: of type shogun::PluginEstimate!
<ipython-input-3-11bd62e40894> in <module>
----> 1 k.put("estimate", sg.create_machine("KMeans"))

SystemError: HistogramWordStringKernel::estimate cannot be updated because it must be: of type shogun::PluginEstimate!

In [4]: k                                                                                                                                                                                                   
Out[4]: HistogramWordStringKernel(num_subscriptions={function}, cache_size=10, lhs=null, rhs=null, lhs_equals_rhs=false, num_lhs=0, num_rhs=0, combined_kernel_weight=1, optimization_initialized=false, properties=0, normalizer=IdentityKernelNormalizer(...), opt_type=FASTBUTMEMHUNGRY, initialized=false, plo_lhs=Vector<0> ), plo_rhs=Vector<0> ), ld_mean_lhs=Vector<0> ), ld_mean_rhs=Vector<0> ), sqrtdiag_lhs=Vector<0> ), sqrtdiag_rhs=Vector<0> ), mean=Vector<0> ), variance=Vector<0> ), estimate=null)

In [5]: k.put("estimate", sg.create_machine("PluginEstimate"))                                                                                                                                              

In [6]: k                                                                                                                                                                                                   
Out[6]: HistogramWordStringKernel(num_subscriptions={function}, cache_size=10, lhs=null, rhs=null, lhs_equals_rhs=false, num_lhs=0, num_rhs=0, combined_kernel_weight=1, optimization_initialized=false, properties=0, normalizer=IdentityKernelNormalizer(...), opt_type=FASTBUTMEMHUNGRY, initialized=false, plo_lhs=Vector<0> ), plo_rhs=Vector<0> ), ld_mean_lhs=Vector<0> ), ld_mean_rhs=Vector<0> ), sqrtdiag_lhs=Vector<0> ), sqrtdiag_rhs=Vector<0> ), mean=Vector<0> ), variance=Vector<0> ), estimate=PluginEstimate(...))
```

### DIFF
--- a/src/shogun/base/constraint.h
+++ b/src/shogun/base/constraint.h
@@ -103,16 +103,46 @@ namespace shogun
 		virtual bool check(const T& val) const = 0;
 	};
 
-
-	template <typename BaseType, typename DerivedType>
-	struct dynamic_cast_checker: generic_checker<BaseType>
+	template <typename T>
+	struct custom_constraint: generic_checker<T>
 	{
-		dynamic_cast_checker(): generic_checker<BaseType>()
+		template <typename Functor>
+		custom_constraint(Functor&& func): m_func(func) 
 		{
-			static_assert(!std::is_base_of_v<SGObject, BaseType>, 
-				"BaseType must be SGObject derived class");
-			static_assert(!std::is_base_of_v<BaseType, DerivedType>, 
-				"DerivedType must be a class derived from BaseType");
+		}
+
+		std::string error_msg() const override
+		{
+			return msg;
+		}
+
+	protected:
+		bool check(const T& val) const override 
+		{
+			try
+			{
+				m_func(val);
+			}
+			catch (const std::exception& e)
+			{
+				msg = std::string(e.what());
+				return false;
+			}
+
+			return true;
+		}
+
+	private:
+		std::string msg;
+		std::function<void(const T&)> m_func;
+	};
+
+
+	template <typename DerivedType>
+	struct castable: generic_checker<std::shared_ptr<SGObject>>
+	{
+		castable(): generic_checker<std::shared_ptr<SGObject>>()
+		{
 		}
 
 		std::string error_msg() const override
@@ -121,14 +151,9 @@ namespace shogun
 		}
 
 	protected:
-		bool check(const BaseType& ptr) const override 
+		bool check(const std::shared_ptr<SGObject>& ptr) const override 
 		{
-			if constexpr(traits::is_shared_ptr<BaseType>{}) {
-				return static_cast<bool>(std::dynamic_pointer_cast<DerivedType>(ptr));
-			}
-			else {
-				return dynamic_cast<DerivedType*>(ptr) != nullptr;
-			}
+			return static_cast<bool>(std::dynamic_pointer_cast<DerivedType>(ptr));
 		}
 	};
 
@@ -275,12 +300,13 @@ namespace shogun
 		}
 
 		template <typename T>
-		void check(const T& val) const
+		std::optional<std::string> check(const T& val) const
 		{
 			if (!constraint_detail::apply(val, m_funcs))
 			{
-				error("{}", constraint_detail::get_error(m_funcs));
+				return constraint_detail::get_error(m_funcs);
 			}
+			return std::nullopt;
 		}
 
 	private:
--- a/src/shogun/base/constraint.h
+++ b/src/shogun/base/constraint.h
@@ -7,11 +7,16 @@
 #ifndef __CONSTRAINT_H__
 #define __CONSTRAINT_H__
 
+#include <shogun/io/SGIO.h>
+#include <shogun/util/traits.h>
+
 #include <string>
 #include <tuple>
 
 namespace shogun
 {
+	class SGObject;
+
 	namespace constraint_detail
 	{
 		template <typename T, typename... Args, std::size_t... Idx>
@@ -86,9 +91,8 @@ namespace shogun
 	template <typename T>
 	struct generic_checker
 	{
-	public:
-		generic_checker(T val) : m_val(val){};
-		bool operator()(T val) const
+		generic_checker() = default;
+		bool operator()(const T& val) const
 		{
 			return check(val);
 		};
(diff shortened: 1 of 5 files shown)


src/shogun/kernel/string/HistogramWordStringKernel.cpp
```diff
@@ -434,14 +434,8 @@ void HistogramWordStringKernel::init()
...
-	SG_ADD(&estimate, "estimate", "Plugin Estimate.");
-
-	add_callback_function("estimate", [this](){
-		if (!std::dynamic_pointer_cast<PluginEstimate>(estimate)) {
-			error("Expected Machine to be PluginEstimate");
-			estimate.reset();
-		}
-	});
+	SG_ADD(&estimate, "estimate", "Plugin Estimate.", ParameterProperties::CONSTRAIN,
+			SG_CONSTRAINT(dynamic_cast_checker<std::shared_ptr<Machine>, PluginEstimate>()));
```
gf712: @jonpsy so here you go, you pass the type you are registering in the first template argument and then the type it must be castable to at runtime :)

jonpsy: Thank you so much!!

gf712: and then you are guaranteed to have defined behaviour if the `put` fails

src/shogun/kernel/string/HistogramWordStringKernel.cpp
```diff
@@ -434,23 +434,17 @@ void HistogramWordStringKernel::init()
...
-			estimate.reset();
-		}
-	});
+	SG_ADD(&estimate, "estimate", "Plugin Estimate.", ParameterProperties::CONSTRAIN,
+		SG_CONSTRAINT(dynamic_cast_checker<std::shared_ptr<Machine>, PluginEstimate>()));
 }
 
 #ifdef DEBUG_HWSK_COMPUTATION
-float64_t CHistogramWordStringKernel::compute_slow(int32_t idx_a, int32_t idx_b)
+float64_t HistogramWordStringKernel::compute_slow(int32_t idx_a, int32_t idx_b)
```
karlnapf: should rather delete all this code....debugging with ifdef is something we try to avoid these days I guess? :)

src/shogun/kernel/string/HistogramWordStringKernel.cpp
```diff
@@ -434,23 +434,17 @@ void HistogramWordStringKernel::init()
...
-	SG_ADD(&estimate, "estimate", "Plugin Estimate.");
-
-	add_callback_function("estimate", [this](){
-		if (!std::dynamic_pointer_cast<PluginEstimate>(estimate)) {
-			error("Expected Machine to be PluginEstimate");
-			estimate.reset();
-		}
-	});
+	SG_ADD(&estimate, "estimate", "Plugin Estimate.", ParameterProperties::CONSTRAIN,
+		SG_CONSTRAINT(dynamic_cast_checker<std::shared_ptr<Machine>, PluginEstimate>()));
```
karlnapf: semantics are here that anything placed with put should be a subclass of Machine, but the type stored is PluginEstimate? Then the latter would be kind of redundant in the specification... is it necessary? Because put already checks that?

gf712: Hmmm I guess I can accept SGObject type

src/shogun/kernel/string/HistogramWordStringKernel.cpp
gf712: sure, I was just fixing the typos :p

src/shogun/kernel/string/HistogramWordStringKernel.cpp
```diff
@@ -434,57 +434,6 @@ void HistogramWordStringKernel::init()
...
-	}
-
-	if (initialized)
-		result /=  (sqrtdiag_lhs[idx_a]*sqrtdiag_rhs[idx_b]) ;
-
-	std::static_pointer_cast<CStringFeatures<uint16_t>>(lhs)->free_feature_vector(avec, idx_a, free_avec);
-	std::static_pointer_cast<CStringFeatures<uint16_t>>(rhs)->free_feature_vector(bvec, idx_b, free_bvec);
-	return result;
+	SG_ADD(&estimate, "estimate", "Plugin Estimate.", ParameterProperties::CONSTRAIN,
+		SG_CONSTRAINT(castable<PluginEstimate>()));
```
gf712, line 429: @karlnapf  should `estimate` now be of type `PluginEstimate` or `Machine`? Basically should the C++ constructor and setters accept `Machine` or `PluginEstimate`? Because now we can check the type easily with `get` we can go back to the more C++ way of doing things when using the "internal" C++ API, e.g. constructor + setter accepting the most specific type possible. (by setter I mean an actual class member function, like `set_estimate(...)`.

karlnapf:
:partying_face:

src/shogun/base/constraint.h
```diff
@@ -86,18 +91,80 @@ namespace shogun
...
+	protected:
+		bool check(const T& val) const override 
+		{
+			try
+			{
+				m_func(val);
+			}
+			catch (const std::exception& e)
+			{
+				msg = std::string(e.what());
```
LiuYuHui, line 128: it seems like that msg can't be assigned in `check` function, as the `check` is `const `.

gf712: ah true.. I didn't test this, feel free to fix it :)

gf712: you can make `msg` `mutable`

Issue #5090, opened 2020-07-07
Refactor Preprocessor


jonpsy:
The preprocessor classes have a few redundant features which need cleanup. I'll list them below for convenience 

- ```m_fitted``` is scattered around throughout the code, which could prove to be fatal. 
-  The ```cleanup()``` method should be removed as objects are destroyed automatically with smart pointers
-  ```fit()``` method could be split into two, namely, ```fit()``` and ```fit_impl()``` where the former does the input-checks and the latter does the actual computation.

Here's how I want to go about solving this:

- Remove ```cleanup()``` method from preprocessor class and derivatives
- Separate fit() method into two wherein it calls the **helper function** fit_impl() which handles the computation which finally sets m_fitted to true. This immediately solves the m_fitted problem, now one could rest easy about manually toggling the m_fitted method inside class implementation.

karlnapf:
Thx!

PR #5091, opened 2020-07-07
[1] Remove cleanup 

jonpsy:
#5090 

@vigsterkr @karlnapf @gf712

### DIFF
--- a/src/shogun/preprocessor/FeatureSelection.cpp
+++ b/src/shogun/preprocessor/FeatureSelection.cpp
@@ -82,13 +82,6 @@ FeatureSelection<ST>::~FeatureSelection()
 
 }
 
-
-template <class ST>
-void FeatureSelection<ST>::cleanup()
-{
-	m_subset->remove_all_subsets();
-}
-
 template <class ST>
 std::shared_ptr<Features> FeatureSelection<ST>::apply_backward_elimination(std::shared_ptr<Features> features)
 {
--- a/src/shogun/preprocessor/FisherLDA.cpp
+++ b/src/shogun/preprocessor/FisherLDA.cpp
@@ -195,13 +195,6 @@ void FisherLDA::solver_classic(
 	}
 }
 
-void FisherLDA::cleanup()
-{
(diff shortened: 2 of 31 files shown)


src/shogun/preprocessor/FeatureSelection.cpp
```diff
@@ -82,13 +82,6 @@ FeatureSelection<ST>::~FeatureSelection()
 
 }
 
-
-template <class ST>
-void FeatureSelection<ST>::cleanup()
-{
-	m_subset->remove_all_subsets();
```
karlnapf, line 89: what about this?

src/shogun/preprocessor/FisherLDA.cpp
```diff
@@ -195,13 +195,6 @@ void FisherLDA::solver_classic(
 	}
 }
 
-void FisherLDA::cleanup()
-{
-	m_transformation_matrix=SGMatrix<float64_t>();
```
karlnapf, line 200: how is this now handled?

karlnapf (changes requested):
The empty ones are no-brainers to remove

But the ones that do stuff that is not just freeing memory, you need to check whether the code relies on the changes being made in cleanup. You can't just delete that code and hope for the best

src/shogun/preprocessor/FeatureSelection.cpp
jonpsy: The author [already used](https://github.com/shogun-toolbox/shogun/blob/f01b18655994c7a4630acc4b67b60912a957217e/src/shogun/preprocessor/FeatureSelection.cpp#L184) ```remove_all_subsets()``` inside the transform method (which is the public interface). If we call it **n** number of times, it will reset m_subset that many times and start anew. I suppose the author only implemented cleanup() because it was purely virtual.

karlnapf: I checked the code, seems ok

karlnapf:
thx!

PR #5096, opened 2020-07-10
[2] Add fit_impl

jonpsy:
#5090 

@vigsterkr btw I added your std::atomic<bool> suggestion ;)

@gf712  @karlnapf

### DIFF
--- a/src/shogun/preprocessor/DensePreprocessor.cpp
+++ b/src/shogun/preprocessor/DensePreprocessor.cpp
@@ -132,12 +132,6 @@ void DensePreprocessor<ST>::fit(std::shared_ptr<Features> f)
 }
 
 template <class ST>
-void DensePreprocessor<ST>::fit_impl(SGMatrix<ST> feature_matrix)
-{
-	not_implemented(SOURCE_LOCATION);
-}
-
-template <class ST>
 SGMatrix<ST>
 DensePreprocessor<ST>::inverse_apply_to_matrix(SGMatrix<ST> matrix)
 {
--- a/src/shogun/preprocessor/DensePreprocessor.cpp
+++ b/src/shogun/preprocessor/DensePreprocessor.cpp
@@ -119,7 +119,7 @@ DensePreprocessor<ST>::inverse_transform(std::shared_ptr<Features> features, boo
 template <class ST>
 void DensePreprocessor<ST>::fit(std::shared_ptr<Features> f)
 {
-	require(f, "Features are not provided");
+	require(f, "No features provided");
 
 	auto dense_features = f->as<DenseFeatures<ST>>();
 	auto num_features = dense_features->get_num_features();
--- a/src/shogun/preprocessor/DensePreprocessor.cpp
+++ b/src/shogun/preprocessor/DensePreprocessor.cpp
@@ -128,7 +128,7 @@ void DensePreprocessor<ST>::fit(std::shared_ptr<Features> f)
 	auto feature_matrix = dense_features->get_feature_matrix();
 	fit_impl(feature_matrix);
 	
-	m_fitted = true;
+	m_fitted.store(true);
 }
 
 template <class ST>
--- a/src/shogun/preprocessor/DensePreprocessor.cpp
+++ b/src/shogun/preprocessor/DensePreprocessor.cpp
@@ -117,6 +117,27 @@ DensePreprocessor<ST>::inverse_transform(std::shared_ptr<Features> features, boo
 }
 
 template <class ST>
+void DensePreprocessor<ST>::fit(std::shared_ptr<Features> f)
+{
+	require(f, "Features are not provided");
+
+	auto dense_features = f->as<DenseFeatures<ST>>();
+	auto num_features = dense_features->get_num_features();
+	require(num_features > 0, "Dimension of provided features {} must be positive", num_features);
+	
+	auto feature_matrix = dense_features->get_feature_matrix();
+	fit_impl(feature_matrix);
+	
+	m_fitted = true;
+}
+
+template <class ST>
+void DensePreprocessor<ST>::fit_impl(SGMatrix<ST> feature_matrix)
+{
+	not_implemented(SOURCE_LOCATION);
+}
+
+template <class ST>
 SGMatrix<ST>
 DensePreprocessor<ST>::inverse_apply_to_matrix(SGMatrix<ST> matrix)
 {
--- a/src/shogun/preprocessor/DensePreprocessor.h
+++ b/src/shogun/preprocessor/DensePreprocessor.h
@@ -79,7 +79,9 @@ template <class ST> class DensePreprocessor : public Preprocessor
 		 */
 		virtual SGMatrix<ST> inverse_apply_to_matrix(SGMatrix<ST> matrix);
 
-		virtual void fit_impl(SGMatrix<ST> feature_matrix);
+		virtual void fit_impl(const SGMatrix<ST>& feature_matrix)
+		{
+		}
 };
 
 }
--- a/src/shogun/preprocessor/DensePreprocessor.h
+++ b/src/shogun/preprocessor/DensePreprocessor.h
@@ -62,6 +62,8 @@ template <class ST> class DensePreprocessor : public Preprocessor
 		/// return a type of preprocessor
 		EPreprocessorType get_type() const override;
 
+		void fit(std::shared_ptr<shogun::Features>) override;
+
 	protected:
 		/** Apply preprocessor on matrix. Subclasses should try to apply in
 		 * place to avoid copying.
@@ -76,6 +78,8 @@ template <class ST> class DensePreprocessor : public Preprocessor
 		 * @return the matrix after applying the preprocessor
 		 */
 		virtual SGMatrix<ST> inverse_apply_to_matrix(SGMatrix<ST> matrix);
+
+		virtual void fit_impl(SGMatrix<ST> feature_matrix);
 };
 
 }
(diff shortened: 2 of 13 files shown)


jonpsy:
- m_fitted is now more structured.
- For classes which don't need fit method, throws not_implemented error

jonpsy:
Sorry to say, I'd have to revert the std::atomic<bool> it's causing errors which are out of the scope of PR.

src/shogun/preprocessor/RFFPreprocessor.h
```diff
@@ -84,6 +82,8 @@ class RFFPreprocessor: public RandomMixin<DensePreprocessor<float64_t>> {
 
 	virtual SGMatrix<float64_t> apply_to_matrix(SGMatrix<float64_t> matrix) override;
 
+	virtual void fit_impl(SGMatrix<float64_t> feature_matrix) override;
```
vigsterkr: remove `virtual`

src/shogun/preprocessor/RescaleFeatures.h
```diff
@@ -64,6 +56,8 @@ namespace shogun
 	protected:
 		virtual SGMatrix<float64_t> apply_to_matrix(SGMatrix<float64_t> matrix);
 
+		virtual void fit_impl(SGMatrix<float64_t> feature_matrix) override;
```
vigsterkr: remove `virtual`

vigsterkr (changes requested):
some minor things.

although i've just realised that Transformer::m_fitted is a simple bool.... it should be eventual changed to `std::atomic<bool>` imo

src/shogun/preprocessor/DensePreprocessor.cpp
```diff
@@ -116,6 +116,27 @@ DensePreprocessor<ST>::inverse_transform(std::shared_ptr<Features> features, boo
...
+	require(num_features > 0, "Dimension of provided features {} must be positive", num_features);
+	
+	auto feature_matrix = dense_features->get_feature_matrix();
+	fit_impl(feature_matrix);
+	
+	m_fitted = true;
+}
+
+template <class ST>
+void DensePreprocessor<ST>::fit_impl(SGMatrix<ST> feature_matrix)
```
gf712:
```suggestion
void DensePreprocessor<ST>::fit_impl(const SGMatrix<ST>& feature_matrix)
```

gf712: Shouldn't this be pure virtual and then make this is a noop, rather  `not_implemented` when it is implemented? The issue with throwing is that you might add this to a `Pipeline`, and this will throw an error because the `Pipeline` object will always do fit + transform.

gf712:
> Sorry to say, I'd have to revert the std::atomic it's causing errors which are out of the scope of PR.

What are the issues with making m_fitted atomic?

src/shogun/preprocessor/DensePreprocessor.cpp
jonpsy: Purely virtual isn't an option because few preproc don't employ fit_impl method. This would cause compile error, I guess we could make it empty function then?

gf712: yes, I think so

jonpsy:
Once this PR #5097  is ready, I could make this support std::atomic<bool>

src/shogun/preprocessor/DensePreprocessor.cpp
```diff
@@ -116,6 +116,21 @@ DensePreprocessor<ST>::inverse_transform(std::shared_ptr<Features> features, boo
 	return std::make_shared<DenseFeatures<ST>>(feat_matrix);
 }
 
+template <class ST>
+void DensePreprocessor<ST>::fit(std::shared_ptr<Features> f)
+{
+	require(f, "Features are not provided");
```
karlnapf: Nit: "No features provided" (better style)

src/shogun/preprocessor/DensePreprocessor.h
```diff
@@ -76,6 +78,10 @@ template <class ST> class DensePreprocessor : public Preprocessor
 		 * @return the matrix after applying the preprocessor
 		 */
 		virtual SGMatrix<ST> inverse_apply_to_matrix(SGMatrix<ST> matrix);
+
+		virtual void fit_impl(const SGMatrix<ST>& feature_matrix)
```
karlnapf, line 82: no error in here?

karlnapf: I like the SGMarix!

karlnapf: what about inplace? it is const

src/shogun/preprocessor/PruneVarSubMean.cpp
```diff
@@ -27,20 +27,17 @@ PruneVarSubMean::~PruneVarSubMean()
...
 }
 
-void PruneVarSubMean::fit(std::shared_ptr<Features> features)
+void PruneVarSubMean::fit_impl(const SGMatrix<float64_t>& feature_matrix)
 {
 
-	auto simple_features = features->as<DenseFeatures<float64_t>>();
-	auto num_examples = simple_features->get_num_vectors();
-	auto num_features = simple_features->get_num_features();
+	auto num_examples = feature_matrix.num_cols;
```
karlnapf, line 33: nice!

karlnapf (commented):
Nice! Some minori-sh comments/questions

src/shogun/preprocessor/DensePreprocessor.h
gf712: The idea is that you shouldn’t get an error if the Transformer doesn’t implement a fit method because it doesn’t make sense. But I think this should be pure virtual function and then each implementation handles this.. that way nothing unexpected happens..

karlnapf: could you explain why it doesnt make sense? 
I am ok with a purely virtual method, just needs more refactoring

Any thoughts about the const matrix and inplace?

src/shogun/preprocessor/DensePreprocessor.cpp
jonpsy: > Shouldn't this be pure virtual and then make this is a noop, rather `not_implemented` when it is implemented? The issue with throwing is that you might add this to a `Pipeline`, and this will throw an error because the `Pipeline` object will always do fit + transform.

@karlnapf  this is why it doesn't make sense(from the second sentence).

jonpsy: @gf712 , what I did instead was make the fit_impl a noop in ```DensePreprocessor```. This way we needn't make it noop in each and every subclass which doesn't use fit_impl, and those who use fit_impl simply override this :)

src/shogun/preprocessor/DensePreprocessor.h
gf712: > could you explain why it doesnt make sense?

ah sorry, yes. It doesn't make sense for something like `LogPlusOne`. There is nothing to `fit`, only to `apply`. So if I put this `Transformer` in a `Pipeline` it will always throw (if we have an exception, instead of a noop)

> Any thoughts about the const matrix and inplace?

Isn't the inplace for `transform` rather than `fit`?

src/shogun/preprocessor/DensePreprocessor.cpp
gf712: sure, I think that should be fine :)

karlnapf: how many subclasses dont have a `fit` method? I guess that would be very few (`LogPlusOne`, which else?)
I'd actually prefer to be as explicit as possible with this. I.e. making the method purely virtual in the base class, and then the classes that don't need it override with as a no-op. That seems to make the structure much more clear. I.e. it is the exception that a subclass has a noop fit method, not the common case. If we make it a noop in the base class we are putting the cart in front of the horse imo

src/shogun/preprocessor/DensePreprocessor.h
karlnapf: ah yes, sorry I keep confusing the two.
See above comment on error/purely-virtual/no-op

src/shogun/preprocessor/DensePreprocessor.cpp
jonpsy: A total of five(Pnorm, NormOne, LogPlusOne, Homogenous, SumOne) don't have fit method. 

I see your point, I drew my inspiration from the ```fit``` implementation we have on ```Transformer.h``` it is defined as noop, so those who don't use it get a noop others override it. We can go with your method as well, but I'd like to know why we didn't use this same methodology on ```fit```?

jonpsy:
std::atomic integrated :)

src/shogun/preprocessor/DensePreprocessor.h
```diff
@@ -62,6 +62,8 @@ template <class ST> class DensePreprocessor : public Preprocessor
 		/// return a type of preprocessor
 		EPreprocessorType get_type() const override;
 
+		void fit(std::shared_ptr<shogun::Features>) override;
```
LiuYuHui:
```suggestion
		void fit(std::shared_ptr<Features>) override;
```

jonpsy: Nice catch ;)

src/shogun/preprocessor/DensePreprocessor.cpp
karlnapf: Interesting, thx for pointing that out. In that case let's stick with what is there then for consistency reasons (i.e. with what you suggested)

Asking why something is the way it is can be difficult in shogun ... many authors, many opinions ;)

jonpsy:
GTM??

gf712:
> GTM??

I think so, I'll let @karlnapf merge it just in case he still needs to do a final check :)

karlnapf (approved):
LGTM

karlnapf:
I'll wait for the CI and then hit the button :)
Thanks!

PR #5097, opened 2020-07-11
basic support for std::atomic with Any

gf712:
support std::atomic parameter registration

### DIFF
--- a/src/shogun/lib/any.h
+++ b/src/shogun/lib/any.h
@@ -201,6 +201,14 @@ namespace shogun
 		virtual void exit_map(size_t* size) = 0;
 
 		template <typename T>
+		void on(std::atomic<T>* val)
+		{
+			T tmp = val->load();
+			on(std::addressof(tmp));
+			val->store(tmp);
+		}
+
+		template <typename T>
 		void on_matrix_row(index_t* rows, index_t* cols, SGMatrix<T>* _v)
 		{
 			enter_matrix_row(rows, cols);
@@ -587,6 +595,8 @@ namespace shogun
 			{
 				return T(value);
 			}
+			else if constexpr (traits::is_atomic_v<T>)
+				return T(value.load());
 			else
 			{
 				// we assert something that is false to see the type T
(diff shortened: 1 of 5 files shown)


jonpsy:
Yay!!

gf712:
@vigsterkr this might not be the most correct implementation, but it should be fine for the purpose of serde + access from interfaces

src/shogun/util/traits.h
```diff
@@ -184,6 +185,29 @@ namespace shogun
...
+		template<typename>
+		struct is_atomic: std::false_type {};
+
+		template <typename T>
+		struct is_atomic<std::atomic<T>> : std::true_type
+		{
+		};
+
+		template <typename T>
+		struct atomic_type {};
```
gf712, line 198: replaces std::atomic<T>::value_type, because it is only available in very recent compilers (gcc-9 and clang trunk)

src/shogun/lib/any.h
```diff
@@ -1329,34 +1349,34 @@ namespace shogun
...
 		{
 			Any::register_caster<T, std::string>(
-			    [](T value) { return std::to_string(value); });
+			    [](const T& value) { return std::to_string(value); });
 		}
 		if constexpr (traits::has_to_string<T>::value)
 		{
 			Any::register_caster<T, std::string>(
-			    [](T value) { return value.to_string(); });
+			    [](const T& value) { return value.to_string(); });
```
gf712, line 1379: made const ref because you cannot copy std::atomic, but it also doesn't make sense to copy all these values into the casting function

src/shogun/lib/any.h
```diff
@@ -965,7 +982,10 @@ namespace shogun
 		 */
 		void set(void** storage, const void* v) const override
 		{
-			mutable_value_of<T>(storage) = value_of(typed_pointer<T>(v));
+			if constexpr(traits::is_atomic_v<T>)
+				mutable_value_of<T>(storage).store(typed_pointer<T>(v)->load());
```
gf712, line 986: cannot assign to std::atomic with `operator=` using a reference on the left handside

src/shogun/lib/any.h
```diff
@@ -200,6 +200,14 @@ namespace shogun
 		virtual void exit_std_vector(size_t* size) = 0;
 		virtual void exit_map(size_t* size) = 0;
 
+		template <typename T>
+		void on(std::atomic<T>* val)
+		{
+			T tmp = val->load();
+			on(std::addressof(tmp));
+			val->store(tmp);
```
gf712, line 208: the `tmp` variable is a copy of the value managed by std::atomic, so need to store the `tmp` value in case it has been changed when visiting on `T*`

PR #5098, opened 2020-07-13
Generate seed for each machine in each fold

gf712:
@karlnapf is this what you had in mind for xvalidation?

### DIFF
--- a/src/shogun/base/SGObject.h
+++ b/src/shogun/base/SGObject.h
@@ -796,6 +796,18 @@ protected:
 			);	
 		}
 
+		if constexpr (std::is_arithmetic_v<T> && !std::is_same_v<T, bool>)
+		{
+			Any::register_visitor<int32_t, ParameterPutInterface<T>>(
+			    [](int32_t* value, auto* visitor) { 
+					*value = utils::safe_convert<int32_t>(visitor->m_value); 
+				});
+			Any::register_visitor<int64_t, ParameterPutInterface<T>>(
+			    [](int64_t* value, auto* visitor) { 
+					*value = utils::safe_convert<int64_t>(visitor->m_value); 
+				});
+		}
+
 		if constexpr (traits::is_functional<T>::value)
 		{
 			if constexpr (!traits::returns_void<T>::value)
@@ -844,6 +856,13 @@ protected:
 					visitor->m_value = *value;	
 				}
 			);
+			if constexpr(std::is_arithmetic_v<T> && !std::is_same_v<T, bool>) 
+			{
+				Any::register_visitor<T, ParameterGetterInterface<int32_t, int32_t>>(
+					[](T* value, auto* visitor) {visitor->m_value = utils::safe_convert<int32_t>(*value);});
+				Any::register_visitor<T, ParameterGetterInterface<int64_t, int64_t>>(
+					[](T* value, auto* visitor) {visitor->m_value = utils::safe_convert<int64_t>(*value);});
+			}
 		}
 	}
 	/** Registers a class parameter which is identified by a tag.
--- a/src/shogun/evaluation/CrossValidation.cpp
+++ b/src/shogun/evaluation/CrossValidation.cpp
@@ -14,7 +14,7 @@
 #include <shogun/lib/observers/ObservedValueTemplated.h>
 #include <shogun/machine/Machine.h>
 #include <shogun/mathematics/Statistics.h>
-#include <shogun/mathematics/UniformIntDistribution.h>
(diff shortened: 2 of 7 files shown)


src/shogun/evaluation/CrossValidation.cpp
```diff
@@ -95,6 +96,12 @@ float64_t CrossValidation::evaluate_one_run(int64_t index) const
 
 	SGVector<float64_t> results(num_subsets);
 
+	std::vector<int32_t> seed_vector(num_subsets);
+	UniformIntDistribution<int32_t> seed_dist(std::numeric_limits<int32_t>::min(), 
+		std::numeric_limits<int32_t>::max());
```
gf712: @vigsterkr not sure how else to generate int32_t values. Should seed be [0, 2^31-1] or [-2^31, 2^31-1]?

vigsterkr: why dont you just use the prng directly?

gf712: because prng returns a uint64_t, which then would get truncated when casted to int32 no?

vigsterkr: why do you need signed int32_t? i mean it's gonna be a seed for another prng no?
for mersenne prng the seed should be actually unsigned 💃 so we have a bug in the Seedable :P

src/shogun/evaluation/CrossValidation.cpp
```diff
@@ -95,6 +96,12 @@ float64_t CrossValidation::evaluate_one_run(int64_t index) const
 
 	SGVector<float64_t> results(num_subsets);
 
+	std::vector<int32_t> seed_vector(num_subsets);
```
vigsterkr: `std::vector<decltype(m_prng::result_type)>`?

src/shogun/evaluation/CrossValidation.cpp
gf712: So Seedable needs to know what the input type of the prng is?

karlnapf:
> @karlnapf is this what you had in mind for xvalidation?

yes! :)  I guess only some details need updating.

src/shogun/mathematics/RandomNamespace.h
```diff
@@ -105,6 +105,25 @@ namespace shogun
...
+		 * @param num_elements number of elements to generate
+		 */
+		template <typename PRNG, typename T>
+		std::vector<typename std::remove_reference_t<PRNG>::result_type> 
+		generate_array(PRNG&& prng, const T& num_elements) 
+		{
+			std::vector<typename std::remove_reference_t<PRNG>::result_type> result;
+			result.reserve(num_elements);
+			std::generate_n(std::back_inserter(result), num_elements, 
+				[prng = std::forward<PRNG>(prng)]() mutable {
```
vigsterkr, line 121: badass!

src/shogun/mathematics/Seedable.h
```diff
@@ -62,7 +62,7 @@ namespace shogun
 			random::seed(object, m_seed);
 		}
 
-		int32_t m_seed;
+		int64_t m_seed;
```
vigsterkr, line 65: these lines should be changed as well to `int64_t`:
https://github.com/shogun-toolbox/shogun/blob/167880c1ea5bc62ec4906f48a86d72427eb9a014/src/shogun/mathematics/Seedable.h#L24
https://github.com/shogun-toolbox/shogun/blob/167880c1ea5bc62ec4906f48a86d72427eb9a014/src/shogun/mathematics/Seedable.h#L31
https://github.com/shogun-toolbox/shogun/blob/167880c1ea5bc62ec4906f48a86d72427eb9a014/src/shogun/mathematics/Seedable.h#L68

vigsterkr (changes requested):

src/shogun/base/SGObject.h
```diff
@@ -844,6 +856,13 @@ class SGObject: public std::enable_shared_from_this<SGObject>
 					visitor->m_value = *value;	
 				}
 			);
+			if constexpr(std::is_arithmetic_v<T> && !std::is_same_v<T, bool>) 
+			{
+				Any::register_visitor<T, ParameterGetterInterface<int32_t, int32_t>>(
```
karlnapf, line 861: that is nice to do that down here!

gf712: maybe.. But this is getting a bit out of hand :D I thought about separating all of this out and adding to a mixin, but I think the resulting multiple inheritance might cause headaches..

PR #5105, opened 2020-08-03
Add meta example for hessian locally linear embedding converter.

Hephaestus12:


gf712:
Did you add data files for these PRs?

Hephaestus12:
> Did you add data files for these PRs?

No, I was waiting for approval regarding the meta example code. Should I go ahead with the datafiles?

gf712:
I think it seems fine. Would be good to see if you still have issues with octave with other transformers

geektoni:
The same issue with the integration test happens here.

PR #5110, opened 2020-08-07
Suggest class name if class name not found

### LINKED ISSUE #4475: Suggest class name in factory instantiation

LiuYuHui:
fixes #4475

### DIFF
--- a/src/shogun/base/base_types.h
+++ b/src/shogun/base/base_types.h
@@ -167,6 +167,20 @@ namespace shogun
 
 	template <typename T>
 	inline constexpr bool is_auto_value_v = is_auto_value<T>::value;
+
+
+	template<typename classType>
+	std::string_view name_lookup()
+	{
+		if constexpr(std::is_same_v<classType, Machine>)
+		{
+			return "Machine";
+		}
+		else 
+		{
+			return "SGObject";
+		}
+	};
 } // namespace shogun
 
 #endif // BASE_TYPES__H
--- a/src/shogun/base/class_list.cpp.templ
+++ b/src/shogun/base/class_list.cpp.templ
@@ -75,11 +75,11 @@ std::string shogun::find_correct_name(const std::string& name)
 {
     using pair = std::pair<int32_t, std::string>;
     std::priority_queue<pair, std::vector<pair>, std::greater<pair>> names;
-    LevenshteinDistance levenshtein;
+    auto levenshtein = std::make_shared<LevenshteinDistance>();
     std::for_each(classes.begin(), classes.end(), 
     	[&](const auto& class_pair){
 			auto class_name = class_pair.first;
-			auto dis = levenshtein.distance(class_name, name);
+			auto dis = levenshtein->distance(class_name, name);
 			names.push(std::make_pair(dis, class_name));
     	});
     return names.top().second;
--- a/src/shogun/base/class_list.cpp.templ
+++ b/src/shogun/base/class_list.cpp.templ
@@ -70,3 +70,17 @@ std::set<std::string> shogun::available_objects()
 	}
 	return result;
 }
+
+std::string shogun::find_correct_name(const std::string& name)
+{
+    using pair = std::pair<int32_t, std::string>;
+    std::priority_queue<pair, std::vector<pair>, std::greater<pair>> names;
+    LevenshteinDistance levenshtein;
+    std::for_each(classes.begin(), classes.end(), 
+    	[&](const auto& class_pair){
+			auto class_name = class_pair.first;
+			auto dis = levenshtein.distance(class_name, name);
+			names.push(std::make_pair(dis, class_name));
+    	});
+    return names.top().second;
+}
--- a/src/shogun/base/class_list.h
+++ b/src/shogun/base/class_list.h
@@ -17,6 +17,7 @@
 
 #include <set>
 #include <string>
+#include <queue>
 
 namespace shogun {
 	class SGObject;
@@ -33,6 +34,9 @@ namespace shogun {
 	 * the requested type and the object's type do not match.
 	 *
 	 */
+
+	std::string find_correct_name(const std::string& name);
+
 	template <class T>
 	std::shared_ptr<T> create_object(
 	    const char* name,
@@ -41,9 +45,9 @@ namespace shogun {
 		auto object = create(name, pt);
 		if (!object)
 		{
+			const auto& correct_name = find_correct_name(name);
 			error(
-			    "Class {} with primitive type {} does not exist.", name,
-			    ptype_name(pt).c_str());
+			    "Class {} does not exist. Did you mean {} ?", name, correct_name);
 		}
 		auto cast = std::dynamic_pointer_cast<T>(object);
 		if (cast == nullptr)
--- a/src/shogun/base/class_list.h
+++ b/src/shogun/base/class_list.h
@@ -28,15 +28,17 @@ namespace shogun {
 	 */
 	std::shared_ptr<SGObject> create(const char* sgserializable_name, EPrimitiveType generic);
 
+	/** Use Levenshtein distance to find the most similar name
+	 *
+	 */
+	std::string find_correct_name(const std::string& name);
+
 	/** Creates new shogun instance, typed.
 	 *
 	 * Throws an exception in case there is no such classname or
 	 * the requested type and the object's type do not match.
 	 *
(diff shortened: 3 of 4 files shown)


gf712:
Nice! Can you show a python code snippet that uses this?

src/shogun/base/class_list.h
```diff
@@ -33,6 +34,9 @@ namespace shogun {
 	 * the requested type and the object's type do not match.
 	 *
 	 */
+
+	std::string find_correct_name(const std::string& name);
```
gf712: can you move this declaration please? It is in the way of doxygen string

LiuYuHui:
for example `sg.create_machine("liblinear")`, such exception will be thrown.
```
<ipython-input-2-a1dac361590b> in <module>
----> 1 sg.create_machine("liblinear")

~/shogun/build/src/interfaces/python/shogun.py in _internal_factory(name, *args, **kwargs)
    107     def _internal_factory(name, *args, **kwargs):
    108 
--> 109         new_obj = _obj(name, *args)
    110         for k,v in kwargs.items():
    111             new_obj.put(k, v)

SystemError: Class liblinear does not exist. Did you mean LibLinear ?
```

gf712:
> for example `sg.create_machine("liblinear")`, such exception will be thrown.
> 
> ```
> ---------------------------------------------------------------------------
> SystemError                               Traceback (most recent call last)
> <ipython-input-2-a1dac361590b> in <module>
> ----> 1 sg.create_machine("liblinear")
> 
> ~/shogun/build/src/interfaces/python/shogun.py in _internal_factory(name, *args, **kwargs)
>     107     def _internal_factory(name, *args, **kwargs):
>     108 
> --> 109         new_obj = _obj(name, *args)
>     110         for k,v in kwargs.items():
>     111             new_obj.put(k, v)
> 
> SystemError: Class liblinear does not exist. Did you mean LibLinear ?
> ```

Nice! @karlnapf could you check if that error message is sufficient, please?

karlnapf:
> > for example `sg.create_machine("liblinear")`, such exception will be thrown.
> > ```
> > ---------------------------------------------------------------------------
> > SystemError                               Traceback (most recent call last)
> > <ipython-input-2-a1dac361590b> in <module>
> > ----> 1 sg.create_machine("liblinear")
> > 
> > ~/shogun/build/src/interfaces/python/shogun.py in _internal_factory(name, *args, **kwargs)
> >     107     def _internal_factory(name, *args, **kwargs):
> >     108 
> > --> 109         new_obj = _obj(name, *args)
> >     110         for k,v in kwargs.items():
> >     111             new_obj.put(k, v)
> > 
> > SystemError: Class liblinear does not exist. Did you mean LibLinear ?
> > ```
> 
> Nice! @karlnapf could you check if that error message is sufficient, please?

Love it. I think it is decent. I guess doing something like `"Machine \"liblinear\ does not exist. Did you mean \"LibLinear\""? is hard (adding the type?).

src/shogun/base/class_list.cpp.templ
```diff
@@ -70,3 +70,17 @@ std::set<std::string> shogun::available_objects()
 	}
 	return result;
 }
+
+std::string shogun::find_correct_name(const std::string& name)
+{
+    using pair = std::pair<int32_t, std::string>;
+    std::priority_queue<pair, std::vector<pair>, std::greater<pair>> names;
+    LevenshteinDistance levenshtein;
```
karlnapf: on stack? Don't we do shogun objects on heap usually?

karlnapf (commented):
Very cute! Did you try some failure cases of this?
Like severe typos?

gf712:
> Love it. I think it is decent. I guess doing something like `"Machine "liblinear\ does not exist. Did you mean "LibLinear""? is hard (adding the type?).
Yes, that should be possible, we just need to overload a struct with a `string_view` member for the name and in `create_object` use `T` to lookup the name. Unfortunately I cannot think of a way to automate that part.
```cpp
template <typename T>
struct name_lookup {
  static constexpr std::string_view name = "SGObject";
};

template <>
struct name_lookup<Machine> {
  static constexpr std::string_view name = "Machine";
};

// get the name
name_lookup<Machine>::name;
```

src/shogun/base/class_list.cpp.templ
gf712: probably should, I think we should be consistent even if here it isn't needed

src/shogun/base/class_list.h
```diff
@@ -17,16 +17,34 @@
...
+	{
+		if constexpr(std::is_same_v<classType, Machine>)
+		{
+			return "Machine";
+		}
+		else 
+		{
+			return "SGObject";
+		}
+	};
```
gf712: I think this should be in base_types.h and also have the same base types that are forward declared there

src/shogun/base/base_types.h
```diff
@@ -167,6 +167,20 @@ namespace shogun
 
 	template <typename T>
 	inline constexpr bool is_auto_value_v = is_auto_value<T>::value;
+
+
+	template<typename classType>
+	std::string_view name_lookup()
```
gf712, line 173: could you just this to a more meaningful name please?

gf712: @vigsterkr do you know if there is another way to lookup the class name and get the name as a string?

src/shogun/base/base_types.h
```diff
@@ -167,6 +167,20 @@ namespace shogun
...
+	{
+		if constexpr(std::is_same_v<classType, Machine>)
+		{
+			return "Machine";
+		}
+		else
```
gf712, line 179: what about the base class names?

src/shogun/distance/LevenshteinDistance.h
```diff
@@ -114,7 +114,15 @@ namespace shogun
 
 		std::shared_ptr<Features>
 		replace_lhs(std::shared_ptr<Features> lhs) override;
+		
+		using Distance::distance;
 
+		int32_t distance(const std::string& lhs, const std::string& rhs)
+		{
+			const SGVector<char> l(lhs.begin(), lhs.end());
+			const SGVector<char> r(rhs.begin(), rhs.end());
```
gf712, line 123: this will have to be cleaned up in the future, this copy is not necessary. But can be done in another pr

src/shogun/base/base_types.h
karlnapf: Not sure if this is of any help: https://github.com/shogun-toolbox/shogun/blob/develop/src/shogun/lib/any.h#L856

gf712: the problem is that it isn't cross platform, i.e. doesn't work with msvc

gf712:
@LiuYuHui can I merge this?

LiuYuHui:
> @LiuYuHui can I merge this?

yes :D

gf712:
Thanks, this is great :)

PR #5111, opened 2020-08-10
Make Machine class stateless

gf712:
@LiuYuHui's main GSoC project. Machine class becomes stateless wrt Features and Labels which means that the user has to provide features and labels when fitting a Machine. This is essentially done by adding the notion of (Non)Parametric Machines.


karlnapf:
Nice to see this coming in :)

gf712:
@LiuYuHui sorry for the delays, I was meant to do this a while ago. I rebased your branch now and it should be good to merge soon. Feel free to review in case I messed up the rebase

src/shogun/machine/Composite.h
```diff
@@ -78,7 +78,7 @@ namespace shogun
 				}, v.second);
 			}
 			m_ensemble_machine->train(current_data, labs);
-			return m_ensemble_machine;
+			return true;
```
gf712, line 81: @LiuYuHui I had to change the return type, since that is the type of Machine::train. In future it might be worth returning the Machine object.

tests/unit/machine/EnsembleMachine_unittest.cc
```diff
@@ -77,13 +78,15 @@ TEST(combinate_composite_and_pipeline, train)
...
-	                		   		->apply_multiclass(test_feats);
+	auto machine = pipeline->over(std::make_shared<NormOne>())
+		 	        	->composite()
+			 	        	->over(std::make_shared<MulticlassLibLinear>())
+            				->over(std::make_shared<MulticlassOCAS>())
+                        	->then(std::make_shared<MeanRule>());
+	
+	machine->train(train_feats, train_labels);
+
+	auto pred = machine->apply_multiclass(test_feats);
```
gf712, line 89: At some point Machines should return a pointer to themselves after calling train (like in sklearn) so that we can use a builder pattern in situations like this

PR #5112, opened 2020-08-12
Add RegressionLabels Encoder

LiuYuHui:


src/shogun/labels/RegressionLabelsEncoder.h
```diff
@@ -0,0 +1,64 @@
...
+#include <shogun/labels/RegressionLabels.h>
+#include <shogun/lib/SGVector.h>
+#include <shogun/util/traits.h>
+
+namespace shogun
+{
+	/** @brief Implements a reversible mapping from
+	 * any form of labels to RegressionLabels.
+	 */
+	class RegressionLabelsEncoder : public LabelEncoder
```
gf712:
```suggestion
	class RegressionLabelEncoder : public LabelEncoder
```

src/shogun/labels/RegressionLabelsEncoder.h
```diff
@@ -0,0 +1,64 @@
+/*
+ * This software is distributed under BSD 3-clause license (see LICENSE file).
+ *
+ * Authors: Yuhui Liu
+ *
+ */
+#ifndef _RegressionLabelsEncoder__H__
+#define _RegressionLabelsEncoder__H__
```
gf712:
```suggestion
#define _REGRESSIONLABELENCODER__H__
```
could you capitalise these to stay consistent please?

gf712 (changes requested):
Looks good! Just some minor changes. I think in the future we should consider adding some optimisation where we check the type of the labels being encoded in order to avoid create a new Labels object for the same type

PR #4793, opened 2019-11-09
fix #4710

vigsterkr:

PR #5118, opened 2020-09-02
Auto dispatch create methods in python

### LINKED ISSUE #4998: auto dispatch create methods in python
We have `get` for python, which automatically returns the correct type, i.e. `get('my_vector')` and `get('my_int')` return a numpy array and an int respectively.

Now, why don't we do the same thing for creating shogun objects via factories? (In Python)

`create_kernel('GaussianKernel')` -> `create('GaussianKernel')`

That should be possible using exactly the same mechanics as in `get`, or do I get this wrong? @gf712 Maybe you can provide pointers for a student to do this

ghost:
This fixes #4998 using the same technique that @gf712 suggested. I created a visitor for interface classes and changed the `create` method in `class_list` to accept the visitor. Then used this visitor in swig to transform the interface object to PyObject.


src/shogun/base/base_types.h
```diff
@@ -113,8 +94,44 @@ namespace shogun
...
 		EvaluationResult, MulticlassStrategy, NeuralLayer,
 		SplittingStrategy, LikelihoodModel, MeanFunction,
-		DifferentiableFunction, Inference, LossFunction,
-		Tokenizer>;
+		Inference, DifferentiableFunction, LossFunction,
+		Tokenizer, CombinationRule, KernelNormalizer, Transformer,
+		MachineEvaluation, StructuredModel, FactorType, ParameterObserver,
+		Distribution, GaussianProcess, Alphabet>;
+
+	namespace types_detail
```
vigsterkr, line 102: awesome!

src/interfaces/python/factory_python.i
```diff
@@ -0,0 +1,184 @@
...
+// but this is faster and should not cause problems
+#define SHOGUN_GET_SWIG_TYPE(name)                                             \
+	SWIGTYPE_p_std__shared_ptrT_shogun__##name##_t
+
+namespace shogun
+{
+	class ShogunInterfaceToPyObject : public InterfaceTypeVisitor
+	{
+	public:
+		virtual void on(std::shared_ptr<SGObject>* v)
```
gf712: minor thing, but now we are trying to always use `override` instead of overwriting virtual functions
```suggestion
		void on(std::shared_ptr<SGObject>* v) override
```

src/interfaces/python/factory_python.i
```diff
@@ -0,0 +1,184 @@
+%{
+#include <shogun/util/visitors/InterfaceTypeVisitor.h>
+
+// we can use:
+// SWIG_TypeQueryModule or SWIG_MangledTypeQueryModule or SWIG_Python_TypeQuery
+// to query the swig_type in run time and be less implementation dependent maybe?
+// but this is faster and should not cause problems
+#define SHOGUN_GET_SWIG_TYPE(name)                                             \
+	SWIGTYPE_p_std__shared_ptrT_shogun__##name##_t
```
gf712: I am wondering if this can be generalised to the other swig interfaces?

src/interfaces/python/factory_python.i
```diff
@@ -0,0 +1,184 @@
...
+		PyObject* pyobj()
+		{
+			// this should never be true since "create" throws on error
+			if (!m_pyobj)
+			{
+				m_pyobj = SWIG_Python_NewPointerObj(
+					nullptr, nullptr, SHOGUN_GET_SWIG_TYPE(SGObject),
+					SWIG_POINTER_OWN);
+			}
+			return m_pyobj;
```
gf712: even if it *should* not be possible to reach this, could you replace this with an exception? The worst thing that can happen is segfault in Python, because then the interpreter crashes. An exception can at least be handled gracefully

gf712: ah nvm, I see that you handle that below

src/interfaces/python/factory_python.i
ghost: override FTW

src/interfaces/python/factory_python.i
ghost: I'm assuming you mean the whole file. I would guess so. `factory_python` is a horrible name for that regard. Any name suggestions?
Edit: I will call it factory_visitors.i for now

src/interfaces/python/factory_python.i
ghost: In this case I return a python null object. I will change the function call to `SWIG_Py_Void()` to be more clear and add a comment.

theartful:
I can see this visitor being integrated into the GetterVisitor (making AnyVisitor inherit from InterfaceTypeVisitor from example, and defaulting the new methods to `on(SGObject)`).
Does the parameter framework currently support "getting" objects of base classes?
If no, I think this structure should be good for now.

src/interfaces/python/factory_python.i
ghost: Scratch that. Throwing is better.

src/interfaces/python/factory_python.i
gf712: yes, that's fine. And yes, I meant if you could for example template `ShogunInterfaceToPyObject`, and the template determines the interface object type, e.g. `ShogunInterfaceFactory<PyObject>`.

src/interfaces/python/factory_python.i
gf712: yes, agreed, throwing would provide more information. Getting `None` from a factory might be odd for a user

gf712:
> Does the parameter framework currently support "getting" objects of base classes?

Not yet, it's been worked on in #4793

src/interfaces/python/factory_python.i
ghost: I have two reasons not to:
1. Added complexity without any other use case at the moment.
2. After reading "GetterVisitor" I don't see any place for this visitor in the future (and maybe not for InterfaceClassVisitor either). Once the parameter framework and AnyVisitor support base types we can get rid of both new visitors, and use GetterVisitor directly, no?

ghost: Since R is working now, I should make this generic as you suggested :"D

gf712: I think so, that would be great! Then it could be easier to add other languages?

ghost: Yeah. This should work with octave as well.

src/shogun/util/visitors/InterfaceTypeVisitor.h
```diff
@@ -0,0 +1,74 @@
...
+		virtual void on(std::shared_ptr<CombinationRule>*) = 0;
+		virtual void on(std::shared_ptr<KernelNormalizer>*) = 0;
+		virtual void on(std::shared_ptr<Transformer>*) = 0;
+		virtual void on(std::shared_ptr<MachineEvaluation>*) = 0;
+		virtual void on(std::shared_ptr<StructuredModel>*) = 0;
+		virtual void on(std::shared_ptr<FactorType>*) = 0;
+		virtual void on(std::shared_ptr<ParameterObserver>*) = 0;
+		virtual void on(std::shared_ptr<Distribution>*) = 0;
+		virtual void on(std::shared_ptr<GaussianProcess>*) = 0;
+		virtual void on(std::shared_ptr<Alphabet>*) = 0;
```
gf712, line 47: I wonder if there could be a way to automatically generate this? Then again the base types probably do not change that often

ghost: We can do type erasure as follows
```C++
virtual void on(std::shared_ptr<void>*) = 0;
```
and then try to `dynamic_cast` for all base types one by one (with generic for loop on types) and dispatch to a generic `on`.
```C++
template <typename BaseClass>
void on(std::shared_ptr<BaseClass>*);
```
 But this is slower.

ghost: We can alaso manually construct the virtual table of `on` methods using a polymorphic lambda. So we would have something like
```C++
// vtable is constructed from one polymorphic lambda
std::tuple<std::function<void(std::shared_ptr<BaseClasses>*)>...> vtable;

template <typename BaseClass>
void on(std::shared_ptr<BaseClass>* ptr)
{
   std::get<TypeToIndex<BaseClass>::value>(this->vtable) (ptr);
}
```
This should be just as efficient as multiple virtual methods, but it requires complex code.

gf712: I think the manual vtable is similar to what we have in Any and use in SGObject
https://github.com/shogun-toolbox/shogun/blob/9b8d856971af5a295dd6ad70623ae45647a6334c/src/shogun/base/SGObject.h#L768
I wonder if the same logic could be used. I guess it would only be possible if we add another base class between SGObject and the "base" classes, where it would register types to be used (somehow) in the interfaces

gf712: The type erasure solution is kind of what we have/had in SGObject but we are moving away from that and just use visitor patterns.

PR #5120, opened 2020-09-05
Return Machine with Composite::train when swig is used

### LINKED ISSUE #5119: Composite meta example causes memory leak in python
I noticed while running the meta examples that the output of "composite.py" is
```
[09/04/20 23:56:07 info] Loaded labels from CSVFile("/home/theartful/Programming/shogun/build/examples/meta/data/classifier_4class_2d_linear_labels_train.dat") as MulticlassLabels
[09/04/20 23:56:07 info] Loaded labels from CSVFile("/home/theartful/Programming/shogun/build/examples/meta/data/classifier_4class_2d_linear_labels_test.dat") as MulticlassLabels
[09/04/20 23:56:07 info] optimization finished, #iter = 295
swig/python detected a memory leak of type 'std::shared_ptr< EnsembleMachine > *', no destructor found.
```
It seems that swig is unable to delete it for some reason.

ghost:
fixes #5119

### DIFF
--- a/src/shogun/machine/Composite.h
+++ b/src/shogun/machine/Composite.h
@@ -56,7 +56,12 @@ namespace shogun
 			m_stages = std::forward<T>(stages);
 		}
 
-		std::shared_ptr<EnsembleMachine> train(
+#ifdef SWIG
+		std::shared_ptr<Machine>
+#else
+		std::shared_ptr<EnsembleMachine>
+#endif
+		train(
 		    const std::shared_ptr<Features>& data,
 		    const std::shared_ptr<Labels>& labs)
 		{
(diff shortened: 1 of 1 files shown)


src/shogun/machine/Composite.h
```diff
@@ -56,7 +56,12 @@ namespace shogun
 			m_stages = std::forward<T>(stages);
 		}
 
-		std::shared_ptr<EnsembleMachine> train(
+#ifdef SWIG
+		std::shared_ptr<Machine>
```
gf712: Shouldn’t it always return Machine?

ghost: I would prefer that, or possibly return nothing at all, and add another function to return the internal machine since it's mostly not needed.

ghost: what do you suggest?

gf712: I like the returning of the fitted object from train. But it should be of type Machine consistently. This is what scikit-learn does, and it works well imo

gf712:
Thanks!

PR #5122, opened 2020-09-07
Make R work again

ghost:
I added a script that fixes three problems:
1. The multiple inheritance issue: https://github.com/swig/swig/issues/1867
2. Garbage collection. Now if we force SGObject destructor to print to stdout and clear R's local variables we can see them  deleted. This did not work before. It was done by attaching a "finalizer" at the function that wraps pointers in `SEXP` (R objects). Originally this was only done for wrapped functions of constructors.
```R
rm(list=ls())
gc()
[09/07/20 09:40:14 info] SGObject destroyed (0x1b1fe9e0)
...
```
3. Handling memory allocation when casting https://github.com/swig/swig/issues/1660. This is more or less copy pasted with small modifications from the generated python bindings. If we only remove the assertion (that previously caused the error) and run valgrind, it will find lots of "definitely lost" memory leaks. Now no "definitely lost" memory leaks exist (although valgrind finds an error when attaching the library, but as far as I can tell it is not related to shogun + finds "possibly lost" similar to cpp example version).


vigsterkr:
@theartful awesome work!

@gf712 @karlnapf any objections to merge? the failures are unrelated....

gf712:
@theartful looks great!

examples/undocumented/python/library_fisher2x3.py
```diff
@@ -1,19 +0,0 @@
...
-
-x=array([[20.0,15,15],[10,20,20]])
-y=array([[21.0,21,18],[19,19,22]])
-z=array([[15.0,27,18],[32,5,23]])
-
-
-parameter_list = [[x,concatenate((x,y,z),1)]]
-
-def library_fisher2x3 (table, tables):
-	pval=Statistics_fishers_exact_test_for_2x3_table(table)
```
ghost, line 13: I removed Statistics and Math classes from swig since there is name clashing in R with these classes.
Now this example failed which only calls a function from the statistics class. I decided to remove the example. If this isn't reasonable I will bring it back + the Statistics class in swig.

gf712: I think it is reasonable to remove this!

gf712:
@karlnapf @vigsterkr can I merge this so that for now we have a functional R interface?

gf712:
@theartful thanks for this!

theartful:
I think there was a Jupyter notebook failing that used "Statistics/Math" classes. I will make another PR to either fix this or readd them to swig,

PR #5126, opened 2020-09-19
Fixed shogun import issue #4990

AllenRLopez:
Changed shogun import statement and updated lines of code where methods from the shogun library were used.

### DIFF
--- a/examples/undocumented/python/kernel_distantsegments.py
+++ b/examples/undocumented/python/kernel_distantsegments.py
@@ -7,12 +7,12 @@ testdat = lm.load_dna('../data/fm_test_dna.dat')
 parameter_list = [[traindat,testdat,5,5],[traindat,testdat,6,6]]
 
 def kernel_distantsegments (fm_train_dna=traindat,fm_test_dna=testdat,delta=5, theta=5):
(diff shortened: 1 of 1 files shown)


examples/undocumented/python/kernel_distantsegments.py
```diff
@@ -7,12 +7,12 @@
 parameter_list = [[traindat,testdat,5,5],[traindat,testdat,6,6]]
 
 def kernel_distantsegments (fm_train_dna=traindat,fm_test_dna=testdat,delta=5, theta=5):
-	from shogun import DistantSegmentsKernel
+	import shogun as sg
 
 	feats_train=sg.create_string_features(fm_train_dna, sg.DNA)
```
AllenRLopez, line 12: @gf712  @karlnapf on lines 12 & 13 before my commit, they were already using "sg."  without importing shogun as sg until I did. Was it a bug? I'm having trouble figuring out why that was working before. If you could help explain, I'd be very grateful, thanks.

gf712: potentially, not sure if the test suite skips this test

gf712:
Thanks for this!

PR #5127, opened 2020-09-23
[WIP] Add Bayesian Optimization

LiuYuHui:
- [ ] Add parameter range 
``` cpp
//Real value
obj->put("C1", Param{1.0, 2.0});
//Categorical value
obj->put("kernel", Param{gaussian_kernel, const_kernel});
```

- [ ] Add BayessianOptimization class


src/shogun/base/AnyParameter.h
```diff
@@ -275,6 +280,7 @@ namespace shogun
 		std::shared_ptr<params::AutoInit> m_init_function;
 		std::function<std::optional<std::string>(Any)> m_constrain_function;
 		std::vector<std::function<void()>> m_callback_functions;
+		std::optional<ParameterRange> m_param_range;
```
LiuYuHui, line 283: the user call `put ` to set this value, and then we can use this informatation in bayesian optimization.

src/shogun/base/AnyParameter.h
```diff
@@ -213,7 +213,12 @@ namespace shogun
 		{
 			return m_properties;
 		}
-
+		using ParameterRange = std::variant<std::pair<Any, Any>, std::vector<Any>>;
```
LiuYuHui, line 216: @vigsterkr  can I assume that type of std::pair<Any, Any> is float64_t/int and std::vector<Any> is kernel?

gf712:
@LiuYuHui do you need the additional `m_param_range` field? Couldn't this just be type erased and stored in the map of Any in sgobject?

LiuYuHui:
> @LiuYuHui do you need the additional `m_param_range` field? Couldn't this just be type erased and stored in the map of Any in sgobject?

you mean like this make_any(std::pair<float64_t, float64_t>)?

LiuYuHui:
I will continue to work on it when I am free.

PR #5129, opened 2020-10-01
Convert pythonic imports for all python/graphical

### LINKED ISSUE #4990: pythonic imports of shogun

sztang:
Combed through and updated the python/graphical folder; partially addresses #4990 and fixed some minor typos/print statements. Seems like we're mostly doing Python2 prints (i.e. no parenthesis), so went ahead and standardized that.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "python.pythonPath": "/Users/SIMON/anaconda3/bin/python"
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,0 @@
-{
-    "python.pythonPath": "/Users/SIMON/anaconda3/bin/python"
-}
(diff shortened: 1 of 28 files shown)


gf712:
Hi thanks for this! We actually do not run CI jobs on these examples (these are usually quite low priority when refactoring and are probably mostly outdated).

sztang:
Figures, I see why everything's still in Python 2. Thanks for merging anyway, new to PRs and learned a lot from it. Will look around for other spots where I can help cleanup.

Issue #5058, opened 2020-06-07
Write a utility function to make a shallow copy of an SGObject


gf712:
Usually `SGObject`s are stored in `shared_ptr`, so when we perform a copy we just increment the control block refcount. We have a clone method to perform deep cloning, but this means we might be copying features, labels and weights, which is not always needed. This can be avoided by passing the respective [ParameterProperties](https://github.com/shogun-toolbox/shogun/blob/78cfc5f4ef6daf196819f900234be7d5455bc3a1/src/shogun/base/AnyParameter.h#L21) to the `make_clone` function, so it would be good to have `make_shallow_copy` free function that sets these flags automatically (something like `ParameterProperties::ALL ^ ParameterProperties::MODEL ^ ParameterProperties::READONLY`), and then just have something like `make_shallow_copy(obj)`.

LiuYuHui:
you mean like this?
``` cpp
template<typename T>
std::shared_ptr<SGObject> make_shallow_copy(std::shared_ptr<T> obj){
   return obj->clone(ParameterProperties::ALL 
              ^ ParameterProperties::MODEL ^ ParameterProperties::READONLY);
}

gf712:
yup, and then find places where you can use it. You could also use the `make_clone` function instead, but it would be the same

gf712:
https://github.com/shogun-toolbox/shogun/blob/78cfc5f4ef6daf196819f900234be7d5455bc3a1/src/shogun/multiclass/tree/RelaxedTree.cpp#L299

gf712:
@karlnapf are those the flags you would expect a shallow copy to use?

karlnapf:
yes, see in the xvalidation code.
But I would name it differently, it is not really a "shallow" copy, as some of the subobjects would be cloned. It is more like .... mmh not sure what a good name would be. But I think it is good to have this function (shouldnt be part of swig though)

5aumy4:
I am new to open source development.
Can I do any help?

yiransii:
Hello! I looked at the source code, and SGObject is still missing the make_shallow_copy() method. I would like to work on this!

yiransii:
^ following the previous comment...

I am trying to run some python example, but I was not able to find shogun.py, as listed in https://github.com/shogun-toolbox/shogun/blob/develop/doc/readme/INTERFACES.md. 
Can anyone help me with this? 

thanks!

LiuYuHui:
> ^ following the previous comment...
[...]

PR #5147, opened 2021-01-05
Add sparse copy

jonpsy:
#5058 

A free function which has default flags as mentioned in the issue. Pretty much same as make_clone, but it "motivates" sparse copying.

SGMatrix, and Features.h seems to be potential contenders for this sparse_copy OR make_clone(for entire deep copy), we can handle them both in this PR? @gf712 

My thought process is to use make_clone for deep copy **only** and sparse_copy for if we want to copy only certain component(and therefore I've changed it in RelaxedTree).


jonpsy:
Re-open when devs are back.

PR #5149, opened 2021-01-07
Chain put

jonpsy:
#5047

I'm using ```shared_from_this()``` , which forces users to create objects using shared pointers. 

In RandomMixin_unittest.cc, https://github.com/shogun-toolbox/shogun/blob/9b8d856971af5a295dd6ad70623ae45647a6334c/tests/unit/mathematics/RandomMixin_unittest.cc#L106

Is it necessary to use std::make_unique? Can't it be make_shared, that'll solve the test case error in this file.


jonpsy:
Re-open when devs are back.

PR #5190, opened 2023-12-09
First shot to fix the reported median bug.

iglesias:
Hi!

It is still work in progress. Just in case, apologies for opening the pull request already.


lisitsyn:
LGTM!

iglesias:
It doesn't seem to be reproducible. At least to me 🤷‍♂️

https://godbolt.org/z/P4b1sbcsW

danlark1:
You should put the define before includes or in the command line, see

https://godbolt.org/z/Tx8rGrqoa

danlark1:
Or, for example, without any randomization

https://godbolt.org/z/vn5e7eW7M

iglesias:
Oh! Thanks a lot for taking a look and getting back!

